PR #31, opened 2017-02-13
Enable C++14 for Qt Creator on OS X

marvrez:
Seems like there's a glitch where the standard libraries aren't included when using the c++14 config flag. This gets fixed by adding the following line to the .pro file:```
QMAKE_CXXFLAGS += -stdlib=libc++ ```

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -9,6 +9,8 @@ CONFIG  += communi
 COMMUNI += core model util
 CONFIG  += c++14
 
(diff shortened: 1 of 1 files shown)


fourtf:
Are you sure that this is needed? I was able to build it on macOS 10.11 just fine.

marvrez:
Hmm, it works now for some reason when I try to rebuild the project without the added line. Nvm then, lol.

fourtf:
Who understands c++ compilers ¯\_(ツ)_/¯

PR #36, opened 2017-04-13
windowmanager destructor free _mainwindow, fix save - remove const in

hemirt:
settingsPath that is being written to from ptree tree, no need for explicit initialization of _windowMutex - it is done by default already


hemirt:
there might also be some thread safety issues

PR #39, opened 2017-04-19
Compile before push DansGame

Bur0k:
DansGame


widgets/accountpopup.cpp
```diff
@@ -1,6 +1,6 @@
 #include "widgets/accountpopup.h"
 #include "channel.h"
-#include "ui_userpopup.h"
+#include "ui_userpopupform.h"
```
Confuseh: it should be `ui_accountpopupform.h` for both

PR #40, opened 2017-04-19
Make runnable again

Bur0k:
Initialize them when they are needed, not on startup. This is THREAD-SAFE since C++11.

Chatterino2 now runs on Linux again


fourtf:
it compiled for windows and I didn't check after I refactored everything on my linux vm ¯\_(ツ)_/¯

PR #47, opened 2017-06-29
fix wrong variable name, openChannelNameDialog on ctrl+t, give focus to new split

hemirt:
the _currentValue should fix scrollbar (atleast it does for me), without it cannot scroll with mousewheel plus dragging the scrollbar with mouse is kinda bugged too


pajlada:
Looks good to me :1st_place_medal:

Issue #72, opened 2017-07-03
Arrow up/Arrow down in chat text input should browse through previously written messages


pajlada:

pajlada:
Fixed in 23eb42013be63c90e1ea4a0a1e959603b4505592

Issue #71, opened 2017-07-03
Ctrl+Enter to send message but not clear the input box


pajlada:

pajlada:
Implemented in 23eb42013be63c90e1ea4a0a1e959603b4505592

PR #80, opened 2017-07-23
Implemented issues #71 and #72

Cranken:
This implements the feature of Issue #71 .


Cranken:
Now also implements #72

Issue #52, opened 2017-07-02
Parse links, make them clickable


pajlada:

pajlada:
Majority of feature fixed by @Cranken in b3b89e80196520035fa3a346220c29fd7df40d8b
Few quirks remaining:
`google.com` is not a link (regex probably needs updating)

pajlada:
Fully fixed by @Cranken in f787d3f4644bf9aed691fd6c5e30e11ee08b231a

PR #81, opened 2017-07-25
Made links open browser

Cranken:
Added link parsing and opening the browser as requested in #52 .
The regex only works on links that are standing alone and are not part of other words e.g. "Kaphttps://google.com" doesn't work.

Oh and also, I have no clue how to only pull request a single commit forsenE .


pajlada:
What you would normally do is make changes in different feature-branches, or just made sure to make your forked repo up to date with chatterino2 beforehand. I'll manually merge your latest commit. Thanks a lot for the contribution!

Apparently I'm not a git expert either, and I couldn't figure out a way to only rebase one of your commits, so I just got a patch from that commit and applied it manually

Issue #50, opened 2017-07-02
Highlights


pajlada:
Highlight message when your username/custom highlight is mentioned in a message

Cranken:
Started working on this, do we want the custom sounds back? Was this feature even used?
And also, should the sound play/taskbar also flash when the custom highlight is triggered or only when the username is mentioned?

pajlada:
Custom sounds was used a decent amount I think. Yeah I think custom highlights should trigger sound/taskbar flashes too

- [ ] Highlight message in chat
- [ ] Option to play sound when any message is highlighted
   - [ ] Option to choose a custom sound to play
- [ ] Option to flash taskbar when any message is highlighted

Personally I would like per-custom-highlight options so that I can add a highlight that does highlight in chat but doesn't play a sound for example, so something like that might be good to think about already.

Cranken:
How should we implement the taskbar flash feature on linux? I never really used Linux much but I think there's a different taskbar on pretty much every distro. QApplication only gives a solution for Windows/MacOS with [QApplication::alert()](https://doc.qt.io/qt-5/qapplication.html#alert)

pajlada:
Try [QApplication::alert()](https://doc.qt.io/qt-5/qapplication.html#alert) and I'll see what happens on Linux

pajlada:
Base highlighting implemented

PR #86, opened 2017-07-30
Added Highlights

Cranken:
Added Highlights #50 , custom sound, custom options for every keyword.

Hope I squashed it right.


pajlada:
Looks good 👍 I want to move away from the old settings system to start using the `pajlada/settings` library in the future

PR #87, opened 2017-07-31
Fixed regex triggering on everything.

Cranken:
Also cleaned up a bit and changed to http as default URL scheme as it's more compatible with every site.


pajlada:
:+1:

PR #91, opened 2017-08-02
[WIP] Support CMake build system

matijakevic:


matijakevic:
To make building even more easier and automated, qmake should be eliminated completely. Therefore we need to create our CMakeLists.txt for libcommuni library Core, Util and Model.

pajlada:
Looking good so far, still haven't tested.

I don't think a CMakeLists.txt for libcommuni should necessarily be required though. What about a FindLibCommuni.cmake file in the cmake folder specifying which headers and sources that needs to be compiled together with Chatterino2?

PR #103, opened 2017-08-25
Quality options for Streamlink

Confuseh:
Implementation is bit scuffed but it works.


bennyhillthebest:
Could it be possible for the user to append options to the Streamlink command? I would like to use mpv instead of VLC with the option "-p mpv"

Confuseh:
probably yeah, ill get on that pajaW

Issue #93, opened 2017-08-04
Changing font should also change chat input font


pajlada:
- [x] Font when creating the Chat Input Widget should be fetched from the Font Manager
 - [ ] When the Current Font is updated in the Font Manager, the Chat Input Widget needs to update its font

kicklogicout:
I pretty much have this one done but I need to figure out why exactly the input text font is not changing at the same time as the font for the messages in the view do. Any pointers? Something to do with the FocusPolicy on ResizingTextEdit perhaps?

![fx6tymh - imgur](https://user-images.githubusercontent.com/15697718/29837526-475be098-8cf0-11e7-83c8-10a62f27141f.gif)

PR #107, opened 2017-09-12
Changing font should also change chat input font (#93)

kicklogicout:
Could do with a review first before merging.


fourtf:
make a new pr whenever you feel like it's ready for the review

PR #117, opened 2017-09-22
Added some functionality to the accountPopup

Cranken:
This adds most of the original buttons from Chatterino 1 to the accountPopup. 
Also adds a user blacklist for highlights.


src/settingsmanager.cpp
```diff
@@ -35,6 +35,7 @@ SettingsManager::SettingsManager()
     , pathHighlightSound(this->settingsItems, "pathHighlightSound", "qrc:/sounds/ping2.wav")
     , highlightProperties(this->settingsItems, "highlightProperties",
                           QMap<QString, QPair<bool, bool>>())
+    , blacklistedUsers(_settingsItems, "blacklistedUsers", "")
```
fourtf: The name _settingsItems was changed to settingsItems in a recent commit

src/settingsmanager.hpp
```diff
@@ -49,6 +49,7 @@ class SettingsManager : public QObject
     Setting<bool> customHighlightSound;
     Setting<QString> pathHighlightSound;
     Setting<QMap<QString, QPair<bool, bool>>> highlightProperties;
+    Setting<QString> blacklistedUsers;
```
fourtf: The name `blacklistedUsers` is not the proper name as it is only used for ignoring highlights from certain users. I think `highlightUserBlacklist` or something along the lines would be a better name.

src/twitch/twitchmessagebuilder.cpp
```diff
@@ -397,13 +397,16 @@ void TwitchMessageBuilder::parseHighlights()
         bool alert;
     };
 
+    QStringList blackList = settings.blacklistedUsers.getnonConst().split("\n",QString::SkipEmptyParts);
```
fourtf: get() should be used instead of getnonConst() if possible

src/twitch/twitchuser.cpp
```diff
@@ -22,6 +23,16 @@ const QString &TwitchUser::getOAuthToken() const
     return this->_oauthToken;
 }
 
+const QString &TwitchUser::getUserId() const
+{
+    return this->_userId;
+}
+
+void TwitchUser::setUserId(QString id)
```
fourtf: minor issue: function should accept `const QString &id` for consistency.

PR #124, opened 2017-10-01
ImageLoader

hemirt:
push emote requests into a dedicated thread and the requests are done synchronously


hemirt:
dont merge this yet, gonna work on a better solution

PR #129, opened 2017-10-11
fix image loader scrollbar crash

hemirt:


src/messages/imageloadermanager.cpp
```diff
@@ -80,9 +83,10 @@ void ImageLoaderWorker::handleLoad(chatterino::messages::LazyLoadedImage *lli, Q
     }
 
     lli->emoteManager.incGeneration();
-    lli->windowManager.layoutVisibleChatWidgets();
+    //lli->windowManager.layoutVisibleChatWidgets();
```
pajlada: Just remove this line, it's useless to leave it commented out when we've moved it elsewhere.

src/messages/imageloadermanager.hpp
```diff
@@ -15,6 +15,8 @@ class ImageLoaderWorker : public QObject
 public slots:
     void handleRequest(chatterino::messages::LazyLoadedImage *lli, QNetworkAccessManager *nam);
     void handleLoad(LazyLoadedImage *lli, QNetworkReply *reply);
+signals:
```
pajlada: I think clang-format needs to be run here, or just add an empty line above `signals:`

PR #134, opened 2017-10-25
Networkmanager

hemirt:
use NetworkManager for all network related tasks


hemirt:
take a look around & tell me what should be done differently or something, i'd like another set of eyes

hemirt:
pajlada changed the api to more oo style instead of the (maybe unclear) functional (? thats how i'd call it) style

PR #135, opened 2017-10-26
Use user's font in chat input

vanolpfan:
https://github.com/fourtf/chatterino2/issues/93 

@fourtf  👉 👌
@pajlada  👉 👌


pajlada:
Does not completely fix #93 - if you change the font mid-way through the font is not updated. This is 50% of the way though, so I have accepted your pull request!

vanolpfan:
All other fonts doesn't update as well. I suppose there should be some global layout update on font change, so idk.

PR #224, opened 2018-01-11
Should fix #190

### LINKED ISSUE #190: Menu buttons on OSX become highlighted when setVisible is done
screen shot 2018-01-01 at 18 48 57

Here's how it looks like on a channel.

Cranken:
I have no way to confirm if it indeed works on Mac but according to the docs it should work.
May need to get @coral or someone with a Mac to test it.


nuuls:
Tested it on mac and it does not fix the issue.

PR #277, opened 2018-02-13
Fix duplicats in message history

Chronophylos:


anibyl:
Twitch IRC doesn't prevent sending same messages from users (even from the same user).

Chronophylos:
It's about the message history. This has nothing to do with twitch

PR #279, opened 2018-02-18
actually display channel ffz emotes as channel ffz emotes

Ckath:
not entirely sure why the whole thing with FillInFFZEmoteData is the way it is so not sure if this is the best solution, but at least now a Channel FFZ Emote is a Channel FFZ Emote again instead of giving everything Global FFZ Emote on calling that function


pajlada:
works, thank you!

PR #306, opened 2018-04-04
correct install instructions for Ubuntu/Debian

Jarel1337:


pajlada:
@Jarel1337 What distro/version did you use for these updated instructions?

Jarel1337:
@pajlada It was Ubuntu Mate 18.04, I don't know if the qtsvg lib is preinstalled on normal ubuntu

pajlada:
@Jarel1337 I downloaded Ubuntu Mate 18.04 myself and gave it a try, and your instructions were correct. It seems that with Ubuntu 18.04 using just `apt` is preferred over `apt-get`, and the svg library was indeed missing. Instead of replacing the existing Ubuntu 16.04 installation instructions I added a new set of instructions here: https://github.com/fourtf/chatterino2#ubuntu-1804
Let me know if those instructions are lacking.

PR #318, opened 2018-04-07
Now highlights the text in the channel change and tab rename dialogue.

Cranken:
Something missing from chatterino 1.


pajlada:
Works great 👍

PR #342, opened 2018-04-12
fix copying <br>Global/Channel ffz emote after copying a ffz emote

hemirt:
when copying FFZ emotes such as "4House ZreknarF Keepo"
the name of the emote ( Image::getName() ) returns the "code + \<br>Global/Channel ffz emote" instead of just the code of the emote
-> "4House\<br/>Channel FFZ Emote ZreknarF\<br/>Global FFZ Emote Keepo "
this fixes correct tooltiping the emote
(and is now in sync with bttv and twitch emotes in the same file)


pajlada:
👍

PR #384, opened 2018-05-15
Added the functionality of having logs in custom folders

apa420:
As well as resetting the custom path to default. ("BUG" Will create folder only if chatterino is restarted)


hemirt:
when you change the log folder, it (as said by apa420) doesnt create the new Whispers, Mentions, Channels/channels folders, and still keeps on writing to the old files (**not sure** if it creates new folders for new splits with chats, so there might be a seperation to channels opened before you changed it and after you changed it), as said after restart uses the new custom one

PR #420, opened 2018-05-24
Changed bit badge overlay text to be consistent with Twitch

LajamerrMittesdine:
Currently it uses "Twitch Bits(x)". Changed to "cheer x" to be consistent with Twitch text overlay.


fourtf:
can you change it to "Twitch Cheer" so it is the same style as all the other tooltips

LajamerrMittesdine:
Have updated to be consistent with Chatterino text overlay design.

PR #443, opened 2018-06-04
Update macOS instructions

Yardanico:
You can easily install qt with brew, so last 3 entries are taken from linux instructions.
Tested on macOS High Sierra (but should work on all macOS versions supported by brew/qt)


pajlada:
If a mac user could try these instructions that would be nice

fourtf:
it should be mentioned that qt creator is "required" to develop chatterino but that you only need qt to build it

Yardanico:
@pajlada well, I'm on hackintosh, but I already installed everything so I can't try this on a clean system.

PR #447, opened 2018-06-05
RFC Compliant URL Matching

LajamerrMittesdine:
Implemented https://gist.github.com/dperini/729294 in C++

This makes URL Validation / Matching RFC compliant.

### DIFF
--- a/src/messages/messagebuilder.cpp
+++ b/src/messages/messagebuilder.cpp
@@ -65,7 +65,7 @@ QString MessageBuilder::matchLink(const QString &string)
     "(?:\\.(?:[1-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))"
     "|"
     // host name
-    "(?:(?:[a-z\\x{00a1}-\\x{ffff}0-9]-*)*[a-z\\x{00a1}-\\x{ffff}0-9]+)"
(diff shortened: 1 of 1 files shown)


fourtf:
This regex is very complicated and doesn't provide a lot of advantages over the current one. It also introduces a more edge cases for example only matching certain ip ranges. 

- [ ] The regex should be a lot simpler to allow other links to be added (e.g. spotify:track..., steam:...)
- [ ] Ip ranges should not be excluded
- [ ] Ftp links can not be opened nor copied properly (it adds http:// at the start).

I think it is easier to make minor changes to the current regex than to replace it with a 30 line long regex.

pajlada:
I like this regex, but would prefer it with a few changes:
 1. Remove the ip range exclusion (at least for local ranges). It's useless to be able to link 200.200.200.200 but not 192.168.1.1
 2. I don't see a reason to allow ftp links, but if they're allowed then more prefixes should be allowed imo (irc://, steam://) like the way it works in the old regex. As fourtf noted, links with another prefix than http(s) cannot be copied properly

LajamerrMittesdine:
Too many issues with the fundamental design of this specific pull request to support the necessary features.

Closing and will give it another shot.

PR #452, opened 2018-06-06
Make the 'add account' popup box go to foreground

DatGuy1:
Previously it just stayed behind. i only tested this in windows since I don't have any OSX/Linux vms but raise() should just work on them.

### DIFF
--- a/src/widgets/logindialog.cpp
+++ b/src/widgets/logindialog.cpp
@@ -1,7 +1,7 @@
 #include "widgets/logindialog.hpp"
 #include "common.hpp"
 #include "util/urlfetch.hpp"
-#if FMT_USE_WINDOWS_H
+#if USEWINSDK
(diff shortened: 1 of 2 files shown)


pajlada:
Tested it in Linux with the i3 window manager. Compiles and runs as expected. Should be tested on Windows before merging at least

DatGuy1:
Already tested on windows, at least with my build.

fourtf:
Please use the USEWINSDK define instead of FMT_USE_WINDOWS_H, then I will merge it.

DatGuy1:
Done. Also, there's probably a few other bugs like this, already found it in look->font select. when c2 is pretty much ready for public release I can go over "popup" widgets and add that

fourtf:
You have to use `#ifdef USEWINSDK` instead of `#if USEWINSDK`

PR #453, opened 2018-06-06
Default 'add split' to channel

DatGuy1:
Just a little UX fix


pajlada:
Looks good to me 👍

pajlada:
Tested and it works well, so I'm merging :+1:

PR #463, opened 2018-06-11
Add the option to manually update channel text.

LajamerrMittesdine:
This adds the option to manually update the channel text.

So that people can get fresh stats when hovering over a header split, without having to wait for it to update / restarting chatterino.

image


fourtf:
There are two issues that I see with this:
1. It adds unnecessary complexity to the ui by adding a feature that only a few users want.
2. It doesn't describe what it does properly, I would assume that most users won't figure out what is meant by "channel stats" and I don't see a proper way of describing it.

PR #497, opened 2018-06-22
Add room mode options to splitheader

DatGuy1:
Transferring a feature from Chatterino 1.

### DIFF
--- a/src/widgets/helper/splitheader.cpp
+++ b/src/widgets/helper/splitheader.cpp
@@ -14,6 +14,7 @@
 
 #include <QByteArray>
 #include <QDrag>
+#include <QInputDialog>
 #include <QMimeData>
 #include <QPainter>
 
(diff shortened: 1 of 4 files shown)


src/widgets/helper/splitheader.cpp
```diff
@@ -134,6 +143,70 @@ void SplitHeader::addDropdownItems(RippleEffectButton *)
...
+    QObject::connect(setsub, &QAction::triggered, this, [setsub, this]() {
+        QString sendcommand = "/subscribers";
+        if (!setsub->isChecked()) {
+            sendcommand.append("off");
+        };
+        this->split->getChannel().get()->sendMessage(sendcommand);
+    });
+
+    QAction *setemote = new QAction("Emote only", this);
+    this->setemote = setsub;
```
pajlada: This is wrongly copypasted as `setsub` instead of `setemote`

pajlada:
One more behavioural thing, if you're not a moderator I don't think the dropdown menu should open on click.
After that I think we're good to merge

DatGuy1:
So just to cast TwitchChannel?

fourtf:
- [ ] there are changes to completionmodel.cpp and .hope which, given the name of the pr, should not be included here. Furthermore you should not read the commands from file but use the commandmanager to access them from memory
- [ ] QTimer::singleShot shold take `this` as the second argument to remove the race condition that could potentially crash.
- [ ] variale names should be in lower "camelCase"

src/widgets/helper/splitheader.cpp
```diff
@@ -56,7 +57,18 @@ SplitHeader::SplitHeader(Split *_split)
 
         // mode button
         auto mode = layout.emplace<RippleEffectLabel>(this).assign(&this->modeButton);
+        this->addModeItems(mode.getElement());
 
+        QObject::connect(mode.getElement(), &RippleEffectLabel::clicked, this, [this] {
+            QTimer::singleShot(80, [&] {
```
pajlada: http://doc.qt.io/qt-5/qtimer.html#singleShot-6 Check out that function, you'll need to pass `this` as the context parameter

PR #524, opened 2018-06-25
Initial work of rewriting match link

LajamerrMittesdine:
This is just some initial work on rewriting the match hyperlink feature.


Yardanico:
:+1: No more highlights of things like "test.test" :joy: :lit: :100:

fourtf:
- The TLDs resource file should be delimited by a newline instead of | so it's not an 11k long line. Also makes it less of a git fiesta.
- The TDLs should be loaded from the resource file instead of being embedded into the source file.
- No temp files should be created by the script or they should at least be deleted afterwards.
- SFTP matching should be removed.

LajamerrMittesdine:
Did a little bit of changes.

Still working on eliminating the rest of the temp files.
Then working on changing messagebuilder.cpp so that it doesn't include the massive length of text inside of it.

LajamerrMittesdine:
Making a newer pull request. Since this one has conflicts.

PR #526, opened 2018-06-25
Fix for issue created with 0245bd5

ch-ems:
'window always on top' didn't have effect on launch.


pajlada:
This is an issue in my signals/settings libraries.

A settings connect automatically invokes the signal right after connection with its current value, whereas a normal signal connect doesn't. I should make that more clear/less easy to mess up.

ch-ems:
sorry, I should've looked into it,
should I close this PR then?

pajlada:
@ch-ems You can leave it, it's still valid. I just wanted to make it clear that the reason this happens in the first place is due to the settings/signal library being poorly designed

Issue #525, opened 2018-06-25
Whispers split header doesn't change colour on new whisper


DatGuy1:
Sorry, hit enter prematurely :)

I would send a video but it's self explanatory. The font colour of the name  doesn't change to white and the header of it stays grey too.

ch-ems:
adding [this flag](https://github.com/fourtf/chatterino2/blob/6375a902a0936e52a93233e28a21d42852d92f51/src/providers/twitch/ircmessagehandler.cpp#L219) only after we add the message to the whisper channel could fix this.

PR #530, opened 2018-06-26
Fix for minor issue with slowmode checkbox and #525

ch-ems:
should whispers containing highlight phrases be sent to the mentions channel though?
seems unnecessary.


fourtf:
Please don't reformat the readConnectionOnlyIDs as clang-format will just turn it to it's original state.

Issue #401, opened 2018-05-17
Support for different rooms in twitch channels


TETYYS:

TETYYS:
Rooms are being discontinued due to too little users

PR #531, opened 2018-06-27
Do not fetch recent messages and viewer list for chatrooms

nforro:
It is possible to connect to a [chatroom](https://dev.twitch.tv/docs/irc/chat-rooms/) using `chatrooms:<channel ID>:<room UUID>` as channel name.

But `https://tmi.twitch.tv/group/user/<channel name>/chatters` endpoint doesn't return anything,
and `https://tmi.twitch.tv/api/rooms/<room ID>/recent_messages` endpoint returns messages from the general chat of a channel, because `room-id` is the same for all chatrooms.

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -123,7 +123,10 @@ void TwitchChannel::setRoomID(const QString &_roomID)
 {
     this->roomID = _roomID;
     this->roomIDchanged.invoke();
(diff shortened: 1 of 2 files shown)


pajlada:
I would prefer to not implement any hacks like this since chatrooms are a planned feature #401 
@fourtf might have a different opinion

nforro:
Ok, I'm fine with that. Feel free to close this then.

fourtf:
Agreeing with @pajlada

PR #533, opened 2018-06-27
Initial rework of matchLink

LajamerrMittesdine:
This one is tested to work correctly but still needs some refactoring for performance reasons.

Though personally I didn't notice any issues with CPU usage or memory with the current state.


pajlada:
Copying fourtf's comment:

    The TLDs resource file should be delimited by a newline instead of | so it's not an 11k long line. Also makes it less of a git fiesta.
    The TDLs should be loaded from the resource file instead of being embedded into the source file.
    No temp files should be created by the script or they should at least be deleted afterwards.
    SFTP matching should be removed.

I will look at this within 2 days

LajamerrMittesdine:
Thank you.

>The TLDs resource file should be delimited by a newline instead of | so it's not an 11k long line. Also makes it less of a git fiesta.
Completed this.
>The TLDs should be loaded from the resource file instead of being embedded into the source file.
This is done. However the resource file is being created and opened everytime matchLink is called. Needs some refactoring work.
>No temp files should be created by the script or they should at least be deleted afterwards.
Did as much as I could here. No more tmp1, tmp2, etc.
>SFTP matching should be removed.
Has been removed.

PR #537, opened 2018-06-29
Fix issue on Windows not matching links correctly.

LajamerrMittesdine:
Fixed an issue where links were not being highlighted correctly because of improper parsing and formatting of the tlds.txt file.

Before code-fix
image

After code-fix
image

I assumed it was because it's only replacing the new line \n and not the return carriage \r

### DIFF
--- a/src/common/LinkParser.cpp
+++ b/src/common/LinkParser.cpp
@@ -17,7 +17,7 @@ LinkParser::LinkParser(const QString &unparsedString)
         t1.setCodec("UTF-8");
 
         // Read the TLDs in and replace the newlines with pipes
-        QString tldData = t1.readAll().replace("\n", "|");
+        QString tldData = t1.readAll().replace(QRegExp("[\\n]"), "|").remove(QRegExp("[\\s]"));
 
         const QString urlRegExp =
             "^"


fourtf:
On Windows a newline is \r\n not only \n.

Only \r was used by the original Mac OS.

LajamerrMittesdine:
Sorry, I might've been wrong on the reasoning. But it does fix the issue that was occurring for me on Windows.

LajamerrMittesdine:
Accidentally closed pull request. Reopening.

pajlada:
I think the correct solution is to leave to code as is (split on `\n`) and ensure that it's that character that's used as a newline in the tld.txt file. Or change to `\r\n` in the code if that's what's used in the tld.txt file

fourtf:
I would suggest just splitting the lines with the regex `\r?\n`

pajlada:
ye fourtf's solution is probably least friction @LajamerrMittesdine

LajamerrMittesdine:
`QString tldData = t1.readAll().replace(QRegExp("[\\n]"), "|").remove(QRegExp("[\\s]"));` is the only one that works for me 100% perfectly. Everyother solution has flaws. Not sure why.

Here's what happens when I try and test it with. `QString tldData = t1.readAll().replace(QRegExp("[\\r?\\n]"), "|");`
image

Here's what happens when I try and test it with. `QString tldData = t1.readAll().replace("\r?\n", "|");`
image

fourtf:
The replace function just matches the raw string `"\r?\n"` (so carriage return, question mark, newline) so you need to pass a regex object.

QRegExp however is deprecated so use QRegularExpression instead.

pajlada:
@LajamerrMittesdine yeah what fourt said, I just fixed it here ec4c2434552df33301c4354fba88989ba04ac5cc so you can see how it can be done

PR #548, opened 2018-07-01
Add viewing logs to UserInfoPopup

DatGuy1:


DatGuy1:
Should be fine to review

fourtf:
Some of these issues may also be present in code written by myself, however I still want them to be addressed in PR's.

- Use the NetworkRequest class instead of QNetworkRequest. (rest of the code uses it for performance reasons)
- Don't capture variables with = in the lambda, capture them explicitly instead to make it clear what is used. (make it clear what is captured)
- Don't stack multiple http requests in lambdas, create functions accordingly. (spaghetti)
- Use auto (when possible and reasonable). (reduces noise level of the code)
- Don't to use message boxes when showing errors, maybe use a label to show the error instead. (annoying UX)
- Instead of showing the results in a MessageBox show them in a ChannelView.
  - Create a Channel
  - Parse the irc with Communi::IrcMessage (.text in the cbenni.com api)
  - Use TwitchMessageBuilder with the IrcMessage to get a Message
  - Add the messages to the Channel
  - Create a Split and set the Channel to the newly created Channel with the log messages.

You may need to create the MessageBox in the gui thread using util::sendToThread when you switch to NetworkRequest.

If this is too much for you we can accept the PR after fixing the first 4 issues.

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    req.onError([this](int errorCode) {
+        this->getOverrustleLogs();
+        return true;
+    });
+
+    req.getJSON([this, channelName](QJsonObject &dataCbenni) {
+        QJsonValue before = dataCbenni.value("before");
+        for (int i = before.toArray().size() - 1; i >= 0; --i) {
+            int index = before.toArray().size() - 1 - i;
```
fourtf: Double reverse

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+            QString message = before[index]["text"].toString();
+            // Hacky way to fix the timestamp
+            message.insert(1, "historical=1;");
+            message.insert(1, QString("tmi-sent-ts=%10000;").arg(before[index]["time"].toInt()));
+            this->answer += message + "<br>";
```
fourtf: Create a `std::vector<MessagePtr>` here instead of storing the answer as a member.

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+        box->setAttribute(Qt::WA_DeleteOnClose);
+        box->show();
+        box->raise();
+        return true;
+    });
+
+    req.getJSON([this, channelName](QJsonObject &rustleData) {
+        if (rustleData.contains("lines")) {
+            QJsonArray messages = rustleData.value("lines").toArray();
+            for (auto i : messages) {
```
fourtf: Use a reference here so the object doesn't need to get copied.

`for (auto &i : messages)`

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    QString channelName = this->channelName_;
+
+    QString urlRustle("https://overrustlelogs.net/api/v1/stalk/" + channelName + "/" +
+                      this->userName + ".json?limit=500");
+
+    NetworkRequest req(urlRustle);
+    req.setCaller(QThread::currentThread());
+    req.onError([this, channelName](int errorCode) {
+        this->hide();
```
fourtf: Call `this->close();` instead of hide so the LogsPopup can get deleted.

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    req.getJSON([this, channelName](QJsonObject &rustleData) {
+        if (rustleData.contains("lines")) {
+            QJsonArray messages = rustleData.value("lines").toArray();
+            for (auto i : messages) {
+                QJsonObject singleMessage = i.toObject();
+                this->rustleTime =
+                    QDateTime::fromSecsSinceEpoch(singleMessage.value("timestamp").toInt()).time();
+                this->answer += QString("%1<br>").arg(singleMessage.value("text").toString());
+            }
```
fourtf: As above, instead of having a member to store the answer just create a `std::vector<MessagePtr>` here instead.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -79,6 +82,12 @@ UserInfoPopup::UserInfoPopup()
 
         user->addStretch(1);
 
+        QObject::connect(viewLogs.getElement(), &RippleEffectLabel::clicked, [this] {
+            LogsPopup *logs = new LogsPopup();
```
fourtf: Use auto here to remove code noise.

`auto logs = new LogsPopup();`

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+        this->hide();
+        QMessageBox *box = new QMessageBox(QMessageBox::Information, "Error getting logs",
+                                           "No logs could be found for channel " + channelName);
+        box->setAttribute(Qt::WA_DeleteOnClose);
+        box->show();
+        box->raise();
+        return true;
+    });
+
+    req.getJSON([this, channelName](QJsonObject &rustleData) {
```
fourtf: No need to call the variable rustleData since the name doesn't collide with anything else. Just call it data or something similar.

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+    req.execute();
+}
+
+void LogsPopup::getOverrustleLogs()
+{
+    this->usedLogviewer = false;
+
+    QString channelName = this->channelName_;
+
+    QString urlRustle("https://overrustlelogs.net/api/v1/stalk/" + channelName + "/" +
```
fourtf: Name doesn't collide with anything else so you can just call it `url`.

src/widgets/dialogs/LogsPopup.cpp
DatGuy1: Error: C2440: 'initializing': cannot convert from 'QJsonValueRef' to 'QJsonValueRef &'

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,142 @@
...
+    req.setCaller(QThread::currentThread());
+
+    req.onError([this](int errorCode) {
+        this->getOverrustleLogs();
+        return true;
+    });
+
+    req.getJSON([this, channelName](QJsonObject &dataCbenni) {
+        QJsonValue before = dataCbenni.value("before");
+        for (int i = before.toArray().size() - 1; i >= 0; --i) {
```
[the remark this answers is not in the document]
DatGuy1: Not sure if this is needed since I simplified this for loop

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+LogsPopup::LogsPopup()
+{
+    this->initLayout();
+    this->resize(400, 600);
+}
+
+void LogsPopup::initLayout()
+{
+    {
```
fourtf: no reason for the extra { }

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+        this->setLayout(layout);
+    }
+}
+
+void LogsPopup::setInfo(ChannelPtr channel, QString userName)
+{
+    this->channel_ = channel;
+    this->userName = userName;
+    this->setWindowTitle("Logs for " + this->userName + " in " + this->channel_->name +
+                         "'s history");
```
fourtf: `User's logs in #channel` would be more compact and brings the most important info to the start of the message.

src/widgets/dialogs/LogsPopup.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+class LogsPopup : public BaseWindow
+{
+public:
+    LogsPopup();
+
+    void setInfo(std::shared_ptr<Channel> channel, QString userName);
+
+private:
+    ChannelView *channelView = nullptr;
```
fourtf: __All__ private member variables should follow the pattern `lowerCamelCase_`

src/widgets/dialogs/LogsPopup.hpp
```diff
@@ -0,0 +1,32 @@
...
+    ChannelPtr channel_ = Channel::getEmpty();
+
+    QString userName;
+
+    bool usedLogviewer = true;
```
fourtf: This variable should be removed

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+    this->getLogviewerLogs();
+}
+
+void LogsPopup::setupView(std::vector<MessagePtr> messages)
```
fourtf: Pass the vector as a reference as it doesn't need to get copied then

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+        };
+        this->setupView(messages);
+    });
+
+    req.execute();
+}
+
+void LogsPopup::getOverrustleLogs()
+{
+    TwitchChannel *twitchChannel = dynamic_cast<TwitchChannel *>(this->channel_.get());
```
fourtf: Also check for nullptr here

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+            // Hacky way to fix the timestamp
+            message.insert(1, "historical=1;");
+            message.insert(1, QString("tmi-sent-ts=%10000;").arg(messageObject["time"].toInt()));
+
+            MessageParseArgs args;
+            auto ircMessage = Communi::IrcMessage::fromData(message.toUtf8(), nullptr);
+            auto privMsg = static_cast<Communi::IrcPrivateMessage *>(ircMessage);
+            TwitchMessageBuilder builder(logsChannel.get(), privMsg, args);
+            if (!builder.isIgnored()) {
```
fourtf: Don't check if the message is ignored since the user wants to see all the messages when he is looking at the logs of a person. (e.g. if they are a mod)

src/widgets/dialogs/LogsPopup.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void LogsPopup::getOverrustleLogs()
+{
+    TwitchChannel *twitchChannel = dynamic_cast<TwitchChannel *>(this->channel_.get());
+
+    QString channelName = twitchChannel->name;
+
+    QString url("https://overrustlelogs.net/api/v1/stalk/" + channelName + "/" + this->userName +
+                ".json?limit=500");
```
fourtf: Also use `.arg()` like in getLogviewerLogs()

src/widgets/dialogs/LogsPopup.cpp
fourtf: Rename the method to `setMessages` or something similar

LajamerrMittesdine:
It's insanely huge and bad discoloration.

image

Here's what the one from master build looks like.

image

LajamerrMittesdine:
In this case, it's cut off text.

image

LajamerrMittesdine:
Tested this for a bit. Everything looks fine to me functionality wise. Not sure if it's a limitation of cbenni but some channels aren't loading badges correctly in the logs.

PR #555, opened 2018-07-02
Fix bug with 'disabled users' textbox

DatGuy1:
Fix 553


fourtf:
This setting will be refactored soon. I'm not sure if it is worth to fix it when it won't be in the next version anyways.

pajlada:
Fixed in 03b7fae09ef3536b743f15ea33e252a277de9ea9

PR #557, opened 2018-07-02
Add option to make links lowercase

DatGuy1:
Seems like clang autoformatted some stuff

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -236,10 +236,13 @@ MessagePtr TwitchMessageBuilder::build()
                     link = Link();
                 } else {
                     if (app->settings->lowercaseLink) {
-                        QRegularExpression httpRegex("\\bhttps?://");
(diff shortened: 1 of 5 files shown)


fourtf:
The setting is supposed to only lowercase the domain name and not the entire url

fourtf:
Looks good but I haven't tested it yet

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -235,6 +235,19 @@ MessagePtr TwitchMessageBuilder::build()
                 if (linkString.isEmpty()) {
                     link = Link();
                 } else {
+                    if (app->settings->lowercaseLink) {
+                        QRegularExpression httpRegex("\\bhttps?://");
+                        QRegularExpression ftpRegex("\\bftps?://");
```
pajlada: These two regular expressions need `QRegularExpression::CaseInsensitiveOption` options added

pajlada:
I will yolo merge, since I think it will work combined with the upstream changes to link parsing

PR #558, opened 2018-07-03
last message indicator pattern

hemirt:
Let's you choose between ver (dotted) and solid pattern for the last message indicator via a QComboBox inside settings-look-messages and stores it in a setting
[http://doc.qt.io/qt-5/qt.html#BrushStyle-enum](Qt::BrushStyle) has more pattern styles, but I dont think any of these would be any good.
Can someone go over the actual box in the Settings-Look-Messages, it feels/looks a bit weird, and I dont know how to set the size correctly 
edit1: this is what i mean https://i.imgur.com/ZddeUU5.png


hemirt:
maybe in the future someone could make the line colored?

src/widgets/settingspages/AppearancePage.cpp
```diff
@@ -103,6 +103,30 @@ AppearancePage::AppearancePage()
                                              app->settings->showMessageLength));
 
         messages.append(this->createCheckBox(LAST_MSG, app->settings->showLastMessageIndicator));
+        {
+            auto *combo = new QComboBox(this);
+            combo->addItems({"Ver", "Solid"});
```
pajlada: Call it "Dotted" instead of "Ver"

hemirt:
renamed to Dotted, fixed incorrect values shown in the combo boxes for this (pattern) and collapse lines, they defaulted to first (pos 0) item in the box instead of the current value

PR #562, opened 2018-07-04
Adding option for bold @username

apa420:
Title
requested by lajamerr (I created the issue)

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -231,7 +231,7 @@ MessagePtr TwitchMessageBuilder::build()
                 QString linkString = this->matchLink(string);
                 auto fontStyle = FontStyle::ChatMedium;
 
-                if (string[0] == '@' && app->settings->usernameBold) {
+                if (string.startsWith('@') && app->settings->usernameBold) {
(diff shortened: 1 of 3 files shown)


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -230,6 +230,12 @@ MessagePtr TwitchMessageBuilder::build()
                 // Actually just text
                 QString linkString = this->matchLink(string);
 
+                if (string[0] == '@' && app->settings->usernameBold) {
```
fourtf: Check if the string has at least one character before accessing the first element.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -230,6 +230,12 @@ MessagePtr TwitchMessageBuilder::build()
...
+                    this->emplace<TextElement>(string, MessageElement::Text, textColor,
+                                               FontStyle::ChatMediumBold);
+                    string = "";
+                }
```
fourtf: Don't duplicate code for no reason. Instead add a FontStyle variable and use it in the emplace<TextElement> below.

`string = "";` is a terrible hack here!

PR #554, opened 2018-07-02
Add highlighting of all messages from a certain user

DatGuy1:

PR #564, opened 2018-07-04
Add highlighting of all messages from a certain user

DatGuy1:
Cherry picked version of #554


src/controllers/highlights/HighlightController.cpp
```diff
@@ -33,6 +34,26 @@ HighlightModel *HighlightController::createModel(QObject *parent)
...
+{
+    auto *model = new UserHighlightModel(parent);
+    model->init(&this->highlightedUsers);
+
+    return model;
+}
+
+bool HighlightController::userContains(const QString &username)
+{
+    std::vector<UserHighlight> userItems = this->highlightedUsers.getVector();
```
fourtf: Should be `const auto &userItems = ...`

src/controllers/highlights/UserHighlight.hpp
```diff
@@ -0,0 +1,121 @@
+#pragma once
```
fourtf: Delete this file and use the `HighlightPhrase` class instead

src/controllers/highlights/HighlightController.hpp
```diff
@@ -20,10 +22,13 @@ class HighlightController
 
     UnsortedSignalVector<HighlightPhrase> phrases;
     UnsortedSignalVector<HighlightBlacklistUser> blacklistedUsers;
+    UnsortedSignalVector<UserHighlight> highlightedUsers;
 
     HighlightModel *createModel(QObject *parent);
     HighlightBlacklistModel *createBlacklistModel(QObject *parent);
+    UserHighlightModel *createUserModel(QObject *parent);
 
+    bool userContains(const QString &username);
```
fourtf: Rename to `isHighlightedUser`

PR #576, opened 2018-07-06
Add channel badges and emotes to LogsPopup

DatGuy1:
Very ugly code for room-id stuff

### DIFF
--- a/src/widgets/dialogs/LogsPopup.cpp
+++ b/src/widgets/dialogs/LogsPopup.cpp
@@ -35,7 +35,7 @@ void LogsPopup::setInfo(ChannelPtr channel, QString userName)
     this->channel_ = channel;
     this->userName_ = userName;
     this->setWindowTitle(this->userName_ + "'s logs in #" + this->channel_->name);
-    this->getLogviewerLogs();
+    this->getRoomID();
 }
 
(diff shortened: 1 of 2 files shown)


pajlada:
you'll need to rework how network requests are made

```cpp
NetworkRequest request("https://url.here.com/xD");

// Optionally set the request type. GET by default
// request.setRequestType(NetworkRequestType::Put);

// Optionally set the caller (the caller decides which thread the request runs in)
// request.setCaller(myQObject);

request.onSuccess([](auto result) {
    auto json = result.parseJson();
    // ...
    return true; // return false if the json returned was invalid in some way
});

// Start performing the request
request.execute();
```

The [NetworkRequest header file](../blob/master/src/common/NetworkRequest.hpp) is a lot cleaner now, so you can check out what other options you can use there

pajlada:
@DatGuy1 What's the status of the PR, do you feel like it's ready to be reviewed?

DatGuy1:
I'm not sure but i'd like to somehow not just display the time since its confusing at least for me as it is right now https://github.com/fourtf/chatterino2/blob/da1627e47dc0f5b76e22242a2bf9c1e65d99462f/src/providers/twitch/TwitchMessageBuilder.cpp#L112

LajamerrMittesdine:
Tested this and everything seems to work fine. 👍

PR #580, opened 2018-07-07
Fix timer stuff

pajlada:


pajlada:
I'm squashing this bad boy to hide these commits
image

PR #582, opened 2018-07-07
Lock UI Scale Hotkeys

hemirt:
Adds a checkbox to make "Ctrl+Plus", "Ctrl+Minus", "Ctrl+MWheelUp", "Ctrl+MWheelDown", "Ctrl+0" do nothing


pajlada:
I think this is the wrong way to go about it, and sets a bad precedent for how to disable features with hotkeys. There should just be an option to "unset" hotkeys in the not-yet-implemented hotkey settings page

PR #587, opened 2018-07-08
Ignored Phrases [WIP]

hemirt:
Adds an option to instead of removing the entire message from view to replace the ignored phrase inside of message with a custom message, (regex included)

- [x] replacement
- [x] removes twitch emotes from the message info inside matched phrase
- [x] shifts position of other twitch emotes not inside matched phrase
- [x] remember removed emotes in case of capture groups of regex and readd them to info
- [x] be able to use own emotes as replacements


LajamerrMittesdine:
Noticing a lot of white space in messages even with no phrases added.

image

PR #589, opened 2018-07-09
Added mention of user by right-clicking on his nickname.

23rd:


pajlada:
Looks good, just not sure I agree with the `@`-adding being tied to the `usernameBold`-setting

pajlada:
I would say remove possibility of choice for now. if the feature is requested at a later point it won't be hard to re-add

PR #609, opened 2018-07-12
Allow @username for various commands

LajamerrMittesdine:
Allow /ignore, /unignore, /follow, and /unfollow to use @ in front of username.

### DIFF
--- a/resources/resources.qrc
+++ b/resources/resources.qrc
@@ -74,7 +74,7 @@
         <file>avatars/pajlada.png</file>
         <file>images/download_update.png</file>
         <file>images/download_update_error.png</file>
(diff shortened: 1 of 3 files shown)


LajamerrMittesdine:
Do not commit yet, this needs to be changed a bit.

fourtf:
The code to remove the first letter of a string if it is available is shared a lot in the codebase for example when removing usernames here or when removing # in front of channel. As the DRY principle suggests it is better to create a function that does this instead of repeating it.

Create a helper function like `QString trimFirstCharacterIf(const QString &str, QChar char)` or similar in a shared header file.

Issue #617, opened 2018-07-13
Be able to click the "ignore highlights" on the user popup


apa420:

pajlada:
Fixed in #618

PR #618, opened 2018-07-13
'ignore highlights' checkbox in user info popup - did request

apa420:
TODO: better way of determening when a button is changed, currently stateChanged gives a pulse when the userinfopopup opens (added stuff so it doesn't fuck up anything, but it could be done better) #617

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -230,31 +230,29 @@ void UserInfoPopup::installEvents()
         });
 
     // ignore highlights
-    QObject::connect(
-        this->ui_.ignoreHighlights, &QCheckBox::clicked, [this](bool gotClicked) mutable {
-            this->ui_.ignoreHighlights->setEnabled(false);
+    QObject::connect(this->ui_.ignoreHighlights, &QCheckBox::clicked, [this](bool checked) mutable {
+        this->ui_.ignoreHighlights->setEnabled(false);
 
-            if (gotClicked) {
-                getApp()->highlights->blacklistedUsers.insertItem(
-                    HighlightBlacklistUser{this->userName_, false});
-                this->ui_.ignoreHighlights->setEnabled(true);
-            } else {
-                const auto &ignoreHighlightsUsersVector =
-                    getApp()->highlights->blacklistedUsers.getVector();
+        if (checked) {
+            getApp()->highlights->blacklistedUsers.insertItem(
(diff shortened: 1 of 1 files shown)


fourtf:
It would be better if the checkbox was in indeterminate state when the username is matched by a regex but not by a non-regex.

Also showing a tooltip when hovering it which says that it gets matched by a regex would be good.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -227,6 +228,33 @@ void UserInfoPopup::installEvents()
...
+    QObject::connect(
+        this->ui_.ignoreHighlights, &QCheckBox::clicked, [this](bool gotClicked) mutable {
+            this->ui_.ignoreHighlights->setEnabled(false);
+
+            if (gotClicked) {
+                getApp()->highlights->blacklistedUsers.insertItem(
+                    HighlightBlacklistUser{this->userName_, false});
+                this->ui_.ignoreHighlights->setEnabled(true);
+            } else {
+                const auto &ignoreHighlightsUsersVector =
```
fourtf: Use a shorter name for this temporary such as `vector` since this is a temporary reference and only used to make the call to `getApp()->highlights->blacklistedUsers.getVector()` shorter.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -227,6 +228,33 @@ void UserInfoPopup::installEvents()
...
+                if (getApp()->highlights->blacklistContains(this->userName_)) {
+                    this->ui_.ignoreHighlights->setToolTip("Name matched by regex");
+                } else {
+                    this->ui_.ignoreHighlights->setEnabled(true);
+                }
+            }
+        });
 }
 
 void UserInfoPopup::setData(const QString &name, const ChannelPtr &channel)
```
fourtf: This function is 77 lines long at the point of this PR. Split this function into multiple reasonably sized sub functions.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -287,8 +318,23 @@ void UserInfoPopup::updateUserData()
             }
         }
 
+        // get ignoreHighlights state
+        bool isIgnoringHighlights = false;
+        auto ignoreHighlightsUsersVector = getApp()->highlights->blacklistedUsers.getVector();
```
fourtf: Use `const auto  &` so the vector doesn't get copied.

Rename the vector name to `vector` (same reason as above).

PR #651, opened 2018-08-01
mac docs update relating to qt

gempir:
give a little more detailed installation guide for the qt installation


pajlada:
Looks good to me 👍

Issue #640, opened 2018-07-24
[Whispers] Play sound and highlight message


cineafx:
At the moment receiving a whisper can easily be missed if you are not actively looking at the chat.
![Example](https://i.nuuls.com/vGbUD.jpeg)

Maybe an option to add "Receiving a whisper" to the Highlights menu
![How it could look like](https://i.nuuls.com/dGId9.png)

cineafx:
Already mentioned by gempir some time ago:

https://github.com/fourtf/chatterino2/issues/268#issuecomment-377890129

PR #652, opened 2018-08-01
Added functionality of highlighting whispers #640

apa420:
Highlight excluded users will also apply to whispers.
(if a user is highlight excluded and they whisper you, the message won't be highlighted)

### DIFF
--- a/src/common/SignalVectorModel.hpp
+++ b/src/common/SignalVectorModel.hpp
@@ -118,7 +118,7 @@ public:
         rowItem.items[column]->setData(value, role);
 
         if (rowItem.isCustomRow) {
-            this->customRowSetData(rowItem.items, column, value, role);
+            this->customRowSetData(rowItem.items, column, value, role, row);
         } else {
             int vecRow = this->getVectorIndexFromModelIndex(row);
             this->vector_->removeItem(vecRow, this);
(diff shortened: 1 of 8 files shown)


ThePridestalker:
Better approach could be making optional the option of muting whispers highlights through a checkbox in case some people don't want to fully mute an specific user. Normally whispers are for serious talk, while chats are random. If you want to mute pings from someone you most likely want to block that person not just mute the highlights imo.

pajlada:
I personally don't see the purpose of having whispers show up in /mentions at all

ThePridestalker:
About the whispers showing in mentions, I agree with pajlada. I think it would be better to display a whispers pop-up window like the emotes one, having the title formatted as: " Whispers from &lt;user&gt;".
A posible approach could be having a button next to the gear (settings button), that manages the whispers notifications, displaying a list with the format: &lt;user&gt; &lt;number of unread messages&gt;. Also it would display a " New Message " button for the case where you are the one that is going to whisper others, using the same steps as for channels, but creating a whispers smaller pop-up window. ( /w in any channel would do the same as the " + ").

The design of that button UI is similar as the one for managing accounts in chatterino.

In the case this would happen, while having a whispers window open, it wouldn't make sound but the task bar would get highlighted.

apa420:
The practical use of this feature, imo, would be that you get a notification on when you've received a whisper. Some people only have one or a few chats open at a time in chatterino and if those chats are very fast it can be really easy to miss out on a whisper that has been sent to you. I do agree that maybe the better approach would be to make every whisper not appear in the /mentions tab and only in the /whispers one, but the whisper still have the option to ping/flash taskbar/have a red background.

PR #675, opened 2018-08-22
Set displayName to UserInfo link as value.

23rd:
This small change makes correct case of nicknames in UserInfoPopup window and right-click mentions.
At first I wanted to add some getDisplayName/setDisplayName methods, but then I realized there was easier way.
As far as I could see in code and test behaviour of program, it doesn't break anything.


pajlada:
@23rd Have you tried this with localized usernames?

23rd:
@pajlada Yep.

pajlada:
Great, looks good to me! Thanks for the contribution

PR #677, opened 2018-08-24
Added tooltip to links via BTTV API.

23rd:
Okay, it's time to do something more useful.
I have noticed that Chatterino tooltips fit perfectly into concept of BTTV link preview.
So I added new `BttvTooltip` class that uses BTTV API to get information about each link. The user can see it by simply hovering over the link. 
API is requested immediately after the link appears in the chat, then information is added to the tooltip.
When we make request, link must be encoded. Then we decode the received response.
Some urls or descriptions in tooltip could be very long, so we should `setWordWrap(true)`, when it link under cursor.
  
If you want, I can add an option to disable BTTV tooltips. But I think it fits perfectly into the Chatterino, uses very few resources and does not require the option to turn off.

This is what it looks like.   

    
You can see response of good link here: https://api.betterttv.net/2/link_resolver/http%3A%2F%2Fgnu.org
And response of wrong link: https://api.betterttv.net/2/link_resolver/http%3A%2F%2Fgggnu.org


pajlada:
I want the option to disable this, and also I don't want to use the BTTV API like that, it feels wrong. Other than that, it looks great :+1: 
If you want to look into implementing the same feature to this API: https://github.com/pajlada/chatterino-api-cache then let me know, otherwise I can do that

23rd:
BTTV API allows you to easily obtain information about the current link, such as a title, description and even redirect short links.

But if I understand you correctly, you do not want to use BTTV API and want to raise your own link_resolver on the server like `https://braize.pajlada.com/chatterino/twitchemotes/set/`?

pajlada:
@23rd you can replace the URL with https://braize.pajlada.com/chatterino/link_resolver/ now and test it

23rd:
@pajlada Seems like everything works well.

PR #678, opened 2018-08-24
Notification when a stream goes live

apa420:
Platform-addition-proof

- [x] A controller handling the different stored channels
- [x] When a channel is live the split-color changes to pink (currently, suggest a better color)
- [x] Playsound when a channel goes live (same playsound as mentions, maybe it's better if it's a different one) (also  the way this is handled should be reworked)
- [x] Channels outside of ones you have in splits can give notifications
- [x] Toast notifications for Windows 8.x and Windows 10, toast will also use the channel avatar from the live channel (this avi is stored in cache/profileusers/platform/channel.png, this has to be reworked since you can now change where the cache is stored)
- [x] Clean up the code a bit

- When a channel in a split goes live the split-tab will turn pink
- Toasts for windows with the channel avatar as the pic in the toast (only win 8.x and 10)
- Options to enable playsound as well as flashing the taskbar on live
- You will be able to get live-status notifications from channels you don't have in splits
- New Settings-page as well as a notification-controller

Library used for the toasts - https://github.com/mohabouje/WinToast

small showcase (it would seem like this image is too big for github **https://i.nuuls.com/mb6tn.gif**)


chatterino.pro
```diff
@@ -444,14 +452,17 @@ HEADERS  += \
     src/BrowserExtension.hpp \
     src/util/FormatTime.hpp \
     src/util/FunctionEventFilter.hpp \
+    src/controllers/notifications/NotificationModel.hpp \
+    src/singletons/Toasts.hpp \
+    src/common/DownloadManager.hpp
```
LajamerrMittesdine: Missing a \

chatterino.pro
```diff
@@ -249,6 +252,9 @@ SOURCES += \
     src/BrowserExtension.cpp \
     src/util/FormatTime.cpp \
     src/util/FunctionEventFilter.cpp \
+    src/controllers/notifications/NotificationModel.cpp \
+    src/singletons/Toasts.cpp \
+    src/common/DownloadManager.cpp
```
LajamerrMittesdine: Missing a \

pajlada:
Does not compile on Linux

The dependency .pri file needs a win32 { ... } wrapper around it (see boost.pri)
All usages of the library needs ifdefs to only run on Windows

pajlada:
Compiles properly on Linux now :+1:

pajlada:
Merging now, thank you for your hard work on this @apa420, and @LajamerrMittesdine for helping test it out!

PR #679, opened 2018-08-25
Fixed double click in emote popup.

23rd:
At the moment, if `linksDoubleClickOnly` option is enabled, emotes from Emote Popup are not added to the input field.
At first sight it seems that line `this->linkClicked.invoke(link);` should be added to `mouseDoubleClickEvent` to fix problem.
But I think `linksDoubleClickOnly` should be ignored, because double click on emotes to add them is not obvious.


pajlada:
Looks good to me :+1: thank you for your contribution!

PR #680, opened 2018-08-25
Added right-click on nickname to reply on whisper.

23rd:
I improved work of the right click on nickname a little. 
Now, if you right click on user that whispered to you, `/w` is automatically added to input field to answer him.

(And finally dynamic_cast is collapsed to condition!)


pajlada:
Looks good to me, thank you sir!

PR #684, opened 2018-08-27
Added hotkey for Emote menu.

23rd:
I dunno why it hasn't been done yet, but I added hotkey `Ctrl+E` to open the Emote menu.
If I have option `Show input box when empty` disabled, then I can open the Emote menu only if I write something. If I want to send one smile, I need to insert space to see button for the Emote menu. 
So I think hotkey here is very suitable.

I hope you have nothing against adding hotkey.


pajlada:

PR #686, opened 2018-08-27
Fixed down key in input field.

23rd:
At the moment, if user writes message and presses key down arrow, his message is simply erased without any chance of returning it.
`currMsg_` saves message only if user presses key up arrow.
If user types text, presses key up arrow, then key down arrow, everything is pretty fine.
But after this, if user appends message and presses key down arrow, message returns to its previous state.
So I fixed it and little improved work of key down arrow.

How it was:

How it's now:


pajlada:
I've noticed that issue as well. will test this one out before merging. but just looking at the code it looks good 👍

PR #695, opened 2018-08-31
Fixed "Inline whispers should not highlight every tab"

### LINKED ISSUE #691: inline whispers should not highlight every tab
- [ ] when you receive a whisper with your name inside it, you shouldnt get a highlight for every notebook/tab you have

- [ ] debatable: i believe /whispers tab should highlight you, even when there is no highlight keyword inside the whisper

apa420:
Fix #691


pajlada:
This is not how the issue should be fixed, the Message is constant for a reason.
If we want channel-specific message flags, there should probably be another argument to addMessage, with additional message flags that are &'ed together by the channel view

PR #697, opened 2018-09-02
Clear highlights scrollbar with messages.

23rd:
Every new feature must has issues. 
So "Clear messages" has one issue too! =)

If user has highlights in scrollbar, they will be shown even after the chat has been cleared.
So this PR just clears highlights in scrollbar with messages.


pajlada:
Thanks!

PR #698, opened 2018-09-02
fixed chatterino badges

apa420:
still has an issue when zooming in


LajamerrMittesdine:
I have no issues with the badges when I zoom in.

src/Application.hpp
```diff
@@ -60,6 +61,7 @@ class Application
     TaggedUsersController *const taggedUsers{};
     ModerationActions *const moderationActions{};
     TwitchServer *const twitch2{};
+    ChatterinoBadges *const badges{};
```
pajlada: This class should not be called "badges" - it's too generic. chatterinoBadges or just badges in a named struct `chatterino`, not sure which is the standard atm

PR #707, opened 2018-09-17
removed move semantics on append call

qubard:
minor change: append takes an lvalue so move semantics are never invoked (?)

### DIFF
--- a/src/messages/MessageBuilder.hpp
+++ b/src/messages/MessageBuilder.hpp
@@ -55,7 +55,7 @@ public:
 
         auto unique = std::make_unique<T>(std::forward<Args>(args)...);
         auto pointer = unique.get();
-        this->append(std::move(unique));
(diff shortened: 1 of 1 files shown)


qubard:
nevermind, apparently it does call move semantics woops

qubard:
nevermind, apparently this does in fact invoke move semantics LOL

qubard:
nvm, the reason this bugged me is i can't find a move ctor on MessageElement but it inherits boost::noncopyable so they are invoked. woops.

qubard:
nvm, i tested it but made the pull first durr

qubard:
https://github.com/fourtf/chatterino2/blob/af7b742a234efa0c34ca352ad0d56c001d5ec96b/src/singletons/NativeMessaging.cpp#L151 malloc :thinking:

PR #708, opened 2018-09-17
replace malloc with make_unique (minor)

qubard:
SeemsGood ?

### DIFF
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -147,8 +147,7 @@ void NativeMessagingServer::ReceiverThread::run()
 
     while (true) {
         try {
-            std::unique_ptr<char> buf(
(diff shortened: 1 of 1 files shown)


pajlada:
@qubard Looks good, feel free to send a PR where you've added yourself to the https://github.com/fourtf/chatterino2/blob/master/resources/contributors.txt#L5 file if you want to appear in the About page of the application

fourtf:
This is **NOT** good!

Instead of allocating `MAX_NUMBER` chars it not allocated a single char and sets the value to `char(MAX_NUMBER)`. It doesn't even crash for me so this could potentially write into the memory of other variables.

Fixed here: https://github.com/fourtf/chatterino2/commit/47c93fb118c7e4d4fc0c167032207961b46456d6#diff-35f1a935f604d0a97996cc061b138755L150

qubard:
thanks for spotting this so soon. my bad

PR #712, opened 2018-09-19
Dont show empty categories in viewer list.

23rd:
How it was:  
image

How it should be:  
image


src/widgets/splits/Split.cpp
```diff
@@ -485,9 +485,12 @@ void Split::showViewerList()
 
         loadingLabel->hide();
         for (int i = 0; i < jsonLabels.size(); i++) {
+            auto currentCategory = chattersObj.value(jsonLabels.at(i)).toArray();
+            // If current category of chatters is empty, dont show this category.
+            if (currentCategory.size() == 0) continue;
```
pajlada: Please replace the `size() == 0` check with `empty()` instead as documented here http://doc.qt.io/qt-5/qjsonarray.html#empty

pajlada:

PR #714, opened 2018-09-19
Improved insertion of emotes from popup.

23rd:
This PR slightly improves insertion of emotes from popup.
If there is character before cursor, space is automatically inserted before emote.

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -152,7 +152,18 @@ void SplitInput::openEmotePopup()
         this->emotePopup_ = std::make_unique<EmotePopup>();
         this->emotePopup_->linkClicked.connect([this](const Link &link) {
             if (link.type == Link::InsertText) {
-                this->insertText(link.value + " ");
+                QTextCursor cursor = this->ui_.textEdit->textCursor();
+                QString textToInsert(link.value + " ");
+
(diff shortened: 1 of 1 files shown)


src/widgets/splits/SplitInput.cpp
```diff
@@ -152,7 +152,18 @@ void SplitInput::openEmotePopup()
         this->emotePopup_ = std::make_unique<EmotePopup>();
         this->emotePopup_->linkClicked.connect([this](const Link &link) {
             if (link.type == Link::InsertText) {
-                this->insertText(link.value + " ");
+                QTextCursor cursor = this->ui_.textEdit->textCursor();
+                QString textToInsert(link.value + " ");
+
+                auto symbolBeforeCursor = 
+                    getInputText()[cursor.position() - 1];
```
pajlada: Am I correct in assuming that if you access an index that doesn't exist (like -1. if cursor.position() is 0 and getInputText() length is 0), it returns a valid value, that is just null (returns true for `isNull`)?

23rd: Yes.
`getInputText()[-1].isNull()` returns `true`.
As you can see on the gif there is a moment when cursor is at the beginning. After inserting emote no error occurs.

pajlada:

PR #719, opened 2018-09-20
Fixed tooltip of twitch emotes in a cheesy fashion

apa420:


23rd:
I mean I'm not sure that `Twitch Emote` line is really needed.

apa420:
There are lines for other emote types like FFZ channel emotes, FFZ Global emotes etc

23rd:
It makes much more sense to add a line to which channel current emote belongs.
And if emote is global add line "Global Twitch Emote".

pajlada:
sure, this works for now until we can add a tooltip with more granularity (Global emote, sub emote etc)

PR #728, opened 2018-09-23
Fixes that timeout/ban messages didn't show

apa420:
This took all in all 4 hours to find FeelsDankMan


pajlada:
Makes sense, and works when I tested it :+1:

PR #729, opened 2018-09-23
Ignored Phrases

hemirt:
Everything should be working now

see #587


hemirt:
someone needs to check the IgnoresPage and decide how it should look like (inside the settings)

pajlada:

PR #734, opened 2018-09-26
Clear scroll bar when switching channels in a ChannelView

TETYYS:


TETYYS:
Oops, messed up commits.

PR #736, opened 2018-09-26
Changed LimitedQueue::space() to signed value, eliminating underflows

TETYYS:


TETYYS:
Probably related to https://github.com/fourtf/chatterino2/issues/630

pajlada:
I will test this for a few days before accepting, but it looks promising

PR #743, opened 2018-10-01
Added scrolling over tabs with wheel to switch channels.

23rd:
heh3


pajlada:
Something with the mouse scroll multiplier or wheel step stuff makes this behave strangely on my i3 Linux setup. I have to scroll my wheel three times before it switches tab

23rd:
I thought that scrolling on each step of the wheel would be very sensitive for users. But if this is normal behavior for you, I will remove the "three-step" scroll.

gempir:
Doesn't work at all on macOS

23rd:
Oops.

gempir:
Woops aswell. @23rd Sorry, works perfectly fine on macOS aswell.

PR #746, opened 2018-10-02
Don't change split header for whisper

DatGuy1:
Prevent normal channel split headers from changing colours when receiving a whisper

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -229,13 +229,16 @@ void IrcMessageHandler::handleWhisperMessage(Communi::IrcMessage *message)
             app->twitch.server->mentionsChannel->addMessage(_message);
         }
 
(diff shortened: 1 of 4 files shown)


TranRed:
Related Issue: #691

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -229,6 +229,7 @@ void IrcMessageHandler::handleWhisperMessage(Communi::IrcMessage *message)
             app->twitch.server->mentionsChannel->addMessage(_message);
         }
 
+        _message->flags.set(MessageFlag::DoNotTriggerNotification);
```
TranRed: I think this needs to be moved one line down. If I understand correctly, this would also make the whispers tab not getting highlighted. But we only want this for all other channels processed in the forEachChannel-Loop later on.

Correct/Ignore me if I'm being stupid again :)

PR #748, opened 2018-10-04
Fixed window moving 1px to the left each restart.

### LINKED ISSUE #747: Main window moves slightly on restart
Main window moves a few pixels to the left everytime you restart it, instantly triggering my OCD when it pokes over to my main screen.

Cranken:
This is more of a "hack" than a real fix but it seems the issue lies in the implementation of Qt.
I've debugged the calls and all values are set/loaded properly yet it still messes up.

Fixes #747


pajlada:
This fixes an issue that occurs when you use chatterino in windowed mode. If you use chatterino in fullscreen mode there are still some issues left with the window position saving/loading.

PR #757, opened 2018-10-04
add menu option to reload subscriber emotes

Ckath:
So basically the `loadEmotes` function from twitchaccount seems perfectly capable of reloading all the subscriber emotes long after init, it even clears the list nicely, its just never done in the current program. This pr just adds a convenient button for it under reload channel emotes.

From my brief testing it seems to work perfectly fine and doesn't break anything, quite convenient for gift subs nanHead


Ckath:
little demo of it 'working', tested as well after a new sub and it seemed to reload all fine. dont mind all those warnings those are normal
demo

pajlada:
Looks good to me, I will not sub to someone to test it. but I'm sure it works :+1:

PR #758, opened 2018-10-04
Fixed missing icon

### LINKED ISSUE #753: Taskbar/Window icon missing
The application doesnt have an icon (windows falls back to the default) in the taskbar + all otherwise created windows don't have an icon as well.

icon

After quick feedback with other chatterino2 users the issue did not appear to be high-DPI-related, although I did replicate it only on my high-DPI laptop.

Using latest master (39a8e8193c3bee320e11e6c57fdc6927b3f3558e)

Chronophylos:
fixes #753


23rd:
It doesn't really fix icon for me in Windows 10 and default DPI.

Issue #26, opened 2017-02-07
Replace blacklisted words with xxx or *** instead of removing the entire message


cineafx:
Instead of blocking the entire message just replace the blocked word with *** like FFZ does https://i.nuuls.com/JZ52d.png

Maybe make it optional or even on per channel basis.
For example: Block all messages with weeb emotes in forsens chat but only replace them with xxx / *** in all other channels.

gempir:
@cineafx Both links end in a 404

cineafx:
I got no clue why the imgur link is broken. I never deleted it. Maybe some inactivity stuff? Will fix when I come home.

The Chatterino 1 issue link is broken because issues got disabled in the chatterino 1 repository.

gempir:
This should definetly be optional. Since some users just prefer the message completly gone.

cineafx:
Updated with new link and yes I agree it should definitely be optional!

hemirt:
You'd probably want to be able to replace `text phrase` with either `text phrase` _default: `xxx` (or *\**?)_ or `regex phrase` and also have the option to remove full message, I think

pajlada:
Currently being worked on in #587 by @hemirt

PR #768, opened 2018-10-06
Fix emote replacement to render emotes correctly

hemirt:


hemirt:
if you had an emote, and a replacement, and in the replacement you used the emote in the first, or last place (with no character before, or after them respectively) they didnt render

hemirt:
close #26 with this

pajlada:
Works, thanks for the explanation :+1:

PR #772, opened 2018-10-06
[WIP]Cleaner code for flash-taskbar

apa420:


fourtf:
The changes don't make the code more readable.

- it adds more code and an extra variable with no gain
- the duration length variable doesn't store the duration length. Instead it is a "boolean" (2500 means true and 0 means false).
- there is an empty {} block

PR #775, opened 2018-10-06
periodically save tab/window layout

Ckath:
it seems to work, the question is if it works well enough. I have no clue about any issues regarding thread safety or other problems this might spawn.

I just don't want the channels and splits I added gone after I turn off pc without closing chatterino properly


23rd:
I dont think this is good idea to rewrite file more than hundred times in every day.

It is better to make 'Save' button. Or you can just close ch2 to save it manually.

Ckath:
probably better to add a save button instead of this yeah, as far as not tying saving to all specific events where it'd make sense to save goes, which I couldn't see an easy way to do. might be better for the time being to just have a button until something proper is made

pajlada:
I improved upon your idea with this commit a947bf74c8222f95f5b31ac4ed3b20ef8acf6537

PR #785, opened 2018-10-08
Fixed message collapsing

### LINKED ISSUE #773: Message collapsing doesn't work
Specifying any value for the **Collapse messages longer than X lines** does nothing, other than somehow rearranging messages with different padding.

flex3r:
fixes #773


fourtf:
Works 👍

PR #790, opened 2018-10-11
Combined emotes of different tiers in Emote Popup.

23rd:
How it was, and how it's now.

Also this PR allows easily add labels to tiers in the future.

image =====>  image


pajlada:
Works perfectly, thank you for your contribution!

PR #795, opened 2018-10-13
ClosedSplits feature (ctrl+shift+t in browsers)

hemirt:
since ctrl+shift+t is already used to open a new tab, the hotkey is ctrl+g atm
when you close a split, you can reopen it with ctrl+g and it opens in the correct tab and also selects it into foreground
when you closed the tab, the split gets open on currently selected tab

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -261,7 +261,8 @@ SOURCES += \
     src/messages/MessageContainer.cpp \
     src/debug/Benchmark.cpp \
     src/common/UsernameSet.cpp \
-    src/widgets/settingspages/AdvancedPage.cpp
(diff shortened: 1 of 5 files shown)


hemirt:
visual bug: first restored split does not alter default tab name
replicate:
close "pajlada", "fourtf" tab gets renamed "<empty>"
ctrl + g restores "fourtf" split, tab still named "<empty>"
ctrl + g restores "pajlada" split, tab gets renamed "fourtf"

pajlada:
Works well, thank you for your contribution!

23rd:
Seems like this PR closes https://github.com/fourtf/chatterino2/issues/339.

PR #796, opened 2018-10-13
correct options text of hiding buttons

Ckath:
this saying the opposite has been annoying me ever since it was added demo


pajlada:

PR #802, opened 2018-10-17
Pphop issue 793

pphop:
Fixes 2/3 of issue 793:

- Last read message marker now behaves like in chaterino 1, resets on tab switch;
- Customizable last read message color;


pphop:
The value of custom marker color is stored as int. The default value of 0 never collides with any custom value since alpha channel value is 255 by default and is not shown in color picker dialog w/o QColorDialog::ShowAlphaChannel option.

src/widgets/settingspages/LookPage.cpp
```diff
@@ -351,11 +352,59 @@ void LookPage::addLastReadMessageIndicatorPatternSelector(
...
+                     colorUpdater([getCurrentColor]() {
+                         return QColorDialog::getColor(getCurrentColor());
+                     }));
+
+    QPushButton *resetButton = new QPushButton("Reset Color");
+    resetButton->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Policy::Fixed);
+
+    QObject::connect(
+        resetButton, &QPushButton::clicked, colorUpdater([]() {
+            return getApp()->themes->tabs.selected.backgrounds.regular.color();
```
pajlada: I think resetting should return 0

src/widgets/splits/SplitContainer.cpp
```diff
@@ -197,6 +197,13 @@ void SplitContainer::addSplit(Split *split)
 
 void SplitContainer::setSelected(Split *split)
 {
+    static SplitContainer *previousSplitContainer;
```
pajlada: Check out Notebook::select instead, and look at the part that's commented out

PR #806, opened 2018-10-20
Moved the red dot live indicator

apa420:
Got triggered by pixels
from

to


fourtf:
still looks ugly as shit but at least it doesn't trigger the ocd

PR #808, opened 2018-10-21
Fix emote scale

pajlada:
The emote scale slider now works again.
It also contributes to the image set calculations, which figures out which of the images we should load from the image set

### DIFF
--- a/src/messages/ImageSet.cpp
+++ b/src/messages/ImageSet.cpp
@@ -58,6 +58,10 @@ const ImagePtr &ImageSet::getImage3() const
 
 const ImagePtr &ImageSet::getImage(float scale) const
 {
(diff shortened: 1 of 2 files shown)


src/messages/ImageSet.cpp
```diff
@@ -58,6 +58,10 @@ const ImagePtr &ImageSet::getImage3() const
 
 const ImagePtr &ImageSet::getImage(float scale) const
 {
+    auto emoteScale = getSettings()->emoteScale.getValue();
+
+    scale += emoteScale - 1.f;
```
fourtf: Shouldn't `scale` be multiplied with `emoteScale`?

pajlada: You're right, that feels like it makes more sense in the emote size transitions with different scales

Issue #810, opened 2018-10-21
Message history doesn't save on CTRL+Enter


pajlada:
Message history (up arrow) doesn't save if you only sent the message with CTRL+Enter and then delete the line

Reproduction steps:
 - Type `ASD` in the message box
 - Press CTRL+Enter
 - Delete the line in the message box (i.e. through CTRL+Backspace)
 - Press up arrow

Expected text in the message box: `ASD`
What's actually in the message box: nothing

PR #811, opened 2018-10-21
save messages sent with ctrl enter to history as well

Ckath:
Should stop messages sent with ctrl+enter from being left out of history(#810), from brief testing it seems to do just that.


fourtf:
Works for me 👍

PR #822, opened 2018-10-24
added toggle for greying out historic messages

pphop:


pajlada:
I personally don't think this is the right long term solution. I would rather see a theme tab in the settings dialog where there's a color picker next to the "Historic message color" field

PR #823, opened 2018-10-24
fixed badges scaling with emote scale slider #817

### LINKED ISSUE #817: Emote scale slider scales badges
(c49f17c-1)

"non-standard" Badges are currently also scaled by emote scale. Not sure if scaling badges is even intended. But it should either scale all or none, preferably none imo.

grafik

pphop:
Fixes #817.
Now doesnt scale badges at all, instead of scaling half of them.


pajlada:
Works well, thank you!

PR #830, opened 2018-10-25
load custom badges for old chat as well

### LINKED ISSUE #716: Generic subscriber badges are displayed instead of custom ones (preloaded messages only)
When I first tried the beta version I saw the subscriber badges correctly. In the nightly version, however, instead of the custom badges I see generic stars, identical for users with different subscription lengths. Here's the screenshot from browser Twitch chat and the same channel in Chatterino (nightly v1.0.723 built 2018-09-20T11:39:19.0000000Z):
image

**Edit:**
I've just noticed that it happens with old messages, preloaded when Chatterino starts. New messages that appear later are displayed with the correct custom sub badges.

Ckath:
the custom mod badges are loaded here already as well, might as well also refresh the custom sub and cheer badges here so they show up in the history.

resolves #716

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -144,6 +144,7 @@ void TwitchChannel::initialize()
 {
     this->refreshChatters();
     this->refreshChannelEmotes();
+    this->refreshBadges();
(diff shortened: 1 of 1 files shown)


pajlada:
Works, thank you

PR #831, opened 2018-10-26
Makes the commands save in json, fixes #372

### LINKED ISSUE #372: The "Trigger" column in the commands settings is not limited to one word
It is however limited to one word when saving and loading the commands which causes command triggers with multiple words to not get saved properly.

`Command("/a b", "c")` will become `Command("/a", "b c")` after closing and reopening chatterino

apa420:


apa420:
Settings library got reworked, not needed anymore

PR #835, opened 2018-10-26
Removed second unnecessary getLogviewerLogs() call.

23rd:
Second getLogviewerLogs() call causes the QMessageBox double open.
Call should only occur when the ID is successfully received.


pajlada:
:+1:

PR #836, opened 2018-10-26
Added closing LogsPopup with QMessageBox.

23rd:
This PR allows user to close LogsPopup window with one click on QMessageBox, as the logs can not be found and window becomes useless.

Of course this PR works only with this: https://github.com/fourtf/chatterino2/pull/835.
Because second QMessageBox will try to close already closed window and chatterino2 will just crash.


pajlada:
Works, thank you

PR #838, opened 2018-10-26
Refactored LogsPopup.

23rd:
1. Removed getting channel ID from third party service and placed getting ID with own implementation which Chatterino2 has already.
2. `roomID_` is only used in one place, so there is no need to have it in header, better to pass it as argument. Together with this removed useless casts `QString -> int -> QString` of `roomID`.
3. TwitchChannel dynamic cast is used in three places only to get `channelName`, I collapsed dynamic cast and placed it in one place.

There will probably be a conflict with my previous PR, because I separated fix, feature and refactor into several PR's. ¯\_(ツ)_/¯

### DIFF
--- a/src/widgets/dialogs/LogsPopup.cpp
+++ b/src/widgets/dialogs/LogsPopup.cpp
@@ -3,6 +3,7 @@
 #include "IrcMessage"
 #include "common/Channel.hpp"
 #include "common/NetworkRequest.hpp"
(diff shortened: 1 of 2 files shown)


23rd:
Conflicts resolved.

pajlada:
Looks good to me :+1: Thank you!

PR #843, opened 2018-10-26
Fixed pause on mouse hover

### LINKED ISSUE #527: Pause chat on hover doesn't work
Pause timeout does not work.

flex3r:
Fixes the intended 1s pause when hovering over a message (if enabled in settings). fixes #527


pajlada:
Works :+1: Thank you!
Future ideas for the pause: Customize the 1s timer, and show some indicator in the split that it's being paused right now

PR #845, opened 2018-10-28
Fixed Unmaximize button in title bar for Windows.

23rd:
There's a pretty unobvious bug here. Windows only.
If user maximize window by drag and drop to top of screen, 'Unmaximize' button starts working only after second click.
The easiest way I've found to fix this bug is to simply change the condition from `this->windowState()` to `_maxButton->getButtonStyle()`.


fourtf:
nice find!

PR #846, opened 2018-10-28
save settings after clicking ok

Ckath:
from what I can tell this is what saves all the preferences correctly, seems to work after clicking Ok and closing client abruptly (kill, turn off computer etc)


fourtf:
seems good

PR #855, opened 2018-10-31
Replaced long switch in WindowManager.cpp.

23rd:
This is just small refactor of `getUiScaleValue()`.
Method `clampUiScale(scale)` does not allow to go beyond, so we can safely use `int scale` as index in array.

### DIFF
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -640,43 +640,9 @@ float WindowManager::getUiScaleValue()
 
 float WindowManager::getUiScaleValue(int scale)
 {
-    switch (clampUiScale(scale))
-    {
-        case -5:
-            return 0.5f;
-        case -4:
-            return 0.6f;
-        case -3:
-            return 0.7f;
-        case -2:
-            return 0.8f;
-        case -1:
-            return 0.9f;
-        case 0:
-            return 1;
-        case 1:
-            return 1.2f;
-        case 2:
-            return 1.4f;
-        case 3:
-            return 1.6f;
-        case 4:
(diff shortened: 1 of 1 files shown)


fourtf:
1. Now it does god knows what when scale <-5 or scale > 10
2. The system will probably be replaced with using actual float values

SLCH:
I think its better to make this convertion into some math function with limits on input data.
Will think on one later

23rd:
@SLCH I thought about it. This would be possible if the scale change was linear. But the values change unevenly.
First, the scale increases by 0.1, then 0.2, then 0.33, then 0.5.

23rd:
Okay.
What you think about this way:
```
float WindowManager::getUiScaleValue(int scale)
{
    auto s = clampUiScale(scale);
    auto multiplier = [s]() { // multiplier
        if (s < 0) return .1f; 
        if (s < 6) return .2f;
        if (s < 8) return .38f;
        return .5f;
    };
    auto offset = [s]() { // offset
        if (s < 6) return 1;
        if (s < 8) return 0;
        return -1;
    };
    return (offset(s) + (s * multiplier()));
}
```

Yeah, this method is not as concise as the method with an array, but it is independent of `uiScaleMin` and `uiScaleMax` values. In the future, the boundaries can be changed and the function will not need to be changed and will give new numbers. (0.4f, 0.3f, 0.2f or 4.5f, 5, 5.5f)


The only narrow point is return of 2.28 instead of 2.33 scale. 
image

fourtf:
As promised this was replaced by using an actual float value for the scale

PR #861, opened 2018-11-03
Fixed copying of emotes with '< >' symbols.

23rd:


23rd:
Okay, I found a ready-made `cleanUpCode()` method in the code. 
Since it works with twitch emotions, it is more logical to put it in `TwitchEmotes.cpp` and make it as static function.

PR #865, opened 2018-11-05
Hide "below" button after clearing messages.

23rd:


23rd:
@fourtf Replaced commit with new changes.

fourtf:
This is good 👍

PR #878, opened 2018-11-15
fix surrogatepairs messing position of emotes to be removed 

hemirt:
mismatch between actual position, and position in unicode codepoints due to characters over 16 bit being encoded as a surrogate pair of two 16 bit characters (QChar) in QString


hemirt:
While laying in bed I've just thought of a possible issue, so don't merge yet

hemirt:
should be done now and working correctly 👌🏻

hemirt:
I left this on the master branch, I want to use it.

PR #898, opened 2018-12-06
BTTV changed to uppercase

Kllrt:
I guess I have autismo or something, but this bothered me for some time, sorry.


fourtf:
With this change it would say `BTTV` for global emotes, and as before `Bttv` for channel emotes and `BetterTTV in the settings.

I suggest that it would be best if it is changed to BetterTTV in all instances as that seems to be the official name on the website www.nightdev.com/betterttv

Kllrt:
You mean also for variables and stuff or just text (which user can see)?

fourtf:
Just text that the user can see

Kllrt:
I guess it should be okay now? 😅

PR #906, opened 2018-12-14
stop emotes from being cut off

### LINKED ISSUE #897: ffz emotes that are higher than wide will get cut off
Seems to be unaffected by emote size or UI scale
image

Ckath:
fixes #897, not entirely sure why this was ever commented out in the first place, works fine like this again though

### DIFF
--- a/src/messages/layouts/MessageLayoutContainer.cpp
+++ b/src/messages/layouts/MessageLayoutContainer.cpp
@@ -165,7 +165,8 @@ void MessageLayoutContainer::breakLine()
         //        {
         if (element->getRect().height() < this->textLineHeight_)
         {
-            //yExtra -= (this->textLineHeight_ - element->getRect().height()) / 2;
+            // yExtra -= (this->textLineHeight_ - element->getRect().height()) /
+            // 2;
         }
 
         element->setPosition(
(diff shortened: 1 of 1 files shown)


23rd:
This PR fixes a trimming of emotes, but breaks a display of effect for (dis/re)connected messages (fc93d7b7388e65abeaa5fa915f287b90e7590a96). 
image

Ckath:
didn't really notice it as it looks like this on mine, guess those lines weren't commented out for no reason at all 
image

Ckath:
oh now theres even conflics, I'll look at again another time if it stays

hemirt:
request for reopen

Ckath:
@hemirt alright, I was going to look at it 'eventually' and just start over looking at random places that messed with scale and made the emotes cutoff

Ckath:
think this does the trick actually, if it does pls merge so I can go back to using upstream from my aur package instead of a fork with one line changed

Ckath:
@pajlada it would be great if this one line could be changed in the `pajlada-dev` branch too so the emote issue is fixed for the time being in the aur package

Ckath:
merged to `pajlada-dev` in #1003 no reason to keep this open

PR #933, opened 2019-01-27
Historic messages behaviour

hemirt:
"Crossed and Greyed" -> default (and is current default)
"Crossed" -> just crossed
"Greyed" -> just greyed
"No change" -> no change in appearance


hemirt:
image

PR #965, opened 2019-02-27
Fix for inline whispers being logged

moe-ka:


pajlada:
Thank you for your contribution! works :+1:

PR #966, opened 2019-02-28
Use ffz/bttv channel emotes for chatrooms

hemirt:
chatrooms being identified by their twitch format of `chatrooms:<channel` ID>:<room UUID>`

### DIFF
--- a/src/providers/twitch/ChatroomChannel.cpp
+++ b/src/providers/twitch/ChatroomChannel.cpp
@@ -1,0 +1,44 @@
+#include "ChatroomChannel.hpp"
+
+#include <QDebug>
+#include "TwitchApi.hpp"
+#include "common/Common.hpp"
+#include "messages\Emote.hpp"
+#include "providers/bttv/BttvEmotes.hpp"
+#include "providers/bttv/LoadBttvChannelEmote.hpp"
+#include "singletons/Emotes.hpp"
+
+namespace chatterino {
+
+ChatroomChannel::ChatroomChannel(const QString &channelName,
+                                 TwitchBadges &globalTwitchBadges,
+                                 BttvEmotes &globalBttv, FfzEmotes &globalFfz)
+    : TwitchChannel(channelName, globalTwitchBadges, globalBttv, globalFfz)
+{
+    auto list = channelName.split(":");
+    if (list.size() > 2)
+    {
+        this->chatroomOwnerId = list[1];
+    }
+}
+
+void ChatroomChannel::refreshChannelEmotes()
+{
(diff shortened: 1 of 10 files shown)


src/providers/twitch/ChatroomChannel.cpp
```diff
@@ -0,0 +1,44 @@
+#include "ChatroomChannel.hpp"
+
+#include <QDebug>
+#include "TwitchApi.hpp"
+#include "common/Common.hpp"
+#include "messages\Emote.hpp"
```
pajlada: this backslash in the include looks off, change it to a normal forward slash

src/providers/twitch/ChatroomChannel.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace chatterino {
+
+ChatroomChannel::ChatroomChannel(const QString &channelName,
+                                 TwitchBadges &globalTwitchBadges,
+                                 BttvEmotes &globalBttv, FfzEmotes &globalFfz)
+    : TwitchChannel(channelName, globalTwitchBadges, globalBttv, globalFfz)
+{
+    auto list = channelName.split(":");
+    if (list.size() > 2)
```
pajlada: what would be the consequence of a chatroom id such as `chatrooms:123456` being input? the first condition at https://github.com/fourtf/chatterino2/pull/966/files#diff-6f295e4ab67461a536026941daa7aab5R97 would be fulfilled, but then this one wouldn't. would it be smarter to have a freestanding function which ensures the validity of the chatroom syntax before attempting to create a chatroom channel?

src/providers/twitch/ChatroomChannel.cpp
hemirt: fixed

src/providers/twitch/ChatroomChannel.cpp
hemirt: any particular place where to put the  [isChatroom](https://github.com/fourtf/chatterino2/blob/3c01af980579db1e97c9393446a257a412cf70af/src/providers/twitch/TwitchServer.cpp#L26-L39) function?

hemirt:
fixed 
> {time} chatrooms:{channel_id}:{room_id} is live
message to show the channel owner name

src/providers/twitch/ChatroomChannel.cpp
pajlada: I think the current positioning is fine :+1:

pajlada:
Works great :+1: Thank you!

PR #968, opened 2019-03-01
Added the broadcaster to the viewer list 

### LINKED ISSUE #967: Broadcaster no longer showing up in the viewer list, so tabcompleting does not work
This has started happening to me today or a couple of days ago. The broadcaster is no longer placed along the moderators on the Chatterino viewerlist, and thus you can't tabcomplete them.

apa420:
Broadcaster was added to the viewer list api
Closes #967


apa420:
A better way of doing it was added, as shown here
or did I misunderstand

PR #989, opened 2019-03-28
disable requests to resolve link info when its disabled

### LINKED ISSUE #988: Disable requests to braize.pajlada.com when link hovering info is disabled
I currently have this option unselected:
image

However, requests to braize.pajlada.com/chatterino/link_resolver/ are still being made.

Pls fix Volvo

Ckath:
Don't make any requests to get link info while the setting isn't enabled, resolves #988

### DIFF
--- a/src/providers/LinkResolver.cpp
+++ b/src/providers/LinkResolver.cpp
@@ -12,9 +12,12 @@ namespace chatterino {
 void LinkResolver::getLinkInfo(
     const QString url, std::function<void(QString, Link)> successCallback)
 {
+    if (!getSettings()->linkInfoTooltip) {
+        successCallback("No link info loaded", Link(Link::Url, url));
+        return;
+    }
     QString requestUrl("https://braize.pajlada.com/chatterino/link_resolver/" +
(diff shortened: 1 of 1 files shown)


hemirt:
@pajlada since youre online, this needs to be fixed ASAP

pajlada:
Change is ok, but I don't see the purpose in attaching a tooltip in the first place if the setting is disabled.

@hemirt I won't currently be merging any Pull Requests, I don't want to make the merge between dev and master any more difficult than it has already been.

Ckath:
I just attached a tooltip because it wont retry getting info on enabling the setting again, so showing no info was loaded seemed more logical than nothing at all. as of now the full url info is attached as tooltip regardless of the setting so it seemed more straight forward to do it this way. might as well add no tooltip at all though if thats preferred.

btw does this mean current pull requests are pretty much useless if theres already such a huge change ongoing small changes like this cant be thrown in? should I be looking at another branch for relevant source?

pajlada:
@ckath I'll look into maybe making a dev dev branch for these pull requests, so the burden of merging falls on me

PR #990, opened 2019-03-30
Adds option to mute ping sound for specific channels

hemirt:
Mutes the sound of the highlight for hightlights coming from that specific channel

https://i.imgur.com/Lg2SyfB.png


src/widgets/splits/SplitHeader.cpp
```diff
@@ -297,6 +298,23 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
 
     moreMenu->addAction(action);
 
+    {
+        auto action = new QAction(this);
+        action->setText("Mute pings");
```
TranRed:
```suggestion
        action->setText("Mute highlight sound");
```
In my opinion, using the term "ping" is creating inconsistent copy, since it is not used anywhere else in the UI. The corresponding settings use the terms "Highlights" and "highlight sound".

hemirt:
(fixed wrong local branching)

apa420:
Tried it and it works well :+1:

src/controllers/pings/PingController.cpp
```diff
@@ -0,0 +1,70 @@
...
+        {
+            channelVector.removeItem(i);
+            i--;
+        }
+    }
+}
+
+bool PingController::toggleMuteChannel(const QString &channelName)
+{
+    if (isMuted(channelName))
```
pajlada: missing `this->` when calling method

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -917,12 +918,13 @@ void TwitchMessageBuilder::parseHighlights(bool isPastMsg)
         if (!isPastMsg)
         {
             if (playSound &&
-                (!hasFocus || getSettings()->highlightAlwaysPlaySound))
+                (!hasFocus || getSettings()->highlightAlwaysPlaySound) &&
```
pajlada: these if-statements could use some cleanup so it's possible to read what they do

hemirt:
@pajlada i think this is a bit more readable then

pajlada:
Please create a new PR (or just fix this one) based on `nightly` instead of `master`. I think the only manual thing you'll need to fix is the chatterino.pro line-ending-fix

hemirt:
@pajlada should be now all fine, tested it works again

pajlada:
Thank you for the contribution!

PR #1003, opened 2019-04-11
stop emotes from being cut off (pajlada-dev)

RAnders00:
See #906 

I had this patch enabled for a few months now and really the only thing it breaks is the reconnected/etc messages. I think those messages being misaligned is preferable to every emote being cut off, personally.

This is the same as #906 but as requested by @Ckath into `pajlada-dev` for testing/development purposes.

### DIFF
--- a/src/messages/layouts/MessageLayoutContainer.cpp
+++ b/src/messages/layouts/MessageLayoutContainer.cpp
@@ -165,7 +165,8 @@ void MessageLayoutContainer::breakLine()
         //        {
         if (element->getRect().height() < this->textLineHeight_)
         {
-            //yExtra -= (this->textLineHeight_ - element->getRect().height()) / 2;
+            // yExtra -= (this->textLineHeight_ - element->getRect().height()) /
+            // 2;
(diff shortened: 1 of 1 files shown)


Ckath:
after the last few commits that make this pr **only uncomment a single line**, the reconnected/etc messages dont get messed up either for me.

btw

RAnders00:
might be worth merging into pajlada-dev for testing

Ckath:
since things are being rewritten on `dev` anyway and development is frozen on main branch, I'll close the other pr once/if this one is merged to pajlada-dev.

RAnders00:
@Ckath would be nice if you could open another PR that's based on the current pajlada-dev that fixes emotes and reconnect message

Ckath:
@RAnders00 if this doesn't already I don't think I can, the reconnect messages and emotes look perfectly fine to me with this in `pajlada-dev`. I've been using it with it manually patched in before for some time now.

RAnders00:
I'll try today and report back.

PR #1005, opened 2019-04-11
[WIP] Apa remove moderated messages v2

apa420:
Not sure if this way of doing it is better.
see #998


apa420:
There is probably an even better way of doing this

PR #1007, opened 2019-04-12
Show ignored users messages anyway setting

### LINKED ISSUE #849: Functionality to show ignored users as moderator/broadcaster
Options from Chatterino 1 seem not to be implemented yet. The Dropdown is empty for me in 2.0.4 and in the nightly.

Chatterino 2: (empty drop down)
grafik

Chatterino 1: (if moderator / if broadcaster / never)
grafik

pphop:
Closes #849 

preview: https://i.nuuls.com/_4PM6.gif

### DIFF
--- a/src/controllers/ignores/IgnoreController.hpp
+++ b/src/controllers/ignores/IgnoreController.hpp
@@ -12,7 +12,7 @@ class Paths;
 
 class IgnoreModel;
 
-enum ShowIgnoredUsersMessages { Never, IfModerator, IfBroadcaster };
+enum class ShowIgnoredUsersMessages { Never, IfModerator, IfBroadcaster };
 
 class IgnoreController final : public Singleton
 {
--- a/src/controllers/ignores/IgnoreController.hpp
+++ b/src/controllers/ignores/IgnoreController.hpp
@@ -12,6 +12,8 @@ class Paths;
 
 class IgnoreModel;
 
+enum ShowIgnoredUsersMessages { Never, IfModerator, IfBroadcaster };
+
 class IgnoreController final : public Singleton
 {
 public:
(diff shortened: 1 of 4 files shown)


src/controllers/ignores/IgnoreController.hpp
```diff
@@ -12,6 +12,8 @@ class Paths;
 
 class IgnoreModel;
 
+enum ShowIgnoredUsersMessages { Never, IfModerator, IfBroadcaster };
```
pajlada: make this an enum class please

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -80,6 +80,18 @@ bool TwitchMessageBuilder::isIgnored() const
         {
             if (sourceUserID == user.id)
             {
+                switch (getSettings()->showIgnoredUsersMessages)
```
pajlada: making the ShowIgnoredUsersMessages thing an enum class instead would require a static cast here ^

src/widgets/settingspages/IgnoresPage.cpp
```diff
@@ -77,8 +77,24 @@ void addUsersTab(IgnoresPage &page, LayoutCreator<QVBoxLayout> users,
...
+        auto &setting = getSettings()->showIgnoredUsersMessages;
+
+        setting.connect(
+            [combo](const int value) { combo->setCurrentIndex(value); });
+
+        QObject::connect(combo,
+                         QOverload<int>::of(&QComboBox::currentIndexChanged),
+                         [&setting](int index) {
+                             if (index != -1)
+                                 setting = index;
```
pajlada: and here ^

pphop: cast here is not required since it uses an int setting, dont know if there is a way to have an actual enum setting.

PR #1012, opened 2019-04-19
Added deleted messages

### LINKED ISSUE #1009: Delete message support
chrome_0vBMc6Q7R6
In normal twitch chat, there is a moderator option that deletes the specific message instead of timing out the user entirely and all the message history.
Currently Chatterino 2 doesn't even show if a message gets deleted. Related to #955 

The button could be added next to name of chatter when you enable the mod icons.

apa420:
- Added {msg-id} to the moderation tab so /delete can be used on messages
 - Added an error message on incorrect /delete usage
 - Added trashcan for moderation buttons
 - Made deleted messages show up as deleted

close #1009


src/controllers/moderationactions/ModerationAction.cpp
```diff
@@ -73,6 +73,10 @@ ModerationAction::ModerationAction(const QString &action)
     {
         this->image_ = Image::fromPixmap(getApp()->resources->buttons.ban);
     }
+    else if (action.startsWith("/delete"))
+    {
+        this->image_ = Image::fromPixmap(getApp()->resources->pajaDank);
```
pajlada: I would like to see a proper resource be put here, possibly a trashcan like Twitch uses?

apa420: Added a trashcan png

Issue #939, opened 2019-01-31
Mentions are hard to read in light themes


CaptainFlint:
Mentions are now very hard to read in light/white theme:

![image](https://user-images.githubusercontent.com/6475044/52069663-41256980-2590-11e9-9198-b28efbd9036d.png)

Nightly v1.0.876 built 2019-01-31T07:32:25.0000000Z

apa420:
This seems to be fixed in the latest version @fourtf

PR #744, opened 2018-10-01
Add hidden setting for highlight color

Chronophylos:
change `highlighting/color` to any hex color like requested in #703

Issue #710, opened 2018-09-18
Option on how streams from toast notifications should be opened


TranRed:
Currently clicking the toast notification opens the stream twitch page. I would like to have an option to change this behaviour to player in browser or streamlink.  This would make the same options available as we have in channel splits. 

![2018-09-18_06-33-48](https://user-images.githubusercontent.com/16636423/45664609-f210f780-bb0c-11e8-83dd-843345b5cc2f.png)

Since I always use "Open player in browser" my current interaction is as follows:
1. seeing the notification
2. going to the respective tab
3. "Open player in browser" from the channel's split options

Issue #703, opened 2018-09-14
Setting for Custom Highlight background color


wigoor:
The ability to change the highlight background color from red to anything else.

fourtf:
There will be no setting for that. Either complete theme support or nothing.

PR #1017, opened 2019-04-22
Opening streams from toast + quickfix highlights in light theme

TranRed:
I kinda messed up with my committing so I guess this is two changes in one .....

First change:
Implements #710

Added options for opening streams from clicking the toas notficiation: (according to options in splits)
- open in browser
- open player in browser
- open in streamlink (needs to be tested, but should in theory work, since I only use the already existing function)
- don't open the stream (in case someone wants to prevent opening streams "by accident")

new setting

Toast with text based on setting

Second change:
Quick fix for #939 

Hidden setting introduced in PR #744 was setting a "default" highlight color. 
I changed it so the defaults depends on the basic hue of the theme. The hidding setting will only overwrite if it is populated in settings.json.

As mentioned in #703 a proper theme support for this should be implemented at one point.

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -144,7 +144,7 @@ public:
         "/highlighting/whisperHighlight/enableSound", false};
     BoolSetting enableWhisperHighlightTaskbar = {
         "/highlighting/whisperHighlight/enableTaskbarFlashing", false};
-    QStringSetting highlightColor = {"/highlighting/color", "#4B282C"};
+    QStringSetting highlightColor = {"/highlighting/color", ""};
 
     BoolSetting longAlerts = {"/highlighting/alerts", false};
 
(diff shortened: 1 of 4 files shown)


TranRed:
Updated title and description because I kinda fucked up my fork. 🥇

pajlada:
I would've done one thing differently, this might be something you want to experiment with in the future:
Store the "Open toast in" setting as an `int`, make it map to an `enum class` storing the various actions, and then figure out a way to map between an int/enum class value and the string value (and vice versa). That way we can change the wording of the settings shown to the user without having to change the actual implementation. You also get a more readable switch-case in `Toasts.cpp`.

Mapping can be done in multiple ways in order of neatness:
1) Bi-directional map (e.g. boost bimap)
2) Keep two maps (one map<enum class, string>, one map<string, enum class>) in parallel
3) Make two functions, one for converting an enum class to string with a simple switch-case, one for converting a string to an enum class with some if/elseif stuff

PR #1019, opened 2019-04-24
Replaced suffix character 0x206D with 0x2800 (fixes spamming)

TranRed:
Adding Braille Blank Patter (0x2800) as suffix still allows spamming in twitch. Most other "blank" characters I tried do not work anymore.
Only negative point I can think of is that this character has width. But it is the only working one I can find.


TranRed:
I was informed the character shows up on certain fonds. With this knowledge it is not a valid solution anymore.

TETYYS:
You can also do `this->messageSuffix_.append("󠀀"); // E0000` instead. Appears to be a smaller character.

PR #1025, opened 2019-04-27
Changed spamcharacter to a better one

### LINKED ISSUE #1024: invisible character code issue
https://github.com/fourtf/chatterino2/commit/40b4c2fefc6d3a4541f784a1e6bbefffc828a156
this shouldnt be a plain character, but an encoded sequence, imo

apa420:
The old character was a UTF-32 one which QChar doesn't properly support
(from testing it only seems to work with a single UTF-16 char and the one @TETYYS used before required 2)
If some font could see this char we should maybe change it, but from testing it seems to be a good 0-width char
close #1024


apa420:
nvm you can see it on a lot of fonts

PR #1030, opened 2019-04-27
Refactored "open from toasts"-implementation

TranRed:
I think I kinda combined suggestions 2&3 from last merge:

2. Keep two maps (one map<enum class, string>, one map<string, enum class>) in parallel 
3. Make two functions, one for converting an enum class to string with a  simple switch-case, one for converting a string to an enum class with  some if/elseif stuff

Please review and maybe squash commits (?)

grafik

### DIFF
--- a/src/singletons/Toasts.cpp
+++ b/src/singletons/Toasts.cpp
@@ -9,6 +9,7 @@
 #include "providers/twitch/TwitchServer.hpp"
 #include "singletons/Paths.hpp"
 #include "util/StreamLink.hpp"
+#include "widgets/helper/CommonTexts.hpp"
 
 #ifdef Q_OS_WIN
 
@@ -88,7 +89,7 @@ public:
     {
         QString openingMode = getSettings()->openFromToast;
         QString link;
-        if (openingMode == "in browser")
+        if (openingMode == OPEN_IN_BROWSER)
         {
             if (platform_ == Platform::Twitch)
             {
@@ -96,7 +97,7 @@ public:
             }
             QDesktopServices::openUrl(QUrl(link));
         }
-        else if (openingMode == "player in browser")
+        else if (openingMode == OPEN_PLAYER_IN_BROWSER)
         {
             if (platform_ == Platform::Twitch)
             {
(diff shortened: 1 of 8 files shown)


src/singletons/Toasts.hpp
```diff
@@ -7,17 +7,29 @@ namespace chatterino {
 
 enum class Platform : uint8_t;
 
+enum class ToastReactions {
```
pajlada: The casing(?) of the enums don't match the rest of the codebase, we use PascalCase, not camelCase
see https://github.com/fourtf/chatterino2/blob/3f41dfeff942d9becb311a4b1769e8445da016af/src/messages/layouts/MessageLayout.hpp#L24-L29

TranRed:
Added a new commit. Changing to Int Setting "broke" the ComboBox I used. I implemented my own combobox for now, not really sure about that implementations. Doesn't feel high quality to me, but works for now. Should prolly be changed in later [refactor](https://geekandpoke.typepad.com/geekandpoke/2010/11/todoxls.html) . I assume it needs to get reworked anyways once we will try to get rid of the mix of "new" (e.g. general, external tools) and "old" (e.g. commands, highlights, notifications) setting pages.

src/widgets/settingspages/NotificationPage.cpp
```diff
@@ -133,4 +134,31 @@ NotificationPage::NotificationPage()
...
+
+    // update when setting changes
+    pajlada::Settings::Setting<int> setting = getSettings()->openFromToast;
+    setting.connect(
+        [toastReactionOptions](const int &index, auto) {
+            toastReactionOptions->setCurrentIndex(index);
+        },
+        managedConnections);
+
+    QObject::connect(toastReactionOptions, &QComboBox::currentTextChanged,
```
TranRed: Couldn't figure out how to properly connect the currentIndexChanged Signal (with int), to update the setting from that.

TranRed: Now the proper event is connected thanks to @pajlada FeelsGoodManClap

src/singletons/Toasts.cpp
```diff
@@ -35,6 +48,38 @@ bool Toasts::isEnabled()
...
+}
+
+QString Toasts::findStringFromReaction(
+    const pajlada::Settings::Setting<int> &value)
+{
+    int i = static_cast<int>(value);
+    return Toasts::findStringFromReaction(static_cast<ToastReactions>(i));
+}
+
+ToastReactions Toasts::findReactionFromString(const QString &string)
```
TranRed: I just realised this is probably not used anymore, because we connected the proper signal to the combobox now. I want to delete the unused before merging. Will check and commit that later!

src/singletons/Toasts.hpp
pajlada: Should be called `ToastReaction` (singular) not `ToastReactions` (plural)

TranRed: done

PR #1038, opened 2019-04-28
remove unprofessional (puke) inside comments

hemirt:
remove unprofessional (puke) inside comments


fourtf:
The entire comment is useless and may be removed.

pajlada:
@hemirt could you push an update to remove the entire comment?
thanks!

hemirt:
@pajlada @fourtf removed comment overall

PR #1039, opened 2019-04-29
implemented fallback logic for corrupted window layout 

TranRed:
1. before saving the window-layout a backup will created to avoid corruption due to crashes while saving

2. when starting chatterino and the window-layout file returns an empty window layout (due to corruption) the backup will be read and the layout will be build from this data

(worked with the file I managed to corrupt and a proper backup file, but it's hard to test 100% since the issue is not as easily reproducable. also works with "non-existing" file if backup file exists)

### DIFF
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -474,12 +474,7 @@ void WindowManager::save()
 
     // save file
     QString settingsPath = getPaths()->settingsDirectory + SETTINGS_FILENAME;
-
-    // sometimes chatterino exits/crashes while saving the window layout
-    // this led to corrupted file and users losing their entire layout
-    // QSaveFile used to avoid that
     QSaveFile file(settingsPath);
-
     file.open(QIODevice::WriteOnly | QIODevice::Truncate);
 
     QJsonDocument::JsonFormat format =
@@ -624,7 +619,7 @@ void WindowManager::incGeneration()
     this->generation_++;
 }
 
-QJsonArray WindowManager::buildWindowArray(QString &settingsPath)
+QJsonArray WindowManager::loadWindowArray(const QString &settingsPath)
 {
     QFile file(settingsPath);
     file.open(QIODevice::ReadOnly);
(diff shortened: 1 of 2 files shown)


fourtf:
This approach is unconventional and seems overcomplicated.

Instead, save the file to SETTINGS_TEMP and then rename it to SETTINGS_FILENAME. Renaming a file is an atomic operation on most filesystems and thus can't corrupt.

Rename "TEMP_FILENAME" since the name is too general and not associated with settings.

pajlada:
The approach fourtf has described is what I use in the settings library

TranRed:
@fourtf, should I still keep the fallback for reading the temporary file in case the settings file is not available? Since you can't rename to an already existing file, there is a (short) state in which the settings file is deleted before renaming the temporary one. 

Or do you think the operations are save enough and we should risk it?

fourtf:
You can (and should) in this case rename to an already existing filename.

fourtf:
Looks like QSaveFile does what you want

TranRed:
That seems pretty neat. Will test that one

TranRed:
Seems like it does not break the original file anymore! 👍 

grafik

pajlada:
Looks good to me :+1:

src/singletons/WindowManager.cpp
```diff
@@ -477,7 +474,12 @@ void WindowManager::save()
 
     // save file
     QString settingsPath = getPaths()->settingsDirectory + SETTINGS_FILENAME;
-    QFile file(settingsPath);
+
+    // sometimes chatterino exits/crashes while saving the window layout
+    // this led to corrupted file and users losing their entire layout
+    // QSaveFile used to avoid that
```
fourtf: This comment could be removed.

src/singletons/WindowManager.cpp
```diff
@@ -622,4 +624,14 @@ void WindowManager::incGeneration()
     this->generation_++;
 }
 
+QJsonArray WindowManager::buildWindowArray(QString &settingsPath)
```
fourtf: Name could be changed to `loadWindowArray` to suggest that it does a disk operation.

I like that you pass the path as a reference but it needs to be changed to a `const QString&` since it should be immutable.

TranRed: you have to understand that I don't really have a clue how shit works in C++ 🤣 😂 👌

Issue #565, opened 2018-07-04
Can't click link to purchase subscription 


CommName:
You can't press link that you get when you try to type in sub only mode.
![image](https://user-images.githubusercontent.com/17132081/42288973-62b7b8f6-7fbd-11e8-83d9-56a1624d835c.png)

pajlada:
Fixed in #1043

PR #1043, opened 2019-05-01
made links in system messages clickable

TranRed:
Needed to make sub ticket in sub only chat-message accessible. 
Solution for  #565 

Please review again if there is a smarter way of implementing this. :)

grafik

### DIFF
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -107,10 +107,10 @@ MessageBuilder::MessageBuilder(SystemMessageTag, const QString &text)
 
     // check system message for links
     // (e.g. needed for sub ticket message in sub only mode)
-    QStringList textFragments = text.split(QRegularExpression("\\s"));
+    const QStringList textFragments = text.split(QRegularExpression("\\s"));
     for (const auto &word : textFragments)
     {
-        auto linkString = this->matchLink(word);
+        const auto linkString = this->matchLink(word);
         if (linkString.isEmpty())
         {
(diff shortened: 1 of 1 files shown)


src/messages/MessageBuilder.cpp
```diff
@@ -104,8 +104,23 @@ MessageBuilder::MessageBuilder(SystemMessageTag, const QString &text)
     : MessageBuilder()
 {
     this->emplace<TimestampElement>();
-    this->emplace<TextElement>(text, MessageElementFlag::Text,
-                               MessageColor::System);
+
+    // check system message for links
+    // (e.g. needed for sub ticket message in sub only mode)
+    QStringList textFragments = text.split(QRegularExpression("\\s"));
```
23rd: Imho, `textFragments` can be `const QStringList`.

src/messages/MessageBuilder.cpp
```diff
@@ -104,8 +104,23 @@ MessageBuilder::MessageBuilder(SystemMessageTag, const QString &text)
...
+    for (const auto &word : textFragments)
+    {
+        auto linkString = this->matchLink(word);
```
23rd: And this one can be `const auto` too.

TranRed: I don't see any reasons not to

TranRed:
@pajlada since we changed MessageBuilder.cpp in #1045 and #1052 : Will this properly merge?

pajlada:
Looks like it automatically be merged, no problem

PR #1044, opened 2019-05-01
Add NixOS Building Instructions

rexim:
Checked Chatterino2 on [NixOS](https://nixos.org/). Builds and works quite well. Here are instructions just in case.

May I also contribute Chatterino2 to https://github.com/nixos/nixpkgs so it's available to everybody who uses NixOS?


fourtf:
merge this shit ASAP

fourtf:
also sure we can add it to the package manager

PR #1055, opened 2019-05-04
Refactored code.

23rd:
This PR slightly refactors code and clears some warnings during compilation.
Removed some unused variables.
Renamed variables with duplicate names within single methods.
Refactored some conditions into a concise form.

### DIFF
--- a/src/common/DownloadManager.cpp
+++ b/src/common/DownloadManager.cpp
@@ -20,9 +20,7 @@ DownloadManager::~DownloadManager()
 
 void DownloadManager::setFile(QString fileURL, const QString &channelName)
 {
-    QString filePath = fileURL;
     QString saveFilePath;
-    QStringList filePathList = filePath.split('/');
     saveFilePath =
         getPaths()->twitchProfileAvatars + "/twitch/" + channelName + ".png";
     QNetworkRequest request;
--- a/src/controllers/accounts/AccountModel.cpp
+++ b/src/controllers/accounts/AccountModel.cpp
@@ -31,12 +31,12 @@ int AccountModel::beforeInsert(const std::shared_ptr<Account> &item,
 {
     if (this->categoryCount_[item->getCategory()]++ == 0)
     {
-        auto row = this->createRow();
+        auto newRow = this->createRow();
 
-        setStringItem(row[0], item->getCategory(), false, false);
-        row[0]->setData(QFont("Segoe UI Light", 16), Qt::FontRole);
+        setStringItem(newRow[0], item->getCategory(), false, false);
+        newRow[0]->setData(QFont("Segoe UI Light", 16), Qt::FontRole);
 
-        this->insertCustomRow(std::move(row), proposedIndex);
+        this->insertCustomRow(std::move(newRow), proposedIndex);
 
         return proposedIndex + 1;
     }
--- a/src/messages/MessageElement.cpp
+++ b/src/messages/MessageElement.cpp
@@ -291,7 +291,6 @@ void TimestampElement::addToContainer(MessageLayoutContainer &container,
 {
     if (flags.hasAny(this->getFlags()))
     {
-        auto app = getApp();
         if (getSettings()->timestampFormat != this->format_)
         {
             this->format_ = getSettings()->timestampFormat.getValue();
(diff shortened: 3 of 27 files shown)


src/common/DownloadManager.cpp
```diff
@@ -46,9 +44,9 @@ void DownloadManager::onDownloadProgress(qint64 bytesRead, qint64 bytesTotal)
     log("Download progress: {}/{}", bytesRead, bytesTotal);
 }
 
-void DownloadManager::onFinished(QNetworkReply *reply)
```
fourtf: The name `reply` was fine before.

src/controllers/accounts/AccountModel.cpp
```diff
@@ -31,12 +31,12 @@ int AccountModel::beforeInsert(const std::shared_ptr<Account> &item,
 {
     if (this->categoryCount_[item->getCategory()]++ == 0)
     {
-        auto row = this->createRow();
```
fourtf: Something like `newRow` or leaving it as `row` is better. `created` doesn't carry much meaning.

src/messages/MessageElement.cpp
```diff
@@ -188,15 +188,15 @@ void TextElement::addToContainer(MessageLayoutContainer &container,
         for (Word &word : this->words_)
         {
             auto getTextLayoutElement = [&](QString text, int width,
-                                            bool trailingSpace) {
```
fourtf: The element which gets creates *is* not a trailing space it *has* a trailing space. `hasTrailingSpace` would be more fitting.

src/messages/layouts/MessageLayout.cpp
```diff
@@ -50,7 +50,7 @@ int MessageLayout::getHeight() const
 
 // Layout
 // return true if redraw is required
-bool MessageLayout::layout(int width, float scale, MessageElementFlags flags)
```
fourtf: No improvement in readability since `this->` is used for all members.

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -369,18 +369,18 @@ void MessageLayoutContainer::paintSelection(QPainter &painter, int messageIndex,
                     {
                         returnAfter = true;
                         index = line.startCharIndex;
-                        for (int i = line.startIndex; i < line.endIndex; i++)
```
fourtf: Renaming the variables doesn't improve readability. An actual change in the code structure with less nesting would be needed for that.

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -390,17 +390,17 @@ void MessageLayoutContainer::paintSelection(QPainter &painter, int messageIndex,
                         int lineIndex2 = lineIndex + 1;
                         for (; lineIndex2 < this->lines_.size(); lineIndex2++)
                         {
-                            Line &line = this->lines_[lineIndex2];
```
fourtf: Same applies here as above

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -268,7 +268,7 @@ void AbstractIrcServer::onDisconnected()
 
     MessageBuilder b(systemMessage, "disconnected");
     b->flags.set(MessageFlag::DisconnectedMessage);
-    auto disconnected = b.release();
```
fourtf: Inconsistent naming: `connectedMsg` and `disconnectedMessage`

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -321,13 +321,13 @@ void TwitchChannel::addPartedUser(const QString &user)
         this->partedUsersMergeQueued_ = true;
 
         QTimer::singleShot(500, &this->lifetimeGuard_, [this] {
-            auto partedUsers = this->partedUsers_.access();
```
fourtf: It is always wrong to capture the partedUsers in the lambda so why not reuse the variable name?

src/messages/layouts/MessageLayoutContainer.cpp
23rd: This name is more relevant to the `lineIndex2` variable.

fourtf: sure

src/singletons/Paths.cpp
```diff
@@ -37,7 +37,7 @@ bool Paths::isPortable()
 
 QString Paths::cacheDirectory()
 {
-    static QStringSetting cachePathSetting = [] {
```
fourtf: The lambda is immediately invoked so the name `getCachePathSetting` doesn't make much sense.

PR #1057, opened 2019-05-05
Added boldness adjustment for fonts

apa420:


FeelsDankMan:
awoo72

apa420:
Since the setting is hardcoded in fourtfs appbase it's kind of hard to change the name of it, since it is in fact in use.

23rd:
After this PR nicknames began to look terrible.

Value `50` is too low, `75` is too much. `57` was perfect.

image

Issue #1058, opened 2019-05-05
Whispers with more than 1 space dont get send but it shows as if it would


ThePridestalker:
Whispering someone using "/w <space> <space> <username>" shows as sent when it is not getting sent.

pajlada:
Fixed in nightly

PR #1060, opened 2019-05-05
fix whispers display and not sending in case of double spaces before name

### LINKED ISSUE #1053: whisper via command issue
When issuing a whisper via a command, your sent whisper does not show up in your "whispers" tab. 

https://i.imgur.com/9htXASC.gifv

hemirt:
fix #1053 #1058

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -35,13 +35,17 @@
 namespace {
 using namespace chatterino;
 
+static const QStringList whisperCommands{"/w", ".w"};
+
 void sendWhisperMessage(const QString &text)
 {
+    // (hemirt) pajlada: "we should not be sending whispers through jtv, but
+    // rather to your own username"
     auto app = getApp();
     app->twitch.server->sendMessage("jtv", text.simplified());
 }
 
-bool appendWhisperMessageLocally(const QStringList &words)
+bool appendWhisperMessageWordsLocally(const QStringList &words)
 {
     auto app = getApp();
 
@@ -139,7 +143,7 @@ bool appendWhisperMessageLocally(const QStringList &words)
     return true;
 }
 
-bool appendWhisperMessageLocally(const QString &textNoEmoji)
+bool appendWhisperMessageStringLocally(const QString &textNoEmoji)
 {
     QString text = getApp()->emotes->emojis.replaceShortCodes(textNoEmoji);
     QStringList words = text.split(' ', QString::SkipEmptyParts);
@@ -151,11 +155,11 @@ bool appendWhisperMessageLocally(const QString &textNoEmoji)
 
     QString commandName = words[0];
 
-    if (commandName == "/w")
+    if (whisperCommands.contains(commandName, Qt::CaseInsensitive))
     {
         if (words.length() > 2)
         {
-            return appendWhisperMessageLocally(words);
+            return appendWhisperMessageWordsLocally(words);
         }
     }
     return false;
@@ -253,11 +257,11 @@ QString CommandController::execCommand(const QString &textNoEmoji,
     // works in a valid twitch channel and /whispers, etc...
     if (!dryRun && channel->isTwitchChannel())
     {
-        if (commandName == "/w")
+        if (whisperCommands.contains(commandName, Qt::CaseInsensitive))
         {
             if (words.length() > 2)
             {
-                appendWhisperMessageLocally(words);
+                appendWhisperMessageWordsLocally(words);
                 sendWhisperMessage(text);
             }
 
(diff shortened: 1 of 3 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -31,6 +32,136 @@
...
+    QStringList words = text.split(' ', QString::SkipEmptyParts);
+
+    if (words.length() == 0)
+    {
+        return false;
+    }
+
+    QString commandName = words[0];
+
+    if (commandName == "/w")
```
pajlada: Make a `QStringList` containing "/w" and ".w" and use its contains operator to check if it's a whisper instead. make sure to supply the `Qt::CaseInsensitive` argument. This comment applies to other places you've touched that checks if something is a whisper

src/controllers/commands/CommandController.cpp
```diff
@@ -31,6 +32,136 @@
...
+        app->twitch.server->forEachChannel(
+            [&messagexD, overrideFlags](ChannelPtr _channel) {
+                _channel->addMessage(messagexD, overrideFlags);
+            });
+    }
+
+    return true;
+}
+
+bool appendWhisperMessageLocally(const QString &textNoEmoji)
```
pajlada: Please don't use operator overloading with these two functions, call them separate names since the parameters are so similar.

src/controllers/commands/CommandController.cpp
```diff
@@ -518,7 +554,15 @@ QString CommandController::execCustomCommand(const QStringList &words,
...
 
-    return result.replace("{{", "{");
+    auto res = result.replace("{{", "{");
+
+    if (dryRun)
+    {
+        return res;
+    }
+
+    return appendWhisperMessageLocally(res) ? (sendWhisperMessage(res), "")
```
pajlada: Expand this to a logical if-statement instead of using a ternary operator. also what does `(sendWhisperMessage(res), "")` do? does that execute sendWhisperMessage and return an empty string?

src/controllers/commands/CommandController.cpp
```diff
@@ -31,6 +32,136 @@
             "/followersoff"                                             \
     }
 
+namespace {
+using namespace chatterino;
+
+void sendWhisperMessage(const QString &text)
+{
+    auto app = getApp();
+    app->twitch.server->sendMessage("jtv", text.simplified());
```
pajlada: please add a comment here stating that we should not be sending whispers through jtv, but rather to your own username

src/controllers/commands/CommandController.cpp
hemirt: yes, a comma operator
sure

src/controllers/commands/CommandController.cpp
hemirt: I don't understand, but okay, but i believe you should be using the chat youre currently writing in, because the error messages get sent on the same channel, thats why I added "jtv" channel to the "global" error messages checks https://github.com/fourtf/chatterino2/blob/e201881516a06f8cd14038ac7e6b32cc034404f8/src/providers/twitch/IrcMessageHandler.cpp#L532

pajlada: Sending whispers to your own channel means you will not be rate-limited, which can happen if you send whispers to jtv

src/controllers/commands/CommandController.cpp
hemirt: https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L158

https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L260

src/controllers/commands/CommandController.cpp
hemirt: https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L48

https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L146

src/controllers/commands/CommandController.cpp
hemirt: https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L563-L571

src/controllers/commands/CommandController.cpp
hemirt: https://github.com/fourtf/chatterino2/blob/b7d04a648ff7ab584c07f5d9d58defe89b393253/src/controllers/commands/CommandController.cpp#L42-L43

hemirt:
resolved comments

pajlada:
Thank you for your contribution! LGTM :+1:

PR #1072, opened 2019-05-18
fix error on remove items in settings due to signal vector model

hemirt:
fixes an access violation (null) when removing items in settings menu

### DIFF
--- a/src/common/SignalVectorModel.hpp
+++ b/src/common/SignalVectorModel.hpp
@@ -188,6 +188,11 @@ public:
         assert(row >= 0 && row < this->rows_.size() && column >= 0 &&
                column < this->columnCount_);
 
+        if (column >= this->rows_[row].items.size())
+        {
+            return Qt::NoItemFlags;
+        }
+
         return this->rows_[row].items[column]->flags();
     }
 


pajlada:
What are some things that crashed before but does not crash now?

hemirt:
@pajlada removing a highlight phrase

pajlada:
> @pajlada removing a highlight phrase

I'm unable to reproduce the issue on Windows and Linux, is there anything more specific you need to do to experience the crash?

hemirt:
i just got nightly, used the debug build and got the crashes on removing it, its probably something about having a different kind of settings json file or inconsistency in it, anyways this check is always correct against crashes (read out of bounds) and its only on the delete functions, so it shouldnt even be a bottleneck at all, it could fix some crashes for some, and do nothing detrimental anyways, ergo i believe it should be put in anyways

apa420:
would be nice to see what could cause it, do you mind posting your config - personal stuff?

hemirt:
@apa420 tried removing everything in %appdata% folder and then the steps, still crashed, guess it wasnt in the files (since they got auto regernerated)

hemirt:
maybe it depends on qt version, mine is 5.11.0 atm

(5.12.3 didnt like the custom title bar 
image
and added the windows on top of it )

apa420:
So does the nightly crash?
Or only the version you built yourself?

hemirt:
nightly downloaded from releases with everything in %appdata%/chatterino2 deleted, crashes @apa420

hemirt:
What even is the purpose of that assert? Shouldnt it just be an if and return NoItemFlags instead?

hemirt:
Shouldnt it just use https://doc.qt.io/qt-5/qmodelindex.html#isValid

PR #1081, opened 2019-05-20
Custom Badges and Custom Emotes API support

hemirt:
Adds support for custom badges and custom emotes API

Just  share a link and it just  works ❗

https://imgur.com/a/YCnv43U

Examples of the api:
https://mikoto.hemirt.com/chatterino/static/emotes2.json
https://mikoto.hemirt.com/chatterino/static/badges2.json

(or a live example https://mikoto.hemirt.com/chatterino/static/emotes.json )

also supports local files through file:/ (or file:///) scheme

(works with gifs too )

fourtf commit  - certification pending


hemirt:
checkfail due #1082

src/widgets/settingspages/EmotesPage.cpp
```diff
@@ -19,6 +19,8 @@ EmotesPage::EmotesPage()
     //    layout.append(this->createCheckBox("Enable FrankerFaceZ emotes",
     //    settings.enableFfzEmotes)); layout.append(this->createCheckBox("Enable
     //    emojis", settings.enableEmojis));
+    //    layout.append(this->createCheckBox("Enable custom emotes",
```
apa420: This is a deprecated page, use GeneralPage instead for this setting

hemirt: added commented setting for documentation

the other enableXEmotes arent on the general page, so I added it nearby them

src/extensions/customemotes/CustomEmotesLoader.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        QString topName =
+            emoteInfo.name.isEmpty() ? "" : (emoteInfo.name + "<br/>");
+
+        for (const auto &subset : jsonSets)
+        {
+            // exists or empty QString
+            auto jsonSubName = subset.toObject().value("name").toString();
+            emoteInfo.subsetNames.push_back(jsonSubName);
+            QString globalTooltip = "<br/>Global Custom Emote<br/>" +
```
apa420: Wouldn't the api name fit better here?
Like "Global Misaka Emote" and for channels have "Channel Misaka Emote" e. g.
and have emotes sets below it, that way we get 3 lines of text instead of 4

src/extensions/customemotes/CustomEmotesLoader.cpp
```diff
@@ -0,0 +1,253 @@
...
+CustomEmotesLoader::CustomEmotesLoader()
+{
+}
+
+void CustomEmotesLoader::loadCustomEmotes()
+{
+    for (const auto &link :
+         getApp()->extensions->customEmotes->items.getVector())
+    {
+        this->load(link);
```
apa420: Making a separate file for the loading of emotes would be preferred.
See bttvEmotes and ffzEmotes

apa420:
Is there a possiblility to bake ffz and bttv into this extension module. Maybe that way we can avoid a lot of duplicate code. Even if it would require to hardcode some stuff specifically for these two extensions.
FFZ and BTTV would act as default extensions in the module.

pajlada:
I have a few concerns with this.
1) If we decide to support custom services for Emotes and Badges, I do not think they should have a third emote/badge API format. They should conform to an already existing API syntax (FFZ imo)
2) Ease of adding custom services for Emotes and Badges is important, having users go into a settings menu and adding a single url should be enough
3) Do we even want to support custom services for Emotes and Badges? This might create friction between Chatterino users - where two Chatterino users all of a sudden don't see the same emotes, causing confusion. Right now we aim to support BTTV and FFZ, and I don't see a problem with supporting further Emote/Badge providers in the future. But there's a pretty big discussion to be had here to decide if and how that should be done.

As for the build failing - I have tried C++17 on my Macbook Pro and it seems to work - I will look into the issue you linked.

hemirt:
> 1. If we decide to support custom services for Emotes and Badges, I do not think they should have a third emote/badge API format. They should conform to an already existing API syntax (FFZ imo)

The api syntax is loosely based on the FFZ with some respects to the BTTV one. I decided to make the emotes API quite a bit simpler.
I think maybe a versioning scheme and maybe putting the `global` and `channels` array of emote objects emoticons under `emotes` object could be preferable.
I decided to change the `sets` object in FFZ api into an array, to allow duplicate names (as per RFC 4627, object names `should` be unique)
Differences to FFZ:
- no "room" object identifying the room
- `sets` is an array of objects and not an object of objects
- `emoticons` array is changed into `global` and `channels` array
- removed redundant entries such as `id, hidden, css, margins, modifier, offset, owner, public`, and others, they aren't being used anyways

> 2. Ease of adding custom services for Emotes and Badges is important, having users go into a settings menu and adding a single url should be enough

That's currently the way it is, or did I misunderstand something?
User is to add a single url into a setting and it works

> 3. Do we even want to support custom services for Emotes and Badges? This might create friction between Chatterino users - where two Chatterino users all of a sudden don't see the same emotes, causing confusion. Right now we aim to support BTTV and FFZ, and I don't see a problem with supporting further Emote/Badge providers in the future. But there's a pretty big discussion to be had here to decide if and how that should be done.

I believe the same could be said about native chat users without extensions, or mobile users.
On other note, I believe this feature could see use in tight small user base communities - a subset of chatterino users which already is a subset of twitch users.

pajlada:
> > 1. If we decide to support custom services for Emotes and Badges, I do not think they should have a third emote/badge API format. They should conform to an already existing API syntax (FFZ imo)
> 
> The api syntax is loosely based on the FFZ with some respects to the BTTV one. I decided to make the emotes API quite a bit simpler.
> I think maybe a versioning scheme and maybe putting the `global` and `channels` array of emote objects emoticons under `emotes` object could be preferable.
> I decided to change the `sets` object in FFZ api into an array, to allow duplicate names (as per RFC 4627, object names `should` be unique)
> 
> Differences to FFZ:
> 
>     * no "room" object identifying the room
> 
>     * `sets` is an array of objects and not an object of objects
> 
>     * `emoticons` array is changed into `global` and `channels` array
> 
>     * removed redundant entries such as `id, hidden, css, margins, modifier, offset, owner, public`, and others, they aren't being used anyways
> 
> 
> (FFZ example: https://api.frankerfacez.com/v1/room/hemirt , BTTV example: https://api.betterttv.net/2/channels/hemirt )
> 
I appreciate the documentation of the differences, however I think we should just pick one of the API syntaxes and go for complete "api parity", minus some things that we don't read. I don't like having to maintain extra API parsers when we really don't have to
> > 1. Ease of adding custom services for Emotes and Badges is important, having users go into a settings menu and adding a single url should be enough
> 
> That's currently the way it is, or did I misunderstand something?
> User is to add a single url into a setting and it works
Neat
> 
> > 1. Do we even want to support custom services for Emotes and Badges? This might create friction between Chatterino users - where two Chatterino users all of a sudden don't see the same emotes, causing confusion. Right now we aim to support BTTV and FFZ, and I don't see a problem with supporting further Emote/Badge providers in the future. But there's a pretty big discussion to be had here to decide if and how that should be done.
> 
> I believe the same could be said about native chat users without extensions, or mobile users.
> On other note, I believe this feature could see use in tight small user base communities - a subset of chatterino users which already is a subset of twitch users.
I believe "The Chatterino experience" tries to be as close to "The Twitch power-user experience" as possible in terms of Emotes and Badges. I don't think straying further is a choice I want to make. So for me this feature is a no.
>I believe this feature could see use in tight small user base communities
For "tight small user base communities" - a fork with this feature could probably serve the same purpose, especially now that it should be easy for forks to set up their own CI both for Windows and mac OS

hemirt:
The FFZ api doesnt seem to offer much description about certain items.
That said making it compatible with FFZ would amount to the custom emotes deployer to use unique IDs for the sets which offers no benefits, and the set would have to overload the `_type` object to announce whether this set of emotes is intended for global usage or channel usage, and then describe which channel it is.
The `emoticons` is a paralel to `global` and `channels`, thats the same and the emotes are the same minus redundant items (leaving only `name `and `urls`)

To summarize it is possible, but offers no benefit, would only make it worse for the intended audience - small communities of semi-techy users (or atleast 1 in the community).

#
>I appreciate the documentation of the differences, however I think we should just pick one of the API syntaxes and go for complete "api parity", minus some things that we don't read. I don't like having to maintain extra API parsers when we really don't have to

I modeled the "parser" as the currently existing solutions, and with it just a json "decompiler", the parser is easy to use, read, and edit. Obviously changes to the `Emote` classes would effect this.

#

I tried to make the changes as much enclosed as possible, and except few "header"-like line changes and the user interface changes (emote popup and completion model), I believe I managed to do so.

> I believe "The Chatterino experience" tries to be as close to "The Twitch power-user experience" as possible in terms of Emotes and Badges. I don't think straying further is a choice I want to make. So for me this feature is a no.

I believe adding extra options and customization points is what makes chatterino powerful and unique.

src/extensions/customemotes/CustomEmotesLoader.cpp
hemirt: That could make the box a bit wide tho

FeelsDankMan:
It looks like a very good addition to chatterino. I don't really understand why stick into "The Twitch power-user experience" when there's so much room for imporvement. This being a great example.  FeelsDankMan

JAremko:
@hemirt  Thank you for your service :vulcan_salute:

hemirt:
Thought about some providers adding global emotes that you might want to only use in specific channels, will take a look into it

PR #1083, opened 2019-05-23
make it possible to search localized names

### LINKED ISSUE #1046: localized name not searchable in chat history
searching for a localized name does not return any results. This would probably feel more natural for native speakers.

grafik

grafik

Ckath:
seems this is the place localized name is in, works out since it makes the search text name + localizedname: message
image

fixes #1046 

PS: I have no clue where to make prs to but I use the nightly personally so I'm just throwing all small fixes to that

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -440,7 +440,7 @@ MessagePtr TwitchMessageBuilder::build()
     this->addWords(splits, twitchEmotes);
 
     this->message().messageText = this->originalMessage_;
-    this->message().searchText = this->userName +
+    this->message().searchText = this->userName + ": " +
                                  this->message().localizedName + ": " +
                                  this->originalMessage_;
 
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -440,7 +440,9 @@ MessagePtr TwitchMessageBuilder::build()
     this->addWords(splits, twitchEmotes);
 
     this->message().messageText = this->originalMessage_;
-    this->message().searchText = this->userName + ": " + this->originalMessage_;
+    this->message().searchText = this->userName +
+                                 this->message().localizedName + ": " +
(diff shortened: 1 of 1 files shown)


hemirt:
Looking at the code I don't believe this should be the way to do it, since I usually search by 
`"hemirt: "` to see user's messages, and this would from a look at the diff break it

Ckath:
hmm good point, maybe structure it like it is in the client "name(localizedname):" if it has a localized name or "name:" if it doesnt?

Ckath:
maybe put the localized name at the front so its localizedname + name? or maybe "name:localizedname: message", its not like the string is ever visible its only used for searching

Ckath:
image
image
seems to work out well enough and not be too weird implementation wise with just another `:  `

hemirt:
the order should probably be the other way around, i dont see most people searching with `バカ: hey` but i sometimes do search with `baka: hey` https://static-cdn.jtvnw.net/emoticons/v1/1846355/1.0

Ckath:
oh yeah `username: message` is a thing you'd do as well, how about this?
```
diff --git a/src/providers/twitch/TwitchMessageBuilder.cpp b/src/providers/twitch/TwitchMessageBuilder.cpp
index 3b1c6296..c5798b06 100644
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -441,6 +441,7 @@ MessagePtr TwitchMessageBuilder::build()
 
     this->message().messageText = this->originalMessage_;
     this->message().searchText = this->userName + ": " +
+                                 this->originalMessage_;
                                  this->message().localizedName + ": " +
                                  this->originalMessage_;
 
```

Ckath:
starting to get pretty stupid though, maybe just switch the order and call it a day so you can have the `username: message` search or search for moonrunenames

pajlada:
I'm fine with just including the localized name before the username, to retain the `userName: message` search, so searchText would be `localizedName userName: message` - imo that makes most sense without bloating the searchText variable to hell

Ckath:
seems good

PR #1085, opened 2019-05-26
Changed Download link

TranRed:
I think we should change the download link from beta to nightly, since I feel that 90% of problems/questions people have with/about beta are answered with "download latest nightly" anyway.


RAnders00:
Also we should consider removing the part that links to Chatterino 1

PR #1088, opened 2019-06-02
Added usercard button to link to twitch's usercard

### LINKED ISSUE #1087: Suggestion: Integrate /user function
Would be cool if the /user function worked in chatterino.

apa420:
close #1087


TranRed:
I think this fix also should include opening the card via the /user chat command.

It currently gives "Sorry, "/user" is not available through this client. " 
Should just open the same way as the new button imho.

apa420:
/user now opens the same way as the button, you can even specify the channel it gets opened in

Issue #976, opened 2019-03-09
Show 3x size emote picture on hover 


ThePridestalker:

rexim:
@pajlada I wanna contribute to Chatterino. May I take this issue?

pajlada:
@rexim absolutely - go ahead!
A few initial opinions/pointers I have:
 - Should probably be opt-in
 - Will need to check "max size" of an emote. Some FFZ emotes don't serve their max size version

pajlada:
Fixed in latest nightly through #1091

PR #1091, opened 2019-06-04
(#976) Implement Emote preview on mouse hover

rexim:
Solution for #976
Requires PR https://github.com/fourtf/appbase/pull/3

I'm not really sure how to properly submit such changes. We need to probably merge https://github.com/fourtf/appbase/pull/3 first and then I'll have to update this PR to make appbase point to the latest commit. Please let me know what would be the best approach.

Also, I'm not familiar with the code base that well, so please let me know if I implemented that feature completely wrong.

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -139,6 +139,7 @@ SOURCES += \
     src/singletons/Toasts.cpp \
     src/singletons/Updates.cpp \
     src/singletons/WindowManager.cpp \
+    src/singletons/TooltipPreviewImage.cpp \
     src/util/DebugCount.cpp \
     src/util/FormatTime.cpp \
     src/util/IncognitoBrowser.cpp \
@@ -301,6 +302,7 @@ HEADERS += \
     src/providers/twitch/TwitchServer.hpp \
     src/providers/twitch/TwitchUser.hpp \
     src/RunGui.hpp \
+    src/singletons/TooltipPreviewImage.hpp \
     src/singletons/Badges.hpp \
     src/singletons/Emotes.hpp \
     src/singletons/helper/GifTimer.hpp \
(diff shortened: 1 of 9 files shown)


rexim:
The CI fails because of https://github.com/fourtf/appbase/pull/3

pajlada:
Neat change so far. What I noticed when testing:
- Emotes can "stick" in the tooltip, you mentioned it might have something to do with ChannelView.

What I would like to see in addition to what you've already done:
- I would like the bool setting to be an integer instead, this leaves us open for the next point
- Implement another value for the setting, instead of just on and off it can also be "show when holding Alt" or something

rexim:
@pajlada 

> Emotes can "stick" in the tooltip, ...

The problem was that TooltipWidget is a singleton and ChannelView is not. There are many ChannelView fighting for a single TooltipWidget.

So the solution was to introduce a centralized place from which the TooltipWidget is updated (TooltipPreviewImage singleton). Different ChannelView-s just put the Image they want to preview to TooltipPreviewImage and forget about it.

Hope that makes sense

> I would like the bool setting to be an integer instead, ...

Done.

> Implement another value for the setting, instead of just on and off it can also be "show when holding Alt" or something

Holding any modifiers while moving the mouse stops sending `ChannelView::mouseMoveEvent()` events for some reason so it's not really clear how to quickly implement that. Maybe we can implement that option in the scope of a separate issue?

rexim:
**UPD** @apa420 implemented the third options in https://github.com/tsoding/chatterino2/pull/1

It has the following limitations:
1. It uses Shift instead of Alt, because of the modifier problem that I mentioned at the end of my previous comment. Shift works for some reason.
2. Pressing Shift does not immediately show the preview. You need at least one additional MouseMove event. Trying to work around that leads to a pretty complicated logic, so it would be better to do in the scope of a separate issue.

pajlada:
Very cool, thanks for your contribution!

Issue #1100, opened 2019-06-16
[Bug] Emote preview & title area


savabozo:
Can speculate on what happens, but it will be easier to look at
<details>
  <summary>screenshot 1</summary>
  
![2019-06-16_16-59-37](https://user-images.githubusercontent.com/47025788/59564463-2d1e6800-9058-11e9-89fc-aa07860d8567.png)

</details>
<details>
  <summary>screenshot 2</summary>
  
![2019-06-16_17-00-50](https://user-images.githubusercontent.com/47025788/59564472-635be780-9058-11e9-8242-79e93421be7a.png)

</details>

Reproduction:
1. Any title of live streamer
2. Hover over emote at chat
3. Hover over title

pajlada:
Thanks for the detailed issue post, I am able to reproduce and I'm pretty sure I know what causes this

pajlada:
Fixed by #1108

PR #1108, opened 2019-06-22
(#1100) Clean TooltipPreviewImage on SplitHeader enterEvent

rexim:
Supa quick fix for #1100

Looks like this bug may come back in other places where we use `TooltipWidget`. For a permanent fix we have to go through all of the places that touch `TooltipWidget` and check if they clean the preview
image correctly.

It would be better to merge TooltipPreviewImage and TooltipWidget together and change `TooltipWidget::setText` to be

```c++
void TooltipWidget::setText(const QString &text, ImagePtr image = ptrnull);
```

But it's impossible right now, because TooltipWidget is located in appbase and `ImagePtr` is not visible there (it's located in the main chatterino repo). Plus we also need an access to `getApp()->windows->gifRepaintRequested`.

First step in implementing the permanent fix would be moving TooltipWidget to chatterino from appbase.

Any thoughts on the situation?

### DIFF
--- a/src/widgets/splits/SplitHeader.cpp
+++ b/src/widgets/splits/SplitHeader.cpp
@@ -10,6 +10,7 @@
 #include "singletons/Resources.hpp"
 #include "singletons/Settings.hpp"
 #include "singletons/Theme.hpp"
+#include "singletons/TooltipPreviewImage.hpp"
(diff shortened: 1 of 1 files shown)


pajlada:
Not sure about a more permanent solution, but this will work fine for now. Thank you!

PR #1117, opened 2019-06-28
Marked VIP badge as "channel authority" badge

RAnders00:


pajlada:
I believe the VIP badge is actually a "Channel" badge - it takes up the same slot as the Broadcaster and Moderator badge

PR #1172, opened 2019-08-03
Improved dmg creation.

23rd:
The current dmg is inconvenient for updates of Chatterino. =(
So I changed the way of dmg file creation ([create-dmg](https://github.com/andreyvit/create-dmg)) and added `/Applications` symlink.
Also this PR adds icon for Chatterino volume.

Before and after patch.
image

image


pajlada:
:+1:

PR #1183, opened 2019-08-04
Fixed switching tabs using trackpad scroll.

23rd:
Unfortunately, my implementation of the switching tabs with the scroll is useless for the Mac users with the trackpad. =(
So this PR fixes it.

[i.imgur.com/GQtSFFZ.mp4](https://i.imgur.com/GQtSFFZ.mp4)


pajlada:
High risk change, I will want to try this out properly before merging it

23rd:
No problem.
120 units was taken from [Qt documentation](https://doc.qt.io/qt-5/qwheelevent.html#angleDelta).

fourtf:
Please use `this->` when accessing members.

pajlada:
Merging so its built on all platforms so I can more easily test it on all platforms.

PR #1184, opened 2019-08-05
Added setting to colorize usernames who have not set own color.

23rd:
This PR adds a new setting to colorize usernames who have not set own color like Twitch does.
It is not always convenient to communicate with gray users. =(
Different colors can help you find the right user faster.

To implement this feature I created a new singleton `UsernameColors` class  that stores the colors of all gray users in the `std::map<int, QColor>`.
I'm not quite sure about the implementation, so I need your opinion.

By the way, as far as I could test, a `if (iterator != this->tags.end())` condition never becomes false and a `iterator.value().toString()` just returns an empty string.

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -23,7 +23,40 @@
 #include <QDebug>
 #include <QMediaPlayer>
 #include <QStringRef>
+#include <QColor>
 #include <boost/variant.hpp>
+
+namespace {
+
+QColor getRandomColor(QVariant value)
+{
+    static const std::vector<QColor> twitchUsernameColors = {
+        QColor(255, 0, 0), // Red
+        QColor(0, 0, 255), // Blue
+        QColor(0, 255, 0), // Green
+        QColor(178, 34, 34), // FireBrick
+        QColor(255, 127, 80), // Coral
+        QColor(154, 205, 50), // YellowGreen
+        QColor(255, 69, 0), // OrangeRed
+        QColor(46, 139, 87), // SeaGreen
+        QColor(218, 165, 32), // GoldenRod
+        QColor(210, 105, 30), // Chocolate
+        QColor(95, 158, 160), // CadetBlue
+        QColor(30, 144, 255), // DodgerBlue
+        QColor(255, 105, 180), // HotPink
+        QColor(138, 43, 226), // BlueViolet
+        QColor(0, 255, 127) // SpringGreen
+    };
+
+    // If someday Twitch will replace all the user-ids with strings
+    // then we just choose a random color.
+    const auto userId = value.toInt();
+    const auto index = (userId ? userId : std::rand())
+        % twitchUsernameColors.size();
+    return twitchUsernameColors[index];
+}
+
+}
 
 namespace chatterino {
 
@@ -466,10 +499,15 @@ void TwitchMessageBuilder::appendChannelName()
 
 void TwitchMessageBuilder::parseUsername()
 {
-    auto iterator = this->tags.find("color");
-    if (iterator != this->tags.end())
+    const auto iterator = this->tags.find("color");
+    if (const auto color = iterator.value().toString(); !color.isEmpty())
     {
-        this->usernameColor_ = QColor(iterator.value().toString());
+        this->usernameColor_ = QColor(color);
+    }
+    else if (getSettings()->colorizeNicknames &&
+        this->tags.contains("user-id"))
+    {
+        this->usernameColor_ = getRandomColor(this->tags.value("user-id"));
     }
 
     // username
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -42,6 +42,8 @@ public:
     BoolSetting hideModerated = {"/appearance/messages/hideModerated", false};
     BoolSetting hideModerationActions = {
         "/appearance/messages/hideModerationActions", false};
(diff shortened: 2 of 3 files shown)


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -466,10 +467,16 @@ void TwitchMessageBuilder::appendChannelName()
...
+    const auto iterator = this->tags.find("color");
+    if (const auto color = iterator.value().toString(); !color.isEmpty())
+    {
+        this->usernameColor_ = QColor(color);
+    }
+    else if (getSettings()->colorizeNicknames &&
+        this->tags.contains("user-id"))
     {
-        this->usernameColor_ = QColor(iterator.value().toString());
+        const auto id = this->tags.value("user-id").toInt();
```
pajlada: User IDs should not be stored as ints, Twitch holds no guarantees about user IDs being convertible to numbers.

src/singletons/UsernameColors.cpp
```diff
@@ -0,0 +1,44 @@
...
+        QColor(30, 144, 255), // DodgerBlue
+        QColor(255, 105, 180), // HotPink
+        QColor(138, 43, 226), // BlueViolet
+        QColor(0, 255, 127) // SpringGreen
+    };
+
+    const auto it = this->usernameRandomColors_.find(userId);
+    if (it == this->usernameRandomColors_.cend())
+    {
+        const auto color = twitchUsernameColors[
```
pajlada: I would prefer if colors weren't fully randomized, rather deterministically chosen, so a user will have the same color all the time even after restarts. this could probably be done using standard C++ <random>

src/singletons/UsernameColors.hpp
```diff
@@ -0,0 +1,22 @@
...
+class UsernameColors
+{
+public:
+    static UsernameColors &getInstance();
+    QColor getColor(const int userId);
+
+    UsernameColors(const UsernameColors &) = delete;
+
+private:
+    UsernameColors()
```
pajlada: replace with `UsernameColors = default;`

src/singletons/UsernameColors.cpp
23rd: Should I store the `usernameRandomColors_` map in a separated file?

pajlada: It should be randomized using a deterministic method (mersienne twister) on the user id so the calling randomColor on the same user id with always produce the same color

23rd: Okay, now I get it, thanks.

src/singletons/UsernameColors.cpp
```diff
@@ -0,0 +1,55 @@
...
+        QColor(30, 144, 255), // DodgerBlue
+        QColor(255, 105, 180), // HotPink
+        QColor(138, 43, 226), // BlueViolet
+        QColor(0, 255, 127) // SpringGreen
+    };
+
+    const auto it = this->usernameRandomColors_.find(userId);
+    if (it == this->usernameRandomColors_.cend())
+    {
+        distribution_ = std::uniform_int_distribution<int>(
```
pajlada: Missing `this->`

src/singletons/UsernameColors.cpp
```diff
@@ -0,0 +1,55 @@
...
+            0, twitchUsernameColors.size() - 1);
+
+        // Generate seed from userId string to get the same random value.
+        // That will help us to have the same color for the user with this
+        // userId all the time even after restarts.
+        auto seed = 0;
+        for (const auto &c : userId.toUcs4()) {
+            seed += c;
+        }
+        twister_.seed(seed);
```
pajlada: Missing `this->`

src/singletons/UsernameColors.cpp
23rd: Sry. =(

src/singletons/UsernameColors.cpp
```diff
@@ -0,0 +1,56 @@
...
+        QColor(30, 144, 255), // DodgerBlue
+        QColor(255, 105, 180), // HotPink
+        QColor(138, 43, 226), // BlueViolet
+        QColor(0, 255, 127) // SpringGreen
+    };
+
+    const auto it = this->usernameRandomColors_.find(userId);
+    if (it == this->usernameRandomColors_.cend())
+    {
+        this->distribution_ = std::uniform_int_distribution<int>(
```
pajlada: Can distribution_ be initialized in the constructor or something instead?

23rd: Done I think.

23rd:
Okay, I thought maybe it would be better to use `std::unique_ptr`.

fourtf:
`unique_ptr` doesn't do anything here except requiring `->` to access members.

fourtf:
I dislike the use of a map to cache the colors for users. I would rather use the modulo of the user-id.

23rd:
Okay, then we can throw out this singleton and implement it with single function in an anonymous namespace.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -23,8 +23,41 @@
 #include <QDebug>
 #include <QMediaPlayer>
 #include <QStringRef>
+#include <QColor>
 #include <boost/variant.hpp>
 
+namespace {
+
+QColor GetRandomColor(QVariant value)
```
pajlada: One last comment: function name should be `getRandomColor`

PR #1196, opened 2019-08-10
🪁 Make endlines consistent

TalVivian:
by turning all CRLF to LF

Most of the files use unix line endings, but about 1/3 has window's line endings.

I've set `core.autocrlf` to `input` on linux and run `git add -A --renormalize`.
It's possible to see changes with `git diff --staged | cat -v | less` ([credits to droid7c2 @ stackoverflow](https://stackoverflow.com/a/39502438))

I checked and fourtf is one of the persons who commit with window's endlines.
May I suggest setting:
- `core.autocrlf` to `true` on windows
- `core.autocrlf` to `input` on linux
as explained [here](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace)

I dislike how git doesn't recognise it as a change and instead marks it as removed and then added lines, but I doubt we can do anything about it.


pajlada:
I've noticed the line-ending-difference too, but it doesn't bother me enough right now to change so much of history of the repo. Also doesn't feel right as a PR to me.

TalVivian:
That's fine. Thanks for considering it 🙏

PR #1197, opened 2019-08-11
🐛 Fixed Travis not uploading dmg and AppImage

AnotherFoxGuy:
Travis needs to have set `overwrite: true` otherwise it won't upload the dmg and the AppImage to GitHub releases


pajlada:
will try this out :+1:

PR #1201, opened 2019-08-12
Fixed dmg Travis deployment.

23rd:
I accidentally found that [one](https://github.com/Chatterino/chatterino2/pull/1172) of my previous commits broke the dmg creation, sorry. =(
At the moment a size of the dmg file is 3.4 MB, so the app does not contain Qt dependencies. And it works if the user updates the previous version, but a clean install does not work.
So, we have to use the `macdeployqt` again.
This PR fixes the mac deployment in some tricky way — we call `macdeployqt` to create the dmg file with dependencies and after this we call `create-dmg` to create the final user-friendly dmg.
It may seem that this is an overhead deployment, but I think it's not a big problem while we use the CI.


pajlada:
@23rd Please fix the merge conflicts

23rd:
Done.

PR #1202, opened 2019-08-12
Fix Settings UI inconsistencies

mateossh:
Hi! A while ago I saw streamer @rexim contributed to this project. I like the idea and although I'm not a C++ developer I wanted to contribute to chatterino as well. My changes are small and related to UI.

- Move "About" tab to make vertical margins consistent
- Increase settings window width to prevent creating horizontal scroll (It is caused by last line in "Attributions" group)
- Fix "show license" link width by using QLabel instead of custom SignalLabel (I couldn't find what's issue in SignalLabel)


fourtf:
Seems good to me, thanks for the contribution!

Issue #591, opened 2018-07-09
Emote completion with colon


Spaenny:
It would be neat to have emote completion with colon like FFZ does currently, maybe make it an option to choose between colon and none for emote completion?

infowolfe:
The current tab completion mechanism for emotes is extremely awkward and doesn't follow twitch "feel" from a ux perspective. Additionally, the tab-complete only seems to work from the start instead of twitch's greedy/lazy completion. if i type :pls it should offer up all `/(.*)pls(.*)/gxi` emotes via popup similar to the twitch UI/UX. Additionally, the current emotes popup/subwindow doesn't have any search/filtering capability, which makes it extremely difficult to quickly find emotes when you've got many channels subscribed.

karl-police:
Not only emote, but also mention function

apa420:
The point of doing : is to specify you want emotes and not usernames.

karl-police:
> The point of doing : is to specify you want emotes and not usernames.

Obviously, I ment that when you do @ the popup appears. I was just talking about the Popup or the list.

Ceono:
Still not a thing?

Mm2PL:
This functionality hasn't been implemented yet, so it is still not a thing.

apa420:
This is a open source project, feel free to implement it yourself! ;) @Ceono

MastaNub:
Is this not gonna get added?

karl-police:
> Is this not gonna get added?

Devs are busy probably, which is okay you know. Maybe it should be put on a to-do flair so it this doesn't get asked anymore.

Thehail:
Waiting for this to be added, good suggestion.

fourtf:
fixed in here https://github.com/Chatterino/chatterino2/commit/f7237dccdd5a4994b057967c3b3cdd7fa5166d55

twitch emotes and emojis will probably be added in a bit

PR #1204, opened 2019-08-14
Make Emote Tab Completion Search for Substrings

leon-richardt:
### Description

At the moment, tab completion only considers emotes that start with the same prefix that is currently under the cursor.  This PR makes tab completion look for the current prefix at any position in emote names. This functionality has been requested in https://github.com/Chatterino/chatterino2/issues/591#issuecomment-431911018 and I certainly find it useful.

Note that this does not change the way username completion works. In order to complete a username, you will still have to type the actual start of the username. (**lada** would not complete to **pajlada**, for example.) 
I would prefer to keep it that way as user names generally don't really follow a pattern like emote names often do. (Like dance emotes often ending on "Pls".)

### Changes

Only minimal changes were required since Qt provides us with all necessary functionality already.
We change the check in the `addString` lambda from `QString::startsWith` to [`QString::contains`](https://doc.qt.io/qt-5/qstring.html#contains).

We also need to update the `filterMode` from `Qt::MatchStartWith` (the default value) to [`Qt::MatchContains`](https://doc.qt.io/qt-5/qt.html#MatchFlag-enum). This is done with [`QCompleter::setFilterMode`](https://doc.qt.io/qt-5/qcompleter.html#filterMode-prop).

### Examples

- In forsen's channel, tabbing on **pls** will cycle through **DonaldPls**, **forsenPls**, **GachiPls**, **nyanPls** and **SourPls**.

- As a forsen subscriber, tabbing on **sen1** will complete to **forsen1**.

- As a pajlada subscriber, tabbing on **shrug** will cycle through **pajaShrugL** and **pajaShrugR**. (Unless you are in a channel with more **shrug** emotes, of course.)

### Note

Please let me know your thoughts on this one, I would gladly appreciate any feedback.

~~Since this is my first time contributing, I am not sure whether this is the right branch to merge with. It looks to me like `master` should be merged back into `nightly` before accepting this PR?~~

### DIFF
--- a/src/common/CompletionModel.cpp
+++ b/src/common/CompletionModel.cpp
@@ -10,6 +10,7 @@
 #include "providers/twitch/TwitchChannel.hpp"
 #include "providers/twitch/TwitchServer.hpp"
 #include "singletons/Emotes.hpp"
+#include "singletons/Settings.hpp"
 
 #include <QtAlgorithms>
 #include <utility>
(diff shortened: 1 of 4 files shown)


fourtf:
It should be merged into master, not nightly. Nightly branch was supposed to be deleted.

leon-richardt:
Oh okay, I didn't know that, thanks for letting me know. I changed the target branch to `master`.

fourtf:
This should definitely be a setting for now! Some users prefer consistently tabbing into the same emotes.

leon-richardt:
Made it a setting under the "Miscellaneous" section. Not quite sure if that is the best place for it though.

Any suggestions for the settings's description? The current phrasing seems kind of awkward to me.

fourtf:
Seems good to me

PR #1209, opened 2019-08-17
Fix Emote Completion Bug

leon-richardt:
This is a followup to #1204. The bug was discussed on pajlada's stream but I did not update the thread in #1204 in time. Sorry for creating a new PR.

### Description

A bug would occur when changing the completion mode from "prefix-only" to "substring" while completing a word.
In that case, the suggestion list was not updated until the next completion attempt on a new word, and the completion failed.

### Fix

This is fixed by forcing a suggestion list refresh whenever the setting's checkbox is updated.

By the way, if anyone knows a better way to get the currently active split, please let me know. Best I could come up with is:
```c++
auto selectedSplit =
    static_cast<SplitContainer *>(
        getApp()->windows->getMainWindow().getNotebook().getSelectedPage())
        ->getBaseNode()
        ->getSplit();
```

### DIFF
--- a/src/common/CompletionModel.hpp
+++ b/src/common/CompletionModel.hpp
@@ -48,6 +48,7 @@ public:
     virtual int rowCount(const QModelIndex &) const override;
 
     void refresh(const QString &prefix);
+    void updateEmoteCompletionMethod();
 
 private:
     TaggedString createUser(const QString &str);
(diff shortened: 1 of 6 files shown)


fourtf:
It seems to me that it'd be easier to connect to the setting's signal and changing the `completionInProgress_` variable in the callback.

leon-richardt:
Alright, didn't know there were signals for changed settings! This indeed makes it easier. 

Definitely squash commits before merging. :sweat_smile:

pajlada:
Cool, works great! Thanks for the contribution! Feel free to make another PR where you add yourself to the [list of contributors](https://github.com/Chatterino/chatterino2/blob/master/resources/contributors.txt#L1) included in the About page.

PR #1222, opened 2019-08-20
Improved NetworkRequest

fourtf:


fourtf:
Yeah seems good, I like it

PR #1230, opened 2019-08-22
Add a copy button near the username in usercards

### LINKED ISSUE #942: Add copy username to clipboard button on modcards

Mm2PL:
Things to do, I couldn't figure out how to:
  [ ] Change the font color on the button ~~to be `#5B5B5B`~~  so the button doesn't look like part of the username

This closes #942


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -58,11 +58,26 @@ UserInfoPopup::UserInfoPopup()
...
         {
-            auto name = vbox.emplace<Label>().assign(&this->ui_.nameLabel);
+            auto name_box = vbox.emplace<QHBoxLayout>();
+            name_box.withoutMargin();
+
+            auto name = name_box.emplace<Label>().assign(&this->ui_.nameLabel);
+
+            LayoutCreator<EffectLabel2> copy_user_name = name_box.emplace<EffectLabel2>(this);
+            copy_user_name->getLabel().setText("Copy");
+            copy_user_name->getLabel().setCentered(true);
```
TranRed: You can change color of the text with copy_user_name->getLabel().setStyleSheet("color: #5B5B5B"); or copy_user_name->getLabel().setStyleSheet("color: rgb(91,91,91)"); if you prefer the RGB notation.

I think the gray is a bit hard to read tbh. Maybe try some other colors:
ia16u3nBLl

Mm2PL: How about this look?

green copy button

RAnders00:
Not sure how achieveable this is, but I associate copyable text with an input field with a copy button. Similar to this:
DD661DAA-6259-4613-B5CD-F1CA40B6B875

TranRed:
You can easily change it to this: (thrashcan used as reference, cause to lazy to add a new resource)
```C++
LayoutCreator<EffectLabel2> copy_user_name = name_box.emplace<EffectLabel2>(this);
copy_user_name->setPixmap(app->resources->buttons.trashCan);
copy_user_name->setScaleIndependantSize(30, 30);
```

Will look like this:
9ORwgEGZDk

Alternatively you could change it into a button, similar to the mod/unmod buttons. I would prefer this solution because it seems more consistent with other buttons on the user card
```C++
auto copy_user_name = name_box.emplace<Button>(this);
copy_user_name->setPixmap(app->resources->buttons.trashCan);
copy_user_name->setScaleIndependantSize(30, 30);
```

fourtf:
Code looks good. However, I dislike the icon but that can be changed later as well. I'll merge this some time next week.

Mm2PL:
Okay I can look and search for a better icon or just make one

PR #1232, opened 2019-08-22
manually send req, this ensures we send it before JOIN

### LINKED ISSUE #1229: Messages received while recent messages load are grey and with no badges
**Describe the bug**
While the recent messages load (or, in other words, before they are loaded in) messages received via the normal IRC connection are missing color and badges.

**To Reproduce**

Open chatterino2, and look at a fast chat around the time of the connect.

**Expected behavior**

Messages load in with color and badges.

**Screenshots**
image
image

**Chatterino version**
2.1.0 (21.08.2019 59b3242-1) which is equivalent to commit c81e1929c9dffab64f58e5e676dc5623c14036b0

**Additional context**

pajlada:
I send the connection itself as a parameter to onConnected now since TwitchServer could not access readConnection, the other alternatives/additions I can think of are:
keep current solution, but rename onConnected to onReadConnectionConnected
replace current solution, make readConnection protected so TwitchServer can access it

Fixes #1229

### DIFF
--- a/src/providers/irc/AbstractIrcServer.cpp
+++ b/src/providers/irc/AbstractIrcServer.cpp
@@ -36,9 +36,9 @@ AbstractIrcServer::AbstractIrcServer()
     QObject::connect(this->readConnection_.get(),
                      &Communi::IrcConnection::privateMessageReceived,
                      [this](auto msg) { this->privateMessageReceived(msg); });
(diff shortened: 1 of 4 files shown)


pajlada:
Looks good to me! :+1:

PR #1237, opened 2019-08-23
Improvements to Message Search

leon-richardt:
**EDIT:** So this PR kind of changed direction regarding what it was originally intended to do. Check this post's history to see the original request.

### Overview

This PR aims to improve message search in a channel, inspired by the way Discord does it.
As a proof of concept, two tags have been implemented so far:

Tag | Description | Example
----- | ----------------- | ------------
`from:` | Show messages sent by specific users | `from:pajlada,fourtf` would only show messages sent by pajlada or fourtf
`has:link` | Show messages containing a link | `has:link` would only show messages containing a link

(See the Screenshots section for some examples.)

In an attempt to be modular about new tags I use a predicate system (see [src/messages/predicates](https://github.com/Chatterino/chatterino2/tree/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates)). Each predicate describes one specific feature a message can have. When calling [`SearchPopup::performSearch`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/widgets/helper/SearchPopup.cpp#L86) the search query is parsed and the desired predicates are generated. Afterwards, every message satisfying all parsed predicates is added to the search result.

### Predicates

Predicates need to be derived from [`MessagePredicate`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates/MessagePredicate.hpp). [`MessagePredicate::appliesTo(const MessagePtr)`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates/MessagePredicate.hpp#L32) must contain the logic deciding whether a given message fulfills the predicate.
So far, three `MessagePredicates` have been implemented: [`AuthorPredicate`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates/AuthorPredicate.hpp), [`LinkPredicate`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates/LinkPredicate.hpp) and [`SubstringPredicate`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/messages/predicates/SubstringPredicate.hpp).

Their implementation is straightforward and (hopefully well enough) documented in-source.

When implementing new predicates, the logic for parsing their tags should be added to [`SearchPopup::parsePredicates`](https://github.com/Chatterino/chatterino2/blob/ad7d7cb2414587197e80f1f8f897daa25a07f858/src/widgets/helper/SearchPopup.cpp#L119).

### Screenshots
<details>
<summary>Click to expand</summary>

Searching for a substring

Searching for every message sent by two users

Searching for every message sent by a specific user that also contains a certain substring

Searching for every message sent by two users that also contains a certain substring

Searching for every message that contains a link and was sent by one of two users

Searching for every message that contains a link, was sent by a specific user and contains a certain substring

</details>

### Open Questions

- What is the best way to inform users about these features? (Entry in the GitHub wiki? In-app tooltip?)
- What other kind of tags would be useful to have?


pajlada:
I would prefer some other way to search instead of using the `@`-prefix. I've come to like the way Discord does it, where you can search for `from:pajlada`, `contains:emoji`, `mentions:pajlada` - I feel like it's a lot easier if we start off using that standard to extend this feature.
Otherwise, super cool feature. would love to press "merge pull request"

leon-richardt:
I agree, that's seems like the most reasonable way to go about this. :)

leon-richardt:
So, this PR as been updated now. Happy to get some feedback on the proposed changes!

leon-richardt:
I fixed the merge conflicts introduced by dd996d1. Please check again whether these changes are okay. Predicate tags also work in the `LogsPopup` search now.

When implementing future predicates, they should definitely be thoroughly tested in the logs search as well, since currently, only the `messageText` and `displayName` attributes of a `Message` are set when retrieving Overrustle Logs.

leon-richardt:
I can't figure out why so many lines are changed in `chatterino.pro`. Locally, it only shows diffs for the added files.

leon-richardt:
Finally managed to fix the giant diff mess in `chatterino.pro`.

src/messages/predicates/MessagePredicate.hpp
```diff
@@ -0,0 +1,36 @@
...
+     * Implementations of `appliesTo` should never change the message's content
+     * in order to be compatible with other MessagePredicates.
+     *
+     * @param message the message to check for this predicate
+     * @return true if this predicate applies, false otherwise
+     */
+    virtual bool appliesTo(const MessagePtr message) = 0;
+};
+
+using MessagePredicatePtr = std::shared_ptr<MessagePredicate>;
```
fourtf: This should be removed after switching to unique_ptr

src/widgets/helper/SearchPopup.cpp
```diff
@@ -113,4 +129,78 @@ void SearchPopup::initLayout()
     }
 }
 
+std::vector<MessagePredicatePtr> SearchPopup::parsePredicates(
+    const QString &input)
+{
+    std::vector<MessagePredicatePtr> predicates;
```
fourtf: This should be a `std::vector<std::unique_ptr<MessagePredicate>>`  since the predicates aren't shared.

src/widgets/helper/SearchPopup.cpp
```diff
@@ -113,4 +129,78 @@ void SearchPopup::initLayout()
     }
 }
 
+std::vector<MessagePredicatePtr> SearchPopup::parsePredicates(
+    const QString &input)
```
fourtf: The functions here are repetitive: The text gets split and iterated in 3 different functions and there seem to be multiple ways to find key:value tags.

I suggest something like this:
```c++
std::vector<MessagePredicatePtr> SearchPopup::parsePredicates(
    const QString &input)
{
    static QRegularExpression predicateRegex(R"(^(\w+):(\w+)$)");

    auto words = input.split(' ', QString::SkipEmptyParts);
    // turn shared into unique ptr later!
    auto preds = std::vector<std::shared_ptr<MessagePredicate>>();
    auto authors = QStringList();

    for (auto it = words.begin(); it != words.end();)
    {
        if (auto match = predicateRegex.match(*it); match.hasMatch())
        {
            QString name = match.captured(1);
            QString value = match.captured(2);

            bool remove = true;

            // match predicates
            if (name == "from")
            {
                authors.append(value);
            }
            else if (name == "has" && value == "link")
            {
                preds.push_back(std::make_unique<LinkPredicate>());
            }
            else
            {
                remove = false;
            }

            // remove or advance
            it = remove ? words.erase(it) : ++it;
        }
    }

    if (!authors.empty())
        preds.push_back(std::make_unique<AuthorPredicate>(authors));

    if (!words.empty())
        preds.push_back(std::make_unique<SubstringPredicate>(words.join(" ")));

    return preds;
}

```

src/messages/predicates/AuthorPredicate.cpp
```diff
@@ -0,0 +1,18 @@
...
+namespace chatterino {
+
+AuthorPredicate::AuthorPredicate(const QStringList &authors)
+    : authors_(authors)
+{
+}
+
+bool AuthorPredicate::appliesTo(const MessagePtr message)
+{
+    if (!message)
```
fourtf: This check can be avoided by taking a `Message&` instead of a `MessagePtr` as a parameter instead.

chatterino.pro
```diff
@@ -122,6 +122,9 @@ SOURCES += \
     src/messages/MessageColor.cpp \
     src/messages/MessageContainer.cpp \
     src/messages/MessageElement.cpp \
+    src/messages/predicates/AuthorPredicate.cpp \
```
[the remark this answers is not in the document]
leon-richardt: Resolved in 7fc41602.

src/messages/predicates/LinkPredicate.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace chatterino {
+
+LinkPredicate::LinkPredicate()
+{
+}
+
+bool LinkPredicate::appliesTo(const MessagePtr message)
+{
+    // Borrowed and slightly abridged from https://stackoverflow.com/a/3809435
+    static const QRegExp urlRegex(
```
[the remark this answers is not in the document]
leon-richardt: Resolved in 74f50635.

src/messages/predicates/MessagePredicate.hpp
leon-richardt: Resolved in 5cd1683a.

src/widgets/helper/SearchPopup.cpp
leon-richardt: Resolved in 5cd1683a.

src/widgets/helper/SearchPopup.cpp
leon-richardt: Resolved in 6e9dbc1e.

src/messages/predicates/AuthorPredicate.cpp
leon-richardt: Resolved in 84be38f8.

src/messages/search/AuthorPredicate.cpp
```diff
@@ -0,0 +1,23 @@
...
+        for (const auto &author : entry.split(',', QString::SkipEmptyParts))
+        {
+            this->authors_ << author;
+        }
+    }
+}
+
+bool AuthorPredicate::appliesTo(const Message &message)
+{
+    return authors_.contains(message.displayName, Qt::CaseInsensitive);
```
fourtf: This should check if it contains either the `displayName` or the `loginName` since users can have localized display names which are different than their login names.

leon-richardt: Resolved in 97ec3433.

Issue #683, opened 2018-08-27
Better user cards


seriousm4x:
* add support to make the buttons on the user card customizable https://imgur.com/a/TXgcz5w
* show previous messages inside the user card from that user (like in bttv) https://imgur.com/a/81A1CCQ

Chronophylos:
you can use `![](link_to_image)` to embed an image

fourtf:
customization of timeout buttons and previous messages is being developed here: https://github.com/Chatterino/chatterino2/pull/1238

PR #1238, opened 2019-08-24
Usercard: customizable timeout buttons

TranRed:
**@23rd thanks for the help with refactoring my shitty code! :)**

This PR implements the a required features from #683. I also added a border to some buttons (i.e. Online Logs & Usercard).

**Customizable timeout buttons**
Currently looks like this:
Usercard:
grafik

Settingspage: 
(for simplicity it only allows values from 1-99. Arbitrary timeouts, e.g. 123 minutes instead of 2 hours can still be done via mod button or /timeout command)
grafik


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -536,25 +554,19 @@ UserInfoPopup::TimeoutWidget::TimeoutWidget()
...
-    addTimeouts("days", {
-                            {"1", 1 * 60 * 60 * 24},
-                            {"3", 3 * 60 * 60 * 24},
-                        });
-    addTimeouts("weeks", {
-                             {"1", 1 * 60 * 60 * 24 * 7},
-                             {"2", 2 * 60 * 60 * 24 * 7},
-                         });
+    const auto timeoutButtons = getSettings()->timeoutButtons.getValue();
+    std::vector<std::pair<QString, int>> t(8);  // Timeouts.
```
23rd: Oh, I missed that `std::pair<QString, int>` could be replaced with `TimeoutButton`.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -536,25 +554,19 @@ UserInfoPopup::TimeoutWidget::TimeoutWidget()
...
+    auto i = 0;
+    std::generate(t.begin(), t.end(), [&] {
+        const auto tButton = timeoutButtons[i];
+        std::pair<QString, int> pair = std::make_pair(
```
23rd: And that could be simplified with `const auto`.

Mm2PL:
Looks really cool so far.

TranRed:
After a lot of refactoring by 23rd, I think the code is in a way better state than before.

src/widgets/settingspages/AdvancedPage.cpp
```diff
@@ -70,9 +71,98 @@ AdvancedPage::AdvancedPage()
                          []() mutable {
                              getSettings()->cachePath = "";  //
                          });
-
+    }
         // Logs end
+
+
+        // Timeoutbuttons
+    {
```
fourtf: All other settings leverage the SignalVectorModel for collection settings so this seems really out of place

src/widgets/dialogs/UserInfoPopup.hpp
```diff
@@ -35,15 +37,16 @@ class UserInfoPopup final : public BaseWindow
     QString userName_;
     QString userId_;
     ChannelPtr channel_;
+    ChannelView *latestMessages_;
```
fourtf: `ChannelView *latestMessages_{}`

Adding the `{}` sets the default value to `nullptr` instead of a random int which makes debugging easier. (This works for most types from "C" like ints and pointers, C++ types do not need this)

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -568,4 +556,25 @@ void UserInfoPopup::TimeoutWidget::paintEvent(QPaintEvent *)
     //    / 2);
 }
 
+void UserInfoPopup::fillLatestMessages()
```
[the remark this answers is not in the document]
TranRed: Hope I did this one correctly. :)

src/widgets/settingspages/AdvancedPage.cpp
TranRed: true, will do a rework.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -25,6 +28,46 @@
 #define TEXT_VIEWS "Views: "
 #define TEXT_CREATED "Created: "
 
+namespace {
```
fourtf: You can also put the anonymous namespace into the chatterino namespace.

src/widgets/splits/Split.cpp
```diff
@@ -636,6 +636,16 @@ void Split::showViewerList()
     viewerDock->activateWindow();
 }
 
+void Split::openSubPage()
```
23rd: I think you should remove the first commit that merges your other branch.

TranRed: True, I only merged that so I can use it in my own compiled version anyway. Not related to this feature.

23rd: In my opinion, it would be better to really drop the first (d3aa3694f2d305b6f38f71809efb5dfc87b51061) and last (18349ed9b3f35ccb59c658a04e2c71b6817c08e3) commits to make the history of changes look a little bit nicer.
One of the many ways to do this is to use `git rebase -i HEAD~25`, manually delete the first and last lines and save the result and then use `git push --force`.
(Or just use any gui git client.)

fourtf:
i could also squash it into a single commit when merging

TranRed:
Considering the already large number of commits I personally would prefer  squashing as well. Keeps history in Master cleaner imho.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -337,6 +386,9 @@ void UserInfoPopup::setData(const QString &name, const ChannelPtr &channel)
     this->updateUserData();
 
     this->userStateChanged_.invoke();
+
+    this->latestMessages_->setChannel(
+        filterMessages(this->userName_, this->channel_));
```
pphop: Should check if `latestMessages_` already holds a channel and delete the old one to not leak it since the filterMessages creates a new channel every time it is called.

5E7EN:
Not sure on the progress of this PR, but it seems like a useful feature nonetheless... 
🤜 BUMP

TranRed:
no progress, but I agreed with fourtf to keep the PR open.
Maybe someone who knows what he's doing can just pick the idea up and reimplement from scratch.

0xRainy:
Sorry about the refs, you can delete them all. Dunno why github does this, when those aren't even around anymore.

TranRed:
latest messages merged with  #1729 How to proceed with the other half? do I close this PR? 🤔

zneix:
I think there's no other PR / Issue related to customizable timeout buttons and I know some users wanted that feature, so maybe it's a good thing to continue with just buttons?

TranRed:
I feel like it would be better to start a clean PR, than continue with this mess. The latest messages part, was the working one anyway.

zneix:
If you feel like it's better to start a fresh PR then go for it :)

0xRainy:
Would love to see you continue the custom timeout feature, been wanting something like this along with keybindings for the buttons in the userinfopopup for a long time.

Click their name, hit a key to issue a custom timeout 👍

fourtf (approved):

TranRed:
Updated PR name and start post

PR #1242, opened 2019-08-25
Fixed Travis build for macOS.

23rd:
This quick PR fixes Travis build for macOS.
Clang on macOS has some problems with `std::optional`. =(
We cannot access the value with the `value()` in macOS, and as a workaround we should use `*should`.


fourtf:
switched to boost::optional in https://github.com/Chatterino/chatterino2/commit/6cb82307e659a260612d228e8f893eb6456f8745

PR #1245, opened 2019-08-26
Moved to static method a casting from Channel to TwitchChannel.

23rd:
There are quite a lot of castings in the code from the Channel to the TwitchChannel, so I think it would be better to move this operation to a separated method.


fourtf:
Having dynamic_casts in almost all places that channels definitely indicates a problem. However, this change doesn't fix the problem. It just makes it seem like the problem is intended. The problem is that twitch related code is in places that it should not be in. The CompletionModel, for example, should not have to know what a TwitchChannel is, it just needs to be able to get all the completion items. This issue would get worse if other providers like Mixer or IRC were added.

PR #1247, opened 2019-08-26
Add user IDs to user cards

### LINKED ISSUE #1243: Add an option to show user ids in user cards

I would like for the client to have an option to show user ids in user cards.

 * Add a checkbox to the settings UI.
 * If the checkbox is checked user IDs will be displayed in user cards on another line, otherwise nothing will be changed compared to the current user card.

* A command to show the user ID

**Additional context**

Photoshop of how it could look like.
screencap_temp

Possibility of adding a copy button to the user ID as well

Mm2PL:
Closes #1243 
This PR adds a user ID field into the user card and a setting to toggle it. The default is disabled not to bother people who don't need it.

Edit: I'm not able to add a copy button in this PR. I will submit a new one when this gets merged that adds the copy button.

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -70,11 +70,7 @@ UserInfoPopup::UserInfoPopup()
                 .assign(&this->ui_.followerCountLabel);
             vbox.emplace<Label>(TEXT_CREATED)
                 .assign(&this->ui_.createdDateLabel);
-            if (getSettings()->showUserIDsInUsercards)
-            {
-                vbox.emplace<Label>(TEXT_USER_ID)
-                    .assign(&this->ui_.userIDLabel);
-            }
+            vbox.emplace<Label>(TEXT_USER_ID).assign(&this->ui_.userIDLabel);
         }
     }
 
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -355,9 +355,13 @@ void UserInfoPopup::updateUserData()
 
         this->userId_ = id;
 
-        this->ui_.userIDLabel->setText(TEXT_USER_ID + this->userId_);
-        // don't wait for the request to complete, just put the user id in the card
-        // right away
+        // this can be nullptr if the setting is not enabled.
+        if (this->ui_.userIDLabel)
+        {
+            this->ui_.userIDLabel->setText(TEXT_USER_ID + this->userId_);
+            // don't wait for the request to complete, just put the user id in the card
+            // right away
+        }
 
         QString url("https://api.twitch.tv/kraken/channels/" + id);
 
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -14,6 +14,7 @@
 #include "widgets/dialogs/LogsPopup.hpp"
 #include "widgets/helper/EffectLabel.hpp"
 #include "widgets/helper/Line.hpp"
+#include "singletons/Settings.hpp"
 
 #include <QCheckBox>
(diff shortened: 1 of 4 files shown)


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -68,6 +70,10 @@ UserInfoPopup::UserInfoPopup()
                 .assign(&this->ui_.followerCountLabel);
             vbox.emplace<Label>(TEXT_CREATED)
                 .assign(&this->ui_.createdDateLabel);
+            if (getSettings()->showUserIDsInUsercards)
+            {
+                            vbox.emplace<Label>(TEXT_USER_ID).assign(&this->ui_.userIDLabel);
+                        }
```
23rd: I think it would be cool to reduce the indentation.

Mm2PL: And this reminds me to run clang-format.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -348,6 +355,10 @@ void UserInfoPopup::updateUserData()
 
         this->userId_ = id;
 
+        this->ui_.userIDLabel->setText(TEXT_USER_ID + this->userId_);
```
23rd:
```
if (this->ui_.userIDLabel)
{
    this->ui_.userIDLabel->setText(TEXT_USER_ID + this->userId_);
}
```

23rd:
By the way, it would be better to have ability to copy ID, otherwise it is not useful to just see the sequence of numbers.

src/widgets/dialogs/UserInfoPopup.cpp
Mm2PL: I don't understand why do that. Can you please explain?

23rd: Oh, sorry, I forgot to explain.

`this->ui_.userIDLabel` exists only when `getSettings()->showUserIDsInUsercards` is `true`. So when it is `false`, then `this->ui_.userIDLabel` is `nullptr` and you call `setText()` from empty object.

I just built your branch, launched the chatterino, opened the usercard, and got crash just because the `showUserIDsInUsercards` setting was not set.

Mm2PL: Oh thank you for expalaining. I will fix it ASAP

apa420:
I'd prefer if this wasn't a toggle able setting and just implemented instead.
- It won't clutter the usercard that much.
- Might be confusing for an average twitch user to see it in the setting

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -348,6 +351,14 @@ void UserInfoPopup::updateUserData()
 
         this->userId_ = id;
 
+        // this can be nullptr if the setting is not enabled.
+        if (this->ui_.userIDLabel)
```
pajlada: shouldn't have to check this now since the setting is removed

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -8,6 +8,7 @@
 #include "providers/twitch/PartialTwitchUser.hpp"
 #include "providers/twitch/TwitchChannel.hpp"
 #include "singletons/Resources.hpp"
+#include "singletons/Settings.hpp"
```
apa420: Don't have to include this anymore since you're not using settings

src/widgets/dialogs/UserInfoPopup.cpp
Mm2PL: done

src/widgets/dialogs/UserInfoPopup.cpp
Mm2PL: removed

PR #1255, opened 2019-08-27
Fetch FFZ channel emotes with channel id instead of name

### LINKED ISSUE #1253: Use user ids in the ffz api
Using usernames is prone to username changes.

Example: https://api.frankerfacez.com/v1/room/id/22580017

apa420:
Since the ffz api now requires id we need to fetch the id first before we can fetch the emotes. That's why bttv and ffz emotes won't load at the same time.
Closes #1253


fourtf:
merged in command line

Issue #1280, opened 2019-09-03
Show online logs in search window


fourtf:
**Is your feature request related to a problem? Please describe.**
When looking at the logs of someone it takes another action to search in the logs.

**Describe the solution you'd like**
Show search input right away in the log popup

**Additional context**
If possible reduce code duplication. Maybe do LogsPopup extends SearchPopup hierachy.

fourtf:
https://github.com/Chatterino/chatterino2/pull/1281

PR #1281, opened 2019-09-03
Allow searching in LogsPopup

fourtf:
LogsPopup now extends SearchPopup.


fourtf:
Would fix  #1280

fourtf:
Would fix  #1280

fourtf:
Would fix  #1280

pajlada:
Looks good to me, will be great with #1237

Issue #1291, opened 2019-09-07
Preview on Hover: Not showing the correct size of a FrankerFaceZ custom mod badge + preview of vip, regular moderator, and broadcaster badge not appearing


rvdog815:
Describe the bug
The Preview on Hover feature is not displaying the correct size preview for custom FrankerFaceZ moderator badges. also after initially opening this issue i have noticed that the preview for the Broadcaster badge and the VIP badge aren't showing up

To Reproduce
Go to settings turn on Preview on Hover and then go to a channel that has a custom FrankerFaceZ moderator badge larger than 18x18 for example Trihex

Expected behavior
The Preview on Hover would show the correct size preview of the custom FrankerFaceZ moderator badge

Screenshots
image of preview viewed on twitch website: ![image](https://user-images.githubusercontent.com/35497363/64466655-f771a400-d0c7-11e9-9711-7413009e3800.png)

image of preview on the latest nightly: ![image](https://user-images.githubusercontent.com/35497363/64466749-79fa6380-d0c8-11e9-8d88-0e0e8371793d.png)

image of preview of VIP badge on the latest nightly: ![image](https://user-images.githubusercontent.com/35497363/64468396-33613500-d0d9-11e9-826f-00ef558d3772.png)

image of preview of Broadcaster badge on the latest nightly: ![image](https://user-images.githubusercontent.com/35497363/64468439-86d38300-d0d9-11e9-8ecb-06e395567393.png)

image of preview of regular Moderator badge on the latest nightly: ![image](https://user-images.githubusercontent.com/35497363/64469208-e5512f00-d0e2-11e9-9f01-054cf8c390a6.png)

Chatterino version
2.1.4-beta-1 (03.09.2019 9ee286f)

pajlada:
Currently looking into fetching the FFZ mod badge more intelligently

pajlada:
#1293 should solve the issue of the FFZ mod badges not scaling properly.
For the other badges (VIP, Broadcaster, Moderator) we just need to make them into `BadgeElement`s instead of `ImageElement`s for them to be "hoverable".

Mm2PL:
Can confirm, broken on 2.1.4 ce56653e. This happens also with the `Twitch Prime Subscriber` badge.

fourtf:
can;t reproduce

PR #1293, opened 2019-09-07
Load mod badge information from the FFZ API instead of assuming the URL.

pajlada:
This lets us load all sizes of the emote if they are available.

Channel with all versions of the mod badge: https://api.frankerfacez.com/v1/room/pajlada
Channel with only one version of the mod badge: https://api.frankerfacez.com/v1/room/apa420
Channel with no mod badge: https://api.frankerfacez.com/v1/room/forsen

A few "controversional" things I've done:
1) Image now has a custom "On successful network request" call, which the FFZ mod badge image uses to inject the green mod background. This is the `customOnSuccess` stuff in Image. As a result of this, `setPixmap` needed to be made public. I couldn't come up with a better solution
2) FFZ mod badge lives in the `FfzEmotes` file, so the name of that file isn't super logical anymore.
3) `FfzEmotes::loadChannel` needs to return information about the mod badge, so I added a new callback to it. I could've also just added it as a parameter to the first callback. I can easily do that upon request, just wasn't sure which felt cleaner (still not sure).

This PR is a response to #1291


fourtf:
On first glance I dislike 1) since this should IMO be handled in a custom `MessageElement` similar to https://github.com/Chatterino/chatterino2/blob/master/src/messages/MessageElement.hpp#L203

pajlada:
@fourtf Are you thinking a custom `MessageElement` type and a custom `ImageLayoutElement` that draws the green mod badge under it?
e.g. `ModBadgeElement` which in addToContainer adds a `ImageWithBackgroundLayoutElement` element which in turn has a paint element that not only draws its own "image pixmap, but also in this case the green mod badge background?

fourtf:
@pajlada, yes

fourtf:
@pajlada I think that reflects the "real world" situation where usually an emotes or badge is just an image but in this instance it's an image and a background. For the OOP structure it would make even more sense to make them extend EmoteElement and ImageLayoutElement

pajlada:
@fourtf cool, will do

src/providers/ffz/FfzEmotes.cpp
```diff
@@ -79,8 +79,37 @@ namespace {
 
         return {Success, std::move(emotes)};
     }
-    std::pair<Outcome, EmoteMap> parseChannelEmotes(const QJsonObject &jsonRoot)
+    std::tuple<Outcome, EmoteMap, boost::optional<EmotePtr>>
+        parseChannelResponse(const QJsonObject &jsonRoot)
```
fourtf: Making a `parseChannelEmotes` and a `parseModBadge` function separately would work just as well. Here, The result is a tuple so need to either have comments or read the code to understand what the tuple items are.

fourtf: Also, unrelated to this PR, the Outcome in the tuple could be removed it is unused.

fourtf:
looks good to me

PR #1309, opened 2019-09-12
Sort File Names When Updating File List

leon-richardt:
This PR sorts the acquired sources and headers list before writing it to the project file.
This should prevent unwanted diffs and guarantee consistent output.

To test, checkout this branch and run `python3 update_filelist.py`. If everything works correctly, no diffs should be produced in the `chatterino.pro` file.

### DIFF
--- a/update_filelist.py
+++ b/update_filelist.py
@@ -10,12 +10,16 @@ data = ""
 
 with open(filename, 'r') as project:
     data = project.read()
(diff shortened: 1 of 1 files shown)


leon-richardt:
@Mm2PL reported that his file list gets sorted differently. He is on an older `sort` version though. Additional testing much appreciated.

leon-richardt:
Now uses built-in Python sorting instead of relying on the `sort` command.

Issue #874, opened 2018-11-13
/version command


pajlada:
Built-in command to print the version of chatterino you're using.
Not sure whether it should print a system-message, or actually print the message in chat.

Example output would be something like:
`Chatterino 2.0.4 Nightly (dcda714), built January 3rd 2015 16:05:21`

fourtf:
This should only happen if a lot more features (such as settings) are exposed through text commands. Just having this one command would feel out of place.

TranRed:
I would expect to see this information in the "about" tab.

pajlada:
Version information is available in the About tab now, which is fine by me :ok_hand:

PR #1315, opened 2019-09-16
Add Version Information to "About" Page

leon-richardt:
### Description

This PR adds information about the program version to the "About" page.

As the program's version is also displayed in the window title (and to avoid code duplication), I put the "information parsing logic" in its own class at `common/Version.cpp`. Not sure whether this is the best solution.

The text in the version information box is copyable so that users can easily paste it when filing bug reports.

Also kind of related to #874.

### Screenshot

Version information in the "About" page


leon-richardt:
I hope the rebase doesn't screw up this branch's history. pajaDent

PR #1320, opened 2019-09-19
Better Highlights

### LINKED ISSUE #1216: 🎶 Multiple differentping sounds
No ability to distinguish which highlight plays ping sound. Having multiple ping sounds would solve this. For moderation you might want to get notified on certain patterns which have nothing to do with normal pings.

Allow selecting Ping sounds per highlight.

Don't do it.

**Additional context**

leon-richardt:
### Description

This PR aims to add more options for highlights, such as user-defined colors and sounds for different phrases and users. See screenshots for examples below.

Also resolves #1216, I guess.

### TODO List

- [x] User-defined colors for highlights
- [x] User-defined sounds for highlights
- [x] Accurate scrollbar highlight colors
- [x] Make color picker behave as expected
- [x] Make TableView reasonably good looking

### Additional Considerations & Questions
- [x] Check out FFZ highlighting strategy
    - Looks pretty smiliar to what I'm trying to do.
- [x] How to go about existing subscription messages on theme changes?
   - Theme changes are respected.
- [x] What should happen if a message matches multiple highlights with different sounds/colors?
    - First highlight takes priority.

### Screenshots

Phrase highlights settings page

User highlights settings page

Phrase highlights being applied

User highlights being applied


leon-richardt:
Builds fail on MacOS because (apparently?) the linker command is broken: It uses `-framework\ Security` instead of `-framework Security` and `-framework\ Foundation` instead of `-framework Foundation`. This wasn't the case on previous builds. 🤔

leon-richardt:
I tried rebasing some commits but apparently that messed up the branch history. 
So I guess squashing is required.

pajlada:
Testing this now

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -939,23 +954,33 @@ void TwitchMessageBuilder::parseHighlights()
...
         {
             this->highlightAlert_ = true;
         }
 
-        if (userHighlight.getSound())
+        if (userHighlight.hasSound())
         {
             this->highlightSound_ = true;
+            // Use custom sound if set, otherwise use the fallback sound
+            if (!userHighlight.getSoundUrl().toString().isEmpty())
```
pajlada: Replace with a `HighlightPhrase::hasCustomSound` that does the same thing

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -77,6 +96,7 @@ TwitchMessageBuilder::TwitchMessageBuilder(
     , action_(_ircMessage->isAction())
 {
     this->usernameColor_ = getApp()->themes->messages.textColors.system;
+    this->highlightSoundUrl_ = getFallbackHighlightSound();
```
pajlada: We shouldn't initialize `highlightSoundUrl_` if we don't need it. If we really need to initialize it in the constructor, I'd rather we just set it to `QUrl("qrc:/sounds/ping2.wav")` instead of making filesystem calls in every constructor.

src/widgets/helper/ScrollbarHighlight.hpp
```diff
@@ -6,17 +6,49 @@ class ScrollbarHighlight
 {
 public:
     enum Style : char { None, Default, Line };
-    enum Color : char { Highlight, Subscription };
+    enum Type { Highlight, Subscription, Invalid };
```
pajlada: should be possible to make this an `enum class`

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -92,7 +99,8 @@ HighlightingPage::HighlightingPage()
                 // Case-sensitivity doesn't make sense for user names so it is
                 // set to "false" by default & no checkbox is shown
                 view->setTitles({"Username", "Flash\ntaskbar", "Play\nsound",
```
pajlada: Makes sense to make an additional comment/modify the existing comment now that "Case-sensitivity" is part of the titles here, despite it not being rendered

pajlada:
Two more things we mentioned in chat, that I'll post here as well for book-keeping:
 - Setting a custom sound on a highlight should enable the "Play sound" column too if it's disabled. Note for future self: Make sure to test what happens if we clear the custom sound, ensure it doesn't enable play sound column then.
- Explore possibility of adding alpha to the custom color picker. if that works, that'd be dank

src/providers/twitch/TwitchMessageBuilder.cpp
leon-richardt: Turns out we don't need this at all. :sweat_smile:

Rsolved in d44d803f.

src/controllers/highlights/HighlightModel.hpp
```diff
@@ -13,6 +13,18 @@ class HighlightModel : public SignalVectorModel<HighlightPhrase>
 {
     explicit HighlightModel(QObject *parent);
 
+public:
+    // Used here and in UserHighlightModel
```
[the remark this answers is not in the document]
leon-richardt: Resolved in d938649d.

src/providers/twitch/TwitchMessageBuilder.cpp
leon-richardt: Resolved in d938649d.

src/widgets/helper/ScrollbarHighlight.hpp
leon-richardt: Resolved in 399ff7da.

src/widgets/settingspages/HighlightingPage.cpp
leon-richardt: Hope I understood this correctly. If so, resolved in 18a3b599.

leon-richardt:
Don't merge yet, need to check if ScrollbarHighlights still work after https://github.com/Chatterino/chatterino2/pull/1320/commits/ace0fe0a25b6cd5b5fb3edab8e44bedf28206848.

fourtf:
First off this is a great addition and allows more useful customization. Unfortunately since this is an unexplored area in Chatterino this will need some refinement until it can be merged into master. Software development is an iterative process after all.

First off there are some usability issues:
- Default color looks unpleasant in [light mode](https://i.imgur.com/ADY3z14.png)
- No way to reset color to default or even select the default color without knowing the exact color code.
- No way to change color of subs.
- Highlight colors don't update after changing the color.
- I can't imagine using any of the [default colors](https://i.imgur.com/7gFBask.png). I propose that instead there should be a set of default colors which have the same lightness as the default color and work in light and dark mode.
- "Custom colors" in the color dialog seems dated, instead I'd like to see it removed or replace with "recent colors".
- I feel like "Play sound" and "Custom sound" could be merged into a single column.
- The two new columns not being selecteable [looks out of place](https://i.imgur.com/gafnl2V.png)

Other things:
- In the code subscription messages seem to be treated different than highlights. I propose that subscriptions get added to the list of custom highlight messages (like own username/whispers)
- Do we need the #rrggbb:aaa format? Is QColor::name(QColor::HexArgb) and QColor's constructor not sufficient?

leon-richardt:
Those are all good points. I will try to address them over the next week but it might take me a while to get around every single one.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -939,23 +952,33 @@ void TwitchMessageBuilder::parseHighlights()
...
         }
 
         if (this->highlightAlert_ && this->highlightSound_)
         {
-            // Break if no further action can be taken from other
-            // usernames Mostly used for regex stuff
-            break;
+            // Usernames "beat" highlight phrases: Once a username highlight
+            // has been applied, no further highlight phrases will be checked
+            return;
```
Mm2PL: I think this should be communicated through the UI.

leon-richardt:
Sorry, this is taking so long. Quick status update on what has been achieved so far:

- Subscriptions have been added to the preset highlights (in the same category as self highlights and whispers). This means that you can now also set a custom color and sound for subscription messages.

- Highlight colors are updated after changing the color (only on already highlighted messages, old messages aren't parsed again).

- Columns have been given the `Qt::ItemIsSelectable` flag and now look consistent with the other columns.

- Serialization of highlight colors makes use of the `QColor::name(QColor::HexArgb)` instead of using an own format.

Updated TODO:

- [ ] Devise a set of default colors that work well in dark and light mode.

- [ ] Design an own color picker dialog that is more suited to this use case (e.g. recent colors or option to reset to default color).

- [ ] Figure out a sensible way to merge "Play sound" and "Custom sound" colors.

src/providers/twitch/TwitchMessageBuilder.cpp
leon-richardt: See 84dd044.

Mm2PL:
I found some weird behavior: when importing settings from an old version all highlights will be black, not the old red/pinkish color.

leon-richardt:
> I found some weird behavior: when importing settings from an old version all highlights will be black, not the old red/pinkish color.

This is probably due to the change from the "custom" format of #rrggbb:aaa to ´QColor::name(QColor::HexArgb)`. This shouldn't be much of a problem when merged because there are no old settings to import.

Mm2PL:
It has problems when importing from the current `master`.
monkaS

leon-richardt:
> Can this be a helper function?

The whole part here is still under heavy development, so much may change here. I will definitely re-request a review once I feel like I'm happy with the state of the code.

leon-richardt:
> It has problems when importing from the current `master`.
> monkaS

It should have been fixed by  b87b760!

src/widgets/dialogs/ColorPickerDialog.hpp
```diff
@@ -0,0 +1,90 @@
...
+namespace chatterino {
+
+/**
+ * @brief A custom color picker dialog.
+ *
+ * This class exists because QColorPickerDialog did not suit our use case.
+ * This dialog provides buttons for recently used and default colors, as well
+ * as a color picker widget identical to the one used in QColorPickerDialog.
+ */
+class ColorPickerDialog : public BaseWindow
```
pajlada: This can be made into a `BasePopup` instead of a `BaseWindow` - it will make pressing "Escape" act as pressing the "Cancel" button.

leon-richardt:
Okay, most of the stuff mentioned in https://github.com/Chatterino/chatterino2/pull/1320#issuecomment-546689864 is done now. I couldn't figure out a nice way to merge the "Play Sound" and "Sound Path" columns though.
I attached some screenshots of the default colors in light and dark theme, as well as a picture of the color picker.

I would appreciate a re-review of this PR. :smile: 

<details>
  <summary>Screenshots</summary>

  Default colors in dark theme

Default colors in light theme

Custom color picker dialog

</details>

src/widgets/dialogs/ColorPickerDialog.hpp
leon-richardt: See 61d5fdb.

src/widgets/dialogs/ColorPickerDialog.cpp
```diff
@@ -0,0 +1,365 @@
...
+            });
+    }
+}
+
+void ColorPickerDialog::initHtmlColor(LayoutCreator<QWidget> &creator)
+{
+    auto html = creator.setLayoutType<QGridLayout>();
+
+    // Copied from Qt source for QColorShower
+    QRegularExpression regExp(
```
pajlada: Will this be compiled every time the function is called?

src/widgets/helper/QColorPicker.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
pajlada: Please include the original license from the files these were copied from. explain the difference as you explained it below

src/widgets/helper/QColorPicker.cpp
```diff
@@ -0,0 +1,244 @@
+#include "widgets/helper/QColorPicker.hpp"
```
pajlada: Please include the original license from the files these were copied from. explain the difference as you explained it in the header file

src/widgets/dialogs/ColorPickerDialog.cpp
leon-richardt: Good catch – made `static` in f4d82f2.

src/widgets/helper/QColorPicker.hpp
leon-richardt: See 0d51d90.

src/widgets/helper/QColorPicker.cpp
leon-richardt: See 0d51d90.

Mavvi87:
My request have been answered HandsUp
But now on to some other question............ how do I implement this to my chatterino?

leon-richardt:
> how do I implement this to my chatterino?

Since this feature is not part of a release yet, you would have to opt for the [nightly builds](https://github.com/Chatterino/chatterino2/releases/tag/github-actions-nightly) for the time being.

Mavvi87:
Ah, so this will be featuring in next update? Basicly just have to wait?

leon-richardt:
> Ah, so this will be featuring in next update? Basicly just have to wait?

Yes, if no major issues pop up with this feature, it will be included in a future release. It's hard to give an ETA on that though.

Mavvi87:
> > Ah, so this will be featuring in next update? Basicly just have to wait?
> 
> Yes, if no major issues pop up with this feature, it will be included in a future release. It's hard to give an ETA on that though.

That's awsome, dude! Thank you so much, I can't wait!
I'm just a regular user who loves Chatterino 2 and has it up 24\7, but its so nice to be heard.

Issue #1296, opened 2019-09-07
I dont see any messages while connected to dorm internet


RevanSakero:
Hello,
When I am at dorm, I cant see any messages in chatterino. Its probably because they block some ports at dorm (game ports, discord voice chat ports, etc.). Is it any way how to fix it without using VPN?
Thanks

SLCH:
#450 should solve this.
Btw, do you see messages in usual twitch chat?

RevanSakero:
Yes, I do

RAnders00:
This can be solved by connecting to `irc.chat.twitch.tv:443`, but connecting on port 443 is undocumented.

RevanSakero:
Can you please tell me how to do it? Thank you

RAnders00:
It's a change chatterino would have to make, so unless you want to get into compiling chatterino yourself, you can't do it currently I'm afraid.

In the meantime, a solution like a VPN will most likely work.

RevanSakero:
Ok, thank you

PR #1324, opened 2019-09-21
Allow customization of Twitch chat server parameters through ENV variables

pajlada:
This would leave an option for https://github.com/Chatterino/chatterino2/issues/1296 if compiling their own version isn't viable


fourtf:
We could also add a setting for choosing ws+tsl/tsl/raw irc.

pajlada:
Yup, could also do that. Is there a point in defaulting to 443? It looks like the most compatible one, but I don't know if there are any downsides to using it.

fourtf:
@pajlada that's actually 5Head since it will get around proxies

pajlada:
I am closing the PR and moving default port to 443 instead of 6697

PR #1337, opened 2019-09-24
Fix clang-format Only Reformatting .cpp Files

leon-richardt:
I hope that this isn't depended on the `find` version (similar to how #1309 used to be) but when I run the `clang-format-all.sh` script, only `.cpp` files are operated on. Please check on your own systems.

I confirmed this when replacing the `-exec` part with a `-print`: It does not find any header files.
In this PR, I changed the expression to a regex that matches `.cpp` and `.hpp` files.


fourtf:
I am mildly disappointed that the description wasn't in leet-speak 😞

leon-richardt:
> I am mildly disappointed that the description wasn't in leet-speak 😞

Fixed.

Issue #1264, opened 2019-08-30
show the title when offline


cinemassacres:
example people will type title change PogU in offline chat, but you cant see it in offline chat and know what was changed

RAnders00:
Feature has been added by PR #1347

PR #1347, opened 2019-10-01
Show the title when offline

CommName:
**Description**
This PR solves issue #1264 
The format of the offline tooltip text is different so it can be easily distinguished when a stream is live. Timer for title update is currently set to 15 sec, it might be too low.
**Screenshots**
image

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -89,7 +89,7 @@ TwitchChannel::TwitchChannel(const QString &name,
     , bttvEmotes_(std::make_shared<EmoteMap>())
     , ffzEmotes_(std::make_shared<EmoteMap>())
     , mod_(false)
-    , titleRefreshedTime_(QTime::currentTime().addSecs(-10))
+    , titleRefreshedTime_(QTime::currentTime().addSecs(-titleRefreshPeriod_))
 {
     log("[TwitchChannel:{}] Opened", name);
 
@@ -447,11 +447,11 @@ void TwitchChannel::refreshTitle()
         return;
     }
 
-    if (this->titleRefreshedTime_.elapsed() < 10 * 1000)
+    if (this->titleRefreshedTime_.elapsed() < this->titleRefreshPeriod_ * 1000)
     {
         return;
     }
-    this->titleRefreshedTime_.start();
+    this->titleRefreshedTime_ = QTime::currentTime();
 
     QString url("https://api.twitch.tv/kraken/channels/" + roomID);
     NetworkRequest::twitchRequest(url)
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -463,14 +463,19 @@ void TwitchChannel::refreshTitle()
 
                 const auto document = result.parseRapidJson();
 
-                if (!document.HasMember("status"))
+                auto statusIt = document.FindMember("status");
+
+                if (statusIt == document.MemberEnd())
                 {
                     return Failure;
                 }
 
                 {
                     auto status = this->streamStatus_.access();
-                    status->title = document["status"].GetString();
+                    if (!rj::getSafe(statusIt->value, status->title))
+                    {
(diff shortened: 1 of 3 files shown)


fourtf:
If you are in 50 channels this does 200 extra api requests to twitch per minute and that is not a comfortable amount anymore.

This can only be added if it is loaded on demand when you hover over it.

CommName:
Ok, now it only does a network request when the channel is not live (it didn't load the title). 
There is a noticeable delay (on my machine) if it is loaded on demand.

CommName:
Added on demand, now the question is should I remove the extra api request when a stream is offline or not?

CommName:
Title is not changed that often so the need for extra api request is overkill, so I removed it. 

**What's left:** 
- When id is loaded there is an API request for the title 
- When mouse is over the header of the split there is an on-demand request 

**Known bugs that can occur:**
- If the title is changed while mouse is over the header split title is not automatically changed
- For a slow internet connection, it will show the previous title not a new one he gets from the api request

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -437,6 +438,37 @@ void TwitchChannel::setLive(bool newLiveStatus)
     }
 }
 
+void TwitchChannel::refreshTitle(){
```
pajlada: This function is called from way too many places. It should probably only be called in the roomIdChanged signal, and when the user hovers over the header. And for the "on user hover over header" there needs to be throttling implemented, so the API call can't be spammed over and over again for the same streamer.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -544,6 +550,7 @@ void SplitHeader::updateChannelText()
 
     if (auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get()))
     {
+        twitchChannel->refreshTitle();
```
pajlada: this is the place that I don't want it to be called from, cuz this will be fired 4 times in quick succession when the stream is live and we get new game/title/viewer/live status data

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -437,6 +438,37 @@ void TwitchChannel::setLive(bool newLiveStatus)
...
+    auto roomID= this->roomId();
+    if(roomID.isEmpty()){
```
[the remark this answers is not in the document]
pajlada: Still missing clang-format, if you're on linux you can run `./tools-clang-format-all.sh` - if you give me push permissions to your repo I can just push formatting from my end

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -437,6 +439,51 @@ void TwitchChannel::setLive(bool newLiveStatus)
...
 
+void TwitchChannel::refreshTitle()
+{
+    auto roomID = this->roomId();
+    if (roomID.isEmpty())
+    {
+        return;
+    }
+
+    if (this->titleRefreshedTime_.elapsed() < 10 * 1000)
```
fourtf: 10s should be a constant since it is used in two places.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -437,6 +439,51 @@ void TwitchChannel::setLive(bool newLiveStatus)
...
+    {
+        return;
+    }
+    this->titleRefreshedTime_.start();
```
fourtf: I would prefer assigning `QTime::currentTime()` to it since `start` is very undescriptive of what it does.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -670,6 +680,9 @@ void SplitHeader::enterEvent(QEvent *event)
 {
     if (!this->tooltipText_.isEmpty())
     {
+        dynamic_cast<TwitchChannel *>(this->split_->getChannel().get())
+            ->refreshTitle();
```
fourtf: Check if it is actually a twitch channel (what if we add mixer and it also has titles).

PR #1350, opened 2019-10-02
Fix first line of emotes in emote picker window being shifted.

### LINKED ISSUE #1334: First line of emotes are shifted a bit in emote window
**Describe the bug**
1. First line of emotes are shifted a bit on every tab in emote window
2. Every restart emote window size gets reset (same as in #351)

**To Reproduce**

**Expected behavior**
1. Emotes having the same margin from both sides of a window
2. Emote window position/size should be remembered same as main window is.

**Screenshots**
1. 
image
image

**Chatterino version**
Chatterino Nightly 2.1.5 (commit 367b1970f, 22.09.2019)

Cranken:
Fixes #1334


fourtf:
Seems good to me 👍

PR #1351, opened 2019-10-02
Fix tooltips in chat not showing on ctrl hover.

### LINKED ISSUE #1292: CTRL breaks badge descriptions
**Describe the bug**
Pressing CTRL makes badge descriptions stop appearing when hovering over them.

**To Reproduce**
Press CTRL, hover over the channel badge while still pressing CTRL.

**Expected behavior**
The description should appear even when pressing CTRL.

**Screenshots**
pepega

**Chatterino version**
2.1.3

**Additional context**
Not sure it it's planned behaviour. Seems like a bug to me.

Cranken:
Removes some dead code that prevented tooltips on badges etc. from showing while holding the control/alt modifiers.

Fixes #1292

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1038,14 +1038,6 @@ void ChannelView::leaveEvent(QEvent *)
 
 void ChannelView::mouseMoveEvent(QMouseEvent *event)
 {
-    if (event->modifiers() & (Qt::AltModifier | Qt::ControlModifier))
-    {
-        this->unsetCursor();
-
-        event->ignore();
-        return;
-    }
-
     /// Pause on hover
     if (float pauseTime = getSettings()->pauseOnHoverDuration;
         pauseTime > 0.001f)


pajlada:
Everything seems to work as expected using your branch. However, before I merge this PR, I would like to try to understand what it does/did.

Cranken:
I've looked back to the original commit to understand what this section is supposed to do. It seems like it was something which allowed you to alt+click a split to drag it around. This functionality isn't implemented in any other part anymore. 
The same code segment existed in the mousePressEvent and mouseReleaseEvent where it was removed some time ago.

As the split dragging that is currently implemented works in a different way so this seems like it was just some forgotten part left in.
Maybe @fourtf knows what this part was supposed to do as he originally commited this [here](https://github.com/Chatterino/chatterino2/commit/0a3c6e29c7fd9ab3d28462a32bee42219b2a09fc#diff-a017f9d917094b48d33b9784fadbf5a9R585).

pajlada:
ctrl+alt movement is something I tested as well, and that works on master and on your branch (hold ctrl+alt, press the "move" icon, move split around, release mouse)

fourtf:
The alt+ctrl overlay is now implemented as a widget that's above the split so this is (as stated) not used anymore.

PR #1355, opened 2019-10-06
🐛 Fixed OpenSSL version

AnotherFoxGuy:
The OpenSSL version needs to be the same as described here: https://wiki.qt.io/Qt_5.13_Tools_and_Versions
Otherwise, chatterino2 fails to connect with the message `this platform does not support OpenSSL`

### DIFF
--- a/conanfile.txt
+++ b/conanfile.txt
@@ -1,5 +1,5 @@
 [requires]
-OpenSSL/1.0.2s@conan/stable
+OpenSSL/1.1.1b@conan/stable
 boost/1.69.0@conan/stable
 
 [generators]


fourtf:
Afaik We use two versions of openssl, 1.0.2 and 1.1.1 and I didn't see a way to add both as a dependency to Conan.

AnotherFoxGuy:
What is the feature that requires OpenSSL 1.0.2?

fourtf:
I think it's websocketpp which in turn includes boost asio.

Both the 1.0.2 and 1.1.1 dlls are loaded when chatterino starts so I'm assuming that they are necessary.

AnotherFoxGuy:
https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_WINDOWS.md#openssl
It seems that OpenSSL 1.0 was only needed for QT

fourtf:
I tried running chatterino without the 1.0.2 dlls and everything seems to work so I'll merge this and see if people start complaining that stuff doesn't work.

PR #1375, opened 2019-10-20
Make a command that shows the Chatterino user card.

### LINKED ISSUE #1143: Enhance /user to be able to show the Chatterino 2 Usercard
There should be an option to open the Chatterino 2 Usercard without clicking a user name.
My suggestion would be to enhance /user by a parameter i.e. ```-ch```
Alternatively make a new command like /chuser or /ch2user or similar.

Idea by dyeoxyde on discord:
grafik

Mm2PL:
closes #1143 

Things changed:
  - Added `/usercard` command that opens the Chatterino user card under the cursor.
     - Syntax is `/usercard USER`. Opening a user card with the context of another channel is not possible, yet.
  - Added a failure callback to PartialTwitchUser::getId(), because it was required to show that an account doesn't exist.


src/providers/twitch/PartialTwitchUser.hpp
```diff
@@ -21,6 +21,10 @@ class PartialTwitchUser
 
     void getId(std::function<void(QString)> successCallback,
                const QObject *caller = nullptr);
+
+    void getId(std::function<void(QString)> successCallback,
+               std::function<void(void)> failureCallback,
```
fourtf: Remove the inner `void` so it's `std::function<void()>`.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -384,6 +388,21 @@ void UserInfoPopup::updateUserData()
...
+        // this can occur when the account doesn't exist.
+        this->ui_.followerCountLabel->setText(TEXT_FOLLOWERS +
+                                              QString(TEXT_UNAVAILABLE));
+        this->ui_.viewCountLabel->setText(TEXT_VIEWS +
+                                          QString(TEXT_UNAVAILABLE));
+        this->ui_.createdDateLabel->setText(TEXT_CREATED +
+                                            QString(TEXT_UNAVAILABLE));
+
+        this->ui_.nameLabel->setText(this->userName_ +
+                                     QString("(account doesn\'t exist)"));
```
fourtf: How do we know that the account doesn't exist? What is it was a network error?

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -384,6 +388,21 @@ void UserInfoPopup::updateUserData()
...
+        this->ui_.userIDLabel->setText(QString("ID is not available."));
+        this->ui_.userIDLabel->setProperty("copy-text",
+                                           QString("ID is not available."));
```
fourtf: Why are we using a different error here than in the 3 other labels?

Mm2PL: Using only one now.

src/providers/twitch/PartialTwitchUser.cpp
```diff
@@ -25,8 +25,15 @@ PartialTwitchUser PartialTwitchUser::byId(const QString &id)
 
     return user;
 }
-
+// this override exists not to dirupt other code that might rely on PartialTwitchUser::getId.
```
fourtf: No need for this comment

src/providers/twitch/PartialTwitchUser.cpp
```diff
@@ -25,8 +25,15 @@ PartialTwitchUser PartialTwitchUser::byId(const QString &id)
...
+void PartialTwitchUser::getId(std::function<void(QString)> successCallback,
+                              const QObject *caller)
+{
+    getId(
```
fourtf: Format file

Mm2PL:
I think it's ready for a merge :)

PR #1384, opened 2019-10-23
Switch Global badge Tool-tips from Description to Title

Felanbird:
Currently Chatterino uses the description from the global badge api, which has actively caused "issues" because whomever does this always seems to make a minor or major mistake on the descriptions.

Previously the Bit Leader badges and Direct Relief - Charity 2018 badges were blank because the description fields were blank but they were fixed via a ticket I put in to have them fixed.
image

image

It would also fix the 1000 Gifted Subs badge, as well the recently added founders badge. 
image

image

Twitch Admin and Staff as well would be cleaned up to be Staff and Admin respectively
image
image


pajlada:
This seems like what's being used in Twitch web chat, so us transitioning over to `title` makes sense to me.

PR #1424, opened 2019-11-23
Now also shows months subscribed in tooltip

### LINKED ISSUE #1418: Show x amount of subbed months when mouse is hovering over the badge
**Describe the bug**

when you hover over a badge it does not show sub age
**To reproduce**
hover over a badge
**Screenshots**
https://i.imgur.com/aLEsJPn.png

**Chatterino version**
Chatterino 2.1.5 (commit c6e1a1de2)

**Operating system**
Windows 10 also browser version

apa420:
Now shows months in tooltip when hovering over the subscriber badge.
Since ```@badge-info``` only gives ```subscriber/{months}``` according to https://dev.twitch.tv/docs/irc/tags#globaluserstate-twitch-tags we use first index on the tag.

Closes #1418

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1251,6 +1251,34 @@ void TwitchMessageBuilder::appendTwitchBadges()
                 break;
             }
         }
+        else if (badge.startsWith("founder/"))
+        {
+            if (auto badgeEmote =
+                    this->twitchChannel->globalTwitchBadges().badge("founder",
+                                                                    "0"))
+            {
+                auto badgeInfo = this->tags.find("badge-info");
+                if (badgeInfo != this->tags.end() &&
+                    badgeInfo.value().toString().split(',')[0].startsWith(
+                        "founder/"))
+                {
+                    auto subMonths =
+                        badgeInfo.value().toString().split(',')[0].mid(8);
+                    this->emplace<BadgeElement>(
+                            badgeEmote.get(),
+                            MessageElementFlag::BadgeSubscription)
+                        ->setTooltip(QString((*badgeEmote)->tooltip.string) +
+                                     " (" + subMonths + " months)");
+                }
+                else
+                {
+                    this->emplace<BadgeElement>(
+                            badgeEmote.get(),
+                            MessageElementFlag::BadgeSubscription)
+                        ->setTooltip((*badgeEmote)->tooltip.string);
+                }
+            }
+        }
         else if (badge.startsWith("subscriber/"))
         {
             if (auto badgeEmote = this->twitchChannel->twitchBadge(
@@ -1309,7 +1337,7 @@ void TwitchMessageBuilder::appendTwitchBadges()
             }
         }
     }
-}
+}  // namespace chatterino
 
 void TwitchMessageBuilder::appendChatterinoBadges()
 {
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1256,9 +1256,26 @@ void TwitchMessageBuilder::appendTwitchBadges()
             if (auto badgeEmote = this->twitchChannel->twitchBadge(
                     "subscriber", badge.mid(11)))
             {
-                this->emplace<BadgeElement>(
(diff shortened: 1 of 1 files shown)


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1309,7 +1337,7 @@ void TwitchMessageBuilder::appendTwitchBadges()
             }
         }
     }
-}
+}  // namespace chatterino
```
pajlada: remove this comment

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1251,14 +1251,59 @@ void TwitchMessageBuilder::appendTwitchBadges()
...
         }
+        else if (badge.startsWith("founder/"))
+        {
+            if (auto badgeEmote =
+                    this->twitchChannel->globalTwitchBadges().badge("founder",
+                                                                    "0"))
+            {
+                auto badgeInfo = this->tags.find("badge-info");
+                if (badgeInfo != this->tags.end() &&
+                    badgeInfo.value().toString().split(',')[0].startsWith(
```
23rd: The `badgeInfo.value().toString().split(',')[0]` is used twice, so it can be moved.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1251,14 +1251,59 @@ void TwitchMessageBuilder::appendTwitchBadges()
...
+                        "founder/"))
+                {
+                    auto subMonths =
+                        badgeInfo.value().toString().split(',')[0].mid(8);
```
23rd: I think it would be more flexible to move the `"founder/"` to variable and use a `founderStr.size()` instead of a raw number `8`.

The same thoughts about `"subscriber/"` and `mid(11)`.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1251,14 +1251,59 @@ void TwitchMessageBuilder::appendTwitchBadges()
...
+                    this->emplace<BadgeElement>(
+                            badgeEmote.get(),
+                            MessageElementFlag::BadgeSubscription)
+                        ->setTooltip(QString((*badgeEmote)->tooltip.string) +
+                                     " (" + subMonths + " months)");
```
23rd: It can be simplified with `QString().arg()`.
Something like this.
```
QString("%1 (%2 months)").arg((*badgeEmote)->tooltip.string).arg(subMonths));
```

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1251,14 +1251,59 @@ void TwitchMessageBuilder::appendTwitchBadges()
...
+                }
+                else
+                {
+                    this->emplace<BadgeElement>(
+                            badgeEmote.get(),
+                            MessageElementFlag::BadgeSubscription)
+                        ->setTooltip((*badgeEmote)->tooltip.string);
```
23rd: We emplace the BadgeElement in any case, so we can move this out of a condition and avoid repeating the code.
Something like this.
```
auto badgeInfo = this->tags.find("badge-info");
QString tooltipText;
if (badgeInfo != this->tags.end() &&
    badgeInfo.value().toString().split(',')[0].startsWith(
        "founder/"))
{
    auto subMonths =
        badgeInfo.value().toString().split(',')[0].mid(8);
    tooltipText = QString((*badgeEmote)->tooltip.string) +
        " (" + subMonths + " months)";
}
else
{
    tooltipText = (*badgeEmote)->tooltip.string;
}
this->emplace<BadgeElement>(
        badgeEmote.get(),
        MessageElementFlag::BadgeSubscription)
    ->setTooltip(tooltipText);
```

The same thoughts about a `subscriber` badge emplacement.

23rd:
Oh, this was merged already. =(
I'm so dumb.

pajlada:
feedback still valid as I will be going around cleaning up that function soon

pajlada:
@23rd Please check out my changes made in bcc53c9aa7966baa9051feda0aaee6458d8ab73c - I hope this is a bit more clean

23rd:
Thank you! Now it looks much better and shorter.

PR #1426, opened 2019-11-30
Fixed closing tab with middle button when tab management is not allowed.

23rd:
This PR fixes the possibility to close the tabs in the "Emotes" popup using the middle mouse button.


pajlada:
Looks good to me, thanks for noticing and thanks for fixing it!

PR #1439, opened 2019-12-14
Just removed the mod button

### LINKED ISSUE #1419: Option to hide "unmod" button
image
Personally I've almost clicked this button a bunch of times as well 🌵 

Option to disable the unmod button, could have the option itself disable the mod button as well

Probably put it under Miscellaneous "Disable (Mod)?/Unmod buttons"

Also if this is created the option would have to be disabled by default (to keep mod/unmod buttons enabled by default)

**Additional context**

image

**edit:** 
This entire issue should probably say show/hide instead of disable

Mm2PL:
4HEad
closes #1419


pajlada:
We would like to keep the mod button
The unmod button visibility just needs to change to be the exact same as the mod button visiblity (only visible if you're a broadcaster).

PR #1443, opened 2019-12-16
Fixing bits

### LINKED ISSUE #1402: Bits are summed wrong
**Describe the bug**
When user donates bits with a message that involves multiple bit emotes, the total bit amount is displayed on each emote

**To reproduce**
Write "trihard1 trihard1" while having 2 bits in account

**Screenshots**
7TaUjiK 1 (left is webchat)

**Chatterino version**
Chatterino Nightly 2.1.5 (commit d200f12f4, 27.09.2019) (randers minimod fork)

**Operating system**
Win 10 1903 18362.356

apa420:
Fixes #1402 
This pr fixes some issues with the current bits being displayed.

1. Someone types something similiar to what Tetyys types (he's cheered one bit here):
New:

Old:

2. Bits should now display different numbers instead of all being the same
New:

Old:

3. Bits now get calculated how many are left (one bit cheered here)
New:

Old:

4. If the amount that is left is less than the next cheer string it will make it accordingly smaller
E.g: 1500 bits cheered and with the message (total written 1599):
```Cheer1000 Cheer100 Cheer100 Cheer100 Cheer100 Cheer99 Cheer100```
New:

Old:

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -758,6 +758,7 @@ void TwitchChannel::refreshCheerEmotes()
 
                     cheerEmote.color = QColor(tier.color);
                     cheerEmote.minBits = tier.minBits;
+                    cheerEmote.regex = cheerEmoteSet.regex;
 
                     // TODO(pajlada): We currently hardcode dark here :|
                     // We will continue to do so for now since we haven't had to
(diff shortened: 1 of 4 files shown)


apa420:
I guess this could still need some work, like calculating how many more bits there are left so we don't show too many.

apa420:
Fixed!
Check main comment

PR #1445, opened 2019-12-17
Add AppData manifest

tim77:
Add initial desktop file and AppData manifest file for modern package managers like GNOME Software/Discover KDE. It will be easier to discover and install Chatterino for users in future. Also there is cool donating button. :)

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

If you think name scheme is weird please check this
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

This appdata and UUID name is also mandatory if you plan distibute your app in future as Flatpak package.


resources/com.chatterino.chatterino2.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
```
leon-richardt: Is this compatible with the MIT license ([used by Chatterino 2](https://github.com/Chatterino/chatterino2/blob/master/LICENSE))?

tim77: Yep, this appdata under `CC0-1.0` license which have no restrictions at all. Just a credit to author. I followed this guidelines
https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/#_appdata_xml_file_creation

resources/com.chatterino.chatterino2.appdata.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
```
pajlada: Curious about best practices in terms of copyrighting contributions like this - generally any contribution would fall under the project license (MIT).
I'd prefer we just skip the copyright notice for now. I understand this makes our file not validate the strictest check, but that doesn't seem to be a real concern.

chatterino.pro
```diff
@@ -509,7 +509,7 @@ linux:isEmpty(PREFIX) {
 }
 
 linux {
-    desktop.files = resources/chatterino.desktop
+    desktop.files = resources/com.chatterino.chatterino2.desktop
```
pajlada: The "application id" we use for other OSes has been `com.chatterino.chatterino` - I think we should stay consistent. Please update this in the other changes you've made too.

resources/com.chatterino.chatterino2.appdata.xml
tim77: No prob. I'll fix this shortly.

PR #1446, opened 2019-12-18
Stack bits

### LINKED ISSUE #1313: Bit stacking

No

Bit stacking: takes the first emote/icon of the bit message and stacks all bits into one group to reduce the 1 bit spam people do

Using web chat 4Head

**Additional context**

apa420:
Requires #1443 
Closes #1313 
We have ran out of enums for message flags (32 bit) so it won't dynamically change.

### DIFF
--- a/docs/ENV.md
+++ b/docs/ENV.md
@@ -18,3 +18,17 @@ Used to change the URL that Chatterino2 uses when trying to get emote set inform
 Default value: `https://braize.pajlada.com/chatterino/twitchemotes/set/%1/`  
 Arguments:  
  - `%1` = Emote set ID
+
+### CHATTERINO2_TWITCH_SERVER_HOST
+String value used to change what Twitch chat server host to connect to.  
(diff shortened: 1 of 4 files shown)


alazymeme:
This PR shouldn't close https://github.com/Chatterino/chatterino2/issues/1313 as it is a feature request. The only thing mentioned was a comment regarding the bug that I've now removed.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1286,6 +1286,35 @@ Outcome TwitchMessageBuilder::tryParseCheermote(const QString &string)
...
     int cheerValue = match.captured(1).toInt();
 
+    if (getSettings()->stackBits)
+    {
+        if (this->bitsStacked)
+        {
+            return Success;
+        }
+        else
+        {
```
pajlada: Please remove this else-scope

Issue #1444, opened 2019-12-17
Build with system fmt


tim77:
Hello. Is it possible to build chatterino2 with system `fmt`, not bundled? This could help with pushing chatterino2 into official repos of some Linux distros since some of them have strict requirements as for bundled libs. Maybe also some other libs could be unbundled as well?

pajlada:
We can definitely look into this. Please list the libraries that are most important for you to be system-bound other than fmt.

tim77:
TBH most important is cryprographic libs and OpenSSL already unbundled, which is nice. But seems like you bundled not for no reason other libs since they are patched... Ok, would be nice if we unbundle this:

```
fmt
libcommuni
rapidjson
websocketpp
```

Note: i consulted with other maintainers and seems like this is not mandatory but recommended.

pajlada:
@tim77 Please see my approach #1447 - I've currently only added support to compile with system version of fmt.
In short, you can pass two extra flags for fmt right now:
`FMT_PREFIX=/path/to/fmt` to compile Chatterino with your own source and header folder
`FMT_SYSTEM=1` to link Chatterino with system fmt
Do note that fmt needs to be version 4 for it to work with Chatterino (which is why it's a submodule/included in the project in the first place)

pajlada:
I've merged #1447 into master now, so I'll close this and make a new issue for updating the fmt lib version

PR #1447, opened 2019-12-19
Better support for compiling with system libraries

pajlada:
#1444

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -1,3 +1,10 @@
+# Exposed build flags:
+# from lib/fmt.pri
+#  - FMT_PREFIX ($$PWD by default)
+#  - FMT_SYSTEM (1 = true) (unix only)
+# from lib/boost.pri
+#  - BOOST_DIRECTORY (C:\local\boost\ by default) (Windows only)
+
 QT                += widgets core gui network multimedia svg concurrent
 CONFIG            += communi
 COMMUNI           += core model util
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -2,6 +2,12 @@
 # from lib/fmt.pri
 #  - FMT_PREFIX ($$PWD by default)
 #  - FMT_SYSTEM (1 = true) (unix only)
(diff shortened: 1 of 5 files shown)


pajlada:
@tim77 
I have added build options for:
```
fmt
rapidjson
websocketpp
```

however, for `libcommuni` we use our own fork for some important irc-related fixes.
On my machine, I was only able to use the rapidjson system library, both fmt and websocketpp had issuues with their system variant.
For fmt, it's just a matter of the version being wrong.
For websocketpp, the version was correct but there had been a patch applied by the arch linux maintainer which made it not work for us.

Please let me know if this fixes your issue, and if not, what I can do to fix it.

tim77:
Good news: chatterino2 packaged for official Fedora repos and now [on QA](https://bodhi.fedoraproject.org/updates/FEDORA-2019-8139b22c6b), but many maintainers still insist that we should use system libs and asking why not push to upstream this patches?

I'll try to build with system `rapidjson` now... At least this something. Thank you.

tim77:
> For fmt, it's just a matter of the version being wrong.

Would it possible to port on actual version 6?

xvitaly:
Hardocoding of paths is not good.

You should use pkg-config. qmake has its support. Just use something like this:

```
CONFIG += link_pkgconfig
PKGCONFIG += fmt
```

xvitaly:
`CONFIG += link_pkgconfig` should be added to main config and `PKGCONFIG += fmt` - to `fmt.pri`.

Pkg-config will automatically find library, its headers and specify correct compiler/linker options.

pajlada:
> > For fmt, it's just a matter of the version being wrong.
> 
> Would it possible to port on actual version 6?

Definitely possible yeah



> `CONFIG += link_pkgconfig` should be added to main config and `PKGCONFIG += fmt` - to `fmt.pri`.
> 
> Pkg-config will automatically find library, its headers and specify correct compiler/linker options.

Thanks! looking into this

PR #1476, opened 2020-01-02
Expect a PONG response to our PINGs.

### LINKED ISSUE #1164: Connecting or Disconnecting a VPN requires a manual reconnect
Maybe chatterino can detect this and reconnect automatically

pajlada:
If no PONG was received, force a reconnection.

Fixes #1164

How does this work?
Keep track of sent PINGs and PONGs.
Expect each PING to have a matching PONG.
If a PING does not receive a PONG by the time it's time for us to send another PING message (after 5+ seconds), we stop refreshing the reconnection timer and let it run out on its own (usually after another ~5 seconds).

This way of doing it is a little bit iffy because we have two connections. In the scenario where the entire connection actually dies, both the read and the write connection will die, probably around the same time, and both fire reconnect requests. this will disconnect and reconnect both the read and write connections. But this is not a worse solution than what we already have, it's just not perfect.

### DIFF
--- a/src/providers/irc/IrcConnection2.cpp
+++ b/src/providers/irc/IrcConnection2.cpp
@@ -1,6 +1,15 @@
 #include "IrcConnection2.hpp"
 
+#include "common/Version.hpp"
+
 namespace chatterino {
+
+namespace {
+
+    const auto payload =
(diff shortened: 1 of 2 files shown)


fourtf:
lgtm

PR #1496, opened 2020-01-11
Makes it possible to hide one man spam

hemirt:
Hides (with several options) messages that are too similar and too close either in time, or based on previous messages by that same user

[old settings visual](https://user-images.githubusercontent.com/1310440/72208133-a58ac180-349f-11ea-92b5-7fe1d8359b93.png)

EDIT: new settings visuals
image

### DIFF
--- a/src/common/Channel.cpp
+++ b/src/common/Channel.cpp
@@ -203,20 +203,6 @@ void Channel::addMessagesAtStart(std::vector<MessagePtr> &_messages)
 {
     std::vector<MessagePtr> addedMessages =
         this->messages_.pushFront(_messages);
-    /*
-    if (getSettings()->similarityEnabled)
-    {
-        for (int i = 0; i < addedMessages.size(); ++i)
-        {
-            if (IrcMessageHandler::similarityRecentMessages(addedMessages, i) >
-                getSettings()->similarityPercent.getValue())
-            {
-                addedMessages[i]->flags.set(MessageFlag::Hidden, true);
-                addedMessages[i]->flags.set(MessageFlag::Disabled, true);
-            }
-        }
-    }
-    */
 
     if (addedMessages.size() != 0)
     {
--- a/src/common/Channel.cpp
+++ b/src/common/Channel.cpp
@@ -3,6 +3,7 @@
 #include "Application.hpp"
 #include "messages/Message.hpp"
 #include "messages/MessageBuilder.hpp"
+#include "providers/twitch/IrcMessageHandler.hpp"
 #include "singletons/Emotes.hpp"
 #include "singletons/Logging.hpp"
 #include "singletons/Settings.hpp"
@@ -202,6 +203,20 @@ void Channel::addMessagesAtStart(std::vector<MessagePtr> &_messages)
 {
     std::vector<MessagePtr> addedMessages =
         this->messages_.pushFront(_messages);
+    /*
+    if (getSettings()->similarityEnabled)
+    {
+        for (int i = 0; i < addedMessages.size(); ++i)
+        {
+            if (IrcMessageHandler::similarityRecentMessages(addedMessages, i) >
+                getSettings()->similarityPercent.getValue())
+            {
+                addedMessages[i]->flags.set(MessageFlag::Hidden, true);
+                addedMessages[i]->flags.set(MessageFlag::Disabled, true);
+            }
+        }
+    }
+    */
 
     if (addedMessages.size() != 0)
     {
--- a/src/messages/Message.hpp
+++ b/src/messages/Message.hpp
@@ -33,6 +33,7 @@ enum class MessageFlag : uint32_t {
     Whisper = (1 << 16),
     HighlightedWhisper = (1 << 17),
     Debug = (1 << 18),
+    Hidden = (1 << 19),
 };
 using MessageFlags = FlagsEnum<MessageFlag>;
 
--- a/src/messages/Message.hpp
+++ b/src/messages/Message.hpp
@@ -33,7 +33,7 @@ enum class MessageFlag : uint32_t {
     Whisper = (1 << 16),
     HighlightedWhisper = (1 << 17),
     Debug = (1 << 18),
-    Hidden = (1 << 19),
+    Similar = (1 << 19),
 };
 using MessageFlags = FlagsEnum<MessageFlag>;
 
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -60,60 +60,31 @@ static float relativeSimilarity(const QString &str1, const QString &str2)
 float IrcMessageHandler::similarity(
     MessagePtr msg, const LimitedQueueSnapshot<MessagePtr> &messages)
 {
-    float ret = 0.0f;
-    int bysameuser = 0;
-    for (int i = 1; bysameuser < getSettings()->hideSimilarMaxMessagesToCheck;
+    float similarityPercent = 0.0f;
+    int bySameUser = 0;
+    for (int i = 1; bySameUser < getSettings()->hideSimilarMaxMessagesToCheck;
          ++i)
     {
         if (messages.size() < i)
         {
             break;
         }
-        const auto &prev_msg = messages[messages.size() - i];
-        if (prev_msg->parseTime.secsTo(QTime::currentTime()) >=
+        const auto &prevMsg = messages[messages.size() - i];
+        if (prevMsg->parseTime.secsTo(QTime::currentTime()) >=
             getSettings()->hideSimilarMaxDelay)
         {
             break;
         }
-        if (msg->loginName != prev_msg->loginName)
+        if (msg->loginName != prevMsg->loginName)
         {
             continue;
         }
-        ++bysameuser;
-        ret = std::max(
-            ret, relativeSimilarity(msg->messageText, prev_msg->messageText));
+        ++bySameUser;
+        similarityPercent = std::max(
+            similarityPercent,
+            relativeSimilarity(msg->messageText, prevMsg->messageText));
     }
-    return ret;
-}
-
-float IrcMessageHandler::similarityRecentMessages(
-    const std::vector<MessagePtr> &messages, int pos_size)
-{
-    float ret = 0.0f;
-    int bysameuser = 0;
-    for (int i = 1; bysameuser < getSettings()->hideSimilarMaxMessagesToCheck;
-         ++i)
-    {
-        if (pos_size < i)
-        {
-            break;
-        }
-        const auto &prev_msg = messages[pos_size - i];
-        // is not able to time it, since parseTime equals currentTime during recent messages
-        if (prev_msg->parseTime.secsTo(QTime::currentTime()) >=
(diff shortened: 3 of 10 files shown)


src/common/Channel.cpp
```diff
@@ -202,6 +203,20 @@ void Channel::addMessagesAtStart(std::vector<MessagePtr> &_messages)
...
+        {
+            if (IrcMessageHandler::similarityRecentMessages(addedMessages, i) >
+                getSettings()->similarityPercent.getValue())
+            {
+                addedMessages[i]->flags.set(MessageFlag::Hidden, true);
+                addedMessages[i]->flags.set(MessageFlag::Disabled, true);
+            }
+        }
+    }
+    */
```
apa420: Why is this uncommented

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -671,12 +671,14 @@ void TwitchMessageBuilder::appendUsername()
 
     switch (usernameDisplayMode.getValue())
     {
-        case UsernameDisplayMode::Username: {
+        case UsernameDisplayMode::Username:
```
apa420: Update your clang

src/widgets/settingspages/KeyboardSettingsPage.cpp
```diff
@@ -26,6 +26,8 @@ KeyboardSettingsPage::KeyboardSettingsPage()
     form->addRow(new QLabel("Ctrl + Shift + T"), new QLabel("Create new tab"));
     form->addRow(new QLabel("Ctrl + Shift + W"),
                  new QLabel("Close current tab"));
+    form->addRow(new QLabel("Ctrl + H"),
```
apa420: Don't like that this is a shortcut, maybe add it when custom shortcuts are a thing

src/common/Channel.cpp
hemirt: https://github.com/Chatterino/chatterino2/pull/1496/files/e14be92deafb4aa3b59763361c0876ea7b4b6248#diff-be41aa477922fd78d2e9dac3f5ac1711R105

src/providers/twitch/TwitchMessageBuilder.cpp
hemirt: everytime i do sth, the formatting is scuffed, what are you doing with it all the time

src/widgets/settingspages/KeyboardSettingsPage.cpp
hemirt: it is a shortcut because you might want to see someone spamming - e.g. when you want to moderate spam in one channel, but not want to see spam when youre watching someone else

hemirt: from my use it is extremely useful

src/providers/twitch/TwitchMessageBuilder.cpp
apa420: Formatting doesn't change that often, downloading clang 9.0+ should fix it

src/common/Channel.cpp
hemirt: so what should be done about this

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
 
 namespace chatterino {
 
+static float LongestCommonSubstring(QString str1, QString str2)
```
leon-richardt: This should probably be put into the unnamed namespace and the parameters can be `const &`.

Also, I find the naming a tad bit confusing since this not actually returning the longest common substring. Maybe something like `relativeSimilarity` would be better?

hemirt: yeah i wasnt sure where to put this, and also where to put the actual `IrcMessageHandler::similarity` functions
the parameters should be const & (tho iirc qstrings are COW, but yeah it should be const &)

Yeah this problem is called `Longest common substring problem` so i named it as such, maybe something like `relativeSimilarity` would be better

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
...
+        if (messages[pos_size]->loginName != prev_msg->loginName)
+        {
+            continue;
+        }
+        ++bysameuser;
+        ret = std::max(ret, relativeSimilarity(messages[pos_size]->messageText,
+                                               prev_msg->messageText));
+    }
+    return ret;
+}
```
apa420: To be honest I don't really see a point in trying to check the recent messages. Even if they are spam I don't feel like it's necessary to filter them.

src/widgets/Window.cpp
```diff
@@ -351,6 +351,11 @@ void Window::addShortcuts()
             getApp()->twitch.server->getOrAddChannel(si.channelName));
         splitContainer->appendSplit(split);
     });
+
+    createWindowShortcut(this, "CTRL+H", [this] {
+        getSettings()->hideSimilar.setValue(!getSettings()->hideSimilar);
+        getApp()->windows->forceLayoutChannelViews();
+    });
```
apa420: Like I've commented before, I dislike a button to enable or disable this. This could be added when there is proper custom support for keyboard shortcuts. 
Right now the issue is if someone accidently presses this and wonders why the chat doesn't show all messages or is really slow since there is no indication at all that any message is being removed.

src/widgets/settingspages/KeyboardSettingsPage.cpp
apa420: Remove

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -547,6 +547,34 @@ void GeneralPage::initLayout(SettingsLayout &layout)
         QDesktopServices::openUrl(getPaths()->rootAppDataDirectory);
     });
 
+    layout.addTitle("Similarity");
+    layout.addDescription(
+        "Anti one man spam feature based on the similarity to "
+        "previous messages from the same person.");
```
apa420: Better description.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
...
+    }
+
+    return z == 0 ? 0.f : float(z) / std::max(str1.size(), str2.size());
+};
+
+float IrcMessageHandler::similarity(
+    MessagePtr msg, const LimitedQueueSnapshot<MessagePtr> &messages)
+{
+    float ret = 0.0f;
+    int bysameuser = 0;
```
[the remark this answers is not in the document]
hemirt: changed

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
...
+    for (int i = 1; bysameuser < getSettings()->hideSimilarMaxMessagesToCheck;
+         ++i)
+    {
+        if (messages.size() < i)
+        {
+            break;
+        }
+        const auto &prev_msg = messages[messages.size() - i];
```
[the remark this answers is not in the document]
hemirt: chanegd

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
...
+        }
+        ++bysameuser;
+        ret = std::max(
+            ret, relativeSimilarity(msg->messageText, prev_msg->messageText));
+    }
+    return ret;
+}
+
+float IrcMessageHandler::similarityRecentMessages(
+    const std::vector<MessagePtr> &messages, int pos_size)
```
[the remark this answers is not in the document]
hemirt: removed

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -21,6 +21,101 @@
...
+{
+    float ret = 0.0f;
+    int bysameuser = 0;
```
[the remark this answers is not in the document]
hemirt: removed

src/providers/twitch/IrcMessageHandler.cpp
hemirt: removed, will not check them since there's no immediate way to find their time

src/widgets/Window.cpp
hemirt: You can't accidentally press this button, since this button in default state does nothing, you will have to enable a setting to hide similar messages which does in the name tell you about this ctrl+h shortcut

i find this shortcut extremely useful

src/widgets/settingspages/KeyboardSettingsPage.cpp
hemirt: as stated on comment above in window.cpp

src/widgets/settingspages/GeneralPage.cpp
hemirt: tried a better one

src/messages/Message.hpp
```diff
@@ -33,6 +33,7 @@ enum class MessageFlag : uint32_t {
     Whisper = (1 << 16),
     HighlightedWhisper = (1 << 17),
     Debug = (1 << 18),
+    Hidden = (1 << 19),
```
pajlada: This flag should be called something with "Similar" or something instead, since it doesn't really mean hidden, it just means that this message was marked as too similar to previous messages according to your settings.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -133,7 +199,33 @@ void IrcMessageHandler::addMessage(Communi::IrcMessage *_message,
...
-        builder.triggerHighlights();
+
+        if (getSettings()->similarityEnabled)
+        {
+            bool isMyself =
+                msg->loginName ==
+                getApp()->accounts->twitch.getCurrent()->getUserName();
+            bool hideMyself = getSettings()->hideSimilarMyself;
+
+            if ((!isMyself || hideMyself) &&
```
pajlada: This type of if-case gets confusing quickly - if we move this scope (everything within `if similarityEnabled`) into a separate function we can control the logic with early outs instead, making the flow easier to understand

src/widgets/settingspages/KeyboardSettingsPage.cpp
```diff
@@ -26,6 +26,8 @@ KeyboardSettingsPage::KeyboardSettingsPage()
...
+                 new QLabel("Hide/Show similar messages"));
```
pajlada: Maybe a small note here that this requires the similarity feature to be enabled

hemirt: added directions

src/providers/twitch/IrcMessageHandler.cpp
hemirt: moved to function

src/messages/Message.hpp
hemirt: renamed

pajlada:
Looks good to me, tested it and it works. The settings might need some touching up, but I can't think of a better way to phrase things while keeping the functionality so I'll merge this as is. Thank you!

PR #1502, opened 2020-01-15
Synchronize Clipboard with Primary Selection on Linux when copying

rexim:
Hello!

I'm a pretty heavy user of Middle Mouse click and Shift+Insert on Linux (those shortcuts pull data from the primary selection rather than the clipboard, the difference between them is summarize really well here: https://unix.stackexchange.com/a/139193). And the fact that Chatterino is the only app on my setup that does not sync with the selection makes it really inconvenient for me. So I went ahead and tried to fix it. Maybe the fix will be useful for other Linux users.

Please, let me know what you think.

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -210,6 +210,7 @@ SOURCES += \
     src/singletons/TooltipPreviewImage.cpp \
     src/singletons/Updates.cpp \
     src/singletons/WindowManager.cpp \
+    src/util/Clipboard.cpp \
     src/util/DebugCount.cpp \
     src/util/FormatTime.cpp \
     src/util/FunctionEventFilter.cpp \
@@ -412,6 +413,7 @@ HEADERS += \
     src/singletons/Updates.hpp \
     src/singletons/WindowManager.hpp \
     src/util/Clamp.hpp \
+    src/util/Clipboard.hpp \
     src/util/CombinePath.hpp \
     src/util/ConcurrentMap.hpp \
     src/util/DebugCount.hpp \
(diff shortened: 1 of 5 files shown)


fourtf:
Feature seems good to me just why do we pass the same QClipboard to the function every time?

rexim:
@fourtf in the original code sometimes we passed `QApplication::clipboard()`, sometimes we passed `QGuiApplication::clipboard()`, even tho they return the same object on my setup (I checked in gdb) I thought that it might be important. If it is not, please let me know and I will remove the clipboard argument from the function.

fourtf:
"For any GUI application using Qt, there is [...] one QApplication object [...]. For non-QWidget based Qt applications, use QGuiApplication instead, as it does not depend on the QtWidgets library." https://doc.qt.io/qt-5/qapplication.html#details

So QApplication inherts QGuiApplication so you can use Qt without QtWidgets (there is also QCoreApplication if you only use QtCore). So the QApplication instance and the QGuiApplication should be the same.

apa420:
From what I understand this will copy the text to both clipboards and not one? Could be annoying to some people, but maybe I'm wrong about that.

rexim:
@apa420 this is how many Linux applications actually work ¯\_(ツ)_/¯ Just checked with Chromium. I Ctrl+C'ed a text and it was automatically available in both Clipboard and Primary Selection.

pajlada:
This looks fine to me. The concept of having multiple clipboards, imo, is outdated. Most applications, as you mentioned, already fill both clipboards. I run a separate service on my computer to synchronize the clipboards but it would be nice to not have to do that.

ldesgoui:
I'd be interested to also have selection clipboard fill up when text is selected

fourtf:
So this is only supposed to fill the other clipboard as well and not actually copy the text while selecting it?

rexim:
@fourtf yes

Issue #1493, opened 2020-01-09
Won't connect to chat


Kakernak:
Will not connect to a chat, despite message in the chat stating 'reconnected ' when I hover over chat tab I see 'offline' despite streams not being offline, no messages are seen and no users can be seen. 

How to replicate: Launching the application and it not working.

![chatterino uh](https://user-images.githubusercontent.com/59707991/72100566-25d7e800-3323-11ea-8226-091100b9b982.png)

Chatterino version: 2.1.7

Operating system: windows 10

Additional info: Clicking reconnect does nothing. I've read through everything over here and on reddit trying different things, nothing worked (relogging, adding exceptions to firewall etc.)

pajlada:
This feels like an environment-based problem. Do you have any sort of firewall installed on your machine that might be interfering?

fourtf:
May be resolved by https://github.com/Chatterino/chatterino2/pull/1513

Could you try the newest nightly and see if this fixes the issue?

apa420:
Is this still an issue @Kakernak

fourtf:
inactive

PR #1513, opened 2020-01-24
Connect to Twitch IRC server using port 443 instead of 6697

marcmos:
I experience connection issues similar to these described in #1493 (no messages and users displayed) when running Chatterino in a firewalled environment (outgoing TCP connections to port 6697 blocked). It seems that a breaking change was introduced in PR #1449 (builds before this change worked well on the same machine/environment).

It seems that default Twitch IRC connection port was [accidentally(?) set from 443 to 6697](https://github.com/Chatterino/chatterino2/pull/1449/files#diff-cf97ac2da2381327578c41c94df81a77R62), as opposed to [defaults described in docs](https://github.com/Chatterino/chatterino2/pull/1449/files#diff-17a44c0332fd031df859f133db848ac1R30). Proposed change actually sets the default of `CHATTERINO2_TWITCH_SERVER_PORT` environment value to `443` and hopefully resolves issue #1493.

### DIFF
--- a/src/common/Env.cpp
+++ b/src/common/Env.cpp
@@ -59,7 +59,7 @@ Env::Env()
           "https://braize.pajlada.com/chatterino/twitchemotes/set/%1/"))
     , twitchServerHost(
           readStringEnv("CHATTERINO2_TWITCH_SERVER_HOST", "irc.chat.twitch.tv"))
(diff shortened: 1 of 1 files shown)


fourtf:
Seems to be accidental, thanks for debugging this and finding the issue!

Issue #1523, opened 2020-01-27
Whispers showing in mentions when not pinged


alazymeme:
**Describe the bug**
When `Whispers` is enabled under highlights, all whispers show in your mentions box (regardless of if you were mentioned or not)

**To reproduce**
1. Settings > Highlights > Turn on Whispers
2. Make someone whisper you and it should show in your `/whispers` and `/mentions` box

**Screenshots**
https://i.nuuls.com/hWJ-q.png
https://i.nuuls.com/_Nb-F.png

**Chatterino version**
Chatterino Nightly 2.1.7 (commit 497ce2d2f, )

**Operating system**
Windows 10

leon-richardt:
Thanks for the report, this should be fixed by #1524.

PR #1524, opened 2020-01-28
Better Highlights: Rework Highlight Parsing Order

leon-richardt:
This PR is in response to #1523.

Whispers are now only added to the `/mentions` tab if they also match a user name or phrase highlight. On a related note, the `highlightVisual_` member has been removed as it is no longer necessary.


leon-richardt:
One possible "pain point" I noticed: The `Highlighted` message flag now has a semantic meaning of "matches-a-user-or-phrase-highlight".

PR #1544, opened 2020-02-11
Control + End scrolls to the bottom of chat

### LINKED ISSUE #1533: Enable CTRL+END to scroll to bottom / resume scrolling (standard Windows shortcut)
In Chatterino you can click "More messages below" to resume scrolling. It would be nice to if you could use the standard shortcut `ctrl`+`end` to do this too. It doesn't look like there's a keyboard shortcut for this already, but would be good to have `ctrl`+`end` too even if there is.

`ctrl`+`end` works across all normal text fields in Windows, such as in this text box I'm writing in this in now, and in other chat apps (e.g. Discord). It's also used in KDE, Gnome and Emacs. On MacOS the equivalent is `Cmd`+`↓`.

The counterpart is `ctrl`+`home` to scroll to the top, which would make sense to implement too, but is less important.

* Reference: https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts
(listed under "Go to end of document")

BigHeadGeorge:
Closes #1533, but there are a few other shortcut suggestions in there that would nice along with this one.

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -9,6 +9,7 @@
 #include "singletons/Theme.hpp"
 #include "util/LayoutCreator.hpp"
 #include "widgets/Notebook.hpp"
+#include "widgets/Scrollbar.hpp"
 #include "widgets/dialogs/EmotePopup.hpp"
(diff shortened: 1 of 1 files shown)


fourtf:
Looks good to me 👍

Issue #869, opened 2018-11-08
Support Emoji 11.0


LeaPhant:
Emoji 11.0 has been released in June 2018 and is supported by Apple, Google, Twitter and EmojiOne.
Would be nice to have the all the new emoji that have been added! 🐘

pajlada:
One issue with supporting the new emoji data is that the emoji data provider we rely on has been forced to remove support for EmojiOne

https://github.com/iamcal/emoji-data

so it's unfortunately not as easy as just running a script

This is definitely something we would want to be able to keep up with though, so maybe it's just a matter of making a smarter script

TranRed:
They also haven an issue over there, maybe we can have an eye on that.

https://github.com/iamcal/emoji-data/issues/145

savabozo:
https://github.com/iamcal/emoji-data/issues/145#issuecomment-493588183
avaliable on master

alazymeme:
![jif](https://cdn.betterttv.net/emote/5fedbfca9d7d952e4059e40c/1x) @zneix

pajlada:
Fixed in 6b0ce396d6f8d727aa641273d00e1013b86c2f3a (#1555 )

PR #1555, opened 2020-02-15
Update to Emoji v13 (2020)

pajlada:
The plan is to:
 - [ ] Update the emoji stuff to support Unicode 12.1 emojis and their new tone stuff
 - [ ] Upload emojis to a proper CDN instead of using a random server


RAnders00:
https://www.jsdelivr.com/package/npm/emoji-datasource-twitter?path=img%2Ftwitter%2F64 twitter emojis are on jsdelivr too as far as i can see

RAnders00:
An issue I've found: Some modifiers are not rendered as images when there are images for them, for example 🦲 (EMOJI COMPONENT BALD) renders as https://abs-0.twimg.com/emoji/v2/svg/1f9b2.svg when used on twitter.com (used as a reference for their Twitter set of emoji images) however in chatterino it does not render: 
image

Emojipedia also lists the twitter image for the "bald" modifier alone: https://emojipedia.org/bald/
image

zneix:
Fixing this would close #869

alazymeme:
Hello any news on this pr? I'm patiently waiting to use the pinching hand emoji in chatterino 🤏

fourtf:
Is this still being worked on or can I close?

pajlada:
Leave this up

alazymeme:
Looking forward to using the new emojis in chatterino 😁

alazymeme:
New year, new emojis 😄 can't wait for this to be implemented!

alazymeme:
Can't wait to prove these nerds wrong when I can see the window emoji 😲

Mm2PL:
Can't wait for this to be merged

src/providers/emoji/Emojis.cpp
```diff
@@ -409,11 +381,15 @@ QString Emojis::replaceShortCodes(const QString &text)
...
         {
+            printf("No match for %s\n", matchString.toStdString().c_str());
             continue;
         }
 
         auto emojiData = emojiIt.value();
 
+        printf("Found match for %s: '%s'\n", qUtf8Printable(matchString),
+               qUtf8Printable(emojiData->unifiedCode));
+
```
zneix, line 392: We should not use `printf` here

zneix (changes requested):
Looks good to me, new emojis work fine. Worth noting there's a very slow emoji load at the first launch, but since it seems to be cached and loads quicker later I guess it's fine :+1:

pajlada:
> Looks good to me, new emojis work fine. Worth noting there's a very slow emoji load at the first launch, but since it seems to be cached and loads quicker later I guess it's fine +1

Thanks for the review!
The slow speed is at least just as slow as it's been before, not slower, as it's hosted on the same provider & server and the filesizes are very similar for these set of emojis.

Still interested in finding a "true CDN" that isn't slow af, but the b2+cloudflare combo was megaslow

alazymeme:
We stopped using JoyPixels didn't we? Was looking at a basic CDN like this: https://cdnjs.com/libraries/emoji-toolkit

pajlada:
> We stopped using JoyPixels didn't we? Was looking at a basic CDN like this: https://cdnjs.com/libraries/emoji-toolkit

We did, yes

pajlada:
Waiting for #2453 to be merged in before running final tests on this and merging it

Issue #1542, opened 2020-02-10
Allow reordering elements in list settings


fourtf:
**Is your feature request related to a problem? Please describe.**
Elements in lists in the settings cannot be reordered. However the order is relevant (highlight processing order/moderation button order).

**Describe the solution you'd like**
Add buttons that move the item one step up or down.

**Screenshot**
![20200210_17h35m53s_grim](https://user-images.githubusercontent.com/6320036/74169951-6e3d3b00-4c2c-11ea-8e2a-f61457765f0b.png)

leon-richardt:
Being able to reorder elements via drag-and-drop would also be nice, if possible.

PR #1557, opened 2020-02-16
Custom highlight color in tabs

### LINKED ISSUE #1543: Use custom highlight colors in tabs
Currently when a highlight goes off in a chat that you have been pinged in, the taskbar of that chat flashes red. Why not have it so the taskbar of that certain chat tab flashed the color of the highlight which pinged you.  

**Additional context**
If my explanation didn't make enough sense, here is a screenshot of what I am saying. https://imgur.com/a/KKHfq2X This tab would instead flash the color of your chosen highlight

2qar:
Closes #1543 
Tabs will highlight whatever color is set for that highlight instead of highlighting red.

Example:
Highlight


fourtf:
what happens if there are multiple highlights?

BigHeadGeorge:
The color of the tab will be whichever highlight happened last. Should there be a priority (eg. username > highlight phrases)?

fourtf:
I'm not sure, perhaps the order of the highlights? Maybe it would also be possible to highlight the tab using multiple colors. None of this is really thought out, just some ideas.

leon-richardt:
> I'm not sure, perhaps the order of the highlights?

That would be nice in combination with #1542. 😄

BigHeadGeorge:
> Maybe it would also be possible to highlight the tab using multiple colors.

Maybe split up like this?
Highlight split

fourtf:
That's what I was thinking off but I'm not sure if it's the right way to go

BigHeadGeorge:
Sorry for the slow response, here's my ideas so far:
* Leaving it and adding #1542 would keep it simple, but won't represent all highlights in that tab
* Splitting the colored part would represent more highlights, but looks kinda ugly and would get cluttered fast
* Flash between the different highlight colors? (probably distracting)

Maybe there could be a setting for this, maybe I'm just overcomplicating it, idk. I think keeping it one color is the right way to go from a visual standpoint.

src/widgets/splits/SplitContainer.cpp
```diff
@@ -200,17 +200,15 @@ void SplitContainer::addSplit(Split *split)
...
                 this->tab_->setHighlightState(state);
             }
-        });
-
-    split->getChannelView().tabHighlightColorRequested.connect(
-        [this](std::shared_ptr<QColor> color) {
             if (color != nullptr)
+            {
                 this->tab_->setHighlightColor(color);
+            }
```
pajlada: this should jump into the above `if (this->tab_ != nullptr)` scope

fourtf:
I'll consider if this should be added.

alazymeme:
> I'll consider if this should be added.

Have you considered yet?

alazymeme:
...

jammehcow:
@ALazyMeme he's still considering :)

xHeaveny:
time to revive this
image

jammehcow:
@fourtf this is beyond stale now, have you made a decision?

MrAuro:
Is there anyone against this change? This PR has been open for _years_ now. 

Since GitHub issues doesn't have the polls feature that the discussions have, react with 👍 if you are for this change and 👎 if you are against it.

Mm2PL:
@fourtf, @pajlada, @zneix, it's about time to decide if this PR is good or one that we should close.

Felanbird:
We missed 2 years 🎆 🎈 - I'll try to remember next year that Sept 26th is the anniversary

Infinitay:
Do I dare suggest for an option for channels you are subscribed in to be highlighted?

fourtf:
I wasn't aware that this was still open. Looks good to me 👍

pajlada:
Closing in favour of https://github.com/Chatterino/chatterino2/pull/5579

PR #1559, opened 2020-02-20
Scroll Using Click-and-Drag with Middle Mouse Button

### LINKED ISSUE #1505: Beginn scrolling by clicking middle mouse button

A lot of software (emg. Browsers) start scrolling when clicking the middle mouse button. Chatterino would be more in line when it also supported that.

**Additional context**

Behaviour should be similar to the implementation in browsers.

leon-richardt:
### Description
Allows a behavior similar to web browser behavior: After clicking the middle mouse button, a user can move the mouse up or down to scroll the chat in the desired direction.
Resolves #1505.

### Additional Questions & Considerations
- [x] Make scrolling behave consistently with different screen scales.
- [x] Figure out good trade-off between `scrollTimer_` interval and UX.
- [x] Use a custom cursor pixmap for showing different scroll states, like used in browsers.


src/widgets/helper/ChannelView.cpp
```diff
@@ -130,6 +130,27 @@ ChannelView::ChannelView(BaseWidget *parent)
     this->clickTimer_->setSingleShot(true);
     this->clickTimer_->setInterval(500);
 
+    // TODO(leon): What interval to choose?
+    this->scrollTimer_.setInterval(20);
+    QObject::connect(&this->scrollTimer_, &QTimer::timeout, this, [this] {
+        const qreal delta = this->currentMousePosition_.y() -
+                            this->lastMiddlePressPosition_.y();
+
+        // TODO(leon): How large should the "grace area" be?
```
fourtf: doesn't account for scale, so if you're using 4k then it's only 5 px

fourtf:
It doesn't visualize the starting point like in chrome which not expected and bad for learnability (user doesn't know what's going on)

leon-richardt:
> Use a custom cursor pixmap for showing different scroll states, like used in browsers.

That's what this TODO was supposed to describe, guess I didn't word it very nicely. It's definitely ooj the list though. 👍

fourtf:
20 for the interval seems fine to me

src/widgets/helper/ChannelView.cpp
leon-richardt: I believe that this is fixed by 8f2db4d. Can't test though because I don't have a high DPI monitor.

leon-richardt:
> Chat line height affects the scrolling speed.
> If all chat lines are short enough to fit on one line in your window, the scrolling will be smooth.
> If some chat lines span more than one line, the scrolling over that line will be jagged.

Any idea on what might be causing this?

pajlada:
> > Chat line height affects the scrolling speed.
> > If all chat lines are short enough to fit on one line in your window, the scrolling will be smooth.
> > If some chat lines span more than one line, the scrolling over that line will be jagged.
> 
> Any idea on what might be causing this?

It looks to be reproducible when using the scrollbar like normal too, so I wouldn't say this is a blocker to merge the PR.

leon-richardt:
Left and right clicks now also disable scrolling.

cinemassacres:
the dev of ffz is talking to the dev of bttv  a bit about getting overlay emotes added to the BTTV API. so we won't need this hack  If that happens, then overlay emotes will start to just work.

leon-richardt:
> the dev of ffz is talking to the dev of bttv a bit about getting overlay emotes added to the BTTV API. so we won't need this hack If that happens, then overlay emotes will start to just work.

I believe you meant to post this on #1613?

pajlada:
Works great for me, the stutter on different message heights is still a thing but it's a scrollbar thing, which I think isn't up to this PR to fix

PR #1568, opened 2020-02-24
Better Highlights: Fix Wrong Color for Migrated Phrases

### LINKED ISSUE #1565: Better highlights: When upgrading version, "migrated" notification settings have black as default color
**Describe the bug**
When you upgrade chatterino2 to a version with the new highlighting updates, old phrases you had before the upgrade all have black set as their highlight color.
I don't think this should be the default experience for most users. I would expect the color to be the default "red" that was used before.

**To reproduce**
see above

**Screenshots**
example of my experience when upgrading

image
image

**Chatterino version**

`Chatterino Nightly 2.1.7 (commit 53e52e189, )`

**Operating system**
windows 10

**Additional information**

leon-richardt:
Prior to this commit, no default color was set when an "old" highlight phrase (one added prior to #1320 / 5957551) was deserialized. This commit makes highlight phrases uses the default self-highlight color for these situations. This approach is reasonably sensible since that color is also similar to the old highlight color.

Fixes #1565.

### DIFF
--- a/src/controllers/highlights/HighlightPhrase.hpp
+++ b/src/controllers/highlights/HighlightPhrase.hpp
@@ -70,6 +70,8 @@ public:
     const QUrl &getSoundUrl() const;
     const std::shared_ptr<QColor> getColor() const;
 
+    static constexpr QColor FALLBACK_COLOR = QColor(127, 63, 73, 127);
+
 private:
     QString pattern_;
     bool hasAlert_;
@@ -133,14 +135,7 @@ struct Deserialize<chatterino::HighlightPhrase> {
 
         auto _color = QColor(encodedColor);
         if (!_color.isValid())
-        {
-            /*
-             * As a fallback, use the default self-highlight color. We can't use
-             * getApp()->themes here because the Application instance is not
-             * initialized when this function is called.
-             */
-            _color = QColor(127, 63, 73, 127);
-        }
+            _color = chatterino::HighlightPhrase::FALLBACK_COLOR;
 
         return chatterino::HighlightPhrase(_pattern, _hasAlert, _hasSound,
                                            _isRegex, _isCaseSensitive,
--- a/src/controllers/highlights/HighlightPhrase.hpp
+++ b/src/controllers/highlights/HighlightPhrase.hpp
@@ -74,7 +74,7 @@ public:
      * XXX: Use the constexpr constructor here once we are building with
      * Qt>=5.13.
      */
-    static QColor FALLBACK_COLOR;
+    static QColor FALLBACK_HIGHLIGHT_COLOR;
 
 private:
     QString pattern_;
@@ -139,7 +139,7 @@ struct Deserialize<chatterino::HighlightPhrase> {
 
         auto _color = QColor(encodedColor);
         if (!_color.isValid())
-            _color = chatterino::HighlightPhrase::FALLBACK_COLOR;
+            _color = chatterino::HighlightPhrase::FALLBACK_HIGHLIGHT_COLOR;
 
         return chatterino::HighlightPhrase(_pattern, _hasAlert, _hasSound,
                                            _isRegex, _isCaseSensitive,
--- a/src/controllers/highlights/HighlightPhrase.hpp
+++ b/src/controllers/highlights/HighlightPhrase.hpp
@@ -132,6 +132,15 @@ struct Deserialize<chatterino::HighlightPhrase> {
         chatterino::rj::getSafe(value, "color", encodedColor);
 
         auto _color = QColor(encodedColor);
+        if (!_color.isValid())
+        {
+            /*
+             * As a fallback, use the default self-highlight color. We can't use
+             * getApp()->themes here because the Application instance is not
+             * initialized when this function is called.
+             */
+            _color = QColor(140, 84, 89, 127);
+        }
 
         return chatterino::HighlightPhrase(_pattern, _hasAlert, _hasSound,
                                            _isRegex, _isCaseSensitive,
--- a/src/controllers/highlights/HighlightPhrase.hpp
+++ b/src/controllers/highlights/HighlightPhrase.hpp
@@ -75,6 +75,7 @@ public:
      * Qt>=5.13.
      */
     static QColor FALLBACK_HIGHLIGHT_COLOR;
+    static QColor FALLBACK_SUB_COLOR;
 
 private:
     QString pattern_;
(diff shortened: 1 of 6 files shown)


leon-richardt:
I'm wondering whether it would more reasonable to use a `#define` for the default self-highlight color instead of having to manually update the color in two places in the code?
I'd appreciate some feedback on that idea.

pajlada:
Go ahead and make a constexpr QColor or something for the default color
It's also worth checking whether that's the old color - I think the alpha value might be different than what it was before (or maybe since alpha just wasn't applied before)

leon-richardt:
> It's also worth checking whether that's the old color - I think the alpha value might be different than what it was before (or maybe since alpha just wasn't applied before)

The old highlight color is `#4B282C` for dark themes, and `#BD8489` for light themes; both at alpha value 255.

> Go ahead and make a constexpr QColor or something for the default color

I'm not sure how to go about this. Just sticking a new `static` variable to `BaseTheme`?

src/controllers/highlights/HighlightPhrase.hpp
```diff
@@ -132,6 +132,15 @@ struct Deserialize<chatterino::HighlightPhrase> {
...
 
         auto _color = QColor(encodedColor);
+        if (!_color.isValid())
+        {
+            /*
+             * As a fallback, use the default self-highlight color. We can't use
+             * getApp()->themes here because the Application instance is not
+             * initialized when this function is called.
+             */
+            _color = QColor(127, 63, 73, 127);
```
fourtf: Why not either
- remove the color from Theme.cpp
- or use the value from Theme.cpp here

src/singletons/Theme.cpp
```diff
@@ -48,7 +48,7 @@ void Theme::actuallyUpdate(double hue, double multiplier)
         this->splits.resizeHandleBackground = QColor(0, 148, 255, 0x20);
     }
 
-    this->messages.backgrounds.highlighted = QColor(140, 84, 89, 127);
+    this->messages.backgrounds.highlighted = QColor(127, 63, 73, 127);
```
fourtf: Why was this color changed?

RAnders00: When you mix this color (alpha = 127) with the default background it gets you exactly the old highlight color

src/controllers/highlights/HighlightPhrase.hpp
leon-richardt: > * or use the value from Theme.cpp here

I'm not sure how I would go about that - `getApp()->themes.messages.backgrounds.highlighted` wouldn't work here because the `Theme` singleton has not been initialized yet when this function is called.

Please let me know if I'm being dank and missing something obvious here.



> * remove the color from Theme.cpp

Yeah, I guess that would work. Since the highlight color is theme-independent now, this may be the way to go?

src/singletons/Theme.cpp
leon-richardt: Also see https://github.com/Chatterino/chatterino2/issues/1565#issuecomment-590441625.

fourtf: So it works for light+dark backgrounds?

RAnders00: Light theme:
image

White theme:
image

Dark theme:
image

Black theme:
image

Depends on your taste whether you think it's good.

leon-richardt: In any case, it would probably be "better" as it is very close, if not identical, to the original highlight color in the dark theme (which is what most people use, I suppose). Guess it would minimize irritation when users are migrated to the new highlight system.

src/controllers/highlights/HighlightPhrase.hpp
leon-richardt: > remove the color from Theme.cpp

I went ahead and did just that in 92e211a9. Let me know if you want that changed, I'm resolving this convo for now.

src/singletons/Theme.cpp
RAnders00: That's why I suggested the color in the first place. Every existing user is going to go through the migration phrase, and you risk a bad user experience if their old experience does not translate to the new system transparently. (I can already imagine the waves of whining users - "This update sucks.")

leon-richardt:
CI fails on Ubuntu because the `QColor(int, int, int, int)` constructor is only `constexpr` since Qt 5.13 (according to [this comment](https://bugreports.qt.io/browse/QTBUG-56232?focusedCommentId=477405&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-477405)), and the [CI uses Qt 5.12.7](https://github.com/jurplel/install-qt-action/commit/f9c8e97d65206a771accd4d0536d7aaa28a214b3). Not sure why it does not fail on Windows and Mac.

src/controllers/highlights/HighlightPhrase.hpp
```diff
@@ -70,6 +70,8 @@ class HighlightPhrase
     const QUrl &getSoundUrl() const;
     const std::shared_ptr<QColor> getColor() const;
 
+    static constexpr QColor FALLBACK_COLOR = QColor(127, 63, 73, 127);
```
pajlada: is it possible to just remove constexpr here until Qt 5.13 is LTS'ed?

RAnders00:
```suggestion
    static QColor FALLBACK_COLOR = QColor(127, 63, 73, 127);
```

leon-richardt:
Is there still anything keeping this PR from being merged? (Apart from a lack of time to review, perhaps.)
I would really like to get this closed.

RAnders00:
Check the CI - it currently does not build outside of windows.

leon-richardt:
Ah, my bad -- thought the change from 1f3b356 worked on all platforms.

src/providers/colors/ColorProvider.cpp
```diff
@@ -89,7 +88,7 @@ void ColorProvider::initTypeColorMap()
     {
         this->typeColorMap_.insert(
             {ColorType::Subscription,
-             std::make_shared<QColor>(backgrounds.subscription)});
+             std::make_shared<QColor>(HighlightPhrase::FALLBACK_COLOR)});
```
pajlada: This should have its own fallback color, to emulate the sub color. Right now, it would fall back to the red highlight color which doesn't seem right :astonished:

leon-richardt: Thanks, see 5fe20b3.

pajlada:
Thank you very much @leon-richardt

Issue #1552, opened 2020-02-15
Separating emote tabbing from user tabbing


alazymeme:
I'd like to see an option (similar to chatterino 1), but "religiously" follows those rules, with no lenience. In this case, in order for a user to be tab completed, they would **have** to prefix the username with @ - but emotes do not require any prefix (or maybe even an optional : or something)

![image](https://user-images.githubusercontent.com/12804673/74590097-0e44fb00-5057-11ea-8ddc-16b281ab76fc.png)

pajlada:
I'd see a setting "Only tab usernames with an @" or something as useful :+1:

PR #1573, opened 2020-02-28
Add option for username tab completion only with an @ prefix

mammothb:
This attempts to fix https://github.com/Chatterino/chatterino2/issues/1552

When the "Only search for username autocompletion with an @" option is enabled, usernames will be omitted from the `CompletionModel` unless there's an "@" prefix

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -125,6 +125,8 @@ public:
         "/behaviour/autocompletion/smallStreamerLimit", 1000};
     BoolSetting prefixOnlyEmoteCompletion = {
         "/behaviour/autocompletion/prefixOnlyCompletion", true};
+    BoolSetting userOnlyCompletionWithAt = {
+        "/behaviour/autocompletion/userOnlyCompletionWithAt", true};
 
     FloatSetting pauseOnHoverDuration = {"/behaviour/pauseOnHoverDuration", 0};
     EnumSetting<Qt::KeyboardModifier> pauseChatModifier = {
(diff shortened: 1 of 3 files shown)


src/singletons/Settings.hpp
```diff
@@ -125,6 +125,8 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/behaviour/autocompletion/smallStreamerLimit", 1000};
     BoolSetting prefixOnlyEmoteCompletion = {
         "/behaviour/autocompletion/prefixOnlyCompletion", true};
+    BoolSetting userOnlyCompletionWithAt = {
+        "/behaviour/autocompletion/userOnlyCompletionWithAt", true};
```
fourtf: I believe this should be called `userCompletionOnlyWithAt` but otherwise seems good.

fourtf:
Looks good to me :+1:

src/singletons/Settings.hpp
mammothb: Changed

src/singletons/Settings.hpp
```diff
@@ -125,6 +125,8 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/behaviour/autocompletion/smallStreamerLimit", 1000};
     BoolSetting prefixOnlyEmoteCompletion = {
         "/behaviour/autocompletion/prefixOnlyCompletion", true};
+    BoolSetting userCompletionOnlyWithAt = {
+        "/behaviour/autocompletion/userCompletionOnlyWithAt", true};
```
pajlada: This changes the default behaviour - we should probably default this to `false`

mammothb: Changed

Issue #662, opened 2018-08-11
Emote menu doesn't remember position


wigoor:
Emote menu does not remember its position after a restart and spawns in its default center position.

All chatts spawn in the center, if you set it up on one chat/emote menu and swap (#pajlada -> #fourtf) it will be in the center untill setup again.

PR #1580, opened 2020-03-01
Make emote popup remember last position

mammothb:
This attempts to fix https://github.com/Chatterino/chatterino2/issues/662

Emote popup position is stored in "window-layout.json". `closeEvent` is reimplemented in `EmotePopup` so that the popup position is saved through `WindowManager`.

### DIFF
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -274,6 +274,16 @@ Window *WindowManager::windowAt(int index)
     return this->windows_.at(index);
 }
 
+QPoint WindowManager::emotePopupPos()
+{
+    return this->emotePopupPos_;
+}
+
+void WindowManager::setEmotePopupPos(QPoint pos)
+{
+    this->emotePopupPos_ = pos;
+}
+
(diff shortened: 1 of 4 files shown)


mammothb:
I switched to use `moveTo` (I hope that's what you meant by using existing code). If the popup is positioned on another monitor which gets disconnected, it is automatically repositioned on the monitor that's still connected. So I'm not sure if there needs to be special handling for that.

bennyhillthebest:
on my Windows 10 machine it looks like the emote popup window slowly moves diagonally down-right every time it is closed and reopened.

mammothb:
@bennyhillthebest I realized that this was caused by `offset=true` in `moveTo` and have made another PR (https://github.com/Chatterino/chatterino2/pull/1630) to fix it.

bennyhillthebest:
Sorry to bother you again, but it looks like the emote position saving mechanism is broken again. I don't know if you have the time or will to fix it again, though. Thanks anyways!!!

mammothb:
@bennyhillthebest created a PR (https://github.com/Chatterino/chatterino2/pull/1752) that should fix this.

PR #1587, opened 2020-03-01
Replace create-dmg with builddmg

pajlada:
I believe this will make more stable mac builds


pajlada:
I have verified that this works on my mac on catalina

@gempir could you please verify that the latest artifacts works on your mac? Thank you!
The artifact can be downloaded from here: https://github.com/Chatterino/chatterino2/suites/491260491/artifacts/2390872 (requires you to be logged in with a github account)

In the "About" page it should say commit `82773` something

gempir:
@pajlada works on my machine (8841c7fd6)

PR #1594, opened 2020-03-08
very stupid and ugly color code testing

pajlada:
Current progress:
image


Mm2PL:
Would this work if Twitch IRC's server returned valid color codes?

pajlada:
> Would this work if Twitch IRC's server returned valid color codes?

Right now, the implementation is specifically not enabled on Twitch Servers, so no

Mm2PL:
@pajlada, Is there a chance that this will be merged any time soon? pepeLaugh

PR #1603, opened 2020-03-14
Add rendering for new lines.

Mm2PL:
Adds rendering for new lines in messages, currently only way to get these is to use a proxy.


fourtf:
Which "newlines" does this render?

Mm2PL:
This PR adds support for the newline character `\n`. I think it's time to close this.

PR #1626, opened 2020-04-10
option to hide viewer count and duration of the stream when you hover on the split

### LINKED ISSUE #1621: Please add the option to hide viewer count and duration of the stream when you hover on the title
I use chatterino as a streamer, i already disabled the viewer count and the duration of the stream, but when i hover they're still there, i get kinda anxious when i lose or gain viewers and i don't want it to affect my performance, and when i see i've streamed close to 2 hours it usually just changes my mood instantly, i wish i couldn't see how much i've streamed.

So please add the option to disable the viewer count/duration of the stream when you hover on the title, it would be very useful!, thank you

apa420:
Fixes #1621


leon-richardt:
Should there be a separate option for this? I think using the existing setting for this would be fine.

PR #1635, opened 2020-04-18
Clarified that Redeemed Highlights are from Channel Points, not Bits

pajlada:
# Description

Previously the highlight setting said that the Redeemed highlights were bought with Bits
image
However, they're actually redeemed by Channel Points, so I updated the text.
image


pajlada:
Seems good paj!

PR #1636, opened 2020-04-18
Attempt downscaling the avatar pixmap to fit the button

### LINKED ISSUE #1611: Image scaling without interpolation
**Describe the bug**
Profile pictures are scaled down in a bad way, I think this is because there is no interpolation used

**To reproduce**
Open user card of a user that has sharp lines in their profile picture. The effect is quite noticeable e.g. on manga-style drawings etc.

**Screenshots**
(Open the image in a new tab and set zoom level to 100% to see the effect properly)
image
The black border around the goat is all jagged and missing in some places.

Just the scaled-down profile picture from chatterino (cut from the screenshot above):
image

**Chatterino version**
Chatterino Nightly 2.1.7 (commit 95c909b33, )

**Operating system**
windows 10

**Additional information**
on my computer the little preview thing on the usercard is 88 by 88 pixels.
I downloaded the profile picture from the example above and scaled it using GIMP with various interpolation algorithms.

Original image:
image

Scaled to 88x88 with Interpolation set to None:
image

Linear:
image

Cubic:
image

I think this illustrates the difference between no interpolation and linear/cubic well enough.

It's probably worth exploring whether there are other less noticeable places in Chatterino where images are scaled without interpolation, e.g. in emotes/emojis or possibly other places I can't think of in the moment.

pajlada:
This uses a smooth transformation (bilinear filtering)

Fixes #1611 ???

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -512,7 +512,9 @@ void UserInfoPopup::loadAvatar(const QUrl &url)
             // might want to cache the avatar image
             QPixmap avatar;
             avatar.loadFromData(data);
-            this->ui_.avatarButton->setPixmap(avatar);
+            auto size = this->ui_.avatarButton->size();
+            this->ui_.avatarButton->setPixmap(avatar.scaled(
+                size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation));
         }
         else
         {


fourtf:
closing in favor of https://github.com/Chatterino/chatterino2/commit/06cb88f6c28096b3246e033e92e952e71e1bfcee since if you resize the button later (e.g. dpi changes) it will resize again.

PR #1642, opened 2020-04-22
macOS update checker fix

dnsge:
# Description
On macOS, checking the web API for an update would always result in the "Failed to load version information" message. Also, the update message dialog could sometimes be off-screen to the left making the message unreadable.

Chatterino expected a JSON field "portable_download" for both Windows and macOS, but the web API only provides that field for the `/win/` endpoint. This results in macOS Chatterino always complaining that it couldn't check for an update.

Screenshot before:
before

Screenshot after: (I locally bumped version to test)
after

### DIFF
--- a/src/util/InitUpdateButton.cpp
+++ b/src/util/InitUpdateButton.cpp
@@ -16,7 +16,7 @@ void initUpdateButton(Button &button,
         dialog->setActionOnFocusLoss(BaseWindow::Delete);
 
         auto globalPoint = button.mapToGlobal(
-                    QPoint(int(-100 * button.scale()), button.height()));
+            QPoint(int(-100 * button.scale()), button.height()));
 
         // Make sure that update dialog will not go off left edge of screen
         if (globalPoint.x() < 0)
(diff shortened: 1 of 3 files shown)


src/util/InitUpdateButton.cpp
```diff
@@ -14,8 +14,17 @@ void initUpdateButton(Button &button,
...
+        auto globalPoint = button.mapToGlobal(
+            QPoint(int(-100 * button.scale()), button.height()));
+
+        // Make sure that update dialog will not go off left edge of screen
+        if (globalPoint.x() < 0)
+        {
+            globalPoint.setX(0);
+        }
+
+        dialog->move(globalPoint);
```
23rd, line 27:
```suggestion
        dialog->move(button.mapToGlobal(
            QPoint(std::max(0, int(-100 * button.scale())), button.height())));
```

dnsge: Unfortunately this is hard to make work cleanly while keeping the 100px offset because the x-coordinate must be max'd with 0 _after_ mapping to global coordinates, which (to my knowledge) can only be done through the getter and setter. 
`std::max(0, int(-100 * somePositiveValue))` is always 0 so the popup will always match the edge of the update button:
pic
To be honest I don't really see the importance of the 100px offset but I didn't want to change the design too much.

23rd: Hmm, you are right, sorry.

PR #1644, opened 2020-04-23
Handle middle mouse button on links differently

### LINKED ISSUE #1506: open links  by clicking middle mouse button

 Browsers start opening links when clicking the middle mouse button. Chatterino should do that also

**Additional context**

Behaviour should be similar to the implementation in browsers.

gempir:
Fixes #1506

Now links can be opened with middle mouse button, the scroll cursor won't activate on links anymore.

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1384,7 +1384,8 @@ void ChannelView::mousePressEvent(QMouseEvent *event)
                 {
                     break;
                 }
-            } else
+            }
+            else
             {
                 if (this->isScrolling_)
                     this->disableScrolling();
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1408,7 +1408,8 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
     QPoint relativePos;
     int messageIndex;
 
-    bool foundMessage = tryGetMessageAt(event->pos(), layout, relativePos, messageIndex);
+    bool foundMessage =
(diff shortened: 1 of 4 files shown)


gempir:
I fixed the 4 listed bugs. 

I think it would be a really cool feature to open the twitch usercard in the browser when middle clicking a user, if that's something we want then I'll implement it in this PR. (If I figure out how :D)

It's certainly a new feature and maybe not 100% expected but I think it's very useful to moderators especially.

pajlada:
> I fixed the 4 listed bugs.
> 
> I think it would be a really cool feature to open the twitch usercard in the browser when middle clicking a user, if that's something we want then I'll implement it in this PR. (If I figure out how :D)
> 
> It's certainly a new feature and maybe not 100% expected but I think it's very useful to moderators especially.

I think for this PR that feature shouldn't exist.
For now, you can use the links type to make sure only links with type `Url` are triggered by middle mouse button

gempir:
I make use of the type now. 

I also opened an issue for the feature suggestion #1651

src/widgets/helper/ChannelView.cpp
```diff
@@ -1439,25 +1461,35 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
...
-            this->disableScrolling();
+        {
+            const MessageLayoutElement *hoverLayoutElement =
+                layout->getElementAt(relativePos);
+
+            if (hoverLayoutElement == nullptr ||
+                hoverLayoutElement->getLink().isUrl() == false)
+            {
+                return;
+            }
```
pajlada, line 1482: `layout` is not guaranteed to exist yet, because we haven't checked the `foundMessage` variable yet. I'd probably just check the `foundMessage` value here instead of moving code around

PR #1664, opened 2020-05-02
Thumbnail for links

gempir:
# Description
With https://github.com/Chatterino/api/pull/14 I'm adding a thumbnail to the response of the link resolver. 

In this PR I'm adding this functionality to chatterino so you can preview links. This will definetely be off by default in the settings. 

I'm still trying to figure out how to resize the image (set a max size) any pointers would be great! 

Also I wanted to open this Draft to pose the the question if it's okay that this can potentially lead to a IP leak when someone posts a valid img link and you hover over it with the setting enabled. 

If that's not okay we either need to proxy images, which I think is a bit much. Or we only allow thumbnails for trusted urls? Not sure if that list should be in chatterino or the api. 

image

# TODOs

- [x] add max size for thumbnail
- [x] introduce setting for thumbnail (should be off by default)
- [x] think about IP leaking, since all urls with an image would return a thumbnail not hosted on chatterino api
- [x] update thumbnail size on config update

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -161,7 +161,7 @@ public:
     /// Links
     BoolSetting linksDoubleClickOnly = {"/links/doubleClickToOpen", false};
     BoolSetting linkInfoTooltip = {"/links/linkInfoTooltip", false};
-    FloatSetting thumbnailSize = {"/appearance/thumbnailSize", 0};
+    IntSetting thumbnailSize = {"/appearance/thumbnailSize", 0};
     BoolSetting unshortLinks = {"/links/unshortLinks", false};
     BoolSetting lowercaseDomains = {"/links/linkLowercase", true};
 
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -292,6 +292,7 @@ public:
         "/misc/attachExtensionToAnyProcess", false};
     BoolSetting hideViewerCountAndDuration = {
         "/misc/hideViewerCountAndDuration", false};
+    BoolSetting askOnImageUpload = {"/misc/askOnImageUpload", true};
 
     /// Debug
     BoolSetting showUnhandledIrcMessages = {"/debug/showUnhandledIrcMessages",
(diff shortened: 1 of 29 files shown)


alazymeme:
I like this :+1:

Mm2PL:
Maybe the IP leak problem could be semi-solved using an option.
By default we could allow only certain services, like `imgur` or `i.nuuls.com`, the user could then decide to load more thumbnails.

gempir:
We will solve this IP leak issue by caching all images besides YT/Twitch Clips in the chatterino-api

gempir:
This is ready for review now. Requires https://github.com/Chatterino/api/pull/14 to show any thumbnails.

API only supports gif, png and jpeg. Gifs are static for now, I might fix them at a later date in another PR.

And it looks something like this 

image

src/singletons/Settings.hpp
```diff
@@ -161,6 +161,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     /// Links
     BoolSetting linksDoubleClickOnly = {"/links/doubleClickToOpen", false};
     BoolSetting linkInfoTooltip = {"/links/linkInfoTooltip", false};
+    FloatSetting thumbnailSize = {"/appearance/thumbnailSize", 0};
```
pajlada, line 164: Can we make this an IntSetting instead of a FloatSetting

src/widgets/helper/ChannelView.cpp
```diff
@@ -1288,7 +1289,22 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
         }
         else
         {
-            tooltipPreviewImage.setImage(nullptr);
+            auto element = &hoverLayoutElement->getCreator();
+            auto thumbnailSize = getSettings()->thumbnailSize;
+            if (thumbnailSize == 0.f)
```
pajlada, line 1294: remember to modify this when we change the setting from a float to an int

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -521,6 +521,31 @@ void GeneralPage::initLayout(SettingsLayout &layout)
...
+        },
+        [](auto args) {
+            if (args.value == "Small")
+                return 100.f;
+            else if (args.value == "Medium")
+                return 200.f;
+            else if (args.value == "Large")
+                return 300.f;
+
+            return fuzzyToFloat(args.value, 0.f);
```
pajlada, line 547: remember to modify this when we change the setting from a float to an int

PR #1671, opened 2020-05-06
Show visual feedback when BTTV and FFZ emotes are loaded

thekalio:
Upon joining a channel or pressing F5, BTTV and FFZ emotes are
(re)loaded. This change adds visual feedback of the network requests and
their outcome, in the form of a system message in the associated
channel's chat window. These are the 8 possible system messages that can
be shown:

BTTV: emotes reloaded.
BTTV: this channel has no emotes.
BTTV: failed to fetch emotes. (timed out)
BTTV: failed to fetch emotes. (unknown error)
FFZ: emotes reloaded.
FFZ: this channel has no emotes.
FFZ: failed to fetch emotes. (timed out)
FFZ: failed to fetch emotes. (unknown error)

### DIFF
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -144,13 +144,14 @@ void BttvEmotes::loadChannel(TwitchChannel &channel, const QString &channelId,
 {
     NetworkRequest(QString(bttvChannelEmoteApiUrl) + channelId)
         .timeout(3000)
-        .onSuccess([callback = std::move(callback), &channel](auto result) -> Outcome {
-            auto pair = parseChannelEmotes(result.parseJson());
-            if (pair.first)
-                callback(std::move(pair.second));
-            channel.addMessage(makeSystemMessage("BTTV: emotes reloaded."));
-            return pair.first;
-        })
+        .onSuccess(
+            [callback = std::move(callback), &channel](auto result) -> Outcome {
+                auto pair = parseChannelEmotes(result.parseJson());
+                if (pair.first)
+                    callback(std::move(pair.second));
+                channel.addMessage(makeSystemMessage("BTTV: emotes reloaded."));
+                return pair.first;
+            })
         .onError([channelId, &channel](auto result) {
             if (result.status() == 203)
             {
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -141,28 +141,31 @@ void BttvEmotes::loadEmotes()
 
 void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
                              const QString &channelId,
-                             std::function<void(EmoteMap &&)> callback)
+                             std::function<void(EmoteMap &&)> callback,
+                             bool manualRefresh)
 {
     NetworkRequest(QString(bttvChannelEmoteApiUrl) + channelId)
         .timeout(3000)
-        .onSuccess([callback = std::move(callback),
-                    channel](auto result) -> Outcome {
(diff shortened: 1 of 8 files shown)


src/providers/bttv/BttvEmotes.cpp
```diff
@@ -138,16 +139,41 @@ void BttvEmotes::loadEmotes()
         .execute();
 }
 
-void BttvEmotes::loadChannel(const QString &channelId,
+void BttvEmotes::loadChannel(TwitchChannel &channel, const QString &channelId,
```
pajlada, line 142: We're passing in a reference to TwitchChannel here, which will be sent to an asynchronous function. We have no way of ensuring that the TwitchChannel is still alive once we get the response from the network call.

The way this is solved can be seen in the TwitchChannel::refreshXXXEmotes call, where we send in a weak pointer to the channel instead of a straight up reference.

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -138,16 +139,41 @@ void BttvEmotes::loadEmotes()
...
+                    callback(std::move(pair.second));
+                channel.addMessage(makeSystemMessage("BTTV: emotes reloaded."));
+                return pair.first;
+            })
+        .onError([channelId, &channel](auto result) {
+            if (result.status() == 203)
+            {
+                // User does not have any BTTV emotes
+                channel.addMessage(
+                    makeSystemMessage("BTTV: this channel has no emotes."));
```
pajlada, line 160: For any public-facing messages, I think we should stick to the preferred branding of the service. Here it's "BetterTTV". and making the messages just a proper sentence (e.g. "This channel has no BetterTTV channel emotes") feels better to me.

src/providers/ffz/FfzEmotes.cpp
```diff
@@ -192,32 +195,39 @@ void FfzEmotes::loadChannel(
...
-                        std::move(modBadgeCallback)](auto result) -> Outcome {
+                    modBadgeCallback = std::move(modBadgeCallback),
+                    &channel](auto result) -> Outcome {
             auto json = result.parseJson();
             auto emoteMap = parseChannelEmotes(json);
             auto modBadge = parseModBadge(json);
 
             emoteCallback(std::move(emoteMap));
             modBadgeCallback(std::move(modBadge));
+            channel.addMessage(makeSystemMessage("FFZ: emotes reloaded."));
```
pajlada, line 206: For any public-facing messages, I think we should stick to the preferred branding of the service. Here it's "FrankerFaceZ". and making the messages just a proper sentence (e.g. "FrankerFaceZ channel emotes reloaded.") feels better to me.

src/providers/bttv/BttvEmotes.cpp
thekalio: Fixed.

src/providers/bttv/BttvEmotes.cpp
thekalio: Fixed.

src/providers/ffz/FfzEmotes.cpp
thekalio: Fixed.

thekalio:
Thanks for the feedback! I made the requested changes.

As for whether to show the status messages upon entering a new channel... I agree it may be too "debuggy" for most users. The newest commit suppresses non-error messages upon entering a channel, but they will still be shown when refreshing through F5 or "Reload channel
emotes" in the split header menu.

pajlada:
Thanks for your contribution!
Feel free to send another PR if you want to be added to the Contributors list: https://github.com/Chatterino/chatterino2/blob/master/resources/contributors.txt#L1

PR #1674, opened 2020-05-06
Open usercard on mention click

### LINKED ISSUE #1598: Usercard popup from clicking a mention
As this issue was previously closed, I'll make a new issue without the duplicate stuff.
ffz has a neat feature where you are able to click @mentions in chat and it will show their usercard. Would like to see something similar in ch2. I think this along with the integrated logs in the usercard would make for an extremely useful feature.

dnsge:
Clicking on a mention in chat will now open up the usercard popup. The behavior is the same as clicking the username of an author of a message.

[GIF of it in action](https://media.giphy.com/media/MBkxJ5B1rnOl8Tsrj3/giphy.gif)

Closes #1598


leon-richardt:
One issue I found: It only works when the user has the option `Bold @usernames` enabled. Also I find it confusing to only work for "@" mentions and not for just the user name, although this behavior is probably up for debate.
(I am aware that this would significantly increase complexity of the implementation.)

dnsge:
You're right, I missed adding the `setLink` for non-bolded mentions. Just fixed in f7d41d5.

As for all usernames in chat being clickable, it can definitely be implemented though I'm not sure if it would add too much overhead for message processing. Maybe it could be added as a toggleable setting.

apa420:
I think that doing it only for @ messages is the only real viable solution, I don't really see the point of checking the entire viewer list for each word in the message, doesn't seem viable at all.

Mm2PL:
It probably opens an invalid user card.

dnsge:
Yes, it opens an invalid user card. I wrote up a little bit of regex validation that will also handle things like `@username,` or `@username!!!` to extract the proper username (though that doesn't guarantee that the user exists on twitch).

As a side note, I did write a simple implementation of checking the channel viewer list for each word and didn't see any visible performance impact when comparing it side by side even on a huge 30k viewer stream. So I suppose that feature could work if we wanted.

dnsge:
Unfortunately, a user's color is only sent in the tags of an IRC message, so it would be pretty hard to set the color. You could potentially keep a map of names and colors that updates with each message sent, but that only works if the mentioned user has recently sent a message in the chat. See https://discuss.dev.twitch.tv/t/users-chat-color/1681/2 

In commit 1dcea4d, you can find the implementation for linking usernames without an @. I'm not sure if the settings label is super clear, though. If anyone comes up with something better let me know. 
https://github.com/Chatterino/chatterino2/blob/1dcea4d4143bbca62794097eac41469cd0111b81/src/widgets/settingspages/GeneralPage.cpp#L514-L515

was the best I could think of.

0xRainy:
Thanks, and the fact that you can only get a username's color if they've sent a message you have seen is not bad at all.

If the username is colored that means they are in chat and you have seen them (the most common use case for @ing somebody) and if it isn't then you know that they aren't around. I think that behavior is fine. The implementation could be a lookup for their last message and grabbing the color, instead of storing it.

This might not work too well with the default number of messages kept in chat but I am also trying to increase that to a crazy high number, right now I'm testing 100,000 messages. (Still don't know if it is actually keeping that many messages around but I can definitely scroll back way further)

dnsge:
Yeah, unfortunately, AFAIK the only way to get present users is through the `tmi.twitch.tv` endpoint every five minutes or through `JOIN` and `PART` IRC messages, though those are only sent on <1000 viewer streams.

0xRainy:
Would it be possible to turn every username into a link to the UserInfoPopup, for example usernames in ban/timeout messages?

This would be useful to check why somebody was timed out if their message is already scrolled way up and they were timed out later.

0xRainy:
Amazing work on the timeout/ban links!

Do you think you can take a look at https://github.com/Chatterino/chatterino2/pull/1729#issuecomment-643664403, I'm out of my depth and it is related to this PR.

dnsge:
@0xRainy I've added a fix for clicking mentions within the usercard to [this PR](https://github.com/0xRainy/chatterino2/pull/2).

0xRainy:
Merged this with master and discovered an issue.

When mentioning somebody with @ , the message is reported to be sent in console but no message appears in chat and does not get sent to twitch.

Also with setting `Mention users without prefix or suffix` on mentions with a `,` aren't links and with the setting turned off and `Mention users with a comma` turned on, they aren't links either.  The same goes for mentions with suffix : but I assume that is intended as it isn't in the regex match.

dnsge:
Are you sure this wasn't an issue with Twitch? They were having server issues around the time you commented.

0xRainy:
Yeah I just tested again and my other test messages were being sent.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -511,6 +511,8 @@ void GeneralPage::initLayout(SettingsLayout &layout)
     layout.addCheckbox("Show parted users (< 1000 chatters)", s.showParts);
     layout.addCheckbox("Lowercase domains (anti-phishing)", s.lowercaseDomains);
     layout.addCheckbox("Bold @usernames", s.boldUsernames);
+    layout.addCheckbox("Mention users without prefix or suffix",
```
pajlada, line 514: This should probably be something along the lines of "Try to find usernames without @ prefix"

The variable name should also somewhat match that

src/messages/MessageBuilder.hpp
```diff
@@ -68,6 +72,9 @@ class MessageBuilder
     }
 
 private:
+    TextElement *emplaceSystemTextAndUpdate(const QString &text,
```
pajlada, line 75: Could you write a comment clarifying what this does?

src/widgets/settingspages/GeneralPage.cpp
dnsge: Updated in 44263e7

src/messages/MessageBuilder.hpp
dnsge: Added in 44263e7

dnsge:
Packaging step failed on ubuntu for unknown reasons but built fine :)

pajlada:
Could you add a changelog entry in the form of "what this change adds (#1674)"? I'll merge after that :+1:

PR #1680, opened 2020-05-09
Added chatterino as the application 

apa420:
I've started using dwm and wanted to make a special rule for the chatterino classname to all go to the same tag (Desktop). But chatterino doesn't right now provide a WM_CLASS variable in X11 for those windows. This will add that and a little more info about the application.

# Description

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -21,6 +21,10 @@ int main(int argc, char **argv)
 {
     QApplication a(argc, argv);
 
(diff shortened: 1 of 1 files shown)


apa420:
Also I didn't add organizationname since it messed up where the settings were stored (from chatterino to chatterino/chatterino)

gempir:
I think this might have broken Dark Mode support in macOS? 
image

PR #1685, opened 2020-05-12
move tooltip when no space is left on the screen to opposite side of cusor

### LINKED ISSUE #1683: Tooltip under the mouse when at the edge of screen
**Describe the bug**
Tooltips can grow pretty large now since Chatterino supports thumbnails for links now with up to 300px size. It can happen that the tooltip so big that it will clamp to the edge of the screen and cause it to go over/under the mouse and link itself causing visual bugs (sometimes the thumbnail disappears completely)

This is a problem with all kind of tooltips not just with thumbnails, but especially obvious with those.

**To reproduce**
Bring chatterino close to the edge of one of your screens and hover a tooltip that would go beyond the screen (it will then attach to the right/bottom of the screen and lay under/over the cursor)

**Screenshots**
iBCmK

**Chatterino version**
8532c6d3bca4e6c3e2da279df1224ebc489c2dc2

**Operating system**
Windows 10

gempir:
Moves the tooltip so it can no longer land under the mouse.
fixes #1683 

And it looks something like this:
toolip2

### DIFF
--- a/src/widgets/BaseWindow.cpp
+++ b/src/widgets/BaseWindow.cpp
@@ -571,6 +571,9 @@ void BaseWindow::moveIntoDesktopRect(QWidget *parent)
     QRect s = desktop->availableGeometry(parent);
     QPoint p = this->pos();
 
+    bool stickRight = false;
+    bool stickBottom = false;
+
     if (p.x() < s.left())
     {
         p.setX(s.left());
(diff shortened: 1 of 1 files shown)


apa420:
I would prefer it to keep the links to the right always like they currently are done instead of flipping them to the right when there's no space.
Like this:
https://i.nuuls.com/2mLX9.mp4

gempir:
I updated the behaviour and updated the gif in the original post. Now it will stick to the right like before but flip up once no space is left

SLCH:
Could you reduce the blinking/switching behavior while viewing the same link.

Like make it not use the position of sticking middle of toolbar side to the mouse? But again, questionable, could use sides if not much vertical space.
To preferably use corners, on no space slide on bottom/top side of toolbar?
Also not to render if no space at current chosen position.

gempir:
I could look into that in a seperate PR. The way tooltips are rendered and updated would probably have to reworked quite a bit.

PR #1693, opened 2020-05-17
improve tooltips by only positioning the element once we know it's in…

gempir:
Tooltips were always a little bit flickering because they first get positionioned and then get moved to fit into the screen. 

Now I first calculate the correct position of the tooltip and then position it. 
I also removed the stay in screen option since nothing needed it and I don't see why it would. 

If we need it again it's trivial to add again.

### DIFF
--- a/src/widgets/BaseWindow.cpp
+++ b/src/widgets/BaseWindow.cpp
@@ -220,6 +220,18 @@ void BaseWindow::init()
 #endif
 }
 
+void BaseWindow::setStayInScreenRect(bool value)
+{
+    this->stayInScreenRect_ = value;
+
+    this->moveIntoDesktopRect(this, this->pos());
+}
+
+bool BaseWindow::getStayInScreenRect() const
+{
+    return this->stayInScreenRect_;
+}
+
 void BaseWindow::setActionOnFocusLoss(ActionOnFocusLoss value)
 {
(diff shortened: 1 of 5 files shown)


gempir:
@SLCH does this fix your issue with tooltips?

SLCH:
Looks nice, flicker is gone.

One thing though.
<details>
<summary>gif</summary>
I'm not sure, should this be fixed? Looks fine to me, but still kinda flickery.

Peek 2020-05-17 15-53


</details>

gempir:
@SLCH I think that only happens because your cursor changes vertical position a tiny bit and then there is not enough space

SLCH:
Yes, that's what i'm talking about. Such situation does exist, but should it be fixed? I think it's fine as it is now.

gempir:
What would the fix even be? I don't see a good solution for the problem.

SLCH:
Knowing link text's geometry could do it.
Don't know if it is accessible easily in ch2 right now though.

apa420:
This looks great as it is!

SLCH:
Agreed with apa.
I'm just nit-picking to perfection at this point.

gempir:
@pajlada

gempir:
@pajlada I reverted the stayInScreenRect changes

PR #1702, opened 2020-05-22
Thumbnails for stream inside of the splitheader tooltip

apa420:
# Description
Fetches a thumbnail for a stream on twitch and puts it into the splitheader tooltip.
This is the only thing I miss from only running streamlink

Issues right now:
- ~You have to hover over the header twice before it actually displays the thumbnail.~
- ~I think it downloads the thumbnail every time you hover it~
- Setting in general for disabling it because of data usage

How it currently looks:

### DIFF
--- a/src/widgets/splits/SplitHeader.cpp
+++ b/src/widgets/splits/SplitHeader.cpp
@@ -90,9 +90,12 @@ namespace {
             <p class=\"center\">%1%2%3%4%5%6 for %7 with %8 viewers</p>")
             .arg(s.title.toHtmlEscaped())
             .arg(s.title.isEmpty() ? QString() : "<br><br>")
-            .arg(thumbnail.isEmpty() ? "Couldn't fetch thumbnail"
-                                     : "<img src=\"data:image/jpg;base64, " +
-                                           thumbnail + "\"/><br>")
+            .arg(getSettings()->thumbnailSizeStream.getValue() > 0
+                     ? (thumbnail.isEmpty()
+                            ? "Couldn't fetch thumbnail"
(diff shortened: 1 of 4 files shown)


apa420:
Maybe there just shouldn't be a setting for disabling it, a tiny thumbnail shouldn't be that much bigger in size than some emotes

gempir:
> Maybe there just shouldn't be a setting for disabling it, a tiny thumbnail shouldn't be that much bigger in size than some emotes

I think a setting would be overkill. But I'm wondering about this from a streamers perspective. What if a channel has a thumbnail that breaks ToS. Rare case but people have been banned before for showing images hosted by twitch that were against the ToS (user pictures)

apa420:
You'd need to hover the split header for this to appear. I don't see this different from clicking someones name in chat.

leon-richardt:
> Maybe there just shouldn't be a setting for disabling it, a tiny thumbnail shouldn't be that much bigger in size than some emotes

I would appreciate a setting for disabling it honestly. (You can turn off images in emote tooltips too, for example.)

src/widgets/splits/SplitHeader.cpp
```diff
@@ -569,12 +569,25 @@ void SplitHeader::updateChannelText()
...
+            {
+                case 1:
+                    url.append("-80x45.jpg");
+                    break;
+                case 2:
+                    url.append("-160x90.jpg");
+                    break;
+                case 3:
+                    url.append("-360x180.jpg");
+                    break;
```
pajlada, line 588: I would like to see a `default` case where url is set to an empty string or something which can then be checked in the `if`-case below

PR #1704, opened 2020-05-24
Add ability to highlight messages based on user badges

### LINKED ISSUE #1696: Badge based highlights
**What should be added?**
The ability to set highlights based on badges. This is a very useful feature in FFZ that I would like to see come to chatterino.
image

**Why should it be added?**
In the 10k+ viewer streamers the streamer will normally be talking mostly to these people as much of the rest of the chat will be emote spam so it helps keep up with the flow of whats going on. These days I tend to put twitch on the TV but then still need to open popup chat on my pc or chatterino. It tends to be harder to keep up with chatterino.

dnsge:
# Description

Adds badge-based highlights to Chatterino. Currently, selectable badges are hardcoded. Users can select from the following badges: 

- Broadcaster
- Admin
- Staff
- Moderator
- Verified
- VIP
- ~~Global Moderator~~

Demo video: https://i.nuuls.com/8Fc_-.mp4

Closes #1696


Mm2PL:
The global moderator program shut down in December 2018 https://blog.twitch.tv/en/2018/12/13/thank-you-global-moderators-4d44cfccf22/

dnsge:
I removed the global moderator badge from the dropdown in 19728ad. Thanks for pointing that out.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1166,6 +1166,53 @@ void TwitchMessageBuilder::parseHighlights()
...
+
+                // Use custom sound if set, otherwise use fallback sound
+                if (highlight.hasCustomSound())
+                {
+                    this->highlightSoundUrl_ = highlight.getSoundUrl();
+                }
+                else
+                {
+                    this->highlightSoundUrl_ = getFallbackHighlightSound();
+                }
```
23rd, line 1203:
```suggestion
                this->highlightSoundUrl_ = highlight.hasCustomSound()
                                               ? highlight.getSoundUrl()
                                               : getFallbackHighlightSound();
```

src/common/GlobalBadges.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+void GlobalBadges::getBadgeIcons(const QList<DisplayBadge> &badges,
+                                 BadgeIconCallback callback)
+{
+    QListIterator i(badges);
+    while (i.hasNext())
+    {
+        this->getBadgeIcon(i.next(), callback);
+    }
```
23rd, line 111:
```suggestion
    for (const auto &item : badges)
    {
        this->getBadgeIcon(item, callback);
    }
```

src/widgets/dialogs/BadgePickerDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void BadgePickerDialog::initializeValues()
+{
+    QListIterator i(this->badges_);
+    while (i.hasNext())
+    {
+        auto item = i.next();
+        this->dropdown_.addItem(item.displayName(), item.identifier());
+    }
```
23rd, line 47:
```suggestion
    for (const auto &item : this->badges_)
    {
        this->dropdown_.addItem(item.displayName(), item.identifier());
    }
```

src/widgets/dialogs/BadgePickerDialog.cpp
```diff
@@ -0,0 +1,90 @@
...
+boost::optional<DisplayBadge> BadgePickerDialog::getSelection() const
+{
+    QString identifier = this->dropdown_.currentData().toString();
+    QListIterator i(this->badges_);
+    while (i.hasNext())
+    {
+        auto item = i.next();
+        if (item.identifier() == identifier)
+            return item;
+    }
```
23rd, line 73:
```suggestion
    const auto i = this->dropdown_.currentIndex();
    if (i >= 0 && i < this->badges_.size())
    {
        return this->badges_[i];
    }
```

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -130,6 +143,53 @@ HighlightingPage::HighlightingPage()
...
+                    d->setWindowTitle("Choose badge");
+                    if (d->exec() == QDialog::Accepted)
+                    {
+                        auto s = d->getSelection();
+                        getSettings()->highlightedBadges.append(
+                            HighlightBadge{s->badgeName(), s->badgeVersion(),
+                                           s->displayName(), false, false, "",
+                                           ColorProvider::instance().color(
+                                               ColorType::SelfHighlight)});
+                    }
```
23rd, line 184:
```suggestion
                    if (d->exec() != QDialog::Accepted)
                    {
                        return;
                    }
                    auto s = d->getSelection();
                    if (!s)
                    {
                        return;
                    }
                    getSettings()->highlightedBadges.append(
                        HighlightBadge{s->badgeName(), s->badgeVersion(),
                                       s->displayName(), false, false, "",
                                       ColorProvider::instance().color(
                                           ColorType::SelfHighlight)});
```

src/providers/twitch/TwitchBadges.cpp
```diff
@@ -19,43 +19,46 @@ void TwitchBadges::loadTwitchBadges()
...
 
-                    (*badgeSets)[key][vIt.key()] =
-                        std::make_shared<Emote>(emote);
+                        (*badgeSets)[key][vIt.key()] =
+                            std::make_shared<Emote>(emote);
+                    }
                 }
             }
-
+            this->loaded.invoke();
```
23rd, line 61: Can't get why an extra scope is needed here.

dnsge: `this->badgeSets_.access()` returns an `AccessGuard` that will be unlocked when it goes out of scope. `badgeSets` must be unlocked before `this->loaded.invoke()` is called, otherwise functions called by `invoke()` would be deadlocked if they try to access `badgeSets`.

src/widgets/settingspages/HighlightingPage.cpp
dnsge: I think it's better to stick with the convention already used:
https://github.com/Chatterino/chatterino2/blob/6d5ba0c442681fb30afd01f787a10cd3922c10aa/src/widgets/dialogs/SelectChannelDialog.cpp#L153-L156

https://github.com/Chatterino/chatterino2/blob/6d5ba0c442681fb30afd01f787a10cd3922c10aa/src/widgets/helper/NotebookTab.cpp#L77-L81

But I did add the optional checking (4ac85770)
```c++
if (!s)
{
    return;
}
```

src/providers/twitch/TwitchMessageBuilder.cpp
dnsge: Changed in 4ac85770

src/common/GlobalBadges.cpp
dnsge: Changed in 4ac85770

src/widgets/dialogs/BadgePickerDialog.cpp
dnsge: Changed in 4ac85770

src/widgets/dialogs/BadgePickerDialog.cpp
dnsge: Changed in 4ac85770

23rd:
@dnsge Please check [this pull request](https://github.com/dnsge/chatterino2/pull/1) to current branch.

23rd:
So twitch badge resources should be removed?

src/common/GlobalBadges.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    if (this->badgesMap_.contains(identifier))
+    {
+        callback(identifier, this->badgesMap_[identifier]);
+    }
+    else
+    {
+        auto parts = identifier.split(".");
+        auto badge = this->badges_->badge(parts.at(0), parts.at(1));
+        if (badge)
```
23rd, line 90:
```suggestion
        if (const auto badge = this->badges_->badge(parts.at(0), parts.at(1)))
```

dnsge:
> So twitch badge resources should be removed?

https://github.com/Chatterino/chatterino2/blob/6d5ba0c442681fb30afd01f787a10cd3922c10aa/src/messages/MessageBuilder.cpp#L36-L37

This is the only place where the resources are being used (99% sure), and it's only the automod badge. This could probably be replaced with `GlobalBadges` now, or it could stay. Unless there's some usage I didn't notice or we want them for the future, I suppose the rest can be removed.

src/common/GlobalBadges.cpp
dnsge: Changed in 20bd714

src/common/GlobalBadges.cpp
```diff
@@ -0,0 +1,124 @@
+#include "GlobalBadges.hpp"
```
pajlada, line 1: I would much prefer if we were able to reuse the `TwitchBadges` class for this. This means that if the user does not have a network connection, we might not be able to provide a "pretty name" of a badge, nor an icon, but I believe that's worth it

`loadEmoteImage` can basically be replaced with `TwitchBadges->badge(badge_identifier, "")`

src/controllers/highlights/HighlightBadge.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    other.color_);
+}
+
+HighlightBadge::HighlightBadge(const QString &badgeName,
+                               const QString &badgeVersion,
+                               const QString &displayName, bool hasAlert,
+                               bool hasSound, const QString &soundUrl,
+                               QColor color)
+    : badgeName_(badgeName)
+    , badgeVersion_(badgeVersion)
```
pajlada, line 24: What's the purpose of `badgeVersion`?

pajlada:
Would also love if you could resolve the merge conflicts.
In short, the highlight code has been moved to `SharedMessageBuilder`. It's fine to keep the highlighting of twitch-specific things in there for now, I will clean up `SharedMessageBuilder` at a later point.

src/controllers/highlights/HighlightBadge.cpp
dnsge: `badgeVersion` is the version component of the badge sent in IRC tags: https://dev.twitch.tv/docs/irc/tags#globaluserstate-twitch-tags (under 'badges')

Both the name and version of a badge must match for it to be highlighted

src/common/GlobalBadges.cpp
dnsge: `GlobalBadges` wraps TwitchBadges to manage the downloading of the actual badge image, but still uses `TwitchBadges` underneath to load https://badges.twitch.tv/v1/badges/global/display. `GlobalBadges` uses callbacks to update the UI without using a QPainter as normal badges/emotes do. I think that `GlobalBadges` is valuable because  `TwitchBadges` requires the UI to be rebuilt when emotes are loaded (currently after waiting/polling because loading an image has no signal) but `GlobalBadges` fixes those issues.

Hopefully that was coherent

alazymeme:
Looks like this one's been pushed to the bottom of the list, can we fix merge conflicts and merge it soon? PauseChamp

src/messages/SharedMessageBuilder.cpp
```diff
@@ -30,6 +30,34 @@ namespace {
         }
     }
 
+    QStringList parseTagList(const QVariantMap &tags, const QString &key)
```
pajlada, line 33: Duplicate of TwitchMessageBuilder's `parseTagList` - fine to keep duplicate for now

src/messages/SharedMessageBuilder.cpp
```diff
@@ -30,6 +30,34 @@ namespace {
...
+    {
+        auto iterator = tags.find(key);
+        if (iterator == tags.end())
+            return QStringList{};
+
+        return iterator.value().toString().split(
+            ',', QString::SplitBehavior::SkipEmptyParts);
+    }
+
+    std::vector<Badge> parseBadges(const QVariantMap &tags)
```
pajlada, line 43: Duplicate of TwitchMessageBuilder's `parseBadges` - fine to keep duplicate for now

src/controllers/highlights/HighlightBadge.cpp
pajlada: > Currently this is used only for subscriber, to indicate the exact number of months the user has been a subscriber. This number is finer grained than the version number in badges. For example, a user who has been a subscriber for 45 months would have a badge-info value of 45 but might have a badges version number for only 3 years.

We don't have the ability to highlight on subscribers, and we don't have a way in the UI to highlight based on the version, so this feels like a future failure case.

Example: If someone has `vip` highlighted for version `1` (which the user has no control over) then Twitch updates the version in the IRC tags to update the image, then the highlight doesn't go through anymore.

src/common/GlobalBadges.cpp
pajlada: `GlobalBadges` still doesn't feel like it serves a proper purpose for me. Every useful addition could (and should) fit into the `TwitchBadges` class instead.

Add the `getBadgeIcon` variants to `TwitchBadges`.
Make `TwitchBadges` a singleton.
Modify the places that previously owned their own `TwitchBadges` instance use the singleton instead.

pajlada (changes requested):
I've gone ahead and fixed the merge conflicts, some changes are still required, in particular the `GlobalBadges` class and how we use `badgeVersion`.

It's fine if the changelog checker and test checkers fail for now, I have help you resolve those when it's ready to merge.

dnsge:
@pajlada I've gone ahead and made the requested changes, should be good to go pending the multiplatform build checks. 😄 

Also, do we want to add any more badges to allow users to highlight? Currently, just the following are available:
- Broadcaster
- Admin
- Staff
- Moderator
- Verified
- VIP

For reference, they're set here:
https://github.com/dnsge/chatterino2/blob/5631d33a34966283f4a091570950d0ec9c79374b/src/widgets/settingspages/HighlightingPage.cpp#L34-L41

zneix:
> do we want to add any more badges to allow users to highlight?

Twitch has recently added predictions badges. I added proper message elements for those in #2668. You could try and add support for those in this PR.

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -38,6 +38,8 @@ namespace {
         {"Moderator", "moderator"},
         {"Verified", "partner"},
         {"VIP", "vip"},
+        {"Predicted One", "predictions/blue-1,predictions/pink-1"},
+        {"Predicted Two", "predictions/pink-2,predictions/blue-2"},
```
zneix, line 42: Correct me if I'm wrong, but I believe that `Predicted pink` and `Predicted blue` would fit a bit better in here, As those are usually differed by colors in twitch's frontend.

dnsge: This was one thing that I was confused about, as I had never seen Pink 1 or Blue 2 in predictions before. Just keeping Blue 1 and Pink 2 is probably all that's needed, but I wanted to cover the possible edge case. I can combine them.

src/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ set(SOURCE_FILES main.cpp
         common/DownloadManager.hpp
         common/Env.cpp
         common/Env.hpp
+        common/GlobalBadges.cpp
+        common/GlobalBadges.hpp
```
pajlada, line 30:
```suggestion
```

src/widgets/dialogs/BadgePickerDialog.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+BadgePickerDialog::BadgePickerDialog(QList<DisplayBadge> badges,
+                                     QWidget *parent)
+    : QDialog(parent)
+{
+    this->dropdown_ = new QComboBox;
+    auto vbox = new QVBoxLayout(this);
+    auto buttonBox = new QHBoxLayout;
+    auto okButton = new QPushButton("OK");
+    auto cancelButton = new QPushButton("Cancel");
```
pajlada, line 17: Use https://doc.qt.io/qt-5/qdialogbuttonbox.html instead of a 'home made' layout/QPushButtons

src/common/GlobalBadges.cpp
pajlada: :+1:

src/widgets/dialogs/BadgePickerDialog.cpp
pajlada: Fixed in [`e63110e` (#1704)](https://github.com/Chatterino/chatterino2/pull/1704/commits/e63110ef2fda3c8b4e352cca52b35e18748a044b)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added badge highlighting. (#1704)
```
zneix, line 5: This should be more descriptive and have Major changelog entry instead of minor. It's definietly one of the bigger things so far since v2.3.0

src/widgets/dialogs/BadgePickerDialog.cpp
```diff
@@ -0,0 +1,71 @@
+#include "BadgePickerDialog.hpp"
+#include <QSizePolicy>
+#include "singletons/Resources.hpp"
+
+#include "providers/twitch/TwitchBadges.hpp"
```
zneix, line 5: What's with the weird order of those includes?

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -90,7 +107,7 @@ HighlightingPage::HighlightingPage()
                     "Play notification sounds and highlight messages from "
                     "certain users.\n"
                     "User highlights are prioritized over message "
-                    "highlights.");
+                    "and badge highlights.");
```
zneix, line 110: What's the exact order here? This message says that User higlights are the most important, but then there's no information whether message or badge highlights are more important.

pajlada (approved):
Looks good to me :+1:

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -209,73 +278,92 @@ HighlightingPage::HighlightingPage()
...
+    {
+        using Column = BadgeHighlightModel::Column;
+        if (clicked.column() == Column::SoundPath)
+        {
+            this->openSoundDialog(clicked, view, Column::SoundPath);
+        }
+        else if (clicked.column() == Column::Color)
+        {
+            this->openColorDialog(clicked, view, tab);
+        }
```
zneix, line 366: Can this be a switch(tab)-case statement instead? Maybe it's not necessary, but since we're using fancy enums instead of strings it might be worth looking into in case we're gonna add any other kind of highlights in the future.

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -209,73 +278,92 @@ HighlightingPage::HighlightingPage()
...
+            if (tab == HighlightTab::Messages)
             {
-                view->getModel()->setData(clicked, selected,
-                                          Qt::DecorationRole);
-
-                if (fromMessagesTab)
+                /*
+                 * For preset highlights in the "Messages" tab, we need to
+                 * manually update the color map.
+                 */
```
zneix, line 319: nitpick: this comment could be moved above line 314.

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -209,73 +278,92 @@ HighlightingPage::HighlightingPage()
 
     // ---- misc
     this->disabledUsersChangedTimer_.setSingleShot(true);
-}  // namespace chatterino
+}
```
zneix, line 281: Why was the namespace comment removed from here?
```suggestion
} // namespace chatterino
```

src/providers/twitch/TwitchBadges.hpp
```diff
@@ -5,8 +5,14 @@
 #include <unordered_map>
 
 #include "common/UniqueAccess.hpp"
+#include "messages/Image.hpp"
+#include "util/DisplayBadge.hpp"
 #include "util/QStringHash.hpp"
 
+#include "pajlada/signals/signal.hpp"
+
+#include <shared_mutex>
```
zneix, line 14: Why is `<unordered_map>` all the way at the top?

src/providers/twitch/TwitchBadges.cpp
```diff
@@ -7,60 +7,95 @@
 
 #include "common/NetworkRequest.hpp"
 #include "common/Outcome.hpp"
+#include "common/QLogging.hpp"
 #include "messages/Emote.hpp"
```
zneix, line 11: A bit unrelated, but I believe that Qt-specific includes can be moved under our includes in quotation marks.

src/controllers/highlights/HighlightBadge.cpp
```diff
@@ -0,0 +1,120 @@
+#include "HighlightBadge.hpp"
+#include "singletons/Resources.hpp"
```
zneix, line 2:
```suggestion
#include "HighlightBadge.hpp"

#include "singletons/Resources.hpp"
```

zneix (changes requested):

src/widgets/settingspages/HighlightingPage.cpp
pajlada: This had been accidentally added at some point, so this is just a minor unrelated fix

src/widgets/dialogs/BadgePickerDialog.cpp
pajlada: We don't have a standard of include order, so I'll mark this resolved for now.

src/providers/twitch/TwitchBadges.hpp
pajlada: We don't have a standard of include order, so I'll mark this resolved for now.

src/providers/twitch/TwitchBadges.cpp
pajlada: We don't have a standard of include order, so I'll mark this resolved for now.

CHANGELOG.md
pajlada: I have updated this, please re-review.

src/widgets/settingspages/HighlightingPage.cpp
pajlada: I have updated this, please re-review [`46295e9` (#1704)](https://github.com/Chatterino/chatterino2/pull/1704/commits/46295e9e6827739a87ac18180d6853adb6e6a0a8)

src/widgets/settingspages/HighlightingPage.cpp
pajlada: Fixed :ok_hand:

src/widgets/settingspages/HighlightingPage.cpp
zneix: That explains it well, thank you for your time :+1:

CHANGELOG.md
zneix: I guess what you wrote is fine, good stuff.

zneix (approved):

dojima:
It would be nice to be able to do this for a few of the more common badges like bots, subscribers, and Turbo users specifically.  I'm aware of the thread that mentions being able to filter subscribers into splits, but that's distinct from highlighting.

zneix:
> It would be nice to be able to do this for a few of the more common badges like bots, subscribers, and Turbo users specifically. I'm aware of the thread that mentions being able to filter subscribers into splits, but that's distinct from highlighting.

There's currently no difference between those badges in Chatterino - those all fall under "vanity" slot. An enhancement adding `MessageElementFlag`s for all of those badges would have to be made in order to support that, but there's no real reference for all the twitch badges iirc. I might be wrong though; there's an [undocumented endpoint](https://badges.twitch.tv/v1/badges/global/display?language=en) with many badges, but I wouldn't trust it as it's very old and might be some stuff missing.

> edit: Also, there's a lot of bugginess with the way prediction badges are highlighted. Whichever badge is reordered up or down last within the 'Badges' tab ('Predicted Blue' or 'Predicted Pink') seems to override a lot of the other things configured to be highlighted. As an example, reordering 'Predicted Pink' overrode the highlight settings for both 'Predicted Blue' and a user I had configured in the 'Users' tab.

You should open a new issue for this instead of commenting on an old merged PR - notifications from those can get lost quite easily.

PR #1708, opened 2020-05-29
Fix "Always on top" setting for macOS

### LINKED ISSUE #1166: 'Always on Top' not working as it should
Hello,

On macOS Mojave I have installed the latest version of Chatterino2. Upon selecting the option 'Always on Top', I notice when clicking away the program does not float as it should. When using Chatty and selecting 'Float on Top', it works as it should and floats on every window that is being used. Any ideas?

Thank you.

dnsge:
# Description

This fixes the "Always on top" setting on macOS. Previously, it only worked on Windows. It would be great if someone could check if this implementation also works on Linux.

Closes #1166

### DIFF
--- a/src/widgets/BaseWindow.cpp
+++ b/src/widgets/BaseWindow.cpp
@@ -214,9 +214,14 @@ void BaseWindow::init()
         });
     }
 #else
-//    if (getSettings()->windowTopMost.getValue()) {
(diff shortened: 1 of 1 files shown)


pajlada:
Tested in XFCE4 and works as expected :+1: (the settings dialog gets a bit weird on toggle but I don't think that's an issue of this PR)

PR #1709, opened 2020-05-29
Added possibility to change image uploader's form body with Enviroment Variables

zneix:
Nuuls' uploader software accepts `attachment` as a name of an image form field, but some other hosts don't accept that and use other names.  
This PR makes it possible to change that value with enviroment variable and with this change I can upload images to other uploaders that use multipart form data (like my server).

Since the URL of uploader can be changed with EVs, I thought it would be cool to have this small thing changeable as well.  
This PR also adds a missing newline in docs about changeable uploader's URL.

### DIFF
--- a/docs/ENV.md
+++ b/docs/ENV.md
@@ -20,7 +20,7 @@ Arguments:
  - `%1` = Emote set ID
 
 ### CHATTERINO2_IMAGE_UPLOADER_URL
-Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.
+Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.  
 Default value: `https://i.nuuls.com/upload`
 
 Arguments:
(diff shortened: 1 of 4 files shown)


docs/ENV.md
```diff
@@ -20,7 +20,7 @@ Arguments:
  - `%1` = Emote set ID
 
 ### CHATTERINO2_IMAGE_UPLOADER_URL
-Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.
+Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.
```
Mm2PL, line 23: Minor nitpick: you seem to have left some spaces at the end of this line.

zneix: @Mm2PL I added those on purpose, so there's a visible newline between description and Default value

src/util/NuulsUploader.cpp
```diff
@@ -43,6 +43,7 @@ void uploadImageToNuuls(RawImageData imageData, ChannelPtr channel,
     const static QString contentType =
         QString("multipart/form-data; boundary=%1").arg(boundary);
     static QUrl url(Env::get().imageUploaderUrl);
+    static QString FormBody(Env::get().imageUploaderFormBody);
```
pajlada, line 46: we use `camelCase` for local variables, not `PascalCase`

zneix: gotcha, updated the variable name

PR #1712, opened 2020-05-30
Added logging for uploaded images

### LINKED ISSUE #1678: Logging uploaded images

It's easy to lose i.nuuls.com links for images uploaded through chatterino.

I would like for all links for uploaded images to be saved in a file somewhere on the user's file system.

Searching the logs for all i.nuuls.com links coming from accounts registered in the client, that has an advantage of showing the context, while the disadvantage is that to find the links all the logs files would have to be searched, which takes a lot of CPU time.

**Additional context**

This could be done by opening a CSV file in *append* mode and writing a new line whenever uploading an image.

zneix:
Closes #1678  
Added logging upload information to a file located in `{messageLogDirectory}/ImageUploader.csv`

I am not sure if the path I chose is okay, since `messageLogDirectory` contains logged messages in general, but there was no better existing path for that.  
Maybe creating another directory for misc logs can be an alternative, but I don't think this is necessary (as for now).

Logged stuff is comma-separated and in following format.
Tell me if the order and types of logged information are okay

1. Path to original file (may be empty on clipboard uploads)
2. Link to uploaded image
3. Current timestamp
4. Current channel's name

I also added uploader's responses to system messages that are being generated and send in current channel upon successful file upload - makes it possible to automatically preview images and I think it's a neat addition.

### DIFF
--- a/src/util/NuulsUploader.cpp
+++ b/src/util/NuulsUploader.cpp
@@ -95,8 +95,8 @@ void uploadImageToNuuls(RawImageData imageData, ChannelPtr channel,
             if (uploadQueue.empty())
             {
                 channel->addMessage(makeSystemMessage(
-                    QString("Your image has been uploaded to ")
-                        + result.getData() + QString(" .")));
+                    QString("Your image has been uploaded to ") +
+                    result.getData() + QString(" .")));
                 uploadMutex.unlock();
             }
             else
--- a/src/util/NuulsUploader.cpp
+++ b/src/util/NuulsUploader.cpp
@@ -105,7 +105,7 @@ void uploadImageToNuuls(RawImageData imageData, ChannelPtr channel,
             {
                 channel->addMessage(makeSystemMessage(
                     QString("Your image has been uploaded to ") +
-                    result.getData() + QString(" .")));
+                    result.getData()));
                 uploadMutex.unlock();
             }
             else
@@ -184,7 +184,7 @@ void upload(const QMimeData *source, ChannelPtr channel,
                 else
                 {
                     channel->addMessage(makeSystemMessage(
-                        QString("Cannot upload file: %1, Couldn't convert "
+                        QString("Cannot upload file: %1. Couldn't convert "
                                 "image to png.")
                             .arg(localPath)));
                     uploadMutex.unlock();
@@ -212,7 +212,7 @@ void upload(const QMimeData *source, ChannelPtr channel,
             else
             {
                 channel->addMessage(makeSystemMessage(
-                    QString("Cannot upload file: %1, not an image")
+                    QString("Cannot upload file: %1. Not an image.")
                         .arg(localPath)));
                 uploadMutex.unlock();
                 return;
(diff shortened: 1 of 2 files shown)


src/util/NuulsUploader.cpp
```diff
@@ -86,6 +92,32 @@ void uploadImageToNuuls(RawImageData imageData, ChannelPtr channel,
...
+                out << originalFilePath + QString(",")
+                    << result.getData() + QString(",")
+                    << QDateTime::currentSecsSinceEpoch()
+                    << QString(",%1\n").arg(channel->getName());
+                // image path (can be empty)
+                // image link
+                // timestamp
+                // channel name
+                csvFile.close();
+            }
```
pajlada, line 119: put this in a separate function

src/util/NuulsUploader.cpp
```diff
@@ -62,20 +64,24 @@ void uploadImageToNuuls(RawImageData imageData, ChannelPtr channel,
...
             {
                 channel->addMessage(makeSystemMessage(
-                    QString("Your image has been uploaded. %1 left. Please "
-                            "wait until all of them are uploaded. About %2 "
-                            "seconds left.")
+                    QString(
+                        "Your image has been uploaded to %1 %2 left. Please "
+                        "wait until all of them are uploaded. About %3 "
+                        "seconds left.")
+                        .arg(result.getData() + QString(""))
```
pajlada, line 84: Re-add the . after the "to %1" part

zneix: Re-added with a space between, because without a space before the dot, links are invalid.

src/util/NuulsUploader.cpp
zneix: Moved it, hope it's okay :+1:

zneix:
Please ignore the force-push, I messed up branches on my fork and pushed to wrong one on accident

zneix:
@alazymeme 
> The ImageUploader.csv file is not respective of where I specifically set my log folder to be - it still stores it in %appdata%/chatterino2

Thanks for pointing that out, fixed it in 32aa59a

<br>

> Could you please fix the extra space, before the full stop at the end of the sentence when it confirms your image has been uploaded? As this would screw up the URL, are we better off just not having it there? https://i.nuuls.com/yM-9B.png

If the link is invalid doesn't that defeat the whole purpose of having it in system message? I added link to system message to make it possible to quickly preview the link.

<br>

> The centre column doesn't make any sense to me - are we able to add some headers?

That's a timestamp and alright, I added headers to the beginning of `ImageUploader.csv`

<br>

> What is that random empty column at the start of the file?

That's the path to an image - it isn't always present, because of clipboard uploads, but if you drag'n'drop file(s) to chatterino, then the path to those files will be logged.

alazymeme:
> > If the link is invalid doesn't that defeat the whole purpose of having it in system message? I added link to system message to make it possible to quickly preview the link.
> 

My apologies, I wasn't clear. Can we remove the punctuation when there is a link there, but have it there if the link is invalid?

PR #1721, opened 2020-06-05
Display authors name on hover

### LINKED ISSUE #1720: View emotion author on hover
**What should be added?**
View emotion author on hover

**Why should it be added?**
I think it is convenient and useful.

Roger-Chavez:
closes #1720

I have updated chatterino so that it shows the authors name for both bttv and ffz on hover.

### DIFF
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -77,6 +77,10 @@ namespace {
 
                 auto id = EmoteId{jsonEmote.value("id").toString()};
                 auto name = EmoteName{jsonEmote.value("code").toString()};
+                auto author = EmoteAuthor{jsonEmote.value("user")
+                                              .toObject()
+                                              .value("name")
+                                              .toString()};
                 // emoteObject.value("imageType").toString();
 
                 auto emote = Emote({
@@ -86,7 +90,10 @@ namespace {
                         Image::fromUrl(getEmoteLinkV3(id, "2x"), 0.5),
                         Image::fromUrl(getEmoteLinkV3(id, "3x"), 0.25),
                     },
-                    Tooltip{name.string + "<br />Channel BetterTTV Emote"},
+                    Tooltip{name.string + "<br />Channel BetterTTV Emote" +
+                            ((author.string.isEmpty())
+                                 ? "<br />Channel Emote"
+                                 : "<br />By : " + author.string)},
                     Url{emoteLinkFormat.arg(id.string)},
                 });
 
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -93,8 +93,8 @@ namespace {
                     },
                     Tooltip{name.string + "<br />Channel BetterTTV Emote" +
                             ((author.string.isEmpty())
-                                 ? "<br />By : " + userName.toUtf8()
-                                 : "<br />By : " + author.string)},
+                                 ? "<br />By: " + userName.toUtf8()
+                                 : "<br />By: " + author.string)},
                     Url{emoteLinkFormat.arg(id.string)},
                 });
 
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -91,10 +91,10 @@ namespace {
                         Image::fromUrl(getEmoteLinkV3(id, "2x"), 0.5),
                         Image::fromUrl(getEmoteLinkV3(id, "3x"), 0.25),
                     },
-                    Tooltip{name.string + "<br />Channel BetterTTV Emote" +
+                    Tooltip{name.string + "<br>Channel BetterTTV Emote" +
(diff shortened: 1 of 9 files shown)


src/providers/bttv/BttvEmotes.cpp
```diff
@@ -77,6 +77,10 @@ namespace {
 
                 auto id = EmoteId{jsonEmote.value("id").toString()};
                 auto name = EmoteName{jsonEmote.value("code").toString()};
+                auto author = EmoteAuthor{jsonEmote.value("user")
+                                              .toObject()
+                                              .value("name")
+                                              .toString()};
```
pajlada, line 83: Is the Author just never part of the BTTV Emote response?
If it's never included, we probably shouldn't try to do any parsing since we don't know the scheme.

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -86,7 +90,10 @@ namespace {
                         Image::fromUrl(getEmoteLinkV3(id, "2x"), 0.5),
                         Image::fromUrl(getEmoteLinkV3(id, "3x"), 0.25),
                     },
-                    Tooltip{name.string + "<br />Channel BetterTTV Emote"},
+                    Tooltip{name.string + "<br />Channel BetterTTV Emote" +
+                            ((author.string.isEmpty())
+                                 ? "<br />Channel Emote"
+                                 : "<br />By : " + author.string)},
```
pajlada, line 96: If no author information is found, we simply shouldn't add any text at all.
Right now, all BTTV Channel Emotes have a tooltip that look like this:
```
FeelsKKonaMan
Channel BetterTTV Emote
Channel Emote
```

src/providers/bttv/BttvEmotes.cpp
Roger-Chavez: Channel Emotes and Shared emotes return different JSON objects. Shared Emotes contains the "user" object which contains the "name" of the person who uploaded the emote (what I'm referring to as the author of the emote.)

Here are what the JSON objects look like:

## Shared Emotes 
sharedEmotes

## Channel Emotes
channelEmotes

src/providers/bttv/BttvEmotes.cpp
Roger-Chavez: Should I prevent it from printing anything if 'author' is an empty string? 

Yea now that I look at it it doesn't look good.

That should only be printing for emotes that are uploaded by the user. So emotes that are within the 'channelEmotes'.

For 'sharedEmotes' it looks like this for me. 
sharedEmotesExample

Roger-Chavez: Leaving it blank also didn't sit well with me considering that everything else aside from twitch emotes displayed the user that posted the emote (AKA author). 

Since "channelEmotes" contain emotes published by the user of the current channel  (AKA author) I checked to see if the current channels username was stored in the "TwitchChannel" class. This class contains the "loadChannel" function which is used to load the channel emotes among other data. Sure enough, the "userName" is stored within that class I opted to pass it through to load channel and parseChannelEmotes.

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -86,7 +91,10 @@ namespace {
                         Image::fromUrl(getEmoteLinkV3(id, "2x"), 0.5),
                         Image::fromUrl(getEmoteLinkV3(id, "3x"), 0.25),
                     },
-                    Tooltip{name.string + "<br />Channel BetterTTV Emote"},
+                    Tooltip{name.string + "<br />Channel BetterTTV Emote" +
+                            ((author.string.isEmpty())
+                                 ? "<br />By : " + userName.toUtf8()
+                                 : "<br />By : " + author.string)},
```
zneix, line 97: @Roger-Chavez Looks like you've got an extra space here (`By :`).

Roger-Chavez:
I'm working on getting the channel name to show under twitch emotes

pajlada:
Thanks a lot for your contribution @Roger-Chavez !

If you would like to be added to the contributors list in the clients about page, you can read the instructions over here and do that in a separate PR
https://github.com/Chatterino/chatterino2/blob/master/resources/contributors.txt#L1-L6

zneix:
@Roger-Chavez are you still working on showing twitch channel names on twitch emotes?

P.S. I also wanted to note here that while BTTV emotes show username (full-lowercase), FFZ ones show display-name with capital letters and sometimes "japanese names", which I think should be changed to show username instead. Maybe that can be changed with another PR about channel names for twitch emotes.

Roger-Chavez:
@pajlada Yes I can make the names displayed for FFZ and BTTV consistent. I can also finish up the twitch emotes display. I was having some issues the scope within the get request, should I join the chatterino discord if I have any questions about that?  

Also do I need to create a new issue?

zneix:
Sure, create a new pull request/issue for twitch emotes and for any help you can join Chatterino discord (or pajladas twitch chat), ping me there and we can have a look at it in the morning.

PR #1728, opened 2020-06-09
Removed annoying help tooltip for splits

zneix:
It was unnecessarily taking too much space.

Before https://cdn.zneix.eu/EF8zEL8.png

I eventually removed it entirely after fourtf's comment.
~~After https://cdn.zneix.eu/6eRX2ze.png~~
~~~~

We were talking about it with supinic on his stream and there was a suggestion to disable / remove this completely. Having an option to "disable help / tooltips" would be nice, but there's not so many of those things currently - maybe something could be added in the future if chatterino will have more help tooltips.

### DIFF
--- a/src/widgets/splits/SplitHeader.cpp
+++ b/src/widgets/splits/SplitHeader.cpp
@@ -653,16 +653,14 @@ void SplitHeader::mousePressEvent(QMouseEvent *event)
 {
     switch (event->button())
     {
-        case Qt::LeftButton:
-        {
+        case Qt::LeftButton: {
             this->dragging_ = true;
 
             this->dragStart_ = event->pos();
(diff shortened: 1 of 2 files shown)


apa420:
I've gotten annoyed by this tooltip a lot. Right now if you click the splitheader and wait a bit the tooltip pops up. Then it disappears after a while and it doesn't follow the cursor. Even if you tab a program in front of chatterino the tooltip won't dissapear.

fourtf:
Remove tooltip instead?

zneix:
@fourtf I wasn't sure if you and pajlada would be okay with me straight up purging stuff, but after all it looks like useless feature so I certainly agree 👍

PR #1729, opened 2020-06-10
Added recent messages to UserInfoPopup

### LINKED ISSUE #1726: Add the Search window to every Usercard and auto-fill with username
**What should be added?**
Adding CTRL-F to open logs was great but what would be amazing is to have that whole search window also part of every usercard popup and auto-fill the search field with the username to get only their messages and those who mentioned them.

Nothing needs to change with the search window, just add it onto the bottom of the usercard window.

**Why should it be added?**
Clicking a user to instantly see and be able to search their logs and mentions from the usercard is 1000% faster than CTRL-F and type username.

This would add a ton of usability to the usercard itself, rather than having to further open other windows and clicking/typing. Very important when a chat is moving fast and you need to check a log and timeout/ban quickly.

0xRainy:
Code copied from https://github.com/Chatterino/chatterino2/pull/1238
Resolves #1726

# Description

Adds a window to the bottom of the UserInfoPopup that contains recent messages for that user (only the messages that are still in the chat window).

Cut the code from the multiple feature containing PR https://github.com/Chatterino/chatterino2/pull/1238, and changed the popup to a non-frameless window so that it can be resized easily and increased the default size to accommodate the new log portion.

I don't know C++ so this is pure copy+paste. If there are any issues, I can try and fix it but it would probably be done faster by somebody else. 

Built and tested on Win10 64bit and it works great.

Further improvement could be made by:
- Displaying the users messages from locally stored log files 
- Adding a search box above or below the log portion
- Including mentions from other users
- Allowing multiple UserInfoPopup windows to be open at once
- Implementing QSizeGrip and reverting to a frameless window

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,8 +55,7 @@ jobs:
       - name: Install dependencies (Windows)
         if: startsWith(matrix.os, 'windows')
         run: |
-            REM We use this source (temporarily?) until choco has updated their version of conan
-            choco install conan -y -s="https://api.bintray.com/nuget/anotherfoxguy/choco-pkg"
+            choco install conan -y
 
             refreshenv
         shell: cmd
--- a/docs/ENV.md
+++ b/docs/ENV.md
@@ -20,7 +20,7 @@ Arguments:
  - `%1` = Emote set ID
 
 ### CHATTERINO2_IMAGE_UPLOADER_URL
-Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.
+Used to change the URL that Chatterino2 uses when trying to paste an image into chat. This can be used for hosting the uploaded images yourself.  
 Default value: `https://i.nuuls.com/upload`
 
 Arguments:
@@ -29,6 +29,10 @@ Arguments:
 Notes:
  - If you want to host the images yourself. You need [Nuuls' filehost software](https://github.com/nuuls/filehost)
  - Other image hosting software is currently not supported.
+
+### CHATTERINO2_IMAGE_UPLOADER_FORM_BODY
+Used to change the name of an image form field in a request to the URL that Chatterino2 uses when trying to paste an image into chat. This can be used when your image uploading software accepts a different form field than default value.  
+Default value: `attachment`
 
 ### CHATTERINO2_TWITCH_SERVER_HOST
 String value used to change what Twitch chat server host to connect to.  
--- a/src/common/Aliases.hpp
+++ b/src/common/Aliases.hpp
@@ -34,3 +34,4 @@ QStringAlias(Url);
 QStringAlias(Tooltip);
 QStringAlias(EmoteId);
 QStringAlias(EmoteName);
+QStringAlias(EmoteAuthor);
--- a/src/common/Channel.cpp
+++ b/src/common/Channel.cpp
@@ -59,6 +59,11 @@ bool Channel::isEmpty() const
     return this->name_.isEmpty();
 }
 
+bool Channel::hasMessages() const
+{
+    return !this->messages_.empty();
+}
+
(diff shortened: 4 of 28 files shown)


0xRainy:
Don't know why the formatting is different on the commit than in my Visual Studio 2017, I will swap the include though.

https://github.com/Chatterino/chatterino2/pull/1729/checks?check_run_id=756344899#step:6:33, https://github.com/Chatterino/chatterino2/pull/1729/checks?check_run_id=756344899#step:6:42, and https://github.com/Chatterino/chatterino2/pull/1729/checks?check_run_id=756344899#step:6:10 are correct in my local files. I'll fix it if someone can tell me how.

TranRed:
Not knowing C++ and copying code from someone who also doesn't know C++. I like your style 👍 

Actually would like to see it merge, I kept my own version around for a while when I had it. It's a good addition to the usercard.

zneix:
I built chatterino from [latest commit `c2041de`](c2041de) on Debian 10 and it looks like I can't move the Popup. I am able to do that on [latest nightly `39cc421`](39cc421), so I think that recent changes broke this feature. Though, I am not sure if that's just me, so can someone confirm that?

0xRainy:
I checked the logs and I built off of that commit https://github.com/Chatterino/chatterino2/commit/39cc421f95d6e7af5f4719994b38e1190f1c5f24, and I don't have a deb install right now so I can't test it.  Looking over the code not sure what could be causing it, can you try building from https://github.com/Chatterino/chatterino2/pull/1729/commits/a665108d3f911a0b5437d2b6e46e22235c100d5e

0xRainy:
> 2\. You can press a users name in the UserInfoPopup's list of recent messages which opens a new UserInfoPopup. We should find a way to disable this functionality from within the UserInfoPopup, this is not a blocker though so don't spend too much time on it.

Would you be ok with just allowing multiple UserInfoPopup windows to be open at once instead? This is something that I wanted to do but wasn't sure how, didn't spend a lot of time looking around.

0xRainy:
Discovered an issue, with #1674 merged, when clicking on a mention from within the UserDialogPopup a new usercard is opened but is not populated with the users messages even after clicking refresh.

pajlada:
I snuck in a sneaky commit which enables you to close the UserInfoPopup using the Escape button (worked on Linux before because that's a Popup standard)

pajlada:
If you could swap the linux window flag to Dialog that'd be an acceptable solution for me, however it would also need the `setActionOnFocusLoss` flag removed (resizing a window or moving a window in Linux means it loses focus, thus deleting the UserInfoPopup)

If that's not a problem on windows, feel free to wrap it in an ifndef Q_OS_LINUX

0xRainy:
Ok so having `BaseWindow::FramelessDraggable` makes the window unable to be resized no matter what I do (QSizeGrip or flag as Dialog instead of Popup).

Using `BaseWindow::EnableCustomFrame` as well as changing from window flag Popup to Dialog should fix any dragging issues on linux and it matches the theme of the other windows. I did have to remove the `setActionOnFocusLoss` flag because it closes the window on windows as well.

You can test these changes with this branch https://github.com/0xRainy/chatterino2/tree/UserDialogWithLogsMultipleOpen and it looks like this


The other option is a frameless window that isn't resizable but works the same way https://github.com/0xRainy/chatterino2/tree/UserDialogWithLogsMultipleOpenFrameless, it looks like this


(the refresh button moving around is fixed in the branches just not in the gifs)

Aside from choosing a style, this is ready to go.

0xRainy:
> Discovered an issue, with #1674 merged, when clicking on a mention from within the UserInfoPopup a new usercard is opened but is not populated with the users messages even after clicking refresh.

This is now [fixed](https://i.imgur.com/ntz9MYz.mp4), thank you @dnsge

pajlada:
Works great for me on Arch Linux btw using i3. thanks for the contribution!

PR #1730, opened 2020-06-11
Fix popout player url

alazymeme:
# Description

It appears that twitch has changed the popout player url and that the previous url does not work. I've updated chatterino2 to reflect that :)

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -85,8 +85,7 @@ TwitchChannel::TwitchChannel(const QString &name,
     , ChannelChatters(*static_cast<Channel *>(this))
     , subscriptionUrl_("https://www.twitch.tv/subs/" + name)
     , channelUrl_("https://twitch.tv/" + name)
-    , popoutPlayerUrl_("https://player.twitch.tv/?channel=" + name +
-                       "&parent=twitch.tv")
+    , popoutPlayerUrl_("https://player.twitch.tv/?parent=twitch.tv&channel=" + name)
     , globalTwitchBadges_(globalTwitchBadges)
     , globalBttv_(bttv)
     , globalFfz_(ffz)
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -85,7 +85,7 @@ TwitchChannel::TwitchChannel(const QString &name,
     , ChannelChatters(*static_cast<Channel *>(this))
     , subscriptionUrl_("https://www.twitch.tv/subs/" + name)
     , channelUrl_("https://twitch.tv/" + name)
-    , popoutPlayerUrl_("https://player.twitch.tv/?channel=" + name)
+    , popoutPlayerUrl_("https://player.twitch.tv/?channel=" + name + "&parent=twitch.tv")
     , globalTwitchBadges_(globalTwitchBadges)
     , globalBttv_(bttv)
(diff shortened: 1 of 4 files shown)


zneix:
You seem to have rushed this PR a bit because link to popout player is also in `Toasts.cpp`, `Split.cpp` and `SplitHeader.cpp`.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -85,7 +85,8 @@ TwitchChannel::TwitchChannel(const QString &name,
     , ChannelChatters(*static_cast<Channel *>(this))
     , subscriptionUrl_("https://www.twitch.tv/subs/" + name)
     , channelUrl_("https://twitch.tv/" + name)
-    , popoutPlayerUrl_("https://player.twitch.tv/?channel=" + name)
+    , popoutPlayerUrl_("https://player.twitch.tv/?channel=" + name +
+                       "&parent=twitch.tv")
```
zneix, line 89: @alazymeme I think that format `?parent=twitch.tv&channel=" + name` will be a bit better, because it's in a one string only.

dnsge:
Fixes #1731

TranRed:
I just wanted to do the same, nice to see there is already a PR.

I personally think the parent should never be twitch.tv here, but the embedding site/calling application. So I would suggest chatterino or chatterino.com. Probably only chatterino, since it is not an embed on the web page. Based on: https://dev.twitch.tv/docs/embed/video-and-clips/

@pajlada any opinions/twitch insides? The docu about the parent key isn't very detailed on first glance. Maybe you know more

alazymeme:
> I just wanted to do the same, nice to see there is already a PR.
> 
> I personally think the parent should never be twitch.tv here, but the embedding site/calling application. So I would suggest chatterino or chatterino.com. Probably only chatterino, since it is not an embed on the web page. Based on: https://dev.twitch.tv/docs/embed/video-and-clips/
> 
> @pajlada any opinions/twitch insides? The docu about the parent key isn't very detailed on first glance. Maybe you know more

`Domain that will be embedding Twitch. You must have one parent key for each domain your site is hosted or embedded on.` - this would indicate that the way it currently is; is actually correct. Since chatterino itself is not embedding the player.

src/providers/twitch/TwitchChannel.cpp
alazymeme: > @alazymeme I think that format `?parent=twitch.tv&channel=" + name` will be a bit better, because it's in a one string only.

According to twitch's dev docs, the correct usage is `?channel=channel&parent=sitethatwillembedthecontent`. However twitch re-organizes them anyway. So I'm not sure if it NEEDS to be one way or the other...

TranRed:
> 
> 
> > I just wanted to do the same, nice to see there is already a PR.
> > I personally think the parent should never be twitch.tv here, but the embedding site/calling application. So I would suggest chatterino or chatterino.com. Probably only chatterino, since it is not an embed on the web page. Based on: https://dev.twitch.tv/docs/embed/video-and-clips/
> > @pajlada any opinions/twitch insides? The docu about the parent key isn't very detailed on first glance. Maybe you know more
> 
> `Domain that will be embedding Twitch. You must have one parent key for each domain your site is hosted or embedded on.` - this would indicate that the way it currently is; is actually correct. Since chatterino itself is not embedding the player.

makes sense. I saw aiden also used twitch.tv so it should be fine, if the man himself is doing this as best solution :)

PR #1733, opened 2020-06-11
Adds parent=twitch.tv

aidenwallis:
This is needed to make it compatible with the new twitch embed changes: https://discuss.dev.twitch.tv/t/twitch-embedded-player-updates-in-2020/23956


aidenwallis:
failed to miss it was answered by another, closed

Issue #43, opened 2017-04-27
Commandline options


Bur0k:
suggested:
-join CHANNEL1,CHANNEL2,....

pajlada:
Note to self for command line arguments: check out boost program options http://www.boost.org/doc/libs/1_64_0/doc/html/program_options/tutorial.html#idp435720656

Icarus77:
Does removing the "wontfix" tag mean you'll add this eventually? Great software, but it not working with things like SL Twitch GUI kind of sucks.

azizLIGHT:
Any progress on this front?

jp1995:
Yearly Jeping. Would be awesome to open specific channels for my lidl script

zneix:
If I read correctly, this should have more priority than settings, right?  
For example, running `chatterino --channels=pajlada,zneix,mm2pl` would only open those 3 channels, regardless if I have 0, 10 or 100 channels in settings - and preferrably all of them would open in the new tabs, right?

omrisim210:
Boost is mentioned in a comment from a few years ago, and I wonder why not just use `QCommandLineParser` since Chatterino already uses Qt?

fourtf:
There's a PR for this and other issues so I'll close

PR #1739, opened 2020-06-14
Made --version argument output consistent with AboutPage

zneix:
When using `chatterino --version` it now also outputs commit (and date of build if on nightly), just like in Settings -> About:

I think it's a useful addition, especially when you work with nightly builds and want to quickly identify version without having to launch the app and going to settings.


Mm2PL:
related: #43

Issue #1722, opened 2020-06-05
Move wiki pages to markdown files.


Mm2PL:
**What should be changed?**
I would like for the Chatterino 2 [wiki](https://github.com/chatterino/chatterino2/wiki) to be moved to markdown files.

**Why should it be changed?**
I would like this change because contributing to wiki pages is hard, while files are easy to PR.

PR #1741, opened 2020-06-15
More image uploader services and json logging

zneix:
I wanted to extend #1709 and #1712 by adding more config options for image uploader and changing CSV logs to JSON.

1. Aside from changing form field, I added an option to add extra header(s) defined by user, so it is possible to use hosts like imgur / s-ul, which need an authentication header.  
I decided that it will be the best to parse headers as a string. I know this may look wrong, but usually it is only one auth header and redesigning Settings page for handling names and values separetly is not worth it imo - it would also look too complicated.

2. I added handling JSON responses from those servers, so it is possible to tell chatterino where image link is in the JSON object.

3. Some services, especially imgur, support "deletion links", so I also added an option to specify those in settings.

To achieve all of above, I removed enviroment variables responsible for changing uploader server's URL and form field name (#1709) and added 5 options to `Settings -> External Tools -> Image Uploader` section. I am not sure if that's a good place for those settings, but I took under consideration that Chatterino uses here an external host for uploading images, thus that is a 3rd-party. That is how it looks now:

There is a link to a guide, which is temporarily [this gist](https://gist.github.com/zneix/14d828c145f755394e5ab0301c314c2a) with help and configuration examples for less advanced users. I first wanted to ask a Developer to add it as a wiki page, but I think that once #1722 is done, I can commit it as a markdown file to docs in this repo.

Note: For testing purposes, I created a new imgur application to obtain Client-ID, which is used in calls to imgur API. It is only for anonymous uploads. Devs can create a new anonymous app [here](https://api.imgur.com/oauth2/addclient) if they wish to do so.

---

That change is mainly focused on having an ability to delete uploaded content, so you will not "permaleak" a photo of your face or any other sensitive data by an accident after dragging it onto Chatterino by a mistake.

---

I also changed the logging method from CSV (added in #1712) to JSON (as fourtf and pajlada suggested) - it's now one array with objects that contain information about successful uploads.  
Fourtf suggested that I should use [QSaveFile](https://doc.qt.io/qt-5/qsavefile.html), but since it only only supports WriteOnly mode, I had to use [QFile](https://doc.qt.io/qt-5/qfile.html) to read existing data. If that's not efficient, I can use QFile for both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
 # Changelog
 
 ## Unversioned
-- Major: We now support image thumbnails coming from the link resolver. This feature is off by default and can be enabled in the settings with the "Show link thumbnail" setting. This feature also requires the "Show link info when hovering" setting to be enabled (#1664)
-- Major: Added image upload functionality to i.nuuls.com. This works by dragging and dropping an image into a split, or pasting an image into the text edit field. (#1332)
-- Minor: You can now open the Twitch User Card by middle-mouse clicking a username. (#1669)
-- Minor: Emotes in the emote popup are now sorted in the same order as the tab completion (#1549)
-- Minor: Removed "Online Logs" functionality as services are shut down (#1640)
-- Bugfix: Fix preview on hover not working when Animated emotes options was disabled (#1546)
-- Bugfix: FFZ custom mod badges no longer scale with the emote scale options (#1602)
-- Bugfix: MacOS updater looked for non-existing fields, causing it to always fail the update check (#1642)
+- Major: We now support image thumbnails coming from the link resolver. This feature is off by default and can be enabled in the settings with the "Show link thumbnail" setting. This feature also requires the "Show link info when hovering" setting to be enabled ([#1664](https://github.com/Chatterino/chatterino2/pull/1664))
+- Major: Added image upload functionality to i.nuuls.com with an ability to change upload destination. This works by dragging and dropping an image into a split, or pasting an image into the text edit field. ([#1332](https://github.com/Chatterino/chatterino2/pull/1332), [#1741](https://github.com/Chatterino/chatterino2/pull/1741))
+- Minor: You can now open the Twitch User Card by middle-mouse clicking a username. ([#1669](https://github.com/Chatterino/chatterino2/pull/1669))
+- Minor: User Popup now also includes recent user messages ([#1729](https://github.com/Chatterino/chatterino2/pull/1729))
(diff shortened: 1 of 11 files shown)


zneix:
Also, since image uploader will now support more than just i.nuuls.com, should it be renamed from `NuulsUploader.cpp` to just `ImageUploader.cpp` ? Just to keep consistency.

pajlada:
also, please move the documentation into a file in the repo (docs/IMAGEUPLOADER.md maybe).
The first example in the documentation should also be i.nuuls.com (the default)

pajlada:
Request my review once that's done and I'll try it out a bit more in-depth, in general the "design choices" are :+1:

zneix:
> The ImageUploader.json file is not created if it doesn't already exist (at least on Linux)

I totally missed that and left insufficient OpenModeFlag by a mistake.  


Moved docs to this repo under the suggested name and left a short comment for default image uploader.

CHANGELOG.md
```diff
@@ -1,13 +1,15 @@
...
+- Minor: You can now open the Twitch User Card by middle-mouse clicking a username. ([#1669](https://github.com/Chatterino/chatterino2/pull/1669))
+- Minor: User Popup now also includes recent user messages ([#1729](https://github.com/Chatterino/chatterino2/pull/1729))
+- Minor: BetterTTV / FrankerFaceZ emote tooltips now also have emote authors' name ([#1721](https://github.com/Chatterino/chatterino2/pull/1721))
+- Minor: Emotes in the emote popup are now sorted in the same order as the tab completion ([#1549](https://github.com/Chatterino/chatterino2/issues/1549))
+- Minor: Removed "Online Logs" functionality as services are shut down ([#1640](https://github.com/Chatterino/chatterino2/issues/1640))
+- Bugfix: Fix preview on hover not working when Animated emotes options was disabled ([#1546](https://github.com/Chatterino/chatterino2/issues/1546))
+- Bugfix: FFZ custom mod badges no longer scale with the emote scale options ([#1602](https://github.com/Chatterino/chatterino2/pull/1602))
+- Bugfix: MacOS updater looked for non-existing fields, causing it to always fail the update check ([#1642](https://github.com/Chatterino/chatterino2/pull/1642))
```
pajlada, line 13: Please undo the link additions here, I rely on GitHub being smart enough to autolink these at some point, and not linking them makes them cleaner to read in a text editor

zneix: Alright, I reverted those changes.

PR #1748, opened 2020-06-21
Advanced channel filters

### LINKED ISSUE #518: Filterable /mentions tab
When you open  /mentions it would be neat if you could set chatterino set to just post the mentionings of a specific channel. (The channel has to be open) 
It would be neat if you are mod/caster and could filter the mentions to his channel/the channel he mods/watches.
suggest mention

### LINKED ISSUE #1632: Add filters to channels.

I would like for each channel to have filters, a search string that is matched to all messages. It could be specified like `channel_name filters filters filters...`. The channel change popup could be changed to make the filters a separate input box.

not doing it

**Additional context**

It would be pretty easy to reuse the existing search predicate system.

### LINKED ISSUE #1643: Filter Message by username color
Some chats contain a lot of very new chatters which are mostly just spam and unintersting.

It would be nice to be able to filter messages by attributes of the user that posted them. 
My specific usecase is filtering users without a color. (Show up grey in chatterino by default)

I think on its own the feature is decent, but it would be way better to somehow make multiple conditions etc for a message to appear. So that maybe if a user with no color types a new message but he is actually a subscriber then his message will show up.

This doesn't have to be only for colors, it could be generally filtering of messages by attributes of the user like sub status, badges etc. 

I have had several people telling me they would love that feature, that's why I opened an issue.

dnsge:
# Description
Adds custom channel filters complete with their own mini-language. Filters can be created in settings, and applied by clicking the three dots to open the Split menu and selecting "Set filters".

I tried my hand at creating an icon for the filters tab in settings, but I'd like feedback on its design.

Example filters:
- `message.content contains "hello"`
- `message.length <= 40 || author.subscribed`
- `(message.length <= 10 || message.content contains "look at this") && !(message.content contains "LULW")`

Valid but useless filters:
- `2 + 3 == 5`
- `10 / 2 == 3`
- `20 % 3 == 2`

**Available per-message variables:**
| Variable | Type | Description |
| - | - | - |
| **Author** | | User who sent the message|
| `author.badges` | List<String> | List of author's badges |
| `author.color` | Color | Color code of author, or none |
| `author.name` | String | Display name of author |
| `author.subbed` | Bool | Whether author is subscribed |
| `author.sub_length` | Int | How long author has been subscribed (or zero) |
| **Channel** | | The channel where the message was sent |
| `channel.name` | String | Channel name |
| `channel.watching` | Bool | Whether the channel is being watched (requires Chatterino extension) |
| **Flags** | | Message-specific flags |
| `flags.highlighted` | Bool | Whether the message is highlighted |
| `flags.points_redeemed` | Bool | Whether the message was redeemed through channel points |
| `flags.sub_message` | Bool | Whether the message is a sub/resub/gift message |
| `flags.system_message` | Bool | Whether the message is a system message |
| `flags.whisper` | Bool | Whether the message is a whisper |
| **Message** | | Actual message sent |
| `message.content` | String | Message content |
| `message.length` | Int | Message length |

**Available operators:**
| Operator | Description |
| - | - |
| `&&` | And |
| `\|\|` | Or |
| `==` | Equals |
| `!=` | Not equals |
| `%` | Modulus |
| `<` | Less than |
| `<=` | Less than or equal to |
| `>` | Greater than |
| `>=` | Greater than or equal to |
| `!` | Not |
| `contains` | String, List, or Map contains |
| `startswith` | String or List starts with text or string, respectively |
| `endswith` | String or List ends with text or string, respectively |
| `+` | Add |
| `-` | Subtract |
| `*` | Multiply |
| `/` | Divide (integer) |

**Literals:**
| Name | Example |
| - | - |
| Int | `123`, `5` |
| String | `"Hello there"`, `"Escaped \" quote"` |

**Notes about the order of operations:**
- Math operations are evaluated from left to right, not by MDAS. `2 + 3 * 4` yields `20`, not `14`.
- `a && b || c && d` is evaluated as `(a && b) || (c && d)`
- `a || b && c || d` is evaluated as `a || (b && c) || d`

#### Still working on:
- [x] Persist assigned filters after relaunching Chatterino
- [x] User feedback for parsing of filter
- [x] Verifying that all types work correctly
- [x] Fixing order of operations 
- [x] Documentation 
- [x] Properly filtering messages from splits
- [ ] Fixing edge cases

Closes #1643 
Closes #1632 
Closes #518

This PR should probably be squashed into one commit


fourtf:
Why does this introduce a language for defining filters and not a GUI? Is this accessible to users that are not coders?

dnsge:
I figured that a language would be simpler to implement compared to creating a GUI, mostly because I'm new to Qt and I needed a way to persist the filters anyways. I did write up information for creating filters [here](https://github.com/Chatterino/chatterino2/blob/c1c39d386f2a630c845007e92df35ae747dce090/docs/Filters.md), but I realize that it could probably use some step-by-step instructions for creating your own filters for those unfamiliar with programming. 

I'll improve the documentation somewhat and I can try my hand at creating a GUI editor that wraps the language (i.e. opening the editor, creating the filter, and saving it as text). Maybe I'll replace operators like `&&` and `||` with `and` and `or` to make it simpler.

fourtf:
I think having a gui for simple filtres (only and, no nesting) and leaving this for more complex queries might be fine

dnsge:
Here is a video demoing the simple channel filter editor added in b7afec8: https://i.nuuls.com/qVd4u.mp4

Mm2PL:
How complete is this PR? I would like to test it out on my main setup if it's stable :)

dnsge:
@Mm2PL It's 99% complete, I've been using it on my [own branch](https://github.com/dnsge/chatterino2/tree/latest-features) for the past month or so. These last couple of commits have fixed up the few bugs that existed with my implementation. If you end up using it and notice any issues, you can leave a comment here and I'll try to fix it.

onbjerg:
What is preventing this from being merged?

Mm2PL:
From trying to integrate filters into my Chatterino i found a couple things:
  - I found no way to do a starts with on a string, which would be nice for detecting bot commands and filtering those
  - Splits that have filters look the same as those without. Maybe a filter icon like the moderation one could work, or just changing the split title.

Otherwise it works great for me BroBalt

dnsge:
Just added `startswith`/`endswith` functionality in 23b8863. 

Split titles will now have a ` - filtered` at the end when filters are enabled:
Screen Shot 2020-09-05 at 12 32 54 AM

onbjerg:
Would it be possible to display the applied filter names somewhere?

leon-richardt:
Interesting catch over at https://github.com/Chatterino/chatterino2/issues/1927#issuecomment-687923084. Not sure how easy it is to fix this with the Channel model Chatterino uses (although I haven't looked into the code of this implementation too much).

dnsge:
Yeah I saw that, I should be able to copy attributes from one channel to the proxy channel. I'll take a look into that.

Mm2PL:
Your PR seems to cause logging of the same message multiple times. Could you look into that?

dnsge:
@Mm2PL Thanks for catching that. Fixed in be59384.

zneix:
Just wanted to mention that all documentation has been already moved to [chatterino wiki](https://github.com/chatterino/wiki), so you'll have to commit documentation to that repo.

src/widgets/helper/ChannelView.cpp
```diff
@@ -603,29 +666,68 @@ void ChannelView::setChannel(ChannelPtr channel)
...
     {
         this->connections_.push_back(tc->liveStatusChanged.connect([this]() {
             this->liveStatusChanged.invoke();  //
         }));
     }
 }
 
+void ChannelView::setFilters(const QList<QUuid> &ids)
+{
+    this->channelFilters_ = new FilterSet(ids);
```
pajlada, line 698: use if non-smart pointer again - revisit this

src/widgets/helper/ChannelView.hpp
```diff
@@ -178,11 +183,20 @@ class ChannelView final : public BaseWidget
     LimitedQueueSnapshot<MessageLayoutPtr> snapshot_;
 
     ChannelPtr channel_;
+    ChannelPtr underlyingChannel_;
     ChannelPtr sourceChannel_;
 
     Scrollbar *scrollBar_;
     EffectLabel *goToBottom_;
 
+    FilterSet *channelFilters_ = nullptr;
```
pajlada, line 192: smart pointer here please

src/widgets/settingspages/FiltersPage.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include "controllers/filters/FilterModel.hpp"
+#include "singletons/Settings.hpp"
+#include "util/LayoutCreator.hpp"
+#include "widgets/dialogs/ChannelFilterEditorDialog.hpp"
+#include "widgets/helper/EditableModelView.hpp"
+
+#include <QTableView>
+
+#define FILTERS_DOCUMENTATION \
```
pajlada, line 11: the documentation is good, but should before merging, move to the wiki instead. https://github.com/Chatterino/wiki

pajlada (changes requested):
Overall really cool PR - would be a heavy user of this myself

src/controllers/filters/parser/Types.cpp
```diff
@@ -0,0 +1,378 @@
...
+        case STARTS_WITH:
+            opText = "startswith";
+            break;
+        case ENDS_WITH:
+            opText = "endswith";
+            break;
+        default:
+            opText = "";
+            break;
+    }
```
23rd, line 328:
```suggestion
    const auto opText = [&]() -> QString {
        switch (this->op_)
        {
            case AND:
                return "&&";
            case OR:
                return "||";
            case PLUS:
                return "+";
            case MINUS:
                return "-";
            case MULTIPLY:
                return "*";
            case DIVIDE:
                return "/";
            case MOD:
                return "%";
            case EQ:
                return "==";
            case NEQ:
                return "!=";
            case LT:
                return "<";
            case GT:
                return ">";
            case LTE:
                return "<=";
            case GTE:
                return ">=";
            case CONTAINS:
                return "contains";
            case STARTS_WITH:
                return "startswith";
            case ENDS_WITH:
                return "endswith";
            default:
                return QString();
        }
    }();
```

src/common/WindowDescriptors.cpp
```diff
@@ -68,6 +68,21 @@ namespace {
         return descriptor;
     }
 
+    const QList<QUuid> loadFilters(QJsonValue val)
+    {
+        QList<QUuid> filterIds;
+
+        if (!val.isUndefined())
+        {
+            for (const auto &id : val.toArray())
```
23rd, line 77:
```suggestion
            const auto array = val.toArray();
            filterIds.reserve(array.size());
            for (const auto &id : array)
```

src/controllers/filters/FilterRecord.hpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    FilterRecord(const QString &name, const QString &filter, const QUuid &id)
+        : name_(name)
+        , filter_(filter)
+        , id_(id)
+    {
+        this->parser_ = new filterparser::FilterParser(filter);
+        this->valid_ = this->parser_->valid();
+    }
```
23rd, line 42:
```suggestion
    FilterRecord(const QString &name, const QString &filter, const QUuid &id)
        : name_(name)
        , filter_(filter)
        , id_(id)
        , parser_(std::make_shared<filterparser::FilterParser>(filter))
        , valid_(this->parser_->valid())
    {
    }

    FilterRecord(const QString &name, const QString &filter)
        : FilterRecord(name, filter, QUuid::createUuid())
    {
    }
```

src/controllers/filters/FilterRecord.hpp
```diff
@@ -0,0 +1,122 @@
...
+        return this->parser_->execute(context);
+    }
+
+private:
+    QString name_;
+    QString filter_;
+    QUuid id_;
+    bool valid_;
+
+    filterparser::FilterParser *parser_;
```
23rd, line 80:
```suggestion

    std::shared_ptr<filterparser::FilterParser> parser_;
    bool valid_;
```

src/controllers/filters/parser/FilterParser.cpp
```diff
@@ -0,0 +1,298 @@
...
+     * flags.whisper
+     *
+     * message.content
+     * message.length
+     *
+     */
+
+    using MessageFlag = chatterino::MessageFlag;
+
+    QStringList badges;
```
23rd, line 39:
```suggestion
    QStringList badges;
    badges.reserve(m->badges.size());
```

src/controllers/filters/parser/FilterParser.cpp
```diff
@@ -0,0 +1,298 @@
...
+                    watchingChannel->getName().compare(
+                        m->channelName, Qt::CaseInsensitive) == 0;
+
+    bool subscribed = badges.contains("subscriber");
+    int subLength = (subscribed && m->badgeInfos.count("subscriber") != 0)
+                        ? m->badgeInfos.at("subscriber").toInt()
+                        : 0;
+
+    return {
+        {"author.badges", badges},
```
23rd, line 55:
```suggestion
        {"author.badges", std::move(badges)},
```

src/controllers/filters/parser/FilterParser.cpp
```diff
@@ -0,0 +1,298 @@
...
+    {
+        if (this->tokenizer_.hasNext())
+        {
+            this->errorLog(QString("Missing closing parentheses: got %1")
+                               .arg(this->tokenizer_.preview()));
+        }
+        else
+        {
+            this->errorLog("Missing closing parentheses at end of statement");
+        }
```
23rd, line 170:
```suggestion
        const auto message =
            this->tokenizer_.hasNext()
                ? QString("Missing closing parentheses: got %1")
                      .arg(this->tokenizer_.preview())
                : "Missing closing parentheses at end of statement";
        this->errorLog(message);
```

src/controllers/filters/parser/Types.cpp
```diff
@@ -0,0 +1,378 @@
...
+    QString opText;
+    switch (this->op_)
+    {
+        case NOT:
+            opText = "!";
+            break;
+        default:
+            opText = "";
+            break;
+    }
```
23rd, line 374:
```suggestion
    const auto opText = [&]() -> QString {
        switch (this->op_)
        {
            case NOT:
                return "!";
            default:
                return QString();
        }
    }();
```

src/widgets/dialogs/ChannelFilterEditorDialog.cpp
```diff
@@ -0,0 +1,240 @@
...
+                {
+                    auto validator = new QIntValidator;
+                    validator->setBottom(0);
+                    this->valueInput_->setValidator(validator);
+                }
+                else
+                {
+                    this->valueInput_->setValidator(nullptr);
+                }
+            }
```
23rd, line 128:
```suggestion
            const auto isNumber = (index == 1);
            const auto isVariable = (index == 2);
            this->valueInput_->setVisible(!isVariable);
            this->varCombo_->setVisible(isVariable);

            this->valueInput_->setValidator(
                isNumber ? (new QIntValidator(0, INT_MAX)) : nullptr);
```

src/widgets/dialogs/ChannelFilterEditorDialog.cpp
```diff
@@ -0,0 +1,240 @@
...
+    this->layout_->addLayout(this->left_->layout());
+    this->layout_->addWidget(this->opCombo_);
+    this->layout_->addLayout(this->right_->layout());
+    this->layout_->setContentsMargins(5, 5, 5, 5);
+
+    QObject::connect(  //
+        this->opCombo_, QOverload<int>::of(&QComboBox::currentIndexChanged),
+        [this](int index) {
+            // disable if set to "(nothing)"
+            this->right_->setEnabled(realBinaryOps.at(index) != "");
```
23rd, line 203:
```suggestion
            this->right_->setEnabled(!realBinaryOps.at(index).isEmpty());
```

src/widgets/dialogs/ChannelFilterEditorDialog.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+QLayout *ChannelFilterEditorDialog::BinaryOperationSpecifier::layout() const
+{
+    return this->layout_;
+}
+
+QString ChannelFilterEditorDialog::BinaryOperationSpecifier::expressionText()
+{
+    QString opText = realBinaryOps.at(this->opCombo_->currentIndex());
+    if (opText == "")
```
23rd, line 229:
```suggestion
    if (opText.isEmpty())
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -556,39 +557,101 @@ ChannelPtr ChannelView::channel()
...
+        underlyingChannel->messagesAddedAtStart.connect(
+            [this](std::vector<MessagePtr> &messages) {
+                std::vector<MessagePtr> filtered;
+                std::copy_if(  //
+                    messages.begin(), messages.end(),
+                    std::back_inserter(filtered), [this](MessagePtr msg) {
+                        return this->shouldIncludeMessage(msg);
+                    });
+
+                if (filtered.size() > 0)
```
23rd, line 613:
```suggestion
                if (!filtered.empty())
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -603,29 +666,68 @@ void ChannelView::setChannel(ChannelPtr channel)
...
 }
 
+void ChannelView::setFilters(const QList<QUuid> &ids)
+{
+    this->channelFilters_ = new FilterSet(ids);
+}
+
+const QList<QUuid> ChannelView::getFilterIds() const
+{
+    if (this->channelFilters_ == nullptr)
```
23rd, line 703:
```suggestion
    if (!this->channelFilters_)
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -603,29 +666,68 @@ void ChannelView::setChannel(ChannelPtr channel)
...
+}
+
+FilterSet *ChannelView::getFilterSet() const
+{
+    return this->channelFilters_;
+}
+
+bool ChannelView::shouldIncludeMessage(const MessagePtr &m) const
+{
+    if (this->channelFilters_ != nullptr)
```
23rd, line 718:
```suggestion
    if (this->channelFilters_)
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -731,15 +834,17 @@ void ChannelView::messageAddedAtStart(std::vector<MessagePtr> &messages)
...
-    for (size_t i = 0; i < messages.size(); i++)
+    if (this->showScrollbarHighlights())
     {
-        highlights.push_back(messages.at(i)->getScrollBarHighlight());
-    }
+        std::vector<ScrollbarHighlight> highlights;
+        highlights.reserve(messages.size());
+        for (size_t i = 0; i < messages.size(); i++)
+        {
+            highlights.push_back(messages.at(i)->getScrollBarHighlight());
```
23rd, line 843:
```suggestion
        for (const auto &message : messages)
        {
            highlights.push_back(message->getScrollBarHighlight());
```

src/widgets/helper/SearchPopup.cpp
```diff
@@ -40,6 +41,9 @@ ChannelPtr SearchPopup::filter(const QString &text, const QString &channelName,
             }
         }
 
+        if (accept && filterSet != nullptr)
```
23rd, line 44:
```suggestion
        if (accept && filterSet)
```

src/controllers/filters/parser/FilterParser.cpp
```diff
@@ -0,0 +1,298 @@
...
+{
+    return this->builtExpression_->execute(context).toBool();
+}
+
+bool FilterParser::valid() const
+{
+    return this->valid_;
+}
+
+Expression *FilterParser::parseExpression(bool top)
```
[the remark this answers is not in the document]
dnsge: I never really considered adding any more dependencies, though it certainly would have made this feature simpler.

src/controllers/filters/FilterRecord.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include <memory>
+
+namespace chatterino {
+
+class FilterRecord
+{
+public:
+    bool operator==(const FilterRecord &other) const
+    {
+        return std::tie(this->name_, this->filter_, this->id_) ==
```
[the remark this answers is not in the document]
dnsge: What's the correct way to do this? I adopted the code from here, which does the same thing as far as I can tell:

https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/controllers/highlights/HighlightPhrase.cpp#L17-L25

alazymeme:
Quick question: Is it as simple as adding a filter with `author.subbed` and the author of the message would have to be subscribed? If so, it isn't working for me.

image

Secondly, using `author.subscribed` completely crashes chatterino for me.

dnsge:
@alazymeme 
`author.subbed` is valid and works for me:
Screen Shot 2020-09-29 at 2 14 17 PM
Filters are only applied to new messages, not messages that are already present in the channel. Perhaps that's what happened?

As for creating and applying a filter `author.subscribed`, my Chatterino doesn't crash. Maybe it's a platform-specific bug?
Screen Shot 2020-09-29 at 2 17 25 PM

src/controllers/filters/FilterRecord.hpp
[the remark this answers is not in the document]
pajlada: It's only the `this->filter` variable that's the issue - everything else will compare fine.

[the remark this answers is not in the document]
pajlada: Think of it like this:
```c++
std::unique_ptr<filterparser::FilterParser> filterA;
std::unique_ptr<filterparser::FilterParser> filterB;

if (filterA == filterB) {
  return true;
}

return false;
```

No matter the value of filterA and filterB, the equailty check will be made on the pointers themselves, not the actual value stored within.

See https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_cmp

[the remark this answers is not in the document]
dnsge: `this->filter_` is a `QString` so I think the current code is fine. `this->parser_` is the `filterparser::FilterParser` pointer.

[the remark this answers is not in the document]
pajlada: Oops! You're right :+1: My bad

src/controllers/filters/parser/Tokenizer.hpp
```diff
@@ -0,0 +1,74 @@
...
+    "message.length"          // Int
+};
+
+static const QStringList friendlyValidIdentifiers = {
+    "author badges",        "author color",       "author name",
+    "author has no color?", "author subscribed?", "author sub length",
+    "channel name",         "/watching channel?", "highlighted?",
+    "redeemed points?",     "sub/resub message?", "whisper message?",
+    "message text",         "message length",
+};
```
pajlada, line 31: these two QStringLists are not the same length. let's try to prevent that from happening again by building a QMap<QString, QString> where the key is the identifier and the value is the identifier friendly name, so you can just do validIdentifiers = map.keys(); friendlyValidIdentifiers = map.values();

it's "system message" missing btw

pajlada (changes requested):

src/controllers/filters/parser/Tokenizer.hpp
dnsge: fixed in 1da48d1 👍

pajlada (approved):

pajlada:
Will be merging this in as is once #2081 has been merged into master. Thanks a lot for your contribution!

gempir:
This is an amazing feature that makes Chatterino stand out from a lot of other chat clients. Thank you!

PR #1759, opened 2020-06-25
Added higher tier information to sub badge tooltip

zneix:
With recent Twitch's changes for higher tier subs the sub badge tooltip was also updated in webchat and I though it would be a nice idea to have that behaviour in chatterino:

Webchat https://cdn.zneix.eu/XmR2hR6.png
webchat

Chatterino https://cdn.zneix.eu/6z3X5yX.png
chatterino

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1325,6 +1325,9 @@ void TwitchMessageBuilder::appendTwitchBadges()
             auto badgeInfoIt = badgeInfos.find(badge.key_);
             if (badgeInfoIt != badgeInfos.end())
             {
+                // badge.value_ is 4 chars long if user is subbed on higher tier
+                // (tier + amount of months with leading zero if less than 100)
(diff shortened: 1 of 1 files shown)


Mm2PL:
In my opinion`Tier %d %d-Month Subscriber (%d months)` would sound better 🙂 What do you think?

zneix:
I think having just "month-value" in brackets is nice, but what you suggested may look a bit confusing (imo) due to tier's number and badge descriptions' number being next to each other:
asd

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1325,8 +1325,14 @@ void TwitchMessageBuilder::appendTwitchBadges()
             auto badgeInfoIt = badgeInfos.find(badge.key_);
             if (badgeInfoIt != badgeInfos.end())
             {
+                const auto &subTier =
```
pajlada, line 1328: Could you add a comment describing how the sub tier is calculated with some examples?

zneix: Added in b74c7de. Hope it's clear enough

PR #1760, opened 2020-06-25
Move wiki

### LINKED ISSUE #1722: Move wiki pages to markdown files.

Mm2PL:
# Description
Move the wiki to markdown files.
Before merging: please put links to the files on the wiki pages they are replacing.
Closes #1722


zneix:
I think you should also edit this here https://github.com/Chatterino/chatterino2/blob/8e202c37556ae969718d9be6e736e700be647bb1/src/widgets/helper/EditableModelView.cpp#L99-L100

Mm2PL:
In my opinion fourtf should change that redirect to the file, since changing this in a new version would break the link in older ones. If the redirect is changed the link will still work in older versions.

@fourtf I summon you to this conversation ;)

pajlada:
@Mm2PL Just change the url :+1: 
Wiki shouldn't be purged either, it should just maybe link to the markdown version at the top in big text

PR #1763, opened 2020-06-27
Document FreeBSD build instructions

herrhotzenplotz:
Build succeeded and worked on FreeBSD 12.1 (see underneath). I felt like documenting the dependencies from the ports collection.

This has been tested under the following conditions:

```
$ uname -apKU
FreeBSD triton 12.1-RELEASE FreeBSD 12.1-RELEASE r354233 GENERIC  amd64 amd64 1201000 1201000
```

It may also work on older versions and (should definitely) on FreeBSD 13.0-CURRENT. Other processor architectures might support chatterino as well depending on the tier (and thus ports availability). See [https://www.freebsd.org/platforms/](https://www.freebsd.org/platforms/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Current releases are available at [https://chatterino.com](https://chatterino.co
 ## Nightly build
 You can download the latest Chatterino 2 build over [here](https://github.com/Chatterino/chatterino2/releases/tag/nightly-build)
 
-You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.  
+You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ Current releases are available at [https://chatterino.com](https://chatterino.co
 ## Nightly build
 You can download the latest Chatterino 2 build over [here](https://github.com/Chatterino/chatterino2/releases/tag/nightly-build)
 
-You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.  
+You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.
```
pajlada, line 13: Please revert this change
```suggestion
You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.  
```

herrhotzenplotz:
Should be fine now.

pajlada:
Thanks for the contribution!

PR #1787, opened 2020-07-05
Message menu now holds a shared pointer to a layout

### LINKED ISSUE #1783: Copy message crash
**Describe the bug**
A crash occurs when you try to copy a message that has been gone from the chat for a while (with text selected?).
When it happened the first time I had some text selected accidentally (don't know for how long) then I tried to copy a new recent message and it crashed.

**To reproduce**
1. join a fast chat 
1. wait for the message buffer to fill 
1. scroll up to the very beginning of the message buffer
1. select some text (not sure if required but i did this the first time)
1. right click any message
1. wait a minute or so
1. click "copy message"

**Screenshots**
it was too long for gfycat (you can skip the middle) https://youtu.be/hXHsha01IME

**Chatterino version**
Chatterino 2.1.7 (commit 845c8fa63)

**Operating system**
Windows 10

pajlada:
# Description
This ensures that the layout will survive for the lifetime of the menu,
     so any of the menu actions can with confidence do things with the
     layout, not having to worry whether it's dead or not.

This means that the user, while having the message menu open, could have
one extra MessageLayout alive. I have ensured that when the menu dies
the reference to the shared pointer dies with it.

Fixes #1783

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1570,12 +1570,12 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
     }
 
     // handle the click
-    this->handleMouseClick(event, hoverLayoutElement, layout.get());
+    this->handleMouseClick(event, hoverLayoutElement, layout);
 }
 
 void ChannelView::handleMouseClick(QMouseEvent *event,
                                    const MessageLayoutElement *hoveredElement,
-                                   MessageLayout *layout)
+                                   MessageLayoutPtr layout)
 {
     switch (event->button())
     {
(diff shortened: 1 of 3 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -1640,13 +1640,15 @@ void ChannelView::handleMouseClick(QMouseEvent *event,
...
+    const MessageLayoutElement *hoveredElement, MessageLayoutPtr layout)
 {
     const auto &creator = hoveredElement->getCreator();
     auto creatorFlags = creator.getFlags();
 
-    static QMenu *menu = new QMenu;
-    menu->clear();
+    auto menu = new QMenu;
+    connect(menu, &QMenu::aboutToHide, [menu] {
+        menu->deleteLater();  //
```
leon-richardt, line 1650: Here.

pajlada: Correct, this is just to avoid funky auto-formatting by `clang-format`

Without the trailing `//`, the `connect` statement becomes a single line
`    connect(menu, &QMenu::aboutToHide, [menu] { menu->deleteLater(); });`

leon-richardt: Okay, just making sure it wasn't missed by accident. 😄

PR #1809, opened 2020-07-14
Channel point rewards

Cranken:
Adds parsing and showing of channel point rewards in channels.

Currently there is still one unfixable drawback: We can't get the channel point icon. Twitch uses their gql endpoint for the image and we can't/shouldn't access that.

- [X] Show purple highlight next to reward messages
- [x] Requires an addition to pajladas signal library: https://github.com/pajlada/signals/pull/7
- [x] Show title(type,cost,image) for reward
- [x] Show rewards without user message
- [x] `F9` to print a debug message

### DIFF
--- a/src/messages/Message.hpp
+++ b/src/messages/Message.hpp
@@ -35,6 +35,7 @@ enum class MessageFlag : uint32_t {
     Debug = (1 << 18),
     Similar = (1 << 19),
     RedeemedHighlight = (1 << 20),
+    RedeemedChannelPointReward = (1 << 21),
 };
 using MessageFlags = FlagsEnum<MessageFlag>;
 
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -4,6 +4,7 @@
 #include "messages/Link.hpp"
 #include "messages/MessageColor.hpp"
 #include "singletons/Fonts.hpp"
+#include "src/messages/ImageSet.hpp"
 
 #include <QRect>
 #include <QString>
@@ -323,6 +324,7 @@ private:
     std::vector<Word> words_;
 };
 
+// Forces a linebreak
 class LinebreakElement : public MessageElement
 {
 public:
@@ -334,4 +336,16 @@ public:
     static LinebreakElement &instance();
 };
 
+// Image element which will pick the quality of the image based on ui scale
+class ScalingImageElement : public MessageElement
+{
+public:
+    ScalingImageElement(ImageSet images, MessageElementFlags flags);
+
+    void addToContainer(MessageLayoutContainer &container,
+                        MessageElementFlags flags) override;
+
+private:
+    ImageSet images_;
+};
 }  // namespace chatterino
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -108,8 +108,6 @@ enum class MessageElementFlag {
     // e.g. BTTV's SoSnowy during christmas season
     ZeroWidthEmote = (1 << 31),
 
-    Linebreak = (1 << 32),
-
     Default = Timestamp | Badges | Username | BitsStatic | FfzEmoteImage |
               BttvEmoteImage | TwitchEmoteImage | BitsAmount | Text |
               AlwaysShow,
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -40,6 +40,9 @@ enum class MessageElementFlag {
     BttvEmoteImage = (1 << 6),
     BttvEmoteText = (1 << 7),
     BttvEmote = BttvEmoteImage | BttvEmoteText,
+
+    ChannelPointReward = (1 << 8),
+
     FfzEmoteImage = (1 << 10),
     FfzEmoteText = (1 << 11),
     FfzEmote = FfzEmoteImage | FfzEmoteText,
@@ -326,12 +329,10 @@ private:
 class LinebreakElement : public MessageElement
 {
 public:
-    LinebreakElement();
+    LinebreakElement(MessageElementFlags flags);
 
     void addToContainer(MessageLayoutContainer &container,
                         MessageElementFlags flags) override;
-
-    static LinebreakElement &instance();
 };
 
 // Image element which will pick the quality of the image based on ui scale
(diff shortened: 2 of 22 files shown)


fourtf:
but channel point rewards already show?

Cranken:
The bit rewards show, channel point rewards do not. Atleast they didn't do in the latest nightly when I tested it.
What is currently shown is only the message, not the title,reward and cost.
E.g.
chrome_2020-07-14_11-43-20
Currently only the message is shown.

src/providers/twitch/ChannelPointReward.hpp
```diff
@@ -0,0 +1,34 @@
...
+        .arg(x)
+
+namespace chatterino {
+class ChannelPointReward
+{
+public:
+    ChannelPointReward(const rapidjson::Value &reward);
+    ChannelPointReward() = delete;
+
+    QString getRewardId() const;
```
fourtf, line 18: remove getters

src/providers/twitch/ChannelPointReward.hpp
```diff
@@ -0,0 +1,34 @@
...
+    QString getChannelId() const;
+    QString getRewardTitle() const;
+    int getRewardCost() const;
+    ImageSet getRewardImage() const;
+
+private:
+    QString rewardId;
```
fourtf, line 25: make public

src/providers/twitch/ChannelPointReward.hpp
```diff
@@ -0,0 +1,34 @@
...
+    QString channelId;
+    QString rewardTitle;
+    int rewardCost;
+    ImageSet rewardImage;
+
+    QString parseImage(const rapidjson::Value &image, const char *key);
```
fourtf, line 31: turn into freestanding function

src/providers/twitch/ChannelPointReward.hpp
```diff
@@ -0,0 +1,34 @@
...
+    QString getRewardId() const;
+    QString getChannelId() const;
+    QString getRewardTitle() const;
+    int getRewardCost() const;
+    ImageSet getRewardImage() const;
+
+private:
+    QString rewardId;
+    QString channelId;
+    QString rewardTitle;
```
fourtf, line 27: remove "reward" prefix in these 3 variables

src/providers/twitch/PubsubClient.cpp
```diff
@@ -1093,6 +1113,49 @@ void PubSub::handleMessageResponse(const rapidjson::Value &outerData)
...
+            return;
+        }
+
+        if (pointEventType == "reward-redeemed")
+        {
+            if (!rj::getSafe(msg, "data", msg))
+            {
+                if (!rj::getSafe(msg, "redemption", msg))
+                {
+                    if (!rj::getSafe(msg, "reward", msg))
```
fourtf, line 1132: don't stack ifs so deep

src/messages/Message.hpp
```diff
@@ -35,6 +35,7 @@ enum class MessageFlag : uint32_t {
     Debug = (1 << 18),
     Similar = (1 << 19),
     RedeemedHighlight = (1 << 20),
+    RedeemedChannelPointReward = (1 << 21),
```
fourtf, line 38: isn't this the same as `RedeemedHighlight` ?

Cranken: No. A channel point reward also uses a different IRC tag.

fourtf:
Code looks good so far!

Cranken:
After some more digging around on twitch I found some limitations:

- Twitch uses their private gql endpoint to retreive the possible rewards a channel offers. This means we can not just query an api to do the same, so we will have to use the pubsub topic.
- The message highlight reward is a seperate tag in IRC. This means no `custom-reward-id` is attached to the IRC message. 
This also means we don't receive a pubsub message which corresponds to the highlight reward and therefore can not display the reward info for highlights.

Cranken:
So, I've just implemented the parsing and layout for the point rewards.
Currently they it looks like this:
mpv_2020-07-15_14-28-42.
I've tried to replicate the look of the image I've posted above although if someone has some more input on the design/layout, I welcome it.
There is some margin on top which is a result of the reward image. I don't know what to do about that except maybe centering the reward text vertically if that's possible.

Cranken:
Also another meme I found while investigating: We don't actually receive a custom image for the channel reward points in any pubsub message so we will have to use the default image for now(or until twitch offers a public api for it...).

src/messages/MessageElement.hpp
```diff
@@ -39,6 +40,9 @@ enum class MessageElementFlag {
     BttvEmoteImage = (1 << 6),
     BttvEmoteText = (1 << 7),
     BttvEmote = BttvEmoteImage | BttvEmoteText,
+
+    ChannelPointReward = (1 << 8),
```
Cranken, line 44: I've squeezed it into a free value since the underlying type for the enum is only 32 bits in some systems.
I've tried to change the underlying type to something like a `std::uint64_t` but I encountered some issues with the `FlagEnum` container which I didn't know how to solve.

Cranken:
Does anyone have more input on this? Otherwise this is ready to review/merge.

alazymeme:
> Does anyone have more input on this? Otherwise this is ready to review/merge.

Now that you mention it and I'm testing it, is there potentially a way to add some padding to the text and make the icon smaller? Just looks a touch out of place:

image

Cranken:
> > Does anyone have more input on this? Otherwise this is ready to review/merge.
> 
> Now that you mention it and I'm testing it, is there potentially a way to add some padding to the text and make the icon smaller? Just looks a touch out of place:
> 
> image
@alazymeme 
I'll see what I can do, as long as it doesn't require a large rework in the current layout/rendering logic.

Evanito:
text-free redemptions don't scroll the chat box until another message comes in, or you scroll up THEN down. just scrolling down doesnt work

Cranken:
> text-free redemptions don't scroll the chat box until another message comes in, or you scroll up THEN down. just scrolling down doesnt work

@Evanito 
I can't reproduce this issue. Can you provide some more information/examples of when and how this happened?
How it looks for me: https://giant.gfycat.com/WarmheartedSillyHomalocephale.webm

Evanito:
doing a textless redemption:
image
you can see that: 
the redemption message has started below screen, 
has extended the scrollbar, 
has not autoscrolled, 
and i am unable to manually scroll down as it still thinks it is locked to the bottom

next user message:
image
scrolls past it and resumes normally

redemptions where the user enters a message scrolls normally

PR #1830, opened 2020-07-28
Update the PR Template

alazymeme:
# Description

Found a detailed pr template that looks nice. Modified a little bit, should suit quite nicely.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,40 @@
-# Description
+## Pull request checklist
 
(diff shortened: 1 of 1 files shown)


pajlada:
I don't like the template for PRs - something similar could probably be used to modify the issue templates if anything.
The PR title, branch name, and small description someone writes should be enough for 90% of PRs.
If your PR is so heavy that it needs all of this, then you should take it upon yourself to add the relevant information.

PR #1832, opened 2020-08-01
Alphabetized and Updated Contributor List

mfmarlow:
# Description

I alphabetized and added myself to the contributor list.


leon-richardt:
I was under the impression that the contributor list is not supposed to be in alphabetical order, hence the comment in line 33:
`# If you are a contributor add yourself above this line`

Issue #1820, opened 2020-07-23
New usercard does not auto-close anymore


RAnders00:
**What should be added?**
In the nightly builds of chatterino the usercard with the added chat history now no longer closes automatically. It used to be that the popup would close automatically when its focus was removed.

**Why should it be added?**
I liked the old behaviour more. I don't think there is really much of a use case to keeping the user popup open. It's especially annoying when you click on a name accidentally, or when you want to give a user a ban/timeout quickly as a moderator. You now have to explicitly press the escape button or the close button, and god forbid you focus chatterino before closing the popup, because the you have to bring the popup back to focus using the task bar or alt-tabbing.

rainstack:
It's probably a rare use-case, but the non-closing usercards have been helpful for modding. Makes it easier to follow and backtrack conversations to find the source of issues.

dnsge:
Should this be a setting then? Chatterino could default to the old auto-close behavior and have a setting to keep them open.

apa420:
I'd like this to be a toggleable setting imo.
I personally dislike the windows being open, but I can see the appeal for modding.

PR #1833, opened 2020-08-02
Added setting to auto close user popup

mammothb:
This attempts for fix https://github.com/Chatterino/chatterino2/issues/1820.

Apply `userPopup->setActionOnFocusLoss(BaseWindow::Delete);` only when the user has enabled the setting to "automatically close user popup when it loses focus". This will not close user popups created before toggling the option.

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -126,7 +126,7 @@ public:
     BoolSetting showParts = {"/behaviour/showParts", false};
     FloatSetting mouseScrollMultiplier = {"/behaviour/mouseScrollMultiplier",
                                           1.0};
-    BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", false};
+    BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", true};
     // BoolSetting twitchSeperateWriteConnection =
     // {"/behaviour/twitchSeperateWriteConnection", false};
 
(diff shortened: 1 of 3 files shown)


src/singletons/Settings.hpp
```diff
@@ -126,6 +126,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting showParts = {"/behaviour/showParts", false};
     FloatSetting mouseScrollMultiplier = {"/behaviour/mouseScrollMultiplier",
                                           1.0};
+    BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", false};
```
leon-richardt, line 129: Personally, I think the setting should be `true` by default since that is the behavior in the current stable version (2.1.7).

mammothb: Changed to be `true` by default

src/singletons/Settings.hpp
```diff
@@ -126,6 +126,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting showParts = {"/behaviour/showParts", false};
     FloatSetting mouseScrollMultiplier = {"/behaviour/mouseScrollMultiplier",
                                           1.0};
+    BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", true};
```
RAnders00, line 129: yay for making the default to be `true`. Good addition!

zneix:
Just wanted to point out that it's a good practice to also update changelog.md accordingly with new changes, because that's a very cool addition but I had no idea it existed until I looked through merged Pull Requests due to boredom.

PR #1838, opened 2020-08-06
Update QualityPopup.hpp

Danis-Enikeev:
Now this window automatically goes in floating mode in dwm (not sure if it works for i3 though).

### DIFF
--- a/src/widgets/dialogs/QualityPopup.hpp
+++ b/src/widgets/dialogs/QualityPopup.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <QDialog>
 #include <QComboBox>
+#include <QDialog>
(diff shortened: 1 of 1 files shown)


Mm2PL:
There seems to be a BasePopup class, maybe you could use that instead of QPopup.

Danis-Enikeev:
Tried the BasePopup with `setWindowFlags(Qt::Dialog)` but it didn't produce the desired result. It might be a good idea to rewrite the BasePopup class so it inherits from something QDialog based instead of BaseWindow.

pajlada:
Closing in favour of #1843

pajlada:
@Danis-Enikeev #1843 is being merged into master now - thanks for making me look into it!

PR #1870, opened 2020-08-16
fix a crash using middle click scroll on chat window

fourst4r:
Pull request checklist:

# Description
Middle click scrolling caused `QGuiApplication::screenAt(this->pos())` to return null and crash on Linux Mint 20, easy fix.

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1966,8 +1966,7 @@ void ChannelView::disableScrolling()
 
 void ChannelView::scrollUpdateRequested()
 {
-    const qreal dpi =
-        QGuiApplication::screenAt(this->pos())->devicePixelRatio();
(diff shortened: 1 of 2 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -1966,8 +1966,7 @@ void ChannelView::disableScrolling()
 
 void ChannelView::scrollUpdateRequested()
 {
-    const qreal dpi =
-        QGuiApplication::screenAt(this->pos())->devicePixelRatio();
+    const qreal dpi = this->devicePixelRatio();
```
pajlada, line 1969: `devicePixelRatio()` on a Widget returns an integer, not a qreal. Use `devicePixelRatioF()`

After that change has been done, the fun testing can begin :smile:

fourst4r: done!

PR #1881, opened 2020-08-20
dont ignore md files for CI

RAnders00:
Pull request checklist:

# Description

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,15 +1,7 @@
 ---
 name: Build
 
-on:
-  push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
+on: [push, pull_request]
 
 jobs:
   build:


.github/workflows/build.yml
```diff
@@ -1,15 +1,7 @@
...
 
-on:
-  push:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
-  pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
```
leon-richardt, line 12: Do we have to run the `build` workflow on those paths? Asking since only the `check-formatting` workflow is required for merging.

RAnders00:
@alazymeme why the thumbs down

.github/workflows/build.yml
RAnders00: I must have gotten them mixed up when I wrote the above description. `build` is actually required

image

leon-richardt: Interesting. On other PRs (e.g. #1883), the `check-formatting` workflow has the "Required" badge:
image

RAnders00: Oh right, the workflow is called `Check formatting` and the job inside the workflow is `build`, that's probably why. So you were right.

leon-richardt: That means we should be able to keep the ignores in the `build.yml` file, right? Or am I mixing something up now?

RAnders00: just applied your suggestion, thanks

PR #1883, opened 2020-08-21
Add option for truecase username autocompletion

### LINKED ISSUE #1199: Tab complete "Real Case" user names instead of login names (display name over login name)
Currently users get added to the list as their messages show up in here:
https://github.com/Chatterino/chatterino2/blob/3679125e2f959dbd7cb981bc25ab56b6a82e188a/src/common/Channel.cpp#L73-L78

and once in a while we fetch a full list from api:
https://github.com/Chatterino/chatterino2/blob/aac27c2a0d8fca560e4daa184075bb871e41f968/src/providers/twitch/TwitchChannel.cpp#L677-L699

The problem is that while we can fix the first one by just swapping to display name, the second requires an extra call to api to aquire the display name. I wrote a POC in python to show how it can be extracted.

```python3
#!/usr/bin/env python3
import argparse
import math
import requests

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--client-id', required=True)
    parser.add_argument('--target-channel', required=True, help='for example pajlada')
    parser.add_argument('--api', choices=['helix', 'kraken'], required=True)
    args = parser.parse_args()

    url = f'https://tmi.twitch.tv/group/user/{args.target_channel}/chatters'
    response = requests.get(url)
    response.raise_for_status()
    chatters = [chatter for group in response.json().get('chatters').values() for chatter in group]

    mapping = dict()
    at_once = 100
    for i in range(math.ceil(len(chatters) / at_once)):
        start = i * at_once
        end = start + at_once
        current_chatters = chatters[start:end]

        if args.api == 'helix':
            url = f'https://api.twitch.tv/helix/users?login={"&login=".join(current_chatters)}'
            headers = {'Client-Id': args.client_id}
            response = requests.get(url, headers=headers)
            response.raise_for_status()

            for user in response.json().get('data'):
                mapping[user['login']] = user['display_name']
        elif args.api == 'kraken':
            url = f'https://api.twitch.tv/kraken/users?login={",".join(current_chatters)}'
            headers = {'Client-Id': args.client_id,
                       'Accept': 'application/vnd.twitchtv.v5+json'}
            response = requests.get(url, headers=headers)
            response.raise_for_status()

            for user in response.json().get('users'):
                mapping[user['name']] = user['display_name']

    for login, display in mapping.items():
        # stolen from official twitch chat source code PepeS:
        # user.isIntl = user.login && user.displayName && user.displayName.trim().toLowerCase() !== user.login;
        if display.strip().lower() == login:
            print(f'{login}: {display}')
        else:
            print(f'{login}: {display}({login})')
```

```bash
 python3 main.py --client-id <REDACTED> --target-channel pajlada --api helix
```

I bet it can be nicely cached too. There is one caveat and it's the fact that some display names might be in CJK. Right now chatterino doesn't support tab completion with them, while twitch client is happy to accept for example japanese input and tab complete it.

Sadly I lack skills to do it properly in C++/Qt5 so I'll leave it as an exercise to the reader :alarm_clock:

PS: For now I swapped to `message->displayName` and commented out the request and it kinda works.

### LINKED ISSUE #1715: Lowercase username autocompletion
It would be nice that autocompleting usernames lowercases them.
That would prevent auto timeouts for capslock or special characters abusing usernames

fanway:
# Description

Added an option to lowercase usernames on autocompletion. By default, an option is set to true in order not to break a previously expected behavior. Unchecked gives us a truecase autocompletion.
Also fixes a case when some autocomplete usernames have truecase and some don't. Previously, nicknames loaded from the api call are in lowercase, but IrcMessageHandler uses displayName to add recent chatters, so we might get different behavior.
Closes #1199.
Closes #1715.


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -676,26 +681,40 @@ void TwitchChannel::refreshChatters()
...
+    {
+        getHelix()->getUserByName(
+            this->getName(),
+            [this](const auto &user) {
+                this->addRecentChatter(user.displayName);
+            },
+            [] {
+                // on failure
+            });
+    }
```
leon-richardt, line 717: This piece of code leads to quite different behavior depending on whether the setting is enabled or not.

If it is enabled, the behavior is the same as currently, i.e. users connected to the channel are fetched directly from TMI. This includes users that have not sent a message since you connected to the channel.

If the setting is not enabled, the piece of code will only add the channel owner's display name to the chatters list.
The problem¹  with this approach is that you can only autocomplete chatters that already typed something since you joined the chat room (and users from `recent-messages`). Users that connected to TMI but have not sent a message since you joined the channel are not considered. This feels inconsistent.

----
¹: It is arguable if people *actually* consider this a problem. In my opinion, the behavior should feels consistent regardless of whether the settings is applied or not.

fanway:
@fourtf i think to just static_cast back to TwitchChannel will be better solution.

alazymeme:
Could you please update your branch so I can test?

fanway:
Sure.

alazymeme:
This doesn't work for me. I unchecked the option and all my usernames are still being converted to lowercase. In saying this, the same behaviour occurs in 2.2.2 (commit b67e20d96, 27.09.2020). Guess I can't help you with testing.

fanway:
@alazymeme try to close and open chatterino again.

alazymeme (approved):
Works on Windows 10 for me

fanway:
Maybe i should add (requires restart) to the option. It basically happens because the UsernameSet needs to be cleared.

leon-richardt:
> Maybe i should add (requires restart) to the option. It basically happens because the UsernameSet needs to be cleared.

Could also clear the set (we would need to add an interface for that but I don't think that would be a problem)

alazymeme:
> Maybe i should add (requires restart) to the option. It basically happens because the UsernameSet needs to be cleared.

Maybe add a popup to prompt a restart?

`Enabling truecase username autocompletion requires a Chatterino restart. Would you like to restart?`
`Yes` | `No`

fanway:
> > Maybe i should add (requires restart) to the option. It basically happens because the UsernameSet needs to be cleared.
> 
> Could also clear the set (we would need to add an interface for that but I don't think that would be a problem)

I agree that clearing the set is pretty straightforward. But i'm not so sure how to force it on every opened tab on the setting change.

leon-richardt:
> I agree that clearing the set is pretty straightforward. But i'm not so sure how to force it on every opened tab on the setting change.

In order to get all open splits, you could check out [this](https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L19) function. Also refer to the [use](https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L87) further below.

fanway:
> > I agree that clearing the set is pretty straightforward. But i'm not so sure how to force it on every opened tab on the setting change.
> 
> In order to get all open splits, you could check out [this](https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L19) function. Also refer to the [use](https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L87) further below.

But how to catch the setting change?

leon-richardt:
> But how to catch the setting change?

Chatterino's settings emit a signal when they are changed (see [pajlada/settings](https://github.com/pajlada/settings) for more info).  Quick real-use example can be found [here](https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/helper/NotebookTab.cpp#L49), or a little more "self-contained" [here](https://github.com/pajlada/settings/blob/6de3a27784745959ab261df219880cddc9263ff9/src/test/signal.cpp#L6).

src/common/ChannelChatters.cpp
```diff
@@ -1,72 +1,77 @@
...
+{
+    *this->chatters_.access() = set;
+}
+
+void ChannelChatters::clearChatters()
+{
+    this->chatters_.access()->clear();
+}
+
+}  // namespace chatterino
```
apa420, line 77: Fix your line endings please :)

src/common/ChannelChatters.hpp
```diff
@@ -1,37 +1,38 @@
...
+    // combines multiple joins/parts into one message
+    UniqueAccess<QStringList> joinedUsers_;
+    bool joinedUsersMergeQueued_ = false;
+    UniqueAccess<QStringList> partedUsers_;
+    bool partedUsersMergeQueued_ = false;
+
+    QObject lifetimeGuard_;
+};
+
+}  // namespace chatterino
```
apa420, line 38: Also fix line endings here

src/common/ChannelChatters.cpp
apa420: use windows+linux ones

apa420: or whatever the project uses, I don't remember

fanway: Whats what I actually did. For some reason, these 2 files have dos line endings. All other files have unix. I changed it to unix ones.

src/common/ChannelChatters.hpp
fanway: Same here.

src/common/ChannelChatters.cpp
apa420: If they are correct now that's fine!

alazymeme:
Any chance this could be merged? 😃

pajlada (dismissed):
"true case" auto completion is the correct way, lower-case auto completion should be removed fully with this, not kept as a setting

fanway:
@pajlada kind of worked on another pr to use display/localized channel names in split title (whats why i changed `refreshChatters` to `fetchDisplayName`). Maybe i should name it different for this pr?

pajlada:
> @pajlada kind of worked on another pr to use display/localized channel names in split title (whats why i changed `refreshChatters` to `fetchDisplayName`). Maybe i should name it different for this pr?

Sorry, I did not read through this pr 100% before making my initial comment. The comment is still valid, but there's more to this issue than what I originally commented.

My ideal scenario would be:
Chatters are still loaded and inserted into the list of usernames.
If the chatter is already in the list, don't override someone's displayname.

FUTURE POTENTIAL: Have some job in the background fetching display names for users?

fanway:
> Sorry, I did not read through this pr 100% before making my initial comment. The comment is still valid, but there's more to this issue than what I originally commented.
> 
> My ideal scenario would be:
> Chatters are still loaded and inserted into the list of usernames.
> If the chatter is already in the list, don't override someone's displayname.
> 
> FUTURE POTENTIAL: Have some job in the background fetching display names for users?

I think i need to explain a little bit. Right now, chatters for autocompletion are loaded every 5 minutes from TMI endpoint. They are all in lowercase that's why autocompletion is in lowercase, because when a new message is sent, it tries to `addRecentChatter`, but because of `UsernameSet` it doesn't get updated. There is no way to make them truecase without a bunch of helix api calls (even twitch web doesn't do so). So i tried to use recent-chatters api instead TMI endpoint to load some previous chatters. Additionally, since where might be a not recent-chatters listener, i added `fetchDisplayName` for broadcaster (and because most of the time it's not present in recent-chatters and i think it's an important one for autocompletion). And i actually wanted to use `fetchDisplayName` to make split headers use display/localized names, but i will do it in another pr i guess.

That said, i don't think updating on the go (then new message is sent) is right either because we get different behavior (some nicknames will be lowercased, some don't).

fanway:
@pajlada can you please elaborate a little bit more on how you want this pr to be ended up implemented?

pajlada:
@fanway I believe the right way to go about achieving more "Display Name" tab completions than "Login Name" tab completions, is to modify UsernameSet to allow updating a name and prioritize storing the display name over the login name.

For example:

Chatters list is fetched.
Username `pajlada` is added to the UsernameSet.

When Chatterino User wants to tab complete pajlada, he types `paj` and presses <TAB>. Chatterino only knows of the full lowercase login name returned by the Chatters list, so Chatterino tab completes to `pajlada`.

`Pajlada` later types in chat. Chatterino tried to add `Pajlada` as a recent chatter, and UsernameSet should be intelligent enough to know that what's currently stored for `Pajlada` (`pajlada`) is a login name, since it's fully lowercase, and should attempt to update the stored name to `Pajlada` instead.

We need to be able to identify whether a username could be a login name or a display name, for example checking if the name contains any capital letters or not. Fully lowercase = assumed login name, anything else = assumed display name.

This means that if `Pajlada` has written in chat, and then the Chatterino list is fetched again, where the username is `pajlada` instead, `Pajlada` should *NOT* be overwritten by `pajlada` because the former is a Display Name and the latter is a login name.

Because we're never sure what's what in this case, we would have some potentiel errors if users change their display names halfway through.

Another option would be to, when adding something to UsernameSet, add a "priority" of sorts, where we know that a username is a login name, we *only* add it if it hasn't been added before, or if we know that a username is a display name, we know we *always* overwrite anything that's already in UsernameSet.

Let me know if my rambling was helpful

leon-richardt:
If we decide to go the route @pajlada suggested, we could also explore the possibility of caching the mapping "login name" -> "display name" client-side. I think in the long run, this would work out pretty well since in most channels, a majority of chatters are regulars.

Although some additional thought must be put towards deciding for how long names should be cached.

fanway:
@pajlada Thank you, it was really descriptive :) So, it seems like you are alright with updating on the go when new messages arrive and @leon-richardt seems too, so i guess that's the route.

fanway:
@leon-richardt caching looks quite interesting, but i think it would be applicable only for small channels, in the big ones i think it'll bloat right away, but i'm not so sure here.

pajlada (commented):
Let me know when you want me to review things again please :heart:

fanway:
@pajlada yep, my bad. It's ready to be reviewed

src/common/ChannelChatters.cpp
```diff
@@ -66,7 +66,12 @@ void ChannelChatters::addPartedUser(const QString &user)
 }
 void ChannelChatters::setChatters(UsernameSet &&set)
 {
-    *this->chatters_.access() = set;
+    this->chatters_.access()->merge(std::move(set));
+}
+
+void ChannelChatters::clearChatters()
```
pajlada, line 72: Remove this function since it's unused

src/common/ChannelChatters.hpp
```diff
@@ -18,6 +18,7 @@ class ChannelChatters
     void addJoinedUser(const QString &user);
     void addPartedUser(const QString &user);
     void setChatters(UsernameSet &&set);
+    void clearChatters();
```
pajlada, line 21: Remove this function since it's unused

src/common/UsernameSet.hpp
```diff
@@ -77,6 +77,8 @@ class UsernameSet
     std::pair<Iterator, bool> insert(QString &&value);
 
     bool contains(const QString &value) const;
+    void clear();
```
pajlada, line 80: Remove this function since it's unused

src/common/UsernameSet.cpp
```diff
@@ -68,6 +98,30 @@ bool UsernameSet::contains(const QString &value) const
     return this->items.count(value) == 1;
 }
 
+void UsernameSet::clear()
+{
+    this->items.clear();
+    this->firstKeyForPrefix.clear();
+}
```
pajlada, line 105: Remove this function since it's unused

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
 - Minor: Tab and split titles now use display/localized channel names (#2189)
 - Minor: Add a setting to limit the amount of historical messages loaded from the Recent Messages API (#2250, #2252)
+- Minor: Made username autocompletion truecase
```
leon-richardt, line 43: We should also append the relevant issue numbers here (#1199 and #1883?)

src/common/UsernameSet.cpp
```diff
@@ -43,15 +43,45 @@ std::set<QString>::size_type UsernameSet::size() const
...
+            if (QString::compare(*iter, value, Qt::CaseSensitive) != 0)
+            {
+                this->items.erase(iter);
+            }
+            else
+            {
+                return {iter, false};
+            }
+        }
+    }
```
pajlada, line 60: Could we make this into an internal helper function at the top of the file? Since it's uused in the ::merge function

pajlada (changes requested):

leon-richardt (changes requested):

pajlada:
Works great from testing - just a bit of code cleanup left then we're good to merge!

src/common/ChannelChatters.cpp
fanway: Oh yea, used them before last changes, forgot to clean it up

src/common/UsernameSet.cpp
fanway: You meant in another insert with move semantics? Sure

src/common/UsernameSet.cpp
```diff
@@ -4,6 +4,27 @@
...
+    return {set.end(), true};
+}
+
```
leon-richardt, line 27: Can we put this into the anonymous namespace? Like here for example:

https://github.com/Chatterino/chatterino2/blob/805e33e766a6019d26284851f5821fc00237751c/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L16-L31

leon-richardt (changes requested):

src/common/UsernameSet.cpp
fanway: Sure

src/common/UsernameSet.cpp
```diff
@@ -4,6 +4,31 @@
 
 namespace chatterino {
 
+namespace {
+    using namespace chatterino;
```
pajlada, line 8: This line is not necessary as we are already inside the chatterino namespace

```suggestion
```

pajlada (changes requested):

pajlada (approved):

zneix:
> Added an option to lowercase usernames on autocompletion. By default, an option is set to true in order not to break a previously expected behavior. Unchecked gives us a truecase autocompletion.

@fanway I might be missing something as I didn't follow this pull request, but wasn't this supposed to be an option that we can toggle/disable? I personally prefer to have lowercased username completions as some display names are often channel emotes in some chats.

leon-richardt:
Yeah, the setting got removed over review rounds. The initial PR description just hasn't been updated to reflect that :smile:

fanway:
@zneix yep, basically that @leon-richardt said, sorry for the confusion with pr title/description

PR #1914, opened 2020-08-24
(#1874) Disable updates on non-Windows, Linux or macOS platforms

### LINKED ISSUE #1874: UpdateWidget always reports "Failed to load version information" on FreeBSD
**Describe the bug**
Upon startup of chatterino the UpdateWidget (?, not important) on the top-left appears with a annoying big red arrow. If you click it, it reports: "Failed to load version information". This is a bug, because updates on FreeBSD are only available by compiling from sources.

**To reproduce**
Build and login to Twitch on FreeBSD.

**Screenshots**
2020-08-16-170505_81x47_scrot

**Chatterino version**
Chatterino 2.1.7 (commit 88f45a284)

**Operating system**
FreeBSD 12.1-RELEASE-p8 amd64

This might be resolved by adding https://notitia.chatterino.com/version/chatterino/freebsd/... . Otherwise, a quick hack might be to just lookup the Linux version info (which is really hackish).
The second approach is implemented here: https://github.com/herrhotzenplotz/chatterino2/commit/bacbcdd0e7313b43bb5c2f39d051494d409021f4 . It works and I am currently using every day.
It can easily be changed to use the first approach by changing the define of `CHATTERINO_OS` (see commit message)
Please tell me what is the desired approach. For the first one, someone else has to fix the service behind that URL. Once that is done, I'd open a PR. Otherwise I would just submit this patch.

herrhotzenplotz:
Pull request checklist:

# Description

As described #1874, only Windows, macOS and GNU/Linux are officially
supplied with builds. Thus checking for updates is unnecessary if we
are e.g. on FreeBSD, OpenBSD, illumos distros and possibly other
Operating Systems.

This fixes #1874 by ifdef-ing the 3 officially supported platforms
when checking for updates. Otherwise a debug warning will be emitted
and in the settings the checkbox for beta-updates is hidden and
replaced by a message, explaining the reason for why there are no beta
updates.

2020-08-24-102912_541x103_scrot

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -116,8 +116,13 @@ int Application::run(QApplication &qtApp)
 
     this->windows->getMainWindow().show();
 
+#if defined(Q_OS_WIN) || defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
     getSettings()->betaUpdates.connect(
         [] { Updates::instance().checkForUpdates(); }, false);
+#else
+    qDebug() << "Checking for beta updates disabled because OS doesn't appear "
+                "to be one of Windows, GNU/Linux or macOS.";
+#endif
     getSettings()->moderationActions.delayedItemsChanged.connect(
         [this] { this->windows->forceLayoutChannelViews(); });
 
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -116,13 +116,8 @@ int Application::run(QApplication &qtApp)
 
     this->windows->getMainWindow().show();
 
-#if defined(Q_OS_WIN) || defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
     getSettings()->betaUpdates.connect(
         [] { Updates::instance().checkForUpdates(); }, false);
-#else
-    qDebug() << "Checking for beta updates disabled because OS doesn't appear "
-                "to be one of Windows, GNU/Linux or macOS.";
-#endif
     getSettings()->moderationActions.delayedItemsChanged.connect(
         [this] { this->windows->forceLayoutChannelViews(); });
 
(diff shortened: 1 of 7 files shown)


src/Application.cpp
```diff
@@ -116,8 +116,13 @@ int Application::run(QApplication &qtApp)
 
     this->windows->getMainWindow().show();
 
+#if defined(Q_OS_WIN) || defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
```
fourtf, line 119: I would rather have this check in a function inside the Updates.hpp/cpp so all the update logic is grouped in one file

herrhotzenplotz: The reason why I did it this way is, that otherwise the signal will be connected, resulting in unnecessary calls to `checkForUpdates`. For that reason, having that check at those two places isn't as bad as you claim, but if you think that it's better to have an immediate return in checkForUpdates, I'll respect that.

src/singletons/Updates.cpp
```diff
@@ -232,6 +232,12 @@ void Updates::installUpdates()
 
 void Updates::checkForUpdates()
 {
+#if !defined(Q_OS_WIN) && !defined(Q_OS_LINUX) && !defined(Q_OS_MACOS)
```
fourtf, line 235: I would also put this check into a separate function or a #define since it is still duplicated in the GeneralPage

herrhotzenplotz: @fourtf I'd move it to Version.{hpp,cpp} and add something like Version::instance().isSupportedOS() and dispatch based on that. Is that a better approach in your opinion?

fourtf:
LGTM 👍

PR #1924, opened 2020-08-31
fix clang check when clang++ is called c++

swills:
The check for clang fails when `clang++` is installed as `c++`. This fixes it for me.


fourtf (approved):

swills:
wait

swills:
Don't merge this, there's a mistake here, let me revise.

swills:
Should be good now, sorry.

PR #1929, opened 2020-09-01
use packaged qtkeychain on FreeBSD

swills:
Since FreeBSD has packages of qtkeychain, use those when possible.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,5 +9,5 @@ task:
     git submodule update
     mkdir build
     cd build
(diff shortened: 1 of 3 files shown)


pajlada (changes requested):
I would like to move the system-specific things into the .pri-file itself, similar to how `/lib/rapidjson.pri` is set up.

Optimally, having something like QTKEYCHAIN_SYSTEM and QTKEYCHAIN_PREFIX flags would be best, but for the purpose of this PR, I'm fine it's just a

`freebsd {
    INCLUDEPATH += /usr/local/include/qt5keychain
    LIBS += -Lqt5keychain
} else {
   ORIGINAL INCLUDE PORTION
}```

basically moving the LIBS and INCLUDEPATH portion from the ci script over to the `lib/qtkeychain.pri` file instead.

pajlada (approved):

PR #1940, opened 2020-09-03
Argument parsing rework

### LINKED ISSUE #43: Commandline options

zneix:
Pull request checklist:

# Description

Makes use of `QCommandLineParser` and parses user input adding few dank features:
- Shows detailed description with `--help`:

- Handles joining only channels from command line, for example: `chatterino --channels pajlada,zneix,mm2pl` will start chatterino, but only with those 3 channels open in each split. I think that comma as split character is fine since it can't be a part of twitch username.
- To make sure this doesn't overwrite existing configuration in `window-layout.json`, saving in WindowManager is disabled when application was invoked with `--channels` argument.

Small TODO:
- [x] Add documentation
- [x] Rework handling `--channels` - it works, but it still uses strings to parse JSON, maybe there's a better solution?

Closes #43 

---

Notes to take a look at before merging:
1. Everything works fine on linux, however on Windows: `--help` gives you dialog window with help message, but `--version` doesn't show anything and quits - that was also a thing before, but we can fix it by using native `addVersionOption` method, however that will only output app name + version without commit, removing functionality added in #1739 which I don't wanna do.
2. Due to nature of how help message works, `--crash-recovery` is now shown in `--help` menu. I don't think it's a big issue but I can't seem to find an option to hide that either.


alazymeme:
Closing a stale issue? image

fourtf:
I suppose this should also disable settings from being changed and saved?

zneix:
> I suppose this should also disable settings from being changed and saved since they are not synced between instances?

Agree, it is ignoring saving layout to `window-layout.json` if the application was started with `-c` flag.

> Another solution would be to use IPC to add a new window to the original chatterino instance

Not sure what do you mean here, this doesn't create separate instances of chatterino - it only generates its own layout based on channels supplied in parameter while ignoring layout in `window-layout.json`.

fourtf:
> Agree, it is ignoring saving layout to `window-layout.json` if the application was started with `-c` flag.

I mean it might be good to extend this to the regular settings as well, not just the layout of the windows.

> Not sure what do you mean here, this doesn't create separate instances of chatterino - it only generates its own layout based on channels supplied in parameter while ignoring layout in `window-layout.json`.

Right now (unless I am mistaken) running chatterino with `--channel x` will open a new instance of chatterino instead of communicating to the running instance that it should open the channels in a new window.

pajlada (changes requested):
The `--crash-recovery` option should be hidden, see https://doc.qt.io/qt-5/qcommandlineoption.html#Flag-enum

As fourtf mentioned in the comments about, when chatterino has been launched with the `--channels` option, settings saving should also be disabled.

More open feedback: `--channels foo,bar` will now open two tabs with `foo` in the first tab, and `bar` in the second tab. This assumes Twitch channels, this should probably be specified somehow. And how the channels are opened should probably also be settable somehow.
random thoughts:
`--twitchchannels foo,bar` or `--channels twitch:foo,twitch:bar`
Maybe making the character inbetween change whether something opens in a new tab or not?
So maybe `--channels foo+bar` is a single tab with foo and bar horizontally split, and `--channels foo/bar` is vertically split? Not sure what characters are best to use.

If the "changing how channels are set up" ends up being too big, you're fine to implement that in a separate PR imo, just give a bit of thought to it so the `,` character can stay the same behaviour

zneix:
I changed the way how you specify channels - it's now `platform:channelName`, where platform is just a single letter. I don't know how IRC channel integration in Chatterino works, so I didn't implement those, but that can be easily added if there'll be need for it.

Also changed tab delimeter character to `;` because of following:
As for opening more splits in one tab, I thought of a following idea: `--channels t:foo,t:bar;t:asd,t:test` would open 2 tabs with 2 splits each:
1. containing Twitch channels `foo` and `bar`
2. containing Twitch channels `asd` and `test`

This could be done, but handling multiple vertical and horiziontal splits in one tab is horrible and at this point I think it's a bit overcomplicated, so I don't know if I should try implementing multiple splits in one tab with more parsed JSON or just drop that altogether, wait for eventual rework of `WindowManager.cpp` and then maybe even remove parsed JSON.

pajlada:
would prefer to use a character other than -, which could be part of a username atm.
maybe `:` instead?
and if no prefix is supplied, having Twitch be the default seems sane for this Twitch chat client

mlvzk:
I tried this and it works well, I use it in a script that runs streamlink and chatterino.

mlvzk (approved):

zneix:
Yeah, it is working fine with current code however I should get rid of JSON parsing and it is possible because #1964 is merged now. I just gotta rework that one bit and it will be good to go. I have to find some time in the week and I hope I can make it before this weekend. Thanks for approving though :)

fourtf (approved):

PR #1964, opened 2020-09-13
Split up Window Layout loading into a loading and application stage

pajlada:
Pull request checklist:

# Description

Previously, we were creating UI elements at while we were reading the window-layout.json file.
We now read the window-layout.json file fully first, which results in a
WindowLayout struct which is built up of a list of windows with a list
of tabs with a root node which contains containers and splits.
This WindowLayout can then be applied.

This will enable PRs like #1940 to start Chatterino with Window Layouts
that aren't defined in a json file.

This commit has deprecated loading of v1 window layouts (we're now on v2). If a v1 window layout is there, it will just be ignored and Chatterino will boot up as if it did not have a window layout at all, and on save that old window layout will be gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Deprecate loading of "v1" window layouts. If you haven't updated Chatterino in more than 2 years, there's a chance you will lose your window layout.
 - Minor: Disable checking for updates on unsupported platforms (#1874)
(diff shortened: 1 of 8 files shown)


leon-richardt:
Possibly related to following error which user `Turtoise#6731` posted on Discord?
parent-window_error

pajlada:
> Possibly related to following error which user `Turtoise#6731` posted on Discord?
> parent-window_error
> 
> 

More likely #1940

Issue #720, opened 2018-09-21
Problem with Languages using the Arabic Script - Hebrew and Arabic are reversed


menchxCodes:
Languages which use the Arabic Script write and read text from right to left. In chatterino however, when a message is written in these languages show up in the wrong direction, meaning the words are right-to-left, however the order in the sentence is left-to-right which makes it unreadable. Attached is a picture to better clarify what I mean:

![chatterino-rtl-issue](https://user-images.githubusercontent.com/43462510/45862686-284fb080-bd73-11e8-93fc-e3b020bf8ac6.JPG)

menchxCodes:
Hey Pajlada,

> یک دو سه

It also means "One two three" in case you were wondering.

fourtf:
Oh no

Wither-x:
What is the approximate time that when this bug will be fixed? Chatterino is unusable in Arabic channels.

fourtf:
we don't do time estimations

leon-richardt:
@menchxCodes @Wither-x @Nishyo

I just tried to implement a rudimentary way of rendering RTL languages in #1970. However, I don't use any RTL languages. If you get the chance, it would be great if you could download the version with this feature from [here](https://github.com/Chatterino/chatterino2/actions/runs/257936186) (make sure to get the right file for your operating system) and try it out.

Please feel free to report back any issues with that implementation. Thanks! :smile:

Wither-x:
Sorry for being late. I was busy when I got the email notice and totally forgot about it until now.. 😅

 It seems to be working the right way but there's a weird spacing problem 
![image](https://user-images.githubusercontent.com/71391066/93468973-7c58e580-f8f8-11ea-8b08-eb2dbaf1af34.png)

And tbh I didn't expect to get a respond with a fix as fast as you did.

leon-richardt:
Thanks for the reply. Any chance to paste a message producing weird spacing so I can test myself?

Wither-x:
Of course, here:

-  السلام عليكم 
- وعليكم السلام ورحمة الله وبركاته 

While I was trying to copy this happened.. https://gyazo.com/614653c208cbe127244a9b633e312a81

leon-richardt:
> Of course, here:
> 
> * السلام عليكم
> * وعليكم السلام ورحمة الله وبركاته
> 
> While I was trying to copy this happened.. https://gyazo.com/614653c208cbe127244a9b633e312a81

[...]

PR #1970, opened 2020-09-16
Make RTL languages render the correct way

leon-richardt:
Pull request checklist:

# Description

Implement a very naive fix for #720. It is the easiest way I could come up with. It would be great if an Arabic- or Hebrew-speaking user could test this to verify that the rendering works as expected.

This is what the example from https://github.com/Chatterino/chatterino2/issues/720#issuecomment-423440590 looks like in web chat and this implementation. (Which is the only reference point I have, really.)
### Chatterino:
chatterino
### Web Chat:
web_chat

**Please replace this code with anything more reasonable if you get the chance. I would also be fine with not merging this.** I'm not sure I would want this code in my own code base either. I wanted to offer the "fix" nonetheless. :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unversioned
+- Bugfix: Rudimentary support for right-to-left-languages (#1970)
 
 ## 2.3.0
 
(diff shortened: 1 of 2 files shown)


leon-richardt:
TODO:
- Fix https://github.com/Chatterino/chatterino2/issues/720#issuecomment-694205166
- Fix https://github.com/Chatterino/chatterino2/issues/720#issuecomment-694218194
- See if it's possible to find a cleaner solution

fourtf:
Can this be added in reverse order when layouting the message instead?

leon-richardt:
> Can this be added in reverse order when layouting the message instead?

Yeah, this will be the next thing to try

leon-richardt:
There is no point in keeping this open when I obviously don't have time to work on this. Closing so someone else might feel more comfortable tackling this.

PR #2001, opened 2020-09-27
Changed Streamer Mode settings

### LINKED ISSUE #1997: Allow individual toggling of the streamer mode features
**What should be added?**
Toggles for each of the streamer mode features (turning off image previews and avatars, etc)

**Why should it be added?**
Not all streamers have their chat on their main monitor. It should be enabled by default but have toggles to turn it off.

### LINKED ISSUE #2003: Add more streamer mode options
Add an option to:
- Mute pings while obs is open
- Only allow mentions from your own channel in your mentions window

All of these should be enabled by default.

Why?
- Someone can go into another channel and spam ping you as mod/vip (same concept as discord)
- Someone can go into another channel and mention you with racism or other tos shit

zneix:
Pull request checklist:

# Description

Original idea was to close #1997, but @alazymeme wants a toggle for each of those modules, so we came up with following solution:

Create another section in Settings called `Streamer Mode` (very similar to how it is in Discord), where you could set Streamer Mode to `enable`, `disable` or `detect obs` and then decide what said mode does, for example it could do every (or none) of following:
- Hide images in link tooltips
- Hide images in usercards
- Hide viewer count and stream length in Header Split tooltip (?)
- Mute pings from your own tab / all tabs

Closes #1997
Closes #2003 (second part of the issue can be fixed with #1748)


zneix:
Any idea why [Windows artifact](https://github.com/Chatterino/chatterino2/pull/2001/checks?check_run_id=1172496906) contains only 3 files? Something broke while building?

alazymeme (changes requested):
```
E:\GitHub\chatterino\chatterino2\src\util\StreamerMode.cpp:30: error: C2561: 'chatterino::isInStreamerMode': function must return a value
```

zneix:
> I get this error while trying to compile:

Yeah, function returned wrong type because of me being donk. Fixed in 2e201fe

alazymeme (approved):

alazymeme:
> Closes #1997

Also please remove this, it doesn't fully fix the issue (as discussed you plan this in a later pr)

apa420:
In my opinion instead of disabling and enabling avatars in streamer mode. How about a setting for `disable`,`enable` or `detect obs` the streamermode.

zneix:
> 
> 
> In my opinion instead of disabling and enabling avatars in streamer mode. How about a setting for `disable`,`enable` or `detect obs` the streamermode.

I updated PR Description with an idea of how we can handle the whole thing better

apa420:
Would you also add an option for enable/disable and detect the streamermode? :)

zneix:
> 
> 
> Would you also add an option for enable/disable and detect the streamermode? :)

Yes, I will try to add that as explained: "you could set Streamer Mode to enable, disable or detect obs"

src/util/StreamerMode.cpp
```diff
@@ -25,46 +27,48 @@ const QStringList &broadcastingBinaries()
 bool isInStreamerMode()
 {
 #ifdef USEWINSDK
-    if (IsWindowsVistaOrGreater())
+    if (!IsWindowsVistaOrGreater() || !getSettings()->hideImagesInStreamerMode)
```
fourtf, line 30: This doesn't make sense here. Move it to the place in the code where it checks if it should display the images.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -678,6 +678,9 @@ void GeneralPage::initLayout(SettingsLayout &layout)
     layout.addCheckbox(
         "Hide viewercount and stream length while hovering the split",
         s.hideViewerCountAndDuration);
+    layout.addCheckbox(
```
fourtf, line 681: Instead add a new section similar to the following:

> ### Streamer Mode
> 
> Chatterino can automatically change behavior if it detects that "OBS Studio" running. Select which things you? want to change while streaming:
> 
> [ ] Hide unsafe user content (Profile pictures and thumbnails)

This allows people to understand what the feature does and how it triggers (obs being open).

src/singletons/Settings.hpp
```diff
@@ -316,6 +316,8 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/misc/attachExtensionToAnyProcess", false};
     BoolSetting hideViewerCountAndDuration = {
         "/misc/hideViewerCountAndDuration", false};
+    BoolSetting hideImagesInStreamerMode = {"/misc/hideImagesInStreamerMode",
```
fourtf, line 319: something like `/misc/streamerMode/hideUserContent` would maybe make more sense?

fourtf (changes requested):

src/singletons/Settings.hpp
apa420: maybe even `/streamerMode/hideUserContent` might be ok

src/widgets/settingspages/GeneralPage.cpp
zneix: Resolved in b9503b7

src/singletons/Settings.hpp
zneix: >
> maybe even /streamerMode/hideUserContent might be ok

I went ahead and applied this suggestion in b9503b7, however split link thumbnails and usercard avatars to two separate settings as alazymeme requested in his issue.

src/util/StreamerMode.cpp
zneix: After some changes in b9503b7 this fixed now.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -472,6 +473,48 @@ void GeneralPage::initLayout(SettingsLayout &layout)
...
+        s.enableStreamerMode,
+        [](auto val) {
+            switch (val)
+            {
+                case StreamerModeSetting::Enabled:
+                    return "Yes";
+                case StreamerModeSetting::Disabled:
+                    return "No";
+                case StreamerModeSetting::DetectObs:
+                    return "Detect OBS";
```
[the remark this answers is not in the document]
zneix: Fixed in 00925f1a

zneix:
This is now ready for review, I've fixed all addressed changes and integrated all the settings into a new sub-section in `Settings -> General`. This is how it currently looks with default options:
https://cdn.zneix.eu/iAj3Dvq.png
zulul

Only thing that I am not sure about is whether I want to move `/misc/hideViewerCountAndDuration` setting to `/streamerMode/hideViewerCountAndDuration` in `Settings.hpp`. Imo it makes more sense for that feature to be integrated into Streamer Mode, but moving it could break this feature for some users that are already using that setting.
I would like to hear some feedback on this.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -472,6 +473,48 @@ void GeneralPage::initLayout(SettingsLayout &layout)
             return fuzzyToInt(args.value, 0);
         });
 
+    layout.addTitle("Streamer Mode");
+    layout.addDescription(
+        "Chatterino can automatically change behavior if it "
+        "detects that \"OBS Studio\" is running.\nSelect which "
+        "things you want to change while streaming");
+    layout.addDropdown<int>(
```
apa420, line 481: If you use an enum maybe you should make something similar to addKeyBoardModifierSetting (a function in this file)

apa420: Also this shouldn't be a setting you can type stuff in

apa420 (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added an option to toggle hiding user avatars while OBS is running (#2001, commit 9739d8d)
```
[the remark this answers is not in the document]
apa420: Make this a major instead and mention streamer mode?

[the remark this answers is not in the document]
apa420: Something like  - Major: Added streamer mode configuration

src/widgets/settingspages/GeneralPage.cpp
zneix: Fixed both things in 1d08fed8, huge thanks for noticing them!

src/singletons/Settings.hpp
```diff
@@ -177,7 +178,16 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting unshortLinks = {"/links/unshortLinks", false};
     BoolSetting lowercaseDomains = {"/links/linkLowercase", true};
 
-    /// Ignored phrases
+    /// Streamer Mode
+    IntSetting enableStreamerMode = {"/streamerMode/enabled",
```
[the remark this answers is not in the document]
zneix: Fixed in 1d08fed8

CHANGELOG.md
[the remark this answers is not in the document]
zneix: Fixed in 1d08fed8. Hope it's good enough.

src/widgets/settingspages/GeneralPage.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <QDebug>
 #include <QLabel>
+#include <QPushButton>
```
apa420, line 5: This shouldn't be needed?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1,5 +1,7 @@
 #include "GeneralPage.hpp"
 
+#include <QDesktopServices>
+#include <QFileDialog>
```
apa420, line 4: Is this needed?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -472,6 +475,50 @@ void GeneralPage::initLayout(SettingsLayout &layout)
...
+            {
+                return StreamerModeSetting::Enabled;
+            }
+            else if (args.value == "No")
+            {
+                return StreamerModeSetting::Disabled;
+            }
+            else
+            {
+                // default to Detecting if OBS.exe is running
```
apa420, line 511: I don't understand this comment. There probably shouldn't even be a comment here.

apa420: default to, means it's actively defaulting to something. Wording is different from default is

zneix: I tried following this code here:
https://github.com/Chatterino/chatterino2/blob/b67e20d9620c415229bebc190f08d70b13e0cf31/src/widgets/settingspages/GeneralPage.cpp#L321-L324

alazymeme:
I don't like this Streamer Mode image showing when I hover over links - can I have the option just to not display anything?

zneix:
> I don't like this Streamer Mode image showing when I hover over links - can I have the option just to not display anything?

I wanted to be consistent with Streamer Mode image displayed for usercard avatars. Thought it could be a nice addition but I can remove it if it's more of a useless feature than enhancement.

alazymeme:
> I can remove it if it's more of a useless feature than enhancement.

It's not the biggest issue, some people (like me) would just rather the option to disable it (even though I probably wont use the features)

apa420:
Enabling streamer mode + link thumbnails crashes chatterino for me.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1316,13 +1316,24 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
...
+                    getSettings()->streamerModeHideLinkThumbnails &&
+                    !element->getThumbnail()->url().string.isEmpty())
+                {
+                    tooltipPreviewImage.setImage(
+                        Image::fromPixmap(getResources().streamerMode));
+                }
+                else
+                {
+                    tooltipPreviewImage.setImage(element->getThumbnail());
+                }
```
23rd, line 1335:
```suggestion
                const auto isHideLink =
                    isInStreamerMode() &&
                    getSettings()->streamerModeHideLinkThumbnails &&
                    !element->getThumbnail()->url().string.isEmpty();
                auto thumb =
                    isHideLink ? Image::fromPixmap(getResources().streamerMode)
                               : element->getThumbnail();
                tooltipPreviewImage.setImage(std::move(thumb));
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -472,6 +473,49 @@ void GeneralPage::initLayout(SettingsLayout &layout)
...
+            }
+            else if (args.value == "No")
+            {
+                return StreamerModeSetting::Disabled;
+            }
+            else
+            {
+                return StreamerModeSetting::DetectObs;
+            }
+        },
```
23rd, line 511: Wondering why this part is so compicated.
```suggestion
        s.enableStreamerMode, [](int value) { return value; },
        [](DropdownArgs args) {
            return static_cast<StreamerModeSetting>(args.index);
        },
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -1316,13 +1316,24 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
         {
             auto element = &hoverLayoutElement->getCreator();
             auto thumbnailSize = getSettings()->thumbnailSize;
-            if (thumbnailSize == 0 || isInStreamerMode())
+            if (thumbnailSize == 0)
```
23rd, line 1319: Can be simplified.
```suggestion
            if (!thumbnailSize)
```

src/util/StreamerMode.cpp
```diff
@@ -24,47 +26,60 @@ const QStringList &broadcastingBinaries()
...
+    if (getSettings()->enableStreamerMode.getEnum() ==
+        StreamerModeSetting::Enabled)
+    {
+        return true;
+    }
+    if (getSettings()->enableStreamerMode.getEnum() ==
+        StreamerModeSetting::Disabled)
+    {
+        return false;
+    }
```
23rd, line 38:
```suggestion
    const auto mode = getSettings()->enableStreamerMode.getEnum();
    if (mode == StreamerModeSetting::Enabled)
    {
        return true;
    }
    else if (mode == StreamerModeSetting::Disabled)
    {
        return false;
    }
```

apa420: Case statement might also work here.

src/widgets/helper/ChannelView.cpp
apa420: I don't really mind this being more explicit

src/widgets/helper/ChannelView.cpp
```diff
@@ -1316,13 +1316,24 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
...
             {
                 tooltipPreviewImage.setImage(nullptr);
             }
             else
             {
-                tooltipPreviewImage.setImage(element->getThumbnail());
+                if (isInStreamerMode() &&
+                    getSettings()->streamerModeHideLinkThumbnails &&
+                    !element->getThumbnail()->url().string.isEmpty())
```
apa420, line 1327:
```suggestion
                    getSettings()->streamerModeHideLinkThumbnails)
```
This here will make chatterino not crash :).
Also I don't understand in the first place why you would check if the element has a thumbnail url?

zneix: > I don't understand in the first place why you would check if the element has a thumbnail url?

I do that to check if link actually has a thumbnail to not display Streamer Mode image for links that have no thumbnail. I was looking on how to do that and found out that links without thumbnail URL has obviously no thumbnail, so I used that to determine whether we should display the static image or not. Maybe there's a better way to do that, but this is just a quick solution I found yesterday.

zneix: Should be fixed by e0f24d7.

zneix:
I moved `/misc/hideViewerCountAndDuration` setting to `/streamerMode/hideViewerCountAndDuration`. I know that with this change users that were using this feature might have to set it again, however, based on original request of that feature from #1621, I can tell that this is only/mostly used by streamers, so they should notice that with major change that this PR brings to Chatterino and it shouldn't be that much of an issue in the end.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -472,6 +473,28 @@ void GeneralPage::initLayout(SettingsLayout &layout)
...
         });
 
+    layout.addTitle("Streamer Mode");
+    layout.addDescription(
+        "Chatterino can automatically change behavior if it "
+        "detects that \"OBS Studio\" is running.\nSelect which "
+        "things you want to change while streaming");
+
+    layout.addDropdown<std::underlying_type<StreamerModeSetting>::type>(
+        "Enable Streamer Mode", {"No", "Yes", "Detect OBS (win)"},
```
pajlada, line 483: change (win) to (Windows only)

pajlada (changes requested):

src/widgets/settingspages/GeneralPage.cpp
apa420: The issue with that is in the settings it won't show the entire text in the dropdown.

pajlada: then the other option would be to just not show that option if you're not on windows

zneix: Changed that to Windows only and made the dropdown menu a bit wider, hope that's okay.

pajlada (approved):

PR #2002, opened 2020-09-27
Fix links and grammar and shit

alazymeme:
Pull request checklist:

# Description

Updated some links in some pages + fixed up some grammar

### DIFF
--- a/src/widgets/helper/EditableModelView.cpp
+++ b/src/widgets/helper/EditableModelView.cpp
@@ -116,8 +116,8 @@ void EditableModelView::addRegexHelpLink()
 {
     auto regexHelpLabel =
         new QLabel("<a href='"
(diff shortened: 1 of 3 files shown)


src/widgets/helper/EditableModelView.cpp
```diff
@@ -116,8 +116,8 @@ void EditableModelView::addRegexHelpLink()
 {
     auto regexHelpLabel =
         new QLabel("<a href='"
-                   "https://chatterino.com/help/regex'>"
```
fourtf, line 119: This is there so the link can be changed after the fact.

fourtf:
The links are to the chatterino website so they can be changed after the fact. However since this commit again contains multiple changes at once I'll close it.

PR #2011, opened 2020-09-30
updated the readme.md 

cinemassacres:
Pull request checklist:

# Description

updates llvm to be 10.0 instead of 9.0

### DIFF
--- a/BUILDING_ON_WINDOWS.md
+++ b/BUILDING_ON_WINDOWS.md
@@ -29,7 +29,7 @@ Note: This installation will take about 1.5 GB of disk space.
 
 ## OpenSSL
 ### For our websocket library, we need OpenSSL 1.1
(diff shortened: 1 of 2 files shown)


zneix:
I don't think that Ubuntu 19.10 (as described in [workflow file](https://github.com/Chatterino/chatterino2/blob/master/.github/workflows/check-formatting.yml)) uses clang-format v10 in its apt repositories, thus such change might lead towards confusion, where users will be using clang-format v10, but formatting check on pull requests will be screaming at them for errors in their formatting because v9 - differs a bit.

leon-richardt:
> I don't think that Ubuntu 19.10 (as described in [workflow file](https://github.com/Chatterino/chatterino2/blob/master/.github/workflows/check-formatting.yml)) uses clang-format v10 in its apt repositories

Appears to be the case according to [Ubuntu Launchpad](https://launchpad.net/ubuntu/+source/llvm-defaults/0.49~exp1). The official 19.10 repositories [definitely do not contain](https://launchpad.net/ubuntu/eoan/+search?text=clang-format) `clang-format-10`.

PR #2016, opened 2020-10-01
Commands are now saved upon pressing Ok in the settings window

gempir:
Currently commands are only saved on exiting chatterino. I shutdown my computer without closing chatterino beforehand and always lose my new commands because of that.

With this PR commands are saved once you press the Ok Button in the Settings, similar to the rest of the settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,5 +73,8 @@ rapidjson/*
 
 Thumbs.db
 
-#I HATE MAC
-.DS_Store
+# I HATE MAC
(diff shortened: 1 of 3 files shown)


alazymeme (changes requested):
Good Evening Sir/Madam

It appears that your pull request is missing the "Changelog Entry". Could you please kindly fix?

Kind Regards
ALazyMeme

gempir:
Added changelog entry

PR #2020, opened 2020-10-02
Recent messages timestamp fix

### LINKED ISSUE #1286: NOTICE and USERNOTICE messages loaded from recent messages service have incorrect timestamp
**Describe the bug**
When you load a channel where `NOTICE`s (or `USERNOTICE`s) are in the last 500 messages (e.g. a sub-mode-on message), then the timestamp on those messages have the wrong timestamp.

**To Reproduce**
- find a channel with a message like `This room is now in emote-only mode.`, where that message is at least 1 minute old (so you can notice the difference)
- reopen chatterino
- observe that the timestamp on the message is the time of opening chatterino

**Expected behavior**
The timestamp is calculated from the `rm-received-ts` tag, and not the current time. E.g.:

```
@rm-received-ts=1567706276587;historical=1;msg-id=emote_only_on :tmi.twitch.tv NOTICE #trihex :This room is now in emote-only mode.
```

is a message that the `recent-messages` service returns. `rm-received-ts` is **always** set.

**Screenshots**

image

**Chatterino version**
2.1.2

**Additional context**

dnsge:
Pull request checklist:

# Description

Fixes inconsistent timestamps on messages received through the recent-messages API. 

Closes #1286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Bugfix: Handle symlinks properly when saving commands & settings (#1856, #1908)
 - Bugfix: Starting Chatterino in a minimized state after an update will no longer cause a crash
 - Bugfix: Modify the emote parsing to handle some edge-cases with dots and stuff (#1704, #1714)
+- Bugfix: Fix recent messages timestamp (#1286)
 
 ## 2.2.0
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - Bugfix: Handle symlinks properly when saving commands & settings (#1856, #1908)
 - Bugfix: Starting Chatterino in a minimized state after an update will no longer cause a crash
 - Bugfix: Modify the emote parsing to handle some edge-cases with dots and stuff (#1704, #1714)
+- Bugfix: Fix recent messages timestamp (#1286)
```
RAnders00, line 19:
```suggestion
- Bugfix: Fixed timestamps being incorrect on some messages loaded from the recent-messages service on startup (#1286, #2020)
```

RAnders00 (changes requested):

pajlada (approved):
:+1: Looks good to me! Thanks for your contribution.

PR #2021, opened 2020-10-03
Line on new day

### LINKED ISSUE #1016: Add a message separating days after 12 am
Some messages may have been sent weeks or days apart.
When loading chat history and encountering such, i suggest to add a date label split.

<details>
  <summary>Currently at Ch2 (screenshot)</summary>
  

</details>

<details>
  <summary>Desirable behaviour (screenshot of example in another app)</summary>
  

(it is Telegram Desktop)
</details>

apa420:
Pull request checklist:

# Description

Closes #1016
(There is currently an issue that will be fixed by #2020)

### DIFF
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -151,6 +151,10 @@ public:
 
     virtual void addToContainer(MessageLayoutContainer &container,
                                 MessageElementFlags flags) = 0;
+    virtual QTime getTime()
+    {
+        return QTime(0, 0);
+    };
 
     pajlada::Signals::NoArgSignal linkChanged;
 
@@ -278,6 +282,8 @@ public:
                         MessageElementFlags flags) override;
 
     TextElement *formatTime(const QTime &time);
+
+    QTime getTime() override;
 
 private:
     QTime time_;
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -6,9 +6,9 @@
 #include "singletons/Fonts.hpp"
 #include "src/messages/ImageSet.hpp"
 
+#include <QDateTime>
 #include <QRect>
 #include <QString>
-#include <QTime>
 #include <boost/noncopyable.hpp>
 #include <cstdint>
 #include <memory>
@@ -151,9 +151,9 @@ public:
 
     virtual void addToContainer(MessageLayoutContainer &container,
                                 MessageElementFlags flags) = 0;
-    virtual QTime getTime()
+    virtual QDateTime getTime()
     {
-        return QTime();
+        return QDateTime();
     };
 
     pajlada::Signals::NoArgSignal linkChanged;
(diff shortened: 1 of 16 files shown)


src/messages/MessageElement.hpp
```diff
@@ -151,6 +151,10 @@ class MessageElement : boost::noncopyable
 
     virtual void addToContainer(MessageLayoutContainer &container,
                                 MessageElementFlags flags) = 0;
+    virtual QTime getTime()
+    {
+        return QTime(0, 0);
```
23rd, line 156:
```suggestion
        return QTime();
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -917,8 +917,24 @@ void ChannelView::drawMessages(QPainter &painter)
...
 
+        bool isPastMidnight = false;
+        if (i != messagesSnapshot.size() - 1)
+        {
+            auto elements = &messagesSnapshot[i]->getMessage()->elements;
+            auto elementsAfter =
+                &messagesSnapshot[i + 1]->getMessage()->elements;
+
+            for (int j = 0; j < elements->size(); j++)
+                if ((*elements)[j]->getTime() != QTime(0, 0))
```
23rd, line 928:
```suggestion
                if (!(*elements)[j]->getTime().isNull())
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -917,8 +917,24 @@ void ChannelView::drawMessages(QPainter &painter)
...
+                    for (int k = 0; k < elementsAfter->size(); k++)
+                        if ((*elementsAfter)[k]->getTime() != QTime(0, 0))
```
23rd, line 930:
```suggestion
                        if (!(*elementsAfter)[k]->getTime().isNull())
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -917,8 +917,24 @@ void ChannelView::drawMessages(QPainter &painter)
...
+                            isPastMidnight =
+                                (*elements)[j]->getTime().hour() >
+                                (*elementsAfter)[k]->getTime().hour();
```
23rd, line 933:
```suggestion
                            isPastMidnight = (*elements)[j]->getTime() >
                                             (*elementsAfter)[k]->getTime();
```

23rd:
I think it might be easier if we look not for the last message of the previous day, but for the first message of the next day.
You can check my [PR](https://github.com/apa420/chatterino2/pull/4) to your current branch.

src/widgets/helper/ChannelView.cpp
apa420: Reason I wanted the hour was because I was unsure if I'd ever get a race condition where one message gets sent before the other one. But that case might not happen.

pajlada (changes requested):
Wouldn't work for this scenario:
```
01:00 (2020-10-04)
01:01 (2020-10-04)
01:02 (2020-10-04)
----------------
01:00 (2020-10-05)
01:02 (2020-10-06) <-- NO LINE ADDED HERE :(
```

apa420:
Will use QDateTime for messages

fourtf (changes requested):
I would like the new date to be included as well

apa420:
This approach is very stupid, will implement with a timer that send a signal on a new day instead!

PR #2024, opened 2020-10-03
Changed some default settings

### LINKED ISSUE #1992: Enable live channels in tabs by default
**What should be added?**
This setting seems useful and not harmful to normal users.

Example: image

apa420:
Pull request checklist:

# Description

Closes #1992


src/singletons/Settings.hpp
```diff
@@ -79,7 +79,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting hideModerationActions = {
         "/appearance/messages/hideModerationActions", false};
     BoolSetting colorizeNicknames = {"/appearance/messages/colorizeNicknames",
```
gempir, line 81: This is going to cause a lot of confusion to existing Chatterino users, it should at least be reflected in the changelog.

PR #2029, opened 2020-10-04
Improve "Login expired!" message

leon-richardt:
Pull request checklist:

# Description

The message display when an OAuth token expires repeatedly lead to users asking about what they should do in order to fix it. This PR changes the error message in order to clarify that users need to re-add their account. Additionally, a link to the "Accounts" page is included in the message (see video).

Video: https://i.nuuls.com/v0qWI.mp4


fourtf:
Looks good to me 👍 Thank you for your contribution!

PR #2035, opened 2020-10-05
Make "moderation mode" setting persist

### LINKED ISSUE #2032: Moderation Mode not saying enabled after relaunch
**Describe the bug**
After enabling moderation mode it does not stay enabled.

**To reproduce**
Simply enable moderation mode for a channel, close your Chatterino, relaunch Chatterino and moderation mode is not enabled.

**Chatterino version**
Chatterino Nightly 2.2.2 (commit 42daac088, 04.10.2020)

**Operating system**
Windows 10

dnsge:
Pull request checklist:

~~`CHANGELOG.md` was updated, if applicable~~
(fixes bug in unversioned change)

# Description
Fixes a small bug introduced in #1964. "Moderation mode" would be written to the settings file but wouldn't be loaded. 

Closes #2032

### DIFF
--- a/src/common/WindowDescriptors.cpp
+++ b/src/common/WindowDescriptors.cpp
@@ -76,6 +76,7 @@ void SplitDescriptor::loadFromJSON(SplitDescriptor &descriptor,
 {
     descriptor.type_ = data.value("type").toString();
     descriptor.server_ = data.value("server").toInt(-1);
+    descriptor.moderationMode_ = root.value("moderationMode").toBool();
     if (data.contains("channel"))
     {
         descriptor.channelName_ = data.value("channel").toString();


pajlada (approved):
Thanks!

fourtf:
Where in the code does it get loaded?

dnsge:
The `moderationMode` value is read from the split descriptor in these two places:

https://github.com/Chatterino/chatterino2/blob/fdb24449e721f68b22e9125caa3866414116c806/src/widgets/splits/SplitContainer.cpp#L718

https://github.com/Chatterino/chatterino2/blob/fdb24449e721f68b22e9125caa3866414116c806/src/widgets/splits/SplitContainer.cpp#L750

The issue was that `splitNode.moderationMode_` wasn't being read from the JSON file and instead defaulted to `false`.

https://github.com/Chatterino/chatterino2/blob/fdb24449e721f68b22e9125caa3866414116c806/src/common/WindowDescriptors.hpp#L37-L38

PR #2057, opened 2020-10-10
Made the current channels emotes appear at the top of the emote picker

Y0dax:
Pull request checklist:

# Description

 Made the current channels emotes appear at the top of the emote picker popup. This makes them easier to find and the user no longer has to scroll down the picker to get them. The remaining are ordered as nornal.

image


pajlada (approved):
:+1:

PR #2059, opened 2020-10-10
Viewer list usability improvements

Y0dax:
Pull request checklist:

# Description

- Added channel name to viewer list window title.
- Synced viewer list font settings with global setting.
- Improved distinction of labels in the list by using the theme accent color and adding a space between the categories.

9pt font and 12pt font
imageimage

Light and white themes 9pt font
imageimage

EDIT:
Reordered Moderators and VIPs to match twitch
image


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Improve UX of Viewer List window and sync font with settings.
```
TranRed, line 5: I would suggest changing this to "Minor: Improved UI of the viewer list window"

src/widgets/splits/Split.cpp
```diff
@@ -635,7 +636,9 @@ void Split::showViewerList()
     for (auto &x : labels)
     {
         auto label = new QListWidgetItem(x);
-        label->setBackgroundColor(this->theme->splits.header.background);
+        label->setBackground(this->theme->splits.header.background);
+        label->setForeground(QColor(169, 112, 255));
```
TranRed, line 640: I don't think we use twitch purple anywhere else in the application. It feels out of place to me to just do it in one very specific UI. Someone correct me if I'm wrong or they feel otherwise.

Y0dax: @TranRed The color is arbitrary to me, but I think it should be different than the usernames or vice versa. Twitch has white labels and purple usernames. Chatterino does not really have an accent color that I could tell, so I was unsure of what to go with. 

Do we have a theme color that makes sense here?

CHANGELOG.md
Y0dax: Done

src/widgets/splits/Split.cpp
Y0dax: Done

TranRed: I think the closest we have to an accent color is focusedText. But this is just my interpretation. No guarantee it's intended to use in this way.
https://github.com/Chatterino/chatterino2/blob/95c909b33755294a0227ded3c6941ec87d9d50fe/src/singletons/Theme.cpp#L54-L55

If we introduce an accent color it should go into Theme.hpp/cpp though, so it can be reused in other places.

pajlada: In the latest master, there's an `accent` variable you can use from theme - try that out for the color of the "headers"

Y0dax: I updated the headers to use the new theme accent color. It looks great on dark theme.

I don't know where this app/team stands on accessibility, but the theme color for text on Light and White themes is not AA or AAA accessible.

Results here:
White https://webaim.org/resources/contrastchecker/?fcolor=00AEEF&bcolor=FFFFFF
Light https://webaim.org/resources/contrastchecker/?fcolor=00AEEF&bcolor=E6E6E6
Dark https://webaim.org/resources/contrastchecker/?fcolor=00AEEF&bcolor=191919
Black https://webaim.org/resources/contrastchecker/?fcolor=00AEEF&bcolor=000000

TranRed: I believe the entire white/light theme family is a bit (i.e. very) orphaned.

Y0dax:
Ready for review again. See last screenshot for latest look.

pajlada (approved):
:+1: Thank you for your contribution!

Issue #2058, opened 2020-10-10
Add right click "search with google" action


RAnders00:
**What should be added?**
When right-clicking on some selected text, there should be a "Search with google..." option that opens the browser with a google search for the selected phrase.

**Why should it be added?**
Can be useful to quickly google things other chatters are talking about.

This is also a feature in google chrome and other modern browsers, I personally use it quite often in the browser and I imagine it could be similarly useful in Chatterino. I've also seen "Search with google..." being an option in the Intellij IDEA console where it can be used to quickly google for error messages etc.

TranRed:
maybe we could even offer a advanced setting for alternate search engines for people who dislike google

leon-richardt:
> advanced setting for alternate search engines

I suggest such setting to be located on the "External Tools" settings page.

Y0dax:
What alternate search engines do we think make sense here? 

From popularity: 
Bing?
Yahoo?

leon-richardt:
> What alternate search engines do we think make sense here?

I believe the idea here is to support custom query schemas, similar to what to the "Custom stream player" and "Image Uploader" section in the "External tools" settings.

TranRed:
I was thinking about duckduckgo. I think this is popular with the "no google"-demographic

TranRed:
Also was Leon said.

ALazyMeme:
I think a better solution would to somehow have the ability to use commands in a right click function on a message. From there, a user can make a `/openurl` command and use whatever search engine they want.

remd:
#2067 provided an implementation of this feature but was closed until discussion on this issue was finished, however discussion seems to have stalled out without any significant contention between participants.

I humbly ask that #2067 be resurrected and the discussion here continues as needed.

ALazyMeme:
> #2067 provided an implementation of this feature but was closed until discussion on this issue was finished, however discussion seems to have stalled out without any significant contention between participants.
> 
> I humbly ask that #2067 be resurrected and the discussion here continues as needed.

What other discussion would you like? Seems like my suggestion would offer the right amount of customization, but wouldn't specifically fall under this issue.

RAnders00:
I don‘t like alazymeme‘s suggestion. Seems like using commands like that would result in a  bad user experience for a simple feature that would offer a better user experience if implemented explicitly. (It would allow the UI to be custom-fit for the task for example, instead of repurposing the commands for that)

LosFarmosCTL:
[...]

PR #2067, opened 2020-10-12
Search with google feature

Y0dax:
Pull request checklist:

# Description

Added a basic search with google feature for selected text in channel messages. (#2058)

There is ongoing discussion regarding allowing different search engines, but I figured it would be good to get the basic feature out there which can be enhanced later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added search with google feature (#2058)
 - Major: Added Streamer Mode configuration (under `Settings -> General`), where you can select which features of Chatterino should behave differently when you are in Streamer Mode. (#2001)
 - Minor: Made the current channels emotes appear at the top of the emote picker popup. (#2057)
 - Minor: Added viewer list button to twitch channel header. (#1978)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added search with google feature (#2058)
```
leon-richardt, line 5: Please place "Minor" entries in the changelog below the "Major" entries.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1706,11 +1706,18 @@ void ChannelView::addContextMenuItems(
...
     if (!this->selection_.isEmpty())
     {
         menu->addAction("Copy selection",
                         [this] { crossPlatformCopy(this->getSelectedText()); });
+
+        menu->addAction("Search with Google", [this] {
+            auto query =
+                QString(QUrl::toPercentEncoding(this->getSelectedText()));
+            QDesktopServices::openUrl(
+                QUrl("https://google.com/search?q=" + query));
```
leon-richardt, line 1719: I do not like this string being hardcoded here. Consider extracting the opening of the link into an own class, especially in regards to custom search query schemas potentially being added later on.

leon-richardt (changes requested):
Important note: Should probably wait for @fourtf or @pajlada to express their opinion on #2058 before investing too much time into implementation.

Y0dax:
Closing while discussion is ongoing.

PR #2073, opened 2020-10-13
Fix: ColorButtons in settings don't respect search

leon-richardt:
Pull request checklist:

# Description

Prior to this commit, adding a `ColorButton` to a `SettingsLayout` via `SettingsLayout::addColorButton` lead to the button not respecting search queries. This is because they are not added to the layout in a `Group`.

This commit fixes the behavior, causing color buttons to behave like every other settings widget.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@
 - Bugfix: Fixed timestamps being incorrect on some messages loaded from the recent-messages service on startup (#1286, #2020)
 - Bugfix: Fixed timestamps missing on channel point redemption messages (#1943)
 - Bugfix: Fixed tooltip didn't show in `EmotePopup` depending on the `Link preview` setting enabled or no (#2008)
-- Bugfix: Fixed "Last message line color" button not respecting search queries (#2073)
 
 ## 2.2.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - Bugfix: Fixed timestamps being incorrect on some messages loaded from the recent-messages service on startup (#1286, #2020)
 - Bugfix: Fixed timestamps missing on channel point redemption messages (#1943)
 - Bugfix: Fixed tooltip didn't show in `EmotePopup` depending on the `Link preview` setting enabled or no (#2008)
+- Bugfix: Fixed "Last message line color" button not respecting search queries (#2073)
```
pajlada, line 32: No changelog entry is needed if the fix happened to a feature that was added in the same changelog section, you could if anything edit the original feature changelog entry and add this pr number to it

pajlada (changes requested):

CHANGELOG.md
leon-richardt: Right, my bad. Didn't realize it was unversioned. Reverted in d188eed9.

pajlada (approved):

PR #2084, opened 2020-10-18
Wait for smooth scrolling animation to add new message

Pizzabelly:
Pull request checklist:

# Description

This fixes a bug in smooth scrolling where scrolling at the same time as a new message was added would cause a visual glitch that looks like the text jumps up and back instantly. To test it you can go in a fast chat and once the message queue fill up, and it starts scrolling to keep your place in the chat, you can scroll up and down to see it.

I'm not sure if this is a good fix because I couldn't fully understand the smooth scrolling code, and this seems like a workaround. So if there is a better way to fix this I'll just close this.


alazymeme:
Can we fix the merge conflicts so this can be reviewed and hopefully merged soon?

zneix (approved):
When it comes to fixing the visual bug, the PR does the job. Not entirely sure about the code, but that looks good to me. @pajlada could you have another look at this?

pajlada:
This seems like a fair fix for me.

Basically 'pausing' all message additions during the 'smooth scrolling' effect.

pajlada (approved):

fourtf:
@Pizzabelly could you explain how the other event loop works and if there is race conditions that can happen?

Issue #1089, opened 2019-06-02
Can't do moderation actions in the logpopup window


VulpesHD:
There are no moderation buttons when I open up someone's popup window in the search tab.
Usercards or online logs do not work here either.

![Screenshot_15](https://user-images.githubusercontent.com/28871321/58767548-09a1ea80-858d-11e9-8205-ceabcedd64e6.jpg)

TranRed:
kinda duplicate of #1065

fourtf:
duplicate of https://github.com/Chatterino/chatterino2/issues/550

Issue #550, opened 2018-07-01
Channel-specific user info popup actions don't work in special channels


DatGuy1:

pajlada:
For `/mentions`, I think the message should be able to be linked to the channel, so I can time people out from the split. Same thing with the Search window, if the messages are linked to a channel you can have moderation actions in there

fourtf:
Edited to give a choice between fixing it or removing it.

zneix:
This has been fixed in #2088, but do we care about showing channel actions when opening a usercard from channels such as whispers or mentions? @fourtf what's your input on this?

jammehcow:
Also resolved in #3623, can we close this? I don't see anything else to be implemented.

Felanbird:
Only thing I can think of would be whispers actually appearing in the usercard popup, but I assume that would require a full-on refactor from IRC whispers to pubsub whispers, since channel context in `/whispers` would just be `#jtv`.

Mm2PL:
For `Usercard` button in `/whispers` and `/live`, I think we should just hide the button. I can't see any other things to do in this area.

PR #2088, opened 2020-10-19
Retain channel context for user cards opened from search window

Y0dax:
Pull request checklist:

# Description

This change fixes the issue in which channel actions such as moderator buttons were not present on users cards that were opened from the search popup.

This closes (#1089) which was marked as duplicate, but only partially resolves the referred issue (#550) since this will not fix channel actions when viewing a user from channels such as whispers or mentions.

Setting the source channel allows this logic to use the source channel rather than the current channel (which is an empty channel of type none for the search window) when opening the user info.
```
void ChannelView::showUserInfoPopup(const QString &userName)
{
    auto *userPopup = new UserInfoPopup(getSettings()->autoCloseUserPopup);
    userPopup->setData(userName, this->hasSourceChannel() ? this->sourceChannel_
                                                          : this->channel_);
    QPoint offset(int(150 * this->scale()), int(70 * this->scale()));
    userPopup->move(QCursor::pos() - offset);
    userPopup->show();
}
```

image


pajlada (approved):
:+1: Looks good to me, thanks for your contribution!

PR #2089, opened 2020-10-19
Add parent-window option

dnsge:
Pull request checklist:

~~`CHANGELOG.md` was updated, if applicable~~

# Description

Fixes the "unknown option 'parent-window'" error that Windows users experience when using the browser extension by simply ignoring the command line option.

Fixes Chatterino/chatterino-browser-ext/issues/41


pajlada (approved):
Solved in acec41685ed589bc661d0df68b3fcb18bb8319c2 already by doing error-handing ourselves since it doesn't fit for us to not start if Chrome starts sending another parameter, however it's also not wrong to accept the option if we can maybe us it in the future

PR #2108, opened 2020-10-21
Fix link preview doesn't update link info for previous messages after setting is enabled

### LINKED ISSUE #2054: Link preview always shows "link was not loaded" on hover over previous messages
**Describe the bug**
Link preview always shows "No link info loaded" on hover over previous messages, when you set `Link information` setting to enabled.

**To reproduce**

1. Start app with `Link information` setting disabled
2. Post message with a link (or recent-messages would do too)
3. Enable setting, and it will always show "No link info loaded" on hover.

**Chatterino version**
Chatterino 2.2.2 (commit 84613187e)

It happens because link information is fetched in `TwitchMessageBuilder.cpp` only once ( on message build :) )
https://github.com/Chatterino/chatterino2/blob/84613187e1bda102ac72a207dfdf9c51ea416f80/src/providers/twitch/TwitchMessageBuilder.cpp#L455
And setting is checked here
https://github.com/Chatterino/chatterino2/blob/8532c6d3bca4e6c3e2da279df1224ebc489c2dc2/src/providers/LinkResolver.cpp#L18-L22
And it never gets updated afterwards. So you need to restart the app to get previous links resolved.

fanway:
Pull request checklist:

# Description
Fixes #2054
 
Now, on hover, we try to update link info, if previously it wasn't loaded because of the setting


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 - Bugfix: Fixed timestamps missing on channel point redemption messages (#1943)
 - Bugfix: Fixed tooltip didn't show in `EmotePopup` depending on the `Link preview` setting enabled or no (#2008)
 - Bugfix: Fixed Stream thumbnail not updating after using the "Change channel" feature (#2074, #2080)
+- Bugfix: Fixed previous link info weren't updated after `Link information` setting is enabled (#2054)
```
Felanbird, line 37:
```suggestion
- Bugfix: Fixed previous link info not updating after `Link information` setting is enabled (#2054)
```
tiny wording fix on changelog entry

Felanbird (changes requested):

PR #2126, opened 2020-10-25
Fixed contributor avatar links

zneix:
Pull request checklist:

- ~~[ ] `CHANGELOG.md` was updated, if applicable~~
not applicable

# Description

﻿Had to put them in qtresource format so those will actually display in Settings -> About ([Reference](https://doc.qt.io/qt-5/resources.html#compiled-in-resources)).

Question is, can I also put my GitHub avatar (and preferrably avatars of others) to resources folder? 🙂


pajlada:
seems good to me, and you're fine to add your own avatar to the list :+1: let others add their own avatar if they wish now that we have the established format.

if you could also add some instructions regarding avatar adding to the same document (e.g. size limitations, use pngcrush on the avatar etc)

pajlada (approved):

Issue #870, opened 2018-11-10
Show timeout timer


TETYYS:
If you're timed out, show timeout timer in the input box, preferably next to the label of current character count.

This could also include a slowmode timer.

gempir:
Insperation for timer 
![image](https://user-images.githubusercontent.com/9765622/86518421-4f854e00-be32-11ea-81df-315ec4430de7.png)

gempir:
Would be interesting how this could look when hiding input box. Personally I think it might be cool to just have a progress bar then at the bottom of the chat.

fourtf:
This issue is out of scope of the Chatterino project.

x9136:
Yo!

There are chats in which timeouts are given no longer than half an hour. Also, there are chats where you can fight with each other through the bots and the loser gets a timeout for 10-20 mins. It would be useful to be notified when you are allowed to send messages after a timeout. The notifications should work the same way as when your nickname is highlighted.

I might consider such a feature to be out of scope of the Chatterino as well?

PR #2143, opened 2020-10-31
Added placeholder for Split's textEdit

zneix:
Pull request checklist:

# Description
Added a placeholder for textEdit split which currently looks like this:

https://cdn.zneix.eu/FmE1NIn.png
farming commits xdddddddddd


zneix:
Just had a random thoguht - since we already connect to a account change signal, we might as well try connecting to other events if needed and try implementing something like request from #870

src/widgets/splits/Split.hpp
```diff
@@ -118,6 +120,7 @@ class Split : public BaseWidget, pajlada::Signals::SignalHolder
     pajlada::Signals::Connection roomModeChangedConnection_;
 
     pajlada::Signals::Connection indirectChannelChangedConnection_;
+    pajlada::Signals::SignalHolder SignalHolder_;
```
pajlada, line 123: should be `signalHolder_` not `SignalHolder_`

pajlada (changes requested):

src/widgets/splits/Split.hpp
zneix: Fixed in 7ee5716b

zneix: Fixed in 7ee5716b

pajlada (approved):

PR #2149, opened 2020-11-01
Improved textEdit placeholder

zneix:
Pull request checklist:

- ~~[ ] `CHANGELOG.md` was updated, if applicable~~
not applicable

# Description

I applied some suggestions about placeholder text being a bit too 'easily readable' and made it slightly lighter to not mix with color of system messages and message timestamps.
I also read that tooltip might've been causing some issues on non-twitch channels, so I removed it (though, it only makes sense in twitch channels anyway).

### DIFF
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -283,6 +283,11 @@ void Split::setContainer(SplitContainer *container)
 
 void Split::onAccountSelected()
 {
+    if (!this->getChannel()->isTwitchChannel())
+    {
+        return;
+    }
+
     auto user = getApp()->accounts->twitch.getCurrent();
     QString placeholderText;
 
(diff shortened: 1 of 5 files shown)


src/widgets/splits/Split.cpp
```diff
@@ -298,6 +303,19 @@ void Split::onAccountSelected()
     }
 
     this->input_->ui_.textEdit->setPlaceholderText(placeholderText);
+
+    this->updateTooltipColor();
+    this->signalHolder_.managedConnect(this->theme->updated, [this]() {
+        this->updateTooltipColor();  //
```
23rd, line 309:
```suggestion
        this->updateTooltipColor();
```

pajlada: Single line lambxas will be disabled soon tm so the code is good as is

pajlada (changes requested):
Could you update the existing changelog entry to include this pr number?

zneix:
> Could you update the existing changelog entry to include this pr number?

Sure thing, resolved in e99b353d

pajlada (approved):

PR #2167, opened 2020-11-08
Chore: Proper Lambda Formatting

leon-richardt:
Pull request checklist:

- [x] ~`CHANGELOG.md` was updated, if applicable~ (not applicable)

# Description

Introduce a clang-format rule to prevent merging lamdas into single lines. Also remove trailing `//` instances where sensible now.

Note that the automatic re-running of clang-format (required for checks to pass) changed lines all over the place which messes with `git blame` history.


pajlada (approved):
:+1:

PR #2168, opened 2020-11-09
Fixes the dropdown order in channel filter dialog

dnsge:
# Description

Undo's the order change in aa4326f which broke the value dropdown


pajlada (approved):

pajlada:
For future person, might be worth adding a comment to this pointing to the other map/enum/struct or w/e that would need to be updated in unison if we want to reorder this

Issue #2173, opened 2020-11-10
Crash when clicking at an username in the usercard message history


fanway:
**Describe the bug**
Chatterino crashes when clicking at an username in the usercard message history

**To reproduce**
- In a chat click at any username (open usercard)
- In the usercard history click at any username

**Chatterino version**
Chatterino 2.2.3-beta (commit 0eed45ae6)

**Operating system**
macOS Catalina 10.15.7 (19H15)

**Additional information**
gdb shows what crash happens here:
https://github.com/Chatterino/chatterino2/blob/0eed45ae67b67b5202c9c1273435e299c50f69de/src/widgets/BaseWidget.cpp#L50

After a couple of compilations, i found out that bug was introduced in 1ee1e8837f82cafb56fbffa1e4fe83dc2e41bced (523874dc210e41640986299d0b07c0fa6c6e76de works as expected)

pajlada:
Confirmed crashed on my macOS laptop too

Could you try my branch out over at https://github.com/Chatterino/chatterino2/pull/2185 and comment there if it fixes your issue?

fanway:
Sure

fanway:
Fixed by #2185

PR #2185, opened 2020-11-15
Use the main window as the parent to all user cards on macOS

pajlada:
Pull request checklist:

# Description


fanway:
Fixes #2173 for me

PR #2189, opened 2020-11-16
Use display/localized names in tab and split titles

### LINKED ISSUE #2172: Use localized/display names in tab and split titles
**What should be added?**
Use localized/display names in the tabs and split titles. Something like that:

**Why should it be added?**
Twitch supports and uses it, so i think it would be nice to have it in chatterino too.

fanway:
Pull request checklist:

# Description
Closes #2172
Basically the title. Issue have screenshots.


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 - Minor: Allow highlights to be excluded from `/mentions`. Excluded highlights will not trigger tab highlights either. (#1793, #2036)
 - Minor: Flag all popup dialogs as actual dialogs so they get the relevant window manager hints (#1843, #2182, #2185)
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
+- Minor: Now tab and split titles use display/localized channel names (#2189)
```
Felanbird, line 34:
```suggestion
- Minor: Tab and split titles now use display/localized channel names (#2189)
```
Sounds a little bit better

Felanbird (changes requested):

src/widgets/splits/Split.cpp
```diff
@@ -377,6 +377,10 @@ void Split::setChannel(IndirectChannel newChannel)
         this->header_->setViewersButtonVisible(false);
     }
 
+    this->channel_.get()->displayNameChanged.connect([this]() {
```
23rd, line 380:
```suggestion
    this->channel_.get()->displayNameChanged.connect([this] {
```

zneix:
This doesn't sound like a good change to me. There are some users, who set their display names to non-alpha characters while they are more recognized under their login names and have their specific nicknames mean something. Seeing those non-alphanumeric characters in tabs without any reference to what can that mean in alphanumeric might confuse many users (including myself) when you have 2 or more of those tabs next to each other.

fanway:
I was referring to how twitch-web is implemented this (they use only localized names in a stream title), and i think it's a proper use case. But, @zneix  do you think that it would be better to handle them as they do in chat right now (localized name + in the brackets a login name)?

zneix:
>  do you think that it would be better to handle them as they do in chat right now (localized name + in the brackets a login name)?

Now this sounds better, having at least some reference to original name is nice - format `loginName(displayName)` would do its job I guess.

fanway:
Looking good in tabs, but not so sure with (live) in split title

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -754,6 +776,31 @@ void TwitchChannel::refreshChatters()
...
+            if (QString::compare(user.displayName, channel->getName(),
+                                 Qt::CaseInsensitive) == 0)
+            {
+                channel->setDisplayName(user.displayName);
+                channel->setLocalizedName(user.displayName);
+            }
+            else
+            {
+                channel->setLocalizedName(user.displayName + " (" +
+                                          channel->getName() + ")");
```
zneix, line 797: Could make use of `QString().arg()` in here to make things look more simple. Also, to keep consistency with chat messages, shouldn't we keep format `loginName(localizedName)`?

fanway: > Could make use of [`QString().arg()`](https://doc.qt.io/qt-5/qstring.html#arg-12) in here to make things look more simple

Sure
> Also, to keep consistency with chat messages, shouldn't we keep format loginName(localizedName)?

Yep, forgot that chatterino formats it swapped.

zneix (approved):

Felanbird (approved):

pajlada (approved):

pajlada:
Looks good to me - thanks for your contribution!

PR #2192, opened 2020-11-16
Fix FFZ emote assertion crash

dnsge:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

Apparently, the FFZ API can return `null` values for emote URLs. This PR treats `null` values the same as if they were missing from the JSON response.

See https://github.com/Chatterino/chatterino2/issues/2191#issuecomment-728278070 for more info

Fixes #2191

PR #2194, opened 2020-11-17
Remove assert() to prevent crashes when emote.isString() is false

MelonLord145:


swills:
Duplicate of #2192

MelonLord145:
My apologies for submitting a duplicate, however I'll leave this PR open for now as the other PR does not account for when the value returned is the wrong type, as it only checks if the value is null.

fourtf:
merged the other one

PR #2199, opened 2020-11-17
cleanup output in release builds

swills:
Cleans up debugging output in release builds

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -614,6 +614,7 @@ CONFIG(debug, debug|release) {
     message("Building Chatterino2 DEBUG")
 } else {
     message("Building Chatterino2 RELEASE")
+    DEFINES += QT_NO_DEBUG_OUTPUT
 }
 
 message("Injected git values: $$git_commit ($$git_release) $$git_hash")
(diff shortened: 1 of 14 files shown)


fourtf:
Are any of the `qDebug()` logs useful for running in release mode? Then I would suggest moving them to `qInfo()` and `qWarning()` instead.

swills:
> Are any of the `qDebug()` logs useful for running in release mode? Then I would suggest moving them to `qInfo()` and `qWarning()` instead.

Not that I've seen. Are there any you had in mind?

If category based logging (`qCDebug`, `qCInfo`, `qCWarning`) were used, they could be enabled dynamically at run time. I might look at that, but it's a bigger effort.

fourtf:
> they could be enabled dynamically at run time

I don't think that's necessary but every once in a while I see people looking for logs but most of the things we log right now would probably not be of much use.

pajlada:
@swills I pushed some changes to your branch - could you take a peek and see if those make sense?

If everything looks good, please make the branch up-to-date with master and I'll merge this in :ok_hand:

swills:
Merging done

pajlada:
Closing in favour of #2206

PR #2201, opened 2020-11-17
Add ccache to the build where possible

swills:
Won't affect places where ccache isn't available, but speeds up build where it is.

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -8,6 +8,12 @@
 # from lib/boost.pri
 #  - BOOST_DIRECTORY (C:\local\boost\ by default) (Windows only)
 
+CCACHE_BIN = $$system(which ccache)
+!isEmpty(CCACHE_BIN) {
+  load(ccache)
+  CONFIG+=ccache
+}
+
 MINIMUM_REQUIRED_QT_VERSION = 5.12.0
 
 !versionAtLeast(QT_VERSION, $$MINIMUM_REQUIRED_QT_VERSION) {


fourtf:
I have a couple questions.

1. This caches the compilation. Does this have issues after changing source code where it might become incompatible and requires a full rebuild?
2. How does this speed up the compilation? Build tools normally already have incremental builds.
3. In which situations and by how much does this speed up compilation?
4. Can this also be solved by using tools like ninja instead when moving to cmake.

swills:
> I have a couple questions.
> 
>     1. This caches the compilation. Does this have issues after changing source code where it might become incompatible and requires a full rebuild?
> 
>     2. How does this speed up the compilation? Build tools normally already have incremental builds.

Right, it mostly speeds up development building in certain situations, see:

https://ccache.dev/
https://ortogonal.github.io/ccache-and-qmake-qtcreator/

>     3. In which situations and by how much does this speed up compilation?
> 

For me, when doing a clean build with populated cache, this reduces build from 1:47 to 0:10.

>     4. Can this also be solved by using tools like ninja instead when moving to cmake.

I think build tool doesn't matter here.

fourtf:
i don't see an issue with this, maybe @pajlada has some input as well

pajlada:
@swills do I have to specify some flag for the c cache config to be enabled?

swills:
> @swills do I have to specify some flag for the c cache config to be enabled?

Nope, as long as ccache is installed, qmake handles it for you.

pajlada (approved):
Because of our use of precompiled headers, I had to add this config value to my `ccache.conf` file:
`sloppiness = pch_defines,time_macros` as instructed by https://ccache.dev/manual/3.2.4.html#_precompiled_headers

Afterwards, this worked exactly as expected and I saw the same sort of compilation speed increases. Very nice!

Once this has had master merged into it and CI passes I'm good to merge this into master. Feel free to wait with this until #2206 has been fully merged in if it's easier that way

swills:
Rebased but made a mistake, sorry for the noise...

zneix:
@swills I don't want to be picky or rude, but it'd be nice if you used Pull Request template or at least updated `CHANGELOG.md` with an entry with a prefix `Dev: ` when some things like this get added, because it makes sure discoverability of new features and changes is preserved.

swills:
@zneix ok, will do, sorry.

PR #2203, opened 2020-11-17
check path is not empty

swills:
Fixes a `QDir::mkpath: Empty or null file path` error on startup.


fourtf:
I'd prefer https://doc.qt.io/qt-5/qstring.html#isEmpty

fourtf:
Looks good to me 👍 Thank you for your contribution!

PR #2206, opened 2020-11-19
Switch to QT Category logging

swills:
Instead of #2199 perhaps this would be better. This switches to category based logging:

https://doc.qt.io/qt-5/qloggingcategory.html

and disables the debug level of logging, but it can be turned back on at run time using `QT_LOGGING_RULES`. Along with that, this adds a custom logger for websocketpp so that it logs via Qt.

Unfortunately this is kinda large and messy, let me know if you want me to split it up, squash, etc.


alazymeme:
Please also make a PR to add the new environment variable to the Chatterino wiki: https://github.com/Chatterino/wiki/blob/master/docs/Environment%20Variables.md

pajlada:
@swills I was rude and pushed some code to your repo and performed the merge - at your earliest convenience please let me know if the changes I've made are okay with you and I will merge this into master

swills:
Looks great to me! Thanks!

PR #2208, opened 2020-11-19
Specify minimum Qt version to avoid issues

swills:
Title says it all

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -22,6 +22,11 @@ DEFINES           += AB_CUSTOM_THEME
 DEFINES           += AB_CUSTOM_SETTINGS
 CONFIG            += AB_NOT_STANDALONE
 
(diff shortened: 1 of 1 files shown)


leon-richardt:
Nothing for this PR in particular but posting here since it fits the topic: we should also update the required Qt version in the build guides: [Linux](https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_LINUX.md), [Windows](https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_WINDOWS.md), [Mac](https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_MAC.md), and [FreeBSD](https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_FREEBSD.md)

swills:
#2210 is better, closing

PR #2215, opened 2020-11-22
Add Ctrl+F5 as a shortcut for reconnect()

mammothb:
Pull request checklist:

# Description
Added a keyboard shortcut (Ctrl+F5) for "Reconnect"

### DIFF
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -117,6 +117,9 @@ Split::Split(QWidget *parent)
     // F5: reload emotes
     createShortcut(this, "F5", &Split::reloadChannelAndSubscriberEmotes);
 
+    // CTRL+F5: reconnect
+    createShortcut(this, "CTRL+F5", &Split::reconnect);
+
     // F10
     createShortcut(this, "F10", [] {
         auto *popup = new DebugPopup;
@@ -846,6 +849,11 @@ void Split::reloadChannelAndSubscriberEmotes()
         twitchChannel->refreshBTTVChannelEmotes(true);
         twitchChannel->refreshFFZChannelEmotes(true);
     }
+}
+
+void Split::reconnect()
+{
+    this->getChannel()->reconnect();
 }
 
 void Split::dragEnterEvent(QDragEnterEvent *event)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added a keyboard shortcut (Ctrl+F5) for "Reconnect" (#2215)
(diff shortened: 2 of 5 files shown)


zneix (changes requested):
Please make sure to consistently reflect changes to keybinding in `Settings -> Keybindings`. Source for that page can be found in `KeyboardSettingsPage.cpp`.

mammothb:
Do you mean this [line](https://github.com/Chatterino/chatterino2/blob/b98a0df73ba0e82527f360ae3948f0ac8831d919/src/widgets/settingspages/KeyboardSettingsPage.cpp#L69)? Or do you mean I should change the label to something else?

zneix:
> Do you mean this [line](https://github.com/Chatterino/chatterino2/blob/b98a0df73ba0e82527f360ae3948f0ac8831d919/src/widgets/settingspages/KeyboardSettingsPage.cpp#L69)? 

Oh right, I totally missed that file in there. Apologies for the confusion!



> Or do you mean I should change the label to something else?

If I were to change anything, maybe `Reconnect current channel` would fit a bit better?

mammothb:
> If I were to change anything, maybe `Reconnect current channel` would fit a bit better?

The `reconnect()` function will cause all channels to reconnect so I'm not sure if specifying "current channel" will be an accurate description of the functionality.

zneix:
Oh yes, I believe you're right I just checkedthe source deeper and I mixed that up with regular reconnection. Once again I am sorry for messing up ;/

zneix (approved):

pajlada (commented):
I am hesitant towards adding a hotkey for something this disruptive when the reason for its existence is a bug (Chatterino sometimes not reconnecting on its own). Is this an action you have to take often?

mammothb:
It occurs irregularly but I had to reconnect a couple of times today, which prompted this modification. And as some people I talked to would rather get it from the official build than a custom build from me, I decided to open this PR.

alazymeme:
I think it's a nice addition. I won't personally use it, but I see it being quite useful if I do ever need it.

pajlada (approved):
Approved under the condition that this will not stay forever - will be removed in a future patch soon :tm:

src/widgets/splits/Split.cpp
```diff
@@ -118,6 +118,9 @@ Split::Split(QWidget *parent)
     // F5: reload emotes
     createShortcut(this, "F5", &Split::reloadChannelAndSubscriberEmotes);
 
+    // CTRL+F5: reconnect
+    createShortcut(this, "CTRL+F5", &Split::reconnect);
```
23rd, line 122: Why not just use a lambda?

PR #2220, opened 2020-11-24
Made #channel link in /mentions jump to #channel split

### LINKED ISSUE #2087: When channel name of a mention is clicked, two broken webpages open
**Describe the bug**
Double clicking on the channel name to the left of a message in the /mentions channel opens two web pages. One is just the name of the channel and the other appears to be a guid. Neither are real sites.

**To reproduce**
Double click the channel name to the left of a mention message

**Screenshots**
image

**Chatterino version**
Chatterino Nightly 2.2.2 (commit 48a3506d7, 17.10.2020)

**Operating system**
Windows 10

zneix:
Pull request checklist:

# Description

The `#channel` part in `/mentions` tab is a link and as fourtf once said, it was supposed to take you to the channel and message that you were mentioned in - very similar to how it is done on Discord, but it wasn't working properly and even shown those bugged URL links on some environments (See #2087).

I looked at code from channel switcher and in a pretty similar way added it here. And while channel switcher opens up a new channel when the target channel is closed, I assumed that messages in `/mentions` come always from already open channels and there's no real need to make this function open a new channel because I assume that the split with a channel will still be open.

While switching current SplitContainer and the Split itself works flawlessly, I wasn't really able to implement scrolling to certain message - maybe I can look into that further in the future and send another PR with the solution, but for now this is better than nothing and also fixes an open issue.

<details>
<summary>GIF with demonstration</summary>

</details>

Closes #2087

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -2043,6 +2043,41 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
                                        SettingsDialogPreference::Accounts);
         }
         break;
+        case Link::JumpToChannel: {
+            // Get all currently open pages
+            QSet<SplitContainer *> openPages;
+
+            auto &nb = getApp()->windows->getMainWindow().getNotebook();
+            for (int i = 0; i < nb.getPageCount(); ++i)
+            {
+                openPages.insert(
+                    static_cast<SplitContainer *>(nb.getPageAt(i)));
+            }
+
+            for (auto *sc : openPages)
+            {
+                auto splits = sc->getSplits();
+
+                // Find channel with mention in all opened channels
+                // TODO(zneix): Consider opening a channel if it's closed (?)
+                auto it = std::find_if(
+                    splits.begin(), splits.end(), [link](Split *split) {
+                        return split->getChannel()->getName() == link.value;
+                    });
+
+                if (it != splits.end())
+                {
+                    // Select SplitContainer and Split itself where mention message was sent
+                    // TODO(zneix): Try exploring ways of scrolling to a certain message as well
+                    auto &nb = getApp()->windows->getMainWindow().getNotebook();
+                    nb.select(sc);
+
+                    Split *dankSplit = *it;
+                    sc->setSelected(dankSplit);
+                }
+            }
+        }
+        break;
 
         default:;
     }
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -2054,11 +2054,11 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
                     static_cast<SplitContainer *>(nb.getPageAt(i)));
             }
 
-            for (auto *sc : openPages)
+            for (auto *page : openPages)
             {
-                auto splits = sc->getSplits();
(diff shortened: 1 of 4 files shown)


leon-richardt:
Now that we use this in multiple places, it might make sense to provide the [`openPages()`](https://github.com/Chatterino/chatterino2/blob/05aab868455500ede7cc1d3d3f9a2006392d5f12/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp#L19-L30) functionality in some public interface.

(Not sure if appropriate for this PR but just something to keep in mind.)

zneix:
> Not sure if appropriate for this PR

I think that's appropriate, since I've made use of this method and we might as well refactor that to some publicly available function because I bet it might be used somewhere else as well.
I will try something and if that's not okay, I can always revert it.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2043,6 +2043,41 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
...
+            QSet<SplitContainer *> openPages;
+
+            auto &nb = getApp()->windows->getMainWindow().getNotebook();
+            for (int i = 0; i < nb.getPageCount(); ++i)
+            {
+                openPages.insert(
+                    static_cast<SplitContainer *>(nb.getPageAt(i)));
+            }
+
+            for (auto *sc : openPages)
```
pajlada, line 2057: please rename `sc` to `page`

src/widgets/helper/ChannelView.cpp
```diff
@@ -2043,6 +2043,41 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
...
+                    });
+
+                if (it != splits.end())
+                {
+                    // Select SplitContainer and Split itself where mention message was sent
+                    // TODO(zneix): Try exploring ways of scrolling to a certain message as well
+                    auto &nb = getApp()->windows->getMainWindow().getNotebook();
+                    nb.select(sc);
+
+                    Split *dankSplit = *it;
```
pajlada, line 2075: please rename `dankSplit` to `split`

src/widgets/helper/ChannelView.cpp
```diff
@@ -2043,6 +2043,41 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
...
+            {
+                openPages.insert(
+                    static_cast<SplitContainer *>(nb.getPageAt(i)));
+            }
+
+            for (auto *sc : openPages)
+            {
+                auto splits = sc->getSplits();
+
+                // Find channel with mention in all opened channels
```
pajlada, line 2061: Update comment phrasing a bit, maybe something along the lines of:
`Search for channel matching link in page/split container`

pajlada (changes requested):

src/widgets/helper/ChannelView.cpp
zneix: Updated in 012f98be

src/widgets/helper/ChannelView.cpp
zneix: Renamed in 012f98be

src/widgets/helper/ChannelView.cpp
zneix: Fixed in 012f98be

Issue #2219, opened 2020-11-23
The ignore feature crashes Chatterino.


Sypticle:
**Describe the bug**
Adding something to the ignore list will crash Chatterino after some time (not sure how long it takes but I find it to take around 20 minutes and sometimes quicker when minimized.) 

**To reproduce**
Simply add something to the ignore list, something that might get spammed a lot, not exactly sure if that matters or not but for me it crashes when adding KEKW, and does not matter if it's case sensitive, I would also assume the replacement can be anything and would still reproduce the crash. Again, minimizing the window seems to cause the crash to happen quicker.
(also another bug, the window to choose fonts is stuck behind every other window and cannot be accessed.)

**Chatterino version**
Chatterino 2.2.3-beta2 (commit 6499b7c12) also happens on the latest release.

**Operating system**
Windows 10

pajlada:
Would love help in getting full reproduction steps for this issue (preferably a stack trace)

pajlada:
@Sypticle Could you give the latest nightly a try to see if that fixes your issue?
I managed to find a case where Chatterino would freeze completely (but not crash) and that's been fixed in the latest nightly.

You can find the latest nightly here: https://github.com/Chatterino/chatterino2/releases/tag/github-actions-nightly (no installer unfortunately)

Sypticle:
@pajlada Sorry, I actually meant freezes, it totally went over my head when writing the issue.
But I have tested it for about 2 hours now and have had no issues with it.

alazymeme:
Close it ![PAJJED](https://user-images.githubusercontent.com/12804673/102008800-e90c4080-3d7e-11eb-9564-7e50b2c18372.png)

PR #2231, opened 2020-11-28
Fix freeze in Ignore & Replace if replacement followed by emote

pajlada:
Pull request checklist:

# Description

This fixes an issue that could happen when using the "Ignore & Replace" functionality of Ignore.
If a Twitch emote followed a replaced phrase and the replaced phrase differed in length from the replacement phrase, the Twitch emote position data was not fully updating, causing the Twitch emote searching code to get stuck in an endless loop.


leon-richardt:
Fixes (?) https://github.com/Chatterino/chatterino2/issues/2219.

PR #1843, opened 2020-08-08
Use Qt's dialog where applicable

pajlada:
# Description

This should hopefully fix a lot of issues people run into using tiling window managers.
There are a few more places that should start using BaseWindow with the Dialog flag or BasePopup instead, but that can come later. For now, this is in a testable phase.
The windows that should work better now are:
 * Streamlink quality popup (Can temporarily be accessed by typing /xd in a split)
 * Emote input popup (The one that pops up if you type `:fors` and that suggests `forsenE`)
 * Emote menu (Cccessed by pressing the smiley face in the bottom right of a split or pressing CTRL+E)
 * Channel changer dialog (Accessed by pressing CTRL+R or pressing the slim hamburger menu and clicking "Change Channel")
 * Settings dialog (Accessed by pressing CTRL+P or pressing the cogwheel button)
 * Tutorial videos (Accessed by pressing the slim hamburger menu and clicking either of the options under the "How to..." submenu)
 * License dialog (Accessed by pressing CTRL+P, going to the About side-tab, pressing "show license" on some of the licenses)
 * Color picker dialog (Accessed by pressing CTRL+P, going to the Highlights side-tab, selecting a custom color for one of your highlights)
 * User Info Popup (Accessed by typing `/usercard pajlada` or pressing a username in a split)
 * Debug Popup (Accessed by pressing F10)

Test checklist:
 - [x] Linux w/ i3
 - [x] Linux w/ dwm
 - [x] Linux w/ Sway
 - [x] Linux w/ non-tiling WM (KDE or Gnome)
 - [x] Windows
 - [x] MacOS

Template for testing:
```
Popup                    | master (3ee08b9ff) | PR build | Notes
------------------------ | :----------------: | :------: | ----------
Streamlink quality popup | ❌                 | ✔️
Emote input popup        | ✔️                 | ✔️
Emote menu               | ❌                 | ✔️
Channel changer dialog   | ❌                 | ✔️
Settings dialog          | ❌                 | ✔️
Tutorial videos          | ❌                 | ✔️
License dialog           | ❌                 | ✔️
Color picker dialog      | ❌                 | ✔️
User Info Popup          | ❌                 | ✔️
Debug Popup              | ❌                 | ✔️
```

PR #2234, opened 2020-11-28
fix: make LoginWidget child of AccountsPage

leon-richardt:
Pull request checklist:

# Description

Another followup to #1843, this time regarding the `LoginWidget` spawned when adding a new account.


pajlada (approved):
Works well for me on i3 using Arch Linux btw :+1:

Issue #2226, opened 2020-11-25
Chatterino crashes with SIGSEGV after trying to load 1000 messages from recent-messages API


zneix:
**Describe the bug**
I recently tried selfhosting [recent-messages service](https://github.com/zneix/) and after setting message limit to a "little" higher than original instance, but I encountered crashes. Narrowed it down to big channels with 1000 messages in cache crashing Chatterino. I know that I am first person to ever selfhost mentioned service, so this is not as important and probably nobody is as silly as I am to set message limit to 1000 or higher, but this is something that chatterino should handle on it's side and prevent from SIGSEGV'ing.

**To reproduce**
1. Set up recent-messages service on your own with message high limit (I encountered the issue with 1000).
2. Make sure you have channel that has 1000 messages in cache (like `#forsen`).
3. Run Chatterino with custom Environment: `CHATTERINO2_RECENT_MESSAGES_URL=https://recent-messages.zneix.eu/api/v2/recent-messages/%1`
4. Get an error `fish: “chatterino” terminated by signal SIGSEGV (Address boundary error)`.

**Chatterino version**
Chatterino 2.2.3-beta2 (commit 05aab8684)

**Operating system**
Arch Linux 5.9.10-arch1-1

**Additional information**
[none](https://cdn.zneix.eu/q8Hh5mh.png)

Mm2PL:
dupe or related to #1421 [<img src=https://static-cdn.jtvnw.net/emoticons/v1/300196486/1.0 alt=pajaM>](https://cdn.zneix.eu/q8Hh5mh.png)

zneix:
I am honestly amazed and shocked I opened duplicated issue just after original one had it's anniversary few days ago. [xd](https://cdn.betterttv.net/emote/5e21fed68af14b5f1b43e909/1x)

Issue #1421, opened 2019-11-20
Crash if recent messages endpoint returns 1000 messages


RAnders00:
**Describe the bug**
Chatterino crashes on startup if the recent messages API returns 1000 messages.

**To reproduce**
- set environment variable: `CHATTERINO2_RECENT_MESSAGES_URL=https://shitty-customapis-si5z3vz69j2s.runkit.sh/recent-messages-test-1k/%1`
- start c2

**Screenshots**
n/a

**Chatterino version**
`Chatterino Nightly 2.1.5 (commit d200f12f4, 27.09.2019)`

**Operating system**
Windows 10

**Additional information**

Mm2PL:
Still can reproduce on 97b9bfb2

Felanbird:
Due to the implementation of Twitch Announcements, this issue is more prevalent, as Announcements are rendered as 2 separate messages, so when loading historical message if more than 99 announcements are in said history, more than 1000 messages are technically loaded and Chatterino crashes.

See #3820

pajlada:
https://github.com/Chatterino/chatterino2/pull/3798 is being merged in now, once that's merged in this *should* be fixed

Felanbird:
Looks to be fixed 👍 
![image](https://user-images.githubusercontent.com/41973452/174450131-730868be-91b4-4e71-9ebc-f907e2323afe.png)

PR #2252, opened 2020-12-05
Add setting to limit number of historical messages to load on connect

### LINKED ISSUE #2250: Add setting for limiting how many historical messages to fetch from the recent-messages api
**What should be added?**
The ability to limit how many messages are fetched for each channel from the Recent Messages API. This can be achieved by passing the ?limit query parameter to the recent messages API.

**Why should it be added?**
Performance - in case your computer doesn't like you fetching up-to 800 messages (the default) from the API.

API Docs: https://recent-messages.robotty.de/api

pajlada:
Pull request checklist:

# Description

This pull request adds a setting to limit the amount of historical messages to fetch from the Recent Messages API.

Fixes #2250

### DIFF
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -600,6 +600,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                        s.highlightInlineWhispers);
     layout.addCheckbox("Load message history on connect",
                        s.loadTwitchMessageHistoryOnConnect);
+    // TODO: Change phrasing to use better english once we can tag settings, right now it's kept as history instead of historical so that the setting shows up when the user searches for history
+    layout.addIntInput("Max number of history messages to load on connect",
+                       s.twitchMessageHistoryLimit, 10, 800, 10);
 
     layout.addCheckbox("Enable experimental IRC support (requires restart)",
                        s.enableExperimentalIrc);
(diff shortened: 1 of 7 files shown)


RAnders00:
Downloaded the windows build, works fine for me 👍

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -600,6 +600,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                        s.highlightInlineWhispers);
     layout.addCheckbox("Load message history on connect",
                        s.loadTwitchMessageHistoryOnConnect);
+    // TODO: Change phrasing to use better english once we can tag settings, right now it's kept as history instead of historical so that the setting shows up when the user searches for history
+    layout.addIntInput("Max number of history messages to load on connect",
+                       s.twitchMessageHistoryLimit, 10, 800, 10);
```
revolter, line 605: @pajlada, Why is the maximum limit only `800`?

zneix: IIRC this is how much randers' service returns at max, so there's no need for a higher limit.
Also, loading ~950 or more messages crashes Chatterino, see #2226

killertofus: > 
> 
> IIRC this is how much randers' service returns at max, so there's no need for a higher limit.
> Also, loading ~950 or more messages crashes Chatterino, see #2226

you meant #1421

PR #2266, opened 2020-12-08
Improve the "You are banned" message

### LINKED ISSUE #2110: Improved banned message
**What should be added?**

When a user gets permanently banned from chat, the messages that is shown to them should be something along the lines of 
> You were banned and disconnected from chat, if you get unbanned you still need to reconnect.

or at least convey this information in some way.

**Why should it be added?**

Because when I ban people then unban them (when they are using chatterino), they think that they haven't been unbanned, since twitch now disconnects users that are banned and chatterino never informs them, or tries to reconnect. I also do not want to whisper every new person I do this too. (Because I can't send them from chatterino), ultimately this is a selfish request.

Mm2PL:
Pull request checklist:

# Description

Closes #2110 
## Adds:
  - *You were forced off the channel, try reconnecting* message
  - *You have been banned* message
  - `Link::Reconnect`
## Removes:
  - nothing
## Changes
  - How the *You are banned message* is handled.

## Notes
An unexpected `PART` doesn't exactly mean being banned. If you get banned you get a Pubsub message saying that you were  after the `PART`.

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -700,7 +700,7 @@ std::vector<MessagePtr> IrcMessageHandler::parseNoticeMessage(
                                         .arg(curUser->getUserName());
         const auto loginPromptText = QString(" Try adding your account again.");
 
-        MessageBuilder builder;
+        auto builder = MessageBuilder();
         builder.message().flags.set(MessageFlag::System);
 
         builder.emplace<TimestampElement>();
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -25,7 +25,6 @@ MessagePtr generateBannedMessage(bool confirmedBan)
 {
     const auto linkColor = MessageColor(MessageColor::Link);
     const auto accountsLink = Link(Link::Reconnect, QString());
-    const auto curUser = getApp()->accounts->twitch.getCurrent();
     const auto bannedText =
         confirmedBan
             ? QString("You were banned from this channel!")
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -38,7 +38,7 @@ MessagePtr generateBannedMessage(bool confirmedBan)
                   "If you believe you have been unbanned, try reconnecting.")
             : QString("Try reconnecting.");
 
-    auto builder = MessageBuilder();
+    MessageBuilder builder;
     builder.message().flags.set(MessageFlag::System);
 
     builder.emplace<TimestampElement>();
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -26,9 +26,11 @@ MessagePtr generateBannedMessage(bool confirmedBan)
     const auto linkColor = MessageColor(MessageColor::Link);
     const auto accountsLink = Link(Link::Reconnect, QString());
     const auto curUser = getApp()->accounts->twitch.getCurrent();
-    const auto bannedText = confirmedBan
-                                ? QString("You were banned from this channel!")
-                                : QString("You were forced off this channel.");
+    const auto bannedText =
+        confirmedBan
+            ? QString("You were banned from this channel!")
+            : QString(
+                  "Your connection to this channel was unexpectedly dropped.");
 
     const auto reconnectPromptText =
         confirmedBan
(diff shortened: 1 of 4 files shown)


Mm2PL:
@pajlada, could you please review this?

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -19,6 +19,38 @@
...
+namespace {
+using namespace chatterino;
+MessagePtr generateBannedMessage(bool confirmedBan)
+{
+    const auto linkColor = MessageColor(MessageColor::Link);
+    const auto accountsLink = Link(Link::Reconnect, QString());
+    const auto curUser = getApp()->accounts->twitch.getCurrent();
+    const auto bannedText = confirmedBan
+                                ? QString("You were banned from this channel!")
+                                : QString("You were forced off this channel.");
```
leon-richardt, line 31: I feel like this message on its own does not provide users with enough information about what happened. (Was does "forced off" mean?)

I'm not sure what a better message would be but I think even something like `Your connection to this channel was dropped.` would be preferable. Open to discussion though

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -19,6 +19,38 @@
...
+
+    const auto reconnectPromptText =
+        confirmedBan
+            ? QString(
+                  "If you believe you have been unbanned, try reconnecting.")
+            : QString("Try reconnecting.");
+
+    auto builder = MessageBuilder();
```
leon-richardt, line 39:
```suggestion
    MessageBuilder builder;
```

src/providers/twitch/IrcMessageHandler.cpp
Mm2PL: The message is meant to convey the user's account being forcibly `PART`ed from the channel, *forced off the channel*.

src/providers/twitch/IrcMessageHandler.cpp
Mm2PL: Applied in 1d19dea4

src/providers/twitch/IrcMessageHandler.cpp
leon-richardt: Yes, that makes sense to me and people familiar with TMI. I just think it could lead to confusion for the average user.
Maybe `Your connection to this channel was dropped by Twitch.` or similar could be a viable compromise?

leon-richardt (approved):

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -19,6 +19,40 @@
 
 #include <unordered_set>
 
+namespace {
+using namespace chatterino;
+MessagePtr generateBannedMessage(bool confirmedBan)
+{
+    const auto linkColor = MessageColor(MessageColor::Link);
+    const auto accountsLink = Link(Link::Reconnect, QString());
+    const auto curUser = getApp()->accounts->twitch.getCurrent();
```
zneix, line 28: Unused variable
```suggestion
```

src/providers/twitch/IrcMessageHandler.cpp
zneix: Maybe it's not a good idea to make code reviews at 5am without sleeping because you can't be sure if it's you seeing things or @mm2pl applying suggested change in wrong place (at line 703).

src/widgets/helper/ChannelView.cpp
```diff
@@ -2080,6 +2080,10 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
             }
         }
         break;
+        case Link::Reconnect: {
+            this->underlyingChannel_.get()->reconnect();
```
zneix, line 2084: Can this be used here instead? I did not have a way to test it, but I'm pretty sure it's the same in a nicer form.
```suggestion
            this->channel()->reconnect();
```

zneix (changes requested):

src/widgets/helper/ChannelView.cpp
Mm2PL: When I tested it with `this->channel()`, it didn't work.

src/providers/twitch/IrcMessageHandler.cpp
leon-richardt: Good catch @zneix, you are definitely correct here.

@Mm2PL, could you apply the suggested change in the correct place? (And maybe also revert the change from 1d19dea?)

Mm2PL: will fix when i can

src/widgets/helper/ChannelView.cpp
Mm2PL: `ChannelView::channel_` is different from `ChannelView::underlyingChannel_` and doesn't seem to be connected to the read/write connections.

Mm2PL:
I attempted to fix my own fork and force pushed on the wrong branch :-)

zneix (approved):
I am a really big fan of this kind of link-based solutions, code looks good and I'm looking forward to having this merged :+1:

pajlada (approved):

pajlada:
@Mm2PL Looks good to me! Please get this branch up-to-date with master and I'll merge this in

PR #2271, opened 2020-12-11
Added clip creation support 🎬

zneix:
# Pull request progress checklist

- [x] Clip creation.
- [x] Chat command (e.g. `/clip`).
- [x] Context menu entry / button.
- [x] Keybind for clip creation.
~~- [ ] Handling disabled clips and changing fail message accordingly.~~
impossible unless we will have consistent HTTP status codes in `NetworkResponse`
- [x] Changelog entry.

# Linked pull requests

Requires chatterino/login#8 before merging.

# Description

This pull request adds the possibility to take clips of your favourite god gamers directly from chatterino.
Alongside that, there are some smaller things added, such as `CopyToClipboard` and link and new scope - `clips:edit`.

❗ Before testing this, make sure your Bearer Token has the new required scope.

Examples:

https://cdn.zneix.eu/xa25jfN.png
Success

https://cdn.zneix.eu/lGnwyLB.png
Fail


zneix:
I think I added all kinds of accessibility that all users can use as they want to.
Notes about those features:

1. The keyboard shortcut is copied directly from webchat, but I can change it if desired.

2. The header split option is only visible in channels that are live, it currently looks like this:
https://cdn.zneix.eu/wMfeX5N.png

Though, I've discussed available options and something that gempir proposed sounds also like a good idea to me, I am looking for some feedback on this one:
https://cdn.zneix.eu/6R8qD4Q.png

src/widgets/Window.cpp
```diff
@@ -393,7 +393,7 @@ void Window::addShortcuts()
         getApp()->windows->forceLayoutChannelViews();
     });
 
-    createWindowShortcut(this, "CTRL+K", [this] {
+    createWindowShortcut(this, "CTRL+K", [] {
```
leon-richardt, line 396: This change should probably be in a separate PR?

src/widgets/splits/SplitHeader.cpp
```diff
@@ -418,45 +434,46 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
...
+                             mhAction->setChecked(getSettings()->isMutedChannel(
+                                 this->split_->getChannel()->getName()));
+                         });
+        mhAction->connect(mhAction, &QAction::triggered, this, [this]() {
             getSettings()->toggleMutedChannel(
                 this->split_->getChannel()->getName());
         });
 
-        moreMenu->addAction(action);
+        moreMenu->addAction(mhAction);
```
leon-richardt, line 476: These changes should definitely be in a separate PR

src/widgets/Window.cpp
leon-richardt, line 396: This change should probably be in a separate PR

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -927,6 +927,69 @@ void TwitchChannel::refreshCheerEmotes()
...
+                ->setLink(Link(Link::Url, clip.editUrl));
+            this->addMessage(builder.release());
+        },
+        [this] {
+            const QString failureText =
+                "Failed to create a clip. Since this feature is new "
+                "and it requires extra scopes";
+            const auto loginPromptText =
+                QString(" try adding your account again.");
+            const auto accountsLink = Link(Link::OpenAccountsPage, QString());
```
leon-richardt, line 976: Since these are `const`, we could make them `static` or maybe put them up with other constants [here](https://github.com/Chatterino/chatterino2/blob/3d32055a675448676e11a82da91cbff1e082af3f/src/providers/twitch/TwitchChannel.cpp#L38-L39)

src/providers/twitch/api/Helix.cpp
```diff
@@ -315,6 +315,46 @@ void Helix::getGameById(QString gameId,
...
+            for (const auto &jsonClip : data.toArray())
+            {
+                clips.emplace_back(jsonClip.toObject());
+            }
+
+            successCallback(clips[0]);
+            return Success;
+        })
+        .onError([failureCallback](NetworkResult result) {
+            // clip creation failed LUL
```
leon-richardt, line 349: Pointless comment imo, since the code below it repeats the same information

src/providers/twitch/api/Helix.cpp
```diff
@@ -315,6 +315,46 @@ void Helix::getGameById(QString gameId,
...
+            auto root = result.parseJson();
+            auto data = root.value("data");
+
+            if (!data.isArray())
+            {
+                failureCallback();
+                return Failure;
+            }
+
+            std::vector<HelixClip> clips;
```
leon-richardt, line 338: I'm wondering why this is a vector. Is there a possibility for the endpoint to return more than one clip? (I don't see any mention of this at https://dev.twitch.tv/docs/api/reference#create-clip)

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -927,6 +927,69 @@ void TwitchChannel::refreshCheerEmotes()
...
+            builder.emplace<TextElement>("Clip created!",
+                                         MessageElementFlag::Text,
+                                         MessageColor::System);
+            // clip link
+            builder
+                .emplace<TextElement>("Copy link to clipboard",
+                                      MessageElementFlag::Text,
+                                      MessageColor::Link)
+                ->setLink(Link(Link::CopyToClipboard,
+                               "https://clips.twitch.tv/" + clip.id));
```
leon-richardt, line 958: I would prefer the `"https://clips.twitch.tv/"` to be in a constant or something rather than a magic string

leon-richardt (changes requested):
Really like this feature and the implementation overall, great idea to integrate it the way you did :+1: I do have a few remarks though

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -927,6 +927,69 @@ void TwitchChannel::refreshCheerEmotes()
...
+void TwitchChannel::createClip()
+{
+    if (!this->isLive())
+    {
+        this->addMessage(makeSystemMessage(
+            "Cannot create clip while the channel is offline!"));
+        return;
+    }
+
+    this->addMessage(makeSystemMessage("Creating clip..."));
```
Mm2PL, line 939: In my opinion, this would be the most (second one to this being `Helix::createClip()`) convenient place to add it.

Mm2PL (changes requested):
I overall like the idea of your PR, however it lacks any kind of rate limiting of the clip requests ([they have their own rate limits](https://dev.twitch.tv/docs/api/reference#create-clip)), which means the user can hold Alt+X or repeatedly send `/clip` and that could get Chatterino rate limited. Please add that some kind of rate limiting or allow to only create one clip at a time. For example I used a [`QMutex`](https://github.com/Chatterino/chatterino2/blob/df722a72c18457c71833e03cfad9caccff3cf963/src/util/NuulsUploader.cpp#L43) ([docs](https://doc.qt.io/qt-5.15/qmutex.html)) and [`QTimer::singleShot`](https://github.com/Chatterino/chatterino2/blob/df722a72c18457c71833e03cfad9caccff3cf963/src/util/NuulsUploader.cpp#L193-L196) ([docs](https://doc.qt.io/qt-5.15/qtimer.html)) to achieve this in the image uploader. 

Example of what can happen when the user spams the clip button:
Example of what can happen when the user spams the clip button

src/providers/twitch/api/Helix.cpp
zneix: Yes I agree, I forgot to remove it after debugging, good catch :+1:

src/providers/twitch/api/Helix.cpp
zneix: At least from what docs say, there's no possibility for this endpoint to return more than 1 clip, however due to lack of professional knowledge on how to handle the array I just followed an example from `Helix::fetchUsers`. I Will look into just getting the first element of `data` object from response and return it.

src/providers/twitch/TwitchChannel.cpp
zneix: Sure thing, I didn't think of that. Will change this part :+1:

src/widgets/Window.cpp
zneix: Eh, it was just a very little change that meant to fix an annoying warning from my editor, but sure, I'll revert it.

src/widgets/splits/SplitHeader.cpp
zneix: I had a feeling those changes might've been not welcome, though after changing some related code it annoyed me so much I had to change it. You've got a point thoguh and I will revert those changes :+1:

src/providers/twitch/TwitchChannel.cpp
zneix: I just assumed that this was going to be unchanged on Twitch's side so I allowed myself to "hardcode" this value in the said magic string to make things more simple. Though, I am a fan of the approach you've suggested and I will move that into a constant based on `QString` :+1: thanks for pointing this out

src/widgets/splits/SplitHeader.cpp
zneix: Though, I will keep the declaration of `twitchChannel` as it is useful in the PR and makes other lines of code more readable as well.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -927,6 +927,69 @@ void TwitchChannel::refreshCheerEmotes()
...
+                .emplace<TextElement>("edit it in browser.",
+                                      MessageElementFlag::Text,
+                                      MessageColor::Link)
+                ->setLink(Link(Link::Url, clip.editUrl));
+            this->addMessage(builder.release());
+        },
+        [this] {
+            const QString failureText =
+                "Failed to create a clip. Since this feature is new "
+                "and it requires extra scopes";
```
Felanbird, line 973:
```suggestion
                "Failed to create a clip. Since this feature is new";
```
I'm not 100% on declaring 'new' in the error, but the scopes part is irrelevant to the average user.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added clip creation support. You can create clips with `/clip` command, `Alt+X` keybind or `Create a clip` option in split header's context menu. (#2271)
```
```suggestion
- Major: Added clip creation support. You can create clips with `/clip` command, `Alt+X` keybind or `Create a clip` option in split header's context menu. [This requires a new scope so re-authentication will be required to use it.] (#2271)
```
moved the 'new scope' information to the changelog instead

Felanbird (changes requested):

CHANGELOG.md
zneix: yeah I forgot to include that in the changelog entry, good catch :+1:

src/providers/twitch/TwitchChannel.cpp
zneix: I guess that's fair, makes the error message shorter which I think is good because I also realized I have to include something about streamer having clips disabled, so I'll just edit that myself. Thanks for the input :+1:

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -37,6 +37,13 @@ namespace chatterino {
 namespace {
     constexpr int TITLE_REFRESH_PERIOD = 10;
     constexpr char MAGIC_MESSAGE_SUFFIX[] = u8" \U000E0000";
+    static const QString clipsLink("https://clips.twitch.tv/%1");
+    static const QString failureText =
+        "Failed to create a clip. Since this feature is new "
+        "and it requires extra scopes";
+    static const QString loginPromptText =
+        QString(" try adding your account again.");
+    static const auto accountsLink = Link(Link::OpenAccountsPage, QString());
```
leon-richardt, line 46: Sorry I just noticed this now (and it is kind of nitpicky but I will mention it nonetheless):
- Constants should be set in upper-case according to project conventions
- We use three different kind of initialization here: "canonical" `QString` constructor, implicit `QString` copy constructor and explicit `QString` copy constructor. I would prefer settling for one of these for consistency (I prefer the canonical version, as used for `clipsLink`.

Again, sorry I did not mention this in the first round of review

leon-richardt (changes requested):

src/providers/twitch/TwitchChannel.cpp
zneix: Seems like I yolo copy-pasted the code and messed up kinds of initialization, will fix.

src/providers/twitch/api/Helix.cpp
```diff
@@ -315,6 +315,41 @@ void Helix::getGameById(QString gameId,
...
+                return Failure;
+            }
+
+            HelixClip clip(data.toArray()[0].toObject());
+
+            successCallback(clip);
+            return Success;
+        })
+        .onError([failureCallback](NetworkResult result) {
+            // clip creation failed LUL
```
leon-richardt, line 344: I believe you missed this one when applying suggestions :D

leon-richardt (changes requested):
One more thing and then I'm content :smile:

leon-richardt (approved):

src/providers/twitch/TwitchChannel.cpp
zneix: Thanks for pointing it out and more importantly giving me an excellent example of code in action. I think there's no point in having few clips being created at once, so I guess I will go with the `QMutex`. Thanks again :heart:

Felanbird (approved):

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -986,6 +993,7 @@ void TwitchChannel::createClip()
                 ->setLink(ACCOUNTS_LINK);
 
             this->addMessage(builder.release());
+            clipCreationMutex.unlock();
```
leon-richardt, line 997: One more thing :tm: :

I think we should consider using [QMutexLocker](https://doc.qt.io/qt-5/qmutexlocker.html) here since mutex locking/unlocking is often a source of bug when dealing with different logical branches. In fact, using a QMutexLocker is also recommended in the [QMutex docs](https://doc.qt.io/qt-5/qmutex.html#details).

Furthermore, this seems to be the Qt equivalent of the idiomatic [`std::lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard) from the C++ standard library.

zneix: I didn't get through the whole QMutex's documentation, thus I had no idea about this locker class existance, I will try to redo this :+1:

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -944,6 +945,10 @@ void TwitchChannel::createClip()
         return;
     }
 
+    if (QTime().currentTime().msecsTo(this->timeNextClipCreationAllowed_) > 0)
```
leon-richardt, line 948: Can we use [`QTime::operator<`](https://doc.qt.io/qt-5/qtime.html#operator-lt) instead?

leon-richardt (changes requested):

src/providers/twitch/TwitchChannel.cpp
zneix: I assumed calling provided method is better since it exists just for this, but I guess your solution is cleaner.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -37,6 +37,14 @@ namespace chatterino {
 namespace {
     constexpr int TITLE_REFRESH_PERIOD = 10;
     constexpr char MAGIC_MESSAGE_SUFFIX[] = u8" \U000E0000";
+    static const QString CLIPS_LINK("https://clips.twitch.tv/%1");
+    static const QString FAILURE_TEXT("Failed to create a clip - either the "
+                                      "streamer has clips disabled or you "
+                                      "haven't re-authenticated yet. ");
+    static const QString LOGIN_PROMPT_TEXT(
+        "Click here to add your account again.");
+    static const auto ACCOUNTS_LINK = Link(Link::OpenAccountsPage, QString());
```
leon-richardt, line 46:
```suggestion
    static const Link ACCOUNTS_LINK(Link::OpenAccountsPage, QString());
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -927,6 +935,76 @@ void TwitchChannel::refreshCheerEmotes()
...
+        this->addMessage(makeSystemMessage(
+            "Cannot create clip while the channel is offline!"));
+        return;
+    }
+
+    if (QTime().currentTime() < this->timeNextClipCreationAllowed_ ||
+        this->isClipCreationInProgress)
+    {
+        return;
+    }
```
leon-richardt, line 951: Should we also post a system message here informing the user that a clip creation is currently in progress or that they must wait x more seconds before being able to create a new clip?

leon-richardt (changes requested):

src/providers/twitch/TwitchChannel.cpp
zneix: I didn't want to have unnecessary message while clip creation is in progress since it's rather straight-forward that you can't create other clips in this short period of time.
I'm unsure about cases when you are on "clip cooldown". I thought about maybe adding some information in FAQ about it or in changelog entry.

leon-richardt: Yeah, I'm not sure about this one either 👍 I can agree with both ways of going about it

leon-richardt (approved):

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -37,6 +37,14 @@ namespace chatterino {
...
     constexpr int TITLE_REFRESH_PERIOD = 10;
     constexpr char MAGIC_MESSAGE_SUFFIX[] = u8" \U000E0000";
+    static const QString CLIPS_LINK("https://clips.twitch.tv/%1");
+    static const QString FAILURE_TEXT("Failed to create a clip - either the "
+                                      "streamer has clips disabled or you "
+                                      "haven't re-authenticated yet. ");
+    static const QString LOGIN_PROMPT_TEXT(
+        "Click here to add your account again.");
+    static const Link ACCOUNTS_LINK(Link::OpenAccountsPage, QString());
+    static const int CLIP_CREATION_COOLDOWN = 5000;
```
leon-richardt, line 47: Since these are already in the unnamed namespace, we should be able to drop the `static` keyword here

leon-richardt, line 47: Thought on making this a `constexpr` to keep in line with `TITLE_REFRESH_PERIOD`?

src/widgets/helper/ChannelView.cpp
```diff
@@ -2079,6 +2079,9 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
                 }
             }
         }
+        case Link::CopyToClipboard: {
+            crossPlatformCopy(link.value);
+        }
```
leon-richardt, line 2084: You are missing a `break` statement for the new case. This means that the above case will fall through to this when clicking on a channel name in the mentions split, and that channel is no longer open in any tab.

(In that case, the clicked channel name will be copied to your clipboard.)

```suggestion
        }
        break;
        case Link::CopyToClipboard: {
            crossPlatformCopy(link.value);
        }
```

leon-richardt (changes requested):

src/widgets/helper/ChannelView.cpp
zneix: Looks like I've forgotten about the `break` in that Pull Request. Will add :+1:

zneix:
@fourtf can you review this code / give your input on this pull request? I don't want it to become stale.

src/providers/twitch/api/Helix.cpp
```diff
@@ -390,6 +391,9 @@ void Helix::createClip(QString channelId,
                 << result.getData();
             failureCallback();
         })
+        .finally([finallyCallback] {
+            finallyCallback();
+        })
```
leon-richardt, line 396:
```suggestion
        .finally(finallyCallback)
```

leon-richardt (changes requested):

leon-richardt (approved):

pajlada:
@zneix I snuck in a `HelixClipError` enum that's now a parameter of the `Helix::createClip` `failureCallback` - error code 401 seems to be if the user lacks authentication, and 503 if the streamer doesn't allow clips. I haven't added a message output for the "unknown" error case yet, please go ahead and do so when you've looked through my commit if you agree with the way it's been implemented.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -938,6 +947,100 @@ void TwitchChannel::refreshCheerEmotes()
...
+                        .emplace<TextElement>(LOGIN_PROMPT_TEXT,
+                                              MessageElementFlag::Text,
+                                              MessageColor::Link)
+                        ->setLink(ACCOUNTS_LINK);
+                }
+                break;
+
+                case HelixClipError::Unknown:
+                default: {
+                    // TODO: Implement on unknown error.
```
pajlada, line 1028: Please implement this @zneix

pajlada (changes requested):

pajlada (approved):

PR #2278, opened 2020-12-13
Add Recent Messages privacy disclaimer.

### LINKED ISSUE #1311: Create opt-in system for recent-messages and other third-party services
I'm trying to make my recent_messages service more compliant with the law

A section in the installer where the user is shown
- that Chatterino2 can fetch messages from the recent_messages service
- it will send each channel name you visit to a third-party service for that,
- and that channels you visit will have their messages stored
- that you can opt out of having the bot collect messages in your own channel by using the bot at https://www.twitch.tv/recent_messages
and with a link to the privacy statement (`https://recent-messages.robotty.de/privacy`), and two buttons at the bottom "Accept terms and use service", "Decline terms and don't use service" (or something like that)

Don't do it i guess, but it would be nice, also to keep the users more informed

**Additional context**

Mm2PL:
Pull request checklist:

# Description
Adds a disclaimer about the recent messages service. Closes #1311.

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -148,8 +148,8 @@ int Application::run(QApplication &qtApp)
             "open for the purposes of showing you the message history. You can "
             "learn more <a "
             "href=\"https://recent-messages.robotty.de/\">here</a> or "
-            "<a href=\"https://recent-messages.robotty.de/\">view the privacy "
-            "policy here</a>"
+            "<a href=\"https://recent-messages.robotty.de/privacy\">view the "
+            "privacy policy here</a>"
             "<br/>"
             "Would you like to enable support?",
             QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -145,7 +145,7 @@ int Application::run(QApplication &qtApp)
             "Recent messages is a service which allows you to see "
             "messages from before starting Chatterino. If you enable support "
             "for it Chatterino will send the name of every channel you have "
-            "open for the purposes of showing you the message history. You can "
+            "open to a remote service for the purposes of showing you the message history. You can "
             "learn more <a "
             "href=\"https://recent-messages.robotty.de/\">here</a> or "
             "<a href=\"https://recent-messages.robotty.de/privacy\">view the "
@@ -154,7 +154,8 @@ int Application::run(QApplication &qtApp)
             "Would you like to enable support?",
             QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
 
-        if (reply == QMessageBox::Yes)
+        const auto isConfirmed = (reply == QMessageBox::Yes);
+        if (isConfirmed)
         {
             if (!getSettings()->loadTwitchMessageHistoryOnConnect)
             {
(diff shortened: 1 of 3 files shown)


src/Application.cpp
```diff
@@ -137,6 +137,56 @@ int Application::run(QApplication &qtApp)
...
+        auto reply = QMessageBox::question(
+            &this->windows->getMainWindow(),
+            "Do you want to enable Recent Messages support?",
+            "Recent messages is a service which allows you to see "
+            "messages from before starting Chatterino. If you enable support "
+            "for it Chatterino will send the name of every channel you have "
+            "open for the purposes of showing you the message history. You can "
+            "learn more <a "
+            "href=\"https://recent-messages.robotty.de/\">here</a> or "
+            "<a href=\"https://recent-messages.robotty.de/privacy\">view the "
```
23rd, line 151: What if the `CHATTERINO2_RECENT_MESSAGES_URL` environment variable will be replaced?

src/Application.cpp
```diff
@@ -137,6 +137,56 @@ int Application::run(QApplication &qtApp)
...
+                        }
+                    }
+                }
+            }
+            getSettings()->loadTwitchMessageHistoryOnConnect = true;
+        }
+        else
+        {
+            getSettings()->loadTwitchMessageHistoryOnConnect = false;
+        }
```
23rd, line 186:
```suggestion
        const auto isConfirmed = (reply == QMessageBox::Yes);
        if (isConfirmed)
        {
            if (!getSettings()->loadTwitchMessageHistoryOnConnect)
            {
                if (auto selected = this->windows->getMainWindow()
                                        .getNotebook()
                                        .getSelectedPage())
                {
                    if (auto container =
                            dynamic_cast<SplitContainer *>(selected))
                    {
                        for (auto &&split : container->getSplits())
                        {
                            if (auto channel = split->getChannel();
                                !channel->isEmpty())
                            {
                                channel->addMessage(makeSystemMessage(
                                    "You will need to restart chatterino to "
                                    "see historical messages."));
                            }
                        }
                    }
                }
            }
        }
        getSettings()->loadTwitchMessageHistoryOnConnect = isConfirmed;
```

src/Application.cpp
```diff
@@ -137,6 +137,56 @@ int Application::run(QApplication &qtApp)
...
     this->windows->getMainWindow().show();
+    if (!getSettings()->seenRecentMessagesDisclaimer)
+    {
+        auto reply = QMessageBox::question(
+            &this->windows->getMainWindow(),
+            "Do you want to enable Recent Messages support?",
+            "Recent messages is a service which allows you to see "
+            "messages from before starting Chatterino. If you enable support "
+            "for it Chatterino will send the name of every channel you have "
+            "open for the purposes of showing you the message history. You can "
```
RAnders00, line 148:
```suggestion
            "open to a remote service for the purposes of showing you the message history. You can "
```

src/Application.cpp
Mm2PL: Changed in 78ab484, links are now dynamic based on the URL given from the ENV variable.

leon-richardt:
Should we also mention in the info text that the external service can see your IP? (In distinction to Chatterino API requests where we act as a proxy so external providers cannot see user IPs.)

pajlada:
This sort of messaging should probably be part of the installers instead.

However, a privacy notice of some kind could be part of the About page describing the third party APIs we consume. This PR could be the first third party API to be added to the About page

Mm2PL:
Yes it should be a part of the installer, however the portable versions don't have an installer, so it would show up only for some users, not a good idea. The idea of this popup was to ask for consent for sending channel data.

Mm2PL:
This is getting stale. I might reimplement this some time in the future but it isn't now.

Issue #2163, opened 2020-11-07
Settings option to disable the update arrow


bennyhillthebest:
**What should be added?**
An option to disable the update arrow from appearing when the application is launched.

**Why should it be added?**
Chatterino on my system is built and updated seldomly by looking at new features in the commit log (Debian i686), so firstly i don't think i can simply download and install the new update, secondly it's a bit of nag having the arrow appearing everytime the application is launched. Of course it's a very minor request, so do as you like.

pajlada:
I am working on a PR to "normalize" the update skipping behaviour.

On Windows, you skip any sort of update checks if you have downloaded a Nightly version. I will make this apply to macOS and Linux too. To apply this behaviour, you can add the `nightly` mode to your modes file (the modes file is read from the same folder as the chatterino2 binary is located)

See #2164

bennyhillthebest:
Tried setting `bool isNightly{true};` in Modes.hpp or commenting the 19 to 26 lines of Modes.cpp except for `this->isNightly = true;`, but QtCreator laments that it doesn't recognize `PlaceholderText` at line 316 of Split.cpp. 

My coding ability is very limited so i don't know if it is me or the new committed changes, but i'm very doubtful of the latter, since ubuntu-latest build passed. I think i'm dank.

ZoeS17:
@bennyhillthebest  I think the code in `Modes.cpp` is looking for a raw file named modes, at wherever Chatterino is run from. and looking to see if it contains `nightly` on a single line.

pajlada:
Modes are normally part of a Chatterino installation - for example, for the portable installation there will be a file called modes next to the Chatterino application file indicating to it that it should look for settings in the current folder instead of in the operating systems "AppData" folder.

To forcefully add/remove a mode, you would need create (or edit if the file already exists) the `modes` file in the same folder as the Chatterino application file exists. For example, this could be in `/usr/bin` or `/usr/local/bin` if you have installed chatterino using `make install`. What you would do then is `sudo vim /usr/local/modes` and add the line `nightly` to the file.

The reason this process seems a bit weird is because this is kind of a misuse of the modes functionality - it's really there for development-like flags but I think in your case that fits.

bennyhillthebest:
[...]

PR #2284, opened 2020-12-15
Color mentions to match the mentioned users color

### LINKED ISSUE #1963: Change the color of names in messages to match the users name color
Add an option to or just change the color of names in messages to match the users name color.

This should be added because it's a feature in web chat(not sure if its from ffz or bttv thought). It also makes it much easier to tell  when a message is directed towards a specific user rather than just in chat.

### LINKED ISSUE #2121: An option to make @mentions the same color as the mentioned person's username
**What should be added?**
When you @ someone, there should be an option to make their username in your message the same color as their name. For example, if my username is bob, and someone @'d me in chat. The message should read: "@bob [this part in my color] hey! [this part in the normal text color]"

**Why should it be added?**
There's an option like this in the FrankerFaceZ Control Center on Twitch, and it makes it a lot easier to find the people who get @'d.

kiwec:
Pull request checklist:

# Description

2020-12-15-223048_656x354_scrot

Fixes #1963
Fixes #2121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
 - Minor: Tab and split titles now use display/localized channel names (#2189)
 - Minor: Add a setting to limit the amount of historical messages loaded from the Recent Messages API (#2250, #2252)
-- Minor: Color mentions to match the mentioned users (#1963, #2121)
+- Minor: Color mentions to match the mentioned users (#1963, #2121, #2284)
 - Bugfix: Fix crash occurring when pressing Escape in the Color Picker Dialog (#1843)
 - Bugfix: Fix bug where the "check user follow state" event could trigger a network request requesting the user to follow or unfollow a user. By itself its quite harmless as it just repeats to Twitch the same follow state we had, so no follows should have been lost by this but it meant there was a rogue network request that was fired that could cause a crash (#1906)
 - Bugfix: /usercard command will now respect the "Automatically close user popup" setting (#1918)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Major: Added "Channel Filters". See https://wiki.chatterino.com/Filters/ for how they work or how to configure them. (#1748, #2083, #2090, #2200)
 - Major: Added Streamer Mode configuration (under `Settings -> General`), where you can select which features of Chatterino should behave differently when you are in Streamer Mode. (#2001)
-- Major: Color mentions to match the mentioned users. You can disable this by unchecking "Color usernames" under `Settings -> General -> Advanced (misc.)`. (#1963, #2284)
+- Major: Color mentions to match the mentioned users. You can disable this by unchecking "Color @usernames" under `Settings -> General -> Advanced (misc.)`. (#1963, #2284)
 - Minor: Made BetterTTV emote tooltips use authors' display name. (#2267)
 - Minor: Added Ctrl + 1/2/3/... and Ctrl+9 shortcuts to Emote Popup (activated with Ctrl+E). They work exactly the same as shortcuts in main window. (#2263)
 - Minor: Added reconnect link to the "You are banned" message. (#2266)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
 - Minor: Tab and split titles now use display/localized channel names (#2189)
 - Minor: Add a setting to limit the amount of historical messages loaded from the Recent Messages API (#2250, #2252)
+- Minor: Color mentions to match the mentioned users (#1963, #2121)
 - Bugfix: Fix crash occurring when pressing Escape in the Color Picker Dialog (#1843)
 - Bugfix: Fix bug where the "check user follow state" event could trigger a network request requesting the user to follow or unfollow a user. By itself its quite harmless as it just repeats to Twitch the same follow state we had, so no follows should have been lost by this but it meant there was a rogue network request that was fired that could cause a crash (#1906)
 - Bugfix: /usercard command will now respect the "Automatically close user popup" setting (#1918)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Major: Added "Channel Filters". See https://wiki.chatterino.com/Filters/ for how they work or how to configure them. (#1748, #2083, #2090, #2200)
 - Major: Added Streamer Mode configuration (under `Settings -> General`), where you can select which features of Chatterino should behave differently when you are in Streamer Mode. (#2001)
+- Major: Color mentions to match the mentioned users. You can disable this by unchecking "Color usernames" under `Settings -> General -> Advanced (misc.)`. (#1963, #2121, #2284)
 - Minor: Made BetterTTV emote tooltips use authors' display name. (#2267)
 - Minor: Added Ctrl + 1/2/3/... and Ctrl+9 shortcuts to Emote Popup (activated with Ctrl+E). They work exactly the same as shortcuts in main window. (#2263)
 - Minor: Added reconnect link to the "You are banned" message. (#2266)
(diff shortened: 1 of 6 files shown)


TranRed:
should there be a setting for it or do we force this? I'd personally prefer not to use it

leon-richardt:
> should there be a setting for it or do we force this? I'd personally prefer not to use it

I agree that this is a very "opinionated" change, it adds too much visual "clutter" for my liking. I believe there should definitely be a setting for this.

kiwec:
I should have explained that this change is not for visual fluff. I added it because it is necessary to be able to follow conversations in any reasonably large communities.

alazymeme:
> I should have explained that this change is not for visual fluff. I added it because it is necessary to be able to follow conversations in any reasonably large communities.

What yunglpr and tranred said is still applicable. Features that certain people want, others may not want. It should be an option, not a forced feature.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
 - Minor: Tab and split titles now use display/localized channel names (#2189)
 - Minor: Add a setting to limit the amount of historical messages loaded from the Recent Messages API (#2250, #2252)
+- Minor: Color mentions to match the mentioned users (#1963, #2121)
```
leon-richardt, line 43: Please also add the number of this PR (#2284) to the changelog entry

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
         });
     }
 }
+
+const QColor ChannelChatters::getUserColor(const QString &user)
```
leon-richardt, line 68: Unless I'm missing something here, I think we should prevent unnecessary copies and return a `const QColor &` instead. The function can be `const` as well

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
...
+        return search->second;
+    }
+}
+
 void ChannelChatters::setChatters(UsernameSet &&set)
 {
     *this->chatters_.access() = set;
 }
 
+void ChannelChatters::setUserColor(const QString &user, const QColor color)
```
leon-richardt, line 88: Same thing applies here, we could save a copy using `const QColor &color`

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
...
+{
+    auto chatterColors = this->chatterColors_.access();
+    chatterColors->emplace(user.toLower(), color);
```
leon-richardt, line 91: When a user changes their user name color, this change will not be respected by `chatterColors`.
[`std::map::emplace` only inserts the newly constructed element if there is no element with the key in the container.](https://en.cppreference.com/w/cpp/container/map/emplace)

(It's a niche scenario but I think we should still put thought into it.)

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -499,12 +500,14 @@ void TwitchMessageBuilder::addTextOrEmoji(const QString &string_)
 
         if (match.hasMatch() && chatters->contains(username))
         {
+            auto userColor = this->twitchChannel->getUserColor(username);
```
leon-richardt, line 503: nit: can be `const`

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
         });
     }
 }
+
+const QColor ChannelChatters::getUserColor(const QString &user)
+{
+    auto chatterColors = this->chatterColors_.access();
```
leon-richardt, line 70: nit: can be `const`

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
...
+
+    auto search = chatterColors->find(user.toLower());
```
leon-richardt, line 72: nit: can be `const`

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -479,13 +479,14 @@ void TwitchMessageBuilder::addTextOrEmoji(const QString &string_)
         if (match.hasMatch())
         {
             QString username = match.captured(1);
+            auto userColor = this->twitchChannel->getUserColor(username);
```
leon-richardt, line 482: nit: can be `const`

leon-richardt (changes requested):
I have some comments regardless of whether or not we eventually want a setting for this :smile:

kiwec:
Thanks for the feedback. I added a setting and fixed the issues pointed out by leon.

src/common/ChannelChatters.cpp
kiwec: I don't know the threading model that is used in chatterino, but since the color map is behind a mutex, I'll leave it as is.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 - Minor: Don't show update button for nightly builds on macOS and Linux, this was already the case for Windows (#2163, #2164)
 - Minor: Tab and split titles now use display/localized channel names (#2189)
 - Minor: Add a setting to limit the amount of historical messages loaded from the Recent Messages API (#2250, #2252)
+- Minor: Color mentions to match the mentioned users (#1963, #2121, #2284)
```
zneix, line 43: We don't include issue numbers in changelog entries. Pharsing seems a bit weird to me, I hope this makes sense.
```suggestion
- Minor: Added a setting for @mentions to match the mentioned users' chat color (#2284)
```

src/common/ChannelChatters.hpp
```diff
@@ -17,13 +17,16 @@ class ChannelChatters
     void addRecentChatter(const QString &user);
     void addJoinedUser(const QString &user);
     void addPartedUser(const QString &user);
+    const QColor getUserColor(const QString &user);
     void setChatters(UsernameSet &&set);
+    void setUserColor(const QString &user, const QColor &color);
```
zneix, line 22: Follow-up to comment above - `getUserColor` could be moved right above `setUserColor` to keep things in logical order.

src/common/ChannelChatters.cpp
```diff
@@ -64,9 +64,31 @@ void ChannelChatters::addPartedUser(const QString &user)
...
+    if (search == chatterColors->end())
+    {
+        return QColor(MessageColor::Text);
+    }
+    else
+    {
+        return search->second;
+    }
+}
+
```
zneix, line 82: This function could be moved right above ChannelChatters::setUserColor to keep things in logical order.

zneix (changes requested):
I like this feature and I think it's a nice addition overall :+1: Sorry for being a little pedantic over the code though :stuck_out_tongue:

CHANGELOG.md
leon-richardt: This not necessarily true. (For example, we also link #2163 or #2250)

pajlada: We do a bit of both - this is fine imo.

I basically see it as a list of relevant PRs or Issues to the change.

pajlada, line 43: I would classify this as a major change, please also include information that this feature is enabled by default and can be disabled in the Settings window through the "Color @usernames" setting

pajlada:
Looks good - the unresolved comments are still changes I'd like to see - after that I'll give it a whirl and we can get this merged in! Thanks a lot for contributing!

src/common/ChannelChatters.cpp
leon-richardt: Thanks for getting back to the suggestions so quickly.

Would you mind explaining what your concern in this particular case is?
(Not saying there is none by the way :smile: Interested in learning something though)

kiwec:
Pushed the requested changes. :slightly_smiling_face: 

> 2. What will happen when I send something like `@asdasdasdasdasdasdasdxd`? Chatterino will resolve this to a valid mention, but obviously such user doesn't exist - how would this feature behave in such cases?

The default text color will be used (`MessageColor::Text`).

pajlada (approved):
Thanks for your contribution!

zneix (approved):
:+1:

src/common/ChannelChatters.cpp
kiwec: My concern is that if the color map gets updated after the reference is returned and before it is displayed, the reference could point to the wrong data.

Compiler also warned against returning a reference - since I don't do a lot of C++ I don't know how references work, but I'm assuming it's just like pointers.

leon-richardt: This section in the [`std::map::insert_or_assign`](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) leads me to think otherwise:

> No iterators or references are invalidated.

Also, I did not get any compiler warnings with gcc when I tried the change, are you using clang?

(Also, the function could be `const` in either case)

fanway:
Does it continuously populate color cache without ever clearing it?

fanway:
Sorry btw to write after it have been merged xd

PR #2298, opened 2020-12-19
Fix crash when receiving whisper

kiwec:
Pull request checklist:

# Description

The color mentions PR #2284 introduced a crash when receiving whispers.

We save user colors in the relevant twitch channels, but whispers don't have one, so we crash because `this->twitchChannel` is a `nullptr`.

Fix: don't save the color for whispers. Another possible fix would be to save colors globally instead of per-channel.


zneix (approved):
Can confirm it fixes the issue :+1:

Mm2PL (approved):

PR #2306, opened 2020-12-21
Migrated follow and unfollow methods to Helix API

zneix:
Pull request checklist:

# Description

Moved place of declared Kraken methods from `TwitchAccount` class to `Helix` class. Updated api documentation accordingly

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -313,29 +313,35 @@ void CommandController::initialize(Settings &, Paths &paths)
             channel->addMessage(makeSystemMessage("Usage: /follow [user]"));
             return "";
         }
-        auto app = getApp();
 
-        auto user = app->accounts->twitch.getCurrent();
-        auto target = words.at(1);
+        auto currentUser = getApp()->accounts->twitch.getCurrent();
 
-        if (user->isAnon())
+        if (currentUser->isAnon())
         {
             channel->addMessage(
                 makeSystemMessage("You must be logged in to follow someone"));
             return "";
         }
 
+        auto target = words.at(1);
+
         getHelix()->getUserByName(
             target,
-            [user, channel, target](const auto &targetUser) {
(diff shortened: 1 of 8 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -347,29 +353,35 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                    },
+                    [channel, target]() {
+                        channel->addMessage(makeSystemMessage(
+                            "An error occurred while unfollowing " + target));
+                    });
             },
             [channel, target] {
                 channel->addMessage(makeSystemMessage(
-                    "User " + target + " could not be followed!"));
+                    QString("User %1 could not be followed!").arg(target)));
```
pajlada, line 384: While we're modifying this, could we correct the message to include that a user with the name `target` was not found?
Maybe something along the lines of `User %1 could not be followed, no user with that name found.`

src/controllers/commands/CommandController.cpp
```diff
@@ -347,29 +353,35 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            [currentUser, channel, target](const auto &targetUser) {
+                getHelix()->unfollowUser(
+                    currentUser->getUserId(), targetUser.id,
+                    [channel, target]() {
+                        channel->addMessage(makeSystemMessage(
+                            "You successfully unfollowed " + target));
+                    },
+                    [channel, target]() {
+                        channel->addMessage(makeSystemMessage(
+                            "An error occurred while unfollowing " + target));
```
pajlada, line 379: Same pattern could be used here:
`User %1 could not be followed, an unknown error occured.` since it'll be unknown until we have better error handling, although looking at the documentation it seems like it'll most likely just be unknown errors

pajlada (changes requested):
Two nitpicks, other than that looks great! Thanks a lot for your contribution!

pajlada (approved):

PR #2316, opened 2020-12-27
Added automatic streamer mode detection to Linux

zneix:
Pull request checklist:

# Description

This updates handling `Streamer Mode -> Enable -> "Detect OBS"` setting to also work on Linux. Uses `pgrep` to determine whether `obs` process is running or not.

### DIFF
--- a/src/util/StreamerMode.cpp
+++ b/src/util/StreamerMode.cpp
@@ -68,7 +68,7 @@ bool isInStreamerMode()
                 return (p.exitCode() == 0);
             }
 
-            // Fallback to true and showing a warning
+            // Fallback to false and showing a warning
 
             if (shouldShowWarning)
             {
--- a/src/util/StreamerMode.cpp
+++ b/src/util/StreamerMode.cpp
@@ -1,6 +1,14 @@
 #include "StreamerMode.hpp"
 
+#include "Application.hpp"
+#include "common/QLogging.hpp"
+#include "messages/MessageBuilder.hpp"
 #include "singletons/Settings.hpp"
+#include "singletons/WindowManager.hpp"
+#include "widgets/Notebook.hpp"
+#include "widgets/Window.hpp"
+#include "widgets/helper/NotebookTab.hpp"
+#include "widgets/splits/Split.hpp"
 
 #ifdef USEWINSDK
 #    include <Windows.h>
@@ -14,12 +22,14 @@ namespace chatterino {
 
 constexpr int cooldownInS = 10;
 
+bool shouldShowWarning = true;
+
 const QStringList &broadcastingBinaries()
 {
 #ifdef USEWINSDK
     static QStringList bins = {"obs.exe", "obs64.exe"};
 #else
-    static QStringList bins = {};
+    static QStringList bins = {"obs"};
 #endif
     return bins;
 }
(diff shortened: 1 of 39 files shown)


src/util/StreamerMode.cpp
```diff
@@ -32,51 +32,62 @@ bool isInStreamerMode()
...
-    {
-        return false;
-    }
-    static bool cache = false;
-    static QDateTime time = QDateTime();
+#ifdef Q_OS_LINUX
+            QProcess p;
+            // No matter where we're redirecting the output, it still appears in the console for some reason
+            // p.setStandardOutputFile(QProcess::nullDevice());
+            auto exitCode = p.execute("pgrep", {"obs"});
```
pajlada, line 41: use streamingBinaries here

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -326,7 +326,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
 
     ComboBox *dankDropdown =
         layout.addDropdown<std::underlying_type<StreamerModeSetting>::type>(
-            "Enable Streamer Mode", {"No", "Yes", "Detect OBS (Windows only)"},
+            "Enable Streamer Mode", {"No", "Yes", "Detect OBS"},
```
leon-richardt, line 329: Think we should mention that is does not work on Mac (or maybe the Linux way could also be used on Mac?)

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,108 @@ bool isInStreamerMode()
...
+        case StreamerModeSetting::DetectObs:
 
-#ifdef USEWINSDK
-    if (!IsWindowsVistaOrGreater())
-    {
-        return false;
-    }
-    static bool cache = false;
-    static QDateTime time = QDateTime();
+#ifdef Q_OS_LINUX
```
fanway, line 47: Works on mac for me and i believe since MacOS 10.8 pgrep is available by default.
```suggestion
#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
```

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,108 @@ bool isInStreamerMode()
...
+            if (time.isValid() &&
+                time.addSecs(cooldownInS) > QDateTime::currentDateTime())
+            {
+                return cache;
+            }
+            time = QDateTime::currentDateTime();
 
-    time = QDateTime::currentDateTime();
+            QProcess p;
+            p.start("grep", {"-x", broadcastingBinaries().join("|")},
```
fanway, line 60:
```suggestion
            p.start("pgrep", {"-x", broadcastingBinaries().join("|")},
```

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,108 @@ bool isInStreamerMode()
...
+                    QIODevice::NotOpen);
 
-    WTS_PROCESS_INFO *pWPIs = nullptr;
-    DWORD dwProcCount = 0;
+            if (p.waitForFinished(1000))
```
fanway, line 63: [exitCode value is not valid unless exitStatus() returns NormalExit](https://doc.qt.io/qt-5/qprocess.html#exitCode)
```suggestion
            if (p.waitForFinished(1000) && p.exitStatus() == QProcess::NormalExit)
```

src/util/StreamerMode.cpp
zneix: Silly typo, thanks :+1:

src/util/StreamerMode.cpp
zneix: I didn't want to "yolo-add" a feature while I have no way to test this on Mac. I also think `pgrep` isn't available by default on most mac installations, please correct me if I'm wrong.

fanway: 10.8 released like in 2012. So it's fair to say that most mac installation have it.

Mm2PL (approved):

src/util/StreamerMode.cpp
zneix: If someone (like @pajlada or @gempir) could confirm it works on Mac, I'd be glad to make this change and get mac done as well.

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,109 @@ bool isInStreamerMode()
...
 
-    if (pWPIs)
-    {
-        WTSFreeMemory(pWPIs);
-    }
+            qCWarning(chatterinoStreamerMode) << "pgrep execution timed out!";
 
-    cache = false;
+            cache = true;
+            return true;
```
pajlada, line 101: The default return value has always been false - changing this to true should not be part of the same PR that adds discovery to Linux

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,109 @@ bool isInStreamerMode()
...
+            }
+
+            if (pWPIs)
+            {
+                WTSFreeMemory(pWPIs);
+            }
+
+            cache = false;
+#endif
+            return true;
```
pajlada, line 146: The default return value has always been false - changing this to true should not be part of the same PR that adds discovery to Linux

src/util/StreamerMode.cpp
pajlada: Let's leave that to a separate PR, but thanks for letting us know!

src/util/StreamerMode.cpp
```diff
@@ -32,51 +42,109 @@ bool isInStreamerMode()
...
+                            dynamic_cast<SplitContainer *>(selected))
+                    {
+                        for (auto &&split : container->getSplits())
+                        {
+                            if (auto channel = split->getChannel();
+                                !channel->isEmpty())
+                            {
+                                channel->addMessage(makeSystemMessage(
+                                    "Streamer mode got enabled, due to pgrep "
+                                    "missing. Install it to fix the issue."));
```
pajlada, line 91: When the "default on error" value has been reverted back to false, this phrasing needs to update too.
Maybe something along the lines of:
"Streamer mode is set to automatic, but pgrep is missing. Install it to fix the issue or set streamer mode to enabled or disabled manually in the settings."

pajlada (changes requested):

src/util/StreamerMode.cpp
```diff
@@ -32,51 +43,93 @@ bool isInStreamerMode()
...
 
-    if (WTSEnumerateProcesses(WTS_CURRENT_SERVER_HANDLE, NULL, 1, &pWPIs,
-                              &dwProcCount))
-    {
-        //Go through all processes retrieved
-        for (DWORD i = 0; i < dwProcCount; i++)
-        {
-            QString processName = QString::fromUtf16(
-                reinterpret_cast<char16_t *>(pWPIs[i].pProcessName));
+            // Fallback to true and showing a warning
```
pajlada, line 71: Update comment now that the fallback value isn't true

pajlada (changes requested):

pajlada (approved):

PR #2321, opened 2020-12-30
Add 'open in mod view' button to split context menu

Robinlemon:
Pull request checklist:

# Description

Adds an extra button to the split context menu to open the moderation view for a channel when the account selected has moderator permissions.

# Preview

Using anonymous mode: 
Using an account with moderator permissions:

### DIFF
--- a/src/widgets/splits/SplitHeader.cpp
+++ b/src/widgets/splits/SplitHeader.cpp
@@ -366,8 +366,11 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
         menu->addAction(OPEN_PLAYER_IN_BROWSER, this->split_,
                         &Split::openBrowserPlayer);
 #endif
-        menu->addAction(OPEN_MOD_VIEW_IN_BROWSER, this->split_,
-                        &Split::openModViewInBrowser);
+        if (this->split_->getChannel()->hasModRights())
+        {
+            menu->addAction(OPEN_MOD_VIEW_IN_BROWSER, this->split_,
+                            &Split::openModViewInBrowser);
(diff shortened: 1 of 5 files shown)


src/widgets/splits/SplitHeader.cpp
```diff
@@ -366,6 +366,12 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
         menu->addAction(OPEN_PLAYER_IN_BROWSER, this->split_,
                         &Split::openBrowserPlayer);
 #endif
+        if (this->split_->getChannel()->hasModRights())
+        {
+            menu->addAction(OPEN_MOD_VIEW_IN_BROWSER, this->split_,
+                            &Split::openModViewInBrowser);
+        }
+
```
zneix, line 374: nitpick: Why is this inserted in between "Open stream in X" options? Wouldn't it make more sense to move it under "Open in streamlink"?
Since this is not going to be present in all the splits, it might be confusing to my (and maybe other's) muscle memory while trying to open a new stream in streamlink with one click.

zneix (changes requested):
Overall it's a good enhancement :+1: I like it, though I just have one nitpick.

src/widgets/splits/SplitHeader.cpp
Robinlemon: @zneix Good point, and I definitely agree! I've now moved it to be below both player options.

zneix (approved):
LGTM 👍

pajlada (approved):
Looks good to me! Thanks for the contribution!

PR #2344, opened 2021-01-08
Added /chatters command

### LINKED ISSUE #2343: Add a /chatters command
**What should be added?**
Add a `/chatters` command, just shows a total of the current users in chat

**Why should it be added?**
We already have the data in chatterino, why not use it?

zneix:
Pull request checklist:

# Description

As described in #2343, this is meant to replicate BetterTTV extension's behavoir and make use of the `chatter_count` property we're reciving while fetching data from TMI API. If the approach I chose isn't the best, let me know how can I change it.

While I was at it, I replaced deprecated `QString::SkipEmptyLines` with corresponding updated version from `Qt` namespace

Closes #2343

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -439,13 +439,16 @@ void CommandController::initialize(Settings &, Paths &paths)
         "/chatters", [](const auto & /*words*/, auto channel) {
             auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
 
-            auto message =
-                (twitchChannel == nullptr)
-                    ? "The /chatters command only works in Twitch Channels"
-                    : "Chatter count: " +
-                          QString::number(twitchChannel->chatterCount());
+            if (twitchChannel == nullptr)
+            {
+                channel->addMessage(makeSystemMessage(
+                    "The /chatters command only works in Twitch Channels"));
+            }
 
-            channel->addMessage(makeSystemMessage(message));
+            channel->addMessage(makeSystemMessage(
+                QString("Chatter count: %1")
+                    .arg(QString::number(twitchChannel->chatterCount()))));
+
             return "";
         });
 }
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -439,11 +439,13 @@ void CommandController::initialize(Settings &, Paths &paths)
         "/chatters", [](const auto & /*words*/, auto channel) {
             auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
 
-            channel->addMessage(makeSystemMessage(
+            auto message =
                 (twitchChannel == nullptr)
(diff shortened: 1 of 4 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -434,6 +434,22 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            if (twitchChannel == nullptr)
+            {
+                channel->addMessage(makeSystemMessage(
+                    "The /chatters command only works in Twitch Channels"));
+                return "";
+            }
+
+            channel->addMessage(makeSystemMessage(
+                "Chatter count: " +
+                QString::number(twitchChannel->chatterCount())));
```
23rd, line 450:
```suggestion
            const auto message =
                twitchChannel
                    ? "Chatter count: " +
                          QString::number(twitchChannel->chatterCount())
                    : "The /chatters command only works in Twitch Channels";
            channel->addMessage(makeSystemMessage(message));
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -770,9 +775,14 @@ void TwitchChannel::refreshChatters()
...
                 if (!shared)
+                {
                     return Failure;
+                }
+
+                auto data = result.parseJson();
+                this->chatterCount_ = data.value("chatter_count").toInt();
 
-                auto pair = parseChatters(result.parseJson());
+                auto pair = parseChatters(data);
```
23rd, line 785:
```suggestion
                auto pair = parseChatters(std::move(data));
```

src/controllers/commands/CommandController.cpp
leon-richardt: I don't agree that this is easier to read

src/controllers/commands/CommandController.cpp
```diff
@@ -434,6 +434,20 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        "/chatters", [](const auto & /*words*/, auto channel) {
+            auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
+
+            auto message =
+                (twitchChannel == nullptr)
+                    ? "The /chatters command only works in Twitch Channels"
+                    : "Chatter count: " +
+                          QString::number(twitchChannel->chatterCount());
+
+            channel->addMessage(makeSystemMessage(message));
```
pajlada, line 448: use an if case instead of a ternary operator here - readability is more important than compactness

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -78,6 +78,7 @@ class TwitchChannel : public Channel,
     const QString &subscriptionUrl();
     const QString &channelUrl();
     const QString &popoutPlayerUrl();
+    int &chatterCount();
```
pajlada, line 81: return a value here instead of a reference

pajlada (changes requested):

pajlada (approved):

PR #2346, opened 2021-01-09
Add cache limit to `chattersColors`

fanway:
Pull request checklist:

- [x] ~~`CHANGELOG.md` was updated, if applicable~~

# Description

I was late to #2284 merge and raised some concerns about how colors cache never gets cleared. I'm not sure that it's the best solution, feel free to discuss.
Additionally, i've added some changes:
* Changed `chatterColor` -> `chattersColors` for consistency (nitpick, but mistyped myself a lot then searching)
* Changed `chattersColor` type form `std::map` to `std::unordered_map` (not entirely sure why its map in the first place)
* Remove `toLower()` in `setUserColor` (passed `user` is already in lowercase, so its leads to unnecessary string copy)

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -785,6 +785,16 @@ void ChannelView::messageAppended(MessagePtr &message,
     this->lastMessageHasAlternateBackground_ =
         !this->lastMessageHasAlternateBackground_;
 
+    if (auto tc = dynamic_cast<TwitchChannel *>(this->underlyingChannel_.get()))
+    {
+        this->channelConnections_.push_back(messageRef->destroyed.connect(
+            [tc, userName = messageRef->getMessage()->loginName] {
+                if (userName == "")
+                    return;
+                tc->removeUserColor(userName);
+            }));
+    }
+
     if (this->messages_.pushBack(MessageLayoutPtr(messageRef), deleted))
     {
         if (this->paused())
--- a/src/common/ChannelChatters.cpp
+++ b/src/common/ChannelChatters.cpp
@@ -72,10 +72,10 @@ void ChannelChatters::setChatters(UsernameSet &&set)
 
 const QColor ChannelChatters::getUserColor(const QString &user)
 {
-    const auto chatterColors = this->chatterColors_.accessConst();
+    const auto chattersColors = this->chattersColors_.accessConst();
 
-    const auto search = chatterColors->find(user.toLower());
-    if (search == chatterColors->end())
+    const auto search = chattersColors->find(user.toLower());
+    if (search == chattersColors->end())
     {
         // Returns an invalid color so we can decide not to override `textColor`
         return QColor();
@@ -86,8 +86,14 @@ const QColor ChannelChatters::getUserColor(const QString &user)
 
 void ChannelChatters::setUserColor(const QString &user, const QColor &color)
 {
-    const auto chatterColors = this->chatterColors_.access();
-    chatterColors->insert_or_assign(user.toLower(), color);
+    const auto chattersColors = this->chattersColors_.access();
+    chattersColors->insert_or_assign(user, color);
+}
+
+void ChannelChatters::removeUserColor(const QString &user)
+{
+    const auto chattersColors = this->chattersColors_.access();
+    chattersColors->erase(user);
 }
 
 }  // namespace chatterino
--- a/src/common/ChannelChatters.hpp
+++ b/src/common/ChannelChatters.hpp
@@ -20,13 +20,14 @@ public:
     void setChatters(UsernameSet &&set);
     const QColor getUserColor(const QString &user);
     void setUserColor(const QString &user, const QColor &color);
+    void removeUserColor(const QString &user);
 
 private:
     Channel &channel_;
 
     // maps 2 char prefix to set of names
     UniqueAccess<UsernameSet> chatters_;
-    UniqueAccess<std::map<QString, QColor>> chatterColors_;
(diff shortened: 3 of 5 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -785,6 +785,16 @@ void ChannelView::messageAppended(MessagePtr &message,
     this->lastMessageHasAlternateBackground_ =
         !this->lastMessageHasAlternateBackground_;
 
+    if (auto tc = dynamic_cast<TwitchChannel *>(this->underlyingChannel_.get()))
+    {
+        this->channelConnections_.push_back(messageRef->destroyed.connect(
+            [tc, userName = messageRef->getMessage()->loginName] {
+                if (userName == "")
+                    return;
```
zneix, line 793:
```suggestion
                if (userName.isEmpty())
                {
                    return;
                }
```

zneix (changes requested):
You can also add a number of this PR to brackets of related changelog entry (line 7th in `CHANGELOG.md` as of right now).

leon-richardt:
Another possible route to improve performance: For two users with the same color, it is unnecessary to save two identical `QColor` instances in the map. Instead, we could point multiple users to the same `QColor` instance if they share colors.

I would estimate this to be quite effective since a majority of users likely uses one of the default colors.

fanway:
@leon-richardt Would you suggest create something like a map of all colors? In pr `chattersColors` size is ~800 elements in big chats, QColor is 8 bytes, so 6.25 kbytes per channel, seems like a pretty reasonable amount.

leon-richardt:
My (naive) first intuition would be something like a `QSet` containing `QColor`s and a `std::unordered_map` mapping a user name to a reference/pointer to the color in the `QSet`.

On second thought, however, I'm not sure if this makes sense. Even raw pointers are 8 bytes already, and if `QColor`s just take 8 bytes as well, my proposed approach would just add unnecessary complexity.

fanway:
@leon-richardt my bad, QColor is 16 bytes. But i think it's doesn't change much, still more time complexity + another level of indirection.

pajlada:
> @leon-richardt my bad, `QColor` is 16 bytes. Actually idk if its worth to implement, let me know if you think so.

I don't think it's worth the extra complexity

pajlada (changes requested):
Two concerns:
1) The `destroyed` solution seems unsound to me (I have not actually tested it)
 It would seem to me that if a chat looks like this:
```
pajlada: hello
leppunen: hi
leppunen: hi
leppunen: hi
leppunen: hi
pajlada: hello
```
and user leppunen types another message, which would cause the topmost `pajlada: hello` message to be destroyed, we would removed the pajlada color from the chatterscolors, which feels very incorrect as the user just now typed something.

2) this is a tradeoff between memory usage and cpu work - right now I have a hard time seeing the tradeoff is worth it. We need to collect real data about memory usage in both real world and extreme scenarios to see what memory usage we're talking about.

if we are to go forward with this, I'd like to first of all have data from 2) showing memory usage concerns and then coming up with a better solution that doesn't have the issue 1) described

I'm fine merging in the minor changes here, like the switch from map to unordered_map if you want to have that in asap

fanway:
>     1. The `destroyed` solution seems unsound to me (I have not actually tested it)
>        It would seem to me that if a chat looks like this:

If someone types again in chat, it'll try to readd color to chattersColors, but if were are several messages of one user on the screen and one of them got destroyed, which leads to removal, this is definitely unsound (maybe this is exactly what you mean, just want to make sure). On top of my head, maybe to add some counter to `chattersColors` something like `std::unordered_map<string, std::pair<int, QColor>` but i'm not sure here.

>     2. this is a tradeoff between memory usage and cpu work - right now I have a hard time seeing the tradeoff is worth it. We need to collect real data about memory usage in both real world and extreme scenarios to see what memory usage we're talking about.

Its not only a memory issue, i believe even `std::unordered_map` lookup time start to grow linearly after it won't fit into L3 cache (~70k entries for `std::unordered_map<int,int>`, for strings it should be a lot less), i think its easily achievable, for example, if i don't close the app for several days in some big active chat. But i totally agree that collecting data will help a lot.

pajlada:
You could use the debug popup here to show the count of chatters in chatterscolors, that way we could start gathering some data individually

fanway:
Ye, i've done some simple debug to check map size, in like 120k xqc chat (very spammy), it has hit 10k entries within an hour and kept increasing.

fanway:
So, i did some basic profiling:
First of all, it's kind of random to profile memory usage with username length varies all the time, so i checked for edge case using just `std::unordered_map<std::string, QColor>` with unique 25 length string keys (max for twitch username) which lead to ~109 bytes allocated per entry.
The next thing i did was check `chattersColors` size for big streamers with active chat, here some stats:

<table>
<tr>
	<td> <b>Streamer
	<td> <b>Online
	<td> <b>chattersColors entries
	<td> <b>Total size
<tr>
	<td> shroud (mid stream he switched to submode, so i stopped checking)
	<td> 60000-70000
	<td> 11000
	<td> 1.14 mbytes
<tr>
	<td> forsen
	<td> 30000-35000
	<td> 14000
	<td> 1.45 mbytes
<tr>
	<td> xQcOw
	<td> 105000-120000
	<td> 48000
	<td> 4.99 mbytes
<tr>
	<td> random stream
	<td> 3500-3900
	<td> 2800
	<td> 298 kbytes
</table>

Stats was collected within one stream and total size was calculated with 109 bytes per entry in mind.
My main concern was what happens if i don't close app/tab, it'll just populate forever? But i have a feeling it might be some extreme use case. So, if you think its insignificant tradeoff, i'm okay to close the pr.

PR #2350, opened 2021-01-09
feat: add a `finally` callback to NetworkRequests

leon-richardt:
Pull request checklist:

- [x] ~`CHANGELOG.md` was updated, if applicable~ (n/a)

# Description

This patch adds an additional callback to `NetworkRequest` instances. This callback will execute regardless of whether the request failed or succeeded and it will always be executed after the `onSuccess` and `onError` callbacks.

This allows for code de-duplication in some cases (e.g. when unlocking a mutex or updating a timestamp) and can thus reduce the potential of human programming errors. For example, the timestamping logic of #2271 could be simplified with this feature.


zneix (approved):

pajlada (approved):
Looks good to me! Please give me an approval of the extra commit I pushed to the branch if it all looks good then I'll go ahead and merge this in.

leon-richardt (commented):
Looks good, thanks for adding the additional test!

(I can only comment on the commit instead of explicitly approving it, thanks GitHub.)

PR #2360, opened 2021-01-14
Implement /marker command

### LINKED ISSUE #1109: Add /marker command (from twitch chat)
Context: https://help.twitch.tv/s/article/creating-highlights-and-stream-markers?language=en_US#Add

### Steps to reproduce

- Go to a channel you own
- Start streaming
- Send `/marker` command 

### Observed

System message that says `Sorry, "/user" is not available through this client.`

2019-06-22-220115_611x141_scrot

### Expected

Stream marker is successfully added.

zneix:
Pull request checklist:

# Description

Requires chatterino/login#10

Command was tested in webchat and I managed to replicate its behaviour in 100%.

Closes #1109


Mm2PL:
Please add a case for anon users :-)

zneix:
> Please add a case for anon users :-)

Good idea, added it in 7573d59

Mm2PL (approved):

src/providers/twitch/api/README.md
```diff
@@ -107,6 +107,14 @@ Requires `user:edit:follows` scope
      * `widgets/dialogs/UserInfoPopup.cpp` to unfollow a user by unticking follow checkbox in usercard
      * `controllers/commands/CommandController.cpp` in /unfollow command
 
+### Create Stream Marker
+URL: https://dev.twitch.tv/docs/api/reference/#create-stream-marker
```
pajlada, line 111:
```suggestion
URL: https://dev.twitch.tv/docs/api/reference/#create-stream-marker  
```
add two spaces here so it shows up nicely on a new line

src/providers/twitch/api/README.md
```diff
@@ -107,6 +107,14 @@ Requires `user:edit:follows` scope
...
+Requires `user:edit:broadcast` scope
+
+ * We implement this in `providers/twitch/api/Helix.cpp unfollowUser`
```
pajlada, line 114:
```suggestion
 * We implement this in `providers/twitch/api/Helix.cpp unfollowUser`  
```

Adds two new lines here so it shows up nicely on a new line

src/controllers/commands/CommandController.cpp
```diff
@@ -452,6 +453,48 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+    this->registerCommand("/marker", [](const QStringList &words,
+                                        auto channel) {
+        // Avoid Helix calls without Client ID and/or OAuth Token
+        if (getApp()->accounts->twitch.getCurrent()->isAnon())
+        {
+            channel->addMessage(makeSystemMessage(
+                "You need to be logged in to create stream markers!"));
+        }
+
+        auto *twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
```
pajlada, line 466: Please check the return value of `dynamic_cast` to ensure this doesn't crash. Even if non-`TwitchChannel`s don't currently use `CommandController`, it's still a sane thing to do

src/controllers/commands/CommandController.cpp
```diff
@@ -452,6 +453,48 @@ void CommandController::initialize(Settings &, Paths &paths)
 
             return "";
         });
+
+    this->registerCommand("/marker", [](const QStringList &words,
+                                        auto channel) {
+        // Avoid Helix calls without Client ID and/or OAuth Token
+        if (getApp()->accounts->twitch.getCurrent()->isAnon())
+        {
+            channel->addMessage(makeSystemMessage(
```
pajlada, line 462: You're missing a return here to actually stop the request from happening if the user is not logged in

src/controllers/commands/CommandController.cpp
```diff
@@ -452,6 +453,48 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        }
+
+        auto *twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
+
+        // Exact same message as in webchat
+        if (!twitchChannel->isLive())
+        {
+            channel->addMessage(makeSystemMessage(
+                "You can only add stream markers during live streams. Try "
+                "again when the channel is live streaming."));
```
pajlada, line 473: Missing return here - we shouldn't even try creating a stream marker if the channel is live.

pajlada (changes requested):

src/controllers/commands/CommandController.cpp
zneix: I am a little confused because every command is using this solution everywhere.

zneix:
We just tested it with @Mm2PL on livestream and triple-checked scopes of a token that we're using for creating stream marker and scope `user:edit:broadcast` doesn't seem to be required for marker creation. A possible Twitch bug?

pajlada (approved):

pajlada:
Thank you for your contribution! :heart:

Issue #2190, opened 2020-11-16
--channels and/or -c args not working


Sukid:
**Describe the bug**
The command line arguments for launching certain channels aren't working. I've tried every variant I can think of, including
`chatterino -c t:joviandragon`
`chatterino --channels joviandragon`
`chatterino --channels t:joviandragon`

I've also tried changing capitalization, using quotation marks, removing spaces, adding colons... as much as I could think of. the most I got was getting an error message if I tried something like -Ctheemster, although it just says those are unhandled options. Otherwise, I never even get an error message. The application just completely ignores the arguments.

**To reproduce**
Just try to use the --channels command line switch with any given twitch channel. The window will either come up empty or with the channels you were already in and that's it.

**Chatterino version**
Chatterino 2.2.3-beta (commit 7d2b64087)
I've also tried it in the stable 2.2.2 release as well.

**Operating system**
Both Arch Linux and Windows 10

**Additional information**
--version and --help aren't working in Windows 10, but they are working in Linux. --help-all doesn't work in either. --version and --help on Windows just seem to print out nothing, whereas with --help-all the application just launches as though there were no arguments just like with -c.

pajlada:
Confirmed this is a bug - it was lost in some merge conflict. I think @zneix will be looking into this.

Thanks for the report

zneix:
@pajlada correct, sorry for slacking on this one but some code responsible for this function got lost in a merge conflict. I looked into this and I am in the process of restoring that :+1:

pdf:
Behaviour here is slightly different to the original report - no CLI arguments work at all here on Linux (Arch pkg chatterino2-git) as of a6e23e246864458c9d00c469c3b7306086b8d0ed: `--help` and `--version` simply exit with no output, and `--channels` starts the application with no effect.

PR #2368, opened 2021-01-17
fix: properly handle CLI arguments

leon-richardt:
Pull request checklist:

# Description

Fix CLI arguments not being respected. This happened due to the addition of category-based logging (`--help`, `--version`) and changes to the window loading ( `--channels`), respectively.

When handling `--channels`, I took the liberty to refactor the previous version of window description (which relied on generating JSON) to directly building the `WindowLayout`.

Closes https://github.com/Chatterino/chatterino2/issues/2190.

### DIFF
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -315,10 +315,10 @@ void WindowManager::initialize(Settings &settings, Paths &paths)
     }
 
     // No main window has been created from loading, create an empty one
-    if (mainWindow_ == nullptr)
+    if (this->mainWindow_ == nullptr)
     {
-        mainWindow_ = &this->createWindow(WindowType::Main);
-        mainWindow_->getNotebook().addPage(true);
+        this->mainWindow_ = &this->createWindow(WindowType::Main);
+        this->mainWindow_->getNotebook().addPage(true);
     }
 
     settings.timestampFormat.connect([this](auto, auto) {
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -37,8 +37,6 @@
 namespace chatterino {
 namespace {
 
-    const QString WINDOW_LAYOUT_FILENAME(QStringLiteral("window-layout.json"));
-
     boost::optional<bool> &shouldMoveOutOfBoundsWindow()
     {
         static boost::optional<bool> x;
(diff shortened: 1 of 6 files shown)


zneix:
Thanks for fixing my laziness

zneix (approved):
Everything seems to be good, though, you may add this PR's number to the changelog entry about `--channels` parameter.

leon-richardt:
> Everything seems to be good, though, you may add this PR's number to the changelog entry about `--channels` parameter.

ACK, added it. I will keep the separate entry in the changelog as well since `--version` and `--help` were also fixed

src/singletons/WindowManager.cpp
```diff
@@ -297,18 +297,29 @@ void WindowManager::initialize(Settings &settings, Paths &paths)
     assert(!this->initialized_);
 
     {
-        auto windowLayout = this->loadWindowLayoutFromFile();
+        WindowLayout windowLayout;
+
+        if (getArgs().customChannelLayout)
+        {
+            // We would love to use std::optional<T>::value() instead but macOS
+            // does not support it yet. :)
```
fanway, line 305: But it does though :) (from macOS 10.14 (apple clang 10.0)). Right now chatterino min macOS version is 10.13 for some reason, we could just bump it to 10.14 or even 10.15 :)

leon-richardt: Although I agree an argument could be made to explicitly mention in the comment that this refers to 10.13

fanway: It's not actually about macos version, it's about apple clang version. So you could just use normal compiler and you will be fine :) Also, i think a good point to raise version is that some of the external libs (for example boost) already build for 10.15 and it always gives a warning that it's linked to a different version. Sorry for write it here, kind of offtopic

leon-richardt: > It's not actually about macos version, it's about apple clang version.

Aha, I am little confused about that since the error message explicitly mentions 10.14:
```
../src/singletons/WindowManager.cpp:304:58: error: 'value' is unavailable: introduced in macOS 10.14
```

But then again, I am just vaguely familiar with the Apple ecosystem so I don't feel like I can contribute much to the conversation

fanway: Ye, i agree, message is definitely confusing 😄 Sorry, i just like `::value` usage here, so i thought you could just bump macOS min version :)

leon-richardt: I would prefer to use `::value()` as well but it's not my decision to make :man_shrugging: Would love to hear input from @fourtf and @pajlada on this

leon-richardt: Decided to go with `boost::optional` for now, as suggested by @pajlada on stream.

pajlada (approved):

PR #2375, opened 2021-01-18
Added millisecond timestamp formats to settings

### LINKED ISSUE #2372: Milliseconds timestamp format.
# What should be added?

An option under timestamp format to display milliseconds. 
( `00:00:00.000` & `00:00:00.000 am/pm` )

I've seen zneix at least has a working version of something like this from the chatterino on his stream.
I would try to make this myself, but building chatterino seems like its complicated, and I only started learning C++ like a week ago
# Why should it be added?

Its useful information for people / developers making things that integrate with twitch chat, even if the typical user won't find any use for this I think I will, being able to easily see response times and compare them easily to other bots. 😃 And I like seeing information like this.  ( even though when you turn on logging this information is availible in the log files, it is very annoying to check through those )

zneix:
Pull request checklist:

# Description

Adds milisecond timestamp format to settings.
Does this need a changelog entry? I don't think so myself NOIDONTTHINKSO

Closes #2372


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -257,9 +257,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addCheckbox("Highlight messages redeemed with Channel Points",
                        s.enableRedeemedHighlight);
     layout.addDropdown<QString>(
-        "Timestamp format (a = am/pm)",
+        "Timestamp format (a = am/pm, zzz = miliseconds)",
```
fanway, line 260: :)
```suggestion
        "Timestamp format (a = am/pm, zzz = milliseconds)",
```

zneix: I am way too used to a Polish version of this word ;p thank you!

fourtf:
it's so psychotic that `h:mm:ss.zzz a` is missing from the list

pajlada (approved):

PR #2386, opened 2021-01-21
Bump macOS minimum version to 10.14

fanway:
Pull request checklist:

# Description

Right now minimum macOS version is 10.13. For example some external libraries (boost, openssl) have brew formulas only for 10.14 (Mojave). 
Also it would introduce full C++17 (currently used) support for apple clang, which will be useful for example in https://github.com/Chatterino/chatterino2/pull/2368#pullrequestreview-573454781
Maybe it should be bumped to only 10.14 for more compatibility, let me know if you think so.

### DIFF
--- a/BUILDING_ON_MAC.md
+++ b/BUILDING_ON_MAC.md
@@ -1,5 +1,6 @@
 # Building on macOS
 #### Note - If you want to develop Chatterino 2 you will also need to install Qt Creator (make sure to install **Qt 5.12 or newer**)
+#### Note - Chatterino 2 is only tested on macOS 10.14 and above - anything below that is considered unsupported. It may or may not work on earlier versions
 1. Install Xcode and Xcode Command Line Utilites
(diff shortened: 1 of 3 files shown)


zneix:
I know it's not specified yet, but maybe you should also include this information regarding minimum mac version in `BUILDING_ON_MAC.md` document.

pajlada:
I think 10.14 makes more sense in that case - staying in line with whatever brew decides to do seems like a good choice

fanway:
@pajlada ye, i agree, 10.14 seems more reasonable, do you think it worth to add information about minimum version to `BUILDING_ON_MAC.md` as @zneix mentioned?

pajlada:
> @pajlada ye, i agree, 10.14 seems more reasonable, do you think it worth to add information about minimum version to `BUILDING_ON_MAC.md` as @zneix mentioned?

yeah I think so - I took the freedom of just adding that note to `BUILDING_ON_MAC.md` and pushing it to your branch

PR #2404, opened 2021-01-24
Update HighlightingPage.cpp

### LINKED ISSUE #2403: Remove "Case-sensitive" column for user highlights
**What should be removed?**
The column to toggle case-sensitivity in the "Users" tab on the "Highlights" page in the settings. (See screenshot.)

image

**Why should it be removed?**
The case of a user name does not contribute to its uniqueness. Thus, this column serves no real purposes but adds visual clutter and may potentially confuse users.

AweesneW:
Applies changes mentioned in issue #2403

Closes #2403


src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -99,7 +99,7 @@ HighlightingPage::HighlightingPage()
                         .getElement();
 
                 view->addRegexHelpLink();
-                view->getTableView()->horizontalHeader()->hideSection(4);
+                view->getTableView()->horizontalHeader()->hideSection(5);
```
leon-richardt, line 102:
```suggestion
                view->getTableView()->horizontalHeader()->hideSection(
                    HighlightModel::Column::UseRegex);
                view->getTableView()->horizontalHeader()->hideSection(
                    HighlightModel::Column::CaseSensitive);
```

Your change gets rid of the "Case-sensitive" column but the "Enable regex" column is then shown instead (se image)
.

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -99,7 +99,7 @@ HighlightingPage::HighlightingPage()
...
```
leon-richardt, line 103: Your change gets rid of the "Case-sensitive" column but the "Enable regex" column is then shown instead (see image).

image

```suggestion
                view->getTableView()->horizontalHeader()->hideSection(
                    HighlightModel::Column::UseRegex);
                view->getTableView()->horizontalHeader()->hideSection(
                    HighlightModel::Column::CaseSensitive);
```
This suggestion hides both columns and also uses the existing `Column` enum which is more readable. :smile:

leon-richardt (changes requested):

pajlada (changes requested):
Please add a changelog entry for this change - you can get hints for phrasing in #2403

pajlada (approved):

PR #2411, opened 2021-01-27
Added x-attach-split-to-window command line arg

fourtf:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

Adds a flags to attach a split to a parent WND on windows.

PR #2427, opened 2021-01-31
Ensure SwitchSplitItem container can never be null

pajlada:
There were two ways to create a SwitchSplitItem:
a) A Container only: This would let the user switch to a Tab only, but not change which split in the focus was focused
b) A split only: This would programmatically fill in the container (by querying the split) and let the user switch to the Tab and change focus to the split

Since split had no way to programmatically get its container, we can
instead require that the container is always filled in. Where this class
is used (QuickSwitcherPopup.cpp:97 and QuickSwitcherPopup:108) we do
have the container pointer already, so the logic for creating a
SwitchSplitItem has now changed to:
a) A container but no split
b) A container and a split

Pull request checklist:

# Description

### DIFF
--- a/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp
+++ b/src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp
@@ -94,7 +94,7 @@ void QuickSwitcherPopup::updateSuggestions(const QString &text)
             if (split->getChannel()->getName().contains(text,
                                                         Qt::CaseInsensitive))
             {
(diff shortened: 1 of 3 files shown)


fourtf:
This might collide with https://github.com/Chatterino/chatterino2/pull/2411

pajlada:
> This might collide with #2411

This PR is to be merged into #2411 - not into master. I didn't want to make an edit straight into your PR :dancers:

pajlada:
LGTM :+1: 
image

PR #2443, opened 2021-02-10
Added /streamlink command

zneix:
Pull request checklist:

# Description

Usage: /streamlink [channel], use "channel" literally to open current channel in streamlink.

I am not so sure about using 'channel' for opening the current channel in streamlink, it's just a concept and I may change it to just `/streamlink` opening the current channel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,8 +50,7 @@
 - Minor: Update the listing of top-level domains. (#2345)
 - Minor: Properly respect RECONNECT messages from Twitch (#2347)
 - Minor: Added human-readable formatting to remaining timeout duration. (#2398)
-- Minor: Added `/streamlink` command. Usage: `/streamlink <channel>`. You can also use the command without arguments in any twitch channel to open it in streamlink
-. (#2443)
+- Minor: Added `/streamlink` command. Usage: `/streamlink <channel>`. You can also use the command without arguments in any twitch channel to open it in streamlink. (#2443)
 - Bugfix: Fix crash occurring when pressing Escape in the Color Picker Dialog (#1843)
 - Bugfix: Fix bug where the "check user follow state" event could trigger a network request requesting the user to follow or unfollow a user. By itself its quite harmless as it just repeats to Twitch the same follow state we had, so no follows should have been lost by this but it meant there was a rogue network request that was fired that could cause a crash (#1906)
 - Bugfix: /usercard command will now respect the "Automatically close user popup" setting (#1918)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,8 @@
 - Minor: Update the listing of top-level domains. (#2345)
 - Minor: Properly respect RECONNECT messages from Twitch (#2347)
 - Minor: Added human-readable formatting to remaining timeout duration. (#2398)
-- Minor: Added `/streamlink` command. Usage: /streamlink [channel], use "channel" literally to open current channel in streamlink. (#2443)
+- Minor: Added `/streamlink` command. Usage: `/streamlink <channel>`. You can also use the command without arguments in any twitch channel to open it in streamlink
+. (#2443)
 - Bugfix: Fix crash occurring when pressing Escape in the Color Picker Dialog (#1843)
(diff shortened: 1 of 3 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -538,6 +539,32 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        }
+
+        if (!channel->isTwitchChannel())
+        {
+            return "";
+        }
+
+        QString channelName =
+            (words[1] == "channel") ? channel->getName() : words[1];
+        qDebug() << channelName;
```
leon-richardt, line 560:
```suggestion
```

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 - Minor: Update the listing of top-level domains. (#2345)
 - Minor: Properly respect RECONNECT messages from Twitch (#2347)
 - Minor: Added human-readable formatting to remaining timeout duration. (#2398)
+- Minor: Added `/streamlink` command. (#2443)
```
leon-richardt, line 53: I think we should also add a brief explanation of what the command does

src/controllers/commands/CommandController.cpp
```diff
@@ -538,6 +539,32 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         return "";
     });
+
+    this->registerCommand("/streamlink", [](const QStringList &words,
+                                            ChannelPtr channel) {
+        if (words.size() < 2)
+        {
+            channel->addMessage(makeSystemMessage(
+                "Usage: /streamlink [channel], use \"channel\" to open current "
```
leon-richardt, line 548: Do we have an established syntax for commands and especially command arguments? In [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form) (which is also commonly used in man pages), `[ ]` is used to denote that an argument is optional. In this case, the argument is required, and cannot be omitted.

If we have already established this syntax elsewhere, it's fine by me. Otherwise, I suggest changing this to keep in line with common conventions.

src/controllers/commands/CommandController.cpp
```diff
@@ -538,6 +539,32 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         return "";
     });
+
+    this->registerCommand("/streamlink", [](const QStringList &words,
+                                            ChannelPtr channel) {
+        if (words.size() < 2)
```
leon-richardt, line 545: Should we also print usage when more than one argument is provided?

leon-richardt (changes requested):

src/controllers/commands/CommandController.cpp
zneix: Sure thing, I went with [] because I changed my mind about whether `"channel"` would be required or not, I guess I will change this to `< >` to keep consistency with EBNF. :+1:

src/controllers/commands/CommandController.cpp
zneix: I don't think so, I was thinking about maybe making something like accepting `best` / `worst` as further arguments and then, after a little rework of streamlink functions, make it autoselect best / worst quality, but I am unsure about it.

src/controllers/commands/CommandController.cpp
```diff
@@ -538,6 +539,31 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         return "";
     });
+
+    this->registerCommand(
+        "/streamlink", [](const QStringList &words, ChannelPtr channel) {
```
pajlada, line 544: I think `/streamlink` without any parameters should be legal when used in Twitch channels.
Using `/streamlink pajlada` should be legal in any channel, not just Twitch channels.

src/controllers/commands/CommandController.cpp
```diff
@@ -538,6 +539,31 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            if (words.size() < 2)
+            {
+                channel->addMessage(makeSystemMessage(
+                    "Usage: /streamlink <channel>, use \"channel\" literally "
+                    "to open current channel in streamlink."));
```
pajlada, line 549: If we allow `/streamlink` to be called without parameters, the literal `channel` parameter could go away too, making this usage message a bit more clear

pajlada (changes requested):

pajlada (approved):

PR #2447, opened 2021-02-12
Support for building with the CMake build system

AnotherFoxGuy:
## Description
Pull request for building chatterino2 with cmake

Status: Builds and runs on windows without any issues

------------------

**Pull request checklist:**
- [x] Fix tests
- [ ] Update build instructions
- [x] Test build on Windows
- [ ] Test build on Linux
- [ ] Test build on Mac
- [ ] Update `CHANGELOG.md`


pajlada:
A few notes

- Windows depending on something like conan or vcpkg (should be able to support both) is fine, however it should *not* be needed on macOS or Linux, so we need to make sure conan/vcpkg-things are cleaned up
- Tests should *not* be removed - modify the root `CMakeLists.txt` so that if `BUILD_TESTS` is not set to true we build the actual application. In future, both should be built if `BUILD_TESTS` is set to true
- Don't use a glob to find files
- humanize is not used in chatterino anymore, so no need to keep any buildfiles for it

pajlada:
also why do we need FindSignals/FindSerialize when they already implement reasonable (I think) CMakeLists.txt files? the add_subdirectory works already

pajlada:
with globbing being removed, we should not need to use the `generate_source_groups` function which should remove that dependency

src/CMakeLists.txt
```diff
@@ -0,0 +1,272 @@
...
+        widgets/splits/ClosedSplits.{hpp,cpp}
+        widgets/splits/EmoteInputItem.{hpp,cpp}
+        widgets/splits/EmoteInputPopup.{hpp,cpp}
+        widgets/splits/Split.{hpp,cpp}
+        widgets/splits/SplitContainer.{hpp,cpp}
+        widgets/splits/SplitHeader.{hpp,cpp}
+        widgets/splits/SplitInput.{hpp,cpp}
+        widgets/splits/SplitOverlay.{hpp,cpp}
+        widgets/dialogs/IrcConnectionEditor.ui
+        )
```
pajlada, line 212: Only .cpp files should be part of the `add_executable` code (and .ui/qrc etc)

AnotherFoxGuy: The header files should be added to `add_executable` to appear as dependencies in the Makefile, and also to be added in the Visual Studio project (otherwise the header files will be marked as external dependencies)

pajlada: SeemsGood :+1: been taught wrong, googling modern examples shows you're right

.github/workflows/test.yml
```diff
@@ -20,29 +20,16 @@ jobs:
...
-        with:
-          aqtversion: '==1.1.1'
-          cached: ${{ steps.cache-qt.outputs.cache-hit }}
-          extra: --external 7z
-
-      # LINUX
       - name: Install dependencies (Ubuntu)
         if: startsWith(matrix.os, 'ubuntu')
         run: |
+            curl "https://raw.githubusercontent.com/AnotherFoxGuy/ci-scripts/main/install-conan.sh" | sudo bash
```
pajlada, line 32: we should not be using conan when building the tests, or we should be using both conan and not conan through the matrix function

and don't run a random curl shell script to install conan when it's so few lines on ubuntu
```
      - name: Install Conan (Ubuntu)
        run: |
            sudo apt-get update && sudo apt-get -y install python3 python3-pip python3-setuptools
            pip3 install --user conan
            echo "$HOME/.local/bin" >> $GITHUB_PATH
```

pajlada:
https://github.com/AnotherFoxGuy/chatterino2/pull/18

AnotherFoxGuy:
@pajlada Could you maybe try to fix the https://github.com/AnotherFoxGuy/libcommuni-cmake build on Linux with QT 5.15.2? I don't have a Linux system to test on.
Its probably a compile flag I forgot to port over to the cmake files.

pajlada:
> @pajlada Could you maybe try to fix the [AnotherFoxGuy/libcommuni-cmake](https://github.com/AnotherFoxGuy/libcommuni-cmake) build on Linux with QT 5.15.2? I don't have a Linux system to test on.
> Its probably a compile flag I forgot to port over to the cmake files.

The errors seem to come from moc files being included inside a `namespace` directive

so the files basically look like this:
```cpp
// xd.cpp
namespace Communi {
#include "moc_xd.cpp"
}
```

```cpp
// moc_xd.cpp
#include <memory>
```

which leads to us basically having one of the stdlib include-files inside the Communi namespace, confusing the compiler.
This seems to be fine with some compilers & some build systems for some reason, but it doesn't fly with cmake with makefiles on Linux

This **seems** to be by design https://github.com/communi/libcommuni/commit/70381bcf so I don't know what the long-term fix is.

https://github.com/AnotherFoxGuy/libcommuni-cmake/pull/1 This PR fixes the compilation of libcommuni with cmake on my local machine, I don't know

pajlada:
Found a different solution to the issue which might be easier to merge into libcommuni master, which is just pre-include the `memory` header, I've updated the above-linked PR. It makes a little bit more sense now why this might work on something like visual studio but not gcc/make or w/e since their "compilation units" function quite differently

pajlada:
Transitionally: I would like to provide both CMake and QMake build options - I think it makes sense to utilize the "matrix build" functionality of GitHub Actions for this, and this is something that I'd like to work on and PR onto this PR.

pajlada:
https://github.com/AnotherFoxGuy/chatterino2/pull/20

PR #2340, opened 2021-01-05
Add custom hotkeys.

Mm2PL:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

This pull request adds custom hotkeys to Chatterino.
- Closes #1592, 
Definitely one of those PRs which you squash and merge.

# TODO
  - [x] Add defaults for hotkeys.
  - [x] Make it add new default hotkeys.
  - [x] Move all hotkeys/menu actions to new system. Done?
    - [-] ~~Fucking Ctrl+C~~ not doable with current systems and knowledge 
  - [x] Add user interface. ~~(another pr?)~~
    - [x] new settings tab
    - [x] modifying hotkeys
    - [x] saving hotkeys
  - [x] Add actions for menu items
  - [x] Implement usage of `HotkeyController::shortcutsForScope()` in
    - [x] `EmotePopup`
    - [x] `SelectChannelDialog` (`Escape => close, Return => accept, Ctrl+Tab => next tab, Ctrl+Shift+Tab => prev tab`)
    - [x] `SettingsDialog` (`Ctrl+F => search in settings`)
    - [x] `Split`
    - [x] `SplitInput` (figure out how to override default shortcuts from QTextEdit)
    - [x] `UserInfoPopup` (`Escape => close`)
    - [x] `Window`
    - [x] `Window` debug shortcuts, is this even worth it?
    - [x] non-obvious actions
    - Something else?
  - [x] Bind commands to a hotkey. It's easy enough and will allow for less code duplication.
  - [x] Make sure to deprecate the old shortcut creation methods.
  - [x] fix delet
  - [x] Add developer docs on adding new hotkeys
  - [x] live reloading
  - [x] Add usage documentation Chatterino/wiki#57
  - [x] Check if I broke : and @-completion menus, with ↑ and ↓.
  - [ ] Make tab-completion use new system
  - [x] MAKE SURE TO TEST WITH WEIRD KEYBOARD LAYOUTS like Dvorak, Cyrillic keyboards, etc.
     - [x] I checked the Dvorak layout (Linux/X11), works exactly as before this
     - [x] I checked Russian layout (Linux/X11), works exactly as before this
     - [ ] needs Wayland testing for Dvorak
     - [x] boring_nick tested Cyrillic keyboards on wayland ~~needs Wayland testing for Russian or other Cyrillic keyboard~~
     - [x] needs Windows testing for Dvorak
     - [x] needs Windows testing for Russian or other Cyrillic keyboard 
[...]

PR #2464, opened 2021-02-16
Register "Enter" shortcut to confirm select channel dialog

woranov:
Pull request checklist:

# Description

Registered numpad's enter key as a shortcut to confirm the select channel dialog.


Mm2PL:
This PR conflicts with #2340, there I've essentially rewritten the way that shortcuts work in Chatterino to allow customization. IMO it would make way more sense to add it as a default there than to deal with the merge conflicts created by using by creating this PR.

PR #2489, opened 2021-03-02
feat: /streamermode command

### LINKED ISSUE #2487: Streamer Mode toggle by chat command such as /streamer
**What should be added?**
A command in chat to toggle Streamer Mode On/Off
**Why should it be added?**
User would not have to open settings to toggle Streamer Mode. Automatic setting is undesirable because it keeps Streamer Mode on when OBS is running but not streaming.

KararTY:
Pull request checklist:

# Description

Still not sure how this code should look like. So if the feature is considered, change code as you see fit.

Closes #2487


pajlada:
While the command might be useful, I'm afraid we would deprecate it in favour of some more generic 'set setting' command too soon.

Please checkout the newly created issue #2635 and give your feedback so we know how to proceed with the functionality

PR #2500, opened 2021-03-06
Fix macOS building with qt6 by default

### LINKED ISSUE #2494: MacOS CI attempts to use Qt6
`macos-latest` builds use `qt6` as the default which is not supported. Now Qt5 also seems to do this
https://github.com/Chatterino/chatterino2/runs/2035040185?check_suite_focus=true

~~https://github.com/Chatterino/chatterino2/pull/2340/checks?check_run_id=2032889985~~

@pajlada

pajlada:
Pull request checklist:

# Description

Fixes #2494

### DIFF
--- a/.CI/CreateDMG.sh
+++ b/.CI/CreateDMG.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 echo "Running MACDEPLOYQT"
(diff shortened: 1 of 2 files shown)


pajlada:
I tested this on my macos machine using big sur and the new .dmg files (even though they are smaller) still work as expected.

pajlada:
image

PR #2501, opened 2021-03-07
Add tray icon

### LINKED ISSUE #1358: Minimize Chatterino to system tray
Basically, you have to have chatterino open, even if you dont want to use it directly. This is used for the window inside the stream which replaces the normal chat.

Either:

- Minimize main chatterino window and remove it from the taskbar and show only the icon in the system tray. Only show the chatterino windows used in browsers ( if you have more than 1 window ) 
- Have the chatterino (program) window be the one that replaces the chat window in the browser.

Closing the window by clicking (X) but that also closes the window in the browser.

**Additional context**
Kinda like spotify, which has a 'if you click close, minimize to system tray.'

univrsal:
Pull request checklist:

# Description
Added an optional tray icon. The tray icon can be used when the main window is minimized, closed or both. Clicking the tray icon will toggle the visibility of all windows. Right clicking it will open a context menu to show or close the application so users have an option to exit if they set Chatterino to hide when closing the main window.
image

I only tested this on linux with KDE, so it would have to be tested on other platforms as well, but I assume Qt takes care of the tray icon fairly well. I hope I put everything in the right place, I didn't really know where to put the enum for the tray icon setting, so if there's some issues I'd be happy to correct them.

Closes #1358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Add ability to minimize and close Chatterino to system tray (#2501)
 - Minor: Added `is:first-msg` search option. (#3700)
 - Minor: Added quotation marks in the permitted/blocked Automod messages for clarity. (#3654)
 - Minor: Adjust large stream thumbnail to 16:9 (#3655)
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -58,6 +58,14 @@ enum UsernameDisplayMode : int {
     LocalizedName = 2,             // Localized name
     UsernameAndLocalizedName = 3,  // Username (Localized name)
 };
+
+enum TrayAction : int {
+    AskMe = 0,              // Default
+    MinimizeToTray = 1,     // Minimize to tray icon
+    MinimizeToTaskBar = 2,  // Minimize to task bar
+    CloseChatterino = 3     // Close program
+};
+
 /// Settings which are availlable for reading and writing on the gui thread.
 // These settings are still accessed concurrently in the code but it is bad practice.
 class Settings : public ABSettings, public ConcurrentSettings
@@ -112,6 +120,10 @@ public:
     BoolSetting showTabLive = {"/appearance/showTabLiveButton", true};
     BoolSetting hidePreferencesButton = {"/appearance/hidePreferencesButton",
                                          false};
+    EnumSetting<TrayAction> minizeTrayAction = {
+        "/appearance/minimizeTrayAction", AskMe};
+    EnumSetting<TrayAction> closeTrayAction = {"/appearance/closeTrayAction",
+                                               AskMe};
     BoolSetting hideUserButton = {"/appearance/hideUserButton", false};
     BoolSetting enableSmoothScrolling = {"/appearance/smoothScrolling", true};
     BoolSetting enableSmoothScrollingNewMessages = {
--- a/src/widgets/BaseWindow.cpp
+++ b/src/widgets/BaseWindow.cpp
@@ -169,11 +169,18 @@ void BaseWindow::init()
                 TitleBarButton *_exitButton = new TitleBarButton;
                 _exitButton->setButtonStyle(TitleBarButtonStyle::Close);
 
-                QObject::connect(_minButton, &TitleBarButton::leftClicked, this,
-                                 [this] {
-                                     this->setWindowState(Qt::WindowMinimized |
-                                                          this->windowState());
-                                 });
+                QObject::connect(
+                    _minButton, &TitleBarButton::leftClicked, this, [this] {
+                        if (this->isMainWindow())
+                        {
+                            handleMinimizeEvent(nullptr);
+                        }
+                        else
+                        {
+                            this->setWindowState(Qt::WindowMinimized |
+                                                 this->windowState());
+                        }
+                    });
                 QObject::connect(_maxButton, &TitleBarButton::leftClicked, this,
                                  [this, _maxButton] {
                                      this->setWindowState(
@@ -298,6 +305,11 @@ bool BaseWindow::supportsCustomWindowFrame()
 #endif
 }
 
+FlagsEnum<BaseWindow::Flags> const &BaseWindow::getFlags()
+{
+    return flags_;
+}
+
 void BaseWindow::themeChangedEvent()
 {
     if (this->hasCustomWindowFrame())
@@ -361,6 +373,16 @@ void BaseWindow::wheelEvent(QWheelEvent *event)
                 getSettings()->getClampedUiScale() - 0.1);
         }
     }
+}
+
+bool BaseWindow::shouldHandleTrayEvent(bool visible)
+{
+    return isVisible() != visible;
+}
+
+bool BaseWindow::isMainWindow()
+{
+    return false;
 }
 
 void BaseWindow::onFocusLost()
@@ -712,6 +734,95 @@ void BaseWindow::updateScale()
     for (auto child : this->findChildren<BaseWidget *>())
     {
         child->setScale(scale);
+    }
+}
+
+void BaseWindow::handleMinimizeEvent(QEvent *event)
+{
+    auto showMessage = [&] {
+        QMessageBox msgBox(QMessageBox::Question, "Chatterino",
+                           "What should we do when you press the "
+                           "minimize button? Minimizing to the taskbar "
+                           "is the default behavior.",
+                           QMessageBox::NoButton, this);
+        auto *taskbar =
+            msgBox.addButton("Minimize to task bar", QMessageBox::YesRole);
+        auto *tray =
+            msgBox.addButton("Minimize to tray", QMessageBox::ApplyRole);
+
+        msgBox.setWindowModality(Qt::ApplicationModal);
+        msgBox.setWindowFlag(Qt::WindowContextHelpButtonHint, false);
+        msgBox.setWindowFlag(
+            Qt::Popup,
+            true);  // Prevents users from being able to minimize the message box on linux (KDE)
+        msgBox.exec();
+
+        auto *result = static_cast<QPushButton *>(msgBox.clickedButton());
+        if (result == taskbar)
+        {
+            return TrayAction::MinimizeToTaskBar;
+        }
+        return TrayAction::MinimizeToTray;
+    };
+
+    auto action = getSettings()->minizeTrayAction.getEnum();
+    if (action == TrayAction::AskMe)
+    {
+        action = showMessage();
+        getSettings()->minizeTrayAction.setValue(action);
+    }
+
+    if (action == TrayAction::MinimizeToTaskBar)
+    {
+        this->setWindowState(Qt::WindowMinimized | this->windowState());
+    }
+    else if (action == TrayAction::MinimizeToTray)
+    {
+        getApp()->windows->setVisibilityAll(false);
+        if (event)
+            event->ignore();
+    }
+}
+
+void BaseWindow::handleCloseEvent(QEvent *event)
+{
+    auto action = getSettings()->closeTrayAction.getEnum();
+    auto showMessage = [&] {
+        QMessageBox msgBox(
+            QMessageBox::Question, "Chatterino",
+            "What should we do when you press the close button? Closing "
+            "Chatterino is the default behavior.",
+            QMessageBox::NoButton, this);
+        auto *close =
+            msgBox.addButton("Close Chatterino", QMessageBox::YesRole);
+        auto *tray =
+            msgBox.addButton("Minimize to tray", QMessageBox::ApplyRole);
+
+        msgBox.setWindowFlag(Qt::WindowContextHelpButtonHint, false);
+        msgBox.setWindowFlag(
+            Qt::Popup,
+            true);  // Prevents users from being able to minimize the message box on linux (KDE)
+        msgBox.exec();
+
+        auto *result = static_cast<QPushButton *>(msgBox.clickedButton());
+        if (result == close)
+        {
+            return TrayAction::CloseChatterino;
+        }
+        return TrayAction::MinimizeToTray;
+    };
+
+    if (action == TrayAction::AskMe)
+    {
+        action = showMessage();
+        getSettings()->closeTrayAction.setValue(action);
+    }
+
+    if (action == TrayAction::MinimizeToTray)
+    {
+        getApp()->windows->setVisibilityAll(false);
+        if (event)
+            event->ignore();
     }
 }
 
--- a/src/widgets/Window.cpp
+++ b/src/widgets/Window.cpp
@@ -66,6 +66,39 @@ Window::Window(WindowType type)
     if (type == WindowType::Main)
     {
         this->resize(int(600 * this->scale()), int(500 * this->scale()));
+
+        getSettings()->tabDirection.connect([this](int val) {
+            this->notebook_->setTabDirection(NotebookTabDirection(val));
+        });
+
+        actionShow_ = new QAction("Show", this);
+        actionExit_ = new QAction("Exit", this);
+        trayContextMenu_ = new QMenu(this);
+
+        trayContextMenu_->addAction(actionShow_);
+        trayContextMenu_->addSeparator();
+        trayContextMenu_->addAction(actionExit_);
+
+        trayIcon_ = new QSystemTrayIcon(this);
+        trayIcon_->setVisible(false);
+        trayIcon_->setIcon(QIcon(":/icon.ico"));
+        trayIcon_->setToolTip(Version::instance().fullVersion());
+        trayIcon_->setContextMenu(trayContextMenu_);
+
+        connect(actionShow_, &QAction::triggered, this, [] {
+            getApp()->windows->setVisibilityAll(true);
+        });
+        connect(actionExit_, &QAction::triggered, this, [] {
+            QApplication::exit();
+        });
+
+        connect(trayIcon_, &QSystemTrayIcon::activated, this, [](auto reason) {
+            if (reason == QSystemTrayIcon::ActivationReason::Trigger)
+            {
+                getApp()->windows->setVisibilityAll(
+                    !getApp()->windows->getMainWindow().isVisible());
+            }
+        });
     }
     else
     {
@@ -95,6 +128,11 @@ SplitNotebook &Window::getNotebook()
     return *this->notebook_;
 }
 
+QSystemTrayIcon *Window::getTrayIcon()
+{
+    return trayIcon_;
+}
+
 bool Window::event(QEvent *event)
 {
     switch (event->type())
@@ -122,27 +160,49 @@ bool Window::event(QEvent *event)
             }
         }
         break;
-
+#if !defined(Q_OS_WINDOWS)
+        // We handle this in the custom minimize button on Windows
+        case QEvent::WindowStateChange:
+            if (this->type_ == WindowType::Main && isMinimized())
+            {
+                handleMinimizeEvent(event);
+            }
+            break;
+#endif
         default:;
     }
 
     return BaseWindow::event(event);
 }
 
-void Window::closeEvent(QCloseEvent *)
+bool Window::isMainWindow()
 {
-    if (this->type_ == WindowType::Main)
+    return this->type_ == WindowType::Main;
+}
+
+void Window::closeEvent(QCloseEvent *e)
+{
+    auto action = getSettings()->closeTrayAction.getEnum();
+    if (this->type_ == WindowType::Main &&
+        action != TrayAction::CloseChatterino)
     {
-        auto app = getApp();
-        app->windows->save();
-        app->windows->closeAll();
+        handleCloseEvent(e);
     }
-
-    this->closed.invoke();
-
-    if (this->type_ == WindowType::Main)
+    else
     {
-        QApplication::exit();
+        if (this->type_ == WindowType::Main)
+        {
+            auto app = getApp();
+            app->windows->save();
+            app->windows->closeAll();
+        }
+
+        this->closed.invoke();
+
+        if (this->type_ == WindowType::Main)
+        {
+            QApplication::exit();
+        }
     }
 }
 
--- a/src/widgets/Window.hpp
+++ b/src/widgets/Window.hpp
@@ -2,6 +2,7 @@
 
 #include "widgets/BaseWindow.hpp"
 
+#include <QSystemTrayIcon>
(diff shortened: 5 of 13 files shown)


leon-richardt:
(Merging this closes #1358.)

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
 - Minor: Remove EmojiOne 2 and 3 due to license restrictions. (#1555)
 - Minor: Added `/streamlink` command. Usage: `/streamlink <channel>`. You can also use the command without arguments in any twitch channel to open it in streamlink. (#2443, #2495)
 - Minor: Humanized all numbers visible to end-users. (#2488)
+- Minor: Add tray icon (#2448, #1358)
```
zneix, line 59: I'd make this a Major changelog entry and rephrase to something like `Added ability to minimize Chatterino to system tray` to make it more descriptive to an average end-user.

src/widgets/Window.cpp
```diff
@@ -69,6 +69,26 @@ Window::Window(WindowType type)
         getSettings()->tabDirection.connect([this](int val) {
             this->notebook_->setTabDirection(NotebookTabDirection(val));
         });
+
+        actionExit_ = new QAction("Exit", this);
+        actionShow_ = new QAction("Show", this);
```
zneix, line 74: nitpick: to keep consistency with `addAction()` below, reorder these lines

src/widgets/Window.cpp
```diff
@@ -69,6 +69,26 @@ Window::Window(WindowType type)
...
+        trayContextMenu_ = new QMenu(this);
+
+        trayContextMenu_->addAction(actionShow_);
+        trayContextMenu_->addSeparator();
+        trayContextMenu_->addAction(actionExit_);
+
+        connect(actionExit_, SIGNAL(triggered()), this, SLOT(exitAction()));
+        connect(actionShow_, SIGNAL(triggered()), this, SLOT(showAction()));
```
zneix, line 82: nitpick: to keep consistency with `addAction()` below, reorder these lines

src/widgets/Window.cpp
```diff
@@ -69,6 +69,26 @@ Window::Window(WindowType type)
...
+
+        trayIcon_ = new QSystemTrayIcon(this);
+        trayIcon_->setIcon(QIcon(":/icon.ico"));
+        trayIcon_->setToolTip("Chatterino");
```
zneix, line 86: I'd modify the tooltip to be more consistent with the window title and at least include version.

src/singletons/Settings.hpp
```diff
@@ -17,6 +17,8 @@ using TimeoutButton = std::pair<QString, int>;
 
 namespace chatterino {
 
+enum TrayIconTrigger { OnMinimize, OnClose, OnCloseAndMinimize };
+
```
zneix, line 21: I'd move this enum to a different file, similar to how it is done with `StreamerModeSetting`. Perhaps `Window.hpp` might be a good place for it.

src/widgets/Window.hpp
```diff
@@ -27,6 +28,8 @@ class Window : public BaseWindow
 
     pajlada::Signals::NoArgSignal closed;
 
+    QSystemTrayIcon *getTrayIcon();
+
```
zneix, line 32: Why not put it right under `&getNotebook()` ?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -91,6 +92,14 @@ GeneralPage::GeneralPage()
...
 
     this->initExtra();
+
+    getSettings()->enableTrayIcon.connect([&](bool enable) {
+        if (trayIconTrigger_)
+            trayIconTrigger_->setEnabled(enable);
+        auto *trayIcon = getApp()->windows->getMainWindow().getTrayIcon();
+        if (trayIcon)
+            trayIcon->setVisible(enable);
+    });
```
zneix, line 102: This should be moved to `initExtra` function with a comment

src/singletons/Settings.hpp
```diff
@@ -99,6 +101,9 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting showTabLive = {"/appearance/showTabLiveButton", true};
     BoolSetting hidePreferencesButton = {"/appearance/hidePreferencesButton",
                                          false};
+    BoolSetting enableTrayIcon = {"/appearance/enableTrayIcon", false};
+    EnumSetting<TrayIconTrigger> trayIconTrigger = {
+        "/appearance/trayIconTrigger", TrayIconTrigger::OnMinimize};
```
zneix, line 106: In my humble opinion, this could be enabled by deafult and use `TrayIconTrigger::OnClose` as a default tray icon trigger

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -178,6 +187,39 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     }
     layout.addCheckbox("Show which channels are live in tabs", s.showTabLive);
 
+    layout.addCheckbox("Enable tray icon", s.enableTrayIcon);
+
+    trayIconTrigger_ =
+        layout.addDropdown<std::underlying_type<TrayIconTrigger>::type>(
+            "Hide to tray icon when",
+            {"minimizing", "closing", "minimizing or closing"},
```
zneix, line 195:
```suggestion
            {"Minimizing", "Closing", "Minimizing or Closing"},
```

zneix (changes requested):
Extra notes;
1. Right now, minimizing the application to system tray via tray icon, hides only main window and all other windows like settings dialog, usercards, emote picker, etc. stay open.
2. WHile minimized to tray, closing any other visible window such as usercard or settings dialog results in application quitting - this should not be a case.

univrsal:
> While minimized to tray, closing any other visible window such as usercard or settings dialog results in application quitting - this should not be a case.

This doesn't happen for me, so maybe it's only an issue on windows? I'll try and test it on Windows later on.
> Right now, minimizing the application to system tray via tray icon, hides only main window and all other windows like settings dialog, usercards, emote picker, etc. stay open.

This one is a bit more complicated. I switched to a signal for hiding/showing windows which seems to have fixed the issue mostly from what I can tell. The only issue left is that the settings dialog is always set to visible when clicking the tray icon, but that should be fixable.

zneix:
> This doesn't happen for me, so maybe it's only an issue on windows? I'll try and test it on Windows later on.

I am an arch user, btw. Maybe it's because of i3wm.

pajlada:
How does this handle multiple Chatterino windows? This is mostly a thing for Windows users who use the extension, or StreamLink (or other external viewer programs) users who might start Chatterino with the `-c` flag?

Getting the behaviour right for most users will be a difficult task.
I believe the setting(s) should be more akin to: **Behaviour when pressing the Close button** and **Behaviour when pressing the Minimize button** rather than a single setting trying to control both behaviours.
In addition, the default value for **Close** should be "ask users for future preference" (not sure how that would be shown in the settings page), and the default value for **Minimize** should be the old behaviour (so just minimize to the task bar, not to the tray)

univrsal:
> How does this handle multiple Chatterino windows? This is mostly a thing for Windows users who use the extension, or StreamLink (or other external viewer programs) users who might start Chatterino with the -c flag?

It only reacts to the main window closing or being minimized. I wanted it so I could keep the program open and still get live notifications. I guess it should automatically start minimized if the ``-c`` flag is used? It should maybe also save the state and load it on launch.
> I believe the setting(s) should be more akin to: Behaviour when pressing the Close button and Behaviour when pressing the Minimize button rather than a single setting trying to control both behaviours.

Seems fair.
> In addition, the default value for Close should be "ask users for future preference"

I honestly dislike programs that do that. I'd leave it at the expected behavior and just let users set it in the settings.

pajlada:
> > In addition, the default value for Close should be "ask users for future preference"
> 
> I honestly dislike programs that do that. I'd leave it at the expected behavior and just let users set it in the settings.

Yeah if there is a valid default that works for most people I agree, but this is a pretty polarizing "feature", and the "ask users what their future preference is" has become pretty standard for some applications at least (I think Discord/Spotify at least)

pajlada:
> > How does this handle multiple Chatterino windows? This is mostly a thing for Windows users who use the extension, or StreamLink (or other external viewer programs) users who might start Chatterino with the -c flag?
> 
> It only reacts to the main window closing or being minimized. I wanted it so I could keep the program open and still get live notifications. I guess it should automatically start minimized if the `-c` flag is used? It should maybe also save the state and load it on launch.

That's my bad - I thought the tray icon was always visible a la Discord, but if the tray icon only shows up when you minimize or close the application (depending on the user settings) that makes more sense to me.

One thing to take into consideration with multiple windows, still, is that if Chatterino is started with the -c flag, settings aren't saved so asking the question for future actions wouldn't make sense

univrsal:
> 
> 
> > > How does this handle multiple Chatterino windows? This is mostly a thing for Windows users who use the extension, or StreamLink (or other external viewer programs) users who might start Chatterino with the -c flag?
> > 
> > 
> > It only reacts to the main window closing or being minimized. I wanted it so I could keep the program open and still get live notifications. I guess it should automatically start minimized if the `-c` flag is used? It should maybe also save the state and load it on launch.
> 
> That's my bad - I thought the tray icon was always visible a la Discord, but if the tray icon only shows up when you minimize or close the application (depending on the user settings) that makes more sense to me.
> 
> One thing to take into consideration with multiple windows, still, is that if Chatterino is started with the -c flag, settings aren't saved so asking the question for future actions wouldn't make sense

Currently it is always visible, but I guess it'd make more sense to only show it when Chatterino is hidden.

pajlada:
> > > > How does this handle multiple Chatterino windows? This is mostly a thing for Windows users who use the extension, or StreamLink (or other external viewer programs) users who might start Chatterino with the -c flag?
> > > 
> > > 
> > > It only reacts to the main window closing or being minimized. I wanted it so I could keep the program open and still get live notifications. I guess it should automatically start minimized if the `-c` flag is used? It should maybe also save the state and load it on launch.
> > 
> > 
> > That's my bad - I thought the tray icon was always visible a la Discord, but if the tray icon only shows up when you minimize or close the application (depending on the user settings) that makes more sense to me.
> > One thing to take into consideration with multiple windows, still, is that if Chatterino is started with the -c flag, settings aren't saved so asking the question for future actions wouldn't make sense
> 
> Currently it is always visible, but I guess it'd make more sense to only show it when Chatterino is hidden.

sorry for the double misunderstanding! :D 

I think, for a single Chatterino window, it makes sense to keep the tray icon visible at all times, especially if we want to expand on the tray icon usage to show things like unread messages, have a place to spawn notifications from etc.

I think the question still is how tray icons behave with Chatterino windows other than the main window.

Maybe if Chatterino is started with -c no tray icon should be visible and close button/minimuze button has the same behaviour as now when no tray icon exists.
Same thing if Chatterino is running in "extension mode", we should also make sure no tray icon is visible

univrsal:
> I think, for a single Chatterino window, it makes sense to keep the tray icon visible at all times, especially if we want to expand on the tray icon usage to show things like unread messages, have a place to spawn notifications from etc.
> 
> I think the question still is how tray icons behave with Chatterino windows other than the main window.
> 
> Maybe if Chatterino is started with -c no tray icon should be visible and close button/minimuze button has the same behaviour as now when no tray icon exists.
> Same thing if Chatterino is running in "extension mode", we should also make sure no tray icon is visible

Is the gui even started if it's running in extension mode? From what I can see [here](https://github.com/Chatterino/chatterino2/blob/4ff2de0567dceb07c528e4b2c2d9646762037c8a/src/main.cpp#L61) it doesn't look like it so the icon shouldn't show up. Where is the ``-c`` flag checked? And what does it do? Also what do you mean by a single Chatterino window? No popup and user chats?

univrsal:
I know this isn't quite done, so I just wanted to quickly ask what I should do so this could be merged (if it's even good enough to be merged).

pajlada (changes requested):
These are the changes that are required for further review:
In the settings dialog, instead of having them be checkboxes, it should instead be a dropdown menu of actions.

The actions possible for the "What should we do when you press the Close button" are:
 - "Ask me" <-- DEFAULT OPTION
 - "Minimize to tray icon"
 - "Close Chatterino (Standard behaviour)"

The actions possible for the "What should we do when you press the Minimize button" are:
 - "Ask me" <-- DEFAULT OPTION
 - "Minimize to tray icon"
 - "Minimize to task bar (Standard behaviour)"

The "Ask me" option should open a dialog asking the user what they would like to do, and saving that option to the setting.

After that there's some thinking/testing to be done with regards to multiple Chatterino windows and how that behaviour works, in particular we will need to handle "Ask me" differently if Chatterino is running in a mode where it does not save settings.

univrsal:
> These are the changes that are required for further review:
> In the settings dialog, instead of having them be checkboxes, it should instead be a dropdown menu of actions.
> 
> The actions possible for the "What should we do when you press the Close button" are:
> 
> * "Ask me" <-- DEFAULT OPTION
> * "Minimize to tray icon"
> * "Close Chatterino (Standard behaviour)"
> 
> The actions possible for the "What should we do when you press the Minimize button" are:
> 
> * "Ask me" <-- DEFAULT OPTION
> * "Minimize to tray icon"
> * "Minimize to task bar (Standard behaviour)"
> 
> The "Ask me" option should open a dialog asking the user what they would like to do, and saving that option to the setting.
> 
> After that there's some thinking/testing to be done with regards to multiple Chatterino windows and how that behaviour works, in particular we will need to handle "Ask me" differently if Chatterino is running in a mode where it does not save settings.

Alrighty, I'll try to find some time next week. Does "Ask me" need to be an option in the settings dialog? Wouldn't it be enough to show this dialog the first time when the user minimizes/closes, apply the users choice to the settings dialog and leave it at that?

pajlada:
It doesn't necessarily need to be an option in the list, but I thought it might be the simple way to do it code-wise. Maybe it just shouldn't render in the list. So for action, since you'd never want to set it to "ask me", just default to minimize to task bar and let the use change it in the settings if they'd like. For close we ask still

cherie0:
hey was there any update on this? curious since i was searching around for exactly this.

univrsal:
No, I sadly haven't had time nor motivation to work on this as of late.

srnyx:
How about now, any updates? Using the setting that shows full window name in the taskbar for applications makes Chatterino take up a lot of space (especially when I'm just using it in the Twitch browser using the Chrome extension thing).

leon-richardt:
> How about now, any updates?

You can track the progress of this PR by following the commit history on this branch.

univrsal:
Well, better late than never:
image
image
image

Works on my machine™

srnyx:
@univrsal it should just say `Minimize to tray` instead of `Minimize to tray icon`

univrsal:
> @univrsal it should just say `Minimize to tray` instead of `Minimize to tray icon`

I just copy pasted that from pajlada's message.

Felanbird:
> > @univrsal it should just say `Minimize to tray` instead of `Minimize to tray icon`
> 
> I just copy pasted that from pajlada's message.

He must have hit his head that day, including `icon` in the setting name sounds fucking stupid.

Felanbird:
Initial thoughts (tested with both QtCreator build & nightly artifact): 
1. It seems the minimize setting does not properly hold the window to request what behavior you want, it always minimizes itself and then the prompt will appear next time you go to maximize it again, and because of this will re-minimize itself on confirmation. 
c2_2501_1

(It should be noted that, while clicking the `C` on your taskbar to minimize, the same issue arises, but the prompt does not follow and is still accessible.)
image

2. Clicking `X` on the popup forces the window into tray for both settings, and sets their respective `BoolSetting` to match.
It should either abort and keep the settings set to `Ask me`, or use original defaults (minimize & close) and keep the setting on `Ask me`.
c2_2501_2

3. Closing Chatterino from the taskbar `right-click -> Exit` will force it back to the taskbar on next startup.
c2_2501_3



4. Unrelated to code but reminder, Changelog will have to be dragged out of the depths of hell, and stripped of it's issue #. (we only do PR #'s now).

univrsal:
> Initial thoughts (tested with both QtCreator build & nightly artifact):
> 
>     1. It seems the minimize setting does not properly hold the window to request what behavior you want, it always minimizes itself and then the prompt will appear next time you go to maximize it again, and because of this will re-minimize itself on confirmation.
>        c2_2501_1
> 
> 
> (It should be noted that, while clicking the `C` on your taskbar to minimize, the same issue arises, but the prompt does not follow and is still accessible.) image
> 
>     2. Clicking `X` on the popup forces the window into tray for both settings, and sets their respective `BoolSetting` to match.
>        It should either abort and keep the settings set to `Ask me`, or use original defaults (minimize & close) and keep the setting on `Ask me`.
>        c2_2501_2
> 
>     3. Closing Chatterino from the taskbar `right-click -> Exit` will force it back to the taskbar on next startup.
>        c2_2501_3
> 
>     4. Unrelated to code but reminder, Changelog will have to be dragged out of the depths of hell, and stripped of it's issue #. (we only do PR #'s now).

Looks like 1. and 2. are issues on Windows. I'll try and debug it. 3. is probably a leftover from when the first attempt, the idea was that it should remember whether the program was hidden or not, but I guess that's not really necessary/useful.

univrsal:
So windows needed some extra code, because of the custom toolbar, but it seems to work now, I'll have to test it on linux again though

https://user-images.githubusercontent.com/8353672/167714926-a31782f3-4bcd-494b-9d41-b13a04933876.mp4

kornes (commented):
beside small things below, one more:
when user minimize chatterino window by clicking on task bar, OS will minimize it by itself while chatterino (with AskMe mode) will still show its MsgBox.

src/widgets/BaseWindow.cpp
```diff
@@ -715,6 +742,86 @@ void BaseWindow::updateScale()
...
+            msgBox.addButton("Minimize to task bar", QMessageBox::YesRole);
+        auto *tray =
+            msgBox.addButton("Minimize to tray icon", QMessageBox::ApplyRole);
+
+        msgBox.setWindowModality(Qt::ApplicationModal);
+        msgBox.setWindowFlags(msgBox.windowFlags().setFlag(
+            Qt::WindowContextHelpButtonHint, false));
+        msgBox.exec();
+
+        auto *result = msgBox.clickedButton();
```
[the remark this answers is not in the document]
univrsal: Initially the message box used the YesRole and NoRole. I assume that when as message box as an abort role it can be closed. Now that it only has YesRole and ApplyRole it can't be closed anymore

univrsal:
> when user minimize chatterino window by clicking on task bar, OS will minimize it by itself while chatterino (with AskMe mode) will still show its MsgBox. Suggested behavior in this case would be not showing msgbox and not changing AskMe settings.

That's how I've now implemented it on windows. I don't think there's an easy way to distinguish these events on linux though. So on linux clicking the taskbar icon just shows the message box.

pritamh:
hey just saw this after commenting in this thread almost a year ago and glad you got to it! was wondering if its currently available for users on most recent update as i dont see the option

LosFarmosCTL:
Alright, here comes the macOS review to make everything more complicated image

This PR in it's current form is problematic on macOS, as the concept of the menu bar (which is where the QSystemTrayIcon is put) differs quite a bit from the system tray on i.e. windows.

How this would be expected to work for a macOS application:

- A single setting "keep in menu bar", that adds a persistent chatterino icon (preferably monochrome) to the menu bar while chatterino is running
- If that setting is enabled, don't close the main chatterino window on clicking the close button and just hide it (like this PR is doing rn)
- On pressing the close button for the first time, display a dialog asking if you want to keep chatterino in the menu bar
- On clicking the menu bar icon, always focus the chatterino window
- Don't change anything about the minimizing behavior, macOS has a separate space for minimized applications in the dock and there is nothing that chatterino should change about that (ofc still un-minimize when clicking the menu bar icon)

Since those changes are fundamentally different from the implementation on windows/linux, I would guess that they'll probably require a lot of separate code for each platform (I haven't looked at the code in-depth so maybe it's less than I expect?) but I am unsure whether it makes sense for you to include that in this PR, without having a macOS system to test on. I could theoretically PR into your fork's branch or just open a separate PR after these changes get merged in, for a macOS implementation, but the current code will definitely have to be hidden, at least in part, behind ifdef's for macOS.

univrsal:
> Since those changes are fundamentally different from the implementation on windows/linux, I would guess that they'll probably require a lot of separate code for each platform (I haven't looked at the code in-depth so maybe it's less than I expect?) but I am unsure whether it makes sense for you to include that in this PR, without having a macOS system to test on. I could theoretically PR into your fork's branch or just open a separate PR after these changes get merged in, for a macOS implementation, but the current code will definitely have to be hidden, at least in part, behind ifdef's for macOS.

Yeah I'd prefer it if I didn't have to deal with macOS, but I'd assume that the changes shouldn't be that complex.

LosFarmosCTL:
> Yeah I'd prefer it if I didn't have to deal with macOS

understandable 😄

Then I'd say it's probably easiest for you to just make this PR non-macOS only and for me to  open another one afterwards adapting it.

alefbeo:
Is this supposed to be live on non-macOS platforms?

Felanbird:
> Is this supposed to be live on non-macOS platforms?

No, the green Open would be a purple Merged.
image
image

skonuksal:
Is this feature available on Nightly? I can't seem to find minimize it to the tray.

Felanbird:
Converting this PR to a draft, as I mentioned above unless the status below the PR title is a purple `[Merged]` - it has not yet made it to the nightly build(and subsequently stable releases).

Nerixyz:
I merged the current master branch on https://github.com/Nerixyz/chatterino2/tree/feat/tray-icon (https://github.com/Chatterino/chatterino2/compare/master...Nerixyz:chatterino2:feat/tray-icon) and cleaned it up a bit (and disabled the stuff on macOS).

In my opinion, there shouldn't be an "ask" option. This would simplify the implementation a bit. Thunderbird for example has a similar feature where you can minimize it to the tray, and it doesn't ask the user (at least last time I reinstalled it). Especially when implementing this as an MVP, the app shouldn't prompt users that updated.

One thing that feels a bit off on Windows in the current implementation is that windows hide as if they're closed. Thunderbird has a minimize animation in contrast, which gives a better indication of what actually happens. It would be fine to keep the current implementation in an MVP.

@univrsal please let us know if you want to continue this PR.

univrsal:
> @univrsal please let us know if you want to continue this PR.

probably not, I just learned to live with the pain and minimize the program.

PR #2504, opened 2021-03-08
Fixed deprecated method QTime::elapsed

zneix:
Pull request checklist:

no changelog entry needed LOOOOOOOOOOOOL

# Description

After I checked out to `cmake-build` branch, I saw many deprecation warnings and most of them\* can be fixed, so I went ahead and did just that.

\* We can't replace `QString::SkipEmptyParts` with newer `Qt::SkipEmptyParts` since the second method was added just in Qt 5.14, so it would not work on our minimal supported version (Qt 5.12). I also left a couple warnings that I didn't understand how to fix, perhaps I could fix them later on.  
I believe there should be no conflicts with both qmake builds and cmake ones, since migration to cmake builds doesn't touch code at all 😎

# Update

Edit: @pajlada wanted me to split changes from this pull request to multiple PRs (#2508, #2509, #2510, #2511, #2512)

This pull request specifically fixes deprecated [`QTime::elapsed`](https://doc.qt.io/qt-5/qtime-obsolete.html#elapsed), which is replaced with [`QTimerElapsed::elapsed`](https://doc.qt.io/qt-5/qtime-obsolete.html#elapsed) method. To keep better consistency, I updated other `QTime` timer in this file to use `QElapsedTimer` class as well

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -36,7 +36,7 @@
 namespace chatterino {
 namespace {
     constexpr char MAGIC_MESSAGE_SUFFIX[] = u8" \U000E0000";
-    constexpr int TITLE_REFRESH_PERIOD = 10;
+    constexpr int TITLE_REFRESH_PERIOD = 10000;
     constexpr int CLIP_CREATION_COOLDOWN = 5000;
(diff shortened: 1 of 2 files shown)


pajlada (approved):
:+1: LGTM

PR #2506, opened 2021-03-09
Fix workflow builds on ubuntu

zneix:
Pull request checklist:

# Description

Added newly required `libxcb-randr0-dev` to packages we need to install. Hopefully this fixes builds on Ubuntu :)


pajlada (approved):
Looks good to me! Thanks for the fix

PR #1982, opened 2020-09-24
Auto-refresh UserInfoPopup recent messages

dnsge:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description
Automatically add new messages to UserInfoPopup dialogs as they are received without having to press the refresh button. Clicking refresh will load the latest messages again from the channel.

<details>
  <summary>Video examples</summary>

[Example 1](https://i.nuuls.com/z3w_-.mp4)
[Example 2](https://i.nuuls.com/o517H.mp4)
</details>

PR #2514, opened 2021-03-12
Removed refresh label from usercard

zneix:
Pull request checklist:

no changelog entry needed

# Description

Ever since #1982 we refresh messages on the usercard automatically, without any need to click the button anymore, so I decided to remove it and it seems that nothing broke.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 - Minor: Added an option to only open channels specified in command line with `-c` parameter. You can also use `--help` to display short help message (#1940, #2368)
 - Minor: Added customizable timeout buttons to the user info popup
 - Minor: Deprecate loading of "v1" window layouts. If you haven't updated Chatterino in more than 2 years, there's a chance you will lose your window layout.
-- Minor: User popup will now automatically display messages as they are received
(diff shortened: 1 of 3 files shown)


Mm2PL (changes requested):
You removed the assignment to `UserInfoPopup::refreshConnection_` but you didn't remove it or its usage in the destructor. Is it still required but somehow used elsewhere in the application?
Also I'd consider this change significant enough for a change log update.

zneix:
> You removed the assignment to `UserInfoPopup::refreshConnection_` but you didn't remove it or its usage in the destructor. Is it still required but somehow used elsewhere in the application?

Good catch! I simply removed the function responsible for refreshing new messages and I haven't seen the assignment of `UserInfoPopup::refreshConnection_` in there, I assume it can be safely removed as there are no other references to it.

> Also I'd consider this change significant enough for a change log update.

I wanted to avoid adding very small things like this to a already HUGE pile of changelog entries, which already is quite overwhelming to be honest. I am still against a changelog entry since this is rather something that should be removed upon merging #1982. and has no longer any practical use what so ever.

pajlada (approved):
Looks good to me :+1:

PR #2523, opened 2021-03-14
Fix windows build not compiling

pajlada:
What we also ended up doing was ensuring that the chatterino2.exe file is actually present before continuing with uploading the artifact.

unfortunately, using `cmd` means we don't get any actual error codes, so this is the check we'll have to do.

If we were able to use powershell for the build command that'd be neater, but there's no clean way I know of to load the vcvars file from powershell

Pull request checklist:

# Description


pajlada:
Looks good to me! Thanks pajlada for the contribution!
image

PR #2524, opened 2021-03-14
🐛 Fixed windows CI not reporting exitcodes correctly

AnotherFoxGuy:
This uses https://github.com/ilammy/msvc-dev-cmd to setup the MSVC environment, and https://github.com/AnotherFoxGuy/conan-onefile for the installation of conan

-----------------------------------

Pull request checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,6 @@ jobs:
           version: ${{ matrix.qt-version }}
 
       # WINDOWS
-
       - name: Cache conan packages
         if: startsWith(matrix.os, 'windows')
         uses: actions/cache@v2.1.4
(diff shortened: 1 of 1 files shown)


pajlada:
`ilammy/msvc-dev-cmd@v1.6.0` seems like a nice solution for this, but I'll keep it for after the cmake PR has been merged into master

.github/workflows/build.yml
```diff
@@ -63,59 +63,49 @@ jobs:
...
 
-      - name: Install dependencies (Windows)
+      - name: Download conan
         if: startsWith(matrix.os, 'windows')
         run: |
-            choco install conan -y
-
-            refreshenv
-        shell: cmd
+            Invoke-WebRequest -Uri "https://github.com/AnotherFoxGuy/conan-onefile/releases/latest/download/conan.exe" -OutFile "C:\vcpkg\conan.exe"
```
pajlada, line 76: Let's revert these changes and keep install conan through choco if possible

pajlada (changes requested):

.github/workflows/build.yml
AnotherFoxGuy: The reason I changed it is because installing conan using choco takes anywhere between 3 to 8 minutes, and this method only takes 5 seconds
Also, this removes the need for a second "Add conan to path" action

pajlada: Yeah let's leave that for another time though - the windows CI is still not the one we have to wait for in the github actions

pajlada: maybe we could just cache the choco results instead?

pajlada: either way something for another PR

pajlada (approved):

PR #2528, opened 2021-03-15
fix: no longer crash on clip creation in "special" channels

### LINKED ISSUE #2527: Clip in '/mentions' crashes app
**Describe the bug**
Chatterino crashes when I trying make clip in '/mentions' panel

**To reproduce**
1) Open Chatterino
2) Open & select '/mentions'
3) Press Alt+X
4) Crash

**Chatterino version**
Chatterino Nightly 2.2.3-beta2 (commit 49bd1bd48, )

**Operating system**
Windows 10 Pro x64 (19042.804)

leon-richardt:
Pull request checklist:

# Description

Fixes #2527.

Clip creation via the `Alt+X` shortcut failed for the `/mentions` and `/whispers` channel. This happened because `Channel::isTwitchChannel` returns `true` for these channels. (Internally, the types are `Channel::Type::TwitchMentions` and `Channel::Type::TwitchWhispers` in anticipation of supporting more platforms in the future.)
This lead to a false assumption in the clip creation logic: Not all channels that display "Twitch content" are able to create clips.

The employed fix checks the channel type in more detail. Clip creation is only allowed if the channel is `Channel::Type::Twitch` or `Channel::Type::TwitchWatching`. This makes sure that there always is an associated channel to generate a clip from.

Additional note: Clip creation did not crash when initiated via the `/clip` command. This is due to the fact that `CommandController::execCommand` makes additional checks regarding the channel context before executing the command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Major: Added clip creation support. You can create clips with `/clip` command, `Alt+X` keybind or `Create a clip` option in split header's context menu. This requires a new authentication scope so re-authentication will be required to use it. (#2271, #2377)
+- Major: Added clip creation support. You can create clips with `/clip` command, `Alt+X` keybind or `Create a clip` option in split header's context menu. This requires a new authentication scope so re-authentication will be required to use it. (#2271, #2377, #2528)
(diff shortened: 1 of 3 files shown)


zneix (approved):
Good catch, I had no idea these channels still return `true` for `isTwitchChannel()` :+1:

pajlada (approved):

Issue #1912, opened 2020-08-24
Add support for reply threads


ALazyMeme:
**What should be added?**
Reply threads, the new things they have in webchat

**Why should it be added?**
It is a new feature added by twitch. Supporting it would be important to bring all the available features of vanilla Twitch chat to Chatterino.

**Additional resources**
https://twitter.com/WarthsYT/status/1293256740440608770
https://discuss.dev.twitch.tv/t/reply-threading/27657

PR #2533, opened 2021-03-16
Added a way to reply to messages

victti:
I'm not the best at coding and my knowledge on C++ is very limited, but I was waiting an update on chatterino that would add a reply feature for months now. So I added it myself, on a really simple way. I wish replies can be added soon to this, but for now I did my own solution.


TranRed:
I believe this might be the implementation of #1912

Mm2PL:
Does the link store the reply until the user presses send (like what Discord or webchat do)?


<details>
<summary>Discord's replies</summary>



</details>


> I believe this might be the implementation of #1912

I don't think this is a full implementation of replies since it's missing (from what I've seen) displaying reply threads of other users.

TranRed:
> > I believe this might be the implementation of #1912
> 
> I don't think this is a full implementation of replies since it's missing (from what I've seen) displaying reply threads of other users.

Just a trail of thought, because I couldn't test yet: 
So you would be seeing your own reply, but not replies to your own messages? 
Or only no replies if two other users are involved?
I imagine this to be confusing to users: "I can only see my replies; Must be a bug!"
Maybe it would be better to implement full reply-support than only half of it.

ALazyMeme:
Doesn't look like there's any further plans to work on this due to the author's fork being deleted. If anyone wants to somehow takeover, this can be reopened.

PR #2537, opened 2021-03-18
Fix app data dir not opening on macOS when "Open AppData" button pressed

### LINKED ISSUE #2531: macOS Open AppData dir broken
Clicking the "Open AppData directory" button does nothing on macOs Big Sur (and previous versions as well).

Should open finder at `$HOME/Library/Application Support/chatterino`

jammehcow:
Pull request checklist:

# Description

This PR aims to fix chatterino's app data directory not opening on macOS when pressing the "Open AppData..." button. When pressing the button on Windows/Linux the relevant directory is opened in the file explorer. When attempting to do this on macOS Finder does not open up the expected directory of `$HOME/Library/Application Support/chatterino`. 

This is caused by macOS requiring `file://` prior to the path in order for it to be opened in finder.

Fixes #2531

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@
 - Bugfix: Fixed hidden tooltips when always on top is active (#2384)
 - Bugfix: Fix CLI arguments (`--help`, `--version`, `--channels`) not being respected (#2368, #2190)
 - Bugfix: Fix Twitch cheer emotes not displaying tooltips when hovered (#2434)
+- Bugfix: Fix a crash on macOS when clicking "Open AppData Directory" in settings (#2531)
 - Dev: Updated minimum required Qt framework version to 5.12. (#2210)
 - Dev: Migrated `Kraken::getUser` to Helix (#2260)
 - Dev: Migrated `TwitchAccount::(un)followUser` from Kraken to Helix and moved it to `Helix::(un)followUser`. (#2306)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@
 - Bugfix: Fixed hidden tooltips when always on top is active (#2384)
 - Bugfix: Fix CLI arguments (`--help`, `--version`, `--channels`) not being respected (#2368, #2190)
 - Bugfix: Fix Twitch cheer emotes not displaying tooltips when hovered (#2434)
-- Bugfix: Fix a crash on macOS when clicking "Open AppData Directory" in settings (#2531)
+- Bugfix: Fix directory not opening when clicking "Open AppData Directory" setting button on macOS (#2531)
 - Dev: Updated minimum required Qt framework version to 5.12. (#2210)
 - Dev: Migrated `Kraken::getUser` to Helix (#2260)
 - Dev: Migrated `TwitchAccount::(un)followUser` from Kraken to Helix and moved it to `Helix::(un)followUser`. (#2306)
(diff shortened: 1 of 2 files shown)


jammehcow:
Current behaviour:
https://user-images.githubusercontent.com/4962764/111604626-6ce62680-87cd-11eb-819a-69b2b9511524.mp4

Expected behaviour:
https://user-images.githubusercontent.com/4962764/111604620-6bb4f980-87cd-11eb-86ea-cf18d954bf0b.mp4

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -455,7 +455,11 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addDescription("All local files like settings and cache files are "
                           "store in this directory.");
     layout.addButton("Open AppData directory", [] {
+#ifdef Q_OS_DARWIN
+        QDesktopServices::openUrl("file://" + getPaths()->rootAppDataDirectory);
+#else
         QDesktopServices::openUrl(getPaths()->rootAppDataDirectory);
+#endif
```
leon-richardt, line 462:
```suggestion
        QDesktopServices::openUrl("file://" + getPaths()->rootAppDataDirectory);
```

I think we could drop the `#ifdef` even, and always use the `file://` prefix on every OS (this is how it's done with the cache directory link.)
Tested on Linux with xdg-open/Nautilus, and seems to work fine. Would like additional tests on other OS and environments before going ahead with that though.

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@
 - Bugfix: Fixed hidden tooltips when always on top is active (#2384)
 - Bugfix: Fix CLI arguments (`--help`, `--version`, `--channels`) not being respected (#2368, #2190)
 - Bugfix: Fix Twitch cheer emotes not displaying tooltips when hovered (#2434)
+- Bugfix: Fix directory not opening when clicking "Open AppData Directory" setting button on macOS (#2531)
```
leon-richardt, line 84: I suggest also adding this PR to the entry:
```suggestion
- Bugfix: Fix directory not opening when clicking "Open AppData Directory" setting button on macOS (#2531, #2537)
```

leon-richardt (commented):
Can confirm that it does not introduce a regression on a Linux/xdg-open/Nautilus setup :+1: (Cannot test on macOS unfortunately)

Check review comments for two small suggestions

CHANGELOG.md
jammehcow: Resolved.

src/widgets/settingspages/GeneralPage.cpp
jammehcow: I suspect this wouldn't work on Windows since it's not a valid path format ([see this](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats)). If that's the case I don't mind reversing this so `file://` is prepended on all OSes bar Windows. 

QT might handle `file://` though so I'll test that tomorrow.

jammehcow: Thinking more like this if that's the case:

```c++
#ifdef Q_OS_WIN
        QDesktopServices::openUrl(getPaths()->rootAppDataDirectory);
#else
        QDesktopServices::openUrl("file://" + getPaths()->rootAppDataDirectory);
#endif

```

jammehcow: Tested this morning on Windows 10 Enterprise. Prefixing `file://` to the string throws this output:
```
ShellExecute 'file://c/Users/USER/AppData/Roaming/Chatterino2' failed (error 5).
```

Since most Linux distros are fine with either do we want to make this a darwin-specific prefix or make it default with a different case for Windows? Either way it's still an ifdef.

leon-richardt: Thanks for testing it 👍

If we need an `#ifdef` either way, I think it's fine to keep the code as it is right now.

jammehcow: Resolving this convo then, sweet as :+1:

pajlada (approved):
LGMT :+1: Thanks for the fix!

PR #2542, opened 2021-03-19
Added system message if no bttv emotes found instead of "unknown error"

MrAuro:
Pull request checklist:

# Description

If no BTTV/FFZ emotes were found a system message now shows saying unknown error. Now if status is 404, a system message will show saying it failed to fetch bttv/ffz emotes.
image


src/providers/bttv/BttvEmotes.cpp
```diff
@@ -182,6 +182,12 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
                     shared->addMessage(makeSystemMessage(
                         "This channel has no BetterTTV channel emotes."));
             }
+            else if (result.status() == 404)
+            {
+                // BTTV emotes not found
+                shared->addMessage(makeSystemMessage(
```
pajlada, line 188: 404 is an "expected error" in case the channel does not have any emotes. Checking for the error like you do is great, but I don't think we need to tell the users that no emotes are found.

src/providers/ffz/FfzEmotes.cpp
```diff
@@ -229,6 +229,12 @@ void FfzEmotes::loadChannel(
                     shared->addMessage(makeSystemMessage(
                         "This channel has no FrankerFaceZ channel emotes."));
             }
+            else if (result.status() == 404)
+            {
+                // FFZ emotes not found
+                shared->addMessage(makeSystemMessage(
```
pajlada, line 235: 404 is an "expected error" in case the channel does not have any emotes. Checking for the error like you do is great, but I don't think we need to tell the users that no emotes are found.

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@
 - Bugfix: Fixed hidden tooltips when always on top is active (#2384)
 - Bugfix: Fix CLI arguments (`--help`, `--version`, `--channels`) not being respected (#2368, #2190)
 - Bugfix: Fix Twitch cheer emotes not displaying tooltips when hovered (#2434)
+- Bugfix: Fix BTTV/FFZ channel emotes saying unknown error when no emotes found
```
pajlada, line 84: Please add the PR number at the end of your changelog entry (so #2542)

pajlada (changes requested):

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -175,7 +175,7 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
             auto shared = channel.lock();
             if (!shared)
                 return;
-            if (result.status() == 203)
+            if (result.status() == 203 || result.status() == 404)
```
zneix, line 178: Before #2304 was merged in, we had inconsistent error codes, thus according to [enum `QNetworkReply::NetworkError`](https://doc.qt.io/qt-5/qnetworkreply.html#NetworkError-enum), HTTP status code `404` used to be `203`, however after merging #2304, we no longer ever return `203`, so I believe this can be simplified
```suggestion
            if (result.status() == 404)
```

src/providers/ffz/FfzEmotes.cpp
```diff
@@ -222,7 +222,7 @@ void FfzEmotes::loadChannel(
             auto shared = channel.lock();
             if (!shared)
                 return;
-            if (result.status() == 203)
+            if (result.status() == 203 || result.status() == 404)
```
zneix, line 225: Before #2304 was merged in, we had inconsistent error codes, thus according to [enum `QNetworkReply::NetworkError`](https://doc.qt.io/qt-5/qnetworkreply.html#NetworkError-enum), HTTP status code `404` used to be `203`, however after merging #2304, we no longer ever return `203`, so I believe this can be simplified
```suggestion
            if (result.status() == 404)
```

zneix (changes requested):

pajlada (approved):

zneix (approved):
LGTM :+1:

PR #2543, opened 2021-03-19
Ensure twitch emote names are escaped

### LINKED ISSUE #995: Hovering over some monkey twitch emotes displays the regex that can be used to produce them, not the text that was used
chatterino_2019-04-06_12-53-59

zneix:
Pull request checklist:

# Description

Fixes an issue pointed out in https://github.com/Chatterino/Chatterino2/issues/995#issuecomment-802786567
Also closes #995 (?)


leon-richardt:
Tested, and works as expected :+1:

RAnders00:
Very nice fix 👍🏻 Thank you @zneix!

src/providers/twitch/TwitchEmotes.cpp
```diff
@@ -75,7 +75,7 @@ EmotePtr TwitchEmotes::getOrCreateEmote(const EmoteId &id,
     if (!shared)
     {
         (*cache)[id] = shared = std::make_shared<Emote>(Emote{
-            EmoteName{name},
+            EmoteName{this->cleanUpEmoteCode(EmoteName{name})},
```
leon-richardt, line 78:
```suggestion
            EmoteName{TwitchEmotes::cleanUpEmoteCode(EmoteName{name})},
```
Just a small suggestion which does not change any logic but is more conventional, I think. This just makes it more obvious to the reader that `TwitchEmotes::cleanUpEmoteCode` is a static method.

leon-richardt (changes requested):

src/providers/twitch/TwitchEmotes.cpp
zneix: Sure, I agree. I just thought that's it's better to use `this` since it's possible to do it in that context - making code a bit shorter.

pajlada (approved):

PR #2587, opened 2021-04-03
Add StartupWMClass property to desktop file

### LINKED ISSUE #1863: Arch Linux Gnome chatterino taskbar icon launches separate empty icon on the taskbar
**Describe the bug**

The chatterino desktop file icon that appears when searching for it on the application search menu is a proper shortcut to launch chatterino like you would through the terminal, and on the taskbar the icon is correct, but when you actually click the icon or launch the program it creates a separate empty icon on the taskbar that doesn't appear under a taskbar tab for the original shortcut. 
**To reproduce**

Launch the app through the terminal with "chatterino" and it will be an empty icon on the taskbar, or click on the shortcut on the taskbar for the same effect
**Screenshots**

Screenshot from 2020-08-14 18-09-10

To the left is the normal chatterino app icon, and the empty icon that i'm hovering over is what appears when the actual program launches
**Chatterino version**

Chatterino 2.1.7
**Operating system**

Arch Linux
**Additional information**

thanks pajlada forsenE

sheybey:
Pull request checklist:

- [ x ] `CHANGELOG.md` was updated, if applicable

# Description

This adds the StartupWMClass property to the desktop file, which window managers can use to determine when chatterino has opened and which window belongs to it.

GNOME uses this property, so this will fix #1863. You can test this by applying this patch to the desktop file of an existing install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,7 +85,7 @@
 - Bugfix: Fix BTTV/FFZ channel emotes saying unknown error when no emotes found (#2542)
 - Bugfix: Fix directory not opening when clicking "Open AppData Directory" setting button on macOS (#2531, #2537)
 - Bugfix: Fix quickswitcher not respecting order of tabs when filtering (#2519, #2561)
-- Bugfix: Fix GNOME not associating Chatterino's window with its desktop entry (#1863)
+- Bugfix: Fix GNOME not associating Chatterino's window with its desktop entry (#1863, #2587)
 - Dev: Updated minimum required Qt framework version to 5.12. (#2210)
 - Dev: Migrated `Kraken::getUser` to Helix (#2260)
 - Dev: Migrated `TwitchAccount::(un)followUser` from Kraken to Helix and moved it to `Helix::(un)followUser`. (#2306)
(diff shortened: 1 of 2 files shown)


sheybey:
Relevant FreeDesktop docs: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html#key-startupwmclass

pajlada (approved):

CHANGELOG.md
```diff
@@ -85,6 +85,7 @@
 - Bugfix: Fix BTTV/FFZ channel emotes saying unknown error when no emotes found (#2542)
 - Bugfix: Fix directory not opening when clicking "Open AppData Directory" setting button on macOS (#2531, #2537)
 - Bugfix: Fix quickswitcher not respecting order of tabs when filtering (#2519, #2561)
+- Bugfix: Fix GNOME not associating Chatterino's window with its desktop entry (#1863)
```
pajlada, line 88:
```suggestion
- Bugfix: Fix GNOME not associating Chatterino's window with its desktop entry (#1863, #2587)
```

pajlada:
Thanks for your contribution @sheybey !

PR #2588, opened 2021-04-03
Add link to video tutorial in README.md

bathtubdev:
Pull request checklist:

# Description

Hello. This is my first ever Pull Request.

- This adds a link to a video tutorial on YouTube that I created that shows how to download and install Chatterino Nightly on Windows, as long as the regular version of Chatterino is already installed.


pajlada (changes requested):
The contents of the PR is fine by me, however the video has a few things that I would like to see changed:

- Record this on a system that doesn't have winrar or 7zip installed
- Using the "Extract here" option should make it easier to just let the user overwrite the 'old' nightly when they download a new version, so it's probably preferred over the 'Extract to chatterino2-sdfhsdkfjhdsf' option.
- No need to advice the user how to move the installation to their Program Files folder - it's not a necessary part of the installation and may fail if the user does not have Administrative privileges.
- Instead of advising people to browse to the GitHub page directly, direct them to the chatterino.com website, press "Download" and click "Nightly Build" (NOTE: We are migrating to a new website very soon so it's probably worth holding off recording for a day, see https://new.chatterino.com for a preview of what it will look like)

Thank you for the initiative! :heart:

ALazyMeme:
Fork has been deleted, let me know if you want this re-opened.

PR #2590, opened 2021-04-03
always update dropdown on split button press

fourtf:
Pull request checklist:

# Description
Always update the menu when pressing the split hamburger button to account for changes because of mod/live status change.


pajlada (approved):
Looks good to me :+1:

PR #2592, opened 2021-04-04
Fix crash that could occur when the user changed the "Custom stream player URI Scheme" setting if the user had closed down and splits in the application runtime

pajlada:
- Manage connections for the customURIScheme setting
- Add changelog entry

Pull request checklist:

# Description

The connection listening to customURIScheme was not being handled properly - it now is

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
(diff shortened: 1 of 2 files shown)


zneix (approved):
I managed to reproduce the crash once on master, however after checking out to this pull request it doesn't happen anymore :+1:

Issue #1484, opened 2020-01-04
Collapsable chat tabs.


tolekk:
**Is your feature request related to a problem? Please describe.**
I hate how half of my chatterino screen is taken up with chat tabs, I want to be able to see more of the actual chat. Also I'm technically limited on the amount of chats I can join due to this.

**Describe the solution you'd like**
Make the chat tabs collapse into a scroll-down or a differently navigated menu so that more chat can fit on the screen. 

**Describe alternatives you've considered**
I have no alternatives on my mind at the moment.

**Additional context**
As most people are probably as lazy as I am, the chat adding menu is not perfect, you have to type in, or copy the whole name which takes more time than optimal in my opinion, that's why I keep all the tabs open. This feature is also helpful for people like me, who host multiple bots for people whose chats I want to keep open so people can ping me.
![image](https://user-images.githubusercontent.com/48994738/71765259-95e50900-2ef2-11ea-9034-5f3803b78d5f.png)

pajlada:
Related issue: #388 
Other ways of displaying tabs could also be explored, maybe a single line horizontal scrollable tab list reasonable (this would hide tabs, so another feature like CTRL+K would help)

apa420:
What about scrolling down and up rows on channels?

![](https://i.nuuls.com/Ds-No.png)

![](https://i.nuuls.com/dfLOA.png)
would be different scroll-things for example

tolekk:
As apa imagined it seems really good, one row that would resize depending on the window width that would be scrollable vertically or horizontally, but in groups of how many tabs fit on the screen - not how it is currently implemented, as I believe current single-tab scrolling is inefficient.

zneix:
@tolekk would hiding tabs be enough? I have a draft pr for it, you can check it out.

ALazyMeme:
I can see @zneix mentioned in his pr that this issue was related. Is there some part of this issue that the pr didn't fix?

pajlada:
Closing since I think #2600 solves for this use-case.

mjeveritt:
if you have a lot of channels, and your window is docked to the side of the screen, you can end up with this scenario: 
<img width="474" height="857" alt="Image" src="https://github.com/user-attachments/assets/10f81f59-d3c3-4807-82d4-fa012541d737" />
where the bottom set of channels is .. Missing .. (hidden).

would it be a total nightmare to add a scrollbar for the channels list/notebook/thingy?
[...]

PR #2600, opened 2021-04-07
Added ability to toggle visibility of tabs

### LINKED ISSUE #2313: Channel Tab Panel Minimization
### The Ability to Collapse/Minimize the Channel Tab Panel.

- There could be an option in the vertical ellipsis or directly left of the chat panel—maybe even a key combination—that can cause the chat panel to collapse towards the left which will hide all of the channel tabs until the chosen method is used again to reveal the channel tabs.

### Desktop Screen Management.

- I believe this feature should be added, or at least considered, because of how many favorite streamers people tend to have and how many of them they will have open; I have approximately fifty streamers (I know; it's probably a lot) I enjoy watching from time to time when my other favorite streamers are offline, and I tend to frequent some of their offline chats. Since I have so many channel tabs active, my Chatterino occupies almost half of my desktop's screen, and that space could be relinquished if I could somehow hide the panel for the channel tabs. While the channel tab panel is collapsed, there could be some sort of notification symbol on it that notifies us each time a streamer goes live until it is expanded.

Even if it's not to be considered, thanks for at least reading it. ღ

zneix:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable.
- [x] Toggle tab visibility with a shortcut (`Ctrl+U`).
- [x] Toggle tab visibility with a Split header context menu entry.
- [x] Show dialog upon hiding tabs for the first time.

# Description

This Pull Request adds an ability to toggle visibility of chat tabs - With a keyboard shortcut (`Ctrl + U`) and a Split header's context menu entry ("Toggle visibility of tabs").

### Open questions
1. Currently, (in both vertical and horizontal tab layouts) special buttons are still visible. Question is whether we gonna get rid of those if tabs are hidden or not. I personally say those *could* stay, but I'm not 100% sure if that's what we want.

<details>
<summary>Demonstration</summary>

hide_tabs_yunglpr

</details>

Related #1484
Closes #2313


jammehcow:
Implements feature of #2313

src/widgets/Notebook.cpp
```diff
@@ -477,51 +529,59 @@ void Notebook::performLayout(bool animated)
         int top = y;
         x = left;
 
+        //        if (this->showTabs_)
+        //        {
```
leon-richardt, line 533:
```suggestion

```

src/widgets/Notebook.cpp
```diff
@@ -542,6 +602,7 @@ void Notebook::performLayout(bool animated)
         {
             this->addButton_->raise();
         }
+        //        }
```
leon-richardt, line 605:
```suggestion

```

src/widgets/Window.cpp
```diff
@@ -356,7 +356,7 @@ void Window::addShortcuts()
 
     // New tab
     createWindowShortcut(this, "CTRL+SHIFT+T", [this] {
-        this->notebook_->addPage(true);
+        auto *newPage = this->notebook_->addPage(true);
```
leon-richardt, line 359:
```suggestion
        this->notebook_->addPage(true);
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -637,6 +637,8 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addCheckbox("Combine multiple bit tips into one", s.stackBits);
     layout.addCheckbox("Ask for confirmation when uploading an image",
                        s.askOnImageUpload);
+    layout.addCheckbox("Show a popup on hiding tabs",
+                       s.informOnTabVisibilityToggle);
```
leon-richardt, line 641: I'm wondering whether it really makes sense to add a setting for this. Showing the popup on first toggle is good IMO. After clicking "Don't show again", I don't think most users would like to re-enable the popup at some point in the future.
Curious to hear more opinions on this though

src/widgets/Notebook.cpp
```diff
@@ -325,6 +325,55 @@ void Notebook::setAllowUserTabManagement(bool value)
...
+            }
+            break;
+
+            case QMessageBox::YesRole:
+            default: {
+                getSettings()->informOnTabVisibilityToggle.setValue(false);
+            }
+            break;
+        }
+    }
```
leon-richardt, line 374:
```suggestion
    if (!value && getSettings()->informOnTabVisibilityToggle.getValue())
    {
        QMessageBox msgBox;
        msgBox.setText("Hide tabs");
        msgBox.setInformativeText(
            "You've just hidden your tabs.\nYou can (un)hide those with "
            "either keyboard shortcut (Default: Ctrl+U) or select an option "
            "in split header's menu.");
        msgBox.addButton(QMessageBox::Ok);
        const auto *dsaButton =
            msgBox.addButton("Don't show again", QMessageBox::YesRole);

        msgBox.setDefaultButton(QMessageBox::Ok);

        msgBox.exec();

        if (msgBox.clickedButton() == dsaButton)
            getSettings()->informOnTabVisibilityToggle.setValue(false);
    }
```
I think this is "more correct" (or at least easier to understand) since comparing the *result* of the message box to a standard button makes little sense conceptually.

leon-richardt (changes requested):
The keybinding `Ctrl+U` should also be added to the "Keybindings" page in the settings.

Other than that, I like the implementation. Feels nice to use from my perspective :)

src/widgets/Notebook.cpp
zneix: Sure I agree, didn't know you can use `.clickedButton()` for that

src/widgets/settingspages/GeneralPage.cpp
zneix: This was more of a setting for testing, however I just went with what we already had for a similar popup - the one that you see upon trying to paste the image to Chatterino.

mm added this lole

Maybe we should get rid of this as well? No idea, just wanted to be consistent

src/widgets/Window.cpp
zneix: No idea why was this changed, but sure why not.

src/widgets/settingspages/GeneralPage.cpp
leon-richardt: I think the use case of that popup is slightly different -- it gives you the ability to cancel an upload. In this case, the popup just informs the user of the action they are about to make.

leon-richardt (approved):

src/widgets/Notebook.cpp
```diff
@@ -325,6 +325,48 @@ void Notebook::setAllowUserTabManagement(bool value)
...
+    // show a popup upon hiding tabs
+    if (!value && getSettings()->informOnTabVisibilityToggle.getValue())
+    {
+        QMessageBox msgBox;
+        msgBox.window()->setWindowTitle("Chatterino - hidden tabs");
+        msgBox.setText("Hide tabs");
+        msgBox.setInformativeText(
+            "You've just hidden your tabs.\nYou can (un)hide those with "
+            "either keyboard shortcut (Default: Ctrl+U) or select an option "
+            "in split header's menu.");
```
jammehcow, line 354: This wording feels a bit odd - especially if you're a new chatterino user. I suggest the following:

```
You've just hidden your tabs!
You can toggle tabs by using the keyboard shortcut (Ctrl+U by default) or clicking "toggle visibility of tabs" in the dropdown next to the stream title.
```

zneix: Thank you very much for feedback from a native english speaker :heart:

jammehcow (changes requested):
Wording changes could be made.

jammehcow (approved):

jammehcow:
Dialog shows up each time CTRL + U is pressed to hide.

Shouldn't it be a one-time thing?

src/widgets/splits/SplitHeader.hpp
```diff
@@ -2,6 +2,8 @@
 
 #include "widgets/BaseWidget.hpp"
 
+#include "widgets/Window.hpp"
```
pajlada, line 5:
```suggestion
#include "widgets/Window.hpp"
```

pajlada:
Functionally the pull request works great for me :ok_hand: 

The interface for toggling visibility of tabs is in the wrong place - it's been put into the Split Header menu which should only control Split/Channel-specific things.

Hotkey is fine, but some other way of accessing it should still be available.

I have heard two different ideas so far:
1) Right clicking tabs/next to the tabs can bring up a menu which allows you to hide/unhide all tabs. This would be possible both when tabs are hidden and not hidden, since a minimu. amount of 'blank space' is always there (Not true if preferences & user button are hidden in Vertical tab mode)

2) Add a new icon next to the Cogwheel/Accounts icons for toggling visibility of tabs. While this is also a simple way to solve it, I'm not 100% sure of adding such a button for a '1% of users feature'.

leon-richardt:
> 2) Add a new icon next to the Cogwheel/Accounts icons for toggling visibility of tabs. While this is also a simple way to solve it, I'm not 100% sure of adding such a button for a '1% of users feature'.

I would much prefer the other solution

zneix:
> Hotkey is fine, but some other way of accessing it should still be available.
> 
> I have heard two different ideas so far:
> 
>     1. Right clicking tabs/next to the tabs can bring up a menu which allows you to hide/unhide all tabs. This would be possible both when tabs are hidden and not hidden, since a minimu. amount of 'blank space' is always there (Not true if preferences & user button are hidden in Vertical tab mode)
> 
>     2. Add a new icon next to the Cogwheel/Accounts icons for toggling visibility of tabs. While this is also a simple way to solve it, I'm not 100% sure of adding such a button for a '1% of users feature'.

As per @leon-richardt's advice I went ahead, and tried my best to improve the UX by moving the context menu entry from split header to NotebookTabs' and also created a new menu with an entry toggling visibility of tabs (for when tabs are hidden).

Hoping this solution is better and that pull request can be merged in soon™

src/widgets/Notebook.cpp
```diff
@@ -325,6 +330,48 @@ void Notebook::setAllowUserTabManagement(bool value)
...
+
+    // show a popup upon hiding tabs
+    if (!value && getSettings()->informOnTabVisibilityToggle.getValue())
+    {
+        QMessageBox msgBox;
+        msgBox.window()->setWindowTitle("Chatterino - hidden tabs");
+        msgBox.setText("You've just hidden your tabs");
+        msgBox.setInformativeText(
+            "You can toggle tabs by using the keyboard shortcut (Ctrl + U by "
+            "default) or right-clicking on tab area and selecting \"Toggle "
```
leon-richardt, line 358:
```suggestion
            "default) or right-clicking on the tab area and selecting \"Toggle "
```
Super pedantic but I couldn't resist :grimacing:

leon-richardt (changes requested):
I tested it and other than this little nitpick, looks good and works just as expected :ok_hand:

src/widgets/Notebook.cpp
zneix: Nitpicks like this are supa cool and useful, no joke. Also pls approve once I fix this.

leon-richardt (approved):

jammehcow (approved):
I couldn't break it, looks good to me :+1:

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added ability to toggle visibility of Channel Tabs - with keyboard shortcut (default: Ctrl+U). (#2600)
```
pajlada, line 5: Update changelog to clarify new UX functionality

```suggestion
- Major: Added ability to toggle visibility of Channel Tabs - This can be done by right-clicking the tab area or pressing the keyboard shortcut (default: Ctrl+U). (#2600)
```

alazymeme:
Works fine on Windows 21H1, leaves a little space, but it appears to be windows related:

Showing tabs:
image

Without tabs:
image

If you have quite a few more tabs, it will minimize down to just a single blank line (as shown in the second screenshot).

pajlada (changes requested):
Small changelog entry nitpick, other than that looks good :ok_hand:

alazymeme:
Found this weird bug too, only seems to occur on windows?

jif

pajlada:
@alazymeme Could you test 'tab area' width/height in horizontal/vertical mode in the latest build please?

alazymeme:
No :tf:

alazymeme:
Line is still there, looks a bit weird now that there's random space on the side (since the space was not there before and the line is evidently moved upon the "creation" of the space)

Bug outlined in https://github.com/Chatterino/chatterino2/pull/2600#issuecomment-850808555
image

Random space:
jif

pajlada:
gotta review paint event to ensure it's aligned with the 'empty space' logic

pajlada:
@alazymeme Once the Windows build has succeeded, could you please take this for one (hopefully last) spin? Thank you!

Known 'quirks': Swapping between horizontal and vertical tab layout will result in the line around buttons (if you have buttons enabled) to not fully settle until you have pressed ok in the settings dialog.

zneix:
I (sadly) tested this on Windows and this seems fine
this shit os is so slow forsenY

alazymeme (approved):
Yes, this works perfectly.

pajlada (approved):
LGTM :+1: 
Would be good to have the phrase "tab area" added to [the glossary](https://wiki.chatterino.com/Glossary/)

Alicyl:
Sorry to intrude upon this lovely little thread, but where can one access the version of Chatterino with this feature (or when can we expect if it's inaccessible to the public)? Nightly builds?

ALazyMeme:
> Nightly builds?

Yes

ichigo-gyuunyuu:
Hello, is there a way to have the tabs be hidden by default on launch and hitting Ctrl + U toggles visibility like usual. Tried to find search the settings but couldn't find any option :/

pajlada:
> Hello, is there a way to have the tabs be hidden by default on launch and hitting Ctrl + U toggles visibility like usual. Tried to find search the settings but couldn't find any option :/

There is not, no

PR #2605, opened 2021-04-08
clang-tidy: use std::move where applicable

Yoitsu:
Pull request checklist:

# Description

Previous PR split into multiple parts, part 1.

std::move added where applicable as per suggested by clang-tidy, and 1 case removed where it was added in error to a trivially copyable object. These are "confident" changes, as applying std::move to non-trivially copyable objects should only result in less temporary copies -> lower overhead -> higher perfomance, unlike const& which can also change code semantics (according to Pajlada's feedback, as apparently const& can degrade perfomance in lambda contexts).


pajlada (approved):

fourtf:
In many of these cases why not pass in the objects as references instead of moving them? seems sus to me

fourtf:
doesn't really matter though

PR #2615, opened 2021-04-11
Right click to add emote to block list

fraserbaigent:
Pull request checklist:

# Description

I use this on my own fork so i can quickly ignore messages (usually spam) containing gifs of weird dancing anime girls.


src/widgets/helper/ChannelView.cpp
```diff
@@ -103,6 +104,12 @@ namespace {
                                 });
         };
 
+        // Ignore messages with this emote (useful for weeb emotes)
```
zneix, line 107: Keep it professional
```suggestion
        // Add current emote to the list of ignored messages
```

zneix (changes requested):
You should also include a changelog entry in `CHANGELOG.md`
Also, there should be probably a check for whether the term is already ignored to avoid flooding your ignores list.

src/widgets/helper/ChannelView.cpp
fraserbaigent: Added guard against multiple additions, really you shouldn't see it once it's blocked but you can still see messages sent before you blocked it; so I also added an unblock command in case you did it accidentally in the same chat.

Had a look around the codebase and couldn't see anything in `SignalVector` to remove by value, I spun up something in the lambda but if this already exists I can just change it.

fraserbaigent:
Added guard against multiple additions, really you shouldn't see it once it's blocked but you can still see messages sent before you blocked it; so I also added an unblock command in case you did it accidentally in the same chat.

Had a look around the codebase and couldn't see anything in `SignalVector` to remove by value, I spun up something in the lambda but if this already exists I can just change it.

src/widgets/helper/ChannelView.cpp
```diff
@@ -103,6 +104,34 @@ namespace {
...
+        {
+            menu.addAction(
+                "Unignore this emote", [name = emote.getCopyString()] {
+                    auto &ignored_messages =
+                        *getSettings()->ignoredMessages.readOnly();
+                    const auto it = std::find_if(
+                        ignored_messages.cbegin(), ignored_messages.cend(),
+                        [&](const IgnorePhrase &phrase) {
+                            return name == phrase.getPattern();
+                        });
```
zneix, line 125: Three things:

1. We should avoid reusing variable names to avoid confusion, just change `name` in this scope to something like `emoteName`
2. Use regex to avoid situations, where one emote could block another. (`e.g. Blocking `FeelsDankMan` would also block `FeelsDankManLostHisHat`).
3. Use `QString()` instead of `""`

```suggestion
        QString emoteName(emote.getCopyString());
        QString emoteIgnoreRegex("(\\s|^)%1(\\s|$)");
        if (!getSettings()->isIgnoredMessage(emoteIgnoreRegex.arg(emoteName)))
        {
            menu.addAction("Ignore this emote", [emoteName, emoteIgnoreRegex] {
                getSettings()->ignoredMessages.append(
                    IgnorePhrase(emoteIgnoreRegex.arg(emoteName), true, true,
                                 QString(), true));
            });
        }
        else
        {
            menu.addAction(
                "Unignore this emote", [emoteName, emoteIgnoreRegex] {
                    auto &ignored_messages =
                        *getSettings()->ignoredMessages.readOnly();
                    const auto it = std::find_if(
                        ignored_messages.cbegin(), ignored_messages.cend(),
                        [&](const IgnorePhrase &phrase) {
                            return emoteIgnoreRegex.arg(emoteName) ==
                                   phrase.getPattern();
                        });
```

zneix (changes requested):

src/widgets/helper/ChannelView.cpp
fraserbaigent: 1 & 3 a bit nitpicky but if you've written the code for me I've no issues :smile:

jammehcow (changes requested):
Would be good to view a list of ignored emotes for them to be easily unignored.

fraserbaigent:
@jammehcow This exists already in the settings menu under blocked emotes, this popup menu command is really just a quick UX way of adding an emote to that list.

This branch is getting a bit stale now, any chance of a merge any time soon? I assume changelog can be done manually at merge time or adjusted and squished as appropriate

alazymeme:
@pajlada 🔔

pajlada (changes requested):
I don't think the right click menu option aligns with what it actually does. "Ignore this emote" to me would just remove the emote from messages, not block the entire message if a message contains the emote.

I'd be ok with the behaviour matches the current phrasing, so the emote image and text would be blocked but the message would be allowed.

PR #2621, opened 2021-04-12
Fixed a semicolon issue

zneix:
Pull request checklist:

# Description

Fixed a colon that was there instead of a semicolon for whatever reason.


pajlada:
I believe the code is good as is (i.e. this PR is unnecessary).

The documentation (correctly) states:
 > Extra headers, that will be included in the request. Header name and value must be separated by colon (:). Multiple headers need to be separated with semicolons (;).
Example: `Authorization: supaKey ; NextHeader: value`.

Each header pair is split on semicolon `;` in https://github.com/Chatterino/chatterino2/blob/master/src/util/NuulsUploader.cpp#L131 and the function you changed is responsible for splitting each header pair into a header name (e.g. Authorization) and a header value (e.g. supaKey) with colon `:`

zneix:
Yes turns out I had a very big brain lag while creating this.

PR #2623, opened 2021-04-12
Add a function in NetworkCommon `parseHeaderList` which parses a header list as a string into a vector of header pairs

pajlada:
Pull request checklist:

# Description

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -134,6 +134,7 @@ SOURCES += \
     src/common/Env.cpp \
     src/common/LinkParser.cpp \
     src/common/Modes.cpp \
+    src/common/NetworkCommon.cpp \
     src/common/NetworkManager.cpp \
     src/common/NetworkPrivate.cpp \
     src/common/NetworkRequest.cpp \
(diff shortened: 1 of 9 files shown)


Bun:
The code suggests colons are not allowed in the header values, but [RFC7230 section 3.2](https://tools.ietf.org/html/rfc7230#section-3.2) suggests otherwise. This would prevent e.g. authorization tokens that use `:` somewhere.

pajlada:
> The code suggests colons are not allowed in the header values, but [RFC7230 section 3.2](https://tools.ietf.org/html/rfc7230#section-3.2) suggests otherwise. This would prevent e.g. authorization tokens that use `:` somewhere.

Thanks for the heads up - I'll modify the code and add some tests for it

pajlada:
> The code suggests colons are not allowed in the header valu

I believe the latest commit remedies this, the limitation I've put on this now instead is that the value, after trimming, must not be empty.
So `Authorization:` is invalid

pajlada:
Code ready for review now, if you could take another peek @Bun that'd be great in addition to :snake: 's eyes :eyes:

zneix (approved):
No idea why did I put off this review, good stuff :+1:

pajlada:
Will update branch and merge this one in after #2634

PR #2628, opened 2021-04-15
Added custom FrankerFaceZ VIP badges

### LINKED ISSUE #2627: Custom VIP Badges (New FrankerFaceZ Feature)
**What should be added?**
Hello, friends! Yesterday, FrankerFaceZ added a feature that allows broadcasters to set a custom image for VIP badges in their chat. Example:
image

These custom badges work like a normal Twitch custom badge, such as sub badges. Just a total image replacement, no background colors or anything like the special way FFZ handles mod badges.

If a channel has a custom VIP badge, there will be URLs for the badge in the APIv1 room response from FrankerFaceZ. For example, from https://api.frankerfacez.com/v1/room/sodapoppin we see:

```json
{
	"room": {
		"id": "sodapoppin",
		...
		"vip_badge": {
			"1": "//cdn.frankerfacez.com/room-badge/vip/sodapoppin/1",
			"2": "//cdn.frankerfacez.com/room-badge/vip/sodapoppin/2",
			"4": "//cdn.frankerfacez.com/room-badge/vip/sodapoppin/4"
		},
		...
	}
}
```

In a channel with no custom VIP badge set, `vip_badge` will be `null`. Additionally, if users don't supply a high enough resolution image, `2` or `4` may be null to indicate their absence. In the future, we'll have our CDN just upscale things on demand, but that isn't happening yet so please don't rely on behavior like that.

**Why should it be added?**
I personally think Twitch should've let people customize their VIP badges from the start, but since they didn't, FFZ is here to help. I believe adding support for this to many different clients will help present a more unified experience to Twitch users, and that is worth it.

Thanks!

zneix:
Pull request checklist:

# Description

your API is shit

Closes #2627

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1132,7 +1132,7 @@ void TwitchMessageBuilder::appendTwitchBadges()
         {
             if (auto customVipBadge = this->twitchChannel->ffzCustomVipBadge())
             {
-                this->emplace<ModBadgeElement>(
+                this->emplace<VipBadgeElement>(
                         customVipBadge.get(),
                         MessageElementFlag::BadgeChannelAuthority)
                     ->setTooltip((*customVipBadge)->tooltip.string);
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1128,6 +1128,18 @@ void TwitchMessageBuilder::appendTwitchBadges()
                 continue;
             }
         }
+        else if (badge.key_ == "vip")
+        {
+            if (auto customVipBadge = this->twitchChannel->ffzCustomVipBadge())
+            {
+                this->emplace<ModBadgeElement>(
+                        customVipBadge.get(),
+                        MessageElementFlag::BadgeChannelAuthority)
+                    ->setTooltip((*customVipBadge)->tooltip.string);
+                // early out, since we have to add a custom badge element here
+                continue;
+            }
+        }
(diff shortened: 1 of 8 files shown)


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1128,6 +1128,18 @@ void TwitchMessageBuilder::appendTwitchBadges()
                 continue;
             }
         }
+        else if (badge.key_ == "vip")
+        {
+            if (auto customVipBadge = this->twitchChannel->ffzCustomVipBadge())
+            {
+                this->emplace<ModBadgeElement>(
```
Mm2PL, line 1135: a bit pedantic: `ModBadgeElement` just feel weird to me here after all this is handling VIP badges.

zneix: Yes, I am aware of this - I will rework all references to custom mod badges and rename those to something along "authority badge" (since both badges use the same channel authority slot). This pull request is basically a pure copyasta, so I think it's better to have less duplicated code by reworking already existing stuff. This is also why the PR is a draft for now. I'll fix it tommorow because I'm tired now.

zneix: Fixed in a23eb20

zneix:
Since these VIP badges are a bit different from mod badges (that were just white pixels on green background) - they are custom 72x72 images with no required background or anything, this means it could be sometimes difficult to distinguish between those and other badges like subscriber ones. Should we add a setting for those?

Mm2PL (approved):
LGTM, looks and works as expected.

pajlada (approved):
Looks good to me! Will rebase master onto this branch and merge it in after #2623

pajlada (approved):

PR #2629, opened 2021-04-15
Added confirmation for usercard actions

### LINKED ISSUE #1886: A confirmation for potentially unwated usercard actions
**What should be added?**
A confirmation box when following/unfollowing someone.

**Why should it be added?**
not to follow or unfollow people by accident

MrAuro:
Pull request checklist:

# Description

Asks the user if they want to follow/unfollow or block/unblock someone when in the usercard
Closes #1886.

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -426,12 +426,19 @@ void UserInfoPopup::installEvents()
             switch (newState)
             {
                 case Qt::CheckState::Unchecked: {
-                    this->ui_.follow->setEnabled(false);
-                    getHelix()->unfollowUser(currentUser->getUserId(),
-                                             this->userId_,
-                                             reenableFollowCheckbox, [] {
-                                                 //
-                                             });
+                    auto box = new QMessageBox(
+                        QMessageBox::Information, "Chatterino 2",
+                        "Unfollow user?", QMessageBox::Yes | QMessageBox::No);
+                    box->setAttribute(Qt::WA_DeleteOnClose);
+                    if (box->exec() == QMessageBox::Yes)
+                    {
+                        this->ui_.follow->setEnabled(false);
+                        getHelix()->unfollowUser(currentUser->getUserId(),
+                                                 this->userId_,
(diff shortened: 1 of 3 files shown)


MrAuro:
I found that this crashes Chatterino when any action is confirmed. I'm not sure what causes this, I am new to QT and c++.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -426,12 +426,19 @@ void UserInfoPopup::installEvents()
...
-                    getHelix()->unfollowUser(currentUser->getUserId(),
-                                             this->userId_,
-                                             reenableFollowCheckbox, [] {
-                                                 //
-                                             });
+                    auto box = new QMessageBox(
+                        QMessageBox::Information, "Chatterino 2",
+                        "Unfollow user?", QMessageBox::Yes | QMessageBox::No);
+                    box->setAttribute(Qt::WA_DeleteOnClose);
+                    if (box->exec() == QMessageBox::Yes)
```
jammehcow, line 433: Not a QT guru but I suspect blocking the UI with this call is your cause. Try using `QObject::connect` to connect to the `clicked` signal. [See here](https://github.com/Chatterino/chatterino2/blob/cb56e414a4fd1b90b25bc03ef5c7de2daeb9ae3d/src/widgets/dialogs/UpdateDialog.cpp#L28) for an example.

QT devs correct me if I'm wrong.

jammehcow: This applies to all 4 usages of `exec()`

fourtf:
Closing the popup and showing a message box doesn't seem like the best UX to me. How about making a button that needs to be pressed twice instead?

Here's a mock-up of what I mean:
image

MrAuro:
Since the create and delete follow endpoints are being deprecated I'm going to close this PR (https://discuss.dev.twitch.tv/t/deprecation-of-create-and-delete-follows-api-endpoints/32351). Went on vacation and completely forgot I made this. I might open another PR with the intended functionality for blocking a user if I have time and learn enough QT & C++ to do so 😄

PR #2633, opened 2021-04-16
Added missing default twitch commands

zneix:
Pull request checklist:

# Description

Added `/vips`.
By the way, we do include standard moderator commands in `TWITCH_DEFAULT_COMMANDS` define, but not the ones available to channel editors (such as `/commercial`, `/host`, `/raid`). Should we add those?

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -28,10 +28,10 @@
 #include <QFile>
 #include <QRegularExpression>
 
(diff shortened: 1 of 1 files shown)


pajlada:
> By the way, we do include standard moderator commands in `TWITCH_DEFAULT_COMMANDS` define, but not the ones available to channel editors (such as `/commercial`, `/host`, `/raid`). Should we add those?

Yes

pajlada (approved):
LGTM :+1:

PR #2634, opened 2021-04-16
Added in:<channels> search predicate

### LINKED ISSUE #2299: Search for channels in /mentions tab
Give us the ability to search for specific channels in the mentions tab: #channelname
(When I get mentioned in a lot of random channels and I only want to search for specific ones out of the mix, I have no easy way of doing it)
CTRL + F => #channelname

talneoran:
Pull request checklist:

# Description
See #2299 for more information.
This PR adds a search predicate (according to #1237) that allows filtering messages based on the channel they were sent in, meant for usage in /mentions. 
This is also inspired by the similar filter in [Discord](https://support.discord.com/hc/en-us/articles/115000468588-Using-Search).
Example usage:
`in:fourtf` would show messages originally sent in fourtf channel.
`in:fourtf,pajlada` would show messages originally sent in either fourtf or pajlada channels.

The things I'm not completely sure about now are:

1. Currently the filter works everywhere, but is only useful in /mentions. If used in a regular split it will either show all messages or none, depending on whether or not the channel list includes the current channel. I'm wondering if there is a need to enforce this filter only being useable in /mentions, or if the current state is fine (at least to me the current functionality makes sense). 
2. The wording of the changelog entry / explanation of the feature.

Closes #2299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added `in:<channels>` search filter to find messages sent in specific channels. (#2299, #2634)
+
 ## 2.3.0
 
 - Major: Added clip creation support. You can create clips with `/clip` command, `Alt+X` keybind or `Create a clip` option in split header's context menu. This requires a new authentication scope so re-authentication will be required to use it. (#2271, #2377, #2528)
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -179,6 +179,7 @@ SOURCES += \
     src/messages/MessageContainer.cpp \
     src/messages/MessageElement.cpp \
     src/messages/search/AuthorPredicate.cpp \
+    src/messages/search/ChannelPredicate.cpp \
     src/messages/search/LinkPredicate.cpp \
     src/messages/search/SubstringPredicate.cpp \
     src/messages/SharedMessageBuilder.cpp \
@@ -406,6 +407,7 @@ HEADERS += \
     src/messages/MessageElement.hpp \
     src/messages/MessageParseArgs.hpp \
     src/messages/search/AuthorPredicate.hpp \
+    src/messages/search/ChannelPredicate.hpp \
     src/messages/search/LinkPredicate.hpp \
     src/messages/search/MessagePredicate.hpp \
     src/messages/search/SubstringPredicate.hpp \
--- a/src/widgets/helper/SearchPopup.cpp
+++ b/src/widgets/helper/SearchPopup.cpp
@@ -165,10 +165,10 @@ std::vector<std::unique_ptr<MessagePredicate>> SearchPopup::parsePredicates(
 {
     static QRegularExpression predicateRegex(R"(^(\w+):([\w,]+)$)");
 
-    auto predicates = std::vector<std::unique_ptr<MessagePredicate>>();
+    std::vector<std::unique_ptr<MessagePredicate>> predicates;
     auto words = input.split(' ', QString::SkipEmptyParts);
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added `in:<channels>` search filter to find messages sent in specific channels. (#2299, #2634)
```
leon-richardt, line 5: Only issue I could potentially see with the way this is phrased, is a possible confusion with [Filters](https://wiki.chatterino.com/Filters/). We could consider changing it as follows:
```suggestion
- Minor: Added `in:<channels>` search predicate to find messages sent in specific channels. (#2299, #2634)
```
Looking for additional input though, as the phrase "search predicate" is not really communicated to the user anywhere in the app. I have opened https://github.com/Chatterino/wiki/issues/58 to address this.

chatterino.pro
```diff
@@ -406,6 +407,7 @@ HEADERS += \
     src/messages/MessageElement.hpp \
     src/messages/MessageParseArgs.hpp \
     src/messages/search/AuthorPredicate.hpp \
+    src/messages/search/ChannelPredicate.hpp \
```
leon-richardt, line 410: You will also need the new files to the `CMakeLists.txt` in order to support both build systems. (Also resolves CI issue)

src/widgets/helper/SearchPopup.cpp
```diff
@@ -167,6 +168,7 @@ std::vector<std::unique_ptr<MessagePredicate>> SearchPopup::parsePredicates(
     auto predicates = std::vector<std::unique_ptr<MessagePredicate>>();
     auto words = input.split(' ', QString::SkipEmptyParts);
     auto authors = QStringList();
+    auto channels = QStringList();
```
leon-richardt, line 171: I think I would prefer `QStringList channels;` instead but keeping it consistent with the lines above is also fine by me.
(Really, the lines above should be changed probably)

leon-richardt (commented):
Tested for a few simple use cases, and works as expected. :+1: I think this is a great use case for message search predicats (although I am probably a little biased :slightly_smiling_face:)

-------

> I'm wondering if there is a need to enforce this filter only being useable in /mentions, or if the current state is fine (at least to me the current functionality makes sense).

I think the current state is fine. The behavior makes sense, and restricting it to `/mentions` now might only limit future use cases

CHANGELOG.md
talneoran: Agreed about the possible confusion, the only reason I didn't write predicate is because I realized users are currently unaware of the term. I can wait for a decision on how to word this.

src/widgets/helper/SearchPopup.cpp
talneoran: Yeah I pretty much went along with how the original change was done. As you said might as well change the lines above too, but would I do this as a part of this PR?

leon-richardt: Can't blame for you for code you didn't write (or for sticking to the convention established in the same file) so I think either way is fine. Just go with what you feel most comfortable with :+1:

CHANGELOG.md
pajlada: I'm fine with keeping the term 'filter' in public terminology, considering we're basing this feature heavily on Discord's search feature, they seem to have originally called them 'search options' but now refer to them as filters https://support.discord.com/hc/en-us/articles/115000468588

pajlada (approved):
LGTM :+1: Will update branch and merge in after #2633

PR #2643, opened 2021-04-18
Keyboard integration for streamlink quality confirmation

M4tthewDE:
Pull request checklist:

# Description

This pr allows the confirmation of the streamlink quality selection by hitting Enter, without having to tab to the "Ok"-Button. 

Not experienced with qt, so feedback appreciated :)

### DIFF
--- a/src/widgets/dialogs/QualityPopup.cpp
+++ b/src/widgets/dialogs/QualityPopup.cpp
@@ -1,5 +1,4 @@
 #include "QualityPopup.hpp"
-#include <qnamespace.h>
 #include "Application.hpp"
 #include "common/QLogging.hpp"
 #include "singletons/WindowManager.hpp"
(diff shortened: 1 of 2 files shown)


src/widgets/dialogs/QualityPopup.cpp
```diff
@@ -46,6 +46,18 @@ void QualityPopup::showDialog(const QString &channelName, QStringList options)
...
+{
+    if (event->key() == Qt::Key_Enter || event->key() == Qt::Key_Return)
+    {
+        this->okButtonClicked();
+    }
+    else if (event->key() == Qt::Key_Escape)
+    {
+        this->cancelButtonClicked();
+    }
+}
```
Mm2PL, line 59: Use `createWindowShortcut` instead

Mm2PL (changes requested):

zneix:
I think this is such a small thing that it doesn't necessarily require a changelog entry.

zneix:
I think instead of adding a shortcut for Enter, rather changing the default button or something along these lines should be done instead. That'd be consistent with all other button behavior's among the codebase and cleaner in my opinion. I'll to fiddle around this one for a little bit now.
Just making a comment before pajlada gets to see this PR.

pajlada:
Fixed in #3169

PR #2647, opened 2021-04-19
Fixed deprecated method QWheelEvent::delta

zneix:
And some other method that's also replaced with QWheelEvent::angleDelta

Pull request checklist:

# Description

Reference: https://doc.qt.io/qt-5/qwheelevent-obsolete.html#delta and https://doc.qt.io/qt-5/qwheelevent-obsolete.html#orientation

### Changes in behavior

Change from `event->delta()` to `event->angleDelta().y()` makes it, so you can no longer scroll horizontally (with trackpad / touchpad) to select next/previous tab (until now, you were able to do it, but I believe this is wrong anyways).

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1098,7 +1098,7 @@ void ChannelView::drawMessages(QPainter &painter)
 
 void ChannelView::wheelEvent(QWheelEvent *event)
 {
-    if (event->angleDelta().x())
+    if (!event->angleDelta().y())
     {
         return;
     }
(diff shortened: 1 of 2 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -1098,8 +1098,10 @@ void ChannelView::drawMessages(QPainter &painter)
 
 void ChannelView::wheelEvent(QWheelEvent *event)
 {
-    if (event->orientation() != Qt::Vertical)
+    if (event->angleDelta().x())
```
pajlada, line 1101:
```suggestion
    if (!event->angleDelta().y())
```

If we only care about the vertical scroll value, maybe it makes more sense to check if there's any vertical scroll value rather than checking for the absence of a horizontal scroll value

PR #2658, opened 2021-04-22
Replace #define in CommandController with QStringList

zneix:
Pull request checklist:

# Description

Reference: https://clips.twitch.tv/SleepyLovelyDotterelMau5-ef2JGpYn8biLhxs8

Makes code more clean. This wall of strings may be quite big, but it integrates well with git diffs since it's all in new lines instead.


pajlada (approved):
LGTM :+1:

PR #2668, opened 2021-04-23
Add ability to hide Twitch Prediction badges

### LINKED ISSUE #2664: Option to hide prediction badges in chat
**What should be added?**
A new option in settings to hide the prediction badges

**Why should it be added?**
Would really help to clean up the chat during predictions and is afaik the only badge thats not hideable right now

zneix:
Pull request checklist:

# Description

This makes predictions badges being acknowledged as an actual separate message element on its own, thus making it possible to configure those (which I also added in this PR).

Examples of messages with those badges: 
```
@badge-info=predictions/No;badges=predictions/pink-2,vip/1;color=#DAA520;display-name=Somso2e;emotes=;flags=;id=cc518b04-801a-473d-85c4-af139d7dd520;mod=0;room-id=99631238;subscriber=0;tmi-sent-ts=1619164205224;turbo=0;user-id=153573952;user-type= :somso2e!somso2e@somso2e.tmi.twitch.tv PRIVMSG #zneix :-tags
```
```
@badge-info=predictions/Yes;badges=predictions/blue-1;color=#0000FF;display-name=Hpixpoke;emotes=;flags=;id=02cfc1c1-8ce8-4d7b-b9e7-64b6d8da1a30;mod=0;room-id=99631238;subscriber=0;tmi-sent-ts=1619164206713;turbo=0;user-id=108748290;user-type= :hpixpoke!hpixpoke@hpixpoke.tmi.twitch.tv PRIVMSG #zneix :-tags
```

I also reordered MessageElementFlag enums and added some missing comments regarding Chatterino badges (no idea whether we should keep those really or not - `10:39:34.373 pajlada: I wonder if we need to keep a list of badges in general`).

Closes #2664

### DIFF
--- a/src/messages/MessageElement.hpp
+++ b/src/messages/MessageElement.hpp
@@ -44,58 +44,65 @@ enum class MessageElementFlag : int64_t {
     ChannelPointReward = (1LL << 8),
     ChannelPointRewardImage = ChannelPointReward | TwitchEmoteImage,
 
-    FfzEmoteImage = (1LL << 10),
-    FfzEmoteText = (1LL << 11),
+    FfzEmoteImage = (1LL << 9),
+    FfzEmoteText = (1LL << 10),
     FfzEmote = FfzEmoteImage | FfzEmoteText,
     EmoteImages = TwitchEmoteImage | BttvEmoteImage | FfzEmoteImage,
     EmoteText = TwitchEmoteText | BttvEmoteText | FfzEmoteText,
 
-    BitsStatic = (1LL << 12),
-    BitsAnimated = (1LL << 13),
+    BitsStatic = (1LL << 11),
+    BitsAnimated = (1LL << 12),
 
     // Slot 1: Twitch
     // - Staff badge
     // - Admin badge
     // - Global Moderator badge
-    BadgeGlobalAuthority = (1LL << 14),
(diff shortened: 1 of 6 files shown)


src/messages/MessageElement.hpp
```diff
@@ -44,58 +44,65 @@ enum class MessageElementFlag : int64_t {
...
+             BadgeSubscription | BadgeVanity | BadgeChatterino | BadgeFfz,
 
-    ChannelName = (1LL << 19),
+    ChannelName = (1LL << 20),
 
-    BitsAmount = (1LL << 20),
+    BitsAmount = (1LL << 21),
 
-    ModeratorTools = (1LL << 21),
+    ModeratorTools = (1LL << 22),
```
Mm2PL, line 106: what's the reason behind reordering these? if you see this i von zulul

zneix: I thought I could re-order those in order to eliminate gaps
Pajlading

But if that's more of an issue than a step into right direction I might as well revert it.

pajlada (approved):

src/messages/MessageElement.hpp
pajlada: :+1: Agreed that this refactoring is fine

PR #2671, opened 2021-04-23
Added is:<flags> search predicate

### LINKED ISSUE #2653: Search Predicate for Disabled Messages
**What should be added?**
Add a search predicate for disabled messages (i.e., messages that were removed by a timeout). This predicate should simply check for the `MessageFlag::Disabled` flag.
Suggested syntax: `is:disabled`. Open to other suggestions though.

**Why should it be added?**
Allows you to quickly check automatically timed out messages, e.g. to see if they warrant a longer timeout or permanent ban or to copy that dank new ASCII copypasta.

This feature should be "almost free" in regards to additional code complexity and does not harm the current user experience.

talneoran:
Pull request checklist:

# Description

This PR adds a search predicate that allows filtering messages based on some of their flags.
Usage: `is:<flags>` where flags is a comma seperated list of one of the following currently supported options:

- `deleted`: Checks for `MessageFlag::Disabled`
- `sub` or `subscription`: Checks for `MessageFlag::Subscription`
- `timeout`: Checks for `MessageFlag::Timeout`
- `highlighted`: Checks for `MessageFlag::Highlighted`
- `system`: Checks for `MessageFlag::System`

Examples:
`is:deleted` would show deleted (disabled) messages
`is:sub,timeout` would show messages that are either a subscription message or a timeout message

Obviously if this is added I'm open to changes and additions to the current options and their naming (these are just the options that I initially thought of given the available flags), as well as changes to the phrasing of the changelog entry.

Closes #2653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added `is:<flags>` search filter to find messages of specific types. (#2653, #2671)
+
 ## 2.3.0
 
 - Major: Added custom FrankerFaceZ VIP Badges. (#2628)
--- a/src/messages/search/MessageFlagsPredicate.cpp
+++ b/src/messages/search/MessageFlagsPredicate.cpp
@@ -2,34 +2,31 @@
 
 namespace chatterino {
 
-MessageFlagsPredicate::MessageFlagsPredicate(const QStringList &flags)
+MessageFlagsPredicate::MessageFlagsPredicate(const QString &flags)
     : flags_()
 {
     // Check if any comma-seperated values were passed and transform those
-    for (const auto &entry : flags)
+    for (const auto &flag : flags.split(',', QString::SkipEmptyParts))
     {
-        for (const auto &flag : entry.split(',', QString::SkipEmptyParts))
+        if (flag == "deleted" || flag == "disabled")
         {
-            if (flag == "deleted" || flag == "disabled")
-            {
-                this->flags_.set(MessageFlag::Disabled);
-            }
-            else if (flag == "sub" || flag == "subscription")
-            {
-                this->flags_.set(MessageFlag::Subscription);
-            }
-            else if (flag == "timeout")
-            {
-                this->flags_.set(MessageFlag::Timeout);
-            }
-            else if (flag == "highlighted")
-            {
-                this->flags_.set(MessageFlag::Highlighted);
-            }
-            else if (flag == "system")
-            {
-                this->flags_.set(MessageFlag::System);
-            }
+            this->flags_.set(MessageFlag::Disabled);
+        }
+        else if (flag == "sub" || flag == "subscription")
+        {
+            this->flags_.set(MessageFlag::Subscription);
+        }
+        else if (flag == "timeout")
+        {
+            this->flags_.set(MessageFlag::Timeout);
+        }
+        else if (flag == "highlighted")
+        {
+            this->flags_.set(MessageFlag::Highlighted);
+        }
+        else if (flag == "system")
+        {
+            this->flags_.set(MessageFlag::System);
         }
     }
 }
--- a/src/messages/search/MessageFlagsPredicate.cpp
+++ b/src/messages/search/MessageFlagsPredicate.cpp
@@ -10,7 +10,7 @@ MessageFlagsPredicate::MessageFlagsPredicate(const QStringList &flags)
     {
         for (const auto &flag : entry.split(',', QString::SkipEmptyParts))
         {
-            if (flag == "deleted")
+            if (flag == "deleted" || flag == "disabled")
             {
                 this->flags_.set(MessageFlag::Disabled);
             }
--- a/src/messages/search/MessageFlagsPredicate.cpp
+++ b/src/messages/search/MessageFlagsPredicate.cpp
@@ -1,0 +1,42 @@
+#include "messages/search/MessageFlagsPredicate.hpp"
+
+namespace chatterino {
+
+MessageFlagsPredicate::MessageFlagsPredicate(const QStringList &flags)
+    : flags_()
+{
+    // Check if any comma-seperated values were passed and transform those
+    for (const auto &entry : flags)
+    {
+        for (const auto &flag : entry.split(',', QString::SkipEmptyParts))
+        {
+            if (flag == "deleted")
+            {
+                this->flags_.set(MessageFlag::Disabled);
+            }
+            else if (flag == "sub" || flag == "subscription")
+            {
+                this->flags_.set(MessageFlag::Subscription);
+            }
+            else if (flag == "timeout")
+            {
+                this->flags_.set(MessageFlag::Timeout);
+            }
+            else if (flag == "highlighted")
+            {
+                this->flags_.set(MessageFlag::Highlighted);
+            }
+            else if (flag == "system")
+            {
+                this->flags_.set(MessageFlag::System);
+            }
+        }
+    }
+}
+
+bool MessageFlagsPredicate::appliesTo(const Message &message)
+{
+    return message.flags.hasAny(flags_);
+}
+
+}  // namespace chatterino
--- a/src/messages/search/MessageFlagsPredicate.cpp
+++ b/src/messages/search/MessageFlagsPredicate.cpp
@@ -1,4 +1,3 @@
-
 #include "messages/search/MessageFlagsPredicate.hpp"
 
 namespace chatterino {
--- a/src/messages/search/MessageFlagsPredicate.cpp
+++ b/src/messages/search/MessageFlagsPredicate.cpp
@@ -1,3 +1,4 @@
+
 #include "messages/search/MessageFlagsPredicate.hpp"
 
 namespace chatterino {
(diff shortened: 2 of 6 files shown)


src/messages/search/MessageFlagsPredicate.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+MessageFlagsPredicate::MessageFlagsPredicate(const QStringList &flags)
+    : flags_()
+{
+    // Check if any comma-seperated values were passed and transform those
+    for (const auto &entry : flags)
+    {
+        for (const auto &flag : entry.split(',', QString::SkipEmptyParts))
+        {
+            if (flag == "deleted")
```
zneix, line 13: Just a nitpick, but you might add a check for `"disabled"` as well for all those nerds out there, who will use it instead of deleted.

src/messages/search/MessageFlagsPredicate.cpp
```diff
@@ -0,0 +1,42 @@
...
+            {
+                this->flags_.set(MessageFlag::Disabled);
+            }
+            else if (flag == "sub" || flag == "subscription")
```
zneix, line 17: Another nitpick: Maybe `"subscribed"` could be added / replace `"subscription"`.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Added `is:<flags>` search filter to find messages of specific types. (#2653, #2671)
```
zneix, line 5: I'll now try to make a wiki page about already existing search predicates. After that's done, you should probably make a PR to chatterino/wiki with a list of the flags that you can use with `"is"` and append link to the wiki page with those to the changelog entry.

src/messages/search/MessageFlagsPredicate.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * specified by user friendly names for the flags.
+ */
+class MessageFlagsPredicate : public MessagePredicate
+{
+public:
+    /**
+     * @brief Create a MessageFlagsPredicate with a list of flags to search for.
+     *
+     * @param flags a list of names for the flags a message should have
+     */
```
leon-richardt, line 23: I propose adding the user-friendly names allowed (`deleted`, `sub`, `highlighted` etc.) to the documentation here

src/messages/search/MessageFlagsPredicate.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace chatterino {
+
+using MessageFlags = FlagsEnum<MessageFlag>;
+
+/**
+ * @brief MessagePredicate checking for message flags.
+ *
+ * This predicate will only allow messages with a list of flags.
+ * specified by user friendly names for the flags.
```
leon-richardt, line 14:
```suggestion
 * Specified by user friendly names for the flags.
```

leon-richardt (changes requested):
Looks good so far, just asking for some documentation changes for now. :smile: Will compile and properly test later today

src/messages/search/MessageFlagsPredicate.cpp
leon-richardt: Not 100% sure about this one :thinking: My "fear" is that it could easily be confused with [Filters](https://wiki.chatterino.com/Filters/) only showing the _messages_ of subscribed users.

talneoran: Yeah I think "subscription" fits "is" more, and "is:subscribed" could be confused with the author being subscribed (which might be another good option but not in this context). Either way though if you think it should be an option I can add it.

leon-richardt (changes requested):
One thing I found while testing: if you provide multiple `is:<flags>` predicates, the flags will be OR'ed over all predicates provided. Ideally, only the flags separated by commas should be OR'ed and separate predicates should be AND'ed. I believe this could be fixed by creating a separate `QStringList flags` for every new `is:<flags>` keyword (or by clearing the list before parsing a new predicate).

For illustration, here is an example. In the last screenshot, I would expect to only see messages that are deleted *and* highlighted. Instead, messages that deleted *or* highlighted are shown:
<details>

<summary>Screenshots</summary>

Only deleted
Only highlighted
Deleted and highlighted

</details>

talneoran:
@leon-richardt This is actually something I noticed and is also true for the `from:users` and `in:channels` predicates in the current release (I implemented `in:channels` and this predicate out of `from:users` to be consistent so that's why). 

I could fix it for all 3 like you said by creating a new predicate and appending it to the predicate list every time the keyword is parsed (instead of just once for all occurences), and that would also make the outer loop in the constructors for all predicates useless. 
Question is since it applies to multiple predicates do I fix it in a seperate PR and only remove the loop in this one? Or maybe it can also be fixed per predicate by keeping multiple inner lists or something.

leon-richardt:
> @leon-richardt This is actually something I noticed and is also true for the `from:users` and `in:channels` predicates in the current release (I implemented `in:channels` and this predicate out of `from:users` to be consistent so that's why).

I'm not convinced these are 100% comparable. Any given message can only have exactly one author and only appear in exactly one channel while it can have multiple `MessageFlag`s. However, you do bring up a perfectly valid point about consistency.

----

> Question is since it applies to multiple predicates do I fix it in a seperate PR and only remove the loop in this one?

I think _existing_ behavior should be changed in a separate PR.

src/messages/search/MessageFlagsPredicate.cpp
zneix: > Not 100% sure about this one thinking My "fear" is that it could easily be confused with [Filters](https://wiki.chatterino.com/Filters/) only showing the _messages_ of subscribed users.

Yeah, my bad. I got confused first and thought that's what this flag does, but if it's just meant to show sub messages then it's fine as it is. Sorry for causing confusion ;p

talneoran:
> I'm not convinced these are 100% comparable. Any given message can only have exactly one author and only appear in exactly one channel while it can have multiple `MessageFlag`s.

Right I see what you mean now, maybe it is better then to change it specifically for this predicate (and potentially others in the future), since there's no meaning to an AND relation for the other existing ones, and no reason to add seperate list entries for those.

CHANGELOG.md
zneix: Chatterino/wiki#65 has now been merged in, you can try appending new stuff to the list of Search filters and maybe add some extra examples at the end in the same format I made those. As for version note `(added in 2.X.X)` just say `(added in nightly build)` or just nightly.

zneix:
Just noticed that all the sub messages also get `MessageFlag::Highlighted` flag, thus they also show up when you search for highlighted messages:
pajaGhetti

I believe that should/could be fixed before merging, but maybe in another PR.

src/messages/search/MessageFlagsPredicate.cpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        if (flag == "deleted" || flag == "disabled")
+        {
+            this->flags_.set(MessageFlag::Disabled);
+        }
+        else if (flag == "sub" || flag == "subscription")
+        {
+            this->flags_.set(MessageFlag::Subscription);
+        }
+        else if (flag == "timeout")
```
zneix, line 19: Could also include ban there, since it also works for ban messages

```suggestion
        else if (flag == "timeout" || flag == "ban")
```

zneix (changes requested):

talneoran:
>     1. Just noticed that all the sub messages also get `MessageFlag::Highlighted` flag, thus they also show up when you search for highlighted messages:
>
Isn't this because of this setting?
image

Assuming it is, why is it wrong that `is:highlighted` captures highlighted sub messages when that highlight is enabled?

>     2. Also I'm not sure if we want to change timeout/ban messages having `MessageFlag::System`, so your search could be a bit more specific.
Not sure what you are suggesting, the filter still just finds messages with `MessageFlag::System` on. Maybe the meaning of the flag can be changed or maybe this flag shouldn't be supported in this filter, but otherwise do you think there's something worth changing with the implementation in this PR?

zneix:
> Isn't this because of this setting?
> [image](https://user-images.githubusercontent.com/43518924/115997121-9a06cf80-a5ea-11eb-8373-62d484f985f4.png)
> 
> Assuming it is, why is it wrong that `is:highlighted` captures highlighted sub messages when that highlight is enabled?

You're right, I had it enabled - totally forgot that existed. Since it's a setting I guess it can stay this way. You don't have to change anything then I think - I just found that just a little confusing that sub messages come up when I search for highlights.


> Not sure what you are suggesting, the filter still just finds messages with `MessageFlag::System` on.

Again, you're right :+1:

> Maybe the meaning of the flag can be changed or maybe this flag shouldn't be supported in this filter.

Yes, I was thinking that messages having `MessageFlag::Timeout` could be filtered / removed when searching with predicate `is:system`, so those won't show up.

> but otherwise do you think there's something worth changing with the implementation in this PR?

I don't think you should change how / which flags are assigned to messages since that's just a little out of scope of this Pull Request. If anything, try to play around with the point I specified above.

zneix (approved):

pajlada (approved):
Tested and works as expected, looks good to me.
Thanks for your contribution!

Issue #2676, opened 2021-04-23
Failed to automatically fetch BTTV emotes upon opening client (2.3.0) 


Vafarien:
Upon first opening the new version of chatterino 2.3.0 windows client a long wait time for all channels to load and initialize, i navigate to a channel only to find that BTTV emotes failed to load in the channel, only by doing a manual reload (F5) the channels emotes load in propperly, the issue is that to get BTTV wokring on every channel is that you have to manually go to each channel and reload everything.

Steps to reproduce:
1. install 2.3.0 chatterino on windows 64-bit system
2.launch chatterino
3.be in channel and see that BTTV emotes failed to load
4. press F5 to manually reload everytime you open chatterino

Version: Chatterino 2.3.0 (commit 58017a754)

Operating System: Windows 7 64-bit

![image](https://user-images.githubusercontent.com/2349950/115932433-cdf2c100-a441-11eb-8543-b54f13dfbca4.png)
![image](https://user-images.githubusercontent.com/2349950/115932572-0abeb800-a442-11eb-9473-4cc6ad76ad6d.png)

jammehcow:
Can reproduce on Chatterino 2.3.0 (commit `0a51252b7`). Seems to happen at random but definitely exists.

jammehcow:
Appeared to be a genuine timeout from BTTV - cannot reproduce now.

Maybe it's worth putting some focus on the BTTV emote retry mentioned in `BttvEmotes.cpp`.

Vafarien:
No longer able to reproduce issue.

jammehcow:
Could probably be reopened. Seems to be happening across a large set of users.

zneix:
I believe this is a thing just because requests to BTTV API tend to time out sometimes (usually, when you have lots of tabs open and there are concurrent API requests being made at the same time. Worth noting that this was also the case before 2.3.0 was released.

jammehcow:
Should be resolved on the timeout side by #2900 (or I might abandon because of scope and create a new PR) when merged.

PR #2677, opened 2021-04-24
Retry BTTV emote loading on timeout

jammehcow:
Pull request checklist:

# Description

When loading BTTV emotes fails due to BTTV timing out, the loading is retried a maximum of 5 (const) times before giving up.
This aims to fix BTTV timesouts (as reported in #2676).

### DIFF
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -207,7 +207,7 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
             }
             return pair.first;
         })
-        .onError([channelId, channel, manualRefresh](auto result) {
+        .onError([channelId, channel, channelDisplayName, callback, manualRefresh, retryCount](auto result) {
             auto shared = channel.lock();
             if (!shared)
                 return;
@@ -229,6 +229,8 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
                         QString("Failed to fetch BetterTTV channel emotes. (timed out, retrying %1/%2)")
                         .arg(QString::number(retryCount + 1), QString::number(MAX_BTTV_RETRY_COUNT));
                 shared->addMessage(makeSystemMessage(sysMessage));
+
+                loadChannel(channel, channelId, channelDisplayName, callback, manualRefresh, retryCount + 1);
             }
             else
             {
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -158,7 +158,8 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
                              const QString &channelId,
                              const QString &channelDisplayName,
                              std::function<void(EmoteMap &&)> callback,
-                             bool manualRefresh)
+                             bool manualRefresh,
+                             int retryCount)
 {
     NetworkRequest(QString(bttvChannelEmoteApiUrl) + channelId)
         .timeout(3000)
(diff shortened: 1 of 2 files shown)


jammehcow:
Is 5 a reasonable number of times to retry?

I also haven't been able to test this because I'm lazy so will look at tomorrow unless someone else confirms.

leon-richardt:
Just wondering what you think about doing an [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) (e.g. as with TCP retries) in regards to the request rate. If the BTTV server fails to reply due to currently being overloaded, repeatedly spamming requests will likely cause further detriments.
So, my suggestion would be waiting an increasing period of time before retrying. A possible progression in terms of cooldown might be `1s -> 2s -> 4s -> 8s -> 16s` but I'm open to any variations of this.

jammehcow:
@leon-richardt I knew it didn't feel right. I'll add that in.

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -179,8 +179,10 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
         shared->addMessage(makeSystemMessage(sysMessage));
     }
 
+    const int timeout = 2 ^ retryCount * 1000;
```
leon-richardt, line 182: Isn't this doing XOR on `2` and `retryCount * 1000`?

jammehcow: Dumb mistake from a non-cpp dev. Resolved

jammehcow:
segfault when a timeout eventually succeeds in `parseEmotes:105`. I assume it's to do with how `channelDisplayName` is passed around. Not a knowledgeable C++ dev to understand how that works with anonymous methods so pointers would be much appreciated.

pajlada:
I think it would be best if this sort of functionality was implemented as part of the `NetworkRequest` instead, where you could specify `.retry(5)` or something as a parameter.

The logic with the exponential backoff is :ok_hand:

jammehcow:
It appears that FFZ and Twitch calls via `NetworkManager` timeout after 30s whereas BTTV is only 3 seconds.

I'm wondering if instead of implementing a retry thing (which would probably be useful in the long run anyways) we just bump up the value to 30s like the rest. Thoughts @pajlada ?

zneix:
> It appears that FFZ and Twitch calls via `NetworkManager` timeout after 30s whereas BTTV is only 3 seconds.
> 
> I'm wondering if instead of implementing a retry thing (which would probably be useful in the long run anyways) we just bump up the value to 30s like the rest. Thoughts @pajlada ?

I agree with you 100% - in most cases there're even more channel bttv emotes than global bttv emotes, so it makes no sense to have less timeout time for a bigger request. This also seems to be an (ancient) typo (3000 insetad of 30000), introduced here - https://github.com/Chatterino/chatterino2/commit/c768bd9bd9d696c86a7525eabefffebf046d66c3#diff-b626f88008195f94729ebb0738b8c7c883fb01f0d4aceb7bc55f137ee6a24a38R131 Since you're asleep, I will make a pull request for it right now.

jammehcow:
> I agree with you 100% - in most cases there're even more channel bttv emotes than global bttv emotes, so it makes no sense to have less timeout time for a bigger request. This also seems to be an (ancient) typo (3000 insetad of 30000), introduced here - [c768bd9#diff-b626f88008195f94729ebb0738b8c7c883fb01f0d4aceb7bc55f137ee6a24a38R131](https://github.com/Chatterino/chatterino2/commit/c768bd9bd9d696c86a7525eabefffebf046d66c3#diff-b626f88008195f94729ebb0738b8c7c883fb01f0d4aceb7bc55f137ee6a24a38R131) Since you're asleep, I will make a pull request for it right now.

Thanks @zneix . Will close this PR unless anyone desperately needs retry functionality in NetworkManager.

PR #2694, opened 2021-04-26
removed access to theme from TMB + misc fixes

fourtf:
Pull request checklist:


pajlada:
When colorizing usernames, the subsequent text now becomes the same color

How it should be:
2021-04-26-215851_599x53_scrot

What it looks like with this patch:
2021-04-26-215846_401x37_scrot


This line: https://github.com/Chatterino/chatterino2/pull/2694/files#diff-a61469571f1ced2559899f109781be679b0a705e22a97ca651a8b40973d4e726R469 seems to be the problematic one. Removing the `&&` portion fixes it, possibly due to us modifying textColor in the upcoming lines

fourtf:
> 
> 
> When colorizing usernames, the subsequent text now becomes the same color
> 
> How it should be:
> 2021-04-26-215851_599x53_scrot
> 
> What it looks like with this patch:
> 2021-04-26-215846_401x37_scrot
> 
> This line: https://github.com/Chatterino/chatterino2/pull/2694/files#diff-a61469571f1ced2559899f109781be679b0a705e22a97ca651a8b40973d4e726R469 seems to be the problematic one. Removing the `&&` portion fixes it, possibly due to us modifying textColor in the upcoming lines

removed the universal reference

pajlada (approved):
LGTM :+1:

PR #2695, opened 2021-04-26
adds assertInGuiThread to getApp()

fourtf:
Pull request checklist:

# Description
Worth considering adding this since accessing getApp() from another thread as the gui thread can lead to parallelism issues. Didn't assert in my quick test.


pajlada:
Testing this PR now

pajlada (approved):
I've run this for a full day now and have not run into the assertation

PR #2702, opened 2021-04-27
Make AccessGuard use a shared_mutex instead

pajlada:
This means accessConst from UniqueAccess will return a 'read' lock, so
multiple places can read from the value at the same time.

As a part of this rewrite, I went ahead and swapped the manual .lock/.unlock
with `unique_lock` and `shared_lock` members as those already nicely
wrap the lock/unlock and lock_shared/unlock_shared functions.

Pull request checklist:

# Description


pajlada:
This PR is ready for review @fourtf

src/common/UniqueAccess.hpp
```diff
@@ -48,10 +42,13 @@ class AccessGuard
...
 private:
     T *element_{};
-    std::mutex *mutex_{};
-    bool isValid_{true};
+    LockType lock_;
 };
 
+template <typename T>
+using SharedAccessGuard =
+    AccessGuard<const T, std::shared_lock<std::shared_mutex>>;
```
fourtf, line 50: I would remove the const here

fourtf:
I added a small comment, otherwise lgtm 👍

src/common/UniqueAccess.hpp
pajlada: Going to keep the `const` here as per our discussion in chat since without it it's easy to make a difficult to read compile error when working with `SharedAccessGuard`

pajlada (commented):
Merging this later today

image

PR #2713, opened 2021-04-28
Refactored Rename Tab dialog and its context menu

### LINKED ISSUE #2709: Show context in rename dialog
**What should be added?**
When renaming a tab, a simple dialog appears with a single text box. This should explain to the user that they are renaming the tab as well as showing the name of the tab in the title/text box placeholder.

**Why should it be added?**
It is not obvious what the dialog does if a user opens it by accident. You could also misclick and believe that you're renaming a different tab than the one being acted on.

zneix:
Pull request checklist:

# Description

I removed `TextInputDialog` class since it was only useful in `NotebookTab.cpp` and adding to that class was more expensive than constructing `QDialog` object inside `NotebookTab.cpp`.

- Added placeholder to lineEdit which shows default tab name.
- Added some spacing, a label and renamed window title to make things look cleaner.
- Edited Tab context menu to make it easier to understand.

layouts are dank my transparency is even danker

Closes #2709


jammehcow:
Tested, LGTM 👍

jammehcow (approved):

fourtf:
Even though this is the suggested solution from issue thread I'm not sure if this solves the issue accurately. A placeholder should help the user understand what he has to input and I don't think that showing up he channel name does that.

fourtf:
I played around a bit in figma and came up with this:
image

It incorporates some other elements to make it more clear what the dialog is for and what the user has to input. I looked at the right click menu of the tab and tried to improve on it a bit to also make it more clear there.

#2709

zneix:
What I managed to achieve:

layouts are dank my transparency is even danker

src/widgets/helper/NotebookTab.cpp
```diff
@@ -75,15 +74,53 @@ NotebookTab::NotebookTab(Notebook *notebook)
...
+    lineEdit->setText(this->getCustomTitle());
+    lineEdit->setPlaceholderText(this->getDefaultTitle());
+    lineEdit->selectAll();
+
+    vbox->addWidget(new QLabel("Name:"));
+    vbox->addWidget(lineEdit);
+    vbox->addStretch(1);
+
+    auto okButton = new QPushButton("OK");
+    auto cancelButton = new QPushButton("Canel");
```
Mm2PL, line 91: canel forsenT

Mm2PL (changes requested):

src/widgets/helper/NotebookTab.cpp
```diff
@@ -75,15 +74,53 @@ NotebookTab::NotebookTab(Notebook *notebook)
...
+    lineEdit->selectAll();
+
+    vbox->addWidget(new QLabel("Name:"));
+    vbox->addWidget(lineEdit);
+    vbox->addStretch(1);
+
+    auto okButton = new QPushButton("OK");
+    auto cancelButton = new QPushButton("Cancel");
+
+    auto buttonBox = new QHBoxLayout;
```
pajlada, line 93: Use a https://doc.qt.io/qt-5/qdialogbuttonbox.html instead

pajlada (approved):
LGTM :+1: Merge in after adding the changelog entry as agreed

PR #2716, opened 2021-04-29
Hide similar messages by any user

33kk:
Pull request checklist:

# Description

Adds a setting to hide similar messages by any user


src/singletons/Settings.hpp
```diff
@@ -367,6 +367,8 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting colorSimilarDisabled = {"/similarity/colorSimilarDisabled",
                                         true};
     BoolSetting hideSimilar = {"/similarity/hideSimilar", false};
+    BoolSetting hideSimilarBySameUser = {"/similarity/hideSimilarBySameUser",
+                                         true};
```
Felanbird, line 371:
```suggestion
                                         false};
```
While I have no opinion positively or negative on this setting, I 100% believe it should be disabled by default. As most of the other "similarity" settings are disabled by default, and on top of that, having peoples spam suddenly gone would be extremely confusing.

Felanbird (changes requested):

src/singletons/Settings.hpp
33kk: This does not enable R9K by default, I set default to true because it worked like that before. You still have to enable "Hide similar messages"

Felanbird: My mistake then, apologies.

pajlada (approved):

pajlada (approved):

PR #2755, opened 2021-05-08
Fix newlines in stream title causing text in header to go out of bounds

jupjohn:
Pull request checklist:

# Description

When a stream title has newlines at the end of the text it causes the text in the split header to move upwards. This change just removes the newlines from the end of the string (including whitespace) to prevent this from happening.

### Before:

Chatterino2_Jammehcow_NewlineTitlePre

### After:

Chatterino2_Jammehcow_NewlineTitlePost

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Unversioned
 
+- Minor: Hosting messages are now clickable. (#2655)
+- Minor: Messages held by automod are now shown to the user. (#2626)
 - Bugfix: Strip newlines from stream titles to prevent text going off of split header (#2755)
 - Bugfix: Automod messages now work properly again. (#2682)
 - Bugfix: `Login expired` message no longer highlights all tabs. (#2735)
+- Bugfix: Fix a deadlock that would occur during user badge loading. (#1704, #2756)
 
 ## 2.3.1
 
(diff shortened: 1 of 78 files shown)


jammehcow:
could almost just replace any newline. would like to test with newlines in the middle of title

apa420:
You can have newlines anywhere in the title so we need to regex all newlines and replace them as spaces like twitch does. Then probably have to run the qtstring function which removes double spaces or w/e.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -170,6 +170,9 @@ namespace {
         if (settings.headerStreamTitle && !s.title.isEmpty())
             title += " - " + s.title;
 
+        // Strip out phantom newlines/trailing whitespace to prevent formatting issues
+        title.replace(QRegExp("[\n\\s]+$"), "");
+
```
pajlada, line 175: The 'phantom newlines' can be added through the Twitch website when changing your title by simply pressing enter in the title box.

On the Twitch website, all newlines in the title are stripped.

There's a handy dandy QString function in Qt5 for solving this called `simplified` (see https://doc.qt.io/qt-5/qstring.html#simplified)

```suggestion
        if (settings.headerStreamTitle && !s.title.isEmpty())
        {
            title += " - " + s.title.simplified();
        }

```

pajlada (changes requested):
Small change to support titles like:
```
foo
bar
baz
```

which on Twitch shows up as `foo bar baz`

pajlada (approved):

PR #2759, opened 2021-05-08
Add setting to disable custom ffz mod/VIP badges

### LINKED ISSUE #2693: Add a setting to disable FFZ custom VIP/mod badges
**What should be added?**
A toggle setting that lets you disable the custom FFZ badges and see the regular mod/vip badges instead.
FFZ allows disabling both badges seperately, but there could also be a single setting for both.
This setting should probably be enabled by default. 
Optional phrasing: "Use custom FrankerFaceZ moderator and VIP badges" for one setting, or "Use custom FrankerFaceZ VIP badges" seperately (similar phrasing to the FFZ settings).

**Why should it be added?**
Allows choice for users to opt out of an optional FFZ feature. Specifically for VIP badges, since they are completely custom images, it may also prevent confusion with other badges (was mentioned in the PR [here](https://github.com/Chatterino/chatterino2/pull/2628#issuecomment-821570815)).

talneoran:
Pull request checklist:

# Description
Adds a toggle setting that allows disabling using the custom FFZ moderator and VIP badges.
Currently implemented as one setting for both, to avoid more settings, and is enabled by default.

Currently not sure about:
- The location of the setting in the menu and in the settings file.
- Other badge settings update old messages when changed, for example disabling the Chatterino badge setting will remove that badge from already existing messages in chat. However, due to how the current implementation works, this setting doesn't update on old messages, and will only change whether or not the custom badges are used for new messages. Not sure if this is a problem, and if it's worth the effort to make this work when, to my knowledge, there isn't a simple solution that doesn't change the original implementation.

Closes #2693


apa420:
I'd rather have a dropdown menu where you enable the custom badges.
Something like:
Enable custom FFZ badges for:
- VIP and Moderator
- VIP
- Moderator
- None
Where VIP and Moderator are default.

(Before you implement it it'd be better if someone else has an opinion on this)

apa420:
That's a separate issue.

pajlada:
In the long run a dropdown with a string value won't make sense so it's more worth keeping the settings as 'single values', even though it will add a bit more clutter now it can be cleaned up the future if another value were to be added. tl;dr: 2 checkboxes > 1 dropdown menu imo. Setting scope is fine, setting position we've chatted about but it can be added after a separator in the "Visible badges" category

pajlada (approved):
Looks good to me 👌

Issue #2767, opened 2021-05-11
Block List capped at 100 users


LeetHaxor:
Added a bunch of spammers to my block list, but when I go to the block list there's only 20 users listed. If I add more than 20 it shows blank lines as if they were added (but no text on the line), but when I reload chatterino it's capped at 20. They are still ignored on normal twitch chat, but I can see their messages in chatterino (and yes I am using "Enable twitch blocked users"). Is there some limitation in place that prevents more than 20 blocks?

LeetHaxor:
Chatterino Nightly 2.3.1 (commit 2ea233a28, )

alazymeme:
Yes, we don't move past the first page for blocked users.

LeetHaxor:
> 
> 
> Yes, we don't move past the first "page" for blocked users. No idea why

That's unfortunate. I suppose there's not much I can do then. Thanks.

LeetHaxor:
> 
> 
> While the question you asked got an answer, I'd still consider this an issue and something that could be fixed.
Oh okay, reopening for now then!

LeetHaxor:
> 
> 
> Needs to be capped at some level so it doesn't spam too many requests if the user has millions of blocks (maybe ~10000 blocks?).

Makes sense. Currently at 250+, so hopefully something above that is easily achievable. Thanks.

zneix:
A little update on this: on latest nightly block list returns 100 blocks, we'll soon make a fix to respect pagination returned from Twitch's API and try to return all users in the block list.

Foretack:
> Needs to be capped at some level so it doesn't spam too many requests if the user has millions of blocks (maybe ~10000 blocks?).

May I propose adding a setting for users to choose how many blocked users they want to load? Sounds like a good compromise

pajlada:
> > Needs to be capped at some level so it doesn't spam too many requests if the user has millions of blocks (maybe ~10000 blocks?).
> 
> May I propose adding a setting for users to choose how many blocked users they want to load? Sounds like a good compromise

I would rather just increase the amount and cap it somewhere higher. Maybe 1000 (so 10 requests)?

PR #2772, opened 2021-05-12
Load 100 blocked users instead of default 20

iProdigy:
Pull request checklist:

# Description

Work towards #2767 so that at least up to 100 blocked users are returned rather than the default 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Minor: Added a message that displays a new date on new day. (#1016)
 - Minor: Hosting messages are now clickable. (#2655)
 - Minor: Messages held by automod are now shown to the user. (#2626)
+- Minor: Load 100 blocked users rather than the default 20. (#2772)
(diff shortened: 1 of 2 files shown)


fourtf:
For some wild reason the "count" is actually called "first". Haven't tested but should work.

https://dev.twitch.tv/docs/api/reference#get-user-block-list

zneix:
I think that following pagination to load all the blocks is a better solution, however this is better than nothing.

jammehcow:
Might be nicer to paginate the blocked users page so that we can return a decent chunk. Might be too bloated for current settings layout though

pajlada:
> I think that following pagination to load all the blocks is a better solution, however this is better than nothing.

Pagination would definitely be what we want to do, but even with a pagination PR we'd want to increase the limit to 100 so this is a good first step

pajlada (approved):

zneix (approved):
Fair point, I agree with paj.

PR #2779, opened 2021-05-13
Refactored and Migrated to Helix AutoMod message management

### LINKED ISSUE #2778: Migrate AutoMod message allow/deny to Helix
AutoMod message approve/deny have been added to [Helix](https://dev.twitch.tv/docs/api/reference#manage-held-automod-messages). As of tommorow Kraken endpoints will be decomissioned, so we have to migrate ASAP!

zneix:
Pull request checklist:

# Description

Migrated AutoMod approve/deny endpoints to corresponding [Helix endpoint](https://dev.twitch.tv/docs/api/reference#manage-held-automod-messages). Since Helix gives us more descriptive error codes, I've added better handling for error cases (we print and extra system message on every error with customized messages based on error type).

Closes #2778

### DIFF
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -361,7 +361,7 @@ SharedAccessGuard<const TwitchAccount::TwitchAccountEmoteData>
 void TwitchAccount::autoModAllow(const QString msgId, ChannelPtr channel)
 {
     getHelix()->manageAutoModMessages(
-        msgId, QString("ALLOW"),
+        this->getUserId(), msgId, QString("ALLOW"),
         [] {
             // success
         },
@@ -403,7 +403,7 @@ void TwitchAccount::autoModAllow(const QString msgId, ChannelPtr channel)
 void TwitchAccount::autoModDeny(const QString msgId, ChannelPtr channel)
 {
     getHelix()->manageAutoModMessages(
-        msgId, QString("DENY"),
+        this->getUserId(), msgId, QString("DENY"),
         [] {
             // success
         },
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -3,12 +3,16 @@
 #include <QThread>
 
 #include "Application.hpp"
+#include "common/Channel.hpp"
 #include "common/Env.hpp"
 #include "common/NetworkRequest.hpp"
 #include "common/Outcome.hpp"
 #include "common/QLogging.hpp"
 #include "controllers/accounts/AccountController.hpp"
+#include "messages/Message.hpp"
+#include "messages/MessageBuilder.hpp"
 #include "providers/IvrApi.hpp"
+#include "providers/irc/IrcMessageBuilder.hpp"
 #include "providers/twitch/TwitchCommon.hpp"
 #include "providers/twitch/TwitchUser.hpp"
 #include "providers/twitch/api/Helix.hpp"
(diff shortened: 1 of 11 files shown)


apa420 (approved):

src/providers/twitch/api/Helix.hpp
```diff
@@ -307,6 +314,11 @@ class Helix final : boost::noncopyable
                        std::function<void(NetworkResult)> successCallback,
                        HelixFailureCallback failureCallback);
 
+    // https://dev.twitch.tv/docs/api/reference#manage-held-automod-messages
+    void manageAutoModMessages(
+        QString msgId, QString action, std::function<void()> successCallback,
+        std::function<void(HelixAutoModMessageError)> failureCallback);
```
pajlada, line 320: Please use `msgID` here instead of `msgId`

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -354,42 +358,88 @@ SharedAccessGuard<const TwitchAccount::TwitchAccountEmoteData>
...
-        .execute();
+                case HelixAutoModMessageError::MessageNotFound: {
+                    errorMessage += "target message not found.";
+                }
+                break;
+
+                // This would most likely happen if the service is down, or if the JSON payload returned has changed format
+                case HelixAutoModMessageError::Unknown:
+                default: {
+                    errorMessage += "an unknown error occured. You may need to "
```
jammehcow, line 393: Error message doesn't match the description in the comment above. Maybe a better message is `"an unknown error occurred. The response from Twitch was not as expected"`

The user re-authing would fix this the majority of the time (unless I'm missing something because of that comment)

pajlada (changes requested):
I only have pedantic changes to request - functionally it works as well as it can - I will note some of my findings as part of a separate comment

pajlada:
Things that aren't blockers but are good to know for the future:
The Helix API endpoint will send the 'Message not found' error if the user presses **Allow** or **Deny** too quickly - we may want to file a bug report for this or implement some 'retry' system for these requests where if an allow or deny request fails with that code we try again after 2-3 seconds.

Once a phrase has been allowed, the message does not show up again as it did with the Kraken endpoint - this seems to be handled by the `automod-queue.USER_ID.CHANNEL_ID` PubSub topic now and in Twitch web chat the original message is just 'unmarked' as being bad.

`{"type":"MESSAGE","data":{"topic":"automod-queue.11148817.117166826","message":"{\"type\":\"automod_caught_message\",\"data\":{\"content_classification\":{\"category\":\"swearing\",\"level\":2},\"message\":{\"content\":{\"text\":\"fuck anus sex!!!\",\"fragments\":[{\"text\":\"fuck anus sex!!!\",\"automod\":{\"topics\":{\"dating_and_sexting\":7,\"vulgar\":6}}}]},\"id\":\"129e6be0-6eaf-42b6-9a51-959be740d317\",\"sender\":{\"user_id\":\"82008718\",\"login\":\"pajbot\",\"display_name\":\"pajbot\",\"chat_color\":\"#2E8B57\"},\"sent_at\":\"2021-05-13T23:47:12.99980313Z\"},\"reason_code\":\"\",\"resolver_id\":\"11148817\",\"resolver_login\":\"pajlada\",\"status\":\"ALLOWED\"}}"}}`

pajlada:
oh and because it wasn't mentioned in the previous comments - thanks for being quick on this fix! :heart:

src/providers/twitch/TwitchAccount.cpp
zneix: We just return a generic error response, same as in /marker command in CommandController.cpp the comment is a copy-paste from there. No idea if we have to change this in any way or not.

I also added proper handling for 401 now, so it's just an unknown error.

pajlada (approved):
LGTM :+1: Feel free to merge in whenever

PR #2781, opened 2021-05-14
Fix word backspacing not closing/updating colon emote picker

### LINKED ISSUE #2780: Emote picker window stays open when deleting words with Ctrl + Backspace
**Describe your issue**
When using the `:` emote picker and using Ctrl + Backspace to delete the text + colon the emote picker window stays active.
Pressing escape does close the window.

**Screenshots**

https://user-images.githubusercontent.com/4962764/118222668-2282dc80-b4d4-11eb-8306-4247f344ac99.mp4

**OS and Chatterino Version**
Arch Linux (btw), Linux Zen `5.12.3-zen`
Chatterino 2.3.1 (commit 2ea233a2)

jammehcow:
Pull request checklist:

Fixes #2780

At the moment the handler can be called twice when the user types a character because both the text input updates and cursor changes position. If anyone has a good way to block one of the events when the other is triggered please suggest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Bugfix: `Login expired` message no longer highlights all tabs. (#2735)
 - Bugfix: Fix a deadlock that would occur during user badge loading. (#1704, #2756)
 - Bugfix: Tabbing in `Select a channel to open` is now consistent. (#1797)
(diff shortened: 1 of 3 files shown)


pajlada (approved):
I don't think trying to avoid it being called twice is worth the overhead/implementation complexity it would require, so this fix looks good to me :+1:

zneix (approved):

PR #2783, opened 2021-05-15
Deprecate C_DEBUG definition in favour of standard NDEBUG

pajlada:
Pull request checklist:

# Description

The checks for C_DEBUG will need to be reversed now as C_DEBUG defined that debug was enabled, and NDEBUG defines that debug is disabled (NDEBUG=NoDebug)

This PR will fix it so that all the F6 and beyond debug bindings work with CMake properly, sorry @mm2pl if this causes some merge conflicts

- Use NDEBUG instead of C_DEBUG in src/RunGui in the "restart on crash" code
- Use NDEBUG instead of C_DEBUG in the "add debug bindings" code

### DIFF
--- a/src/RunGui.cpp
+++ b/src/RunGui.cpp
@@ -134,7 +134,7 @@ namespace {
     // true.
     void initSignalHandler()
     {
-#ifndef C_DEBUG
+#ifdef NDEBUG
         signalsInitTime = std::chrono::steady_clock::now();
 
         signal(SIGSEGV, handleSignal);
--- a/src/widgets/Window.cpp
+++ b/src/widgets/Window.cpp
@@ -25,7 +25,7 @@
 #include "widgets/splits/Split.hpp"
 #include "widgets/splits/SplitContainer.hpp"
 
-#ifdef C_DEBUG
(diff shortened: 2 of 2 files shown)


jammehcow (approved):

zneix (approved):
~~Looks good, however we could mention it somewhere that debug stuff won't work on qmake anymore. Or is this slowly a process of deprecating old build system?~~

See my comment below

pajlada:
> Looks good, however we could mention it somewhere that debug stuff won't work on qmake anymore. Or is this slowly a process of deprecating old build system?

Did you test this? When I tested qmake, this still worked as expected (i.e. Release builds define NDEBUG, debug builds do not)

zneix:
> > Looks good, however we could mention it somewhere that debug stuff won't work on qmake anymore. Or is this slowly a process of deprecating old build system?
> 
> Did you test this? When I tested qmake, this still worked as expected (i.e. Release builds define NDEBUG, debug builds do not)

Nevermind, I must've built release by accident. Tested now and debug stuff on function keys works on both QMake and CMake.

PR #2800, opened 2021-05-20
Rewrite logic for when to show moderation buttons next to messages

fourtf:
Pull request checklist:

# Description

removes some weird 20 lines of code that are actually two lines of normal code

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -184,8 +184,7 @@ MessagePtr TwitchMessageBuilder::build()
     this->emplace<TimestampElement>(
         calculateMessageTimestamp(this->ircMessage));
 
-    if (!this->senderIsBroadcaster &&
-        !(this->tags.value("user-type") == "mod" && !args.isStaffOrBroadcaster))
+    if (this->shouldAddModerationElements())
     {
         this->emplace<TwitchModerationElement>();
     }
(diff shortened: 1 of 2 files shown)


pajlada:
I'm still having hard times wrapping my head around the if. I'm not sure if it's double negation + && or if it's the nested conditional - my counter-proposal would be to wrap the logic in a private function, something akin to this:
```cpp
bool TwitchMessageBuilder::shouldAddModerationElements() const {
    if (this->senderIsBroadcaster) {
        return false;
    }

    if (this->tags.value("user-type").toString() == "mod" && this->args.isStaffOrBroadcaster) {
        return false;
    }

    return true;
}
```

We wouldn't shave off any lines, but the logic for moderation buttons is clear, concise enough (7 lines of code vs the previous 20-line mess)

This is not a hill I'm willing to die on, so if that's unacceptable I'd be fine with a comment above the code instead

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -231,29 +231,8 @@ MessagePtr TwitchMessageBuilder::build()
...
-                {
-                    addModerationElement = false;
-                }
-            }
-        }
-    }
-
-    if (addModerationElement)
+    if (!this->senderIsBroadcaster &&
+        !(this->tags.value("user-type") == "mod" && !args.isStaffOrBroadcaster))
```
apa420, line 235:
```suggestion
    if (!(this->senderIsBroadcaster ||
        (this->tags.value("user-type") == "mod" && !args.isStaffOrBroadcaster)))
```

Would this fix the confusion @pajlada ?

pajlada: No

pajlada (changes requested):
I'm still having hard times wrapping my head around the if. I'm not sure if it's double negation + && or if it's the nested conditional - my counter-proposal would be to wrap the logic in a private function, something akin to this:
```cpp
bool TwitchMessageBuilder::shouldAddModerationElements() const {
    if (this->senderIsBroadcaster) {
        return false;
    }

    if (this->tags.value("user-type").toString() == "mod" && this->args.isStaffOrBroadcaster) {
        return false;
    }

    return true;
}
```

We wouldn't shave off any lines, but the logic for moderation buttons is clear, concise enough (7 lines of code vs the previous 20-line mess)

This is not a hill I'm willing to die on, so if that's unacceptable I'd be fine with a comment above the code instead

pajlada (approved):

PR #2803, opened 2021-05-21
Add moderation buttons in search popup

### LINKED ISSUE #2148: Add moderation buttons for search (ctrl+f)
If Moderation buttons is enabled you would be able to see and use them if you use Search (ctrl+f) in that chat.

It would help if you wanted to search for a certain phrase to timeout/ban instead of scrolling up.

talneoran:
Pull request checklist:

# Description

This PR adds moderation buttons to messages in a search popup opened from a split where moderation mode is active.

While this does seem to work for the cases I tested, I'm not really sure about the implementation and would like some input:

With the way the existing imlementation of adding the `MessageElementFlag` works, the `ChannelView` object "queries" the parent widget (by `dynamic_cast`), expecting it to be a split and checking for moderation mode in that split. The addition here is checking if the parent widget is actually a search popup, fetching the parent of the search widget which should be a split, and using that split to determine moderation mode. I'm just not sure if this is a good way of going about it. (A similar thing applies to determining the channel for performing the moderation actions when clicking the buttons)

To test, simply add whichever moderation buttons you would like to test, open a split with moderation mode enabled, open a search popup, and try the buttons.

Closes #2148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added moderation buttons to search popup when searching in a split with moderation mode enabled. (#2148, #2803)
 - Minor: Added settings to disable custom FrankerFaceZ VIP/mod badges. (#2693, #2759)
 
 ## 2.3.2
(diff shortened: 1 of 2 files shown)


pajlada (approved):
LGTM :+1: Good quality of life change for moderators!

PR #2805, opened 2021-05-22
Attached Video Viewer

goldbattle:
Pull request checklist:

# Description

I have been experimenting with including an internal / attached video player which would be docked alongside the chat. This would allow for a pretty seamless viewing and chatting experience if possible. I have gone and done a quick demo of what this could look like and it seems to only be a few modifications so far.

Originally I stumbled upon the [StreamLinkerino](https://github.com/JohnCiubuc/StreamLinkerino) project, but I want to make it a more first class feature. Here is an example screenshot of what I have so far:

image
image

What works:
- Clicking the "three dots" menu option in the split opens a second window
- This window will then stream via [streamlink](https://streamlink.github.io/) and the [mpv player](https://mpv.io/)
- Switching between tabs will change the stream to match

My current issues:
- Non-hacky storage of variables / widget (where would I store these variables?)
- How to "attach / dock" this second window to the chat so it resizes and min/maximized with the main window (how to handle the layout?)
- Volume control (probably not very likely with how the stream is attached to the qt window)

Any suggestions or comments on directions would be most welcomed. Thanks again for the great project.

Edit: You can try out the preliminary builds here:
https://github.com/goldbattle/chatterino2/actions/runs/867344185


TranRed:
>     * Switching between tabs will change the stream to match

This sounds really bad for proxy and offline chatters, who chat in other channels while watching streams. Also what will it do if you have 2 or more channels in a tab. Which one will it open? The first one?

goldbattle:
If the tab is a non-live stream, the stream will still be the last live stream a person clicks on.
Then if a user clicks on another live stream then it will switch.

I should make this a toggle in the settings to disable auto switch.
The original motivation was to clearly keep the video in sync with what chat you seeing.

zneix (changes requested):
Changes to build.yml, .gitignore and BUILDING_ON_WINDOWS.md are totally unrelated, please revert those.

goldbattle:
I screwed it up as I was experimenting with doing rebase merging from the master branch.
I will re-open a new PR after I make some more changes and experiment a bit more.
Sorry for the notifications.

goldbattle:
I created another non-draft PR here, sorry to those that got the extra ping in this one.
https://github.com/Chatterino/chatterino2/pull/2854

PR #2816, opened 2021-05-23
Add some cheeky benchmark tests

pajlada:
Pull request checklist:

# Description

We don't do anything super relevant here, but it just shows how we can
go about adding more benchmarks.

Currently, this builds off the same 'broken' building that tests does,
    and they don't share the 'library', so there are definitely
    improvements that can be made to the usability of this and tests.

Once we do more smart benchmarking, we can add this as a CI job and output benchmark info as a github comment to PRs and what-not in case we see performance downfalls


pajlada:
Closing in favour of #3038

PR #2825, opened 2021-05-26
Match game names exactly when using /setgame

### LINKED ISSUE #2636: `/setgame` matching "too loose"
**Describe the bug**
Helix is sometimes horribly incorrect for game titles. This can potentially result in something like `/setgame terraria` actually setting the game to `Sub Terrania` or `Terraria 2`.

**Potential Solution**
Converting game titles from the previous search to lowercase, then best matching according to that. For example...

Below is the response from helix for searching for the game. If we were to convert the game titles to lowercase and compare the user's input to the game titles, we'd find that it best matches to id `31376` as opposed to just matching to the first entry.
image

**EDIT**: This issue is actually in reference to the already added `/setgame` feature. Currently it doesn't match some games exactly. The fix to this would be to convert the query to lowercase, search for the game, see which game best matches and use the id to set the game.

jammehcow:
Pull request checklist:

# Description

When using `/setgame` the first game returned by Helix is used. This PR aims to fix the shotty behaviour of `/setgame` by attempting to match the user's input with the games returned. It would be nice if this matching could also do some kind of fuzzy search but that's beyond the scope of this PR.

Fixes #2636

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -722,7 +722,7 @@ void CommandController::initialize(Settings &, Paths &paths)
         return "";
     });
     this->registerCommand("/setgame", [](const QStringList &words,
-                                         ChannelPtr channel) {
+                                         const ChannelPtr& channel) {
         if (words.size() < 2)
         {
             channel->addMessage(
@@ -747,7 +747,7 @@ void CommandController::initialize(Settings &, Paths &paths)
                         auto status = twitchChannel->accessStreamStatus();
                         getHelix()->updateChannel(
                             twitchChannel->roomId(), matchedGame.id, "", "",
-                            [channel, games, matchedGame](NetworkResult) {
+                            [channel, games, matchedGame](const NetworkResult&) {
                                 channel->addMessage(makeSystemMessage(
                                     QString("Updated game to %1")
                                         .arg(matchedGame.name)));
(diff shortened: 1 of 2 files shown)


alazymeme:
This fixes things for me, thanks!

src/controllers/commands/CommandController.cpp
```diff
@@ -731,23 +731,42 @@ void CommandController::initialize(Settings &, Paths &paths)
         }
         if (auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get()))
         {
+            auto gameName = words.mid(1).join(" ");
```
zneix, line 734: I think this could be made const
```suggestion
            const auto gameName = words.mid(1).join(" ");
```

src/controllers/commands/CommandController.cpp
```diff
@@ -731,23 +731,42 @@ void CommandController::initialize(Settings &, Paths &paths)
...
                             makeSystemMessage("Game not found."));
                     }
                     else  // 1 or more games
                     {
+                        auto matchedGame = games.at(0);
+
+                        if (games.size() > 1)
+                        {
+                            // attempt to find the best looking game by matching lowercase values
+                            for (auto &game : games)
```
zneix, line 752: Same here
```suggestion
                            for (const auto &game : games)
```

zneix (approved):
I tested game changing and it works just fine, other than couple nitpicks I think it's good to go, so just change those things and it LGTM :+1:

jammehcow:
@zneix good nitpicks, thanks!

pajlada (approved):

Issue #2692, opened 2021-04-26
All messages sending at once after reconnecting


ponaire:
Multiple times a day I'll have an issue where all of a sudden messages won't send, and after manually reconnecting all of the messages that didn't send will be sent at once. If you're not a vip or mod, the chat will be spammed with "sending messages too fast", according to the amount of messages you sent while having the issue. When you're a VIP or mod, after reconnecting your messages completely spam the chat. Sometimes its hard to tell if you're having the issue, since not every message sends when you are a vip/mod anyway. 

If I'm having this issue, will closing chatterino delete the messages, preventing them from all sending at once? I need a reliable way to prevent all of the messages sending after I reconnect so I don't completely spam the chat.

I don't think there is a way to reproduce the behavior, it just seems to happen randomly, but it's probably associated with lag

Chatterino 2.3.0 (commit 58017a754)

Windows 10

This isn't a new bug, it's been happening for months now

jammehcow:
I'm wondering what redundancy options we have for that write connection. 

Either we open 2 (yuck) or we add some better monitoring to it. This is a widespread issue and probably ranks up with the UAF channel point crash in terms of annoying a user.

Felanbird:
This was more or less fixed in #2850 - as long as you don't send a whole bunch of messages in the short reconnect window.

Issue #2356, opened 2021-01-12
Write connection dies and doesn't reconnect


pajlada:
**Describe the bug**
I stopped being able to write messages in chat.

This seems to me like the write connection dying, and the ping messages dying.

@pajlada add qDebug messages for when pings are sent and received and let connections be tagged READ/WRITE so we actually know which connection is sending what message

**To reproduce**
N/A

**Screenshots**

**Chatterino version**
commit a6e23e246864458c9d00c469c3b7306086b8d0ed

**Operating system**
Arch Linux btw

**Additional information**

apa420:
Fixed by #2850

Issue #1831, opened 2020-07-31
Dropping the write connection for twitch (Sending messages)


apa420:
**Describe the bug**

Dropping connection the write connection for twitch (Sending messages).
Reading messages is fine, but randomly I can't send any messages anymore.
Restarting or reconnecting fixes this.

**To reproduce**

Wait for twitch to drop the connection...

**Screenshots**

**Chatterino version**

Chatterino Nightly 2.1.7 (commit 8e202c375, 21.06.2020)

**Operating system**

Windows 8.1 (Oops wrote win 10 before :P), pretty sure it also happens on arch

**Additional information**

It might be because I've been having packet loss issues with my internet.
Maybe there is a way to check if this connection dropped so we can reconnect?

Mm2PL:
Sometimes happens for me too. It feels very random though.
My OS: Ubuntu Focal.

gempir:
Can confirm I have had this happen to me multiple times aswell on Windows 10 and Ubuntu. Maybe once a week or so

Felanbird:
Have it, friend complained as well. Both windows 10

apa420:
Related #1500

apa420:
Maybe it's an issue with twitch servers then? Is there some way we can hack a fix for it?

Mm2PL:
> Maybe it's an issue with twitch servers then? Is there some way we can hack a fix for it?

Reconnecting after a dropped message is out of the question, since that can happen with spam filters. What else could be done?
If it is a Twitch issue we would have to bother them to fix it pepeLaugh

gwua:
This has been happening for as long as I can remember. Last time it happened, I was downloading a file from Google Drive, which also caused a Twitch stream I was watching to stutter. OS is Windows 10.

apa420:
I'm a bit unaware on how the connections work, but is it possible to send a PING on the write connection and see if we get a response? And if not try and reconnect? Or is this already something that exists.

gempir:
This has been coming up recently a lot again. Lots of user complaints. Probably because TMI is dying a lot more, but still we could do a better job of reconnecting.

pajlada:
Looking into this further in this PR: https://github.com/Chatterino/chatterino2/pull/2347

SEAPUNK:
One thing I noticed is that it sometimes takes multiple reconnects before it "flushes" the writes - I get multiple "you're sending messages too quickly" messages upon reconnecting.

apa420:
[...]

PR #2850, opened 2021-06-01
Fix write connection reconnection issues

Bun:
Ready for review/feedback. There's a lot of room for (additional) refactoring on the connection logic, but I figured I'd at least fix the issue first.

Pull request checklist:

# Description

Chatterino currently does not handle disconnects of the write connection, which
causes messages to remain buffered until you manually reconnect, after which it
will send all messages at once.

This patch addresses the lack of reconnection specifically, and attempts to
refactor some of the reconnection logic to make it more robust. Backoff behavior
is introduced to *both* read/write cons to avoid reconnection loops.

To replicate the issue, simply lose your (write) connection to TMI. This can be
"simulated" using something like:

    tcpkill -i eth0 host 44.226.36.141 or host 34.217.198.238 or host 100.20.159.232

An alternative/secondary approach would be to detect failures of the underlying
`sendRaw` method (as it returns false if the message can't be sent to the
connection), but better monitoring of the connection state should be sufficient.

The way the code is setup currently doesn't really make it easy to implement
this "cleanly". E.g. calls `initializeConnection` are skipped on reconnect, but
since we're reconnecting with the same parameters, this should have little
actual consequence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Minor: Added `/popout` command. Usage: `/popout [channel]`. It opens browser chat for the provided channel. Can also be used without arguments to open current channels browser chat. (#2556, #2812)
 - Minor: Improved matching of game names when using `/setgame` command (#2636)
 - Bugfix: Fixed FFZ emote links for global emotes (#2807, #2808)
+- Bugfix: Fix reconnecting when IRC write connection is lost
 
 ## 2.3.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Minor: Added `/popout` command. Usage: `/popout [channel]`. It opens browser chat for the provided channel. Can also be used without arguments to open current channels browser chat. (#2556, #2812)
 - Minor: Improved matching of game names when using `/setgame` command (#2636)
 - Bugfix: Fixed FFZ emote links for global emotes (#2807, #2808)
-- Bugfix: Fix reconnecting when IRC write connection is lost
+- Bugfix: Fix reconnecting when IRC write connection is lost (#1831, #2356, #2850)
 
 ## 2.3.2
 
--- a/src/providers/irc/IrcConnection2.cpp
+++ b/src/providers/irc/IrcConnection2.cpp
@@ -1,8 +1,12 @@
 #include "IrcConnection2.hpp"
 
+#include "common/QLogging.hpp"
 #include "common/Version.hpp"
 
 namespace chatterino {
+
+// The minimum interval between attempting to establish a new connection
+const int RECONNECT_MIN_INTERVAL = 15000;
 
 namespace {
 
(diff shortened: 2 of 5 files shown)


fourtf:
big if true

Mm2PL:
huge if factual

leon-richardt:
enormous if accurate

alazymeme:
sizeable if veracious

Bun:
Account switching (including updating your own OAuth token) doesn't work properly, because the disconnect triggered by `AbstractIrcServer::connect` causes it to autoreconnect with the current credentials before `initializeConnection` is called. It does reconnect eventually (after 15s), but that's not how it should work.

Working on a fix.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Minor: Added `/popout` command. Usage: `/popout [channel]`. It opens browser chat for the provided channel. Can also be used without arguments to open current channels browser chat. (#2556, #2812)
 - Minor: Improved matching of game names when using `/setgame` command (#2636)
 - Bugfix: Fixed FFZ emote links for global emotes (#2807, #2808)
+- Bugfix: Fix reconnecting when IRC write connection is lost
```
Felanbird, line 13:
```suggestion
- Bugfix: Fix reconnecting when IRC write connection is lost (#1831, #2356, #2850)
```
adding issue #1831, #2356 and this pr to the changelog

Felanbird: also referencing #2692, even though you stated you were working on the reconnection aspect

Bun:
The behavior should be OK now. (Also shouldn't have force pushed a formatting mistake, oops.)

src/providers/irc/IrcConnection2.cpp
```diff
@@ -13,40 +17,96 @@ namespace {
...
+                this->waitingForPong_ = false;
                 return;
             }
 
-            if (!this->recentlyReceivedMessage_.load())
+            if (this->waitingForPong_.load())
+            {
+                // The remote server did not send a PONG fast enough; close the
+                // connection
+                this->expectConnectionLoss_ = true;
```
pajlada, line 93: This seems like a noop since we already set `this->expectConnectionLoss_ = true;` in the `IrcConnection::close()` method

src/providers/irc/IrcConnection2.cpp
```diff
@@ -13,40 +17,96 @@ namespace {
...
+                     });
+
+    QObject::connect(
+        this, &Communi::IrcConnection::socketStateChanged, this,
+        [this](QAbstractSocket::SocketState state) {
+            if (state == QAbstractSocket::UnconnectedState)
+            {
+                if (!this->expectConnectionLoss_.load())
+                {
+                    // Ensure we always notify our parent that we should
```
pajlada, line 33: Could you expand on the comments here explaining under what circumstances we wouldn't want to call connectionLost here?

pajlada (changes requested):
Looks good to me - I'm fine to merge this in pending some comments being added

Thanks a lot for looking into this!

Bun:
Shuffled some state resets around and improved the comments. Let me know if that's sufficient.

pajlada:
> Shuffled some state resets around and improved the comments. Let me know if that's sufficient.

Perfect, thank you.

I'll merge this into master and have people take this for a spin in the nightly version!

pajlada (approved):

PR #2854, opened 2021-06-03
Attached video player

goldbattle:
Pull request checklist:

# Description

I have been experimenting with including an internal / attached video player which would be docked alongside the chat. This would allow for a pretty seamless viewing and chatting experience if possible. I have gone and done a quick demo of what this could look like and it seems to only be a few modifications so far. Also please see the original draft PR I created https://github.com/Chatterino/chatterino2/pull/2805.

Originally I stumbled upon the [StreamLinkerino](https://github.com/JohnCiubuc/StreamLinkerino) project, but I want to make it a more first class feature. Here is an example screenshot of what I have so far:

image

image

image

## What works:
- Clicking the "three dots" menu option in the split opens a second window
- This window will then stream via [streamlink](https://streamlink.github.io/) and VLC
- Switching between tabs will change the stream to match if the tab / split you click on is live currently
- There is an option in the settings to disable the auto follow behavior which means the "three dot" menu will need to be used.
- Volume of the stream can be controlled with scroll wheel when hovering over the stream (tested on windows)

Any suggestions or comments on directions would be most welcomed. Thanks again for the great project.

You can try out the preliminary builds here:
https://github.com/goldbattle/chatterino2/actions/runs/1276311021


goldbattle:
@pajlada @zneix may I request a review / feedback on this feature? Thanks.

src/singletons/Settings.hpp
```diff
@@ -318,7 +318,13 @@ class Settings : public ABSettings, public ConcurrentSettings
     QStringSetting streamlinkPath = {"/external/streamlink/customPath", ""};
     QStringSetting preferredQuality = {"/external/streamlink/quality",
                                        "Choose"};
-    QStringSetting streamlinkOpts = {"/external/streamlink/options", ""};
+    QStringSetting streamlinkOpts = {
+        "/external/streamlink/options",
+        "--twitch-low-latency --twitch-disable-ads"};
```
pajlada, line 323: We should not be setting default options for an external application

src/util/AttachedPlayer.cpp
```diff
@@ -0,0 +1,61 @@
+#include "AttachedPlayer.hpp"
+
+using namespace chatterino;
```
pajlada, line 3: Should stick this into the chatterino namespace instead of having to use `using namespace`

pajlada (changes requested):
I only have nitpicks for code - the functionality seems nice and I can see a decent amount of people liking this. May need to work on the accessibility of how to install mpv and what-not (and also ensure that we don't break any licenses with this)

Since this is a major addition to the project, we should get @fourtf 's thoughts to ensure it aligns with his ideas for Chatterino.

src/singletons/Settings.hpp
goldbattle: I went ahead and removed them, but the `--twitch-low-latency` option is pretty important for the experience to match the main website. I think most users would appreciate at least this recommendation, but I understand.

Mm2PL: Low latency looks like something that should be in the quality chooser dialog, not hidden somewhere deep in the settings of the application. IMO

goldbattle: What about adding these as "checkboxes" for the options so the user can easily toggle these, and then optionally add more. The issue there is that they might break if the streamlink project renames them at any point.
image

goldbattle: I went an added some nice toggles for the user to use.
image

goldbattle:
Could I request a review from @pajlada and @fourtf? Much appreciated!

zneix:
Sorry for no response on this, I had too many things to take care of first. I will review this once I finish #2905

goldbattle:
@fourtf @pajlada ping for review / feedback on this, thanks!

pajlada:
I'm still for this change, I think enough power-users will find it handy to be worth adding.

I have some thoughts around the way we pass options through to streamlink, but those are implementation details I'd rather take at the end of the implementation.

For now there's a bunch of merge conflicts that have popped up because we haven't gotten around to this PR yet - I would love to still have @fourtf's opinion on whether this functionality fits in his idea of Chatterino

goldbattle:
Thanks for the comments, I went ahead and merged in the latest master.
If you have implementation suggestions feel free to either comment or just give me an explanation on what I should change.

I saw @fourtf made PR https://github.com/Chatterino/chatterino2/pull/3222 also.

killertofus:
could we also use vlc?

goldbattle:
@killertofus I don't know why I didn't try this first, great point. I have moved everything over to using VLC. It did require some logic to kill the streamlink thread directly as VLC will keep playing in the background unlike the original MPV player.

This is much nicer as it allows for using your scroll wheel to adjust the volume of the stream directly. Also most people that have streamlink installed already have VLC, which also has a nicer installer. In contrast to the webkit versions (e.g. PR #3222) this should be cross platform.
 
@pajlada @fourtf I think this is a major feature that should be included, could you please do a review so this can get merged ASAP? Thanks!

SukoXI:
such a nice program... im having troubles w make command coz im not good at this, so can u pls give a pre-build link or smth like that, it would be very helpful, even a yt tutorial wouldn't be bad

goldbattle:
You can find some builds here:
https://github.com/goldbattle/chatterino2/actions/runs/1657018715
[chatterino-windows-x86-64-5.12.10-cmake.zip (2).zip](https://github.com/Chatterino/chatterino2/files/7812550/chatterino-windows-x86-64-5.12.10-cmake.zip.2.zip)

SukoXI:
ThankEgg dude, thank u so much

pajlada:
Closing this for now, please spin up a discussion or reach out to me before working more on this!

PR #2550, opened 2021-03-20
Implemented bit emotes

zneix:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

This pull request adds support for "Tabbing" and picking Bit emotes from emote picker menu. Also works with the "colon-emote-picker". To get this to work, we make a request to the Ivr API directly from chatterino - maybe I will have to redo it to make a request via chatteirno/api, however I didn't want to complicate things this much. Currently, emotes are refreshed every 10 minutes in `TwitchEmotes::loadUserstateEmotes`.

## Little To-Do
- [x] Reload bit emotes on F5
if there'll be any need for that, may be a part of another PR
~~- [ ] Move emoteset fetching to chatteirno/api~~ explained below
- [x] Decide whether or not to refresh emotes periodically
decided this is fine

Would close #1450 #2183 #2388

PR #2856, opened 2021-06-03
Split list of emote sets into bunches when performing Ivr API reqeusts

zneix:
Pull request checklist:

# Description

Added a helper vector of vectors and some logic that splits QStringList that we receive as a function parameter to avoid getting 400's from Ivr API as it supports fetching 100 emote sets at max.

One unrelated change: updated a link to Ivr API endpoint we're using because it has changed.

P.S. Thanks @talneoran for helping to come up with the solution Okayga

### DIFF
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -316,7 +316,7 @@ void TwitchAccount::loadUserstateEmotes(QStringList emoteSetKeys)
                              newEmoteSetKeysVector.begin() + last);
     }
 
-    for (const auto &batch : std::move(batches))
+    for (const auto &batch : batches)
     {
         QStringList emoteSetBatch;
 
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -301,54 +301,80 @@ void TwitchAccount::loadUserstateEmotes(QStringList emoteSetKeys)
         return;
     }
 
-    getIvr()->getBulkEmoteSets(
-        newEmoteSetKeys.join(","),
-        [this](QJsonArray emoteSetArray) {
-            auto emoteData = this->emotes_.access();
-            for (auto emoteSet : emoteSetArray)
-            {
-                auto newUserEmoteSet = std::make_shared<EmoteSet>();
+    // splitting newEmoteSetKeys to batches of 100, because Ivr API endpoint accepts a maximum of 100 emotesets at once
+    constexpr int batchSize = 100;
 
-                IvrEmoteSet ivrEmoteSet(emoteSet.toObject());
+    std::vector<std::vector<QString>> batches;
 
-                newUserEmoteSet->key = ivrEmoteSet.setId;
+    auto newEmoteSetKeysVector = newEmoteSetKeys.toVector();
+    batches.reserve((newEmoteSetKeysVector.size() + 1) / batchSize);
 
-                auto name = ivrEmoteSet.login;
-                name.detach();
-                name[0] = name[0].toUpper();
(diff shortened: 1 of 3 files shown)


src/providers/twitch/TwitchAccount.cpp
```diff
@@ -301,54 +301,80 @@ void TwitchAccount::loadUserstateEmotes(QStringList emoteSetKeys)
...
+                    IvrEmoteSet ivrEmoteSet(emoteSet.toObject());
 
-                    emoteData->allEmoteNames.push_back(cleanCode);
+                    newUserEmoteSet->key = ivrEmoteSet.setId;
 
-                    auto twitchEmote =
-                        getApp()->emotes->twitch.getOrCreateEmote(id, code);
-                    emoteData->emotes.emplace(code, twitchEmote);
+                    auto name = ivrEmoteSet.login;
+                    name.detach();
```
Mm2PL, line 341: Why is this needed?

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -301,54 +301,80 @@ void TwitchAccount::loadUserstateEmotes(QStringList emoteSetKeys)
...
-                newUserEmoteSet->type = QString();
-                newUserEmoteSet->channelName = ivrEmoteSet.login;
+    for (const auto &batch : std::move(batches))
+    {
+        QStringList emoteSetBatch;
+
+        for (const auto &el : batch)
+        {
+            emoteSetBatch.push_back(el);
+        }
```
Mm2PL, line 326: If `batches` was `std::vector<QStringList>` (or something along those lines) from the start you wouldn't have to do this.

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -301,54 +301,80 @@ void TwitchAccount::loadUserstateEmotes(QStringList emoteSetKeys)
...
+                    for (const auto &emote : ivrEmoteSet.emotes)
+                    {
+                        IvrEmote ivrEmote(emote.toObject());
+
+                        auto id = EmoteId{ivrEmote.id};
+                        auto code = EmoteName{ivrEmote.code};
+                        auto cleanCode =
+                            EmoteName{TwitchEmotes::cleanUpEmoteCode(code)};
+                        newUserEmoteSet->emotes.emplace_back(
+                            TwitchEmote{id, cleanCode});
```
Mm2PL, line 357: This constructs a `TwitchEmote` and then passes it to the copy constructor (via `emplace_back`), feels weird to me. I believe you either want
```suggestion
                        newUserEmoteSet->emotes.emplace_back(
                            id, cleanCode);
```

or

```suggestion
                        newUserEmoteSet->emotes.push_back(
                            TwitchEmote{id, cleanCode});
```

Mm2PL (changes requested):
mostly comments :)

src/providers/twitch/TwitchAccount.cpp
zneix: Copy-pasta from `TwitchAccount::loadEmoteSetData`, it wasn't actually introduced here, but in #2550 - this is just indentation change - however if this should be changed I could as well make a change in this PR.

src/providers/twitch/TwitchAccount.cpp
zneix: Thank you!

zneix: This works fine on Qt 5.15, however it doesn't work so well on Qt 5.12, so leaving this workaround for now.

Mm2PL (approved):

pajlada (approved):

PR #2857, opened 2021-06-03
Refactored emote reloading

### LINKED ISSUE #2631: Bit emotes are not present in `twitchAccount->emoteData` in some cases
**Describe the bug**
Sometimes bit emotes are not "tabable" and not present in emote picker. 

**To reproduce**
N/A, this occurs randomly. 

**Screenshots**
N/A

**Chatterino version**
Latest master

**Operating system**
Arch GNU/Linux 

**Additional information**
This occurs because we currently clear `emoteData` upon fetching subscriber emotes from Kraken and fetching other emote sets from USERSTATE message only appends to the - mentioned object, we'd have to find an alternative around it. Alternatively, offer a way of manually reloading bit emotes (currently, those update on its own every USERSTATE, but not more often than 10 minutes).

zneix:
Pull request checklist:

# Description

These changes ensure that upon reloading emotes with `reloadChannelAndSubscriberEmotes` function also calls Ivr API and fetches `USERSTATE` emotes. We store last received `USERSTATE` and use it upon reloading.

Closes #2631


src/widgets/splits/Split.cpp
```diff
@@ -823,9 +823,11 @@ void Split::showSearch()
 
 void Split::reloadChannelAndSubscriberEmotes()
 {
-    getApp()->accounts->twitch.getCurrent()->loadEmotes();
+    auto currentUser = getApp()->accounts->twitch.getCurrent();
     auto channel = this->getChannel();
 
+    currentUser->loadEmotes();
```
Mm2PL, line 829: I don't see the point of this change pajaDent

zneix: I think I needed `currentUser` property to call some other methods on it. Currently it can be reverted, you're correct. Thanks for noticing this!

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -257,41 +255,50 @@ void TwitchAccount::loadEmotes()
                           });
                 emoteData->emoteSets.emplace_back(emoteSet);
             }
+            // Getting userstate emotes from Ivr
+            // Without delaying this, deadlock occurs
+            QTimer::singleShot(1, [this] {
+                this->loadUserstateEmotes();
+            });
```
pajlada, line 262: The singleShot is only necessary because both `loadUSerstateEmotes` and our current scope locks `this->emotes_`. You can 'simply' scope the access to emoteData on line 216 so that it's unlocked before `loadUserstateEmotes` is called. See https://gist.github.com/pajlada/5376a2d3ea3abbc904aeaf83bc46d6c0

pajlada: Fixed this in [`8f691c0` (#2857)](https://github.com/Chatterino/chatterino2/pull/2857/commits/8f691c094d593acb22609584d99aad20d98a8b9a)

pajlada (approved):
These changes look fine to me - please review my two commits and merge this with master if they look good to you @zneix 

Basically what I've changed is we call `loadUserstateEmotes` whenever the `emote-sets` key from `USERSTATE` has changed

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -497,16 +497,12 @@ void IrcMessageHandler::handleUserStateMessage(Communi::IrcMessage *message)
...
+    bool emoteSetsChanged = currentUser->setUserstateEmoteSets(
         message->tag("emote-sets").toString().split(","));
 
-    // we have to call loadUserstateEmotes if this is a first USERSTATE
-    // after currentUserChanged signal emitted
-    if (shouldLoadEmotes)
+    if (emoteSetsChanged)
     {
-        // TODO: figure out a nice way to reload only emotes from USERSTATE
+        currentUser->loadUserstateEmotes();
```
zneix: Note for the future:
My biggest concern about calling `loadUserstateEmotes` here is that even though we filter out already fetched emotesets, we might append some emotes to `emoteData` without flushing it. It could lead to some emotesets being loaded in twice.
Knowing Twitch something like emoteset changing ID could happen. Leaving the comment just if something like this could ever occur.

zneix:
LGTM, Thanks for taking care of the hacky `singleShot` solution!

PR #2862, opened 2021-06-04
Refactored system message received upon trying to send a message while anonymous

zneix:
Pull request checklist:

# Description

I CAAAAAAN'T MegaLUL

Moment when I reconnected, I switched to anonymous mode

This also gets rid of mentioned notice upon sending an empty message (usually happens upon executing a command while being anonymous)

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -328,30 +328,33 @@ void TwitchChannel::sendMessage(const QString &message)
 
     if (!app->accounts->twitch.isLoggedIn())
     {
-        if (!message.isEmpty())
+        if (message.isEmpty())
         {
-            const auto linkColor = MessageColor(MessageColor::Link);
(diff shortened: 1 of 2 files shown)


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -328,11 +328,33 @@ void TwitchChannel::sendMessage(const QString &message)
...
+            return;
+        }
+
+        const auto linkColor = MessageColor(MessageColor::Link);
+        const auto accountsLink = Link(Link::OpenAccountsPage, QString());
+        const auto currentUser = getApp()->accounts->twitch.getCurrent();
+        const auto expirationText =
+            QString("You need to log in to send messages. You can link your "
+                    "Twitch account");
+        const auto loginPromptText = QString(" in the settings.");
```
pajlada, line 342: The spacing is already added elsewhere, so need to add the space here
```suggestion
        const auto loginPromptText = QString("in the settings.");
```

pajlada (approved):

PR #2864, opened 2021-06-04
Fix moderation button time units when not using seconds

### LINKED ISSUE #1719: Moderation buttons not showing the correct time
When using moderation mode, the period for which timeouts are administered is displayed incorrectly when not using seconds as a time unit. Meaning the command /timeout {user} 1d will be displayed as a 1s timeout in modmode. The command will still administer a 1d timeout, though.

Enter a timeout command with any unit that isn't seconds in it, up to 2w of course.

Currently using Chatterino 2.1.7 for Win 10.

talneoran:
Pull request checklist:

# Description
This PR fixes the displayed time unit in moderation buttons, when using time units other than seconds in the action.

For example:
In the current behavior, the action `/timeout {user} 1m` would show as `1s` in the button text. After the fix it would show as `1m`.
This applies to minutes (`m`), hours (`h`), days (`d`), and weeks (`w`).

Note that when specifying a time that isn't divisible by the closest unit, it will be rounded down, i.e. `61m` displays as `1h`.
This is the case with the current behavior, I also considered an alternative implementation that would show the accurate time, however this can lead to some bad looking buttons (`1441m` instead of rounding to `1d=1440m`).
I would like some feedback on which option is preferred.

Fixes #1719


jammehcow:
When putting in a stupidly large timeout value (such as `7000d`) the numeric value can extend past the button into the user's name:

Pre-change:
image

Post-change:
image

Silly edge case but it exists.

alazymeme:
> When putting in a stupidly large timeout value (such as `7000d`) the numeric value can extend past the button into the user's name:
> 
> Pre-change:
> image
> 
> Post-change:
> image
> 
> Silly edge case but it exists.

The maximum timeout duration is 2 weeks. I think we should limit what the user can input in the settings more than anything.

talneoran:
> 
> 
> When putting in a stupidly large timeout value (such as `7000d`) the numeric value can extend past the button into the user's name

Note that this already happens pre-change, just if you specify the stupidly large timeout value in seconds/with no unit:
On nightly, with the action `/timeout {user} 864000000` I get this button:
image
My change only generalizes the implementation to correctly display durations specified in units other than seconds (and also apparently adds weeks to the possible displayed durations)

However, I do agree that some sort of limit can/should be put into place, either on the duration or on the number displayed in the button (can maybe limit the button to display max 2w?). With how this is currently implemented the duration might also suffer from an overflow.
I don't mind adding something like this, just not sure if it should happen as a part of this PR.

jammehcow:
That's fair, happy for that to be fixed in a separate issue.

Mm2PL (approved):
works as expected

pajlada (approved):
:+1:

PR #2866, opened 2021-06-06
Add username autocompletion popup menu

### LINKED ISSUE #1979: Suggest viewers when user types @name
**What should be added?**

When you type @ and start typing, it should suggest you names of people in chat.

Here's an example:

image

**Why should it be added?**

I think it's a very common and very useful feature. Even as I write this Github issue, it has that feature. Even the emotes system has that feature using `:`.

Feature | Tab-completion | Instant search
------------ | ------------- | -------------
Emotes | ✔️ | ✔️ type `:` and start typing
Usernames | ✔️ | ❌

talneoran:
Pull request checklist:

# Description
This PR adds a popup menu for autocompleting usernames with an '@' prefix, similar to the existing one for emotes with ':'.
- Showing the popup is an optional feature with a toggle setting turned on by default, when the setting is off you can still autocomplete as usual.
- The strategy for finding usernames to complete should be the same as for the existing @ autocompletion.

When testing, also test the emote autocompletion menu, as this addition can affect it, and the implementation is also inspired by it.
Also looking for feedback on the appearance of the menu.

Closes #1979


MrAuro:
Emote and username autocompletion menus work as intended for me.

One thing about the appearance. I think it would be nice to have some sort of indicator like DankChat does to indicate that it is a person. I provided some examples below:

DankChat:
DankChat

Current PR:
Current PR

Requested Changes:
Requested Changes

Also the username doesn't seem to be vertically centered.

Mm2PL:
I dislike the code duplication. I would probably have reused the existing :-popup code and allowed it to show text without images. I haven't tested the functionality yet.

talneoran:
> One thing about the appearance. I think it would be nice to have some sort of indicator like DankChat does to indicate that it is a person. 

Not sure about this. The web chat equivalent contains only text, and unlike the DankChat example you provided, here it is clear when you are auto completing an emote vs a username based on the prefix used. Open to more opinions about this though.

>Also the username doesn't seem to be vertically centered.

I think this is also the case in the emote completion popup, just less noticeable. I can look into it though.

> I dislike the code duplication. I would probably have reused the existing :-popup code and allowed it to show text without images.

I mostly agree, it just wasn't instantly obvious what the best way to combine this with the existing emote menu is, and I wanted to get something that works quickly. I can definitely try to do some more combining of the functionality and improve the code. If you have any more specific ideas I'd like to hear them.

talneoran:
I combined the code for the emote completion menu and the new username menu, now definitely test the emote completion as well when testing as it is handled by the same code.
Also usernames should be vertically centered now.

pajlada (approved):
Looks good to me
I don't think we need to differentiate between users and emotes the same way DankChat does considering these two popups are activated in different ways (`@` for user completion and `:` for emote completion).

PR #2880, opened 2021-06-10
Communicate streams that went offline through /live

### LINKED ISSUE #2791: /live doesn't log properly
**Describe your issue**
Title

1. Navigate to Channels -> live
2. Look at live log file
3. See a log with only timestamps
4. 🙍‍♀️

Also while I have this issue open, should discuss moving /live logs to a main folder along with whispers & mentions just for consistency sake? Since live is it's own custom tab option like whispers/mentions.

**Screenshots**
image

**OS and Chatterino Version**
win10 2db50f1

### LINKED ISSUE #2799: unsure if people in `/live` already went offline
**Describe your issue**
I cannot see when streamers go offline in my `/live` tab. So I sometimes look for a stream and it's already offline.

**Proposed solution**
Just add `%s is offline` messages to `/live`

**Chatterino Version**
Chatterino + Dankerino patches 9e5314fc6a1c237b7f36c97b5ce478739a865673 (based on 2db50f1c)

**OS version**
Ubuntu 20.04.01 LTS

Mm2PL:
Pull request checklist:

# Description

Adds `channel is now offline.` messages to the `/live` channel. Closes #2799. Fixes #2791


apa420:
LGMT, but wouldn't it be better to just gray out the previous channel is online message instead maybe (in the /live split)?
I feel like that would make more sense in a /live tab since you'll have a lot of channels going live or offline so it might be hard to track if a channel went offline or is still live.

apa420 (approved):
LGTM, but wouldn't it be better to just gray out the previous channel is online message instead maybe (in the /live split)?
I feel like that would make more sense in a /live tab since you'll have a lot of channels going live or offline so it might be hard to track if a channel went offline or is still live.

zneix (commented):
Overall the PR looks good and does its job well :+1: 

> wouldn't it be better to just gray out the previous channel is online message instead maybe

I agree with apa on this one though. it will get very unclear and hard to follow/connect messages related to one channel and another.

Mm2PL:
> LGTM, but wouldn't it be better to just gray out the previous channel is online message instead maybe (in the /live split)?

I really like this idea. I didn't even think of such a possibility. Thank you for pointing this out. I'll be sure to add this.

apa420:
Only con with the approach I suggested is you won't know when they went offline.
I'm unsure how relevant that is to people.

leon-richardt:
Just spitballing some ideas, maybe striking through the messages (instead of)|(in addition to) graying them out could also be a possibility? Might make it a little more noticeable

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -607,10 +607,30 @@ void TwitchChannel::setLive(bool newLiveStatus)
...
+                // "delete" old 'CHANNEL is live' message
+                LimitedQueueSnapshot<MessagePtr> snapshot =
+                    getApp()->twitch2->liveChannel->getMessageSnapshot();
+                int snapshotLength = snapshot.size();
+
+                int end = std::max(0, snapshotLength - 200);
+                auto liveMessageSearchText =
+                    QString("%1 is live!").arg(this->getDisplayName());
+
+                for (int i = snapshotLength - 1; i >= end; --i)
```
leon-richardt, line 625: nit: Can these `int`s be `size_t`s? [`LimitedQueueSnapshot<T>::size()`](https://github.com/Chatterino/chatterino2/blob/5512437f1bd5b2be7997b52bf36b04930446f3f1/src/messages/LimitedQueueSnapshot.hpp#L25) returns a `size_t`

leon-richardt (changes requested):

Mm2PL:
Also accidentally fixes #2791.

src/providers/twitch/TwitchChannel.cpp
Mm2PL: These `int`s cannot be `size_t`s because if `snapshotLength` is <0 then it will overflow and then `std::max` will pick the obviously bigger overflowed value. Also this code was copypasted from [here](https://github.com/Chatterino/chatterino2/blob/f1e843a67275cd8e9d5a3c738a504d54b7743c42/src/common/Channel.cpp#L249)

Mm2PL:
@leon-richardt 

> Just spitballing some ideas, maybe striking through the messages (instead of)|(in addition to) graying them out could also be a possibility? Might make it a little more noticeable

I've looked into this. This requires a bunch of changes to rendering fonts. Currently the function that renders text doesn't need to know anything about the outside world like *should this message have text be struck-through*.

goldbattle:
It would also be nice if when a stream goes offline, it reports the up time so the user can know how long of a stream they missed.

pajlada:
> It would also be nice if when a stream goes offline, it reports the up time so the user can know how long of a stream they missed.

This is not trivial to achieve since the easily accessible timestamp of "going live" we have might be the same time we started Chatterino.
Would need to make the `/live` tab a bit more smart and add metadata received by the API request to each `Streamer is live!` message.

pajlada (approved):
LGTM :+1:

PR #2891, opened 2021-06-13
Add setting for username display style

### LINKED ISSUE #2889: Add a option to hide peoples international username
FFZ just recently added the option to hide peoples international usernames or hide usernames so only one of them will show instead of having the username in brackets. I think this would be a good feature for most people, as I found out some people I talk to find it very annoying, 
image
image

Mm2PL:
Pull request checklist:

# Description

Closes #2889


zneix:
Can't test this right now, but will it also work with channel names in split header? (e. g. when I select to only see login name, I won't see non-ascii username in brackets next to login).

Mm2PL:
This is just chat.

pajlada (approved):
Thanks for your contribution

PR #2900, opened 2021-06-19
Fix cheermotes not loading when user is connected to many channels

jammehcow:
Pull request checklist:

# Description

This PR aims to fix the cheermote fetch timeouts when users have a large number of channels/splits open. A 10s timeout was added to the Helix request in order to handle the slow response.

I've personally noticed this with 73 individual channels open but it's been reported for counts as low as 20.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Bugfix: Fixed pasting text with URLs included (#1688, #2855)
 - Bugfix: Fix reconnecting when IRC write connection is lost (#1831, #2356, #2850)
 - Bugfix: Fixed bit emotes not loading in some rare cases. (#2856)
+- Bugfix: Fixed cheer emotes timing out when user has many splits open (#2900)
 
 ## 2.3.2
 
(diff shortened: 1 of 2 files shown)


pajlada:
This is not a rejection nor an approval, but we're solving for the problem of the `NetworkRequest` timeout timer starting before the `NetworkRequest` has *actually* started, so while this may solve the issue for now it's a band-aid solution at best.

jammehcow:
> This is not a rejection nor an approval, but we're solving for the problem of the `NetworkRequest` timeout timer starting before the `NetworkRequest` has _actually_ started...

Will investigate, thanks!

jammehcow:
Did some digging and didn't get very far but I assume these are some of the ways to go about it:

 1. Start the timer when the thread is started (maybe I don't understand how the threadpool works)
 2. Start the timer when the socket emits that state change signal. Wasn't able to find a reference to the socket from the worker :\ 

I'm too lost on this one so either someone else if going to have to pick it up/give guidance or I'll abandon it.

pajlada:
> Did some digging and didn't get very far but I assume these are some of the ways to go about it:
> 
>     1. Start the timer when the thread is started (maybe I don't understand how the threadpool works)
> 
>     2. Start the timer when the socket emits that state change signal. Wasn't able to find a reference to the socket from the worker :\
> 
> 
> I'm too lost on this one so either someone else if going to have to pick it up/give guidance or I'll abandon it.

number 2 is the one that seemed reasonable to me too - it's strange that qt has no real 'timeout' support for their network request (or maybe I just haven't found it either)

PR #2901, opened 2021-06-19
Add a privacy disclaimer for recent-messages API

### LINKED ISSUE #1311: Create opt-in system for recent-messages and other third-party services

Mm2PL:
Pull request checklist:

# Description

Attempt 2. Closes #1311

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Major: Added ability to toggle visibility of Channel Tabs - This can be done by right-clicking the tab area or pressing the keyboard shortcut (default: Ctrl+U). (#2600)
+- Minor: Added opt-in system for recent-messages (#1311, #2901)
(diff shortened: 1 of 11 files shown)


pajlada (changes requested):
We've had an offline chat regarding the feature - and will revisit it

PR #2920, opened 2021-06-24
Resize tooltip thumbnails clientsize to prevent memory leaks

zneix:
Pull request checklist:

# Description

By default we do proxy thumbnails for images in the Chatterino API, however in the case of custom resolvers we directly pass a link to the image. Chatterino then attempts to load those links and in the result if a link is enormously big it may crash Chatterino and even make it to eat up all the available RAM before crashing.
By resizing thumbnails that are bigger than 300 pixels (optimal, hardcoded default value used by the Chatterino API) we are ensuring that we won't try to load into memory something way too big - we won't even see any difference anyway, because big images are resized by the API anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed Chatterino eating up all the RAM and crashing after hovering over certain image links with link thumbnails turned on. (#2920)
+
 ## 2.3.3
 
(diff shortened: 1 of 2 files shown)


fourtf:
fixed in https://github.com/Chatterino/chatterino2/commit/ed6ef0b98e1d1f6ec13eac72b44b8f3aac6ce86a

PR #2921, opened 2021-06-25
Fix large timeout duration in moderation buttons

### LINKED ISSUE #2865: Large timeout times in moderation buttons overlaps user's usernames
**Describe your issue**
Inputting overly large timeout times in the moderation button settings, overlaps users' usernames.

**Screenshots**
image

**OS and Chatterino Version**
Windows 10, Chatterino 2.3.2, Commit 508fea638

_Idea sourced from https://github.com/Chatterino/chatterino2/pull/2864_

talneoran:
Pull request checklist:

# Description
Limits the duration displayed in timeout moderation buttons to the max timeout duration, 2 weeks.

Initially I wanted to cut down the duration inputted by the user, but with the current implementation, that will require changing the inputted action after it has already been saved, which I'm not sure about. 
Instead I opted for simply displaying `>2w` in the button when the duration exceeds the max duration, without changing the underlying action. I would like to hear if anyone has better ideas, or thinks that editing the action somehow is still the right way to go.

Fixes #2865

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed large timeout durations in moderation buttons overlapping with usernames or other buttons. (#2865, #2921)
+
 ## 2.3.3
 
 - Major: Added username autocompletion popup menu when typing usernames with an @ prefix. (#1979, #2866)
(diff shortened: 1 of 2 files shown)


jammehcow (approved):
Logic looks fine to me and tested. Would `2+ w` look better than `>2 w` to you?
To me it does but either way conveys it well enough.

talneoran:
I think that looks a bit better as well, maybe I'll wait to hear another opinion about it and then change it.

pajlada:
I'm fine keeping it as `>2`

pajlada (approved):

zneix (approved):

PR #2752, opened 2021-05-06
Added clickable hosting messages

apa420:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description
Closes #2655

Also makes live and offline messages clickable.

Issue #2655, opened 2021-04-22
Make hosting messages clickable


Mm2PL:
**What should be added?**

I would like to have channel names clickable in hosting messages (`Now hosting CHANNEL.` and `CHANNEL is now offline. Exiting host mode.`). So clicking `CHANNEL` would open a usercard.

**Why should it be added?**

useless qol

PR #2922, opened 2021-06-25
Add clickable usernames in host offline messages

talneoran:
Pull request checklist:

# Description
Makes the channel name in `<channel> has gone offline. Exiting host mode.` messages clickable.
Implements missing functionality from #2752 which was described in #2655.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ## Unversioned
 
+- Minor: Channel name in `<channel> has gone offline. Exiting host mode.` messages is now clickable. (#2922)
+
 ## 2.3.3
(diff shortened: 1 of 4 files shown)


jammehcow:
Looks fine to me! Only gripe is that the username's capitalisation in "Now hosting" can differ to that of "has gone offline"

image

talneoran:
> Only gripe is that the username's capitalisation in "Now hosting" can differ to that of "has gone offline"

Not sure why this happens since both messages are handled similarly. Do you know if the original offline message we get (which is the one shown on the current version) has the username capitalized?

pajlada (approved):

PR #2926, opened 2021-06-26
Add /openurl command

### LINKED ISSUE #2461: Add support for opening links from chat commands
**What should be added?**
Support for opening a link when using a chat command

**Why should it be added?**
Could be quite useful for a variety of things like modview

**Examples**
Command: `/report <user>` would open a webpage: twitch.tv/user/report

talneoran:
Pull request checklist:

# Description
Add a `/openurl <URL>` built-in command that allows opening URLs through chat.
This is mostly useful for usage in custom commands, for example a command `/report` with the definition `/openurl twitch.tv/{1}/report` can be used to open the user report page of the user given as the first argument.

Closes #2461.

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -798,7 +798,9 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         QString link(words[1]);
         if (getSettings()->openLinksIncognito && supportsIncognitoLinks())
+        {
             openLinkIncognito(link);
+        }
         else
         {
             QUrl url = QUrl::fromUserInput(link);
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -800,7 +800,13 @@ void CommandController::initialize(Settings &, Paths &paths)
         if (getSettings()->openLinksIncognito && supportsIncognitoLinks())
             openLinkIncognito(link);
         else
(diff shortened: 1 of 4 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -786,6 +787,31 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+
+    this->registerCommand("/openurl", [](const QStringList &words,
+                                         const ChannelPtr channel) {
+        if (words.size() < 2)
+        {
+            channel->addMessage(makeSystemMessage("Usage: /openurl <URL>."));
+            return "";
+        }
+
+        QString link(words[1]);
```
Mm2PL, line 799: Some URLs can include unencoded spaces (` `; 0x20) (sometimes can appear when copying from a browser) it will lead to behavior that might be unpredictable: the URL getting chopped off at the first space. I think you might as well take every word from `words[1]` onward joined by spaces. `QUrl::fromUserInput()` (or QUrl *tolerant* parsing for that matter) should automatically encode these spaces for you.

tldr `/openurl duck.com?q=testing test test` opens `http://duck.com?q=testing` instead of `http://duck.com?q=testing%20test%20test`

Mm2PL (changes requested):
Works as expected except the comment I made.

pajlada (approved):

Mm2PL (approved):
lgtm

zneix (approved):

PR #2947, opened 2021-07-02
Readd removing of unmodified cache files

talneoran:
Pull request checklist:

# Description

The function `clearCache` in `RunGui.cpp`, that is supposed to delete cache files that weren't modified in the past 14 days, currently does nothing. It was originally added in bc55a05, but the changes in 17aa3f7 removed the actual deleting of the files along with the debug output.

This PR readds the deleting of the files (along with minor changes to the code).
In terms of debug output, I added a count of the files that were successfully deleted, but I am not sure if this is needed (instead of the original "[Cache] cleared cache" message), as well as if a debug output is needed per file that is deleted (I omitted this since it can get spammy if many files are removed)

This could also be an opportunity to maybe change the strategy of when we delete files (for example looking at the time the files were last read, and not only at when they were last modified), I'm find with this as it is but wanted to get some more opinions.


zneix (approved):

src/RunGui.cpp
```diff
@@ -145,17 +145,18 @@ namespace {
...
             if (info.lastModified().addDays(14) < QDateTime::currentDateTime())
             {
-                toBeRemoved << info.absoluteFilePath();
+                bool res = QFile(info.absoluteFilePath()).remove();
+                if (res)
+                    ++deletedCount;
             }
         }
+        qCDebug(chatterinoCache)
+            << "[Cache] deleted" << deletedCount << "files";
```
Mm2PL, line 159: `[Cache]` is redundant with categorized logging (`qCDebug(chatterinoCache)`)
```suggestion
            << "Deleted" << deletedCount << "files";
```

Mm2PL (approved):
just a small nit pick

pajlada (approved):

Issue #1274, opened 2019-09-02
Show subscriber emotes of current channel


fourtf:
**Is your feature request related to a problem? Please describe.**
You cannot view the emotes that you would get by subscribing to a channel from Chatterino.

**Describe the solution you'd like**
Show the emotes.

**Additional context**
It is unclear how this would be done (if reasonable) by using the twitch API. However, there's a twitchemotes API (example: https://api.twitchemotes.com/api/v4/channels/23161357 ).

zneix:
I am interested in implementing this, but I am unsure where should that be located. We already have emotes of current channel being at the top of subscriber emotes tab in the emote picker, so should we follow this and display emotes of current channel at the top? On the other hand, if someone like forsen has 60 emotes wouldn't that be a wall of useless stuff for someone?

PR #2951, opened 2021-07-03
Respect follower emotes context, making them only available in their owner channels

### LINKED ISSUE #2946: Follower emotes, once fetched, are in emoteData in all chats which shouldn't be the case
**Describe your issue**
Once we get the emoteset ID of set with channel's follower emotes, emotes and this set stay in `TwitchAccount::emotes_`. Our current implementation does not take it into account that those emotes are only usable in the scope of channel that those belong to.

The ideal solution would be to add a flag to `EmoteSet` struct which would signify whether an emote can be used locally or not and then respect it inside `CompletionModel.cpp`.

**Screenshots**
N/A

**OS and Chatterino Version**
N/A, Every version since bit emotes were implemented

zneix:
Pull request checklist:

# Description

- [x] Migrated to a v2 IVR API endpoint for querying bulk emote sets, which gives us information about emote type.
- [x] Added `local` property to `EmoteSet` struct, which indicates whether emotes from the set can be used only in their origin channel (applies only to follower emotes at the moment)
- [x] Made `CompletionModel` respect `EmoteSet.local` property
   - [x] Removed `TwitchAccountEmoteData.allEmoteNames` because we don't use it anymore - `CompletionModel` needs more than just `EmoteName`, so it's better to just use `EmoteSet.local` while iterating over EmoteMap with globally available Twitch emotes
- [x] Made `EmotePopup` respect `EmoteSet.local` property
- [x] Made `InputCompletionPopup` respect `EmoteSet.local` property

While I was working on the Pull Request, I also introduced new `Helix::getChannelEmotes` method, however in the later stage of the PR I didn't need it anymore, but it's good to keep it for the future (may be useful for #1274's impelentation).

Closes #2946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Username in channel points rewards redemption messages is now clickable. (#2673, #2953)
 - Minor: Channel name in `<channel> has gone offline. Exiting host mode.` messages is now clickable. (#2922)
 - Minor: Added `/openurl` command. Usage: `/openurl <URL>`. Opens the provided URL in the browser. (#2461, #2926)
 - Bugfix: Now deleting cache files that weren't modified in the past 14 days. (#2947)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Minor: Added `/openurl` command. Usage: `/openurl <URL>`. Opens the provided URL in the browser. (#2461, #2926)
 - Bugfix: Now deleting cache files that weren't modified in the past 14 days. (#2947)
 - Bugfix: Fixed large timeout durations in moderation buttons overlapping with usernames or other buttons. (#2865, #2921)
+- Bugfix: Middle mouse click no longer scrolls in not fully populated usercards and splits. (#2933)
 - Bugfix: Fix bad behavior of the HTML color picker edit when user input is being entered. (#2942)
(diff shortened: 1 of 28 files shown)


src/providers/IvrApi.hpp
```diff
@@ -55,14 +56,20 @@ struct IvrEmote {
...
 
-    IvrEmote(QJsonObject root)
-        : code(root.value("token").toString())
+    explicit IvrEmote(QJsonObject root)
+        : code(root.value("code").toString())
         , id(root.value("id").toString())
         , setId(root.value("setID").toString())
-        , url(root.value("url_3x").toString())
+        , emoteType(root.value("type").toString())
+        , imageType(root.value("assetType").toBool())
```
pajlada, line 68: string

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Bugfix: Now deleting cache files that weren't modified in the past 14 days. (#2947)
 - Bugfix: Fixed large timeout durations in moderation buttons overlapping with usernames or other buttons. (#2865, #2921)
 - Bugfix: Fix bad behavior of the HTML color picker edit when user input is being entered. (#2942)
+- Bugfix: Follower emotes are only suggested (emote popup menu, tab completion, emote input menu) in their origin channel, not globally. (#2951)
```
pajlada, line 10: fix phrasing so it cannot be misinterpreted by swedish roleplayers

pajlada (approved):
LGTM :+1: Thanks for the contribution!

PR #2981, opened 2021-07-08
Added the ability to add nicknames for users

### LINKED ISSUE #137: Add alias feature / tagging users
Like mentioned in [Issue #110 of chatterino](https://github.com/fourtf/chatterino/issues/110) an Alias feature would be nice.

With the alias feature you can give another user a permanent alias, to be able to keep track of users who change their displayed names.

apa420:
Pull request checklist:

# Description
Alias for usernames in twitch chat.

Closes #137


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added alias for usernames. (#137, #2981)
```
talneoran, line 5:
```suggestion
- Major: Added aliases for usernames. (#137, #2981)
```

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -704,6 +704,16 @@ void TwitchMessageBuilder::appendUsername()
...
 
+    auto aliases = *getCSettings().aliasNames.readOnly().get();
+
+    for (auto const alias : aliases)
+    {
+        if (alias.getName().toLower() == this->message().loginName.toLower())
+        {
+            usernameText = alias.getReplace();
+        }
+    }
```
talneoran, line 715: Currently if you have 2 aliases for the same user, the last one in the list is used instead of the first one like what seems to be the case for commands. Maybe an early break here?
Might want to just prevent duplicates somehow but that's another thing.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -704,6 +704,16 @@ void TwitchMessageBuilder::appendUsername()
         break;
     }
 
+    auto aliases = *getCSettings().aliasNames.readOnly().get();
+
+    for (auto const alias : aliases)
+    {
+        if (alias.getName().toLower() == this->message().loginName.toLower())
```
talneoran, line 711: Maybe storing the alias name as lowercase instead of converting it every time?

src/widgets/settingspages/AliasesPage.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <QHeaderView>
+
+namespace chatterino {
+
+AliasesPage::AliasesPage()
+{
+    LayoutCreator<AliasesPage> layoutCreator(this);
+    auto layout = layoutCreator.setLayoutType<QVBoxLayout>();
+
+    layout.emplace<QLabel>("Gives an alias to a user.");
```
talneoran, line 21:
```suggestion
    layout.emplace<QLabel>("Give an alias to a user.");
```
Just to be consistent with imperative verbs in other places (such as highlights, filters, etc)

talneoran (commented):
Seems to work so far, some general notes:

- In terms of naming, I'm wondering if "alias" is the best option for users, as opposed to something like "nickname". I'm fine with alias but wondering what others think.
- I think something like this should be controllable through the usercard, like how the similar BTTV feature works. Maybe a usercard button that allows to add or edit your alias for that user. I suppose this could also be implemented on its own after this is merged though.

src/providers/twitch/TwitchMessageBuilder.cpp
apa420: I would assume the compiler would optimize it but I can explicitly write it as well.

apa420:
>     * I think something like this should be controllable through the usercard, like how the similar BTTV feature works. Maybe a usercard button that allows to add or edit your alias for that user. I suppose this could also be implemented on its own after this is merged though.

I'm only doing a base implementation and am honestly not sure if it's worth implementing such a feature in the first place since it's a lot of work for little gain since it only adds a tiny convenience for a feature you'll probably rarely use (at least the way I see it).

src/providers/twitch/TwitchMessageBuilder.cpp
talneoran: By storing I meant when saving the alias in the first place, but this is also fine 👍

talneoran (approved):

src/widgets/dialogs/SettingsDialog.cpp
```diff
@@ -164,6 +165,7 @@ void SettingsDialog::addTabs()
     this->addTab([]{return new GeneralPage;},          "General",        ":/settings/about.svg");
     this->ui_.tabContainer->addSpacing(16);
     this->addTab([]{return new AccountsPage;},         "Accounts",       ":/settings/accounts.svg", SettingsTabId::Accounts);
+    this->addTab([]{return new AliasesPage;},         "Aliases",       ":/settings/accounts.svg");
```
Mm2PL, line 168: Rest of these are aligned
```suggestion
    this->addTab([]{return new AliasesPage;},          "Aliases",       ":/settings/accounts.svg");
```

src/controllers/aliases/AliasesName.hpp
```diff
@@ -0,0 +1,77 @@
...
+class AliasesName
+{
+public:
+    AliasesName(const QString &name, const QString &replace)
+        : name_(name)
+        , replace_(replace)
+    {
+    }
+
+    const QString &getName() const
```
Mm2PL, line 24: Other [getters](https://wiki.chatterino.com/Contributing%20for%20Developers/#members) in the app follow a pattern like
```suggestion
    const QString &name() const
```

src/controllers/aliases/AliasesName.hpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return this->name_;
+    }
+    const QString &getReplace() const
```
Mm2PL, line 28: same thing here
```suggestion
    const QString &replace() const
```

Mm2PL (changes requested):
Here are some things I found in the code. One behvior I find weird is that @-s aren't transformed to use aliases, but that can be added later in another small PR.

src/controllers/aliases/AliasesName.hpp
apa420: I think I copied the code from https://github.com/Chatterino/chatterino2/blob/master/src/controllers/filters/FilterRecord.hpp#L43 but I guess it doesn't really matter.

zneix (changes requested):

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -704,6 +704,18 @@ void TwitchMessageBuilder::appendUsername()
         break;
     }
 
+    auto aliases = *getCSettings().aliasNames.readOnly().get();
+    auto loginLower = this->message().loginName.toLower();
```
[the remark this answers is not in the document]
apa420: https://github.com/Chatterino/chatterino2/blob/0d947a07f887ea616a543a19f9cfe462405a609c/src/providers/twitch/TwitchMessageBuilder.cpp#L765

Similar to ignore replace I'd like to avoid race conditions if someone would potentially edit the settings while this is ran.

apa420:
To explain a bit about my alias implementation.
I personally would only want alias in the most simple sense possible, the name of the chatter gets changed from one name to another client sided.
For example my friend Che has the username just_che on twitch and I'd like it to just display Che for me personally and I don't really care about auto completing it or anything else.

Making the names actual aliases in the code would mean we'd have to separate aliases and normal usernames and you'll get issues where someone sets the alias to a user that already exists and it feels like it's such a mess for features most people simply won't use.

Searching with `user:` auto completion etc won't work with this pr nor do I think the feature should support it.

gempir:
The highlights menu uses the terminology "Username" instead of "Name" so I think the column name here should be adjusted

image

Mm2PL (approved):

pajlada:
> The highlights menu uses the terminology "Username" instead of "Name" so I think the column name here should be adjusted
> 
> image

Fixed

image

pajlada (approved):

PR #2987, opened 2021-07-10
Wiki redirect cleanup

Felanbird:
Pull request checklist:

~~- [ ] `CHANGELOG.md` was updated, if applicable~~ N/A

# Description
Noticed the redirect last week, forgot to actually PR it, found another one earlier.
Might as well just link directly to the wiki.

### DIFF
--- a/src/widgets/settingspages/ExternalToolsPage.cpp
+++ b/src/widgets/settingspages/ExternalToolsPage.cpp
@@ -106,7 +106,7 @@ ExternalToolsPage::ExternalToolsPage()
         const auto description = new QLabel(
             "You can set custom host for uploading images, like "
             "imgur.com or s-ul.eu.<br>Check " +
-            formatRichNamedLink("https://chatterino.com/help/image-uploader",
+            formatRichNamedLink("https://wiki.chatterino.com/Image%20Uploader/",
                                 "this guide") +
             " for help.");
         description->setWordWrap(true);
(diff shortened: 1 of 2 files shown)


src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -106,7 +106,7 @@ ExternalToolsPage::ExternalToolsPage()
         const auto description = new QLabel(
             "You can set custom host for uploading images, like "
             "imgur.com or s-ul.eu.<br>Check " +
-            formatRichNamedLink("https://chatterino.com/help/image-uploader",
+            formatRichNamedLink("https://wiki.chatterino.com/Image%20Uploader/",
```
Felanbird, line 109:
```suggestion
            formatRichNamedLink("https://wiki.chatterino.com/Image%20Uploader",
```
Doing this from web so I don't have to re-open ide, doing it for  https://github.com/Chatterino/wiki/pull/88#issuecomment-867674003

ALazyMeme: so lazy smh

zneix (approved):

xHeaveny (approved):

Mm2PL (approved):

fourtf:
The links are there so they can be changed (indirection) so it doesn't break in the future.

Felanbird:
> The links are there so they can be changed (indirection) so it doesn't break in the future.

Should we be moving existing wiki links to a chatterino.com alternative so this can always be prevented?

PR #2988, opened 2021-07-11
Added "Open in: new tab, browser player, streamlink" in twitch link context menu.

SLCH:
Pull request checklist:

# Description

Added "Open in: new tab, browser player, streamlink" in twitch link context menu.

<details>
<summary>
New - Old
</summary>

New:

Screenshot_20210711_120131

Old:

Screenshot_20210711_120050

</details>


SLCH:
```
Error: fatal: No url found for submodule path 'lib/humanize' in .gitmodules
```
Getting in actions on submodule loading.
This lib was removed around January.
Not sure why action is trying to load it - there is no mention of it in my diff.

ALazyMeme:
Could be because you added the library?
https://github.com/SLCH/chatterino2/commit/a947282522e524693e61b9a57e7bcfdf0baa4826

resources/contributors.txt
```diff
@@ -42,6 +42,7 @@ Iulian Onofrei | https://github.com/revolter | :/avatars/revolter.jpg | Contribu
 matthewde | https://github.com/m4tthewde | :/avatars/matthewde.jpg | Contributor
 Karar Al-Remahy | https://github.com/KararTY | :/avatars/kararty.png | Contributor
 Talen | https://github.com/talneoran | | Contributor
+SLCH | https://github.com/SLCH | | Contributor
```
ALazyMeme, line 45: Isn't this meant to be in a separate pr?

SLCH: i have no idea, could reverse it.

SLCH:
> Could be because you added the library?
> [SLCH@a947282](https://github.com/SLCH/chatterino2/commit/a947282522e524693e61b9a57e7bcfdf0baa4826)

I did not add it. It was removed since my last PR but i did update my master before writing any changes.
And i checked my diff with current master's and there is no mention of said lib.

resources/contributors.txt
SLCH: Reverted

Felanbird: > i have no idea, could reverse it.

Better to be a separate PR incase reverting of this PR happens, so your Contributor entry isn't lost.

zneix: It can be done in the same PR - I did this myself back in the old days. Feel free to include it in this one, I think there's no harm but if you wanna farm some more commits do it separetly.

zneix: > incase reverting of this PR happens

Why would it though? :thinking:

src/widgets/splits/Split.hpp
```diff
@@ -114,6 +114,10 @@ class Split : public BaseWidget, pajlada::Signals::SignalHolder
     void handleModifiers(Qt::KeyboardModifiers modifiers);
     void updateInputPlaceholder();
 
+    void openChannelInBrowserPlayer(ChannelPtr channel);
+    void openChannelInStreamlink(QString channelName);
+    void joinChannelInNewTab(ChannelPtr channel);
```
pajlada, line 119: Could you add brief comments here explaining what the functions do?

src/widgets/splits/Split.cpp
```diff
@@ -163,9 +163,30 @@ Split::Split(QWidget *parent)
...
+            case (FromTwitchLinkOpenChannelIn::Split):
+                this->openSplitRequested.invoke(channel);
+                break;
+            case (FromTwitchLinkOpenChannelIn::Tab):
+                this->joinChannelInNewTab(channel);
+                break;
+            case (FromTwitchLinkOpenChannelIn::BrowserPlayer):
+                this->openChannelInBrowserPlayer(channel);
+                break;
+            case (FromTwitchLinkOpenChannelIn::Streamlink):
```
pajlada, line 182: please remove the parenthesis from the cases

src/widgets/splits/Split.cpp
```diff
@@ -309,6 +330,41 @@ void Split::updateInputPlaceholder()
     this->input_->ui_.textEdit->setPlaceholderText(placeholderText);
 }
 
+void Split::joinChannelInNewTab(ChannelPtr channel)
+{
+    // Code opening a new tab was taken from:
+    // src/widgets/dialogs/switcher/NewTabItem.cpp#22 - NewTabItem::action
```
pajlada, line 336: Useful comment for me now while reviewing it, but unnecessary to keep in the code

pajlada (changes requested):
Small nitpicks, otherwise seems good :ok_hand:

pajlada (approved):

src/widgets/splits/Split.cpp
```diff
@@ -595,13 +649,7 @@ void Split::openWhispersInBrowser()
...
 void Split::openBrowserPlayer()
 {
-    ChannelPtr channel = this->getChannel();
-    if (auto twitchChannel = dynamic_cast<TwitchChannel *>(channel.get()))
-    {
-        QDesktopServices::openUrl(
-            "https://player.twitch.tv/?parent=twitch.tv&channel=" +
-            twitchChannel->getName());
-    }
+    this->openChannelInBrowserPlayer(this->getChannel());
```
zneix, line 652: We could try to just replace (and potentially delete) `Split::openBrowserPlayer` with `Split::openChannelInBrowserPlayer` since we just have 2 functions that do literally the same, but maybe that's something that should be done in separate PR so just making a note for the future.

zneix (approved):
Looks good, just got one thought to share - potentially for the future PR though.

src/widgets/splits/Split.cpp
SLCH: Duplicated cause `Split::openBrowserPlayer` was a qt slot and meant for current Split while `Split::openChannelInBrowserPlayer` takes any channel. Same about `Split::openChannelInStreamlink`.

zneix: > Duplicated cause Split::openBrowserPlayer was a qt slot and meant for current Split

Correct! Passing an empty channel or simply the current split's channel could be the way to go imho.

SLCH: > Correct! Passing an empty channel or simply the current split's channel could be the way to go imho.

I implemented it like this at first (with default channel arg = nulptr) but it wouldn't compile and later changed to current version. 
Something to do with slots. My C++ isn't very good :smile:

PR #2992, opened 2021-07-11
Fixed newly uploaded subscriber emotes not being available

### LINKED ISSUE #2935: Newly created sub emotes are not present in emote picker
Due to recent changes in Twitch's emote backend (they now use `emotesv2_` IDs) Kraken endpoint fetching all user's emotes does not recognize new user emotes anymore. Thus, we are missing some data and we completely "omit" the new emotes.
Worth noting that since we do a workaround with the emotesets separetly from `USERSTATE` message, this only affects newly added T1 (and possibly partner's T2/T3? not sure...) subscriber emotes.

To fix this, we'd need to get a Helix endpoint for getting authenticated user's emotes which should recognize emotes with `emotesv2_` IDs.
There's already a [uservoice ticket](https://twitch.uservoice.com/forums/310213-developers/suggestions/43599900-get-the-authenticated-user-s-emotes-equivalent-to) for that, but from my experience it'll take a while to get this endpoint in Helix...

### LINKED ISSUE #3007: Bit emotes and GIF emotes rendered multiple times in the emote menu
[Windows 10, 2.3.3 Release]

Today I updated to the newest 2.3.3 and noticed that in the emote menu emotes that I've unlocked with bits, as well as GIF emotes, are repeated multiple times. API-related bug, perhaps? Here's the screenshot:
image

zneix:
Pull request checklist:

# Description

- Added a handle function to `GLOBALUSERSTATE` irc message.
- Moved initial loading of user's emotes to `IrcMessageHandler::handleGlobalUserStateMessage` - it's done, so that we're sure we've already have `emote-sets` cached in `TwitchAccount::userstateEmoteSets_`.
- Refactored whole logic regarding (re)loading user's emotes.

Current order of handling emotes:
1. Load emote sets from IVR
1. Set which emotes failed to resolve.
1. Call Kraken's `getUserEmotes`.
1. Fill up TwitchAccount::emoteData with emote sets that were returned in a Kraken call, but aren't present in emoteData.

We still continue to call `TwitchAccount::loadUserstateEmotes` in case of follower emotes (which are not present in `GLOBALUSERSTATE` message) and/or newly obtained emote sets, e.g. on firsst `USERSTATE` message after receiving a gifted sub.

Closes #2935
Closes #3007


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Major: Newly uploaded Twitch emotes are present in emote picker and can be "Tabbed" as well. (#2992)
```
```suggestion
- Major: Newly uploaded Twitch emotes are once again present in emote picker and can be "Tabbed" as well. (#2992)
```
I'm just going to suggest this changelog variation purely because as far as most people are concerned, this was just a bug, and they don't know what had to be done to fix it.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -543,6 +545,28 @@ void IrcMessageHandler::handleUserStateMessage(Communi::IrcMessage *message)
     }
 }
 
+// This will emit only once and right after user logs in to IRC - reset emote data and reload emtoes
```
pajlada, line 548:
```suggestion
// This will emit only once and right after user logs in to IRC - reset emote data and reload emotes
```

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -405,11 +377,19 @@ void TwitchAccount::loadUserstateEmotes()
...
                               });
+
                     emoteData->emoteSets.emplace_back(newUserEmoteSet);
                 }
             },
             [] {
                 // fetching emotes failed, ivr API might be down
+            },
+            [=] {
+                // XXX(zneix): We check if this is the last iteration and if so, call the callback
```
pajlada, line 388: This might be "accurate enough" but if we really want to call the callback once all requests are done, we'll need to do some sort of syncgroup outside of the requests (see https://en.cppreference.com/w/cpp/thread/latch)

pajlada (approved):

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -543,6 +545,24 @@ void IrcMessageHandler::handleUserStateMessage(Communi::IrcMessage *message)
...
+    // set received emote-sets, this time used to initially load emotes
+    // NOTE: this should always return true unless we reconnect
+    auto emoteSetsChanged = currentUser->setUserstateEmoteSets(
+        message->tag("emote-sets").toString().split(","));
+
+    // We should always attempt to reload emotes even on reconnections where
+    // emoteSetsChanged, since we want to trigger emote reloads when
+    // "currentUserChanged" signal is emitted
+    qCDebug(chatterinoTwitch) << emoteSetsChanged << message->toData();
+    currentUser->loadEmotes();
```
Mm2PL, line 563: If emote sets didn't change we don't want to reload?

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -519,6 +499,84 @@ void TwitchAccount::autoModDeny(const QString msgID, ChannelPtr channel)
...
+
+                std::sort(emoteSet->emotes.begin(), emoteSet->emotes.end(),
+                          [](const TwitchEmote &l, const TwitchEmote &r) {
+                              return l.name.string < r.name.string;
+                          });
+                emoteData->emoteSets.emplace_back(emoteSet);
+            }
+
+            // Getting userstate emotes from Ivr - NOIDONTTHINKSO
+            // this->loadUserstateEmotes();
```
Mm2PL, line 573: Commented out code here :) I think it's safe to remove?
```suggestion
```

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -60,6 +60,19 @@ void TwitchIrcServer::initializeConnection(IrcConnection *connection,
 
     qCDebug(chatterinoTwitch) << "logging in as" << account->getUserName();
 
+    // twitch.tv/tags enables IRCv3 tags on messages. See https://dev.twitch.tv/docs/irc/tags
+    // twitch.tv/commands enables a bunch of miscellaneous command capabilities. See https://dev.twitch.tv/docs/irc/commands
+    // twitch.tv/membership enables the JOIN/PART/MODE/NAMES commands. See https://dev.twitch.tv/docs/irc/membership
```
Mm2PL, line 65: `MODE` command is gone
```suggestion
    // twitch.tv/membership enables the JOIN/PART/NAMES commands. See https://dev.twitch.tv/docs/irc/membership
```

Mm2PL (changes requested):
Mm2PL suggested changes.

src/providers/twitch/IrcMessageHandler.cpp
zneix: `emoteSetsChanged` might be false in some very rare cases on very rapid reconnections, however given previous logic, where we triggered `currentUser->loadEmotes();` on `currentUserChanged` signal it'd make sense to attempt reloading emotes here too just in case.
Just to clarify - with old logic, even on those very rapid reconnections we were attempting to reload emotes anyway, as the user has changed from none to whoever you're logged in as.

Mm2PL (approved):

zneix:
Apparently my login from #2856 was wrong, I've fixed it in 9564b3b and it seems to split emote sets correctly now.

PR #2996, opened 2021-07-12
Add logging to experimental IRC

Mm2PL:
Pull request checklist:

# Description

This PR adds logging to IRC. I can't say i like the code but I don't know how to make it better. There is a couple things that are out of scope of this PR, mainly faulty message building:
  - private messages are logged as normal messages
  - some messages are logged as empty
  - ACTION messages aren't logged properly on the sending side


Mm2PL:
uhm @pajlada could you review this?

src/singletons/Logging.cpp
```diff
@@ -25,7 +26,7 @@ void Logging::addMessage(const QString &channelName, MessagePtr message)
     auto it = this->loggingChannels_.find(channelName);
     if (it == this->loggingChannels_.end())
     {
-        auto channel = new LoggingChannel(channelName);
+        auto channel = new LoggingChannel(channelName, platformName);
```
pajlada, line 29: This would be an issue if two channels have the same name but are on different platforms, instead of there being a map with the channelName as the key pointing to a LoggingChannel, it should be a `Map<Platform, Map<ChannelName, LoggingChannel>>`

pajlada (changes requested):

pajlada (approved):

PR #3015, opened 2021-07-15
Cleanup debug logging

zneix:
Pull request checklist:

# Description

- Removed "Drag segment:" log, used to be emitted every time you switched to an irc split
- Removed verbose websocketpp debug output for received frame headers


src/debug/Benchmark.cpp
```diff
@@ -12,12 +12,12 @@ BenchmarkGuard::BenchmarkGuard(const QString &_name)
 BenchmarkGuard::~BenchmarkGuard()
 {
     qCDebug(chatterinoBenchmark)
-        << this->name_ << float(timer_.nsecsElapsed()) / 1000000.0f << "ms";
+        << this->name_ << float(timer_.nsecsElapsed()) / 1000.0f << "ms";
```
pajlada, line 15: This seems incorrect to me, 1 millisecond is 1000000 nanoseconds, not 1000

src/debug/Benchmark.cpp
```diff
@@ -12,12 +12,12 @@ BenchmarkGuard::BenchmarkGuard(const QString &_name)
...
 }
 
 qreal BenchmarkGuard::getElapsedMs()
 {
-    return qreal(timer_.nsecsElapsed()) / 1000000.0;
+    return qreal(timer_.nsecsElapsed()) / 1000.0;
```
pajlada, line 20: Same as the previous change in `BenchmarkGuard`

pajlada (changes requested):

src/debug/Benchmark.cpp
pajlada: Without your change:
image
image

So I imagine your change would be incorrect, feel free to test it yourself too

apa420: I've also previously mixed these up but it's milli (10^3), micro (10^6) and then nano (10^9).

zneix: I must've gotten confused with microseconds. Sorry for this silly change.

pajlada (approved):

PR #3018, opened 2021-07-15
Replace std::string with QString in rapidjson and PubSub Client

zneix:
Pull request checklist:

# Description

Changes all `std::string`s in rj:: namespace and `PubsubClient` class to `QString`s along with some small fixes, like removing unused variable and using [QString-multiarg](https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-arg.md) where applicable.
Initial work for Pubsub overhaul that I have planned.


pajlada (approved):
Seems good to me - Assuming you've tested it you're free to merge

PR #3021, opened 2021-07-15
Cleaned up IrcMessageHandler

zneix:
Pull request checklist:

# Description

- Cleaned up `handleRoomstateMessaage` - removed `broadcaster-lang` tag since it's legacy
- Remove declaration of app, use getApp() instead
- Some early-outs and removals of unnecessary indentation layers
- Use `QString()`s instead of empty strings
- Remove `MODE` messaage handler and related code
- Fixed couple clazy warnings
   - [multi-arg](https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-arg.md)
   - [detaching temporary](https://github.com/KDE/clazy/blob/master/docs/checks/README-detaching-temporary.md)
- Removed invalid namespace comment

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -701,26 +701,6 @@ void IrcMessageHandler::handleUserNoticeMessage(Communi::IrcMessage *message,
     }
 }
 
-void IrcMessageHandler::handleModeMessage(Communi::IrcMessage *message)
-{
-    auto channel = getApp()->twitch.server->getChannelOrEmpty(
-        message->parameter(0).remove(0, 1));
-
-    if (channel->isEmpty())
-    {
-        return;
-    }
-
-    if (message->parameter(1) == "+o")
-    {
-        channel->modList.append(message->parameter(2));
-    }
-    else if (message->parameter(1) == "-o")
-    {
-        channel->modList.append(message->parameter(2));
-    }
-}
-
 std::vector<MessagePtr> IrcMessageHandler::parseNoticeMessage(
     Communi::IrcNoticeMessage *message)
 {
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -758,7 +758,7 @@ std::vector<MessagePtr> IrcMessageHandler::parseNoticeMessage(
 
         return builtMessages;
     }
-}  // namespace chatterino
+}
 
 void IrcMessageHandler::handleNoticeMessage(Communi::IrcNoticeMessage *message)
 {
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -167,7 +167,7 @@ static QMap<QString, QString> parseBadges(QString badgesString)
 {
     QMap<QString, QString> badges;
 
-    for (auto badgeData : badgesString.split(','))
+    for (const auto &badgeData : badgesString.split(','))
     {
         auto parts = badgeData.split('/');
         if (parts.length() != 2)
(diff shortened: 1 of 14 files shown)


src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -332,61 +332,57 @@ void IrcMessageHandler::addMessage(Communi::IrcMessage *_message,
...
+            roomModes.slowMode = it.value().toInt();
+        }
+        if (auto it = tags.find("r9k"); it != tags.end())
+        {
+            roomModes.r9k = it.value() == "1";
+        }
+        if (auto it = tags.find("followers-only"); it != tags.end())
+        {
+            roomModes.followerOnly = it.value().toInt();
+        }
```
jammehcow, line 381: Rather than doing multiple iterations on a single iterator it would be nicer to iterate once and use a switch to apply these modes. I know it's not something you changed but could fit in this PR.

Also my knowledge of C++ iterators might be wrong since I'm thinking of them like an `IEnumerable<T>` from .NET

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -706,28 +701,6 @@ void IrcMessageHandler::handleUserNoticeMessage(Communi::IrcMessage *message,
...
-    if (message->parameter(1) == "+o")
-    {
-        channel->modList.append(message->parameter(2));
-    }
-    else if (message->parameter(1) == "-o")
-    {
-        channel->modList.append(message->parameter(2));
-    }
-}
-
```
jammehcow, line 730: Does TMI not notify of +/-o anymore? Feels weird to take this out.

Mm2PL:
Sorry I misclicked. Mobile btw

src/providers/twitch/IrcMessageHandler.cpp
zneix: TMI does not emit `MODE` message at all.

zneix: Regardless, `channel->modList` was completely unused either way.

Mm2PL: It used to emit `MODE`s, that **was even documented**. :rocket:

zneix: Given @mm2pl's comment, I assume it's fine to completely remove this. Resolving this conversation.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -592,7 +585,7 @@ std::vector<MessagePtr> IrcMessageHandler::parseUserNoticeMessage(
     auto tags = message->tags();
     auto parameters = message->parameters();
 
-    QString msgType = tags.value("msg-id", "").toString();
+    QString msgType = tags.value("msg-id", QString()).toString();
```
pajlada, line 588:
```suggestion
    QString msgType = tags.value("msg-id").toString();
```

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -648,7 +642,7 @@ void IrcMessageHandler::handleUserNoticeMessage(Communi::IrcMessage *message,
     auto parameters = message->parameters();
 
     auto target = parameters[0];
-    QString msgType = tags.value("msg-id", "").toString();
+    QString msgType = tags.value("msg-id", QString()).toString();
```
pajlada, line 645:
```suggestion
    QString msgType = tags.value("msg-id").toString();
```

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -818,7 +789,7 @@ void IrcMessageHandler::handleNoticeMessage(Communi::IrcNoticeMessage *message)
             return;
         }
 
-        QString tags = message->tags().value("msg-id", "").toString();
+        QString tags = message->tags().value("msg-id", QString()).toString();
```
pajlada, line 792:
```suggestion
        QString tags = message->tags().value("msg-id").toString();
```

pajlada (changes requested):
smol changes then fine :ok_hand:

pajlada (approved):

src/providers/twitch/IrcMessageHandler.cpp
zneix: Good point, I'll handle this in another PR though if needed as it's something that was here before and I didn't change this logic.

Mm2PL (approved):

PR #3029, opened 2021-07-17
Added informative messages on issues related to recent-messages

### LINKED ISSUE #2870: Handle recent-messages request errors better
> indicating something with the history API is wrong wouldn't be a bad idea, it could possibly prevent issues like this

Fully agree, I just had a thought about adding at least a system message informing the user about status of the recent-messages request / response. Something similar to responses that you receive upon refreshing third party emotes.

_Originally posted by @zneix in https://github.com/Chatterino/chatterino2/issues/2868#issuecomment-855430414_

zneix:
Pull request checklist:

# Description

Added an generic error message if request to recent-messgaes fails, including status code of the HTTP request.

error message

Added an informative message telling user that recent-messages bot isn't currently joined in the channel, for which they requested historical messages, explaining that messages we received might be incomplete.

dank gap message

Closes #2870


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -801,13 +804,39 @@ void TwitchChannel::loadRecentMessages()
...
         })
+        .onError([weak](NetworkResult result) {
+            auto shared = weak.lock();
+            if (!shared)
+                return;
+
+            shared->addMessage(makeSystemMessage(
+                QString("Recent-messages API responded with an error %1 while "
+                        "retrieving messages. This isn't Chatterino's fault, "
+                        "but rather a temporary issue with the service itself.")
```
leon-richardt, line 837: I don't think we should play the blame game here and drop the phrase "This isn't Chatterino's fault"

pajlada: I think the message `Message history service unavailable (Code: %1)` would be descriptive enough, and give service owners & us enough information to debug

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -801,13 +804,39 @@ void TwitchChannel::loadRecentMessages()
...
+                {
+                    qCDebug(chatterinoTwitch)
+                        << QString("rm error_code=%1, channel=%2")
+                               .arg(errorCode, this->getName());
+                    if (errorCode == "channel_not_joined")
+                    {
+                        shared->addMessage(makeSystemMessage(
+                            "Recent-messages API isn't currently joined to "
+                            "this channel (in progress or due to an issue)."));
+                    }
```
pajlada, line 823: If the `errorCode` is `channel_not_joined` and `messages` is not empty we can deduce that the bot was joined previously but has disconnected, we can inform the user that there may be gaps in the message history: `Message history service recovering, there may be gaps in the message history`

If the `errorCode` is `channel_not_joined` and `messages` is empty I don't think we should print anything, because there's no difference for you vs for the next user who joins and sees nothing other than "Wow! You were the first user here with Chatterino! Congratulations! :rocket:"

zneix: Should be resolved in 3f1d81f1, let me know if it's good enough.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -808,18 +808,19 @@ void TwitchChannel::loadRecentMessages()
...
+                // Notify user about a possible gap in logs if it returned some messages
+                // but isn't currently joined to a channel
                 if (QString errorCode = root.value("error_code").toString();
-                    !errorCode.isEmpty())
+                    !errorCode.isEmpty() && messages.size() > 0)
                 {
                     qCDebug(chatterinoTwitch)
                         << QString("rm error_code=%1, channel=%2")
                                .arg(errorCode, this->getName());
                     if (errorCode == "channel_not_joined")
```
pajlada, line 819: Let's move the `messages.empty` check here (and make it use isEmpty or empty instead of size() > 0)

pajlada (approved):
:+1:

PR #3037, opened 2021-07-18
Fix PubSub client creation/pending topic resolving

### LINKED ISSUE #3014: PubSub creates only 1 connection, strictly limitting amount of topics Chatterino can listen to
As defined here, these are the limits of how many connections and listen requests can PubSub make:
https://github.com/Chatterino/chatterino2/blob/5ead70a314ebee8b93541e6cd6855ad48c0f42b9/src/providers/twitch/PubsubClient.hpp#L47-L48

While those do comply with [twitch docs](https://dev.twitch.tv/docs/pubsub#api-limits), we don't scale connections to 10 at max (`MAX_PUBSUB_CONNECTIONS` isn't used anywhere), making it sometimes impossible to listen to certain topics.
As a proof to my previous statement, I've added a some code ([git diff](https://haste.zneix.eu/raw/hypakuguhu)) to see how many times we actually try to listen to a PubSub topic:

debug popup

Note: 100 listen attempts instead of expected 50 might be caused due to following found by @Bun
thanks cbdg

pajlada:
Pull request checklist:

# Description

We had two issues that combined into one:
1) PubSub clients were created on masse on startup, the first 2 clients that are created were filled and then many more listens were requested, creating one client request each. We now handle this crudely with an atomic bool we check upon client creation request, and set when the connection is open.
2) Whenever a client is connected and we go through the topic listen request queue, we also check if the listen queue is empty. If it's not, we request a new client to be created.

Fixes #3014


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Minor: Received Twitch messages now use the exact same timestamp (obtained from Twitch's server) for every Chatterino user instead of assuming message timestamp on client's side. (#3021)
 - Minor: Received IRC messages use `time` message tag for timestamp if it's available. (#3021)
 - Bugfix: Fixed "smiley" emotes being unable to be "Tabbed" with autocompletion, introduced in v2.3.3. (#3010)
+- Bugfix: PubSub didn't properly try to resolve pending listens when the pending listens list was larger than 50. (#3037)
```
Felanbird, line 10:
```suggestion
- Bugfix: Fixed PubSub not properly trying to resolve pending listens when the pending listens list was larger than 50. (#3037)
```
The original variation makes sense, but it reads as "Bugfix: This was broken" and doesn't sound like this PR specifically fixed/fixes it.

zneix (approved):

PR #3060, opened 2021-07-22
Fixed comma appended to username completion when not at the beginning of the message

### LINKED ISSUE #2918: Mention users with a comma inconsistency
**Describe your issue**
With the changes implemented in f605221, mentioning a user at the end of a message adds a `,` to the mention, when it didn't do that before.

1. Start off your message with a `@`mention [tab completion]
2. Type something else [optional]
3. Finish your message with a `@`mention [tab completion]

The issue is specific to the actual menu, as disabling it causes the original logic to work again.

As an added note while testing this, it was noticed that right-clicking on a username to `@`mention them does not follow this original logic and the `,` appears at the end of messages, replicable before and after the menu was added.

**Screenshots**

https://i.imgur.com/l3FiISb.gif [ mm2pl/dankerino@59645ab equiv to 9fb5ef6 ]

https://i.imgur.com/CDFNPn8.gif [ 97b9bfb ]

**OS and Chatterino Version**
win10

zneix:
Pull request checklist:

# Description

Fixes issue described in #2918. As discussed with other people in chat, I've also went ahead and made right-clicking "username links" follow the same format of appending comma after the username only if it's the first word.

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -569,7 +569,9 @@ void SplitInput::insertCompletionText(const QString &input_)
         else if (text[i] == '@')
         {
             input = "@" + input_ +
-                    (getSettings()->mentionUsersWithComma ? ", " : " ");
+                    (edit.isFirstWord() && getSettings()->mentionUsersWithComma
+                         ? ", "
+                         : " ");
             done = true;
         }
 
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -597,6 +597,11 @@ void SplitInput::clearSelection()
     this->ui_.textEdit->setTextCursor(c);
 }
 
+bool SplitInput::isEditFirstWord() const
+{
+    return this->ui_.textEdit->isFirstWord();
+}
+
 QString SplitInput::getInputText() const
 {
     return this->ui_.textEdit->toPlainText();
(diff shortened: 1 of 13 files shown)


Felanbird (approved):
https://i.imgur.com/pDG4PPn.gif

src/widgets/splits/SplitInput.cpp
```diff
@@ -569,7 +569,9 @@ void SplitInput::insertCompletionText(const QString &input_)
         else if (text[i] == '@')
         {
             input = "@" + input_ +
-                    (getSettings()->mentionUsersWithComma ? ", " : " ");
+                    (edit.isFirstWord() && getSettings()->mentionUsersWithComma
```
talneoran, line 572: With how `isFirstWord` is implemented, is this not equivalent to checking if `i == 0` here? Maybe we want the comma to be appended in more situations though.

zneix: > With how isFirstWord is implemented, is this not equivalent to checking if i == 0 here?

No, in cases like: ` @zn<Tab>` `i == 1`.

> Maybe we want the comma to be appended in more situations though.

Could you expand? I can't think of in what situations we'd want to have comma appended.

Mm2PL (approved):

src/widgets/splits/SplitInput.cpp
zneix: As discussed in chat: 

I agree with this, maybe in the future we could upgrade / expand `isFirstWord()` into something that'd check if all words before the cursor are just mentions and append a comma.

Mm2PL (approved):

pajlada (approved):

PR #3065, opened 2021-07-24
Login process overhaul

zneix:
Pull request checklist:

- [x] Tested on Linux
- [ ] Tested on Windows
   - [ ] Firewall (?)
- [ ] Tested on MacOS

# Description

tl;dr login will be 100% client-side, no copypaste bullshit, everything will be automatic.

Made use of `QHttpServer` to create a HTTP server listening on `localhost:52107` (port opens only after user tries to authenticate).
The module itself is still not fully added to Qt itself (thus the blocked label), but it's very usable and there're rumors of adding it to Qt soon, so I'm guessing it's worth waiting. And if not, it's possible to migrate to boost:beast.
For ArchLinux friends who wanna test this, AUR package `qthttpserver-git` does the job.

### UX
Removed basic and advanced sections in favour of clean layout with description, warning and two buttons as before. "Paste token" button completely deprecates "Advanced login" because now we're able to fetch all other details from token.
Example:
zneixFrontend

### How does process of logging in work now?
1. Chatterino opens id.twitch.tv directly (constructed here https://github.com/Chatterino/chatterino2/pull/3065/files#diff-e181c6b227ade04c67338317f8c6988921b020dbf3f29402f2b08efac573fa73R28-R36)
2. We get redirected to `GET localhost:52107/redirect` - Chatterino's server is listening on this port and will return to the user an HTML page containing improved version of script from https://github.com/chatterino/login, which will make the user extract `access_token` from URL fragment and call `PUT localhost:52107/token`
3. On `PUT localhost:52107/token`, Chatterino checks if request contains `X-Access-Token` header containing access token we received from Twitch.
4. We validate the token (https://dev.twitch.tv/docs/authentication#validating-requests) and get all the details we need from it this way (without having to call Helix's Get Users).
5. After validation, we respond to the request that was earlier made to `PUT localhost:52107/token` with empty body but either 200 for success or 400 in any kind of failure.
6. If everything was successful, we update settings and change the current user (just like we did before).

The process for pasting token from clipboard is the same, except steps 1-3 and 5 are omitted.


fourtf:
Requiring opening an http server can cause problems with the users firewall.

xHeaveny:
work on windows but you have to verify to change firewall settings so for non-admin users not so nice

zneix:
> work on windows but you have to verify to change firewall settings so for non-admin users not so nice

How do you know this, did you build that on Windows? How did you do it, there's no package for `httpserver` on Windows as far as I am aware.

xHeaveny:
> > work on windows but you have to verify to change firewall settings so for non-admin users not so nice
> 
> How do you know this, did you build this PR on Windows? How did you do it, there's no package for `httpserver` on Windows as far as I am aware.

https://www.npmjs.com/package/http-server

Mm2PL:
That's a node module you linked, not the Qt `httpserver` module

zneix:
I think I'm gonna rewrite the implementation in boost:beast instead, so I'm putting this one into draft until I learn that framework.

fourtf:
Sorry but I this is not a good solution. Chatterino shouldn't open a tcp port like that. The same thing was done in Chatterino 1 and it creates too many issues.

jammehcow:
> Sorry but I this is not a good solution. Chatterino shouldn't open a tcp port like that. The same thing was done in Chatterino 1 and it creates too many issues.

This is common practice for most apps when doing auth. What concerns do you have about opening a port for it?
I could understand if the port was open all the time so ideally it would only be open when the user needs to add an account. Haven't read the code so not sure if that's the case.

zneix:
> I could understand if the port was open all the time so ideally it would only be open when the user needs to add an account.

I am aiming for listening on a port only on localhost interface and only for the short period of time - Chatterino would open login page and start listening, user would login, be redirected to the localhost:port, - Chatterino would send back either 200 or 400 and immediately stop listening.

Mm2PL:
I believe this to be the best solution available to us. There is no middle man that can get compromised or steal your tokens, we can refresh the tokens in the app. Yes it does come at a cost of having to temporarily open a localhost port on the firewall. Could there be another way of getting the token locally that wouldn't require a web server?

##### I'm excited for not having to deal with Login authentication expired. Login authentication expired. disconnected

zneix:
> Could there be another way of getting the token locally that wouldn't require a web server?

Actually, yes there is fallback way to do it. Since we don't change the authorization flow that's used to obtain the token, we can just open Twitch's authorization page (id.twitch.tv) and after user authorizes application, he'd be redirected to `localhost:port/redirect#token=xxxxxx`. Copy the token and then pressing new "Paste token from clipboard" button Chatterino will try to validate the pasted token and if the token is valid, all the required properties (user id, login, client id) will be obtained upon successful validation, successfully adding user's account to Chatterino.

I know this may be stupid and perhaps unintuitive, for - some users, but in case the port cannot be opened, - we'd detect the failure and inform user about it, instructing users on how to copy and paste their token - into Chatterino.
With proper instructions in place I believe this should be fine and even Swedish roleplayers should be able to understand it since they already somehow understand how to copy and paste from /client_login.

Issue #3076, opened 2021-07-27
Here is why you can't follow/unfollow from chatterino.


Mm2PL:
I'm placing this issue here to prevent duplicates. YES WE KNOW ABOUT THIS. we cannot do anything.

Twitch are officially removing this functionality with **no replacement**: https://discuss.dev.twitch.tv/t/deprecation-of-create-and-delete-follows-api-endpoints/32351 apparently to stop follow bots.

**OS and Chatterino Version**

ANY that ever existed.

Coldblackice:
Ahh big RIP ☹️

Mm2PL:
It's been well over a month since I opened this issue to prevent duplicates. It's about time to close it.

12932:
Implement through GQL?

LosFarmosCTL:
> Implement through GQL?

Using GQL is against Twitch Developer TOS.

12932:
> > Implement through GQL?
> 
> Using GQL is against Twitch Developer TOS.

Is that something Chatterino/Chatterino developers have agreed to? I certainly don't agree to it

LosFarmosCTL:
https://discuss.dev.twitch.tv/t/is-it-legal-to-use-https-gql-twitch-tv-gql-endpoint/21811

While there is no clear “You cannot use this GQL endpoint” somewhere on Twitch’s website, it is at least in conflict with some of their guidelines and could, if twitch wants to result in actions against your account or you as a person.

Use at your own risk, I have so far not seen Twitch take any specific actions against individuals using it, but definitely not something for chatterino to implement.

PR #3078, opened 2021-07-27
Deprecated /(un)follow commands and respective usercard action

zneix:
Pull request checklist:

# Description

As announced, `/helix/users/follows` endpoints have been decomissioned today.

farewell, beautiful Helix code

CC #3076 #2629


zneix:
For now, `/follow` and `/unfollow` have been marked as deprecated:
saj

And for the time being, follow checkbox in usercard remains but it's permanently disabled:
CoolCat

This maybe isn't the ideal solution as it could confuse users, making them think they can still (un)follow and try to click the checkbox.
We could either:
1. Remove the checkbox and checking if current user follows the user of whom usercard is opened altogether.
2. Think of some alternative, like the one that @mm2pl suggested:

<details>
<summary>Click to expand</summary>

GIMP masters

</details>

src/controllers/commands/CommandController.cpp
```diff
@@ -373,6 +373,22 @@ void CommandController::initialize(Settings &, Paths &paths)
             return "";
         });
 
+    this->registerCommand("/follow", [](const auto &words, auto channel) {
+        channel->addMessage(makeSystemMessage(
+            "Twitch has removed possibility to follow users through "
```
Felanbird, line 378:
```suggestion
            "Twitch has removed the ability to follow users through "
```

src/controllers/commands/CommandController.cpp
```diff
@@ -373,6 +373,22 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            "third-party applications. For more information, see "
+            "https://github.com/Chatterino/chatterino2/issues/3076"));
+        return "";
+    });
+
+    this->registerCommand("/unfollow", [](const auto &words, auto channel) {
+        channel->addMessage(makeSystemMessage(
+            "Twitch has removed possibility to unfollow users through "
```
Felanbird, line 386:
```suggestion
            "Twitch has removed the ability to unfollow users through "
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
+- Major: Deprecated `/(un)follow` commands and (un)following in the usercards as Twitch removed the possibility to (un)follow users.
```
```suggestion
- Major: Deprecated `/(un)follow` commands and (un)following in the usercards as Twitch has removed this feature for 3rd party applications. (#3076, #3078)
```

Felanbird (changes requested):
`Minor English loophole:`
Twitch has taken away our ability to manage follows, not the quality of, or possibility to make changes for the endpoint. 
https://wikidiff.com/ability/possibility

As for the changelog I'm indifferent in the usage of (un)follow 🤷

Felanbird:
As for the future of the button, I'm going to put my vote for keeping the button with its ability show follow status as currently handled, but making the functionality return an error message, as done with the slash commands in this PR.

Possibly removing it in the future if we need button room for some reason.

zneix:
> but making the functionality return an error message, as done with the slash commands in this PR.

As I've stated in my comment under this PR:

> follow checkbox in usercard remains but it's permanently disabled

Do you mean we should keep the functionality of the button but add error messages notifying users that (un)following failed or did you just overlook that part where I said it? I believe there should be no logic for event actions related to that button as it's gonna be literally cosmetic at this point (if we decide to keep it in the first place that is).

Felanbird:
> Do you mean we should keep the functionality of the button but add error messages notifying users that (un)following failed 

Yes, I think if we keep the button it should at least do something indicating to users that the attempt failed.




> I believe there should be no logic for event actions related to that button as it's gonna be literally cosmetic at this point  (if we decide to keep it in the first place that is).

If we to remove the all logic we should just remove the button in its entirety, no need for a cosmetic button.

pajlada (approved):

Felanbird:
Works as intended, the only problem seems to be the loading time, you have hover the button for 1.5~ seconds before the prompt appears. It took me a while to actually get the prompt because I was clicking the "button" and and doing so resets the loading.

image

pajlada:
> Works as intended, the only problem seems to be the loading time, you have hover the button for 1.5~ seconds before the prompt appears. It took me a while to actually get the prompt because I was clicking the "button" and and doing so resets the loading.
> 
> image

That's unfortunately fairly tricky for us to solve without creating a new QCheckBox type and re-implementing it's click handler making it show the tooltip.

Mm2PL:
About the usercard: maybe we could change the text color/effects. For example gray out the `Follow` text or cross it out.

pajlada:
All in favour of removing the checkbox?

Felanbird (approved):

pajlada (approved):

Mm2PL (approved):
lgtm

PR #3080, opened 2021-07-28
Don't copy leading or trailing spaces

### LINKED ISSUE #2449: Copying selection adds an extra space at the end
**Describe the bug**
Copying a selection of text with control + c adds a space at the end of the selection

**To reproduce**
Copy any text 

**Chatterino version**
2.2.2-fix

**Operating system**
Windows 10

**Additional information**
Not sure if intended or bug, but I find it annoying

Mm2PL:
Pull request checklist:

# Description

Closes #2449


Mm2PL:
- [ ] ^C

pajlada (changes requested):
We're looking at a different solution offline where the trailing space isn't added in the first place

fourtf:
this is pretty weird if the user selects one word and the space after it, right? or am I missing something

Mm2PL:
Yeah it doesn't work if the user select a space at the start or end manually. I overlooked that. The other solution pajlada is talking about is not adding a trailing space in layout elements.

pajlada:
> this is pretty weird if the user selects one word and the space after it, right? or am I missing something

The solution we looked at instead was, at the end of bulding a message, just fetch the last entry of the message layout container and set its "hasTrailingSpace" to false

Mm2PL:
This is an old PR that i did not finish, closing.

PR #3081, opened 2021-07-28
Use double spaces instead of Chatterino character when possible

Mm2PL:
Pull request checklist:

# Description

No issue. Can help eliminate issues with users noticing the Chatterino character. Double spaces are way less noticeable (Chatterino, Chatty and webchat don't even render those!) than a space and U+E0000.


zneix:
> when possible

Am I missing soemthing or are you straight-up removing our old solution instead?

Mm2PL:
If there is no space, there is nothing to replace with a double space.

pajlada (approved):

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -375,7 +375,16 @@ void TwitchChannel::sendMessage(const QString &message)
...
             {
-                parsedMessage.append(MAGIC_MESSAGE_SUFFIX);
+                auto spaceIndex = parsedMessage.indexOf(' ');
+                if (spaceIndex == -1)
+                {
+                    parsedMessage.append(MAGIC_MESSAGE_SUFFIX);
+                }
+                else
+                {
+                    // replace only first space with a double space
```
Mm2PL, line 385:
```suggestion
                    // replace the space we found in spaceIndex with two spaces
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -375,7 +375,16 @@ void TwitchChannel::sendMessage(const QString &message)
         {
             if (parsedMessage == this->lastSentMessage_)
             {
-                parsedMessage.append(MAGIC_MESSAGE_SUFFIX);
+                auto spaceIndex = parsedMessage.indexOf(' ');
+                if (spaceIndex == -1)
+                {
+                    parsedMessage.append(MAGIC_MESSAGE_SUFFIX);
```
Mm2PL, line 381:
```suggestion
                    // no spaces found, fall back to old magic character
                    parsedMessage.append(MAGIC_MESSAGE_SUFFIX);
```

zneix (approved):

nilyt:
I didn't even know this was a feature, until I saw this change.
Always considered this behaviour a bug I wanted to fix eventually, as it happens seemingly randomly, breaks building SE pyramids, and was just a nuisance all around (to various people).

Where is this setting (`getSettings()->allowDuplicateMessages`) in the UI, so I can turn it off? Or is it not exposed in the UI (for reasons)?
I'd vote for it to be off by default for the mentioned reasons. At least I can't recall enabling something like that?
Or maybe I did? The name "allowDuplicateMessages" is something I'd associate with 'r9k' more than with what it does.

Mm2PL:
> Where is this setting (getSettings()->allowDuplicateMessages) in the UI, so I can turn it off?

here

PR #3095, opened 2021-07-30
updated BUILDING_ON_LINUX to include chromebook

killertofus:
Pull request checklist:

# Description
as they use a Debian 10 environment  https://support.google.com/chromebook/answer/9145439 was wondering if we should have a  BUILDING_ON_CHROMEBOOK section or keep it here

### DIFF
--- a/BUILDING_ON_LINUX.md
+++ b/BUILDING_ON_LINUX.md
@@ -81,3 +81,11 @@ _`gstreamer-plugins-good` package is retired in Fedora 31, see: [rhbz#1735324](h
 ### Using CMake
 
 1. `cmake .. && make`
+
+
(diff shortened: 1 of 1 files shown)


zneix:
If it uses Debian environment then chromebook users should just  Debian installation instructions. No need for a chunky codeblock telling you just that.
Also let's be real - if anyone using a chromebook would want to develop Chatterino (which I highly doubt would even happen), they'd immidiatelly notice what is the environment of the terminal they're using. It's already noted that Debian ~= Ubuntu right under https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_LINUX.md#ubuntu-1804 so this pull request is pointless.

PR #3102, opened 2021-08-01
Check for ignored phrases/users in channel point redemptions

### LINKED ISSUE #3098: Blocked users' redemptions are visible in chat
If a user is blocked (via Twitch) and uses a redemption without text, the redemption is still visible in chat. For reference, twitch web chat has these redemptions hidden.

This only applies to redemptions without text - if I recall correctly, those are not sent via IRC, unlike those with text. As far as I'm concerned, this behavior is up to clients to implement as the redemptions are sent regardless of blocks.

pajlada:
Pull request checklist:

# Description

Dev note:
Because a channel point message is contextless (i.e. has no channel attached), we have to attach the relevant parameters to the function call itself. The alternative to this would be to make the function non-static, but we'd have to rework TwitchMessageBuilder and SharedMessageBuilder to work with a null ircmessage which is not something I want to do.

Fixes #3098

### DIFF
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -159,6 +159,7 @@ SOURCES += \
     src/controllers/highlights/HighlightModel.cpp \
     src/controllers/highlights/HighlightPhrase.cpp \
     src/controllers/highlights/UserHighlightModel.cpp \
+    src/controllers/ignores/IgnoreController.cpp \
     src/controllers/ignores/IgnoreModel.cpp \
     src/controllers/moderationactions/ModerationAction.cpp \
     src/controllers/moderationactions/ModerationActionModel.cpp \
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -2,6 +2,7 @@
 
 #include "Application.hpp"
 #include "common/QLogging.hpp"
+#include "controllers/ignores/IgnoreController.hpp"
 #include "controllers/ignores/IgnorePhrase.hpp"
 #include "messages/Message.hpp"
 #include "messages/MessageElement.hpp"
@@ -104,20 +105,9 @@ void SharedMessageBuilder::parse()
 
 bool SharedMessageBuilder::isIgnored() const
 {
-    // TODO(pajlada): Do we need to check if the phrase is valid first?
-    auto phrases = getCSettings().ignoredMessages.readOnly();
-    for (const auto &phrase : *phrases)
-    {
-        if (phrase.isBlock() && phrase.isMatch(this->originalMessage_))
-        {
-            qCDebug(chatterinoMessage)
-                << "Blocking message because it contains ignored phrase"
-                << phrase.getPattern();
-            return true;
-        }
-    }
-
-    return false;
+    return isIgnoredMessage({
+        .message = this->originalMessage_,
+    });
 }
 
 void SharedMessageBuilder::parseUsernameColor()
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -106,7 +106,7 @@ void SharedMessageBuilder::parse()
 bool SharedMessageBuilder::isIgnored() const
 {
     return isIgnoredMessage({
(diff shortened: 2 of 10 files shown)


chatterino.pro
```diff
@@ -159,6 +159,7 @@ SOURCES += \
     src/controllers/highlights/HighlightModel.cpp \
     src/controllers/highlights/HighlightPhrase.cpp \
     src/controllers/highlights/UserHighlightModel.cpp \
+    src/controllers/ignores/IgnoreController.cpp \
```
zneix, line 162: Did you forget to include `src/controllers/ignores/IgnoreController.hpp`? Try running `./tools/update-filelist.py`

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -114,44 +114,12 @@ TwitchMessageBuilder::TwitchMessageBuilder(
...
-            return true;
-        }
-    }
-
-    return false;
+    return isIgnoredMessage({
+        /*.message = */ this->originalMessage_,
+        /*.twitchUserID = */ this->tags.value("user-id").toString(),
+        /*.isMod = */ this->channel->isMod(),
+        /*.isBroadcaster = */ this->channel->isBroadcaster(),
```
zneix, line 121: These comments were a part of the old solution, right? Can those be removed now?

src/messages/SharedMessageBuilder.cpp
```diff
@@ -104,20 +105,9 @@ void SharedMessageBuilder::parse()
...
-            qCDebug(chatterinoMessage)
-                << "Blocking message because it contains ignored phrase"
-                << phrase.getPattern();
-            return true;
-        }
-    }
-
-    return false;
+    return isIgnoredMessage({
+        /*.message = */ this->originalMessage_,
```
zneix, line 109: This comment was a part of the old solution, right? Can those be removed now?

Also, I see that here we only pass the message, while in the other calls of `isIgnoredMessage` we pass all the things. I believe it's ok and we check for whether more members of `IgnoredMessageParameters` are present, right? Just willing to understand this logic better.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1261,8 +1231,19 @@ Outcome TwitchMessageBuilder::tryParseCheermote(const QString &string)
...
 void TwitchMessageBuilder::appendChannelPointRewardMessage(
-    const ChannelPointReward &reward, MessageBuilder *builder)
+    const ChannelPointReward &reward, MessageBuilder *builder, bool isMod,
+    bool isBroadcaster)
 {
+    if (isIgnoredMessage({
+            /*.message = */ "",
+            /*.twitchUserID = */ reward.user.id,
+            /*.isMod = */ isMod,
+            /*.isBroadcaster = */ isBroadcaster,
```
zneix, line 1241: These comments were a part of the old solution, right? Can those be removed now?

zneix (changes requested):

src/providers/twitch/TwitchMessageBuilder.cpp
pajlada: The comments still serve the purpose to saying which parameter is what, format might be a bit weird since it's part of the old solution but I think it is still pretty enough

chatterino.pro
pajlada: Row 391 has the header file - the header file was not new

src/messages/SharedMessageBuilder.cpp
pajlada: Comments should not be removed, no

We only pass as many parameters as we need, with the exception of forcing us to include message as an empty string in other places because c++20 struct initialization isn't possible yet. With c++20 struct initialization we can omit .message in the other places too.

src/providers/twitch/TwitchMessageBuilder.cpp
pajlada: no :t-rex:

zneix (approved):

PR #3103, opened 2021-08-01
Prepare CMake for Qt6

zneix:
Pull request checklist:

# Description

Renamed variables of QT5KEYCHAIN to just QTKEYCHAIN due to how qt keychain library changes its target name from `qt5keychain` to `qt6keychain` if we build against Qt6.

⚠ Important note: qtkeychain already uses `BUILD_WITH_QT6` CMake parameter, so I thought it would be convenient to just reuse that. No idea if that's a bad practice to use one variable/option that affects multiple things at once, but pretty sure that in this scope it wouldn't change anything.

Required for Chatterino/libcommuni#2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,10 +13,17 @@ option(BUILD_APP "Build Chatterino" ON)
 option(BUILD_TESTS "Build the tests for Chatterino" OFF)
 option(USE_SYSTEM_PAJLADA_SETTINGS "Use system pajlada settings library" OFF)
 option(USE_SYSTEM_LIBCOMMUNI "Use system communi library" OFF)
-option(USE_SYSTEM_QT5KEYCHAIN "Use system Qt5Keychain library" OFF)
+option(USE_SYSTEM_QTKEYCHAIN "Use system QtKeychain library" OFF)
 option(USE_PRECOMPILED_HEADERS "Use precompiled headers" ON)
+option(BUILD_WITH_QT6 "Use Qt6 instead of default Qt5" OFF)
 
 option(USE_CONAN "Use conan" OFF)
+
+if (BUILD_WITH_QT6)
+    set(MAJOR_QT_VERSION "6")
+else()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -13,11 +13,18 @@ option(BUILD_APP "Build Chatterino" ON)
...
+option(USE_SYSTEM_QTKEYCHAIN "Use system QtKeychain library" OFF)
 option(USE_PRECOMPILED_HEADERS "Use precompiled headers" ON)
+option(BUILD_WITH_QT6 "Use Qt6 instead of default Qt5" OFF)
 
 option(USE_CONAN "Use conan" OFF)
 
+if (BUILD_WITH_QT6)
+    set(MAJOR_QT_VERSION "6")
+else()
+    set(MAJOR_QT_VERSION "5")
```
zneix, line 25: Should this be strings or numbers? No idea how CMake treats things like this

AnotherFoxGuy: In cmake everything is a string, it doesn't have any other types

CMakeLists.txt
```diff
@@ -13,11 +13,18 @@ option(BUILD_APP "Build Chatterino" ON)
 option(BUILD_TESTS "Build the tests for Chatterino" OFF)
 option(USE_SYSTEM_PAJLADA_SETTINGS "Use system pajlada settings library" OFF)
 option(USE_SYSTEM_LIBCOMMUNI "Use system communi library" OFF)
-option(USE_SYSTEM_QT5KEYCHAIN "Use system Qt5Keychain library" OFF)
+option(USE_SYSTEM_QTKEYCHAIN "Use system QtKeychain library" OFF)
```
pajlada, line 16: Make sure you add a dev changelog entry for this

zneix: Hopefully 7f8e05b is a good changelog entry for this one.

pajlada (approved):
:+1:

PR #2210, opened 2020-11-19
Updated minimum required Qt versions

zneix:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

- Updated minimum required and latest Qt versions
- Added Qt version check to chatterino.pro

As discused on Chatterino Discord, we now have methods in our code such as [QOalette#placeholderText](https://doc.qt.io/qt-5/qpalette.html#placeholderText) in https://github.com/Chatterino/chatterino2/blob/master/src/widgets/splits/Split.cpp#L321-L322 which were introduced in Qt 5.12.0, updated documentation to reflect those changes.

PR #3105, opened 2021-08-02
Allow building against Qt 5.11

nilyt:
Pull request checklist:

# Description

Current stable debian Linux is Buster, which still ships Qt 5.11.3, so I've had this patch to build chatterino locally, for the last several months.
Since `QPalette::placeholderText` seems to be the only codebase feature depending on QT 5.12, the build has been running perfectly fine.
The Qt docs say
> Before Qt 5.12, the placeholder text color was hard-coded in the code as QPalette::text().color() where an alpha of 128 was applied.

and based on my testing, theme changes look just fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Bugfix: Fixed comma appended to username completion when not at the beginning of the message. (#3060)
 - Bugfix: Fixed bug misplacing chat when zooming on Chrome with Chatterino Native Host extension (#1936)
 - Bugfix: Channel point redemptions from ignored users are now properly blocked. (#3102)
+- Dev: Allow building against Qt 5.11 (#3105)
 - Dev: Ubuntu packages are now available (#2936)
 - Dev: Disabled update checker on Flatpak. (#3051)
 - Dev: Add logging for HTTP requests (#2991)
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -14,7 +14,7 @@ CCACHE_BIN = $$system(which ccache)
   CONFIG+=ccache
 }
 
-MINIMUM_REQUIRED_QT_VERSION = 5.12.0
+MINIMUM_REQUIRED_QT_VERSION = 5.11.0
 
 !versionAtLeast(QT_VERSION, $$MINIMUM_REQUIRED_QT_VERSION) {
     error("You're trying to compile with Qt $$QT_VERSION, but minimum required Qt version is $$MINIMUM_REQUIRED_QT_VERSION")
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -134,14 +134,18 @@ void SplitInput::themeChangedEvent()
     QPalette palette, placeholderPalette;
 
     palette.setColor(QPalette::WindowText, this->theme->splits.input.text);
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 12, 0))
     placeholderPalette.setColor(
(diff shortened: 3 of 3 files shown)


Mm2PL:
I'm pretty sure we dropped support for qt 5.11 #3093, re-adding it would probably mean reverting other commits. We don't have 5.11 CI so we don't know. As simple as this PR is there might be other things holding us back from doing this like Windows for example.

nilyt:
Well, allowing the codebase to build against Qt 5.11 does not have to equal doing CI runs for it and officially supporting builds on it. I should rename the MR there.

The code cleanup in #3093 seems to have be an old workaround for Qt 5.11.1 on Windows specifically, and the windows build environment is different with no system-wide Qt version, so it's easier to upgrade lib dependencies if that makes sense?
(Though, personally, I'm not a fan of forced compatibility stripping like that, just for the sake of it.)

I understand how you might not want to introduce new code to support older libraries; though qt 5.11 is not quite dead yet in the Linux world, e.g. with debian or ubuntu (or other) LTS releases.

ALazyMeme:
Refer to https://github.com/Chatterino/chatterino2/pull/2210

> ...we now have methods in our code such as QOalette#placeholderText in https://github.com/Chatterino/chatterino2/blob/master/src/widgets/splits/Split.cpp#L321-L322 which were introduced in Qt 5.12.0

zneix:
tl;dr We don't want to support ancient versions of Qt, here's why:

1. We're going to drop support for Qt 5.12 soon anyway and move on to 5.15 LTS being the latest minimum supported version, give or take in 1-2 month(s). So in my opinion there's no purpose for this patch. I understand it might be hard to build / contribute for some on distros with literally ancient packages, but that shouldn't stop up from adding new code.
2. There's also many and many more deprecation messages that want us to use newer code that has been introduced after 5.12 (e.g. https://doc.qt.io/qt-5/qstring-obsolete.html#SplitBehavior-enum, replacement has been introduced in 5.14). The only reason we didn't move on yet is that we want to provide support for 5.12 for the time being, but as mentioned in 1. that's gonna change soon™.
3. I'm in the process of preparing Chatterino to use Qt6 and based on my migration attempts [qt6 branch](https://github.com/Chatterino/chatterino2/tree/zneix/feature%2Fqt6) a lot of old code has been just straight up removed. I'm too lazy to give examples, but look at the diff there - there's at least 2 or 3 methods that existed in pre-5.15, but are now removed in Qt6. That means we're not gonna bother with spaghettifying code just to try to support old code that is gonna be dropped eventually. Updating code should take a priority over supporting ancient legacy.

I understand how this makes it harder to build / contribute for some on distros which ship literally ancient packages, but imo that shouldn't stop up from adding new code. Given that 5.12 is literally the oldest LTS that is supported and debian doesn't even ship it. I don't want to sound rude towards you, it's not my intention, but just compiling Qt or somehow getting a newer version from Debian backports might be your only way out of this.

Before closing, I'll let pajlada and/or fourtf decide the fate of this one.

pajlada:
I generally don't mind supporting older versions of Qt where possible, at least supporting the compilation of it but the user would be on their own with any bug reports related to Qt.

However, since we're in the process of migrating to qt6 where a lot more than just this function would be broken for qt5.11, I'm hesistant towards any long term support.

I'm fine to merge this in as is, but I won't promise to keep it in the codebase for long. Once qt6 rolls in it's probably gone.

With Debian11 coming out soon with qt5.15 support, your problems will hopefully be solvable by just upgrading your OS instead.

pajlada (approved):

PR #3110, opened 2021-08-03
Fix filter crash introduced in #3092

### LINKED ISSUE #3108: Crash on startup when filter attached
**Describe your issue**
Chatterino crash on startup when we have channel filter attached. 

https://i.nuuls.com/P4Mgx.mp4

Steps to reproduce

1. Add filter to a channel
2. Close chatterino
3. Open chatterino again

Commit 3cb1e5158ab360776a98954aff9e79b2b2f5d6d7 works without crash.

**OS and Chatterino Version**
Chatterino Nightly 2.3.3 (commit 784fdd28b, )
Windows 10 21H1

pajlada:
Pull request checklist:

# Description

Reason for crash was a deadlock; we were already inside the channel lock while trying to lock it again to find the channel.
Instead of finding the channel again, we pass the channel all the way down to the filter.

Dev: Removed `FilterRecord::filter(message)` function as it was unused and would've required more changes.

Fixes crash introduced in #3092 
Fix #3108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Minor: Added informative messages for recent-messages API's errors. (#3029)
 - Minor: Added section with helpful Chatterino-related links to the About page. (#3068)
 - Minor: Now uses spaces instead of magic Unicode character for sending duplicate messages (#3081)
-- Minor: Added `channel.live` filter variable (#3092)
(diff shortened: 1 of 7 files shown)


Felanbird (approved):
Crash is fixed. 👍

Issue #3107, opened 2021-08-02
Chatterino keeps disconnecting if in more than 20 channels


zneix:
## Why does Chatterino keep disconnecting?

tl;dr for nerds: Chatterino exceeds 20 JOIN / 10s ratelimit, causing Twitch to terminate connection instantly.

As stated in [Twitch's documentation](https://dev.twitch.tv/docs/irc/guide#rate-limits), attempting to JOIN more than 20 channels per 10 seconds is not allowed. That results in Twitch immidiatelly shutting your connection down.

What happens, Chatterino tries to JOIN all the channels at once without any throttling.
If you have more than 20 channels, Twitch will disconnect Chatterino. Chatterino will try to reconnect, repeat indefinietly.

## What can I do for now?

Until a fix is available you have a few choices:
1. Temporarily remove channels to 20 or less.
2. Use DankChat with less chats or Chatsen.
3. Download a portable version of Chatterino and don't add more than 20 channels to it.

If you're going to go with 1. - before doing so, better copy your window layout, so you may restore it later. It's a file in [Chatterino's folder](https://wiki.chatterino.com/Help/#where-is-my-chatterino-folder-located) in `Settings/window-layout.json`. Copy it somewhere first, then close all channels until you reach 20 or less.
When a fixed version of Chatterino releases, simply replace the new `window-layout.json` with the one you backed up to restore all closed tabs.

Sorry for the sudden issue, I'll try to work on it ASAP. Thanks for understanding.

zneix:
Side note: This rate-limit has been documented on Twitch's documentation I linked for quite a long time now (at least a year or so, if not way longer), we just never followed that and the rate-limit was entirely cosmetic. Seems like an hour or so ago Twitch actually deployed a fix that enforces that rate-limit into effect.

AnatoleAM:
@zneix 
![image](https://user-images.githubusercontent.com/9599344/127935923-911db504-1d2a-4e69-b7d4-1cd8ca6d6be5.png)

Void48:
![unnamed-removebg-preview-2x_Loyaldk-SuperPony_370000 V1 0](https://user-images.githubusercontent.com/85196642/127937198-d12ee83b-264d-4adc-8fb9-fc0ea31ffcbc.png)

jammehcow:
[...]

PR #3112, opened 2021-08-03
Remove JOINs from write connection

pajlada:
- Remove JOINs on write connection
- Add changelog entry

Pull request checklist:

# Description

By not joining on the WRITE connection, we:
a) save some bandwidth
b) reduce the amount of JOINs we do, allowing users to have more initial splits open for now until we add proper rate limiting

Works toward fixing #3107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
 - Major: Added the ability to add nicknames for users. (#137, #2981)
+- Major: Work on rate-limiting JOINs and PARTs. (#3112)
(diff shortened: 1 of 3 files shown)


zneix:
Posting here for context too:
As discussed in chat (and based on my testing back from when I worked on implementing `GLOBALUSERSTATE`) I believe that only things that could get affected is messages handled in `TwitchIrcServer::writeConnectionMessageReceived`, so we should be good to have this merged.

I'm going to test this first though - applied patch from this PR in my "production" Chatterino and seeing what happens.

Mm2PL (changes requested):
Why do we still `PART` from the write connection even without joining?

I would comment on line 218 but I can't.

Mm2PL (approved):

zneix (approved):
Overall I'm good with the changes :+1: tested that myself and the only real difference is that we receive less `USERSTATE` messages (which is obviously to be expected, but doesn't affect us at all).

PR #3115, opened 2021-08-04
Rate limit outgoing JOIN messages

### LINKED ISSUE #3107: Chatterino keeps disconnecting if in more than 20 channels

zneix:
Pull request checklist:

# Description

Added experimental rate-limitting for outgoing JOIN messages with some imitation of Leaky bucket rate limitting. Huge thanks to @leon-richardt for helping me figure out the theory.
Tested it out and Twitch does not kill our connection - we conform to Twitch's 20 JOIN messages / 10s limit by sending at most 18 JOIN messages per 10.5s.

This might not be the best solution and I believe there could be a better one, but after about 24 hours of thinking this is the most sane solution I could come up with and given how important this bug is I think it's better than nothing.

~~There's still debug prints in the code, to make it easier for reviewers to test. After approving the PR I'll remove them.~~ already removed.

Closes #3107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
 - Major: Added the ability to add nicknames for users. (#137, #2981)
-- Major: Fixed constant disconnections with more than 20 channels by rate-limitting outgoing JOIN messages. (#3112, #3115)
+- Major: Fixed constant disconnections with more than 20 channels by rate-limiting outgoing JOIN messages. (#3112, #3115)
 - Minor: Added autocompletion in /whispers for Twitch emotes, Global Bttv/Ffz emotes and emojis. (#2999, #3033)
 - Minor: Received Twitch messages now use the exact same timestamp (obtained from Twitch's server) for every Chatterino user instead of assuming message timestamp on client's side. (#3021)
 - Minor: Received IRC messages use `time` message tag for timestamp if it's available. (#3021)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
 - Major: Added the ability to add nicknames for users. (#137, #2981)
-- Major: Work on rate-limiting JOINs and PARTs. (#3112)
+- Major: Fixed constant disconnections with more than 20 channels by rate-limitting outgoing JOIN messages. (#3112, #3115)
 - Minor: Added autocompletion in /whispers for Twitch emotes, Global Bttv/Ffz emotes and emojis. (#2999, #3033)
 - Minor: Received Twitch messages now use the exact same timestamp (obtained from Twitch's server) for every Chatterino user instead of assuming message timestamp on client's side. (#3021)
 - Minor: Received IRC messages use `time` message tag for timestamp if it's available. (#3021)
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -252,6 +252,7 @@ SOURCES += \
     src/util/LayoutHelper.cpp \
     src/util/NuulsUploader.cpp \
     src/util/RapidjsonHelpers.cpp \
+    src/util/RatelimitBucket.cpp \
     src/util/SplitCommand.cpp \
     src/util/StreamerMode.cpp \
     src/util/StreamLink.cpp \
@@ -509,6 +510,7 @@ HEADERS += \
     src/util/rangealgorithm.hpp \
     src/util/RapidjsonHelpers.hpp \
     src/util/RapidJsonSerializeQString.hpp \
+    src/util/RatelimitBucket.hpp \
     src/util/RemoveScrollAreaBackground.hpp \
     src/util/SampleCheerMessages.hpp \
     src/util/SampleLinks.hpp \
@@ -583,6 +585,7 @@ HEADERS += \
     src/widgets/settingspages/IgnoresPage.hpp \
     src/widgets/settingspages/KeyboardSettingsPage.hpp \
     src/widgets/settingspages/ModerationPage.hpp \
+    src/widgets/settingspages/NicknamesPage.hpp \
     src/widgets/settingspages/NotificationPage.hpp \
     src/widgets/settingspages/SettingsPage.hpp \
     src/widgets/splits/ClosedSplits.hpp \
--- a/src/providers/irc/AbstractIrcServer.cpp
+++ b/src/providers/irc/AbstractIrcServer.cpp
@@ -24,13 +24,11 @@ AbstractIrcServer::AbstractIrcServer()
         QCoreApplication::instance()->thread());
 
     // Apply a leaky bucket rate limiting to JOIN messages
-    this->joinBucket_.reset(new RatelimitBucket(
-        18, 10500,
-        [&](QString message) {
-            qCDebug(chatterinoIrc) << "joining" << message;
-            this->readConnection_->sendRaw("JOIN #" + message);
-        },
-        this));
+    auto actuallyJoin = [&](QString message) {
+        qCDebug(chatterinoIrc) << "joining" << message;
+        this->readConnection_->sendRaw("JOIN #" + message);
+    };
+    this->joinBucket_.reset(new RatelimitBucket(18, 10500, actuallyJoin, this));
 
     QObject::connect(this->writeConnection_.get(),
                      &Communi::IrcConnection::messageReceived, this,
--- a/src/providers/irc/AbstractIrcServer.cpp
+++ b/src/providers/irc/AbstractIrcServer.cpp
@@ -15,6 +15,10 @@ const int RECONNECT_BASE_INTERVAL = 2000;
 // 60 falloff counter means it will try to reconnect at most every 60*2 seconds
 const int MAX_FALLOFF_COUNTER = 60;
 
+// Ratelimits for joinBucket_
+const int JOIN_RATELIMIT_BUDGET = 18;
+const int JOIN_RATELIMIT_COOLDOWN = 10500;
+
 AbstractIrcServer::AbstractIrcServer()
 {
     // Initialize the connections
@@ -34,7 +38,8 @@ AbstractIrcServer::AbstractIrcServer()
         }
         this->readConnection_->sendRaw("JOIN #" + message);
     };
-    this->joinBucket_.reset(new RatelimitBucket(18, 10500, actuallyJoin, this));
+    this->joinBucket_.reset(new RatelimitBucket(
+        JOIN_RATELIMIT_BUDGET, JOIN_RATELIMIT_COOLDOWN, actuallyJoin, this));
 
     QObject::connect(this->writeConnection_.get(),
                      &Communi::IrcConnection::messageReceived, this,
--- a/src/providers/irc/AbstractIrcServer.cpp
+++ b/src/providers/irc/AbstractIrcServer.cpp
@@ -31,9 +31,6 @@ AbstractIrcServer::AbstractIrcServer()
     auto actuallyJoin = [&](QString message) {
         if (!this->channels.contains(message))
         {
-            qCDebug(chatterinoIrc)
-                << QString("attempted to join %1 but it does not exist anymore")
-                       .arg(message);
             return;
         }
         this->readConnection_->sendRaw("JOIN #" + message);
@@ -302,7 +299,6 @@ void AbstractIrcServer::onReadConnected(IrcConnection *connection)
     {
         if (auto channel = weak.lock())
         {
-            qCDebug(chatterinoIrc) << "queued joining" << channel->getName();
             this->joinBucket_->send(channel->getName());
         }
     }
--- a/src/providers/irc/AbstractIrcServer.cpp
+++ b/src/providers/irc/AbstractIrcServer.cpp
@@ -24,13 +24,13 @@ AbstractIrcServer::AbstractIrcServer()
         QCoreApplication::instance()->thread());
 
     // Apply a leaky bucket rate limitting to JOIN mesasges
-    this->bucket_ = std::make_unique<RatelimitBucket>(
+    this->bucket_.reset(new RatelimitBucket(
         18, 10500,
         [&](QString message) {
             qCDebug(chatterinoIrc) << "joining" << message;
             this->readConnection_->sendRaw("JOIN #" + message);
         },
-        this);
+        this));
 
     QObject::connect(this->writeConnection_.get(),
                      &Communi::IrcConnection::messageReceived, this,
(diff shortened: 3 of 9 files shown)


chatterino.pro
```diff
@@ -583,6 +585,7 @@ HEADERS += \
     src/widgets/settingspages/IgnoresPage.hpp \
     src/widgets/settingspages/KeyboardSettingsPage.hpp \
     src/widgets/settingspages/ModerationPage.hpp \
+    src/widgets/settingspages/NicknamesPage.hpp \
```
zneix, line 588: After creating files for `RatelimitBucket` I ran `tools/update_filelist.py` and script also added this missing header on its own, I guess it could be sneaked in as a missing part of #2981.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
 - Major: Added the ability to add nicknames for users. (#137, #2981)
-- Major: Work on rate-limiting JOINs and PARTs. (#3112)
+- Major: Fixed constant disconnections with more than 20 channels by rate-limitting outgoing JOIN messages. (#3112, #3115)
```
zneix, line 7: Decided to rephrase the changelog a little, could some kind of God of War @Felanbird have a look and make sure this is clear and descriptive enough?

pajlada (changes requested):
This solution currently doesn't handle JOINs from when a user opens a new split - let's make sure we solve that in the same PR

zneix:
> This solution currently doesn't handle JOINs from when a user opens a new split - let's make sure we solve that in the same PR

Should be resolved in 61b0f39. Also, since that makes the use of `this->bucket_` outside of `AbstractIrcServer::onReadConnected` I think it's better to move initialization of it to where `AbstractIrcServer` itself is initizalied.

src/util/RatelimitBucket.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    this->queue_.append(message);
+
+    if (this->pending_ < this->limit_)
+    {
+        this->execute();
+    }
+}
+
+void RatelimitBucket::execute()
```
fourtf, line 25: Maybe change `execute` to a name which hints that the function only processes a single element like `handleOne` or `tick` or maybe something else. `execute` itself is a bit generic which makes it hard to guess what it does.

pajlada: Changed the name to `handleOne` and added a small comment about it in the header file

src/util/RatelimitBucket.hpp
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <QList>
+#include <QString>
+
+namespace chatterino {
+
+class RatelimitBucket
```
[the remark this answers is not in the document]
pajlada: `RatelimitBucket` has been made into a `QObject`.
The `Q_OBJECT` macro is not needed as far as I know as we don't do anything with signals or slots. (ref: https://doc.qt.io/qt-5/qobject.html#Q_OBJECT)

src/util/RatelimitBucket.cpp
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return;
+    }
+
+    auto item = queue_.takeFirst();
+
+    this->pending_++;
+    callback_(item);
+
+    QTimer::singleShot(cooldown_, [this] {
```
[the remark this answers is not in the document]
pajlada: We've switched to [this](https://doc.qt.io/qt-5/qtimer.html#singleShot-6) overload of `QTimer::singleShot` which accepts a `QObject` as its context as you suggested.

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -279,12 +279,20 @@ void AbstractIrcServer::onReadConnected(IrcConnection *connection)
...
+            connection->sendRaw("JOIN #" + message);
+        });
+
     // join channels
     for (auto &&weak : this->channels)
     {
         if (auto channel = weak.lock())
         {
-            connection->sendRaw("JOIN #" + channel->getName());
+            qCDebug(chatterinoIrc) << "sending to bucket" << channel->getName();
```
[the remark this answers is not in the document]
pajlada: Phrasing changed to `joining` as that's what will happen immediately as the callback is called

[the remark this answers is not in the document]
pajlada: Nevermind my last comment - I read the wrote qCDebug call. Fixed in a new commit.

src/providers/irc/AbstractIrcServer.hpp
```diff
@@ -88,6 +89,7 @@ class AbstractIrcServer : public QObject
     QObjectPtr<IrcConnection> writeConnection_ = nullptr;
     QObjectPtr<IrcConnection> readConnection_ = nullptr;
 
+    std::unique_ptr<RatelimitBucket> bucket_;
```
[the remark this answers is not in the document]
pajlada: `bucket_` renamed to `joinBucket_`

`joinBucket_` changed to a `QObjectPtr` as suggested

sando:
eb3e75c79ea2cb3bfd2a31e60370cd391f736c13 was crashing for me shortly after start up.

a60d8061c8c489eb033c0739a8cedf91d271423a seems to be working, so far so good! Thanks for the quick work on this. ❤️ 

46 channels open, Windows 11 21H2 (22000.100), chatterino-windows-x86-64-5.12.10-cmake

CHANGELOG.md
talneoran:
```suggestion
- Major: Fixed constant disconnections with more than 20 channels by rate-limiting outgoing JOIN messages. (#3112, #3115)
```

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -23,6 +23,15 @@ AbstractIrcServer::AbstractIrcServer()
     this->writeConnection_->moveToThread(
         QCoreApplication::instance()->thread());
 
+    // Apply a leaky bucket rate limitting to JOIN mesasges
```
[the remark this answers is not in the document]
pajlada: I edited your comment to also fix the `messages` word :D

ElZigSGC:
Work fine for me.
Thanks for the fast answer on that issue and for all the work done.

src/providers/irc/AbstractIrcServer.cpp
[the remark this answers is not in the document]
zneix: These debug prints were only added to ensure my logic works while preparing first commit to this PR and I was planning on removing it after code reviewers agree it works. Maybe it can stay though.

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -23,6 +23,19 @@ AbstractIrcServer::AbstractIrcServer()
...
+        if (!this->channels.contains(message))
+        {
+            qCDebug(chatterinoIrc)
+                << QString("attempted to join %1 but it does not exist anymore")
+                       .arg(message);
+            return;
+        }
+        this->readConnection_->sendRaw("JOIN #" + message);
+    };
+    this->joinBucket_.reset(new RatelimitBucket(18, 10500, actuallyJoin, this));
```
leon-richardt, line 37: Can we turn these magic numbers into named constants?

leon-richardt (changes requested):

src/providers/irc/AbstractIrcServer.cpp
zneix: Yes, I do think so. See 658f28f31eb771ca7fcd156b524de5addb8c3072

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -23,6 +27,20 @@ AbstractIrcServer::AbstractIrcServer()
     this->writeConnection_->moveToThread(
         QCoreApplication::instance()->thread());
 
+    // Apply a leaky bucket rate limiting to JOIN messages
+    auto actuallyJoin = [&](QString message) {
```
Mm2PL, line 31: nit: this variable being named `message` feels weird.

src/util/RatelimitBucket.hpp
```diff
@@ -0,0 +1,40 @@
...
+private:
+    /**
+     * @brief budget_ denotes the amount of calls that can be handled before we need to wait for the cooldown
+     **/
+    int budget_;
+
+    /**
+     * @brief This is the amount of time in milliseconds it takes for one used up budget to be put back into the bucket for use elsewhere
+     **/
+    int cooldown_;
```
Mm2PL, line 26:
```suggestion
    const int cooldown_;
```
It makes no sense for the cooldown to change.

src/util/RatelimitBucket.cpp
```diff
@@ -0,0 +1,45 @@
...
+                                 std::function<void(QString)> callback,
+                                 QObject *parent)
+    : QObject(parent)
+    , budget_(budget)
+    , cooldown_(cooldown)
+    , callback_(callback)
+{
+}
+
+void RatelimitBucket::send(QString message)
```
Mm2PL, line 17: q: Could a situation like this happen?
  1. `RatelimitBucket::send()` is called
  2. it appends the message/channel
  3. meanwhile `handleOne`'s empty check passed and it takes the element for itself
  4. then `send` calls `handleOne` with nothing in the queue

I'm scared that this could happen in extremely random cases and crash the app with a mysterious message that shouldn't happen.

Mm2PL (changes requested):
looks and works as expected, but I have a couple nit picks

src/providers/irc/AbstractIrcServer.cpp
zneix: How else would you name it then?

Mm2PL: I'd suggest `channel` since that's what the message is.

Mm2PL (approved):

src/util/RatelimitBucket.cpp
zneix: > I'm scared that this could happen in extremely random cases and crash the app with a mysterious message that shouldn't happen.

If anything, I'm pretty sure that a crash wouldn't happen. As stated by [Qt documentation](https://doc.qt.io/qt-5/qlist.html#removeFirst):

> The list must not be empty. If the list can be empty, call isEmpty() before calling this function.

Which is exactly what we're doing:

https://github.com/Chatterino/chatterino2/blob/658f28f31eb771ca7fcd156b524de5addb8c3072/src/util/RatelimitBucket.cpp#L29-L34

src/providers/irc/AbstractIrcServer.cpp
zneix: Originally, when I came up with the idea about this rate-limit system in my sleep I thought of it as something that could be used in multiple scopes, maybe outside of just `AbstractIrcServer`.

zneix: On another note though, I think we shouldn't worry about something that might not end up being implemented/utilized at all. So maybe we can just change the name to `channel` as you're suggesting and maybe change to something more generic later if my planned idea ever sees the daylight.

no1dead:
Using this change and haven't run into any issues with it.

pajlada (approved):
After testing with 80 channels this works well for me

We have some further UX work, but this is a great first step

src/providers/irc/AbstractIrcServer.cpp
```diff
@@ -23,6 +27,20 @@ AbstractIrcServer::AbstractIrcServer()
...
+        if (!this->channels.contains(message))
+        {
+            qCDebug(chatterinoIrc)
+                << QString("attempted to join %1 but it does not exist anymore")
```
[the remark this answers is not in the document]
zneix: oh yea, I wanted to purged it, doing it now.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - Major: Newly uploaded Twitch emotes are once again present in emote picker and can be autocompleted with Tab as well. (#2992)
 - Major: Added the ability to add nicknames for users. (#137, #2981)
-- Major: Work on rate-limiting JOINs and PARTs. (#3112)
+- Major: Fixed constant disconnections with more than 20 channels by rate-limiting outgoing JOIN messages. (#3112, #3115)
```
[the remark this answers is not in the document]
zneix: I wanted to have both "generic message for users" and something that covers #3112 in a "tech speak". Maybe it's to be changed, waiting for @Felanbird's opinion on my first unresolved comment on the PR. Good point though :+1:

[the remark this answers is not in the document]
zneix: I'll resolve this comment in favour of https://github.com/Chatterino/chatterino2/pull/3115#discussion_r682489073

src/providers/irc/AbstractIrcServer.cpp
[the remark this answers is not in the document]
zneix: Removed this and other debug print introduced by the PR in f1262fa.

CHANGELOG.md
leon-richardt: @Felanbird Please also consider this: https://github.com/Chatterino/chatterino2/pull/3115#discussion_r682809758 :smile:

leon-richardt (approved):

CHANGELOG.md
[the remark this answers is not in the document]
Felanbird: I think this is good, the only way would be able to better word it for users would probably be to change it to "rate-limiting connections…" but that doesn't technically apply here. The disconnection part should convey it properly.

CHANGELOG.md
zneix: Resolving in favour of https://github.com/Chatterino/chatterino2/pull/3115#discussion_r682911557

Felanbird (approved):
Changelog is good, functionality works as well, tested with 86 channels. 👍

src/providers/irc/AbstractIrcServer.cpp
fourtf: you send a message over a channel it doesn't really make sense to call a message that's being sent a channel, I think this is fine

fourtf (approved):

PR #3130, opened 2021-08-06
Handle moderation mode and filters in split popup

### LINKED ISSUE #3113: Keep filters when opening a split as a popout
**Describe your issue**

When opening a split as a popout by using CTRL/CMD + N right now, the resulting popout window does not keep any filters present in the split before active. Might be an intended behavior, might not be idk, but it would definitely be really handy to have an option to keep filters active.

I have some chats where i have up to 5 or so filters enabled and re-enabling them every time i want that chat as a separate window is really annoying.

**OS and Chatterino Version**

macOS Big Sur, Chatterino 2.3.3

talneoran:
Pull request checklist:

# Description

Respect moderation mode status and active filters when opening a split as a popup.

Fixes #3113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Minor: Moderation mode and active filters are now preserved when opening a split as a popup. (#3113, #3130)
+
 ## 2.3.4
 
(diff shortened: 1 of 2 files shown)


Felanbird (approved):
image

Mm2PL (approved):
lgtm

zneix (approved):

PR #3135, opened 2021-08-07
Show system message when reloading subscriber emotes

jammehcow:
Pull request checklist:

# Description

This PR aims to display a system message when reloading subscriber emotes. This behaviour is in line with the BTTV/FFZ emote reload behaviour. 

I've passed through the channel to `loadEmotes` which might be ugly to some people but it ensures you get a 'reloaded' messages when the reloading has started instead of when the method was called. Also kept the old `loadEmotes()` overload around for the call in `IRCMessagehandler`.

### DIFF
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -228,8 +228,11 @@ void TwitchAccount::loadEmotes(const std::shared_ptr<Channel> &channel)
     this->loadUserstateEmotes([this, channel] {
         // Fill up emoteData with emote sets that were returned in a Kraken call, but aren't present in emoteData.
         this->loadKrakenEmotes();
-        channel->addMessage(makeSystemMessage(
-            "Twitch subscriber emotes reloaded."));
+        if (channel != nullptr)
+        {
+            channel->addMessage(makeSystemMessage(
+                "Twitch subscriber emotes reloaded."));
+        }
     });
 }
 
--- a/src/providers/twitch/TwitchAccount.cpp
+++ b/src/providers/twitch/TwitchAccount.cpp
@@ -200,30 +200,7 @@ SharedAccessGuard<const std::set<QString>> TwitchAccount::accessBlockedUserIds()
 
 void TwitchAccount::loadEmotes()
 {
-    qCDebug(chatterinoTwitch)
-        << "Loading Twitch emotes for user" << this->getUserName();
-
-    if (this->getOAuthClient().isEmpty() || this->getOAuthToken().isEmpty())
-    {
-        qCDebug(chatterinoTwitch)
-            << "Aborted loadEmotes due to missing Client ID and/or OAuth token";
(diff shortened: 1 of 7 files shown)


zneix (changes requested):

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -199,6 +199,11 @@ SharedAccessGuard<const std::set<QString>> TwitchAccount::accessBlockedUserIds()
 }
 
 void TwitchAccount::loadEmotes()
+{
+    loadEmotes(nullptr);
```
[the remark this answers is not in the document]
jammehcow: Fixed!

zneix (approved):
LGTM :+1:

On a side note: Just wondering if wording "Twitch **subscriber** emotes" is correct nowadays. It used to be like that, but now we also have follower and bit emotes as well. Potential wording change would also mean changing that phrasing in `KeybindingsPage.cpp` and couple places in `SplitHeader.cpp`.

jammehcow:
> LGTM +1
> 
> On a side note: Just wondering if wording "Twitch **subscriber** emotes" is correct nowadays. It used to be like that, but now we also have follower and bit emotes as well. Potential wording change would also mean changing that phrasing in `KeybindingsPage.cpp` and couple places in `SplitHeader.cpp`.

Start a separate issue/discussion for that - out of scope image

src/widgets/splits/SplitHeader.cpp
```diff
@@ -916,10 +916,6 @@ void SplitHeader::themeChangedEvent()
     }
 }
 
-void SplitHeader::moveSplit()
-{
-}
-
```
apa420, line 922: Start a separate issue/discussion for that - out of scope image

src/widgets/splits/SplitHeader.hpp
```diff
@@ -85,7 +85,6 @@ class SplitHeader final : public BaseWidget, pajlada::Signals::SignalHolder
     std::vector<pajlada::Signals::ScopedConnection> channelConnections_;
 
 public slots:
-    void moveSplit();
```
apa420, line 88: Just because of your other comment :upside_down_face: :tf:

src/widgets/splits/SplitHeader.cpp
jammehcow: No
I do not think so

src/widgets/splits/SplitHeader.hpp
jammehcow: :tf:

zneix:
> Start a separate issue/discussion for that - out of scope image

Yeah I don't think it should be a part of this one, I'm just sharing my idea here and an issue / PR could be made out of my comment later.

pajlada (approved):
:+1:

zneix (approved):
Pajlada's changes seem nice

PR #3143, opened 2021-08-07
Strip leading @ and trailing , from /user and /usercard commands

sando:
Pull request checklist:

# Description

Strip the leading `@` from the username when using the `/user` and `/usercard` commands. Fixes a minor annoyance when these commands are used along with the *Only search for username autocompletion with an @* option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
-- Minor: Strip leading @ from username in /user and /usercard commands (#3143)
+- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
 - Bugfix: Moderation mode and active filters are now preserved when opening a split as a popup. (#3113, #3130)
 - Bugfix: Fixed a bug that caused all badge highlights to use the same color. (#3132, #3134)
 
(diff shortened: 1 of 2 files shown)


apa420:
wouldn't it just be better to take the @ as input in the /user and /usercard commands instead of stripping the @?

sando:
> wouldn't it just be better to take the @ as input in the /user and /usercard commands instead of stripping the @?

Not sure if I understand what you mean here - keeping the @ opens invalid links in user/usercard.

Admittedly I didn't think about when mentioning users with a comma is enabled, but the option doesn't appear to affect it in this case, e.g. typing `/user username` then tab doesn't append with a comma anyway

apa420:
I was thinking maybe handling it in the /user or /usercard command, but it's really up to preference I guess.

ALazyMeme:
What about the following `,`? Should that be added in the same pr?

Felanbird:
> What about the following `,`? Should that be added in the same pr?

The `,` no longer follows when the tab completion isn't in the first position thanks to the changes implemented in #3060.

Felanbird (approved):

Before this PR:
image

Now:
image
urls work too but extra screenshots

Felanbird:
I did notice one thing specific to `/user`, this doesn't fix `/user @username @channel` I'm not sure if anybody actually ever uses it like this? but it might be nice to fix while we are at it.

sando:
I personally don't use it that way, but it's a good idea

Felanbird (approved):

apa420 (approved):
Ignore my previous comments, I didn't read the code and the description confused me. I thought you had stripped the @ while typing the message. For some reason I thought this has to do with autocomplete since you were writing about @'ing people.

LGTM!

Mm2PL (approved):
Lgtm

pajlada (approved):

sando:
LGTM 👍

brian6932:
Can we do this with /streamlink as well?

PR #3145, opened 2021-08-07
Added links to discussions when trying to create an issue

apa420:
Taken from osu!lazer
This is what it looks like:
https://github.com/ppy/osu/issues/new/choose

# Description
Makes a link to discussions when trying to create an issue when you should instead create an enhancement.


apa420:
Fourtf approved this pr in discord dms.
```
Good idea
LGTM
Paj can merge if he agrees as well
```

pajlada (approved):

PR #3146, opened 2021-08-07
Added regex option to Nicknames

apa420:
Pull request checklist:

# Description
Added regex option to nicknames as well as case sensitivity option which is enabled by default when creating an object.

Regex now support the \1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Minor: Commands are now backed up. (#3168)
 - Minor: Added the ability to open an entire tab as a popup. (#3082)
 - Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
+- Minor: Added regex option to Nicknames. (#3146)
 - Bugfix: Fixed colored usernames sometimes not working. (#3170)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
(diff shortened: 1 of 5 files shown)


Mm2PL (changes requested):
I'm unable to get references to the original username working. That works in ignores and maybe should work here too.

Mm2PL (approved):

xHeaveny (approved):

Felanbird:
I'm unsure if this is something we can fix, but if you leave the username box blank with regex enabled, it places the nicknamed phrase in-between every character of the username.

image

image
Case-sensitivity setting having no effect on this fact.

pajlada:
Real use cases:
 - Anonymize chat by replacing `.+` with `User`. (New use case)
 - Simplifies doing modifications on usernames instead of full replacement, i.e. if I want to add a star to `zneix`'s and `gempir`'s names, I could add the following replacements: `zneix` -> `⭐ \1`, `gempir` -> `⭐ \1` instead of `zneix` -> `⭐ zneix` and `gempir` -> `⭐ gempir`. (Simplifies already possible use case)
 - Filter out offensive phrases from usernames. While most offensive phrases might be caught by Twitch already, some offensive phrases are more personal. (New use case)

Meme use cases:
 - Putting a badge in front of everyone's username by replacing `.+` with `🐝 \1`

pajlada (approved):

zneix (approved):

PR #3148, opened 2021-08-08
Allow resub messages to show in /mentions tab

### LINKED ISSUE #1013: resub messages are not shown in /mentions
Highlight and custom sound is working in resub-messages, but it does not show up in the /mentions tab. Having it there would make the message that pings you easier to identify, when it comes from a huge channel (e.g. forsen)

grafik

grafik

jammehcow:
Pull request checklist:

# Description

This PR aims to fix resub messages not showing in the mentions tab when they trigger a highlight. Previously resub/sub messages were explicitly excluded from the `/mentions` tab (for whatever reason). This clause has been removed.

The sub message block (first clause in the highlight parser) has had the return statement removed so it can handle more highlights. I've added checks to see if the message is a sub message in all other highlight checking blocks so the colour of the sub message isn't overridden.

Fixes #1013


zneix (approved):
Code looks very good, I just need to find a good way to test this and I'll approve & merge the changes.

Edit: I wanted to make a comment instead of approving but fat-fingered, whatever it should just work anyway.

Felanbird (approved):
image

I forgot to set a custom color to test the overriding.

jammehcow:
Using the test message stuff in `Window.cpp` (didn't know that existed):

In the channel:
image

In `/mentions`:
image

pajlada (approved):

PR #3153, opened 2021-08-09
Ignore phrase by specified user

mammothb:
Pull request checklist:

# Description

- Added an additional "User" column to the Ignores -> Messages tab to allow users to set ignore patterns for a specific user
- Intended usecase: when a user is spamming too much of a particular emote and you would like to filter that out while still see the emote when you or other users send it


apa420:
Can't the R9K mode in chatterino do this already? ( Blocking spam by the same user )

Should this only be able to add one user per field.
Which means that if you wanted to add this to ~10 people you'd have to create the same ignore pattern 10 times.
Supporting splitting users with `,` might be a good idea. `user1, user2, user3` or `user1 user2` etc.
Or support regex for the user field like #3146.

mammothb:
This is really just for the niche use case where you only want to replace the "annoying" part of a user's message without blocking the entire message or user. I initially developed this for personal use but someone in chat asked me to make a PR so here it is.

Based on that use case, I'm not sure if people will be applying the pattern to multiple users, at least that hasn't been the case for me. However, if you think that supporting splitting users might still be useful, I'll try to implement that.

src/widgets/settingspages/IgnoresPage.cpp
```diff
@@ -49,7 +49,7 @@ void addPhrasesTab(LayoutCreator<QVBoxLayout> layout)
                     ->initialized(&getSettings()->ignoredMessages))
             .getElement();
     view->setTitles(
-        {"Pattern", "Regex", "Case Sensitive", "Block", "Replacement"});
+        {"Pattern", "User", "Regex", "Case Sensitive", "Block", "Replacement"});
```
apa420, line 52: Line 44 (that github doesn't let me mark)
You should add a comment says something like: "Leaving the User tab empty will make it target all users"

apa420 (changes requested):

src/widgets/settingspages/IgnoresPage.cpp
```diff
@@ -41,15 +41,17 @@ IgnoresPage::IgnoresPage()
 
 void addPhrasesTab(LayoutCreator<QVBoxLayout> layout)
 {
-    layout.emplace<QLabel>("Ignore messages based certain patterns.");
+    layout.emplace<QLabel>(
+        "Ignore messages based certain patterns. "
+        "Leaving the User tab empty will make it target all users");
```
apa420, line 46:
```suggestion
        "Ignore messages based certain patterns.\n"
        "Leaving the User tab empty will make it target all users");
```

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -795,6 +795,12 @@ void TwitchMessageBuilder::runIgnoreReplaces(
         {
             continue;
         }
+        const auto &user = phrase.getUser();
+        if (user.isEmpty() &&
```
apa420, line 799: Shouldn't this be 
```suggestion
        if (!user.isEmpty() &&
```
So it's the same as before?

apa420: Also you can press commit suggestion here on github and it'll automatically do the work for you if you agree with the change.

mammothb:
Sorry that this too longer than intended, splitting users by `,` has been implemented. And I updated the placeholder entry to `user1,user2` to demonstrate that.

pajlada:
Closing this for now, feel free to reach out or create a discussion to discuss your use-case further.

Issue #2326, opened 2021-01-01
Allow for if variables within commands


alazymeme:
**What should be added?**
Add an `{if}` variable for commands

**Why should it be added?**
Could be quite useful for a variety of things

**Example**
`{if:lol,"lol","xd"}`

Mm2PL:
Another take on this idea is a special `/if` built-in command that would take a condition (parsable using the existing filter system), then a separator and if the condition was `true`, another separator and `false` output.

jupjohn:
@ALazyMeme with the introduction of plugins, do you still see a benefit having a dedicated `if` function to commands? I would argue that if you're thinking in `if` statements, Lua isn't too much of a jump.

PR #3155, opened 2021-08-09
Add new command placeholders: {channel.name}, {channel.id}, {stream.game}, {stream.title}, {my.id}, {my.name} 

### LINKED ISSUE #2943: Variable placeholders for custom commands
**Describe your issue**
Add a `{channel}` placeholder for commands that would put the current channel. This would be useful for opening moderation popout widgets, like reward queue, predictions, etc.

**Screenshots**
example

**OS and Chatterino Version**
Windows 10 - Latest Nightly

Mm2PL:
Pull request checklist:

# Description

~~Part of #2943. Doesn't implement alternate text so I'll keep the issue open.~~
Closes #2943.


apa420:
If this closes #2943 that should also be part of the CHANGELOG entry (also shouldn't new changelog entries be at the bottom of the list for the unversioned changes with the same type.)

ALazyMeme:
> There is also #2326

Hmm I don't remember making this susLada

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Add new command placeholders (#3155)
```
pajlada, line 5: More verbose changelog entry

pajlada (changes requested):
We should document the new variables on https://wiki.chatterino.com/Commands/ too

once you have a PR there we can get them merged in at the same time

make sure to include from which version onwards the new variables in https://wiki.chatterino.com/Commands/ are available

Mm2PL:
https://github.com/Chatterino/wiki/pull/113

ALazyMeme:
Is there potentially a use-case for filtered variables here? For example instead of `{channelid}`, we have a filter similar to pajbot `|id`. Could be used in a number of cases like: `{1|id}` which would get the user id of the user in the first argument, or `{channel|id}`, which would get the current channel's id. This would thus remove the `{channelid}` and `{myid}` variables, but opens us up for more options to use it (instead of just being limited to the variables that are already there).

pajbot filters here for ref: https://github.com/pajbot/pajbot/blob/master/docs/variables.md#available-filters

pajlada (approved):

src/widgets/helper/ChannelView.cpp
```diff
@@ -2101,12 +2101,21 @@ void ChannelView::handleLinkClick(QMouseEvent *event, const Link &link,
...
+                    {"msg.id", layout->getMessage()->id},
+                    {"msg.text", layout->getMessage()->messageText},
+
+                    // old placehodlers
+                    {"user", layout->getMessage()->loginName},
+                    {"msg-id", layout->getMessage()->id},
+                    {"message", layout->getMessage()->messageText},
+
+                    // new version of this is inside execCustomCommand
+                    {"channel", this->channel()->getName()},
```
zneix, line 2117: Will we mark these as deprecated and remove them later or are we just gonna keep these as legacy?

Mm2PL: I don't like breaking things, maybe we do it some time should though.

zneix: Yes that shouldn't be changed right now, but maybe a note for users that new variables should be used now and maybe remove this like after over a few stable releases to not keep stale legacy forever is reasonable.

pajlada (approved):

zneix (approved):

PR #3156, opened 2021-08-09
tidy up for loop variable type

enkeyz:
Pull request checklist:

# Description

Tidy up for loop variable type, so it's consistent with return type of size()

### DIFF
--- a/src/singletons/Settings.cpp
+++ b/src/singletons/Settings.cpp
@@ -85,8 +85,7 @@ void ConcurrentSettings::mute(const QString &channelName)
 
 void ConcurrentSettings::unmute(const QString &channelName)
 {
-    for (std::vector<int>::size_type i = 0; i != mutedChannels.raw().size();
-         i++)
+    for (std::size_t i = 0; i != mutedChannels.raw().size(); i++)
     {
         if (mutedChannels.raw()[i].toLower() == channelName.toLower())
         {


src/singletons/Settings.cpp
```diff
@@ -85,8 +85,7 @@ void ConcurrentSettings::mute(const QString &channelName)
 
 void ConcurrentSettings::unmute(const QString &channelName)
 {
-    for (std::vector<int>::size_type i = 0; i != mutedChannels.raw().size();
-         i++)
+    for (std::size_t i = 0; i != mutedChannels.raw().size(); i++)
```
apa420, line 88: I don't think this change is correct if we check https://en.cppreference.com/w/cpp/container/vector/size and see that the vector indices will be determined by the type of the vector.
Which in this case
https://github.com/Chatterino/chatterino2/blob/50308905f42ce14d0de6c5369a515d58a0e559af/src/singletons/Settings.cpp#L24
is a QString.
So the correct change would be:
```suggestion
    for (std::vector<QString>::size_type i = 0; i != mutedChannels.raw().size(); i++)
```
I think...

enkeyz: If you open up the code in an IDE, it's show size() return type as std::size_t.

enkeyz: Screenshot from 2021-08-09 22-38-04

Mm2PL: The `size()` method is defined as follows:
```
      // [23.2.4.2] capacity
      /**  Returns the number of elements in the %vector.  */
      size_type
      size() const _GLIBCXX_NOEXCEPT
      { return size_type(this->_M_impl._M_finish - this->_M_impl._M_start); }
```
(src: `stl_vector.h`)

It actually returns `std::vector<T>::size_type`, not `size_t` directly

apa420: I'm not really arguing against you.
And probably every compiler agrees with you.

But according the standard this will NOT always return size_t and while your compiler will return size_t doesn't mean it will on every compiler.

(We don't do this anywhere else in the code either so your change is valid.)

enkeyz: Even tho you're right, `std::vector<QString>::size_type` is the correct way doing it, it's ugly and never seen it anywhere else.

leon-richardt:
Just wanna mention that at the moment, `std::size_t` and `size_t` are both being used in the Chatterino code base.

They refer to the same type so functionally, there is no difference. With PRs like this, I think it is reasonable to decide on a common style go with in the future. Since `size_t` is being used dominantly in the code base (81 uses at the time of writing vs. 5 uses of `std::size_t`), I would advocate for `size_t` being made convention.

PR #3157, opened 2021-08-09
Fixed Searching for subcategories

### LINKED ISSUE #2707: Allow to search settings by subcategories.
**What should be added?**

I would like for subcategories (smaller headers) to be searchable in the settings.

Example of what I mean:
 - I type `R9K` into the search bar and only this would show. (Current behavior produces no search results)
   

**Why should it be added?**

Categories are searchable, why should the subcategories not be searchable? Also useless QoL

apa420:
Pull request checklist:

# Description
Fixes #2707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
-- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
+- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#2707, #3143)
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
 - Minor: Subcategories in settings are now searchable. (#3157)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
 - Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
+- Minor: Subcategories in settings are now searchable. (#3157)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
 - Bugfix: Moderation mode and active filters are now preserved when opening a split as a popup. (#3113, #3130)
 - Bugfix: Fixed a bug that caused all badge highlights to use the same color. (#3132, #3134)
--- a/src/widgets/settingspages/GeneralPageView.cpp
+++ b/src/widgets/settingspages/GeneralPageView.cpp
@@ -291,22 +291,30 @@ bool GeneralPageView::filterElements(const QString &query)
 
             QWidget *currentSubtitle = nullptr;
             bool currentSubtitleVisible = false;
+            bool currentSubtitleSearched = false;
 
             for (auto &&widget : group.widgets)
             {
-                if (dynamic_cast<SubtitleLabel *>(widget.element))
+                if (auto x = dynamic_cast<SubtitleLabel *>(widget.element))
                 {
+                    currentSubtitleSearched = false;
                     if (currentSubtitle)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,8 +3,9 @@
 ## Unversioned
 
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
-- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
+- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#2707, #3143)
```
```suggestion
- Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
```

CHANGELOG.md
```diff
@@ -3,8 +3,9 @@
...
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
+- Minor: Subcategories in settings are now searchable. (#3157)
```
Felanbird, line 8:
```suggestion
- Minor: Subcategories in settings are now searchable. (#2707, #3157)
```

Felanbird (changes requested):
Misplaced your changelog entry xd

Felanbird (approved):

Only issue I could find was attempting to search the `Custom Stream Player` Subcategory under `External Tools`, as `Custom Sound` appears first, but I assume this is not something to be fixed by this PR.

apa420:
This pr only fixes the generalpage subcategories (I only edited the GeneralPageView after all)

Felanbird:
> This pr only fixes the generalpage subcategories (I only edited the GeneralPageView after all)

Ah yes, I tested wrong when I retried searching from my current build without this pr 😄

src/widgets/settingspages/GeneralPageView.cpp
```diff
@@ -291,22 +291,30 @@ bool GeneralPageView::filterElements(const QString &query)
...
             QWidget *currentSubtitle = nullptr;
             bool currentSubtitleVisible = false;
+            bool currentSubtitleSearched = false;
 
             for (auto &&widget : group.widgets)
             {
-                if (dynamic_cast<SubtitleLabel *>(widget.element))
+                if (auto x = dynamic_cast<SubtitleLabel *>(widget.element))
                 {
+                    currentSubtitleSearched = false;
```
zneix, line 300: Why is this added here, while we have the same exact line in line 304.?

src/widgets/settingspages/GeneralPageView.cpp
```diff
@@ -291,22 +291,30 @@ bool GeneralPageView::filterElements(const QString &query)
 
             QWidget *currentSubtitle = nullptr;
             bool currentSubtitleVisible = false;
+            bool currentSubtitleSearched = false;
 
             for (auto &&widget : group.widgets)
             {
-                if (dynamic_cast<SubtitleLabel *>(widget.element))
+                if (auto x = dynamic_cast<SubtitleLabel *>(widget.element))
```
zneix, line 298: Sorry for nitpicking, but please change the variable's name to something more explanatory.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
 - Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
+- Minor: Subcategories in settings are now searchable. (#2707, #3157)
```
zneix, line 8: As agreed with pajlada, we're now against putting issue numbers into changelog entries. CC @felanbird

```suggestion
- Minor: Subcategories in settings are now searchable. (#3157)
```

zneix (changes requested):
The fix indeed does what it is supposed to do, however I am pretty sure that the freeze that occurs upon putting anything into the settings search box is even longer than it originally was and I am wondering if that can be optimized somehow.
Other than nitpicks from code review, good stuff :+1:

src/widgets/settingspages/GeneralPageView.cpp
apa420: https://github.com/Chatterino/chatterino2/blob/0a51252b761ac4a846725ce6b6e8cf8bba5efdf4/src/widgets/settingspages/GeneralPageView.cpp#L266
I was using the same naming scheme as the code above.

zneix: Okay, I guess sticking to the consistency with code above makes sense then :ok_hand: I was just a little confused myself while reading this but it's ok as the variable is only used once.

PR #3172, opened 2021-08-15
Add optional parameter to /usercard command for opening a usercard in a different channel context.

### LINKED ISSUE #3167: Allow /usercard to use a different channel context
**Describe your issue**
Sometimes, I would like to see a user card in a different channel context. Instead of having to open that channel in a new split, I would prefer to be able to type the chat command `/usercard <user> [channel]` to view `user`s user card in the context of `channel`. If the `channel` argument is omitted, the current behavior (opening the user card in the current channel's context) is used.

Also, the name of the channel the user card belongs to should be mentioned somewhere. This could just be the window title or a separate field on the user card itself.

LosFarmosCTL:
Pull request checklist:

# Description

Closes #3167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,13 @@
 - Minor: Duplicate spaces now count towards the display message length. (#3002)
 - Minor: Commands are now backed up. (#3168)
 - Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
+=======
+- Bugfix: Fixed colored usernames sometimes not working. (#3170)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
 - Bugfix: Moderation mode and active filters are now preserved when opening a split as a popup. (#3113, #3130)
 - Bugfix: Fixed a bug that caused all badge highlights to use the same color. (#3132, #3134)
+- Bugfix: Allow starting Streamlink from Chatterino when running as a Flatpak. (#3178)
 - Dev: Renamed CMake's build option `USE_SYSTEM_QT5KEYCHAIN` to `USE_SYSTEM_QTKEYCHAIN`. (#3103)
 - Dev: Add benchmarks that can be compiled with the `BUILD_BENCHMARKS` CMake flag. Off by default. (#3038)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Minor: Added a setting to hide similar messages by any user. (#2716)
 - Minor: Duplicate spaces now count towards the display message length. (#3002)
 - Minor: Commands are now backed up. (#3168)
+- Minor: Add optional parameter to /usercard command for opening a usercard in a different channel context. (#3172, #3167)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
 - Bugfix: Moderation mode and active filters are now preserved when opening a split as a popup. (#3113, #3130)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
 - Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
+- Minor: Allow resub messages to show in `/mentions` tab (#3148)
 - Minor: Added a setting to hide similar messages by any user. (#2716)
 - Minor: Duplicate spaces now count towards the display message length. (#3002)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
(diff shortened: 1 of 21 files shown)


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -33,7 +33,7 @@
 const QString TEXT_VIEWS("Views: %1");
 const QString TEXT_FOLLOWERS("Followers: %1");
 const QString TEXT_CREATED("Created: %1");
-const QString TEXT_TITLE("%1's Usercard");
+const QString TEXT_TITLE("%1's Usercard - %2");
```
apa420, line 36: It might be a good idea to make some kind of indication to which one's the channel name and which is the user name.
A # like in irc might be the way to go.
```suggestion
const QString TEXT_TITLE("%1's Usercard - #%2");
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Minor: Added a setting to hide similar messages by any user. (#2716)
 - Minor: Duplicate spaces now count towards the display message length. (#3002)
 - Minor: Commands are now backed up. (#3168)
+- Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172, #3167)
```
apa420, line 12: We're now supposed to not include the issue number here.
```suggestion
- Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
```

apa420:
Make sure you run clang on your code otherwise the formatter is going to complain.
https://github.com/Chatterino/chatterino2#get-it-automated-with-qt-creator--beautifier--clang-format

LosFarmosCTL:
> Make sure you run clang on your code otherwise the formatter is going to complain.

Yeah, i apparently did something wrong when setting that up, should be fixed now

src/controllers/commands/CommandController.cpp
```diff
@@ -482,6 +482,21 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+
+        if (words.size() > 2)
+        {
+            QString channelName = words[2];
+            stripChannelName(channelName);
+
+            ChannelPtr channelTemp =
+                getApp()->twitch2->getChannelOrEmpty(channelName);
+
+            if (channelTemp->getType() != Channel::Type::None)
```
Mm2PL, line 494:
```suggestion
            if (channelTemp->isEmpty())
```
I believe that using `isEmpty()` is better because it's clearer to understand what this code does. :)

src/controllers/commands/CommandController.cpp
```diff
@@ -482,6 +482,21 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            {
+                channel = channelTemp;
+            }
```
Mm2PL, line 497: Could there be some sort of a warning that Chatterino doesn't know about the supplied channel?

Mm2PL (changes requested):
Just a few things. :)

src/controllers/commands/CommandController.cpp
LosFarmosCTL: Yeah, i though about it as well, but i am not sure where to show that warning...

Maybe a message in chat, similar to how it works when supplying invalid arguments?

chatterino 08-15 21 48 37

Also, this message should probably be updated in general to include the new argument.

LosFarmosCTL: Ok, I added an error message using a system message displayed in chat, maybe take a quick look at it if it seems okay :)

Mm2PL: The error message works as expected, however now since you don't override `channel` with contents of `channelTemp`, the command always falls back to the current channel.

LosFarmosCTL: Yeah i did that on purpose, would be a weird UX to display an error message but still go through with an action as if nothing happened imo

Could change it back though of course

LosFarmosCTL: Oh yeah no wait i see the problem now FDM

LosFarmosCTL: Should be fixed now

Mm2PL: Feel free to resolve this conversation :)

Mm2PL (approved):

apa420 (approved):
Except for maybe adding a dot to the error message this LGTM.

LosFarmosCTL:
> Except for maybe adding a dot to the error message this lgtm.

Yeah, the error messages seem a bit inconsistent in general, i copied the absent point from the /chatters command directly next to /usercard, but it definitely looks better with one.

src/controllers/commands/CommandController.cpp
```diff
@@ -476,12 +476,33 @@ void CommandController::initialize(Settings &, Paths &paths)
     this->registerCommand("/usercard", [](const auto &words, auto channel) {
         if (words.size() < 2)
         {
-            channel->addMessage(makeSystemMessage("Usage /usercard [user]"));
+            channel->addMessage(
+                makeSystemMessage("Usage /usercard [user] (channel)"));
```
leon-richardt, line 480: I propose sticking to man page syntax (does this syntax have an official name?) for describing command parameters. In this case, that would mean:
```suggestion
                makeSystemMessage("Usage: /usercard user [channel]"));
```

Mm2PL: We already use a different syntax for optional parameters. Maybe we could switch, but I it's or of scope of this PR

LosFarmosCTL: Yeah, I got the [required] (optional) from the /user command, but if <required> [optional] is the way going into the future this should probably be changed, no need to stick to the old one then.

pajlada (approved):

zneix (approved):

PR #3178, opened 2021-08-17
Use flatpak-spawn to run streamlink when running as a flatpak

ilyazzz:
Pull request checklist:

# Description

Currently, when Chatterino is installed as a Flatpak package, streamlink doesn't work because it is not included in the application runtime. This PR makes it so chatterino calls the streamlink command with `flatpak-spawn --host` when it detcets a flatpak runtime, which means the command will be called directly on the host machine, and not within the application's sandbox.

Fixes https://github.com/flathub/com.chatterino.chatterino/issues/1


src/common/Version.hpp
```diff
@@ -29,6 +29,7 @@ class Version
     const QString &dateOfBuild() const;
     const QString &fullVersion() const;
     const bool &isSupportedOS() const;
+    const bool isFlatpak() const;
```
leon-richardt, line 32:
```suggestion
    bool isFlatpak() const;
```

Since we return a copy anyway, this should return a `bool` instead of a `const bool`. (Also see [this note](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-57) in the C++ core guidelines)
(Remember to change the function definition in the source file as well)

leon-richardt (changes requested):
Very nice that you're looking into this 👍

I'm wondering though whether the logic for building the program call to open Streamlink via `flatpak-spawn` could also be integrated into `getStreamlinkProgram()`? I believe that could be nicer in regards to SoC and code reusability

ilyazzz:
> Very nice that you're looking into this +1
> 
> I'm wondering though whether the logic for building the program call to open Streamlink via `flatpak-spawn` could also be integrated into `getStreamlinkProgram()`? I believe that could be nicer in regards to SoC and code reusability

The problem with integrating this into `getStreamlinkProgram()` is that `QProcess` takes in the binary name and the arguments separately. Normally the streamlink binary would be the actual binary name, but when running inside the sandbox the binary name has to be `flatpak-spawn`, with the streamlink path being an argument. It would be nice to centralize it into a single method, but in that case it would have to return both the binary and arguments (which would be empty outside of flatpak).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Bugfix: Fixed a bug that caused all badge highlights to use the same color. (#3132, #3134)
 - Dev: Renamed CMake's build option `USE_SYSTEM_QT5KEYCHAIN` to `USE_SYSTEM_QTKEYCHAIN`. (#3103)
 - Dev: Add benchmarks that can be compiled with the `BUILD_BENCHMARKS` CMake flag. Off by default. (#3038)
+- Bugfix: Use `flatpak-spawn` to call streamlink when running as a Flatapk. Fixes streamlink functionality. (#3178)
```
leon-richardt, line 18: I feel like this changelog entry isn't very user-friendly. Maybe change to something along the lines of:
```suggestion
- Bugfix: Allow starting Streamlink from Chatterino when running as a Flatpak. (#3178)
```

(cc: @Felanbird @ALazyMeme for grammar checks :smile:)

Felanbird: It's in the dev section anyway I was just gonna suggest changing it to `Dev:`, but this works too

leon-richardt: > It's in the dev section

Good catch, didn't even notice that. I still feel like listing this as a "Bugfix" entry adds more value to changelog readers rather than it just being a "Dev" entry

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 - Bugfix: Fixed a bug that caused all badge highlights to use the same color. (#3132, #3134)
 - Dev: Renamed CMake's build option `USE_SYSTEM_QT5KEYCHAIN` to `USE_SYSTEM_QTKEYCHAIN`. (#3103)
 - Dev: Add benchmarks that can be compiled with the `BUILD_BENCHMARKS` CMake flag. Off by default. (#3038)
-- Bugfix: Use `flatpak-spawn` to call streamlink when running as a Flatapk. Fixes streamlink functionality. (#3178)
+- Bugfix: Allow starting Streamlink from Chatterino when running as a Flatpak. (#3178)
```
Felanbird, line 18: bump above dev entries

src/util/StreamLink.cpp
```diff
@@ -165,15 +179,31 @@ void getStreamQualities(const QString &channelURL,
             }
         });
 
-    p->setArguments({channelURL, "--default-stream=KKona"});
+    p->setArguments(p->arguments() +
+                    QStringList({channelURL, "--default-stream=KKona"}));
```
zneix, line 183: A little nitpick
```suggestion
    p->setArguments(p->arguments() +
                    QStringList{channelURL, "--default-stream=KKona"});
```

src/util/StreamLink.cpp
```diff
@@ -165,15 +179,31 @@ void getStreamQualities(const QString &channelURL,
...
+    if (Version::instance().isFlatpak())
+    {
+        program = "flatpak-spawn";
+
+        arguments << "--host" << getStreamlinkProgram();
+    }
+    else
+    {
+        program = getStreamlinkProgram();
+    }
```
zneix, line 204: Maybe makes the code a little less readable, but I feel like we can get rid of this else at the end.
```suggestion
    QString program = getStreamlinkProgram();

    if (Version::instance().isFlatpak())
    {
        program = "flatpak-spawn";

        arguments << "--host" << getStreamlinkProgram();
    }
```

leon-richardt:
I have implemented the concept we discussed in chat in https://github.com/ilyazzz/chatterino2/pull/1. Please consider before merging

leon-richardt (approved):
Tested on non-Flatpak version, works as expected 👍 @ilyazzz verified [it works on Flatpak](https://github.com/ilyazzz/chatterino2/pull/1#issuecomment-903073850) as well. I would just like some additional input on that `REVIEW` comment

leon-richardt:
Tests fail because we get a [502 on some requests ](https://github.com/Chatterino/chatterino2/pull/3178/checks?check_run_id=3388248055#step:9:60)to httpbin.org:
```
/home/runner/work/chatterino2/chatterino2/tests/src/NetworkRequest.cpp:139: Failure
Expected equality of these values:
  result.status()
    Which is: 502
  code
    Which is: 406
```

pajlada (approved):

PR #3187, opened 2021-08-22
Parse usernames out of /mods and /vips messages

Mm2PL:
Pull request checklist:

# Description

### DIFF
--- a/src/providers/twitch/TwitchMessageBuilder.cpp
+++ b/src/providers/twitch/TwitchMessageBuilder.cpp
@@ -1432,4 +1432,70 @@ void TwitchMessageBuilder::deletionMessage(const DeleteAction &action,
     builder->message().timeoutUser = "msg:" + action.messageId;
 }
 
+void TwitchMessageBuilder::modsOrVipsSystemMessage(QString prefix,
+                                                   QStringList users,
+                                                   TwitchChannel *channel,
+                                                   MessageBuilder *builder)
+{
+    builder->emplace<TimestampElement>();
+    builder->message().flags.set(MessageFlag::System);
+    builder->message().flags.set(MessageFlag::DoNotTriggerNotification);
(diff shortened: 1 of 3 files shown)


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1432,4 +1432,70 @@ void TwitchMessageBuilder::deletionMessage(const DeleteAction &action,
...
+                                  MessageColor::System);
+    bool hasColor = false;
+    bool isFirst = true;
+    MessageColor color;
+    for (const QString &username : users)
+    {
+        if (!isFirst)
+        {
+            // this is used to add the ", " after each but the last entry
+            builder->emplace<TextElement>(", ", MessageElementFlag::Text,
```
Felanbird, line 1453:
```suggestion
            builder->emplace<TextElement>(",", MessageElementFlag::Text,
```

Before:
image

After:
image

Felanbird (changes requested):
Also as seen from my suggestion comment, the `,` does not properly wrap around.
image

apa420:
Did you test this on a channel that doesn't have mods or vips?

Felanbird:
> Did you test this on a channel that doesn't have mods or vips?

image

Mm2PL:
@apa420 it doesn't change anything about those cases, this PR only changes success/result msg-ids. :)

apa420:
But returning that there aren't any vips or mods could be a successful request...
Hence why I asked.

Also I don't personally think these messages should use the username color since that's not the case with any other system messages afaik.

zneix:
On the other hand, using colors in these messages could be useful, e.g. when you quickly wanna check which mods are "active", since color is only applied if a user has typed anything in chat before. I could see that being potentially useful in some cases, where you need to get help of a channel moderator.

Issue #1855, opened 2020-08-13
Connections (possibly) not being closed immediately after closing the channel.


Felanbird:
**Describe the bug**

Sometimes after closing a chatterino tab, your connection to that channel is not immediately terminated, and such logging and mentions still run normally.

**To reproduce**

It's not consistently able to be replicated, but in my instance the following happened:
* ``` # Start logging at 2020-08-13 18:11:14 Eastern Daylight Time ```
* Approximately 105 messages were sent before I closed the tab, at around ``` [18:46:17] EST ```
* Chatterino ping sound goes off and I receive a message in /mentions at ``` [18:50:13] EST ```
* Last log is recieved at ``` [19:15:45] ```

**Screenshots**

Additional issue holder.
![image](https://user-images.githubusercontent.com/41973452/90196598-3a3d0100-dd9a-11ea-8685-3df6a960e6a9.png)

**Chatterino version**
Chatterino Nightly 2.1.7 (commit 214afc8fc, 25.07.2020)

**Operating system**
Windows 10 Pro, Version 1903, OS Build 18362.959

cinemassacres:
set a timed ping or ask for a timed ping and it will happen https://streamable.com/kwxuqw

alazymeme:
I had this issue a bit ago and still have the issue: https://github.com/Chatterino/chatterino2/issues/1302

Felanbird:
I do not use the Chatterino extension.

alazymeme:
As I still have the issue, I'm definitely sure that it's not an extension issue

Felanbird:
I was just making sure as cinemassacres reproduction has chatterino extension, and you linked to an issue that stated: 

> **To Reproduce**
> 
> 1. Install chatterino extension

Felanbird:
had the issue again, but this time a manually closing of chatterino was required to close the connection

* ```# Start logging at 2020-08-16 00:04:54 Eastern Daylight Time```
* closed around ```[04:46:35] EST```
* logging continues until ```[16:55:35] EST``` manual restart of chatterino

Felanbird:
https://i.nuuls.com/JqpcS.gif | same as cinemassacres but without chatterino extension, and @5E7EN as the issue holder

PR #3189, opened 2021-08-23
Added /raw command

zneix:
Pull request checklist:

# Description

Added `/raw` command, it makes it possible to send raw data to write connection and it's pretty much reimplementation of the same connection from `IrcCommands.cpp`, but for the Twitch channels.  
I know that this command will not be useful to most likely over 99% of Chatterino users, but believe me that there're usecases and it won't hurt to have it even if there're not many / none.

A few good uses I good think of myself (though most of them are nerdy. maybe because only nerds are going to use these for their weird experiments):

- Brought up by @mm2pl: `/raw @reply-parent-msg-id={msg-id} PRIVMSG #pajlada :Please don't backseat :)` Could be added as a custom command and used in moderation buttons to reply to users directly. The `#pajlada` bit could be replaced with `#{channel}` once #3155 gets merged in, making this usecase more useful in general.
- Setting custom `client-nonce` value with a custom command.
- Various tests against TMI and easier debugging that won't require you to connect to IRC with netcat or openssl from terminal and save you some time, e.g. double spaces, no leading colon in front of the message, basically skipping all custom Chatterino's "message baking" and sending your raw data, as the command name suggests.
- JOINing / PARTing channels on demand. Now, that's something you'd normally not want to do by yourself, but forcing `ROOMSTATE` / `USERSTATE` without sending actual message and not needing to reopen chat which could break your split layout could be a nice thing to have. Possibly retriggering recent-messages (? has to be tested).
- Even more uses that number 1 IRC people and TMI experts could come up with...

In before someone comments: Again, this is not something that's targetting (a vast) majority of Chatterino users but rather a very very niche elite of them that know what they're doing.


Felanbird:
> * JOINing / PARTing channels on demand.

This would probably be a workaround for #1855. I'm unsure if this still happens but being able to force part the channel would fix said issue.

pajlada (approved):

PR #3206, opened 2021-08-28
Make Chatterino colorize usernames on IRC

Mm2PL:
Pull request checklist:

# Description

This was originally made for Dankerino. This allows you to have colorized names on IRC but not Twitch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Minor: Added the ability to open an entire tab as a popup. (#3082)
 - Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
 - Minor: Added regex option to Nicknames. (#3146)
+- Minor: Split colorizing usernames on IRC and Twitch into two settings, originally made for Mm2PL/dankerino (#tbd)
 - Bugfix: Fixed colored usernames sometimes not working. (#3170)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Minor: Added the ability to open an entire tab as a popup. (#3082)
 - Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
 - Minor: Added regex option to Nicknames. (#3146)
-- Minor: Split colorizing usernames on IRC and Twitch into two settings, originally made for Mm2PL/dankerino (#tbd)
+- Minor: Split colorizing usernames on IRC and Twitch into two settings, originally made for Mm2PL/dankerino (#3206)
 - Bugfix: Fixed colored usernames sometimes not working. (#3170)
 - Bugfix: Restored ability to send duplicate `/me` messages. (#3166)
 - Bugfix: Notifications for moderators about other moderators deleting messages can now be disabled. (#3121)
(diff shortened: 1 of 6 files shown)


ALazyMeme:
Upon second thoughts, I can't think of any reasons **not** to have this on. What are some thoughts of removing the irc setting entirely and just having it on all the time?

I understand the whole thing of "give users the option", but unlike the current twitch implementation (basically seeing which users are new), there's no specific reason to have the colours turned off.

@zneix @pajlada @Felanbird @TETYYS

Felanbird:
> What are some thoughts of removing the irc setting entirely and just having it on all the time?

Honestly I'm more surprised that with the recent IRC usercard changes, we didn't just accidentally get perfect twitch color sync.

I wouldn't mind @apa420's opinion on this, since as far as I know he's probably the only other person who uses experimental IRC, personally I don't care what approach we take.

apa420:
I don't really use the irc in chatterino that much, I basically have my own chatroom open there that nobody uses.
I haven't tried this pr out, but does the historical messages you get sent by the irc server also get colorized the same way?

I'm unsure if a setting is needed for this change since there's basically no downside to it considering you can't make all names white in normal twitchchat in chatterino either.
It would make chatterino a bit more consistent in coloring.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Minor: Added the ability to open an entire tab as a popup. (#3082)
 - Minor: Added optional parameter to /usercard command for opening a usercard in a different channel context. (#3172)
 - Minor: Added regex option to Nicknames. (#3146)
+- Minor: Split colorizing usernames on IRC and Twitch into two settings, originally made for Mm2PL/dankerino (#3206)
```
Mm2PL, line 15:
```suggestion
- Minor: Colorizing usernames on IRC, originally made for Mm2PL/dankerino (#3206)
```

src/providers/irc/IrcMessageBuilder.cpp
```diff
@@ -36,6 +37,14 @@ MessagePtr IrcMessageBuilder::build()
...
     // PARSE
     this->parse();
+    if (getSettings()->colorizeNicknamesOnIrc)
+    {
+        this->usernameColor_ = getRandomColor(this->ircMessage->nick());
+    }
+    else
+    {
+        this->usernameColor_ = getApp()->themes->messages.textColors.system;
+    }
```
Mm2PL, line 47:
```suggestion
    this->usernameColor_ = getRandomColor(this->ircMessage->nick());
```

src/singletons/Settings.hpp
```diff
@@ -100,6 +100,9 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/appearance/messages/usernameDisplayMode",
         UsernameDisplayMode::UsernameAndLocalizedName};
 
+    BoolSetting colorizeNicknamesOnIrc = {
+        "/appearance/messages/colorizeNicknamesOnIrc", true};
+
```
Mm2PL, line 105:
```suggestion
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -664,6 +664,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                        s.enableExperimentalIrc);
     layout.addCheckbox("Show unhandled IRC messages",
                        s.showUnhandledIrcMessages);
+    layout.addCheckbox("Colorize usernames on IRC", s.colorizeNicknamesOnIrc);
```
Mm2PL, line 667:
```suggestion
```

Mm2PL (commented):
stupidest way to commit, let's go

Felanbird:
> does the historical messages you get sent by the irc server also get colorized the same way?

yes
> I'm unsure if a setting is needed for this change since there's basically no downside to it considering you can't make all names white in normal twitch chat in chatterino either.
> It would make chatterino a bit more consistent in coloring.

👍

apa420 (approved):

pajlada (approved):

PR #3211, opened 2021-08-30
Add an option to disable the system message displayed on restarting after a crash

LosFarmosCTL:
Pull request checklist:

# Description

This PR adds a simple checkbox in the settings to disable the system message that is being displayed on restarting after a crash.

The message really annoys me in inactive chats (I have my splits set up pretty small, meaning that the message takes away a considerable amount of space), so right now I always restart chatterino again manually to get rid of it, this fixes that issue.


LosFarmosCTL:
I still want chatterino to automatically restart when I am i.e. not at my PC at the moment, I just don't need the message telling me that it restarted

It's a really small thing, but I know a few people that are annoyed by it, so I thought I might as well make a PR out of it.

apa420:
This will cause more confusion than do good. I'm against this change.

LosFarmosCTL:
> Someone might accidentally untick it and then it crashes and restarts and they have no idea why.

Well, i personally do not really care if it's merged in, I can just as well run my own chatterino version with the message disabled locally, I do know some people that have the same issue as me though, just thought it would be nice to have the option for everyone, so I opened it as a PR as well.

Tbh I doubt that anyone would accidentally untick that checkbox somewhere in miscellaneous though, and a lot of the other options there would probably be a lot more confusing if changed unknowingly.


> With that said, this pr is not complete. You should like the nuuls uploader thing and whatever have a checkbox with "Send this message on crash" (or something) on the actual popup.

wdym with that, there is no popup on restarting, only the system message?

LosFarmosCTL:
But there is no popup window displayed on restarting, the only thing is this system message that is being sent into every open split?

chatterino 08-30 21 46 05

apa420:
Oh, then why do you care about a system message? A popup I would understand would be annoying, but not really a system message.

LosFarmosCTL:
Well, a popup would honestly be less annoying to me, you just click "ok" and it's gone.

I have something like 15 splits on my main chatterino tab, most of them semi-active offline chats, and the message appearing in every single one of them completely clutters my main window, for some splits that message is basically the only thing i can still see.

I do have to agree that it is definitely a specific problem, I wouldn't have opened a PR for it if I didnt know some other people that had the same issue as me.

LosFarmosCTL:
Reflecting a bit and asking some more people, I have decided that this is in fact a change not useful enough for most people and I am myself against merging it now, so I will close this PR.

PR #3219, opened 2021-09-03
Raise ratelimit cooldown for JOIN messages

zneix:
Pull request checklist:

# Description

This basically ensures we will actually join all the channels just fine - in some rare cases, even with ratelimitting logic, our JOINs can get lost in a limbo.

Even though we use the "18 JOINs / 10.5s" system now, sometimes sending messages to IRC can hit a small lagspike and (as quoted by @Felanbird, who also keeps getting this issue): "if the lag ends up being over .5s then eventually it will catch up to you and you'll be joining slightly faster than 10s and Twitch say NO". I got this issue a couple times myself as well.


Mm2PL (approved):
Let's hope this fixes issues with server-side lag spikes.

Felanbird (approved):
I was the one testing building at `12500` and I never had any blocks drop.

Only time I had the issue was I was building master's`10500`, so I'm fairly confident this will work :+1:

pajlada (approved):

PR #3222, opened 2021-09-04
added support for watching streams

fourtf:
Pull request checklist:

# Description

Allows watching streams from Chatterino. Only works on windows since it uses webview2.
image
image

## Testing it

1. Be on Windows.
2. Install [webview2](https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section) *or* MS Edge Beta
3. Download StreamView. You can use the prebuilt one from here: [fourtf.com/StreamView-0.zip](https://fourtf.com/StreamView-0.zip)
4. Put it into the Chatterino folder so the `StreamView` folder is on the same level as `chatterino.exe`.
  image
5. Run Chatterino or StreamView/StreamView.exe

### DIFF
--- a/src/widgets/FramelessEmbedWindow.cpp
+++ b/src/widgets/FramelessEmbedWindow.cpp
@@ -15,6 +15,24 @@
 
 namespace chatterino {
 
+namespace {
+#ifdef USEWINSDK
+    void sendToStreamView(void *handle, QJsonDocument doc)
+    {
+        auto json = doc.toJson();
+        json.append('\0');
+
+        COPYDATASTRUCT cds;
+        cds.cbData = static_cast<DWORD>(json.size());
+        cds.lpData = json.data();
+
+        ::SendMessage((HWND)handle, WM_COPYDATA,
+                      reinterpret_cast<WPARAM>(handle),
+                      reinterpret_cast<LPARAM>(&cds));
+    }
+#endif
+}  // namespace
+
 FramelessEmbedWindow::FramelessEmbedWindow()
     : BaseWindow(BaseWindow::Frameless)
 {
@@ -52,9 +70,22 @@ bool FramelessEmbedWindow::nativeEvent(const QByteArray &eventType,
                     getApp()->twitch2->getOrAddChannel(channelName));
             }
         }
+        else if (root.value("type").toString() == "request-show")
+        {
+            QJsonDocument doc{QJsonObject{{"type", "shown"}}};
+
+            auto callback = root.value("callback-handle");
+            sendToStreamView(
+                reinterpret_cast<void *>(callback.toString().toLongLong()),
+                doc);
+
+            this->hostWindowId = callback.toString();
+
+            this->show();
+        }
     }
 
-    return BaseWidget::nativeEvent(eventType, message, result);
+    return BaseWindow::nativeEvent(eventType, message, result);
 }
 
 void FramelessEmbedWindow::showEvent(QShowEvent *)
@@ -89,6 +120,16 @@ void FramelessEmbedWindow::showEvent(QShowEvent *)
 
         ::SendMessage(parentHwnd, WM_COPYDATA, reinterpret_cast<WPARAM>(handle),
                       reinterpret_cast<LPARAM>(&cds));
+    }
+}
+
+void FramelessEmbedWindow::scaleChangedEvent(float)
+{
+    if (!this->hostWindowId.isEmpty())
+    {
+        QJsonDocument doc{QJsonObject{{"type", "fix-bounds"}}};
+        sendToStreamView(
+            reinterpret_cast<void *>(this->hostWindowId.toLongLong()), doc);
     }
 }
 #endif
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Allow watching twitch streams with Chatterino StreamView (Windows only)
 - Minor: Remove TwitchEmotes.com attribution and the open/copy options when right-clicking a Twitch Emote. (#2214, #3136)
 - Minor: Strip leading @ and trailing , from username in /user and /usercard commands. (#3143)
 - Minor: Display a system message when reloading subscription emotes to match BTTV/FFZ behavior (#3135)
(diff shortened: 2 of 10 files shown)


caneleex:
the chat embedding section just goes gray after a while of watching (max 30 seconds), without chatterino crashing or basically doing anything at all. I have no idea how to debug this, so debugging steps would be appreciated

fourtf:
> 
> 
> the chat embedding section just goes gray after a while of watching (max 30 seconds), without chatterino crashing or basically doing anything at all. I have no idea how to debug this, so debugging steps would be appreciated
> 
> 

There is a mechanism to close the chatterino embed after 30s if it doesn't have a parent window, so that may be going wrong. That should be the issue if it's precisely 30s every time. Whats your OS information?

Void48:
> the chat embedding section just goes gray after a while of watching (max 30 seconds), without chatterino crashing or basically doing anything at all. I have no idea how to debug this, so debugging steps would be appreciated
> 
> 

Same for me LULE

caneleex:
> 
> 
> > the chat embedding section just goes gray after a while of watching (max 30 seconds), without chatterino crashing or basically doing anything at all. I have no idea how to debug this, so debugging steps would be appreciated
> > 
> 
> There is a mechanism to close the chatterino embed after 30s if it doesn't have a parent window, so that may be going wrong. That should be the issue if it's precisely 30s every time. Which version of windows are you using, and are you using any special settings?

it seems to be around 30 seconds, yeah. I use Windows 10 19043.1202, and no, I don't have any special settings

fourtf:
> 
> 
> > > the chat embedding section just goes gray after a while of watching (max 30 seconds), without chatterino crashing or basically doing anything at all. I have no idea how to debug this, so debugging steps would be appreciated
> > > 
> > 
> > 
> > There is a mechanism to close the chatterino embed after 30s if it doesn't have a parent window, so that may be going wrong. That should be the issue if it's precisely 30s every time. Which version of windows are you using, and are you using any special settings?
> 
> it seems to be around 30 seconds, yeah. I use Windows 10 19043.1202, and no, I don't have any special settings

Actually that's my fault, I was using the stable Chatterino when testing it. I'll fix it asap.

Void48:
When I open the player my chatterino gets really big/small and blurry Screenshot 2021-09-04 200425
smallchatterino

caneleex:
> 
> 
> I also noticed that when I open the player my Chatterino gets really big/small and blurry Screenshot 2021-09-04 200425
> smallchatterino
> 
> edit: ok I noticed it changed the zoom level in settings for some reason
> 1440p screen btw. apu

locate the StreamView executable -> right click, Properties -> Compatibility -> Change high DPI settings -> check both boxes (Use this setting..., Override high DPI...)

Void48:
> > I also noticed that when I open the player my Chatterino gets really big/small and blurry Screenshot 2021-09-04 200425
> > smallchatterino
> > edit: ok I noticed it changed the zoom level in settings for some reason
> > 1440p screen btw. apu
> 
> locate the StreamView executable -> right click, Properties -> Compatibility -> Change high DPI settings -> check both boxes (Use this setting..., Override high DPI...)

Just did that and Chatterino still gets big/small (blurry) and the Chatterino StreamView is blurry too

fourtf:
> 
> 
> > I also noticed that when I open the player my Chatterino gets really big/small and blurry Screenshot 2021-09-04 200425
> > smallchatterino
> > edit: ok I noticed it changed the zoom level in settings for some reason
> > 1440p screen btw. apu
> 
> locate the StreamView executable -> right click, Properties -> Compatibility -> Change high DPI settings -> check both boxes (Use this setting..., Override high DPI...)

Workarounds like this have no business here since they don't help getting the feature out at all.

fourtf:
> 
> 
> I also noticed that when I open the player my Chatterino gets really big/small and blurry Screenshot 2021-09-04 200425
> smallchatterino
> 
> edit: ok I noticed it changed the zoom level in settings for some reason
> 1440p screen btw. apu

I can reproduce this on all the builds from the CI. It didn't happen to me while developing it but I think I know how to fix it.

ALazyMeme:
Not sure if it's possible to hide the "Expand Chat" button. I assume not:

image

src/widgets/FramelessEmbedWindow.cpp
```diff
@@ -15,6 +15,24 @@
...
+    {
+        auto json = doc.toJson();
+        json.append('\0');
+
+        COPYDATASTRUCT cds;
+        cds.cbData = static_cast<DWORD>(json.size());
+        cds.lpData = json.data();
+
+        ::SendMessage((HWND)handle, WM_COPYDATA,
+                      reinterpret_cast<WPARAM>(handle),
```
TETYYS, line 30: This should be handle of the window that is passing the data, as per winapi docs

ALazyMeme:
In addition to the above, when running the StreamView exe directly, it appears to bug out when you switch channels. Then when I try to type in those bugged channels, the chatbox appear to just move down a little more. As shown in the screenshot, it seems this is now permanently bugged after testing the first time:

https://user-images.githubusercontent.com/12804673/132110545-16a2cc7a-a7a8-4f48-8a59-da3ba0ea91ec.mp4

src/widgets/FramelessEmbedWindow.cpp
fourtf: Good catch

fourtf:
> 
> 
> Not sure if it's possible to hide the "Expand Chat" button. I assume not:
> 
> image

I feel like the user would still want the normal chat so he can do stuff like sending the sub renew message. Also messing with the website is just a giant headache that's not worth getting into.

goldbattle:
I tried linking to the [webview/webview](https://github.com/webview/webview) directly.
Which worked on windows, but on linux I ran into issues with GTK conflicting with the QT `signal` .
After [fixing](https://github.com/goldbattle/chatterino2/commit/f61fd9e701346bf38cb2f43e1dcdc0fd9075f790#diff-9fef35ddb7dc849fedc217ee5f36f5b8d1ee939928dc19f3c89a7d69f3445f66R5-R7) this I ran into issues with player being not supported by the twitch player (same one as below).
https://github.com/goldbattle/chatterino2/tree/test-webview-webview

I also experimented with the QT `webenginewidgets` / `qtwebengine`, but this has issues both on windows and other platforms.
It does make compiling easier, but I don't think video decoding is supported unless you build it from scratch with the correct encoders (I didn't look into this as I already was very deep down the rabit hole).
https://github.com/goldbattle/chatterino2/tree/test-webview-linking
image

PR #3231, opened 2021-09-09
Add `flags.reward_message` filter variable

sando:
Pull request checklist:

# Description

Adds `flags.reward_message` variable for filtering on channel point reward messages.


ALazyMeme:
Please fix formatting and add a new filter on the wiki: https://github.com/Chatterino/wiki

Mm2PL:
How does this functionality differ from points_redeemed?

sando:
> How does this functionality differ from points_redeemed?

`flags.points_redeemed` filters on messages highlighted with channel points, whereas `flags.reward_message` filters only on reward messages - each has a different `MessageFlag`. I did consider whether the two should just be combined but there's probably a valid use-case for wanting one but not the other.

pajlada (approved):

PR #3261, opened 2021-09-30
Add chatter count to viewer list popout

### LINKED ISSUE #2993: Viewerlist total number - Feature Request
**Feature Request**:

Please add a text in the viewer list with the total number.
I think at the top right, next to the search bar is the best place.
This can and should then be shortened from a certain number of viewers, e.g. 1K instead of 1000.

MrAuro:
Pull request checklist:

# Description

Closes #2993

Adds the chatter count to the viewer list popout
example

### DIFF
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -743,6 +743,11 @@ void Split::showViewerList()
             auto obj = result.parseJson();
             QJsonObject chattersObj = obj.value("chatters").toObject();
 
+            viewerDock->setWindowTitle(
+                "Viewer List - " + this->getChannel()->getName() + " (" +
+                QString::number(obj.value("chatter_count").toInt()) +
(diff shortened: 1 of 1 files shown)


MrAuro:
Also could this get the hacktoberfest-accepted label

zneix (changes requested):
I'm wondering if it's possible to not duplicate the viewer list's dock's title, other than that, good Qol addition :+1:

Mm2PL:
IMO you should also format the number to use short notation, ie 1k instead of 1000. There is already a helper function for this.

MrAuro:
> 
> 
> IMO you should also format the number to use short notation, ie 1k instead of 1000. There is already a helper function for this.

Using the existing helper function rounded the chatter count too much for my liking, however in the latest commit I pushed I localized the chatter count to match the /chatters command

PR #3265, opened 2021-10-02
Add keybindings to tab context menu items

talneoran:
Pull request checklist:

# Description

When the "Popup Tab" entry was added to the tab context menu we added it with the shortcut for it. For consistency with that and with the split context menu it makes sense to add the keybinds for other relevant entries. This also increases user awareness of useful shortcuts.

### DIFF
--- a/src/widgets/helper/NotebookTab.cpp
+++ b/src/widgets/helper/NotebookTab.cpp
@@ -60,9 +60,12 @@ NotebookTab::NotebookTab(Notebook *notebook)
         this->showRenameDialog();
     });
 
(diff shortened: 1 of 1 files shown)


zneix:
Thanks for noticing these! Since you're adding these labels already, I remember that when I've added toggling visibility of tabs in #2600 I remember there's actually 2 context menu's handling right-click - one for `NotebookTab.cpp` and another for `Notebook.cpp`:

fdm zneix forgor 💀

You can take care of that one as well and I'll merge this one in!

PR #3268, opened 2021-10-02
Restart the application bundle instead of the underlying executable when crashing on macOS

LosFarmosCTL:
Pull request checklist:

# Description

On macOS, programs are contained in ".app" application bundles. When chatterino restarts on a crash, it currently just looks for the executable path obtained via `QApplication::applicationFilePath()` and starts that again, which for macOS is the underlying unix executable inside of the application bundle. Unfortunately, for macOS those 2 are absolutely different, resulting in i.e. a second chatterino icon being added to the dock, due to the OS not recognizing that it is the same application.

This PR changes the restart logic for macOS to use Apples CoreFoundation API to get the path of the application bundle, and open this using the built-in "open" command. For every other OS, just starting the executable directly is, at least afaik, perfectly fine.


pajlada (changes requested):
Please add a **Bugfix:** changelog entry

pajlada (approved):

PR #3282, opened 2021-10-10
Add new search predicate to enable searching for messages matching a regex

LosFarmosCTL:
Pull request checklist:

# Description

This PR adds a new regex predicate to the search dialog, to enable filtering for messages matching a specific regex.

It also introduces the option to surround a predicates value in quotation marks, to enable predicate values that contain spaces. (The old predicate functionality was kept intact, so everything that worked before still works exactly the same, just some new options)

for example: `regex:"^Kappa [0-9]{1,3}$"`


LosFarmosCTL:
> This will require a wiki entry

Oh yeah true, will open a wiki PR once I get home from work

PR #3285, opened 2021-10-15
Simplify `ResizingTextEdit::isFirstWord` to avoid invalid QString index warnings

### LINKED ISSUE #3234: QCharRef with an index pointing outside the valid range of a QString warning
**Describe your issue**
Upon pressing Tab in Chatterino, a following warning is generated:
`[13:57:35.344] unknown Using QCharRef with an index pointing outside the valid range of a QString. The corresponding behavior is deprecated, and will be changed in a future version of Qt.`

Pretty sure it's actually `CompletionModel::refresh` generating it, since pressing Tab only after text in text edit has changed generates said warning.
It has been around for literally months if not years and it's high time we fix this, because in Qt6 something has changed about it and it's something that crashes on my `zneix/feature/qt6` branch.

**Screenshots**
N/A

**OS and Chatterino Version**
Arch GNU/Linux, master

zneix:
Pull request checklist:

# Description

As pointed out in https://github.com/Chatterino/chatterino2/issues/3234#issuecomment-943734055 I've localized the issue and fixed it.
Turns out that our complicated logic with iterating over `QTextEdit`'s plainText and checking whether it contains space or not can be just replaced with a simple [`QString::contains`](https://doc.qt.io/qt-5/qstring.html#contains-1) which does exactly the same thing as our function did before.
While this might be a bit silly logic for detecting a first word, this is something to think of in the future, not something I'm focusing on approaching in this PR.

I've confirmed it Fixes #3234

### DIFF
--- a/src/widgets/helper/ResizingTextEdit.cpp
+++ b/src/widgets/helper/ResizingTextEdit.cpp
@@ -39,10 +39,11 @@ bool ResizingTextEdit::hasHeightForWidth() const
     return true;
 }
 
-// XXX(zneix): This logic might be a bit stupid, see if it can be improved
 bool ResizingTextEdit::isFirstWord() const
 {
-    return !this->toPlainText().contains(' ');
+    QString plainText = this->toPlainText();
+    QString portionBeforeCursor = plainText.left(this->textCursor().position());
+    return !portionBeforeCursor.contains(' ');
 };
 
 int ResizingTextEdit::heightForWidth(int) const


zneix:
Considering this warning/bug was both quite annoying, noticed by quite some ppl and also breaking for Qt6 builds I've decided to put a changelog entry for it.

pajlada:
Initial fix didn't account for the feature where if you move your cursor to the start an already started message: `<CURSOR> hello!!` and type `@pajl` and press tab, it would include a comma since it was autocompleting as the first word.

The commit I added brings that feature back while keeping the improved (and much simpler) check

Example variable values:
Input box: `Hello @pajl<CURSOR>`
plainText: `Hello @pajl`
portionBeforeCursor: `Hello @pajl`
Autocompleting will insert `@pajlada `

Input box: `Hello @pajl<CURSOR>xD`
plainText: `Hello @pajlxD`
portionBeforeCursor: `Hello @pajl`
Autocompleting will insert `@pajlada `

Input box: `@pajl<CURSOR> Hello xD`
plainText: `@pajl Hello xD`
portionBeforeCursor: `@pajl`
Autocompleting will insert `@pajlada, `

zneix:
Thanks so much for noticing my oversight and making a comment covering it :+1:

PR #3296, opened 2021-10-19
Removed duplicate setting for toggling `Channel Point Redeemed Message` highlights

Felanbird:
Pull request checklist:

# Description

While trying to figure out the issue from `#3293`, I added a new general setting to toggle First Message Highlights, and because of this noticed that this setting existed, despite already being handled in `HighlightModel.cpp`


Felanbird:
relevant code that already handled it
https://github.com/Chatterino/chatterino2/blob/17f98cb2787b12f9667c067f8f3678535047b345/src/controllers/highlights/HighlightModel.cpp#L200-L204

PR #3327, opened 2021-10-30
Add GitHub action to test builds with precompiled headers disabled

### LINKED ISSUE #3320: Add GH action building without precompiled headers
That would be something on ubuntu-latest, just latest cmake I suppose. Would prevent scenarios like in #3308, making sure that contributors, who don't compile without precompiled headers won't make it impossible to compile w/o PCH.

pajlada:
Not worth adding 12 extra builds for this

Pull request checklist:

# Description

Fixes #3320

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,14 +15,21 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-latest]
         qt-version: [5.15.2, 5.12.10]
         build-system: [qmake, cmake]
+        pch: [true]
         exclude:
           - os: windows-latest
             qt-version: 5.12.10
             build-system: cmake
+            pch: true
         include:
           - os: windows-2016
             qt-version: 5.12.10
             build-system: cmake
+            pch: true
+          - os: ubuntu-latest
+            qt-version: 5.15.2
+            build-system: cmake
+            pch: false
       fail-fast: false
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -158,7 +165,12 @@ jobs:
         run: |
             mkdir build
             cd build
-            cmake -DCMAKE_INSTALL_PREFIX=appdir/usr/ -DCMAKE_BUILD_TYPE=Release -DPAJLADA_SETTINGS_USE_BOOST_FILESYSTEM=On ..
+            cmake \
+              -DCMAKE_INSTALL_PREFIX=appdir/usr/ \
+              -DCMAKE_BUILD_TYPE=Release \
+              -DPAJLADA_SETTINGS_USE_BOOST_FILESYSTEM=On \
+              -DUSE_PRECOMPILED_HEADERS=${{ matrix.pch }} \
```
Mm2PL, line 172: These work with `true` and `false`, not only `On` or `Off`, right?

pajlada: Yes, verified by adding the message with PCH status

See https://github.com/Chatterino/chatterino2/runs/4054807765?check_suite_focus=true#step:17:63 and https://github.com/Chatterino/chatterino2/runs/4054807993?check_suite_focus=true#step:17:63

PR #3328, opened 2021-10-30
Fix the reconnection backoff accidentally resetting when thrown out of certain IRC servers

### LINKED ISSUE #3313: [IRC] Chatterino keeps trying to reconnect every 1 second
**Describe your issue**
Chatterino keeps reconnecting to IRC servers every one second in case of failure (?). This is problematic because it triggers rate limiting on remote servers. It should reconnect slower or with a exponential backoff

**OS and Chatterino Version**
Windows 10; 2.3.4

pajlada:
- Only reset reconnect backoff when we can confirm a successful connection was made
- Add changelog entry

Pull request checklist:

# Description

Fixes #3313


Felanbird (commented):
As a note, this caps out at 8 second [1->2->4->8] reconnect time. We discussed 16 seconds on stream but I'm unsure where the limitation is set.

Mm2PL:
@Felanbird it's set right here :) https://github.com/Chatterino/chatterino2/blob/fix/dont-reset-reconnection-backoff/src/providers/irc/IrcConnection2.hpp#L34

Felanbird:
> 
> 
> @Felanbird it's set right here :) https://github.com/Chatterino/chatterino2/blob/fix/dont-reset-reconnection-backoff/src/providers/irc/IrcConnection2.hpp#L34

👍 - I misunderstood the purpose of the `<4>` when playing with it myself

pajlada:
Looks good to me
image

PR #3329, opened 2021-10-30
Track timeout/slowmode status

Bun:
Pull request checklist:

# Description

This PR adds a timer to the chat input box (on the left side of the message length counter) that reflects how long until you can chat again. It supports timeouts (on CLEARCHAT and on error notice) and slowmode (including the error notice), but doesn't query e.g. an existing timeout on application restart.

Some of the choices were made in good faith but may be a bit arbitrary, so please nitpick away and let me know if anything should be changed:
* The `color: gold` text for the timer label
* Minor refactor/cleanup of how notices are handled (nb: the PR makes sure to keep logic with side-effects in the `handle*` function instead of `parse*`, which does mean parsing the notice twice)
* Technically, if the timer is between 1s and 0s Twitch will still block you from sending, ~but at <1s the timer label goes away~. Let me know if this is something that should be "fixed".

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -221,12 +221,14 @@ bool TwitchChannel::isEmpty() const
 void TwitchChannel::resyncChatAgain()
 {
     auto now = std::chrono::steady_clock::now();
-    long seconds = 0;
+    int seconds = 0;
     if (this->timeoutEnds_.has_value())
-        seconds = std::max(
+    {
+        seconds = std::max<int>(
             seconds, (*this->timeoutEnds_ - now) / std::chrono::seconds(1));
+    }
     if (this->slowedEnds_.has_value())
-        seconds = std::max(
+        seconds = std::max<int>(
             seconds, (*this->slowedEnds_ - now) / std::chrono::seconds(1));
     if (seconds <= 0)
     {
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -193,6 +193,11 @@ TwitchChannel::TwitchChannel(const QString &name)
     });
     this->liveStatusTimer_.start(60 * 1000);
 
+    // timeout countdown message
+    QObject::connect(&this->timeoutCounter_, &QTimer::timeout, [this] {
+        this->resyncTimedOut();
+    });
+
     // debugging
 #if 0
     for (int i = 0; i < 1000; i++) {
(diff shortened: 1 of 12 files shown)


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -408,6 +495,18 @@ void TwitchChannel::sendMessage(const QString &message)
     {
         qCDebug(chatterinoTwitch) << "sent";
         this->lastSentMessage_ = parsedMessage;
+        if (!this->hasHighRateLimit())
+        {
+            auto roomModes = *this->accessRoomModes();
+            if (roomModes.slowMode > 0)
+            {
+                // XXX: this'll (incorrectly) reset the countdown if we still
+                // have time remaining if the user was too eager to send a
```
Felanbird, line 504:
```suggestion
                // have time remaining, if the user was too eager to send a
```
also thought about replacing if with because but 🤷

Felanbird (changes requested):
Just a few grammar things, the functionality itself seems to work well.

`note:` I'm not crazy about the setting name but I don't have another suggestion right now.

Felanbird:
image

I found the edge case, but it's probably not that important. 

When you attempt to send a message in a channel with verified chat restrictions (phone number in this example [#bobross]), the timer will start and refresh on each message, although the slowmode restriction is likely not applying.

I don't know if we should handle these messages or just ignore this "issue".

Bun:
>  Being a subscriber does not bypass the slow-mode timer, despite subs being immune to slow-mode.

Whether subs are affected is actually a setting for affiliate+ , in Dashboard -> Settings -> Affiliate: "Allow subscribers to chat freely in your channel when chat is in slow mode" and I believe it's disabled by default. Regardless, there doesn't seem to be a public way to query this. We could err on the safe side I guess and only show the countdown if you end up sending too fast.

> Messages that are sent in channels where you don't meet the verified chat restrictions (phone number etc.. - [#bobross as example]), trigger the timer despite the fact that no message was sent. (Can be handled as you receive back a message).

I considered showing the timer upon receiving USERSTATE instead of after sending the message, but thought it would add extra (UI) latency so decided against it. It does however feel more correct, especially considering things like that^.

> Messages that are sent in hyper-fast chats trigger the timer, despite the fact that no message was sent. (Probably not handle-able as you do not receive a message when these are dropped).

You tested that I assume? For example, for the Twitch-side duplicate message check it doesn't matter if your message was actually shown in chat or not, so I would assume it works like that for slowmode as well. (?)

LosFarmosCTL:
> I believe it's disabled by default.

Yes, it is 100% disabled by default.

Felanbird:
> it's disabled by default. 

I forgot about this fact.

> I considered showing the timer upon receiving USERSTATE instead of after sending the message, but thought it would add extra (UI) latency so decided against it. It does however feel more correct, especially considering things like that^.

I'm indifferent about this one anyway, we are probably better off letting people think said message attempt triggered slow-mode anyway.

> You tested that I assume? For example, for the Twitch-side duplicate message check it doesn't matter if your message was actually shown in chat or not, so I would assume it works like that for slowmode as well. (?)

You're likely to be correct about this one as well, I'll re-test later today when somebody else with a hyper-fast chat goes live.

NS2555:
as the only light/white mode user of chatterino, please ensure the timer has enough contrast in chatterino light/white modes, because judging by the fact it uses a bright yellow color which is not a normal chatterino color, it is going to end up being almost completely invisible

Felanbird:
> 
> 
> as the only light/white mode user of chatterino, please ensure the timer has enough contrast in chatterino light/white modes, because judging by the fact it uses a bright yellow color which is not a normal chatterino color, it is going to end up being almost completely invisible

This is a good point
white
image
light
image

NS2555:
i suggest using a different color hue all together, i tested some things here and there and found out `#00526D` works pretty well 
image

pajlada:
In general, functionality looks good. There's some further code review/refactoring that I'd like to see but will leave it for tomorrow or next week.
Will need a little bit more testing with edge-cases but they're pretty well documented here or in the code already which is nice.

pajlada:
I've let this become stale for long enough that it's not something I'm comfortable implementing right now, given the turbulence with chat over EventSub chat & Helix message sending

PR #3332, opened 2021-10-31
Update contributors guide

ALazyMeme:
Pull request checklist:

# Description

I noticed that the avatars URL has changed + new users have a jfif avatar format (as demonstrated [here](https://avatars.githubusercontent.com/1xelerate)).

### DIFF
--- a/resources/contributors.txt
+++ b/resources/contributors.txt
@@ -6,7 +6,7 @@
 # Name | Link | Avatar (Loaded as a resource) | Title (description of work done).
 
 # Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://avatars.githubusercontent.com/username?size=128).
-# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng (if the file is in png format) and run `./generate_resources.py`.
+# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng (if the file is in png format) and run ./generate_resources.py
 # Contributor is what we use for someone who has contributed in general (like sent a programming-related PR).
 
 fourtf | https://fourtf.com | :/avatars/fourtf.png | Author, main developer
--- a/resources/contributors.txt
+++ b/resources/contributors.txt
@@ -5,7 +5,7 @@
 # Adding yourself? Copy and paste this template at the bottom of this file and fill in the fields in a PR!
 # Name | Link | Avatar (Loaded as a resource) | Title (description of work done).
 
-# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://avatars.githubusercontent.com/username?size=128).
+# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://github.com/username.png?size=128).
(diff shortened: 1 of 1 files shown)


ALazyMeme:
I noticed the sizes for the jfif/jpg formats is smaller (naturally, since it's a more compressed image format). Should I perhaps update the guide to suggest using jpg instead of png?

resources/contributors.txt
```diff
@@ -5,9 +5,9 @@
 # Adding yourself? Copy and paste this template at the bottom of this file and fill in the fields in a PR!
 # Name | Link | Avatar (Loaded as a resource) | Title (description of work done).
 
-# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://github.com/username.png?size=128).
-# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng and run ./generate_resources.py
-# Contributor is what we use for someone who has contributed in general (like sent a programming-related PR)
+# Avatar should be located in avatars/ directory. It's size should be 128x128 (get it from https://avatars.githubusercontent.com/username?size=128).
```
jammehcow, line 8:
```suggestion
# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://avatars.githubusercontent.com/username?size=128).
```

"Its" is the possessive form of "it" - whereas "It's" is a contraction of "it is" :)

jammehcow (changes requested):

resources/contributors.txt
```diff
@@ -5,9 +5,9 @@
 # Adding yourself? Copy and paste this template at the bottom of this file and fill in the fields in a PR!
 # Name | Link | Avatar (Loaded as a resource) | Title (description of work done).
 
-# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://github.com/username.png?size=128).
-# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng and run ./generate_resources.py
-# Contributor is what we use for someone who has contributed in general (like sent a programming-related PR)
+# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://avatars.githubusercontent.com/username?size=128).
```
zneix, line 8: Why is this changed? I don't see a point in doing so, while old method still works and will continue to do so with a fancy redirect.

resources/contributors.txt
```diff
@@ -5,9 +5,9 @@
...
+# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng (if the file is in png format) and run `./generate_resources.py`.
```
pajlada, line 9:
```suggestion
# Make sure to reduce avatar's size as much as possible with tool like pngcrush or optipng (if the file is in png format) and run ./generate_resources.py
```

Removed the backticks as this is not a markdown file, and removed the dot to make it easier to copy the script name

resources/contributors.txt
pajlada:
```suggestion
# Avatar should be located in avatars/ directory. Its size should be 128x128 (get it from https://github.com/username.png?size=128).
```

PR #3379, opened 2021-12-02
Colorize entire split input header instead of just the channel name

LosFarmosCTL:
Pull request checklist:

# Description

Having a lot of splits open in one tab currently results in a really frustrating experience when trying to navigate between them using keyboard shortcuts, since the only way of telling which split is currently active is the small colorized channel name in the split header.

This PR adds the option to draw the entire split header in a blue tint instead of just the channel name text. Adding a new setting for this is imo a good idea here, since there is no need for, what I am guessing is probably the majority of chatterino users, those who mostly have a single split in a tab to have their entire header colorized.

# Screenshots

## White

chatterino 2021-12-03 at 00 47 21

## Light

chatterino 2021-12-03 at 00 47 02

## Dark

chatterino 2021-12-03 at 00 45 37

## Black

chatterino 2021-12-03 at 00 46 13


leon-richardt:
I advocate for making this standard behavior and **removing** the option to switch to the old behavior.

Mm2PL:
Should this really be an option? I'd rather just have the behavior changed.

jammehcow:
Agree with Leon and mm2; I don't see a reason to keep the old behaviour around. I would like a single split to not have a colour though, maybe I'm special.

LosFarmosCTL:
> I would like a single split to not have a colour though, maybe I'm special.

No, I feel the same way, its mainly why I made it a setting.

Maybe make it the default whenever there is more than one split currently opened in a tab?

Felanbird:
I agree with whichever consensus does not like this change, but I also understand the reasoning for it.

LosFarmosCTL:
I changed the behaviour to be on by default but only show the colorized header if there is more than one split in the containing tab, retaining the previous behaviour for single splits.

Mm2PL (approved):

zneix:
I don't know if I am testing this wrong or misunderstanding the Pull Request, however - when there's only 1 split in the tab, it keeps the old behaviour; as in, the split header is still as it was before

zneix:
Found another regression:

LosFarmosCTL:
> I don't know if I am testing this wrong or misunderstanding the Pull Request, however - when there's only 1 split in the tab, it keeps the old behaviour; as in, the split header is still as it was before

See my reply above, I agree with jammeh's take that coloring the entire header is unnecessary for a single split and more distracting than adding any additional value, so I added a check to retain the old behaviour for single splits.

> Found another regression:

That one is definitely not intended though, I'll take a look at that tomorrow.

pajlada:
I agree that there should not be a setting for this, if we wanted this user-customizable imo we would support it through custom themes (which isn't off the table).

I don't think we should change the behaviour depending on the amount of splits in a tab - if it looks off with just one split I think it means that the behaviour might need tweaking.

With this information, I set off to tweak some things, and came up with this solution: Just change the focused background color of the split, but leave the color roughly the same as it was before this PR.

Multi split
image

Single split
image

Single split window not focused
image

Light mode single split focused
image

Light mode multi split with focus
image

LosFarmosCTL:
Had absolutely no time over the last few days, but I integrated pajladas patch now.

zneix (approved):
I have tested all the changes, code looks good to me and I'm happy with the functionality it brings :+1: however if you want to make further changes later on, feel free to. Requested a review from pajlada, he can merge this once he reviews the PR.

LosFarmosCTL:
> however if you want to make further changes later on, feel free to.

Dont have anything further to add to this PR either.

pajlada (approved):

PR #3382, opened 2021-12-04
Fix crash on missing parameters with IRC /kick command

pajlada:
If the user tries to use /kick with less than the 2 required parameters (channel and client), we would crash.

Pull request checklist:

# Description


Felanbird (approved):
Functionality works as intended - `/kick` no longer crashes. 👍

PR #3394, opened 2021-12-07
Move '/live' logs to its own subDirectory

Felanbird:
Pull request checklist:

# Description

This PR moves to relocate `/live` logs to it's own directory, as it is a special channel.

This change as expected will cause users who use `/live` to have 2 different logging files, but from what I've seen, this channel is nowhere near as used as `/mentions` or `/whispers`, and coupling that with how many users do not log, this change shouldn't be too disruptive.

That being said, I've included said information in the changelog entry, and included a changelog entry for that reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Minor: Made join and part message have links to usercards. (#3358)
 - Minor: Show picked outcome in prediction badges. (#3357)
 - Minor: Add support for Emoji in IRC (#3354)
+- Minor: Moved `/live` logs to its own subdirectory. (Logs from before this change will still be available in `Channels -> live`). (#3393)
 - Bugfix: Fix Split Input hotkeys not being available when input is hidden (#3362)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - Minor: Made join and part message have links to usercards. (#3358)
 - Minor: Show picked outcome in prediction badges. (#3357)
 - Minor: Add support for Emoji in IRC (#3354)
+- Minor: Moved `/live` logs to its own subdirectory. (Logs from before this change will still be available in `Channels -> live`). (#3393)
```
Felanbird, line 40:
```suggestion
- Minor: Moved `/live` logs to its own subdirectory. (Logs from before this change will still be available in `Channels/live`). (#3393)
```
Going to put this changelog variation here since it's probably more standard, but I do think `->` is very useful when explaining file paths to users.

pajlada (approved):

PR #3397, opened 2021-12-10
Fix special chars in Windows usernames breaking the storage of commands.json

### LINKED ISSUE #866: Cannot load or save commands if path to chatterino2 directory contains scandinavian characters
commands.json is not loaded nor saved if directory path to chatterino2's config folder contains scandinavian letters. My username has the letter Ä in it and that seems to cause the problem.

Tried moving chatterino2 configs to it's own folder and the issue is gone.

LosFarmosCTL:
Pull request checklist:

# Description

This closes #866, apparently the changes to the settings library were not enough for windows. This issue is not reproducible on macOS for me (I don't know if it is on linux), but had a windows user who told me about it yesterday, and could reproduce it on my windows laptop.

I honestly have no Clue about why toStdString does not work, according to the Qt documentation it should convert it internally using QString.toUtf8() so there _should_ be no issue with something like a german ü, but there is. qPrintable uses toLocal8Bit() to create a byte array with the local 8 bit representation, which for whatever reason works on windows as well (I hate charsets). This is also the method already used for converting the filename QString for the normal settings.json, so I don't see anything this could break, since that is working fine everywhere.


LosFarmosCTL:
Ignore the weird merge commits, i am stupid (:

jammehcow:
Tested with an account called `DeveloperÄ` and commands saved fine!
Windows 10 21H1

https://user-images.githubusercontent.com/4962764/145654649-aa38aa80-9c35-460b-bcc6-9b076af83162.mp4

pajlada (approved):

PR #3403, opened 2021-12-13
Add clear button to settings search bar

acdvs:
Pull request checklist:

# Description

Adds an "x" icon to the right side of the settings search bar for easy unfiltering with a single click. When nothing is entered, the button darkens, and upon typing, the button brightens.

clear_search

I decided to give this a go because text inputs (no matter where) that auto-filter something as the user types typically make more sense UX-wise if they have a clear button to undo the filtering.


Mm2PL (approved):

kornes:
Qt has clearButtonEnabled for inputs, the only difference is button disappear instead of graying out
https://doc.qt.io/qt-5/qlineedit.html#clearButtonEnabled-prop

acdvs:
@kornes The built-in functionality for this uses an ugly icon. Is it possible to change that? If so, I'll update my code.

acdvs:
I was able to find a way to use `QLineEdit::setClearButtonEnabled` with a custom image, so I've updated to that. Much simpler.

pajlada (approved):

pajlada:
Good addition, thank you!

zneix (approved):

PR #3404, opened 2021-12-15
Add search to emote popup

acdvs:
Pull request checklist:

# Description

A search bar at the top of the emote popup that filters all emotes across every tab as you type. Uses a new ChannelView to combine all emote sets and is initially hidden until something is typed.

emote_search

[1]: https://cdn.betterttv.net/emote/60477a2f306b602acc599abf/1x

### DIFF
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -220,6 +220,8 @@ EmotePopup::EmotePopup(QWidget *parent)
                                            this->clearShortcuts();
                                            this->addShortcuts();
                                        });
+
+    this->search_->setFocus();
 }
 
 void EmotePopup::addShortcuts()
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -168,6 +168,9 @@ EmotePopup::EmotePopup(QWidget *parent)
     this->search_ = new QLineEdit();
     this->search_->setPlaceholderText("Search all emotes...");
     this->search_->setValidator(searchValidator);
+    this->search_->setClearButtonEnabled(true);
+    this->search_->findChild<QAbstractButton *>()->setIcon(
+        QPixmap(":/buttons/clearSearch.png"));
     layout->addWidget(this->search_);
 
     QObject::connect(this->search_, &QLineEdit::textChanged, this,
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -143,7 +143,7 @@ EmotePopup::EmotePopup(QWidget *parent)
     auto layout = new QVBoxLayout(this);
     this->getLayoutContainer()->setLayout(layout);
 
-    QRegExp searchRegex("[A-Za-z0-9]*");
+    QRegExp searchRegex("\\S*");
     searchRegex.setCaseSensitivity(Qt::CaseInsensitive);
     QValidator *searchValidator = new QRegExpValidator(searchRegex);
 
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -16,7 +16,10 @@
 #include "widgets/Scrollbar.hpp"
 #include "widgets/helper/ChannelView.hpp"
 
+#include <QAbstractButton>
 #include <QHBoxLayout>
+#include <QRegularExpression>
+#include <QRegularExpressionValidator>
 #include <QTabWidget>
 
 namespace chatterino {
@@ -161,9 +164,9 @@ EmotePopup::EmotePopup(QWidget *parent)
     auto layout = new QVBoxLayout(this);
     this->getLayoutContainer()->setLayout(layout);
 
-    QRegExp searchRegex("\\S*");
+    QRegularExpression searchRegex("\\S*");
     searchRegex.setCaseSensitivity(Qt::CaseInsensitive);
-    QValidator *searchValidator = new QRegExpValidator(searchRegex);
+    QValidator *searchValidator = new QRegularExpressionValidator(searchRegex);
 
     this->search_ = new QLineEdit();
     this->search_->setPlaceholderText("Search all emotes...");
@@ -190,7 +193,9 @@ EmotePopup::EmotePopup(QWidget *parent)
         view->linkClicked.connect(clicked);
 
         if (addToNotebook)
+        {
             this->notebook_->addPage(view, tabTitle);
+        }
 
         return view;
     };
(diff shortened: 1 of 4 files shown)


src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -137,40 +161,62 @@ EmotePopup::EmotePopup(QWidget *parent)
     auto layout = new QVBoxLayout(this);
     this->getLayoutContainer()->setLayout(layout);
 
-    this->notebook_ = new Notebook(this);
-    layout->addWidget(this->notebook_);
-    layout->setMargin(0);
+    QRegExp searchRegex("\\S*");
+    searchRegex.setCaseSensitivity(Qt::CaseInsensitive);
+    QValidator *searchValidator = new QRegExpValidator(searchRegex);
```
pajlada, line 166: Should use [QRegularExpression](https://doc.qt.io/qt-5/qregularexpression.html) instead of `QRegExp`, and [QRegularExpressionValidator](https://doc.qt.io/qt-5/qregularexpressionvalidator.html) instead of `QRegExpValidator`

You'll also add an include to both at the top of the file

src/widgets/dialogs/EmotePopup.hpp
```diff
@@ -29,8 +30,19 @@ class EmotePopup : public BasePopup
     ChannelView *channelEmotesView_{};
     ChannelView *subEmotesView_{};
     ChannelView *viewEmojis_{};
+    ChannelView *searchView_{};
 
+    ChannelPtr channel_;
+    TwitchChannel *twitchChannel_;
+
+    QLineEdit *search_;
```
pajlada, line 38: Add an include to QLineEdit at the top of this file
`#include <QLineEdit>`

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -137,40 +161,62 @@ EmotePopup::EmotePopup(QWidget *parent)
...
 
         view->setOverrideFlags(MessageElementFlags{
             MessageElementFlag::Default, MessageElementFlag::AlwaysShow,
             MessageElementFlag::EmoteImages});
         view->setEnableScrollingToBottom(false);
-        this->notebook_->addPage(view, tabTitle);
         view->linkClicked.connect(clicked);
 
+        if (addToNotebook)
+            this->notebook_->addPage(view, tabTitle);
```
pajlada, line 193:
```suggestion
        if (addToNotebook) {
            this->notebook_->addPage(view, tabTitle);
        }
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -137,40 +161,62 @@ EmotePopup::EmotePopup(QWidget *parent)
...
-    layout->setMargin(0);
+    QRegExp searchRegex("\\S*");
+    searchRegex.setCaseSensitivity(Qt::CaseInsensitive);
+    QValidator *searchValidator = new QRegExpValidator(searchRegex);
+
+    this->search_ = new QLineEdit();
+    this->search_->setPlaceholderText("Search all emotes...");
+    this->search_->setValidator(searchValidator);
+    this->search_->setClearButtonEnabled(true);
+    this->search_->findChild<QAbstractButton *>()->setIcon(
```
pajlada, line 172: Missing include of QAbstractButton in this file, add it to the top

`#include <QAbstractButton>`

src/widgets/dialogs/EmotePopup.hpp
```diff
@@ -29,8 +30,19 @@ class EmotePopup : public BasePopup
     ChannelView *channelEmotesView_{};
     ChannelView *subEmotesView_{};
     ChannelView *viewEmojis_{};
+    ChannelView *searchView_{};
```
pajlada, line 33:
```suggestion
    /**
     * @brief Visible only when the user has specified a search query into the `search_` input, otherwise the `notebook_` with all the other views are visible.
     */
    ChannelView *searchView_{};
```

Correct the comment if it's wrong please!

src/widgets/dialogs/EmotePopup.hpp
```diff
@@ -29,8 +30,19 @@ class EmotePopup : public BasePopup
...
 
+    ChannelPtr channel_;
+    TwitchChannel *twitchChannel_;
```
pajlada, line 36:
```suggestion
    TwitchChannel *twitchChannel_{};
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -313,29 +354,119 @@ void EmotePopup::loadChannel(ChannelPtr _channel)
...
+void EmotePopup::loadEmojis(Channel &channel, EmojiMap &emojiMap,
+                            const QString &title)
+{
+    channel.addMessage(makeTitleMessage(title));
+    channel.addMessage(makeEmojiMessage(emojiMap));
+}
+
+void EmotePopup::filterEmotes(const QString &text)
+{
+    if (text.length() > 0)
```
pajlada, line 374: Given that this only has two states, consider switching the logic around to look something like this instead:
```
if (text.empty()) {
    this->notebook_->show();
    this->searchView_->hide();

    return;
}

// search logic here
```

In addition to that, consider changing the argument name `text` to something more descriptive like `searchText` or `searchQuery`

pajlada (changes requested):
Nice feature addition! Thank you for the contribution

Got some small tweaks/concerns covered in the comments that I'd like adressed before merging this in. Let me know if you have any questions!

Open questions for others who test this:
1) How should pressing Escape behave when the search input is filled? Right now it closes the popup, but I feel like maybe it should clear the search entry first if there's a search query active
2) Should the search input be selected by default? I don't believe it is right now, but I think it would be nice if it was

zneix:
> Should the search input be selected by default? I don't believe it is right now, but I think it would be nice if it was

I believe it should be - we already have this as some sort of standard across different dialogs that implement the search input anyway (e.g. Settings dialog, Message Search dialog)

pajlada:
> > Should the search input be selected by default? I don't believe it is right now, but I think it would be nice if it was
> 
> I believe it should be - we already have this as some sort of standard across different dialogs that implement the search input anyway (e.g. Settings dialog, Message Search dialog)

Got code in my local repository fixing this, and adding support for the hotkey system.

Will push together with a merge when another PR has been merged in to save on CI hours.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -259,22 +310,17 @@ void EmotePopup::addShortcuts()
...
+    this->setWindowTitle("Emotes in #" + this->channel_->getName());
 
-    auto addEmotes = [&](Channel &channel, const EmoteMap &map,
-                         const QString &title,
-                         const MessageElementFlag &emoteFlag) {
-        channel.addMessage(makeTitleMessage(title));
-        channel.addMessage(makeEmoteMessage(map, emoteFlag));
-    };
+    if (this->twitchChannel_ == nullptr)
+        return;
```
zneix, line 323:
```suggestion
    if (this->twitchChannel_ == nullptr)
    {
        return;
    }
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -313,29 +359,117 @@ void EmotePopup::loadChannel(ChannelPtr _channel)
...
+    this->notebook_->hide();
+    this->searchView_->show();
+}
+
+EmoteMap *EmotePopup::filterEmoteMap(const QString &text,
+                                     std::shared_ptr<const EmoteMap> emotes)
+{
+    auto filteredMap = new EmoteMap();
+
+    for (const auto emote : *emotes)
```
zneix, line 464: Don't copy
```suggestion
    for (const auto &emote : *emotes)
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -313,29 +359,117 @@ void EmotePopup::loadChannel(ChannelPtr _channel)
...
+        return;
+    }
+
+    auto searchChannel = std::make_shared<Channel>("", Channel::Type::None);
+
+    auto twitchEmoteSets =
+        getApp()->accounts->twitch.getCurrent()->accessEmotes()->emoteSets;
+    std::vector<std::shared_ptr<TwitchAccount::EmoteSet>> twitchGlobalEmotes{};
+
+    for (const auto set : twitchEmoteSets)
```
zneix, line 393: Don't copy
```suggestion
    for (const auto &set : twitchEmoteSets)
```

zneix (changes requested):

pajlada (approved):

zneix (approved):

PR #3418, opened 2021-12-23
Add setting to prevent or highlight message overflow

acdvs:
Pull request checklist:

# Description

Native Twitch chat prevents the user from typing past 500 characters entirely. Right now, Chatterino only enforces this rule by preventing 501+ characters from being sent. This adds a setting to prevent or highlight input overflow to match the native Twitch functionality.

My motivation to add this comes from some frustration in wanting to send a long message but having to manually delete characters until only 500 were left.

### Prevent overflow
overflow_prevent

### Highlight overflow
overflow_highlight


leon-richardt:
I would certainly like some sort of indication _why_ text input is disabled. (Note that this criticism applies to the current mechanism of failing silentyl too!)
Perhaps using a red background for the input box as well as a message above the input?

Just spitballing here, there might be better solutions.

Mm2PL:
>  Perhaps using a red background for the input box as well as a message above the input?

This already happens with the character counter

acdvs:
@Mm2PL I think the concern there may be that visual feedback needs to be on the input text itself. Maybe something similar to what Twitter does: 
image

Mm2PL (changes requested):
This PR does not respect our character counter logic, please make change it to take that into account.

It's still possible to paste messages from the emote popup and paste text via Ctrl+V and right-click menu. I personally don't like the way you achieved this feature with whitelisting keys that could be pressed. IMO It's way better to use a character inserted or text changed event.

acdvs:
I've updated it to use a textChanged slot instead of a keyPress event which handles both typing and pasting all in one. This still won't allow extra spaces at the end as you mentioned, but I strongly suggest that shouldn't be possible because it makes the UX side of character limitation unclear.

src/widgets/splits/SplitInput.cpp
```diff
@@ -701,6 +700,28 @@ void SplitInput::editTextChanged()
             app->commands->execCommand(text, this->split_->getChannel(), true);
     }
 
+    if (getSettings()->messageOverflow.getValue() == 2)
```
zneix, line 703: While comparing setting's value, you should not assume "magic values" like `2` here, instead make an `enum`, which can look something like https://github.com/Chatterino/chatterino2/blob/51ece94f588a60c578504309f9eeea22a831f0fa/src/util/StreamerMode.hpp?plain=1#L7 and then compare to, e.g. `FooBar::Highlight`

src/singletons/Settings.hpp
```diff
@@ -84,6 +84,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting showEmptyInput = {"/appearance/showEmptyInputBox", true};
     BoolSetting showMessageLength = {"/appearance/messages/showMessageLength",
                                      false};
+    IntSetting messageOverflow = {"/appearance/messages/messageOverflow", 0};
```
zneix, line 87: Following the suggestion below, change type of this setting to `EnumSetting<T>` instead. Following that, change default value to, e.g. `FooBar::Allow`.

src/widgets/helper/ResizingTextEdit.cpp
```diff
@@ -30,6 +33,17 @@ ResizingTextEdit::ResizingTextEdit()
     this->installEventFilter(this);
 }
 
+void ResizingTextEdit::editTextChanged()
+{
+    if (getSettings()->messageOverflow.getValue() == 1 &&
+        this->toPlainText().length() > SplitInput::TWITCH_MESSAGE_LIMIT)
```
zneix, line 39: Same is in the comment in SplitInput.cpp

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - Minor: Show picked outcome in prediction badges. (#3357)
 - Minor: Add support for Emoji in IRC (#3354)
 - Minor: Moved `/live` logs to its own subdirectory. (Logs from before this change will still be available in `Channels -> live`). (#3393)
+- Minor: Add setting to limit message input length (#3418)
```
zneix, line 41:
```suggestion
- Minor: Add setting to limit message input length. (#3418)
```

zneix (changes requested):
Pull request looks nice, I like the idea of letting user know what part of the message won't go through anyway. Got a couple things to point out:

1. Name of the setting in Settings dialog (`Message overflow`) isn't very descriptive to an average end user imho. I can't really come up with a better naming, but would like to hear others' opinions about it. cc @Felanbird
2. I think that `Highlight` option makes sense to be set to as a default setting value instead of `Allow` - bringing in new functionality that isn't enabled by default makes it difficult to discover by users; and I think this isn't anything harmful to set by default.

src/widgets/helper/ResizingTextEdit.cpp
```diff
@@ -30,6 +33,17 @@ ResizingTextEdit::ResizingTextEdit()
...
+    {
+        QTextCursor cursor = this->textCursor();
+        this->setTextCursor(cursor);
+        cursor.deletePreviousChar();
+    }
+}
```
Mm2PL, line 45: Could this logic just sit in SplitInput's `editTextChanged`?

Mm2PL (changes requested):
I've got an issue with the feature as is now. If you type out a very loooong message, change the setting to `Prevent`. And then type a character into the filled input will lag Chatterino for a good while. From my testing it turns out that it's because the `text changed` event triggers itself recursively. With enough characters Chatterino will segfault (~12k is enough), again probably because of a stack overflow. While this is not really a very bad issue per se, I'm thinking that it's a sign of a mistake somewhere. A possible solution would be to take the input text, clip it to 500 chars and then set it back to the input, leaving only one recursive `text changed` event which we can live with. Another solution would be to have some kind of a locking mechanism to make sure that the recursively generated events are safely ignored. 

Sorry for the pedantic review.


###### Fun fact: `bt` in my gdb is still going as I'm correcting/expanding this message.

acdvs:
I appreciate the feedback! I'll make changes soon.

@zneix I spent longer than I'd like trying to find better wording for this. Originally I was using "Input overtype behavior" but that seemed too wordy.

acdvs:
@zneix I've changed the default to `Highlight` based on your comment. It can only be beneficial to the UX as a purely visual feature.

@Mm2PL The `Prevent` behavior now trims the whole input via `QString::left` instead of per-character with a cursor. This also means no extra checks are needed to prevent excessive recursion.  `SplitInput::editTextChanged` will still retrigger just once though.

ALazyMeme:
So after my testing on fecba0e0:
- The PR tries to enforce the highlighting on IRC channels, which isn't correct (I believe it's 512 characters?)
- The "prevent" option doesn't work in IRC

Not sure if you're able to add a separate limit for IRC or something? Then would also need the prevent option fixed.

Otherwise it seems to work as intended.

src/widgets/helper/ResizingTextEdit.cpp
acdvs: I've moved it over, but could there be cases where this feature is needed in other ResizingTextEdit instances?

acdvs:
Apologies for leaving this hanging for so long. Is anyone else able to figure out a good fix for ALazyMeme's feedback? My knowledge is limited.

pajlada:
I fixed the merge conflicts and fixed the issues ALazyMeme mentioned, right now I only have two concerns that I can get to myself later:
1) The setting should, imo, be less accessible. It should be hidden under Advanced somewhere as I think the default "Highlight" is going to be good/preferred for 99% of users. The text for the setting should probably be something different too, although naming things is hard so I can't come up with a direct improvement yet!
2) With the Highlight setting, it's possible to get into the state where the message you're typing is completely red. This is achieved by overflowing your messagebox, selecting all text, then starting to type

https://user-images.githubusercontent.com/962989/198898324-f29d6ae9-5767-40bc-b583-60a3dee3dad8.mp4

pajlada (approved):

PR #3427, opened 2021-12-26
Optimize logs folder size calculation

zneix:
Pull request checklist:

# Description

I've found that using [QDirIterator](https://doc.qt.io/qt-5/qdiriterator.html) is actually a faster method of fetching size of all files within a directory and all of its subdirectories.

little unrelated note: found that old method was found in [this SO thread](https://stackoverflow.com/a/47854799), I think at least. It was introduced in #600.
https://github.com/Chatterino/chatterino2/blob/94f7f09e73f5723f8c121c8a70a76c7994f7ca17/src/widgets/settingspages/ModerationPage.cpp#L26-L57

Decided to keep old functionality in a benchmark that compares 2 ways of approaching this problem.
NOTE: In the benchmark, I didn't know how to get the log path, so I just hardcoded path to my logs directory - you may want to change it upon performing benchmarks on your own.

<details>
<summary>Benchmark results</summary>

My chatterino folder, without logs (takes up 1.3M according to `du -hs`)
built with `cmake -DBUILD_BENCHMARKS=On -DBUILD_APP=Off -DUSE_PRECOMPILED_HEADERS=Off -DCMAKE_BUILD_TYPE=Debug`
```
2021-12-26T16:01:09+01:00
Running ./bin/chatterino-benchmark
Run on (16 X 2937.81 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 4096 KiB (x2)
Load Average: 2.68, 5.62, 4.94
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_ShortcodeParsing             2680 ns         2680 ns       261304
BM_LogsSizeCalculationOld      88422 ns        86499 ns         8018
BM_LogsSizeCalculationNew      60134 ns        59201 ns        12118
```

My chatterino logs (taking up 13G according to `du -hs`)
built with `cmake -DBUILD_BENCHMARKS=On -DBUILD_APP=Off -DUSE_PRECOMPILED_HEADERS=Off -DCMAKE_BUILD_TYPE=Debug`
```
2021-12-26T16:03:41+01:00
Running ./bin/chatterino-benchmark
Run on (16 X 3606.08 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 4096 KiB (x2)
Load Average: 2.33, 4.27, 4.52
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_ShortcodeParsing             2591 ns         2591 ns       249824
BM_LogsSizeCalculationOld  733373814 ns    292873707 ns            2
BM_LogsSizeCalculationNew  593373335 ns    254478882 ns            3
```

</details>


pajlada (changes requested):
Code change itself is great, benchmarked my logs and results are good!

2 nitpicks about the benchmark itself:
1) Rename the file from ModerationPage to Helpers
2) Remove the reliance of settings/the users real logs, and instead generate/download logs as a CMake step. See https://github.com/pajlada/settings/blob/master/CMakeLists.txt#L82-L92 for how we can run scripts as part of the cmake build portion

pajlada:
For reference, I'm also fine with leaving the benchmarking to a separate PR! In which case, the function should still live within ModerationPage for commit history simplicity's sake

zneix:
> For reference, I'm also fine with leaving the benchmarking to a separate PR! In which case, the function should still live within ModerationPage for commit history simplicity's sake

Removed benchmark because I can't be bothered on tinkering it and as per this comment I'm going to just merge this in as it is after you approve this.

pajlada (approved):
Merge when function has moved back to ModerationPage

PR #3437, opened 2021-12-28
Add settings tooltips

### LINKED ISSUE #2146: Description when you hover over a setting
**What should be added?**
A feature that adds a description when you hover over a setting. This would look something like when you hover over a link in chat. I understand that most settings are self explanatory but for the ones that aren't it's really troubling to figure out what they do. Without reading through the source code someone may be missing out and a great feature that they didn't even know existed. I personally struggled to know what some of the settings do. I had to ask the chat for help. 

**Why should it be added?**
This feature would allow for the user to understand what each setting does. For the users that come from using something like bttv they want something where they can easily add all they're settings back without the trouble of figuring out what each setting does. So this could end up with them not like the experience of Chatterino and never using it again. In my option this feature is just a necessary quality of life aspect that should to be added.

acdvs:
Pull request checklist:

# Description

Resolves #2146 

Displays a tooltip when hovering over a setting. This adds another parameter to each `GeneralPageView::add[Widget]` function for easy definition within `GeneralPage`. Uses the built-in Qt tooltip with a modified stylesheet to match the settings page colors.

Example

Initially, this PR adds tooltips for the settings listed in #2146 at the time of writing this.

Not sure about the tooltips for badge visibility. Should those stay on the labels?


src/widgets/settingspages/GeneralPageView.hpp
```diff
@@ -99,15 +99,17 @@ class GeneralPageView : public QWidget
     SubtitleLabel *addSubtitle(const QString &text);
     /// @param inverse Inverses true to false and vice versa
     QCheckBox *addCheckbox(const QString &text, BoolSetting &setting,
-                           bool inverse = false);
-    ComboBox *addDropdown(const QString &text, const QStringList &items);
+                           bool inverse = false, QString toolTipText = nullptr);
```
Mm2PL, line 102: While this isn't a blocker, I dislike the usage of `nullptr` here. To me the `nullptr` literal means one thing: a pointer with the value 0 and QString isn't a pointer per se.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -625,11 +634,16 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                       false,
+                       "Make all clickable links lowercase to deter "
+                       "phishing attempts.");
     layout.addCheckbox("Bold @usernames", s.boldUsernames);
     layout.addCheckbox("Color @usernames", s.colorUsernames);
     layout.addCheckbox("Try to find usernames without @ prefix",
-                       s.findAllUsernames);
+                       s.findAllUsernames, false,
+                       "Attempt to autocomplete usernames "
+                       "without the @ prefix.");
```
Mm2PL, line 646: This isn't related to autocompletion. This is finding usernames in chat.
```suggestion
                       "Find mentions of users in chat "
                       "without the @ prefix.");
```
(will need reformat)

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -665,11 +679,14 @@ void GeneralPage::initLayout(GeneralPageView &layout)
 
     layout.addCheckbox(
         "Only search for emote autocompletion at the start of emote names",
-        s.prefixOnlyEmoteCompletion);
+        s.prefixOnlyEmoteCompletion, false,
+        "When disabled, autocomplete emotes based on any part of the name.");
```
Mm2PL, line 683: I find this tooltip a tad confusing. I'm not sure how it should look like, but it took me a good 30s to understand what is being said here. @Felanbird could you help?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -691,8 +708,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         [](auto args) {
             return args.index;
         },
-        false);
-    layout.addCheckbox("Combine multiple bit tips into one", s.stackBits);
+        false, "Combine consecutive timeout messages into a single message.");
+    layout.addCheckbox("Combine multiple bit tips into one", s.stackBits, false,
+                       "Combine consecutive bit tips into a single message.");
```
Mm2PL, line 713:
```suggestion
    layout.addCheckbox("Combine multiple bit tips into one", s.stackBits, false,
                       "Combine consecutive cheermotes into a single one.");
```
This setting is about combining bits from a single message into a single emote.

Mm2PL (changes requested):
I have some code comments. I feel like there isn't much to notice that these elements have a tooltip associated with them. Maybe adding an 🛈 icon would work or as some apps do it: a question mark with a dashed/dotted underline.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -579,15 +585,17 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                       "e.g., broadcaster, moderator");
     layout.addCheckbox("Subscriber ", s.showBadgesSubscription);
-    layout.addCheckbox("Vanity (prime, bits, subgifter)", s.showBadgesVanity);
+    layout.addCheckbox("Vanity", s.showBadgesVanity, false,
+                       "e.g., prime, bits, sub gifter");
     layout.addCheckbox("Chatterino", s.showBadgesChatterino);
-    layout.addCheckbox("FrankerFaceZ (Bot, FFZ Supporter, FFZ Developer)",
-                       s.showBadgesFfz);
+    layout.addCheckbox("FrankerFaceZ", s.showBadgesFfz, false,
+                       "e.g., bot, FFZ supporter, FFZ developer");
```
Felanbird, line 598:
```suggestion
                       "e.g. Bot, FFZ supporter, FFZ developer");
```

src/widgets/settingspages/GeneralPage.cpp
Felanbird:
```suggestion
        "When disabled, emote tab-completion will complete based on any part "
        "of the name."
        "\ne.g. sheffy -> DatSheffy");
```
minor wording change + an example, formatter doesn't mind what line the example is on 🤷

src/widgets/settingspages/GeneralPage.cpp
Felanbird, line 598:
```suggestion
                       "e.g. Bot, FFZ supporter, FFZ developer");
```

src/widgets/settingspages/GeneralPage.cpp
acdvs: There needs to be emphasis that this concerns multiple cheermotes in a single message and not single cheermotes in multiple messages. How about "Combine consecutive cheermotes (sent in a single message) into one cheermote"?

src/widgets/settingspages/GeneralPageView.cpp
```diff
@@ -362,4 +375,27 @@ void GeneralPageView::updateNavigationHighlighting()
...
+    if (!text.isEmpty())
+    {
+        if (text.length() > MAX_TOOLTIP_LINE_LENGTH)
+        {
+            // match MAX_TOOLTIP_LINE_LENGTH characters, any remaining
+            // non-space, and then capture the following space for
+            // replacement with newline
+            QString maxLineLengthPattern =
+                QString(".{%1}\\S*\\K(\\s+)").arg(MAX_TOOLTIP_LINE_LENGTH);
+            text.replace(QRegularExpression(maxLineLengthPattern), "\n");
```
leon-richardt, line 389: The `QRegularExpression` could be made static to reduce the overhead of parsing the expression

src/widgets/settingspages/GeneralPageView.cpp
```diff
@@ -362,4 +375,27 @@ void GeneralPageView::updateNavigationHighlighting()
     }
 }
 
+void GeneralPageView::addToolTip(QWidget &widget, QString text)
+{
+    if (!text.isEmpty())
```
jammehcow, line 380: Just to reduce nesting, I would recommend inverting this `if` to return early.

src/widgets/settingspages/GeneralPageView.cpp
```diff
@@ -362,4 +375,27 @@ void GeneralPageView::updateNavigationHighlighting()
...
+                QString(".{%1}\\S*\\K(\\s+)").arg(MAX_TOOLTIP_LINE_LENGTH);
+            text.replace(QRegularExpression(maxLineLengthPattern), "\n");
+        }
+
+        widget.setToolTip(text);
+        widget.setStyleSheet("QToolTip {"
+                             "padding: 2px;"
+                             "background-color: #333333;"
+                             "border: 1px solid #545454;"
+                             "}");
```
jammehcow, line 397: This should be moved to a static field (or whatever it's called in C++) to allow reuse across other classes. Would definitely use it for some other tooltips across the application.

jammehcow (changes requested):

pajlada (approved):
Took me long enough, thank you for doing the heavy lifting! :bow:

PR #3442, opened 2021-12-30
Check live status for all closed channels with notifications at once

Mm2PL:
Pull request checklist:

- [x] Execute multiple api calls and add a limit

# Description

No more hundred API calls for channels not attached to a split.


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 - Minor: Added autocompletion for default Twitch commands starting with the dot (e.g. `.mods` which does the same as `/mods`). (#3144)
 - Minor: Sorted usernames in `Users joined/parted` messages alphabetically. (#3421)
 - Minor: Mod list, VIP list, and Users joined/parted messages are now searchable. (#3426)
+- Minor: Check live status for all closed channels with notifications at once (#3442)
```
Felanbird, line 45:
```suggestion
- Minor: Batch checking live status for channels with live notifications that aren't connected. (#3442)
```
`batch` is always the one word that I feel like fits but doesn't always for this stuff so 🤷

zneix: Honestly, is this even worth a changelog entry? not like anything that will impact users has changed in any way.

jammehcow: I think it's worth the entry, but not with this wording.

Maybe `Reduce network calls when checking if a channel is live.' as it's easily understandable by an end user. Yeah it takes out the fact that it's batched now, but that's enough to confuse a non-technical person :)

Felanbird: > I think it's worth the entry, but not with this wording.
> 
> Maybe `Reduce network calls when checking if a channel is live.' as it's easily understandable by an end user. Yeah it takes out the fact that it's batched now, but that's enough to confuse a non-technical person :)

Zneix and I had the discussion in chat which was basically remove it or make it dev

pajlada (approved):

zneix (approved):

PR #3445, opened 2022-01-02
Add subscriber and founder badge highlighting

Infinitay:
Pull request checklist:

# Description

Added the capability to highlight messages from sub or founder badges as they were missing to begin with. FWIW Founder badge highlighting does not fall under the subscriber badge. For example, 1/3/6/etc month sub would highlight via the Subscriber badge highlight, but you would have to have a separate highlight for Founder badge highlighting.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Minor: Added autocompletion for default Twitch commands starting with the dot (e.g. `.mods` which does the same as `/mods`). (#3144)
 - Minor: Sorted usernames in `Users joined/parted` messages alphabetically. (#3421)
 - Minor: Mod list, VIP list, and Users joined/parted messages are now searchable. (#3426)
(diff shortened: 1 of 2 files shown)


Infinitay:
Someone brought this feature up on Discord and I was curious as to why we don't already have it. Found mention of it in #1704 

> There's currently no difference between [bots, subscribers, and turbo] badges in Chatterino - those all fall under "vanity" slot. An enhancement adding MessageElementFlags for all of those badges would have to be made in order to support that, but there's no real reference for all the twitch badges iirc.

However, the subscriber badge is not listed as a vanity badge in `MessageElement`. At the very least I don't see why we shouldn't support subscriber badge highlighting. 
https://github.com/Chatterino/chatterino2/blob/3c4331b8cbb535cf6cbf31586d21683144ff0668/src/messages/MessageElement.hpp#L74

pajlada (approved):

pajlada:
Good addition, thank you! Tested and it works as expected!

PR #3473, opened 2022-01-07
Fix edge case where hotkeys would be empty after a crash, causing Chatterino to be unusable unless the user reset their hotkeys

Mm2PL:
Pull request checklist:

- [ ] ~~`CHANGELOG.md` was updated, if applicable~~ I don't think this should get a changelog entry because custom hotkeys have not been into a stable yet. pajaDent

# Description

This forces a save of settings after hotkeys have been saved into memory. I can't really reproduce the bug so this is the best fix I can do rn.


pajlada (approved):

zneix:
>  I don't think this should get a changelog entry because custom hotkeys have not been into a stable yet. pajaDent

I know I'm late, but just appending PR number to the already existing changelog entry would be something you can do. e.g. `- Major: Added customizable shortcuts. (#2340, #3473)`. Keep that in mind if there's any future PR fixing hotkeys-related functionality - could sneak in number of this PR as well then.

PR #3482, opened 2022-01-10
Added new icons all related to "Suspicious User Controls"

karl-police:
Added them now, so they don't have to be added later, I guess.


Mm2PL:
Do you really have rights to use these icons? I have a slight suspicion that they are just downloaded from the platform's website.

karl-police:
> Do you really have rights to use these icons? I have a slight suspicion that they are just downloaded from the platform's website.

Aren't the other ones also just downloaded?

Mm2PL:
> Aren't the other ones also just downloaded?

I don't think so. I cannot speak for others but I've created the copy icons myself, using a public domain image as an inspiration.

> And they're not downloaded, I just took the SVG from HTML.

I don't think copyright lawyers care about technicalities like this. I don't think we should either.

We also don't have a use for these icons: literally any action to do with the suspicious users feature uses first-party-only APIs which we cannot access because Chatterino is a third-party app.

Felanbird:
I'm going to close this for both of the above reasons, Mm2PL is correct, the suspicious users features that these icons are for are first party and are **extremely** unlikely to be made third party and subsequently implemented into Chatterino, and if they are they should just be included in the PR related to these features.

Also as mentioned, and by your attempted wording technicality, you took these icons and we don't have permission to be using them.

PR #3496, opened 2022-01-13
Don't crash upon right-clicking automod badge

### LINKED ISSUE #2969: Right-clicking AutoMod badge crashes
This happens due to Chatterino trying to load emote data for badge, which doesn't exist because our current badge is custom built.
Crash only occurs after 675f99e9c.

zneix:
Pull request checklist:

# Description

Following up 675f99e9c, added a check whether result of dynamic_cast is not a `nullptr` to prevent any other crashes caused by faulty badge elements.
Also fixed the AutoMod badge and made it an actual instance of `BadgeElement` which is what it should've been from the beginning, instead of `ImageElement`.

Fixes #2969

### DIFF
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -28,6 +28,14 @@ MessagePtr makeSystemMessage(const QString &text, const QTime &time)
     return MessageBuilder(systemMessage, text, time).release();
 }
 
+EmotePtr makeAutoModBadge()
+{
+    return std::make_shared<Emote>(Emote{
+        EmoteName{}, ImageSet{Image::fromPixmap(getResources().twitch.automod)},
+        Tooltip{"AutoMod"},
+        Url{"https://dashboard.twitch.tv/settings/moderation/automod"}});
+}
+
 MessagePtr makeAutomodInfoMessage(const AutomodInfoAction &action)
 {
     auto builder = MessageBuilder();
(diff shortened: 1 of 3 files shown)


src/messages/MessageBuilder.cpp
```diff
@@ -28,6 +28,14 @@ MessagePtr makeSystemMessage(const QString &text, const QTime &time)
     return MessageBuilder(systemMessage, text, time).release();
 }
 
+EmotePtr makeAutoModBadge()
+{
+    return std::make_shared<Emote>(Emote{
+        EmoteName{}, ImageSet{Image::fromPixmap(getResources().twitch.automod)},
+        Tooltip{"AutoMod"},
+        Url{"https://dashboard.twitch.tv/settings/moderation/automod"}});
```
zneix, line 36: This actually redirects to, e.g. `https://dashboard.twitch.tv/u/zneix/settings/moderation/automod`, which might be useful if we'll add something like `jump to "badge's url"` in the future

pajlada (approved):
Good fix, feel free to merge in when you're comfortable!

Felanbird (approved):
Functionality works as intended. Right-clicking the automod badge no longer crashes. 👍

Felanbird:
That being said, the functionality of open/copy link do not work and the 1x badge tool-tip looks out of place compared to all other badges which use a 3x.

https://static-cdn.jtvnw.net/badges/v1/df9095f6-a8a0-4cc2-bb33-d908c0adffb8/1
https://static-cdn.jtvnw.net/badges/v1/df9095f6-a8a0-4cc2-bb33-d908c0adffb8/2
https://static-cdn.jtvnw.net/badges/v1/df9095f6-a8a0-4cc2-bb33-d908c0adffb8/3

But this can be implemented in a separate PR.

Mm2PL (approved):
This fix works nice. 

As we discussed in #zneix the better long term solution would be to load the badge like any other Twitch badge. The `TwitchBadges` singleton should already contain `twitchbot` (version `1`) because it loads all badges.

zneix:
> But this can be implemented in a separate PR.

Yup, after some further thinking I'm going to call it here for this PR and improve the automod badge visually separetly, since this rather focuses on the bugfix than working on visual details.

PR #3503, opened 2022-01-13
Remove first timestamp from genereated AutoMod queue messages

karl-police:
Pull request checklist:

# Description

Remove timestamp from AutoMod, for style purposes and because on Twitch it doesn't show it either. The only thing where the timestamp should show is from the message itself.

Same things is with the "low-trust-users" messages.


pajlada:
Could you provide a **before** and **after** screenshot?

karl-police:
> Could you provide a **before** and **after** screenshot?

Sure.

Before:
image


After:
image


Preferably at some point, we could reduce the gap between the modmail message and the users message that would have been sent in the chat.

@pajlada do you have more of these "builder.emplace" "Elements" or is it based of a template from Qt on something, couldn't find that. So that I could look into more stuff for the UI.

zneix (approved):

pajlada (approved):
Looks better!

pajlada:
> @pajlada do you have more of these "builder.emplace" "Elements" or is it based of a template from Qt on something, couldn't find that. So that I could look into more stuff for the UI.

https://github.com/Chatterino/chatterino2/blob/master/src/messages/MessageBuilder.hpp#L65-L75 That's where we've defined the emplace function that you used just now - all it does is add a Message Element to the Message Builder.

It's a semi-common pattern we follow where we have builders

Mm2PL:
I thought that we had added the timestamps on reward message to be consistent with automod

pajlada:
> I thought that we had added the timestamps on reward message to be consistent with automod

Possibly - we will want to review other "message headers" and see about removing the timestamp of those too I think

pajlada:
Thanks for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

PR #3509, opened 2022-01-14
Move OSes to their own CI files

Mm2PL:
Pull request checklist:

# Description

It's so much easier to block a single OS from building or rerun fucking Windows that doesn't work when you don't need to rerun all other CI with it.

Looks like after I finish this, it's gonna have to get an admin merge.


Mm2PL:
conflict hell, might try this some time later

PR #3510, opened 2022-01-15
Show feedback upon opening a channel in streamlink in more places

### LINKED ISSUE #3488: All ways to open a stream in Streamlink should print an info message
### Checklist

### Describe your issue

When opening a stream in Streamlink via the `/streamlink` command, the message `Opening <stream> in streamlink...` is printed to the currently active split. When using the hotkey or the entry in "three-dots" context menu, no such message is printed.

I find this irritating especially when opening a stream via the hotkey since there usually is a delay between using the hotkey and the player opening up. I would like every method to print the same message.

### Screenshots

When opening via the `/streamlink` command
This is what happens when using the command. No such message is printed with the other methods.

### OS and Chatterino Version

Linux 5.15.13-arch1-1, 82009982

zneix:
Pull request checklist:

# Description

Note: we could perhaps also show feedback when opening stream fails (e.g. channel is offline and/or child process quits with 1), but that's for another PR.

Fixes #3488

### DIFF
--- a/src/util/StreamLink.cpp
+++ b/src/util/StreamLink.cpp
@@ -213,8 +213,15 @@ void openStreamlinkForChannel(const QString &channel)
 
     auto *currentPage = dynamic_cast<SplitContainer *>(
         getApp()->windows->getMainWindow().getNotebook().getSelectedPage());
-    currentPage->getSelectedSplit()->getChannel()->addMessage(
-        makeSystemMessage(INFO_TEMPLATE.arg(channel)));
+    if (currentPage != nullptr)
+    {
+        if (auto currentSplit = currentPage->getSelectedSplit();
+            currentSplit != nullptr)
+        {
+            currentSplit->getChannel()->addMessage(
+                makeSystemMessage(INFO_TEMPLATE.arg(channel)));
+        }
+    }
 
     QString channelURL = "twitch.tv/" + channel;
 
--- a/src/util/StreamLink.cpp
+++ b/src/util/StreamLink.cpp
@@ -1,10 +1,14 @@
 #include "util/StreamLink.hpp"
 
 #include "Application.hpp"
+#include "providers/irc/IrcMessageBuilder.hpp"
 #include "singletons/Settings.hpp"
+#include "singletons/WindowManager.hpp"
 #include "util/Helpers.hpp"
 #include "util/SplitCommand.hpp"
+#include "widgets/Window.hpp"
 #include "widgets/dialogs/QualityPopup.hpp"
+#include "widgets/splits/Split.hpp"
 
 #include <QErrorMessage>
 #include <QFileInfo>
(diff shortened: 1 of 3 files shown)


zneix:
*Requesting a review from @leon-richardt silentyl*

Mm2PL:
From what I've looked at the code, this changes where feedback from `/streamlink CHANNEL` lands. Did you intend to do that? miniDank

src/util/StreamLink.cpp
```diff
@@ -205,6 +209,11 @@ void openStreamlink(const QString &channelURL, const QString &quality,
 
 void openStreamlinkForChannel(const QString &channel)
 {
+    auto *currentPage = dynamic_cast<SplitContainer *>(
+        getApp()->windows->getMainWindow().getNotebook().getSelectedPage());
+    currentPage->getSelectedSplit()->getChannel()->addMessage(
+        makeSystemMessage(QString("Opening %1 in streamlink...").arg(channel)));
```
leon-richardt, line 215:
```suggestion
    static const QString INFO_TEMPLATE("Opening %1 in Streamlink ...");

    auto *currentPage = dynamic_cast<SplitContainer *>(
        getApp()->windows->getMainWindow().getNotebook().getSelectedPage());
    currentPage->getSelectedSplit()->getChannel()->addMessage(
        makeSystemMessage(INFO_TEMPLATE.arg(channel)));
```

This prevents a temporary QString from being created on every function call.

leon-richardt (commented):
:+1:

src/util/StreamLink.cpp
zneix: Oh yea, I didn't think about this... Thank god there's smarter programmers

leon-richardt: It's really a nitpick in this case, not likely to ever cause performance problems. But while we're on it, we might as well make it the most correct it can be :man_shrugging:

leon-richardt (approved):

src/util/StreamLink.cpp
```diff
@@ -205,6 +209,13 @@ void openStreamlink(const QString &channelURL, const QString &quality,
 
 void openStreamlinkForChannel(const QString &channel)
 {
+    static const QString INFO_TEMPLATE("Opening %1 in Streamlink ...");
+
+    auto *currentPage = dynamic_cast<SplitContainer *>(
+        getApp()->windows->getMainWindow().getNotebook().getSelectedPage());
+    currentPage->getSelectedSplit()->getChannel()->addMessage(
+        makeSystemMessage(INFO_TEMPLATE.arg(channel)));
```
pajlada, line 217: `getSelectedPage()` and `getSelectedSplit()` can reasonably return a `nullptr`, so we should check for that

pajlada (changes requested):

pajlada (approved):

PR #3529, opened 2022-01-20
Added `/popup` command to open currently focused split or supplied channel in a new window

### LINKED ISSUE #3517: Add a `/popup` command
### Checklist

### Describe your issue

Add a `/popup [channel]` command
Works similar to the `/popout` command - basically allowing a user to popup a chat without having to open the split. Using the command without the channel name would open the current split in a popup.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 7.3.4 (commit 8e7fe9910) on Windows 10

qooq69:
Pull request checklist:

# Description
Closes #3517


src/controllers/commands/CommandController.cpp
```diff
@@ -697,6 +697,33 @@ void CommandController::initialize(Settings &, Paths &paths)
...
 
+    this->registerCommand(
+        "/popup", [](const QStringList &words, ChannelPtr channel) {
+            QString target(words.size() < 2 ? channel->getName() : words[1]);
+
+            if (words.size() < 2 &&
+                (!channel->isTwitchChannel() || channel->isEmpty()))
+            {
+                channel->addMessage(makeSystemMessage(
+                    "Usage: /popup <channel>. You can also use the command "
```
ALazyMeme, line 708:
```suggestion
                    "Usage: /popup [channel]. You can also use the command "
```

ALazyMeme: Refer here: https://github.com/Chatterino/chatterino2/blob/1d272c399b805f75284bbf6db0511243cd5ccc88/CONTRIBUTING.md#conventions

Felanbird, line 708:
```suggestion
                    "Usage: /popup [channel]. You can also use the command "
```
`[]` optional vs `<>` required - https://github.com/Chatterino/chatterino2/pull/3530#issuecomment-1019238728 for extra reference

Felanbird (approved):
Otherwise Functionality works as intended. 👍

src/controllers/commands/CommandController.cpp
Felanbird: Actually, with the extra explanation, this one might be an actual technicality.

pajlada:
This shares some (but not all) code with [Split::popup](https://github.com/Chatterino/chatterino2/blob/51ece94f588a60c578504309f9eeea22a831f0fa/src/widgets/splits/Split.cpp#L876-L890) - any way we could use the Split::popup() code where applicable? like if the user has not specified a channel name parameter.

pajlada:
I changed it @qooq69 - take a look if this makes sense! Code duplication is back, but it's the only way I could see not having a random Split created and forgotten about

qooq69:
I have made a minor change of using the createwindow() optional value to show the new window. Otherwise looks good to me! :+1:

pajlada (approved):

PR #3530, opened 2022-01-21
Updated contribution guide to include exemptions to the parameters

ALazyMeme:
Pull request checklist:

# Description

In certain scenarios, a parameter is required; whilst also being optional. The response returned should indicate that a parameter is required in order to get the desired outcome on the usage of a command. For example, `/streamlink` is a valid command, in order to open a specific channel; you MUST use the channel parameter, thus making it a required parameter: `/streamlink <channel>`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -223,7 +223,7 @@ Keep the element on the stack if possible. If you need a pointer or have complex
 When informing the user about how a command is supposed to be used, we aim to follow [this standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) where possible.
 
 - Square brackets are reserved for `[optional arguments]`.
-- Angle brackets are reserved for `<required arguments>`.
+- Angle brackets are reserved for `<required arguments>`. There are exemptions in such cases where using the command without a parameter also returns a valid response. In such cases, the parameter becomes required and we should inform users as such: `/streamlink <channel>`.
 - The word _Usage_ should be capitalized and must be followed by a colon.
 - If the usage deserves a description, put a dot after all parameters and explain it briefly.
 


Felanbird:
This is my counter argument ce873aeb241cfca2e540165ae49626b7270a3816
image

sando:
This change seems unnecessarily pedantic - in the current example in CONTRIBUTING.md with the listed conventions, `/streamlink <channel>` implies that I *do* need to specify an argument for the command to work at all which is not the case. `/streamlink [channel]` implies that the argument isn't required for use, but *can* be specified for a change in default behaviour. 

@Felanbird's commit makes more sense to me.

ALazyMeme:
> This change seems unnecessarily pedantic - in the current example in CONTRIBUTING.md with the listed conventions, `/streamlink <channel>` implies that I _do_ need to specify an argument for the command to work at all which is not the case. `/streamlink [channel]` implies that the argument isn't required for use, but _can_ be specified for a change in default behaviour.
> 
> @Felanbird's commit makes more sense to me.

My commit does not oppose felanbird's. Both are technically correct and I believe both should be merged.

zneix (changes requested):
In [utility argument convention](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) it's all explained - an argument is put in angle brackets if invoking a command WITHOUT it results in an error. Invoking `/streamlink` with no arguments succeeds, which makes `[channel]` an OPTIONAL argument, not REQUIRED one.

> Felanbird's commit makes more sense to me.

agreed :+1: although where is that commit made to

ALazyMeme:
> Invoking `/streamlink` with no arguments succeeds, which makes `[channel]` an OPTIONAL argument, not REQUIRED one.

The usage text would only appear if the user somehow types the wrong thing as the channel parameter. Since we are trying to correct the user's actions, it is a required parameter - as they are wanting to open another channel, not open the current channel in streamlink.

I mention again that my commit is not designed to oppose felanbird's - as using `/streamlink [channel]` in the documentation IS correct. If the user is trying to open streamlink for another channel, then the channel parameter becomes required in order to get the desired result.

Felanbird:
One point of note only semi-relevant to this discussion, I believe the usage for `/streamlink` can never be generated because we don't check the input for said command. 
image

KillPainHaters:
I love Chatterino, but does anyone know how to get the Chatterino Special Pepe Badge. lol 😋

pajlada:
CONTRIBUTING.md is not meant to be for users of Chatterino, but to help developers know when to use what type of brackets. Angled brackets imply that the parameter MUST be there, not that the parameter may be there, and if it's not there it will work with a default behaviour anyway.

Because of this, in #3536 I'm updating the usage documentation of streamlink, which was wrong.
Note that I am **not** changing the text in CONTRIBUTING.md because again, **it is not for users**. We could add texts after the example commands with the intent, but that can be another pull request.

Closing in favour of #3536

PR #3536, opened 2022-01-22
Fix /streamlink command usage message

pajlada:
Pull request checklist:

# Description


Mm2PL (approved):

src/controllers/commands/CommandController.cpp
```diff
@@ -663,9 +663,8 @@ void CommandController::initialize(Settings &, Paths &paths)
                 (!channel->isTwitchChannel() || channel->isEmpty()))
             {
                 channel->addMessage(makeSystemMessage(
-                    "Usage: /streamlink <channel>. You can also use the "
-                    "command without arguments in any Twitch channel to open "
-                    "it in streamlink."));
+                    "Usage: /streamlink [channel]. If no channel argument is "
+                    "specified, it will open the current Twitch channel."));
```
zneix, line 667: I guess we can preserve the wording to keep it obvious what will happen, this is a nitpick but I feel like just "will open current channel" is missing something. 
```suggestion
                    "Usage: /streamlink [channel]. If no channel argument is "
                    "specified, it will open the current Twitch channel in streamlink."));
```

pajlada: **BASED** on this suggestion, and more chat offline, we decided on `/streamlink [channel]. Open specified Twitch channel in streamlink. If no channel argument is specified, open the current Twitch channel instead.`

zneix (approved):

Mm2PL (approved):

Felanbird:
I feel like I'm missing something, while I might have tested it on 1d272c39 without these changes (but they aren't code changes) no matter what I input into this command I always receive `Opening <1> in Streamlink ...` 

image
image
image

Unless this is some god of war bug because I don't actually have streamlink setup, but I assume errors should be unrelated.

Mm2PL:
You can only get the usage in an empty channel. It would technically work if the command would be called in IRC but sending stuff on IRC never calls Chatterino commands.

Felanbird (approved):
Wording + `[]` change is good 👍 

Will merge after #3538 - (even tho this already has all checks done 👺)

zneix:
> Unless this is some god of war bug because I don't actually have streamlink setup, but I assume errors should be unrelated.

this might be a side-effect of #3510

Felanbird:
> > Unless this is some god of war bug because I don't actually have streamlink setup, but I assume errors should be unrelated.
> 
> this might be a side-effect of #3510

Seems to be unrelated, I picked a random portable Chatterino I had laying around 3fcb6346f and the issue persists back then, we just don't validate the users input in normal channels.

PR #3544, opened 2022-01-26
Added incremental search to channel search.

qooq69:
Pull request checklist:

# Description
Added incremental search to channel search. Can be toggled in settings under `Use incremental search`.


Infinitay:
Thoughts on having this the default search method to also match emote and tagging auto-completion?

LosFarmosCTL:
I feel like this should be the new default behavior instead of a setting.

If the setting stays though, it should not be called "Incremental search", thats more of a technical term and most users would probably not know what it does. Something in the lines of "Show search results while typing" would be more comprehensible.

qooq69:
My only concern if this becomes the default behavior is that it might be too taxing on low end computers and might lag users.

Regarding the name, my thought process was that people can google the term if they don't fully understand it, though I'm not opposed to changing it.

LosFarmosCTL:
> My only concern if this becomes the default behavior is that it might be too taxing on low end computers and might lag users.

I wouldn't expect this to be especially taxing on your systems resources tbh, but  yeah couldn't hurt to test it in a crappy VM or something.

> Regarding the name, my thought process was that people can google the term if they don't fully understand it, though I'm not opposed to changing it.

Having to look up what the name of a setting means is definitely not a good UX.

sando:
I like it as a setting, 100% agree that "Show search results while typing" is more intuitive.

kornes:
i would vote to make it default behavior and remove setting checkbox, it wont be more taxing than search in few hundred emotes/emojis which is currently the case and had no negative feedback

Felanbird:
I'm fine with implementing this as the new default, as long as it's tested in some way on a low-end machine/vm as there are alot of Chatterino users in recent times without the best computers.

Otherwise a setting is fine, I'm aware we have alot of settings without "advanced" sorting but that doesn't mean we should always stray away from user(s) customization.

but change the setting name cuz the current one sucks.

LosFarmosCTL:
I just tested the changes inside of a 1 CPU Core 2GB RAM crappy VM on my old laptop. Chatterino itself was a bit of a pain to use, but the search did not cause any problems whatsoever and worked absolutely flawless without any lags, hiccups, whatsoever.

So I would strongly advocate for this to be the default behaviour and not another new setting, since I can't really see any reason why anyone would need the old behaviour.

pajlada:
I agree with @LosFarmosCTL 's comment - removing setting and keeping this default is the right way to do this.
If you want to be fancy, you could add a benchmark!

Felanbird (approved):

pajlada (approved):

pajlada:
@Mm2PL Merge in when you're happy!

Mm2PL (approved):
Looks good to me. Maybe we should look at completion/suggestions of parsing search predicate later.

Issue #3545, opened 2022-01-27
Faded text is difficult to read


brawaru:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Currently, when you open Chatterino, all the history before you connected will be faded (has half opacity). I understand it was meant to signify that it is history, however, I feel like this is an inaccessible solution. Faded out text is very difficult to parse if you rely on text contrast due to poor vision or if you have reading problems and need to see characters clearly. One could say ‘oh just adjust your monitor settings’, but this does not improve the situation at all and actually makes the image for anything else worse. 

This issue becomes even more severe when the person who was in history gets muted, Chatterino then double fades out their messages, so they become almost invisible.

### Screenshots

![Screenshot demonstrating contrast differences between messages in history and messages sent after the connection is established](https://user-images.githubusercontent.com/10401817/151306694-01493f42-5c1c-46a3-b1c4-0a6126d956b4.png)

### OS and Chatterino Version

Chatterino Nightly 2.3.4 (commit e2eb73d81, built on ); Windows 10 20H2 (19042)

Brawaru:
Thinking of solutions that can solve this:
- Have an option to use a tiny strikethrough effect for muted accounts instead of fading.
- Do not fade away the history at all, instead create a special block line that would separate history from the actual chat.
   
   E.g. [~~​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​~~ ▲ Chat history before 27/01/22 07:44 ▲ ~~​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​~~]

Infinitay:
This option is available on [Dankerino's](https://github.com/Mm2PL/dankerino) fork
![image](https://user-images.githubusercontent.com/6964154/151308648-ee2fdf4d-df15-4120-afad-23b4c6877086.png)

[...]

PR #3546, opened 2022-01-28
Add option to control whether to gray out history of messages

dead-tech:
I think this should do it for #3545.

This is what the setting looks like in the General Page: 
settings_generalpage

And this is the setting in action:
Screenshot from 2022-01-25 17-10-58

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -838,11 +838,16 @@ void TwitchChannel::loadRecentMessages()
 
             if (!getSettings()->grayOutHistory)
             {
-                const auto date = QDateTime::currentDateTime().toString("dd/MM/yy");
-                const auto time = QTime::currentTime().toString("hh:mm");
-                const auto msg = QString("[​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ▲ Chat history before %1 %2 ▲ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​]").arg(date).arg(time);
-                const auto block_line = makeSystemMessage(msg, QTime::currentTime());
-                allBuiltMessages.emplace_back(block_line);
+                const auto str =
+                    QString(
+                        "[​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ "
+                        "​ ​ ​ ​ ​ ​ ​ ​ ​ ▲ Chat history "
+                        "before %1 %2 ▲ ​ ​ ​ ​ ​ ​ ​ ​ ​ "
+                        "​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​]")
+                        .arg(QDateTime::currentDateTime().toString("dd/MM/yy"),
+                             QTime::currentTime().toString("hh:mm"));
+                const auto msg = makeSystemMessage(str);
+                allBuiltMessages.emplace_back(msg);
             }
 
             postToThread([this, shared, root,
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -836,6 +836,15 @@ void TwitchChannel::loadRecentMessages()
                 }
             }
 
+            if (!getSettings()->grayOutHistory)
+            {
+                const auto date = QDateTime::currentDateTime().toString("dd/MM/yy");
+                const auto time = QTime::currentTime().toString("hh:mm");
+                const auto msg = QString("[​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ▲ Chat history before %1 %2 ▲ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​]").arg(date).arg(time);
+                const auto block_line = makeSystemMessage(msg, QTime::currentTime());
+                allBuiltMessages.emplace_back(block_line);
+            }
+
             postToThread([this, shared, root,
                           messages = std::move(allBuiltMessages)]() mutable {
                 shared->addMessagesAtStart(messages);
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -288,7 +288,8 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addColorButton("Line color",
                           QColor(getSettings()->lastMessageColor.getValue()),
                           getSettings()->lastMessageColor);
-    layout.addCheckbox("Gray-out message history before connecting.", s.grayOutHistory);
+    layout.addCheckbox("Gray-out message history before connecting.",
+                       s.grayOutHistory);
 
     layout.addTitle("Emotes");
     layout.addCheckbox("Enable", s.enableEmoteImages);
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -288,6 +288,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addColorButton("Line color",
                           QColor(getSettings()->lastMessageColor.getValue()),
                           getSettings()->lastMessageColor);
+    layout.addCheckbox("Gray-out message history before connecting.", s.grayOutHistory);
 
     layout.addTitle("Emotes");
     layout.addCheckbox("Enable", s.enableEmoteImages);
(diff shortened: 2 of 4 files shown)


Infinitay:
FWIW you have the wrong screenshot posted for
> And this is the setting in action:

LosFarmosCTL:
> And this is the setting in action:

Wrong screenshot aside, having just compiled this myself, there are 2 things I dislike about how it works at the moment. 

**Screenshot for reference:**
Chatterino 2022-01-28 at 15 43 55

1. The message is too big when used in smaller spits. I would personally just remove the extra space inside of the square brackets, doesnt really look that great in bigger chats either.
2. Duplicate timestamps: No reason to have a timestamp in the message itself, if its just a normal system message that displays a timestamp anyways. While I feel like this is another one of those situations where the message would probably look better without the entire timestamp element (can't find the issue or PR where that discussion happened, but it was regarding automod messages and how they basically introduce a second timestamp in addition to the one from the actual message), I am not sure if there is a way to do that "correctly" right now, so might be OOS for this PR. But either way, if you ask me, one of the timestamps has to go.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -836,6 +836,20 @@ void TwitchChannel::loadRecentMessages()
...
 
+            if (!getSettings()->grayOutHistory)
+            {
+                const auto str =
+                    QString(
+                        "[​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ "
+                        "​ ​ ​ ​ ​ ​ ​ ​ ​ ▲ Chat history "
+                        "before %1 %2 ▲ ​ ​ ​ ​ ​ ​ ​ ​ ​ "
+                        "​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​ ​]")
+                        .arg(QDateTime::currentDateTime().toString("dd/MM/yy"),
```
Mm2PL, line 847: I'd suggest you just set the timestamp on this message instead of using a hardcoded date format. As for the spaces, please use the `Center` `MessageFlag` instead if you for the center effect. Using spaces is inconsistent because Chatterino allows setting non-monospace fonts. It can also look bad with small width-wise windows when it's wrapping to multiple lines.

Consider using [this ↓](https://github.com/Chatterino/chatterino2/blob/master/src/messages/MessageBuilder.cpp#L26) version of `makeSystemMessage` to set the timestamp in the generated message.
```cpp
MessagePtr makeSystemMessage(const QString &text, const QTime &time);
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -288,6 +288,8 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addColorButton("Line color",
                           QColor(getSettings()->lastMessageColor.getValue()),
                           getSettings()->lastMessageColor);
+    layout.addCheckbox("Gray-out message history before connecting.",
```
Mm2PL, line 291: I feel like this isn't descriptive of what the setting does, as it also enables adding a separator after message history not only disables the gray tinting.

@Felanbird, could you help? :)

Mm2PL (changes requested):
I've outlined some things I'd change about the code itself. I've also found a little problem, after disabling the setting, the history is imminently set as not tinted but the message isn't added, effectively creating four states:
  - history grayed-out, no message (current behavior, started up with setting enabled)
  - history grayed-out, message (started up with setting disabled then enabling it)
  - history not grayed-out, no message (started up with setting enabled then disabling it)
  - history not grayed-out, message (started up with setting disabled)

Should we even care about this? I honestly don't know.

src/providers/twitch/TwitchChannel.cpp
dead-tech: I'll go ahead and make these changes, the hard coded string I have used was taken from one of the suggestions in the issue so i thought it was fine to use that.

dead-tech: Also since we remove the hard coded date format from the string, in my opinion the message displayed should also change as "Chat history before" does not make much sense anymore. Any suggestions?

kornes:
I think adding extra message as separator is unnecessary noise. Simple permanent line (maybe a bit thicker than the one used for new messages now) would imo look better for this.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -288,6 +288,38 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addColorButton("Line color",
                           QColor(getSettings()->lastMessageColor.getValue()),
                           getSettings()->lastMessageColor);
+    layout.addSeperator();
+    layout.addCheckbox("Gray-out message history before connecting.",
+                       s.grayOutHistory);
+    layout.addCheckbox("Draw a line below the last message before "
+                       "connecting.",
+                       s.showBeforeConnectingIndicator);
```
Felanbird, line 296:
```suggestion
    layout.addCheckbox(
        "Draw a line below the last message loaded from message history.",
        s.showBeforeConnectingIndicator);
```
something like this? I just included the formatted version(s) encase it's implemented via suggestions

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -288,6 +288,38 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addColorButton("Line color",
                           QColor(getSettings()->lastMessageColor.getValue()),
                           getSettings()->lastMessageColor);
+    layout.addSeperator();
+    layout.addCheckbox("Gray-out message history before connecting.",
+                       s.grayOutHistory);
```
Felanbird, line 293:
```suggestion
    layout.addCheckbox("Gray-out message history.", s.grayOutHistory);
```
I can't really think of a better name for this, I just cut the `before connecting` part 🤷

Felanbird (changes requested):
So here's what I noticed:

As of the current code The functionality of the tinting enable/disable works as intended (without the need of a reconnect/restart.)

The line separator on the other hand, does require a reconnect, which is not consistent with the existing settings `Separate with lines` & `Draw a line below the most recent message before switching applications` as well as the tint changes in this PR.

I do think the existing standard, plus the already functioning tinting, should be used and the `Draw a line below msg history...` should work via button press without the need for a reconnect.

Felanbird:
To prevent more code changes to the current implementation I'll include these:

image
image

TranRed:
I just want to mention that this was already a setting. It got implemented in various forms (#822 and #933) and then later. I think it was removed because the red stripes introduced in #933 felt disruptive. 

Pajlada's comment/suggestion https://github.com/Chatterino/chatterino2/pull/822#issuecomment-432973880seems seems still valid and very relevant to me and seems a very good way to make it more/less/non-faded according to the user's needs.

pajlada:
More options: Add special badge to historic messages, do something with the timestamp, maybe add a small icon next to the timestamp?

The one thing I'm sure about is that we should not add another setting for this.

dead-tech:
> Ideas that came up: https://www.viragegroup.com/wp-content/uploads/2021/11/reporting-en-direct.svg image image
> 
> image
> 
> I think the icon should be fairly discrete, and have a tooltip with some more information

Thank you for your time! I'll look into this tomorrow!

pajlada:
Tested out some badges, and it looks pretty bad.
What I've been experimenting with instead now is making the fade for historic messages a lot lighter, and disabling the "double fade" that would occurr when a historic message was disabled (e.g. due to being timed out)

```patch
diff --git a/src/BaseTheme.cpp b/src/BaseTheme.cpp
index 4c58ccba..0a7a4aa0 100644
--- a/src/BaseTheme.cpp
+++ b/src/BaseTheme.cpp
@@ -165,6 +165,7 @@ void AB_THEME_CLASS::actuallyUpdate(double hue, double multiplier)
     // this->messages.backgrounds.resub
     // this->messages.backgrounds.whisper
     this->messages.disabled = getColor(0, sat, 1, 0.6);
+    this->messages.historic = getColor(0, sat, 1, 0.35);
     // this->messages.seperator =
     // this->messages.seperatorInner =
 
diff --git a/src/BaseTheme.hpp b/src/BaseTheme.hpp
index 2d6ee5cd..61b8bb2b 100644
--- a/src/BaseTheme.hpp
+++ b/src/BaseTheme.hpp
@@ -71,6 +71,7 @@ public:
         } backgrounds;
 
         QColor disabled;
+        QColor historic;
         //        QColor seperator;
         //        QColor seperatorInner;
         QColor selection;
diff --git a/src/messages/layouts/MessageLayout.cpp b/src/messages/layouts/MessageLayout.cpp
index 1135556e..f8bfef91 100644
--- a/src/messages/layouts/MessageLayout.cpp
+++ b/src/messages/layouts/MessageLayout.cpp
@@ -212,11 +212,10 @@ void MessageLayout::paint(QPainter &painter, int width, int y, int messageIndex,
         //        painter.fillRect(0, y, pixmap->width(), pixmap->height(),
         //                         QBrush(QColor(64, 64, 64, 64)));
     }
-
-    if (this->message_->flags.has(MessageFlag::RecentMessage))
+    else if (this->message_->flags.has(MessageFlag::RecentMessage))
     {
         painter.fillRect(0, y, pixmap->width(), pixmap->height(),
-                         app->themes->messages.disabled);
+                         app->themes->messages.historic);
     }
 
     if (!isMentions &&
```

c-jaenicke:
Refined icon exampels and ideas,

1. bright
2. dark 1
3. dark 2
 
A quick mockup of them being used:
history

Issue #2005, opened 2020-09-27
Right click menu for chat messages only appears when right clicking text or emotes, not whitespace


azizLIGHT:
**Describe the bug**
When trying to copy a message in chat, right-clicking the whitespace in the line does nothing. In order to get the copy menu, you must right-click the letters in the text of the message or any emotes. This requires precision.

**To reproduce**
Open any active chat, the faster the better.
Right click whitespace, or space between words
Do not see a right click menu with options: `Copy message` and `Copy full message`

**Screenshots**

This is the menu I am expect to get but do not get, when I right click whitespace
![image](https://user-images.githubusercontent.com/8013867/94370088-c8a3e280-00bb-11eb-85f6-289b4fc43798.png)

**Chatterino version**

Chatterino 2.2.2 (commit 53b18d03c)

**Operating system**

Manjaro 20.1 Mikah

**Additional information**

The majority of the pixels in an individual chat line is whitespace (as opposed to text) and is not capable of causing right clicks at all, this narrows the pixels required for right click activation to half or even less. The space character between words doesn't count as right clickable either. It becomes worse if you use small fonts or zoom out or both, or if the channel moves extremely fast, and you don't enable hover mouse to pause the chat, or if the majority of the text messages are single emotes and hard to copy since everyone's nicks are random-length and the emote moves left and right on every line change as its spammed. Just difficult to copy messages unless you're the shroud of chatterino

Mm2PL:
Fixed in #3566

PR #3566, opened 2022-02-09
Show context menu anywhere in MessageLayout when applicable

Infinitay:
Pull request checklist:

# Description

QoL change to make it so you are able to view a messages context menu right-clicking anywhere in that particular message _split_/dimensions. Currently, you are only able to right click on `MessageLayoutElement`'s in order to view the context menu. The current behavior is annoying to deal with if you want to copy a message but manage to right click between words ignoring the event. This makes it so you are essentially able to right click anywhere within the `MessageLayout` and view the respective context menu for the given message.

<details>
<summary>View an example of `MessageLayoutElement` bounds</summary>

Pardon the poor example though I believe it gets my point across. The green areas are where you are able to right click and view the respective context menu prior to this PR.

image

</details>

https://user-images.githubusercontent.com/6964154/153120725-a38808f4-f72b-43bf-bc4c-85b578a99d52.mp4


sando:
Functionality works as intended. Also tested to ensure separate right-click behaviour on usernames, hyperlinks, emotes is preserved.

Closes #2005.

chatterino_KClUcIXIXi

src/widgets/helper/ChannelView.cpp
```diff
@@ -1860,128 +1854,163 @@ void ChannelView::addContextMenuItems(
 
     auto menu = new QMenu;
     previousMenu = menu;
-
-    // Badge actions
-    if (creatorFlags.hasAny({MessageElementFlag::Badges}))
+    if (hoveredElement != nullptr)
```
jammehcow, line 1857: Wrapping this huge chunk in an `if/else` statement reduces readability of an already large method. I understand why you've done it, looks painful. Someone else might have a better idea on how to clean this up.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 - Minor: Removed timestamp from AutoMod messages. (#3503)
 - Minor: Added ability to copy message ID with `Shift + Right Click`. (#3481)
 - Minor: Colorize the entire split header when focused. (#3379)
+- Minor: Show context menu anywhere in MessageLayout when applicable (#3566)
```
jammehcow, line 56:
```suggestion
- Minor: Show right click context menu anywhere within a message's line. (#3566)
```

Not my best wording but easier to understand for a user.

jammehcow (changes requested):
A few nitpicks/thoughts.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1791,13 +1786,14 @@ void ChannelView::handleMouseClick(QMouseEvent *event,
             }
 
             auto &link = hoveredElement->getLink();
-            if (!getSettings()->linksDoubleClickOnly)
+            if (hoveredElement != nullptr &&
+                !getSettings()->linksDoubleClickOnly)
```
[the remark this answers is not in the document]
Infinitay: I realized and considered this after the initial commit, but I didn't bother because I wanted to keep the same _style_ as I had to null check for the next switch case as well and handle it appropriately.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1813,7 +1809,7 @@ void ChannelView::handleMouseClick(QMouseEvent *event,
             };
 
             auto &link = hoveredElement->getLink();
-            if (link.type == Link::UserInfo)
+            if (hoveredElement != nullptr && link.type == Link::UserInfo)
```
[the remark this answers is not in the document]
Infinitay: I personally would like to have do it as you describe but for the same reason of readability I didn't.

src/widgets/helper/ChannelView.cpp
Infinitay: We could instead handle the logic in a `if (hoveredElement == nullptr)`, but then the menu action order would be different than how it normally is (copy message before all other actions).

IIRC there is goto labels in C? Could probably make use of that in order to handle the logic as `==` and retain the menu action ordering, but I'm not sure if it's the best way to go about it.

pajlada (approved):
:+1: I'm happy with the change and the structure of the code. Would like @jammehcow's opinion again to double check that his concerns have been taken care of before merging this in.

jammehcow (approved):
Changes are fine by me!

zneix (approved):

PR #3573, opened 2022-02-20
Updated build instructions

ooxi:
Pull request checklist:

- [X] `CHANGELOG.md` was updated, if applicable (not applicable)

# Description

Building on Ubuntu 18.04 no longer works as documented

```
CMake Error at lib/settings/CMakeLists.txt:132 (message):
  GCC version must be at least 8.0 if using std filesystem
```

I verified the build instructions on Ubuntu 20.04 and they work as advertised.


pajlada (approved):
:+1: The aim is to support the latest Ubuntu LTS primarily, so this change is all good from my perspective. It's definitely still possible to compile on Ubuntu 20.04 still with some extra tweaks here or flags there but I think this is the correct approach still.

PR #3578, opened 2022-02-26
Purged kraken

zneix:
Pull request checklist:

# Description

rip


Mm2PL:
hail helix, i guess

zneix:
Since we no longer have any Twitch API implementations scattered across different files, like it used to be a thing with Kraken implementation in the distant past, I removed `"we implement this in X"` since all Helix implementations were, are and always be in Helix.cpp. I remember discussing removal of that line with pajlada once and he agreed on that - just explaining my extra, (sorta) unrelated changes.

pajlada (approved):
:+1:

zneix:
I came to a conclusion this is nothing we introduce - Kraken would die and changes would be visible to pretty much everyone, so there's no changelog entry needed anymore, I don't care... could as well just say "twitch broke things" but that isn't anything new.

PR #3580, opened 2022-02-27
:construction_worker: Fix CI build with VS 2022

AnotherFoxGuy:
Pull request checklist:

# Description

Fixed the CI build with `windows-latest`
The first build will now take around 30 min, but subsequential builds will be faster once the cache has been created


Felanbird (approved):
This is obviously a better approach, I only went with downgrading to `windows-2019` because I figured nobody actually cared how we were doing windows builds.

pajlada (approved):

PR #3599, opened 2022-03-08
Update PubSub client to use TLS v1.2

LosFarmosCTL:
Pull request checklist:

# Description

TLS v1.0 (what chatterino currently uses) and TLS v1.1 are deprecated, no longer supported by major web browsers and shouldn't be used anymore.


jammehcow:
TLS 1.2 support was introduced way back in boost asio, won't break any existing distributions. Should be fine to bump.

jammehcow (approved):

Felanbird (approved):
This change fixes an issue a user had with being unable to see pubsub mod actions due to the TLS handshake silently failing.
image

LosFarmosCTL:
> This change fixes an issue a user had with being unable to see pubsub mod actions due to the TLS handshake silently failing.

Yeah, this is exactly how I found out that chatterino is still using TLS v1.0 in the first place, I have the exact same issue. No idea why this fixes it or what even causes the problem in the first place, but since its an outdated protocol I figured bumping it would in general be a good idea.

pajlada (approved):

PR #3602, opened 2022-03-09
chore: use system logical core count for Ubuntu/macOS builds

jammehcow:
Pull request checklist:

# Description

This PR aims to replace the hardcoded `-j8` make flag with a logical core count lookup. 

Ubuntu uses nproc (available on Linux), whereas macOS has to use a sysctl call. Could also use `getconf _NPROCESSORS_ONLN` as it returns the same result on both platforms.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,7 @@ jobs:
               -DPAJLADA_SETTINGS_USE_BOOST_FILESYSTEM=On \
               -DUSE_PRECOMPILED_HEADERS=${{ matrix.pch }} \
               ..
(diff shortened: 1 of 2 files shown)


jammehcow:
Note: Windows already sets an envvar to max out job count, no change needed.

Mm2PL:
Shouldn't just `make -j` work?

jammehcow:
> Shouldn't just `make -j` work?

image
Will leave it as is :)

Mm2PL:
Yeah don't use -j, make spawned so many processes my X and PulseAudio stopped responding lol

pajlada (approved):

PR #3622, opened 2022-03-22
Fixed and updated OpenSSL Windows links in building docs

brian6932:
Pull request checklist:

# Description

Links were broken, seems like there is no live exe for `1.0.2u`, so I swapped it over to an archive, there is a zip, although I kept it this way to avoid any re-documenting

### DIFF
--- a/BUILDING_ON_WINDOWS.md
+++ b/BUILDING_ON_WINDOWS.md
@@ -33,13 +33,13 @@ Note: This installation will take about 1.5 GB of disk space.
 
 ### For our websocket library, we need OpenSSL 1.1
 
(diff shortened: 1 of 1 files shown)


jammehcow:
image I thought I had an alert for OpenSSL version changes, nice pick!

kornes:
Qt ssl should work with openSSL 1.1.1+ from 5.15 according [to docs](https://doc.qt.io/qt-5/ssl.html), so seems like whole paragraph can be dropped?

pajlada (approved):
Looks fine, should revisit OpenSSL version requirements in a separate PR as @kornes suggested

PR #3623, opened 2022-03-22
feat: use message's original channel name when usercard popout is opened

### LINKED ISSUE #2248: Can't access usercard from /mentions tab
**Describe the bug**

When attempting to access the twitch usercard from the /mentions tab, the link fails to correctly generate as there is no "channel" declared for that part of the URL

**To reproduce**

1. Click on a user in /mentions tab
2. Click on 'Usercard'
3. See the twitch error ghost

**Screenshots**
image

image

**Chatterino version**
Tested on 9ebc448bc just to be 100% sure it was never fixed

**Operating system**
Windows 10

**Additional information**
This was an issue once before, the search window lacked channel context as well, which was fixed in 67aadfe

Possibly able to use some sort of context from 49fa9bf ? idk

jammehcow:
Pull request checklist:

# Description

This PR aims to fix an issue where opening a user's popout from the usercard in a non-user channel (/mentions) doesnt open use the channel's name. This caused the popout to have an invalid path and 404.

I'm not saying my fix is 'correct', but it's the best I can do without either passing too much info down or reworking it for the sake of a popout.

Fixes #2248


jammehcow:
oh yeah, fixes #2248

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@
 - Minor: Make Tab Layout setting only accept predefined values (#3564)
 - Minor: Added librewolf, icecat, and waterfox incognito support. (#3588)
 - Minor: Updated to Emoji v14.0 (#3612)
+- Minor: Use original channel when opening a user's popout in /mentions (#3623)
```
Felanbird, line 63: ```Bugfix: Fixed being unable to load Twitch Usercards from the `/mentions` tab. (#3623)```

Felanbird (approved):
Functionality works as intended, fixing the original issue. 👍

CHANGELOG.md
jammehcow: Much better wording, but made it minor.

kornes:
requests to ivr api are also currently broken on /mentions tab due to same underlying issue. Can be fixed in scope of this PR here: 
https://github.com/Chatterino/chatterino2/blob/146816aa1a75202fc016e544625c867a77cf060a/src/widgets/dialogs/UserInfoPopup.cpp#L858

jammehcow:
> requests to ivr api are also currently broken on /mentions tab due to same underlying issue. Can be fixed in scope of this PR here:

Good catch, thanks! Will fix it later today.

kornes:
and one more with same problem - middle mouse button on username
https://github.com/Chatterino/chatterino2/blob/495f4b15f46818104cacf65796b17fbd3a38fe2d/src/widgets/helper/ChannelView.cpp#L1719

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -385,7 +385,7 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, QWidget *parent)
 
         QObject::connect(usercard.getElement(), &Button::leftClicked, [this] {
             QDesktopServices::openUrl("https://www.twitch.tv/popout/" +
```
kornes, line 387: call this instead https://github.com/Chatterino/chatterino2/blob/fb9c3ad42ba4a7361d9c8889a1230a2baf71b1d8/src/util/Twitch.cpp#L8

jammehcow: Thanks. I'll do a more thorough check before I push that.

jammehcow:
I'm wondering if it's worth rejigging this to pass through the original `ChannelPtr` instead of just a name. That way the snapshot of a user's messages can also be resolved for the original channel, mod actions work etc.

@Mm2PL thoughts?

src/widgets/dialogs/UserInfoPopup.cpp
jammehcow: No longer required, reword channel passing

Felanbird:
I was holding-off on reviewing this PR as I wasn't sure if I wanted the `/mentions` tab context to be replaced (for searching through multiple days of mentions from the same user), but I think it's fine for channel context to be used instead.

jammehcow:
There should've been a nullptr check when trying to get the channel from a weak pointer - I wouldn't expect it to crash. Will look into it later today

jammehcow:
@Felanbird should be resolved. If the channel wasn't found I was passing an empty channel but not checking that on the other side. Have tested and works.

Felanbird (approved):
Crash is fixed - and the variation I was secretly hoping for exists. 
When the tab is closed it seems to fallback to the pre-PR behavior which is great.

image
image

pajlada (approved):

PR #3626, opened 2022-03-23
docs(building): add instructions for CLion on Windows

Nerixyz:
Pull request checklist:

# Description

Since the documentation for CMake was very sparse and CLion can't use QMake it wasn't straight forward to set up CLion for me. I want to change that for future visitors and for users of other editors requiring CMake (maybe documentation for setting up VSCode sould be added as well).

I noticed there isn't any mention of conan in the documentation for Windows. For me it was really easy setting it up and having it download files to another drive by changing its configuration. Also the `Using CMake` section doesn't mention conan which I beleive is an error since `USE_CONAN` is set to `ON`.

[Rendered documentation](https://github.com/Nerixyz/chatterino2/blob/windows-clion-docs/BUILDING_ON_WINDOWS.md#buildingrunning-in-clion)


pajlada (approved):
:+1: 
Would just like the imgur links replace with github user content links, but Felanbird has offered to do so after some other PRs have been merged in to save on CI hours

Felanbird (approved):
👍

PR #3627, opened 2022-03-24
Add the ability to lock NotebookTab layout

Infinitay:
Pull request checklist:

# Description

Gives users the ability to lock their tabs on Chatterino to prevent accidental movement.

https://user-images.githubusercontent.com/6964154/159972762-331d7f15-3baa-4f8d-b007-5ff848b4d5d4.mp4

# Additional Comments

I'm not sure if it would be better practice to rely only on the `ChatterinoSetting` rather than also using a private variable `lockNotebookLayout_`. Would appreciate some guidance.

### DIFF
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -41,22 +41,22 @@ Notebook::Notebook(QWidget *parent)
         },
         QKeySequence("Ctrl+U"));
 
-    lockNotebookLayoutAction_ = new QAction("Lock Tab Layout", &this->menu_);
+    this->lockNotebookLayoutAction_ = new QAction("Lock Tab Layout", &this->menu_);
 
     // Load lock notebook layout state from settings
     this->setLockNotebookLayout(getSettings()->lockNotebookLayout.getValue());
 
-    lockNotebookLayoutAction_->setCheckable(true);
-    lockNotebookLayoutAction_->setChecked(this->lockNotebookLayout_);
+    this->lockNotebookLayoutAction_->setCheckable(true);
+    this->lockNotebookLayoutAction_->setChecked(this->lockNotebookLayout_);
 
     // Update lockNotebookLayout_ value anytime the user changes the checkbox state
-    QObject::connect(lockNotebookLayoutAction_, &QAction::triggered,
+    QObject::connect(this->lockNotebookLayoutAction_, &QAction::triggered,
                      [this](bool value) {
                          this->setLockNotebookLayout(value);
                      });
 
     // Append it to our current menu actions
-    this->menu_.addAction(lockNotebookLayoutAction_);
+    this->menu_.addAction(this->lockNotebookLayoutAction_);
 }
 
 NotebookTab *Notebook::addPage(QWidget *page, QString title, bool select)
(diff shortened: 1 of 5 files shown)


Infinitay:
I considered adding the menu action to `NotebookTab`s themselves, but midway I realized maybe it wouldn't be an ideal implementation. Originally, I considered giving each `NotebookTab` its own respective menu action. However, that would mean storing them all and updating them when the layout is (un)locked.

Another way I suppose would be to instead to re-use the menu action created it `Notebook` and add it to the `NotebookTab`s.

Would be a QoL to add the option to the `NotebookTab`s allowing users to (un)lock the layout with ease rather than clicking the on the _blank space_ in the header. Additionally, it would follow inline with the existing ShowTabs (_Toggle Visibility of Tabs_) menu action being added to the _blank space_ in the header **and** every `NotebookTab`.

Felanbird:
> Would be a QoL to add the option to the `NotebookTab`s allowing users to (un)lock the layout with ease rather than clicking the on the [_blank space_](https://user-images.githubusercontent.com/6964154/159972361-4957450a-14f6-4883-a290-172ae1aa388a.png) in the header. Additionally, it would follow inline with the existing ShowTabs (_Toggle Visibility of Tabs_) menu action being added to the _blank space_ in the header **and** every `NotebookTab`. On the other hand, would reduce the menu action clutter if we just keep it only in `Notebook`/the [_blank space_](https://user-images.githubusercontent.com/6964154/159972361-4957450a-14f6-4883-a290-172ae1aa388a.png).

This is the important discussion, as visibility of this feature is the worry against it - the more visible it is the more likely somebody will accidentally click it.
From my experience with `toggle visibility of tabs`, users who have accidentally pressed it do not think to click in the empty spot where their channels used to be. My immediate assumption would be that this pattern would continue, especially if your tabs are still there, thinking to click the small empty area would be overlooked.

Infinitay:
Personally I'm conflicted on adding a button as a _tab button_ for a visual indication. It would be nice, and I suppose in most cases it is more than fine. However, it would be annoying to have a special case where only the additional button would cause a new line where currently it wouldn't.

image

Or I suppose the true behavior is that the new _te_ tab would be moved down on the new line and the _+_ and _🔒_ buttons next to it to fit all three together.

Although again, this is my only complaint with it and thought I should bring it up despite it probably not being the case for the majority of Chatterino users. Personally I rename and reorganize my tabs to make sure theres not just one or two tabs on a new line to free some space for chat viewability.

src/widgets/Notebook.cpp
```diff
@@ -35,11 +35,28 @@ Notebook::Notebook(QWidget *parent)
     this->addButton_->setHidden(true);
 
     this->menu_.addAction(
-        "Toggle visibility of tabs",
+        "Toggle Visibility of Tabs",
```
Felanbird, line 38: image

Infinitay: Following the existing style for `NotebookTab` actions.

image

Channel actions on the other hand are a different case
image

Felanbird: but this is not a PR for visibility of tabs now is it 🙂

Infinitay: Fair enough

pajlada (approved):
:+1: 
Added a commit to fix it not saving in settings
Added the menu item to the notebook tabs - I think it's worth it (see felanbird's comment)
Did some minor refactoring

Felanbird:
Also I'll accept that visibility of tabs casing change as a separate PR :)

PR #3630, opened 2022-03-25
Open a new tab/split by shift/alt + middle clicking a username

badoge:
Pull request checklist:

# Description
Just more ways to quickly open new tabs/splits :)

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1718,17 +1718,25 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
             {
                 if (event->modifiers() == Qt::ShiftModifier)
                 {
-                    this->openChannelIn.invoke(
-                        hoverLayoutElement->getLink().value,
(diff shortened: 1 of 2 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -1716,6 +1716,22 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
             else if (hoverLayoutElement->getFlags().has(
                          MessageElementFlag::Username))
             {
+                if (event->modifiers() == Qt::ShiftModifier)
```
Infinitay, line 1719: Perhaps check for the Alt modifier condition here as well instead of having two blocks?

pajlada (commented):
Is there any previous art/other similar software that handles shift+middle click or ctrl+middle click in similar ways? otherwise this feels very yolo-added

We already have things that are yolo-added, so I'm just trying to make sure that new things we add have some proper logic/documentation/discoverability/configurability to them.

PR #3633, opened 2022-03-26
Add an action to select the word under the cursor

Mm2PL:
See #3628

Pull request checklist:

# Description

Adds a new action to Split Input which selects the word under the cursor. There is no default binding, you have to add a hotkey yourself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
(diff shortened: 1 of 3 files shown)


jammehcow (approved):
Tested, works as expected!

Felanbird (approved):
Works as expected, comparing to the 20 seconds I used this feature in vscode after GaZaTu suggested it.

pajlada (approved):

PR #3634, opened 2022-03-27
Add guide for building chatterino2 on Windows with vcpkg

Ian321:
…e, msbuild and msvc

Pull request checklist:

# Description

- Wrote a guide on how to build the project using vcpkg on windows.
- Fixed build errors as they came up


pajlada:
Looks good! Once you think this is finished I'd love to have the guide combined into the original Windows guide as a subcategory instead of its own document. Let me know if you think this doesn't make sense

Ian321:
The current BUILDING_ON_WINDOWS.md is already 190 lines long and I thought of this as a clean new guide, as almost nothing from the current one applies to this.

The 2 tasks not done are:
- [ ] automated build
- [ ] fixing the debug build

vcpkg.json
```diff
@@ -0,0 +1,19 @@
...
+  "name": "chatterino",
+  "version": "2.0.0",
+  "dependencies": [
+    "benchmark",
+    "boost-foreach",
+    "boost-interprocess",
+    "boost-random",
+    "boost-variant",
+    "gtest",
+    "openssl",
```
kornes, line 12: openssl defaults to 3.0.2 from vcpkg, had to lock it to 1.1.1 while testing to produce working build
```
  "overrides": [
        { "name": "openssl", "version-string": "1.1.1n" }
    ]
```
other than that works fine and would put it as preferred build way for windows 👍

Ian321: Could you share some logs? I had no issues with the default version.

kornes: it did build, but failed on all network https connections on startup. Figured it might be openssl since conan is also locked to 1.1.1, didn't troubleshoot further after it worked.

Ian321: No need, it somehow used the openssl outside of vcpkg (which was 1.1.1). I updated the dependencies.

pajlada (approved):

PR #3643, opened 2022-04-02
Improve Twitch PubSub connection reliability

Nerixyz:
Pull request checklist:

# Description

Improve connection stability

## Additional changes

- To simplify "json enum" to C++ enum, the `magic_enum` library has been added as a dependency
- PubSubClient moved to its own file
- PubSub Messages are mostly built as structs initially, and then finally serialized into a json string as needed
- PubSub messages now all have their own file
- Mentions of `Pubsub` have been changed to `PubSub` to be consistent with capitalization
- TwitchAccountManager accountUpdated signal now uses boost2 so we can define a callback order (This fixed account-changing moderation action subscribe/unsubscribing)
- We now unsubscribe from whispers when the Twitch account changes
- use QJson instead of rapidjson where possible

## Old description

This prevents a deadlock if a connection fails since `addingClient` will never be set to `false`.

The [documentation for websocketpp](https://docs.websocketpp.org/reference_8handlers.html#autotoc_md114) says:
> Either open or fail will be called for each connection. **Never both.** Connections that fail will **never have a close handler called.**

Since we don't listen for fails, `addingClient` will be `true` forever if the first connection fails, effectively resulting in a deadlock (the `PubSub topic backlog` will never decrease). This PR attempts to fix this bug.

There's a new bug however: If the connection will always be closed, then we will spam Twitch with requests. Maybe this should be throttled?


Felanbird:
__________
I've tested the current code for the past 4 days and had no issues, both redemptions and automod continue to work in comparison to before where they'd just ⚰️. 
As well tested how the connection would handle a VPN being enabled and disabled (as this killed the connection very consistently before), and it seems to work as well.
also mod actions work too but I only tested them a few minutes ago cuz I was just thinking of the important ones most users see.

Felanbird:
`Minor: Better handle Pubsub reconnections. This will help prevent certain functions appearing to be broken [Automod/Channelpoint Redemptions/Mod Actions]. (#3643)`

I do think this is a bugfix, but the concept of this fix belongs as high up as it can be in the changelog, I'd probably stick it at the top of the changelog (because apparently people actually do read the changelog, who knew.)

src/providers/twitch/PubSubClient.cpp
```diff
@@ -0,0 +1,214 @@
...
+        qCDebug(chatterinoPubSub)
+            << "Client::close(): Error getting con:" << ec.message().c_str();
+        return;
+    }
+
+    conn->close(code, reason, ec);
+    if (ec)
+    {
+        qCDebug(chatterinoPubSub)
+            << "Client::close(): Error closing:" << ec.message().c_str();
```
zneix, line 62: minor nitpick: "hardcoding log message origins" is something that should be avoided since we make use of logging categories, which with `QT_MESSAGE_PATTERN` environment variable can show exact origin of the log message.
Example content of mentioned variable which is helpful: `"[%{time hh:mm:ss.zzz}] %{function} %{message}"`

Though by the way, I think that's something that should rather be specified in contribution guidelines to maintain specific and consistent 'format' for logging messages.

src/providers/twitch/PubSubClient.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include "util/DebugCount.hpp"
+#include "util/Helpers.hpp"
+#include "util/RapidjsonHelpers.hpp"
+
+#include <exception>
+#include <thread>
+
+namespace chatterino {
+
+static const char *PING_PAYLOAD = "{\"type\":\"PING\"}";
```
zneix, line 18: This could be made a string literal.

src/providers/twitch/PubSubClient.hpp
```diff
@@ -0,0 +1,74 @@
...
+class PubSubClient : public std::enable_shared_from_this<PubSubClient>
+{
+public:
+    struct UnlistenPrefixResponse {
+        std::vector<QString> topics;
+        QString nonce;
+    };
+
+    // The max amount of topics we may listen to with a single connection
+    static constexpr std::vector<QString>::size_type listensPerConnection = 50;
```
zneix, line 34: This static could/should be full caps

pajlada (approved):

pajlada:
image

PR #3648, opened 2022-04-03
Use login name for NotificationController lookup

### LINKED ISSUE #3646: Live status Helix call now requires displayname capitalization
### Checklist

### Describe your issue

Due to what I can only assume as an internal Twitch change, our `live notifications` no longer work if the account you want notifications for falls into both of the following:
1. Your `live notifications -> selected channels` does not use correct display-name capitalization `e.g. xQcOW`
2. You do not have a tab open with the channel in question.

Due to the fact that when a tab is open, we send `user_id` based helix calls, when a channel is open the issue in question does not arise and the `live notification` is sent correctly.

### Screenshots

Example 1: All 3 channels `bobross`, `lol_nemesis` & `pajlada` are using incorrect display-name capitalization
After the display-name capitalization is fixed, the live status call works and the notifs appear as intended.
c2_3646_1

Example 2: Channel `BobRoss` has correct capitalization, while `lol_nemesis` & `pajlada` do not, but channel `pajlada` has a tab open.
c2_3646_2

### OS and Chatterino Version

Chatterino 2.3.5-beta1 28fa6b8 | windows 10

iProdigy:
Pull request checklist:

# Description
Fixes #3646
* Add `userLogin`, `gameName` to `HelixStream`
* Store `userLogin` instead of `userName` from streams response
* Lookup lowercase `name` in `liveStreams` set

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Bugfix: Fixed rendering of moderator announcements. (#3639)
+- Bugfix: Fixed display name dependence of live status calls. (#3646)
 
 ## 2.3.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Bugfix: Fixed display name dependence of live status calls. (#3646)
+- Bugfix: Fixed live notifications for usernames containing uppercase characters. (#3646)
 
 ## 2.3.5
 
--- a/src/controllers/notifications/NotificationController.cpp
+++ b/src/controllers/notifications/NotificationController.cpp
@@ -178,12 +178,12 @@ void NotificationController::fetchFakeChannels()
                 std::unordered_set<QString> liveStreams;
                 for (const auto &stream : streams)
                 {
-                    liveStreams.insert(stream.userName);
+                    liveStreams.insert(stream.userLogin);
                 }
 
                 for (const auto &name : batch)
                 {
-                    auto it = liveStreams.find(name);
+                    auto it = liveStreams.find(name.toLower());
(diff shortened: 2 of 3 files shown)


iProdigy:
Aside: [this logic](https://github.com/Chatterino/chatterino2/blob/96e2d1f97174751e0f59f2c3baeb2a6697bcc196/src/providers/twitch/TwitchChannel.cpp#L742-L769) can be simplified now due to the `gameName` field, but someone else can do this refactor

Mm2PL:
I'd say that a bugfix changelog entry is good for this. Also looks like this is my fault, added in https://github.com/Chatterino/chatterino2/pull/3442

Felanbird:
> I'd say that a bugfix changelog entry is good for this. Also looks like this is my fault, added in #3442

I took the assumption of `Work towards` meaning that another PR would be required to fix the underlying issue - and if necessary that changelog could include `(#thisPR, #thatPR)` 

but as I test this PR I can see the calls are already fixed 🙂 
image

iProdigy:
Thanks for the confirmation felan, was planning to review the logic in `NotificationController::checkStream` if this alone didn't fix it

have added a changelog entry now

Felanbird (approved):
Issue is resolved - live notifications follow the specified case-sensitivity, but use intended display-name casing when the tab is open.

image

image

pajlada (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed display name dependence of live status calls. (#3646)
```
pajlada, line 5:
```suggestion
- Bugfix: Fixed live notifications for usernames containing uppercase characters. (#3646)
```

PR #3658, opened 2022-04-07
add "show live tabs only" option

KogasaPls:
Pull request checklist:

# Description
This PR adds a "show live tabs only" option that hides tabs which are not marked as live. By default, it is disabled and has no hotkey bound to it. While active, it locks the tab layout, since it's not clear how a change in layout should translate when offline tabs are unhidden. When tabs are marked as live/not live, they will automatically be inserted/hidden appropriately. 

A video demonstration of the core functionality can be found [here](https://streamable.com/8v900l).

I'd appreciate your feedback. Thank you.

### DIFF
--- a/src/controllers/hotkeys/ActionNames.hpp
+++ b/src/controllers/hotkeys/ActionNames.hpp
@@ -206,7 +206,7 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
           }},
          {"setShowLiveOnly",
           ActionDefinition{
-              "Show live tabs only",
+              "Only show live tabs",
               "[on, off, or toggle. default: toggle]",
               0,
               1,
--- a/src/controllers/hotkeys/ActionNames.hpp
+++ b/src/controllers/hotkeys/ActionNames.hpp
@@ -203,6 +203,13 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
               "[on, off, or toggle. default: toggle]",
               0,
               1,
+          }},
+         {"setShowLiveOnly",
+          ActionDefinition{
+              "Show live tabs only",
+              "[on, off, or toggle. default: toggle]",
+              0,
+              1,
           }}}},
 };
 
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -34,13 +34,29 @@ Notebook::Notebook(QWidget *parent)
 
     this->addButton_->setHidden(true);
 
+    // create actions at once, so we can disable one checkbox with the other
+    this->showLiveOnlyAction_ = new QAction("Show live tabs only", this);
     this->lockNotebookLayoutAction_ = new QAction("Lock Tab Layout", this);
 
-    // Load lock notebook layout state from settings
+    // Load checkbox states from settings
     this->setLockNotebookLayout(getSettings()->lockNotebookLayout.getValue());
+    this->setShowLiveOnly(getSettings()->showLiveOnly.getValue());
 
+    this->showLiveOnlyAction_->setCheckable(true);
+    this->showLiveOnlyAction_->setChecked(this->showLiveOnly_);
+    QObject::connect(this->showLiveOnlyAction_, &QAction::triggered,
+                     [this](bool value) {
+                         this->setShowLiveOnly(value);
+                         this->lockNotebookLayoutAction_->setEnabled(!value);
+                         this->lockNotebookLayoutAction_->setChecked(
+                             value || this->lockNotebookLayout_);
+                     });
+
+    //check and disable if we're only showing live tabs
     this->lockNotebookLayoutAction_->setCheckable(true);
-    this->lockNotebookLayoutAction_->setChecked(this->lockNotebookLayout_);
+    this->lockNotebookLayoutAction_->setEnabled(!this->showLiveOnly_);
+    this->lockNotebookLayoutAction_->setChecked(this->lockNotebookLayout_ ||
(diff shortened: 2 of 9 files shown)


src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -202,6 +202,13 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
               "[on, off, or toggle. default: toggle]",
               0,
               1,
+          }},
+         {"setShowLiveOnly",
+          ActionDefinition{
+              "Show live tabs only",
```
jammehcow, line 208: Wording was a bit backwards
```suggestion
              "Only show live tabs",
```

src/widgets/Notebook.cpp
```diff
@@ -390,6 +404,21 @@ void Notebook::setShowTabs(bool value)
     }
 }
 
+bool Notebook::getShowLiveOnly() const
+{
+    return this->showLiveOnly_;
+}
+void Notebook::setShowLiveOnly(bool value)
```
jammehcow, line 411: nitpick: formatting
```suggestion
}

void Notebook::setShowLiveOnly(bool value)
```

src/widgets/Notebook.cpp
```diff
@@ -390,6 +404,21 @@ void Notebook::setShowTabs(bool value)
...
+{
+    this->showLiveOnly_ = value;
+    this->showLiveOnlyAction_->setChecked(value);
+    //check and disable the "lock tabs" button
+    this->lockNotebookLayoutAction_->setEnabled(!value);
```
jammehcow, line 416: nitpick: formatting
```suggestion
    this->showLiveOnlyAction_->setChecked(value);

    // check and disable the "lock tabs" button
    this->lockNotebookLayoutAction_->setEnabled(!value);
```

src/widgets/Notebook.hpp
```diff
@@ -80,6 +83,10 @@ class Notebook : public BaseWidget
         QWidget *selectedWidget{};
     };
 
+    const QList<Item> allItems()
+    {
+        return allItems_;
+    }
     const QList<Item> items()
```
jammehcow, line 90: nitpick: formatting
```suggestion
    }

    const QList<Item> items()
```

src/widgets/Window.cpp
```diff
@@ -660,6 +660,53 @@ void Window::addShortcuts()
...
+                 getSettings()->showLiveOnly.setValue(true);
+                 this->notebook_->setShowLiveOnly(true);
+             }
+             else if (mode == 2)
+             {
+                 auto value = this->notebook_->getShowLiveOnly();
+                 getSettings()->showLiveOnly.setValue(!value);
+                 this->notebook_->setShowLiveOnly(!value);
+             }
+             return "";
```
jammehcow, line 708: You've essentially duplicated your if/else statements here. I'd recommend that you move all of your `mode` conditional logic into the first lot of statements and remove the `mode` variable altogether.

jammehcow (changes requested):
Other than the changes I've suggested as part of the review, I've tested and found two bugs:

1. Toggling this feature and opening a new tab causes an empty spit to appear in the first position (on top of settings/accounts buttons + the first tab):

https://user-images.githubusercontent.com/4962764/162606396-07d2b6e2-16ef-4edb-aaaf-bdb908b835c9.mp4

2. Ctrl + Tab allows you to navigate all tabs even when they're hidden, but the quick switcher (Ctrl + K) does not switch to the hidden channels.

https://user-images.githubusercontent.com/4962764/162606401-087b0040-8e37-4e3f-b316-fd37360a9412.mp4

Nice work so far but has a few kinks that need ironing out.

pajlada:
Super interesting feature! Would love to get this merged in. Haven't had a time to review the actual code yet but thank you for working on it!

src/widgets/Window.cpp
KogasaPls: Thanks for the review! I simplified this bit in 51ac47b.

KogasaPls:
I've addressed both bugs in this last set of changes.

CTRL+(SHIFT)+Tab will now cycle through only the _displayed_ tabs, and the channel switcher will now correctly focus offline tabs. Would it be better for the channel switcher filter out offline tabs instead?

KogasaPls:
I've noticed that I can't change the channel of a split while the setting is enabled. I'll look into fixing that.

zneix:
I don't like the idea this "feature" brings, wouldn't this clash with "toggle visibility of tabs"? shouldn't rather already existing tab visibility 0(hidden)/1(visible) be changed to something like an enum? e.g. 0/1/2 - hidden, all visible, only live visible, etc... if someone decides to add more things and maintain better future-compatibility.

jammehcow:
Agree with @zneix on this one, combining the two similar features into a single setting would make this a whole lot nicer to work with in the future - should we need to add something else around tab visibility.

maxcamorra:
Will I stay in the chat of the offline streamer as well or will it automatically leave these and only stay inside the live streamer?

leon-richardt:
> Will I stay in the chat of the offline streamer as well or will it automatically leave these and only stay inside the live streamer?

Toggling this setting yields a strictly visual change. You will stay connected to offline chat rooms in the background, only their tabs will not be shown in the notebook.

pajlada (changes requested):
Small questions/tweaks, then I'll do a re-review

pajlada:
Closing in favour of https://github.com/Chatterino/chatterino2/pull/4358

PR #3662, opened 2022-04-08
refactor(TwitchChannel): remove getGameById call

iProdigy:
Pull request checklist:

- [ ] `CHANGELOG.md` was updated, if applicable - let me know if this warrants an entry

# Description

Removes unnecessary helix call (`getGameById`) now that `HelixStream` contains `gameName` (follow-up from https://github.com/Chatterino/chatterino2/pull/3648#issuecomment-1086924644)

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -735,39 +735,8 @@ void TwitchChannel::parseLiveStatus(bool live, const HelixStream &stream)
     {
         auto status = this->streamStatus_.access();
         status->viewerCount = stream.viewerCount;
-        if (status->gameId != stream.gameId)
-        {
-            status->gameId = stream.gameId;
-
-            if (!stream.gameId.isEmpty())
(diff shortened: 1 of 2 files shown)


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -735,39 +735,8 @@ void TwitchChannel::parseLiveStatus(bool live, const HelixStream &stream)
...
         auto status = this->streamStatus_.access();
         status->viewerCount = stream.viewerCount;
-        if (status->gameId != stream.gameId)
-        {
-            status->gameId = stream.gameId;
-
-            if (!stream.gameId.isEmpty())
-            {
-                // Resolve game ID to game name
-                getHelix()->getGameById(
```
iProdigy, line 745: I believe this is the only use of `getGameById` - should the method be removed from the Helix file? (it simply just calls `fetchGames` anyways)

pajlada (approved):

PR #3683, opened 2022-04-17
Add chatter count for each category in viewer list

badoge:
Pull request checklist:

# Description
Adds the number of chatters to each category label in the viewer list.
image

### DIFF
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -1009,8 +1009,7 @@ void Split::showViewerList()
 
                 auto label = formatListItemText(
                     QString("%1 (%2)")
-                        .arg(labels.at(i))
-                        .arg(localizeNumbers(currentCategory.size())));
+                        .arg(labels.at(i), localizeNumbers(currentCategory.size())))
                 label->setForeground(this->theme->accent);
                 chattersList->addItem(label);
                 foreach (const QJsonValue &v, currentCategory)
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -998,18 +998,6 @@ void Split::showViewerList()
                     .arg(localizeNumbers(obj.value("chatter_count").toInt())));
 
             loadingLabel->hide();
-            QList<QListWidgetItem *> labelList;
-            for (int i = 0; i < labels.size(); i++)
-            {
-                auto label = formatListItemText(
-                    QString("%1 (%2)")
-                        .arg(labels.at(i))
-                        .arg(localizeNumbers(chattersObj.value(jsonLabels.at(i))
-                                                 .toArray()
-                                                 .size())));
-                label->setForeground(this->theme->accent);
-                labelList.append(label);
-            }
             for (int i = 0; i < jsonLabels.size(); i++)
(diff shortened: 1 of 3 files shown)


Mm2PL:
Even more settings. Why is it detrimental to have this always on?

kornes:
good UX is based on showing user only data which he needs, we already show total number of chatter in window tittle. From my pov adding extra visual noise to UI which serves little purpose for 99% of users just doesn't fit new default behavior. 

Adding more settings is another thing which noone likes, that's why i suggested to add it in local scope of viewers list only, where it would be found only when someone checked context menu. This would be middle ground with no negative effects for both sides and allow more customization which is always nice.

LosFarmosCTL:
How is this behavior being on by default affecting anyone negatively? I fail to see how this is 'UI clutter', showing the number of entries in a list is a very standard behavior of any UI and its not like the viewer list is some dashboard with 100 graphs and control panels, I highly doubt 3 chars more at the end of a list header will annoy anyone, while providing value for people who need it.

Hiding this behind some context menu would just result in 90% of users not even knowing it existed.

kornes:
to rephrase my point: if some data, no matter of its size in UI, gonna be useful for very small portion of users if any (and that's my main assumption here) i just wouldn't show it by default.

Did any of you guys ever found yourself asking "how many vips or mods are watching?" i never did. But maybe its just me-thing and i already spammed here too much, so lets others talk :D

treuks:
Look, I'm just going to interject in real quick, this feature is great, I really like it.

There is no need to make it a setting, there is no need to be able to toggle this, THIS DOES NOT REMOVE ANYTHING, IT ONLY ADDS SOMETHING. It's neither a side-grade nor a down-grade.

It's not extra information you don't need, it fits in perfectly, like it has always been there. I have genuinely wondered for how much people are in the viewer list right now, and in what groups they are precisely.

If Chatterino was a minimalistic application, we would not have all the features we have right now. Don't come in here with this minimalistic approach.

Please, leave it as it is, in my opinion, there are no visual changes needed.

Felanbird (approved):
good addition 👍

Mm2PL (approved):

src/widgets/splits/Split.cpp
```diff
@@ -1014,7 +1007,12 @@ void Split::showViewerList()
                 if (currentCategory.empty())
                     continue;
 
-                chattersList->addItem(labelList.at(i));
+                auto label = formatListItemText(
+                    QString("%1 (%2)")
+                        .arg(labels.at(i))
+                        .arg(localizeNumbers(currentCategory.size())));
```
zneix, line 1013: If there are multiple substitutions of the same type, [`multi-arg overload`](https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-arg.md) should be used instead of making unnecessary temporary heap allocations with multiple `.arg(QString string)` calls:
```suggestion
                        .arg(labels.at(i), localizeNumbers(currentCategory.size())))
```

zneix (changes requested):

pajlada (approved):

zneix (approved):

Issue #2544, opened 2021-03-20
Chat title not DPI adjusting on macOS


jackitsdotcom:
**What should be added?**
The title of the chat window is much smaller than the text within, and cannot find any setting to adjust this.
<img width="638" alt="Screen Shot 2021-03-20 at 12 21 54 pm" src="https://user-images.githubusercontent.com/7483527/111856500-e8370c00-8976-11eb-8b1d-5d06933b9024.png">

It would be great if this either better matched the OS native DPI settings, or even better, was adjustable in the Settings.

**Why should it be added?**
Would allow title to be seen in the same way as chat, including viewer count, title, game etc. ]

jammehcow:
Can reproduce on macOS Catalina 10.15.5 with chatterino `2.2.3-beta2 (commit 6499c7c12)`.

zneix:
Related: #2412

Issue #1076, opened 2019-05-18
Font size in tabs on OSX is inconsistent in relation to other font sizes on OSX


coral:
![Screenshot 2019-05-18 at 15 40 27](https://user-images.githubusercontent.com/2163363/57975706-4eba0000-7983-11e9-8569-c2d60c62b1e9.png)

jammehcow:
Ties in with #1075? Still present using Catalina and most recent chatterino version. I suspect DPI scaling isn't working as expected for a bunch of elements on macOS.

Issue #1075, opened 2019-05-18
Font size in account settings on OSX is inconsistent


coral:
![Screenshot 2019-05-18 at 15 38 52](https://user-images.githubusercontent.com/2163363/57975695-11557280-7983-11e9-9d62-5bcb27edc10e.png)

jammehcow:
Needs label 'high dpi'

PR #3687, opened 2022-04-19
Fix tooltips ignoring zoom/font changes

kornes:
Pull request checklist:

# Description

Currently due to a bug, tooltips (emotes, badges, link preview, split header) are completely ignoring changes of zoom level or font size/style from settings. This PR fixes that.

Example with 1.3x zoom. Left side shows fixed tooltip, with zoom matching rest of the app. Right side is current behavior.
screen

tested on windows only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Minor: Add information about the user's operating system in the About page. (#3663)
 - Bugfix: Fixed live notifications for usernames containing uppercase characters. (#3646)
 - Bugfix: Fixed certain settings dialogs appearing behind the main window, when `Always on top` was used. (#3679)
+- Bugfix: Fixed tooltips not reacting to zoom and font changes. (#3687)
 - Dev: Use Game Name returned by Get Streams instead of querying it from the Get Games API. (#3662)
 
 ## 2.3.5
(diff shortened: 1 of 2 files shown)


jammehcow (approved):
Looks good to me, thanks for attacking some of these niggly little issues in Chatterino!

Felanbird (approved):
Works as intended 👍

LosFarmosCTL (approved):
Changes work as intended, I wanted to at least mention though that this has the unfortunate side effect of making all tooltips horrible on macOS, due to the existing scaling issues. (#1076 #1075 #2544) Really annoying, but not an issue with this PR, just makes fixing that general problem a bit more pressing.

chatterino 2022-04-20 at 15 59 53chatterino 2022-04-20 at 16 00 04

imagechatterino 2022-04-20 at 16 04 08

kornes:
so if i understand correctly, with this change mac tooltips are getting extra tiny (right screen) on default zoom?
might look into it

LosFarmosCTL:
Nah, the entire macOS UI is already extra tiny, the tooltips just weren’t affected previously because they were missing the scaling.

jammehcow:
I wouldn't be comfortable merging this until we fix the DPI scaling or have some decent workaround for it. As great as this feature is, it will negatively impact macOS users + high DPI resolutions.

kornes:
agree, made some progress on this, will PR tomorrow for tests

kornes:
closing it, moved to #3690

Issue #2412, opened 2021-01-28
Inconsistent/Too Small Interface Size in Ubuntu 20.04


scchow:
**Describe the bug**
1. The font size of the interface is very small. This includes the tab bar, settings/user accounts/update icons, and the streamer name.
2. The font size of the settings menu is also very small.
3. The font size of the send message box seems overly large.
4. See attached images for details.

**To reproduce**
1. Cloned commit [055db0c](https://github.com/Chatterino/chatterino2/commit/055db0ccfb52691991ff2a6190c3cd37c6a2df59)
2. Ran the following commands to build
    1. `sudo apt install qtcreator qtmultimedia5-dev libqt5svg5-dev libboost-dev libssl-dev libboost-system-dev libboost-filesystem-dev`
    2. `mkdir build`
    3. `cd build`
    4. `qmake ..`
    5. `make`
    6. `cd bin`
    7. `./chatterino`
 3. After starting up chatterino, the font size is inconsistent and illegible.

**Screenshots**

![image](https://user-images.githubusercontent.com/11098209/106087655-4d8a2180-60d9-11eb-9713-76048da5dd2c.png)

The font in the top bar matches the system font size and is what I would like the standard text size to be. Notice that the streamer name and settings icons are small while the send message box is large.

The font size of the chat was adjusted using Zoom and text size options in the settings, but they have no effect on the interface size.

![image](https://user-images.githubusercontent.com/11098209/106088183-50d1dd00-60da-11eb-953e-4e16c2eb0bd6.png)

The font size of the settings menu is also small when compared to the font size of the top bar.

**Chatterino version**
Chatterino 2.2.3-beta2 (commit 055db0ccf)

 
**Operating system**
Ubuntu 20.04 (5.4.0-52-generic)

**Additional information**
- Monitors:
  - Monitor 1: 1920x1080: 27"
  - Monitor 2: 2560x1440: 27"
  - Monitor 3:  1920x1080: 24"
- Potentially Related Issues
  - #2351 : Similar problem in MacOS, but their preferences menu is not affected/has the right size font. 
  - #2348 - From my tests, it does not seem like it matters which monitor the terminal window in which `./chatterino` is run makes a difference.

pajlada:
What are your DPI settings for your system / your screens?

scchow:
`runnding xdpyinfo` reports my monitors are running at 96x96 dpi.

[...]

Issue #2351, opened 2021-01-09
Interface font size is very small


revolter:
**Describe the bug**

The font size of the interface got very small.

**To reproduce**

It looked (but I'm not sure) that every time I used the search window, the font got smaller and smaller, and it's the same even when restarting the app.

**Screenshots**

<img width="481" alt="image" src="https://user-images.githubusercontent.com/5748627/104106165-56e35380-52bc-11eb-9d51-abf7c7fd9dd7.png">

<img width="402" alt="image" src="https://user-images.githubusercontent.com/5748627/104106194-ae81bf00-52bc-11eb-9759-30d74231efbb.png">

The preferences window is ok:

<img width="915" alt="Screenshot 2021-01-09 at 20 54 39" src="https://user-images.githubusercontent.com/5748627/104106221-e557d500-52bc-11eb-8ef4-02ce92fca607.png">

**Chatterino version**

Chatterino 2.2.3-beta2 (commit 6499b7c12)

**Operating system**

macOS 11.1 (20C69)

**Additional information**

scchow:
I am also running into the same issue. 

# Problems

1. The font size of the interface is very small. This includes the tab bar, settings/user accounts/update icons, and the streamer name.
2. The font size of the settings menu is also very small.
3. The font size of the send message box seems overly large.
4. See attached images for details.

# Things I've Tried

1. Modifying the zoom level and text size only changes the size of the text in the chat.
2. Moving the Chatterino window to different monitors with different resolutions to not fix the problem.


# System Information
- Chatterino version: 2.2.3-beta2 (commit [055db0c](https://github.com/Chatterino/chatterino2/commit/055db0ccfb52691991ff2a6190c3cd37c6a2df59))
  - Built and run with the following commands after cloning the repository:
    1. `sudo apt install qtcreator qtmultimedia5-dev libqt5svg5-dev libboost-dev libssl-dev libboost-system-dev libboost-filesystem-dev`
    2. `mkdir build`
    3. `cd build`
    4. `qmake ..`
    5. `make`
    6. `cd bin`
    7. `./chatterino`
- Operating System: Ubuntu 20.04 (5.4.0-52-generic)
- Monitors:
  - Monitor 1: 1920x1080: 27"
  - Monitor 2: 2560x1440: 27"
  - Monitor 3:  1920x1080: 24"


# Images

[...]

Issue #2348, opened 2021-01-09
[2.2.3-beta2] - Startup size scaling issue


Swiftpaws:
**Describe the bug**
Chatterino seems to have some problem with the startup size (scaling issue)

My Setup:
I am using three monitors: 1080p (100% scaling) - 1440p (main) (125% scaling) - 2160p (175% scaling)

Initially:
Chatterino seems to have no issues with calculating the startup size on my main monitor, but on the other two it cannot seem to get it quite right:

1. Main monitor 1440p at 125% scaling
	-  Ok - Chatterino always starts with the same dimensions

2. Second monitor 2160p at 175% scaling
	- Not Ok -> Startup size is different after locking windows and signing in

![image](https://user-images.githubusercontent.com/18440081/104091589-9ee38580-527e-11eb-9752-53348e43122b.png)

3. Third monitor 1080p
	- Not Ok -> Startup size is different every time ??? window gets smaller

![image](https://user-images.githubusercontent.com/18440081/104091636-deaa6d00-527e-11eb-88ae-050e26c2cfd6.png)

**Probable cause**
During my testing I could not figure out why there was no clear pattern to the problem. Locking Windows changes what exactly?

But then it occurred to me -> *Chatterino seems to factor in the scaling of the monitor where the mouse cursor is on, not the one where the application will actually open*

So by moving the mouse cursor to another monitor with different scaling the startup size will not be equal to the last used size.

**Chatterino version**
Chatterino 2.2.3-beta2 (commit 6499b7c12)

**Operating system**
Windows 10 - 2004 - 21286

SLCH:
Could be related with #2322

Swiftpaws:
Hmm It could be related but aside from the wrong calculation of startup dimensions I have no other issues - scaling of text and other ui elements seems to work fine.

After some more testing I am pretty confident in my findings under "probable cause". Depending on how this use case is implemented here it might very well be a QT related bug though.

Issue #2322, opened 2020-12-30
Not able to interact with the majority of the UI


erbrinks:
**Describe the bug**
Not able to click on anything besides the tabs to select the channel and delete that channel. (Not able to close the application, click settings, login, minimize, or click to maximize).

Able to add splits but I cannot interact with the UI that pops up to specify the split. However, I can remove splits and interact with the prompt that pops up asking for confirmation.

As far as I know, I am able to type in the chat if I could login due to the fact that I can type in the text box.

In terms of looks, everything is zoomed in and the text is blurry. Saw something similar to that visually when I was looking for solutions but did not see a fix for it. Emotes are working fine as far as I can tell.

**To reproduce**
Launching the application (administrator mode or not).

**Screenshots**
![image](https://user-images.githubusercontent.com/56411406/103369802-118f7c80-4a99-11eb-9633-ceb1add0b9d0.png)

**Chatterino version**
Tried version 2.2.2-fix and 2.2.4-beta2. The application worked awhile ago and I was using it all the time but I am not sure of the version that I was using. It broke when I was using that version so I am not sure it would make a difference anyway.

**Operating system**
Windows 10 (Latest updates)

**Additional information**
Noticed the issue after I ran the application on my vertical monitor awhile ago. Not sure if this is related in any way. Does not work on either monitor now.

Main monitor: 3840 x 2160
Vertical monitor: 1080 x 1920

alazymeme:
Sounds like a DPI issue. What DPI are both monitors set to? Chatterino works just fine vertically for me:

![image](https://user-images.githubusercontent.com/12804673/103397381-dfa71480-4b83-11eb-8f84-1ef327b49c9c.png)

erbrinks:
I went in and changed the scaling from 150% on my main monitor to 100% and it worked on both monitors. I changed it back without closing the program and it kept working. However, if I keep it at 150% and close the app then re-open it, the same issue arises. 

It's not a really big deal, but it would be annoying to have to change the scaling from 150% to 100% and back again when I want to use the app. 

If this is something that is not on your end and cannot be patched I totally understand. Thanks for the input and quick response though!

PR #3690, opened 2022-04-22
Font/UI scaling improvements

kornes:
Pull request checklist:

# Description

This PR aims to improve scaling of fonts and UI in high DPI.
MacOS will be mostly affected, but Windows also had few small fixes (tested with 175% windows scaling thing which is not real DPI but still exposed problems). Linux should be not affected, as its excluded from DPI modifier (for now at least).

All DPI calculations were moved to single place in [Fonts::getOrCreateFontData()](https://github.com/Chatterino/chatterino2/pull/3690/files#diff-e32d8ff0be4a2aa86c6157281840cc5faacc08d30d57a7e8e0e7906179df621bR112) which now takes widget pointer for that purpose (as each child window can end up on different screen using different DPI). Windows and MacOS are set to use same scaling for now.

This PR also includes tooltip scaling fix #3687

To test on all platforms:
- if linux is really not affected at all
- if scaling on Retina/high DPI displays looks OK or initial font size is still too small and has to be bumped
- if zoom between 0.5x - 4.0x doesn't break any UI elements right away
- if scaling is uniform across all widgets and texts
- is every text readable and fully visible (main window, viewers list, emote picker, user card) and doesn't go out of bounds
- if multiple monitors at different DPI are available, move chatterino windows between them while changing zoom level

there will be more things to fix for sure, but heres start


jammehcow:
@LosFarmosCTL are you able to do some macOS testing for this one?

LosFarmosCTL:
> @LosFarmosCTL are you able to do some macOS testing for this one?

Sure, I'll do some testing sometime today.

LosFarmosCTL (changes requested):
Changes look overall very good to me on macOS, huge improvement across basically the entire application compared to before.

chatterino 2022-04-23 at 17 10 15

chatterino 2022-04-23 at 17 12 04 chatterino 2022-04-23 at 17 12 09

chatterino 2022-04-23 at 17 12 58 chatterino 2022-04-23 at 17 13 05

chatterino 2022-04-23 at 17 13 25 chatterino 2022-04-23 at 17 13 29

chatterino 2022-04-23 at 17 51 16 chatterino 2022-04-23 at 17 51 10

I did find a few cases where the scaling is unchanged though, and still too small compared to testing on a windows machine though:

### 1. The chatter list dialog:

chatterino 2022-04-23 at 17 14 41

### 2. The hotkey table in settings:

chatterino 2022-04-23 at 17 47 32

### 3. The account list in settings:

chatterino 2022-04-23 at 17 48 07

All other settings pages and tables (i.e. for filters, etc.) have always been fine though even before this PR, it's only those 2 that are off.

kornes:
Thanks for testing!
can you post screen of how Filters tab looks for reference? on windows Accounts and Filters tab appears to use same font size

LosFarmosCTL:
### Filters:

image

### Hotkeys:

image

kornes:
both mentioned tabs were the only one using custom fonts, now removed so all settings tabs should look the same.

Now, categories need some clear visibility so i used background coloring instead (qt's dark gray):
img

img2

if people wont like this solution, could revert it and make bigger fonts work, but i think its not bad :D

jammehcow:
Probably fixes all of the following issues under the `High DPI` label:
 - #2544
 - #2412
 - #2351
 - #2322
 - #2348
 - #1076
 - #1075
 - #755
 - #754
 - #752
 - #750
 - #749

@LosFarmosCTL sorry for ping again, would you be able to verify the macOS ones (and any others you might be able to reproduce)? 

Tested #1092 and sadly this doesn't affect it, so not a DPI issue.

LosFarmosCTL:
> @LosFarmosCTL sorry for ping again, would you be able to verify the macOS ones (and any others you might be able to reproduce)?

No problem, the macOS specific ones are all fixed by this yeah (#2544 #2351 #1076 #1075). Not entirely sure if this addresses all high DPI ones though, i.e. #750 is should definitely be unaffected right? For the rest its probably a good idea if someone with a high DPI display takes another look at them.

Felanbird:
@RAnders00 Do you mind testing this PR - as you are the issue holder of 5/12 of these issues 😁

jammehcow:
Did some testing, appears that messages + the container are unaffected. I'd imagine that we'd want to pass DPI scaling through right down to the message element from their container (which would end up fixing #750 and #1092 in the process). 

Also, live changes to the DPI cause the split's input to mangle its size on Linux (and I assume this will come up when moving between different DPI monitors):

160 DPI:
image

96 DPI (changed with app running): 
image

kornes:
> Did some testing, appears that messages + the container are unaffected. I'd imagine that we'd want to pass DPI scaling through right down to the message element from their container

Decided to skip Message*.cpp files, because chat scaling seems to be working fine and even your 2 screens from different dpi looks fine in this aspect. Will open another PR for #1092, about #750 not sure if its still valid on qt 5.15 but will test it this week with other linked issues.

Did your linux tests worked well outside of runtime dpi changes? or fonts on screen appeared wonky even at normal start.
Runtime DPI change is still poorly handled (didn't touch that in this pr) and will need lot more research and testing, but impact is small as its mostly affecting users who dock their laptops and/or swap to external monitors with app running. 
Moving windows between monitors worked fine from my testing so far and rescale between DPIs with no issues.

Xyncgas:
image

Xyncgas:
I recommand in the future abandon absolute sizing (px) but use a size pallet that's built with the window's dimension / the system's dimension, so if you set the font size to MySizePallet.Get(15) (15th element) it will automatically adaptive to the dimension of the enviornment

Brog33:
I'd like to do some testing on Ubuntu.

Caf:
This is still not merged in the main branch right? I just have to wait?   :p

Brog33:
> This is still not merged in the main branch right? I just have to wait? :p

yeah, if this works as described, it will be a big deal for linux and macos. hope a merge will happen soon.

dead-claudia:
What's the status of this PR?

pajlada:
Would love to get this tested by folks with different setups

On my X11/Arch Linux btw system, there's no difference at all.
On my remote desktop mac running macOS sonoma with no scaling, everything looks great. The "Change channel" popup is now readable, tabs are now readable, split headers are now readable

LosFarmosCTL:
I recently bought a high-DPI MacBook, so I could finally test this on macOS for both low- and high-DPI displays.

On my old Hackintosh, with a standard 1080p monitor, the changes look great, everything is at a reasonable scale and legible.

For high-DPI, there still appear to be some issues though:

image

The tab titles and the change channel dialog are still way too small, while they appear fine on low-DPI screens.

Some stuff works completely fine on high-DPI as well though, such as the table cell content on settings pages:

image

Overall the only things I could find that don't work are the tab titles, message input box and the change channel dialog.

Nerixyz:
The context menus and labels are too big on Windows:

chatterino_2024-03-23_13-19-25

LosFarmosCTL:
Tooltips appear fine on high-dpi macOS:

image

Labels have the same issues as tab titles, etc.

SCR-20240323-maou

Both are fine on a low-dpi screen.

Nerixyz:
> Labels have the same issues as tab titles, etc.

Can you try and comment out https://github.com/kornes/chatterino2/blob/f2d5f60a74e58b3c16acf1d59f5c03c4b97d345e/src/singletons/Fonts.cpp#L97-L105, like I did in https://github.com/Chatterino/chatterino2/pull/3690#issuecomment-2016477569?

LosFarmosCTL:
> Can you try and comment out https://github.com/kornes/chatterino2/blob/f2d5f60a74e58b3c16acf1d59f5c03c4b97d345e/src/singletons/Fonts.cpp#L97-L105, like I did in [#3690 (comment)](https://github.com/Chatterino/chatterino2/pull/3690#issuecomment-2016477569)?

image  


That's definitely a big improvement, but there is still something different going on compared to other areas of the UI.

Best example imo, the button scaling (and text in general) in the change channel dialog, compared to the settings:

image

image


That change is enough to make everything legible though.

Nerixyz:
Hmm... I also have the bug that stuff in the select channel popup isn't sized correctly. Except for me, it's slightly too big. Changing https://github.com/kornes/chatterino2/blob/f2d5f60a74e58b3c16acf1d59f5c03c4b97d345e/src/widgets/BaseWindow.cpp#L779 to use `this->qtFontScale()` instead of `this->scale()` fixes this, but the account-switch button is too small (probably fixed by https://github.com/Chatterino/chatterino2/pull/4868).

kornes:
stalled and most issues were already addressed

Issue #1738, opened 2020-06-14
Search in all tab


Leeqo:
**What should be added?**
 option to ctrl+f for all channels open 

**Why should it be added?**
i want to be able to search either a user or a phrase in all my tabs open
my amazing design https://i.imgur.com/8a7ysPI.jpg :)

apa420:
What about doing ctrl+shift+f for global search instead.

jammehcow:
A default hotkey of `Ctrl + Shift + F` sounds good to me - the user can change hotkeys as of 2.3.5.

As for implementing this feature, I think passing a list of channels/`ChannelViews` to the search popup is the easiest solution. If a user has multiple splits of the same channel, only one should be snapshotted but filters from all splits for that channel should be applied.

Pretty keen to implement this, sounds useful.

PR #3694, opened 2022-04-25
feat: add global channel search support

jammehcow:
Pull request checklist:

# Description

I will write a description with my changes tomorrow, I am sleeping.

TODO:
 - [x] Apply filters from each channel
 - [x] Add channel name elements before messages (same as `/mentions` tab)
 - [x] Clean up the multi-channel snapshot logic
 - [x] User info popup support with correct channel as context
 - [ ] ~Mod mode support~ not supporting as mod mode is a channel flag, not a message flag

Implements #1738


src/widgets/helper/SearchPopup.cpp
```diff
@@ -88,17 +87,22 @@ void SearchPopup::addShortcuts()
...
+        this->channelView_->setSourceChannel(channel.channel());
+        this->channelName_ = channel.channel()->getName();
+    }
+    else if (this->searchChannels_.size() == 1)
+    {
+        this->channelView_->setSourceChannel(
+            std::make_shared<Channel>("multichannel", Channel::Type::None));
+    }
+
+    this->searchChannels_.append(std::ref(channel));
```
kornes, line 105: checking for duplicate channels here will avoid combining snapshots and iterating messages later on

jammehcow: That would be nice, but the reason I'm not checking is you can have multiple splits of the same channel with different filters. I want to respect all filters when searching - not just the first split's.

jammehcow: Also we'll still need to build this monolithic snapshot no matter what as each channel needs to be snapshotted in order to search it.

kornes: didn't consider filters so make bigger sense now, waiting to test how it works out 👍

jammehcow:
Issue: using the wrong time when sorting. Needs to pull the TMI time rather than the message's `TimeStampElement` as it's not always granular.

jammehcow:
@pajlada any concerns about me making `MessageView#getFlags()` public? Makes this feature a lot easier as it needs to add & remove flags.

src/messages/Message.hpp
```diff
@@ -63,6 +63,7 @@ struct Message : boost::noncopyable {
     QString timeoutUser;
     QString channelName;
     QColor usernameColor;
+    long receivedTime;
```
kornes, line 66: 2 things here:
- it will also hold backwards dates for historical messages, so calling it something like sentTime or produceTime would fit better
- not a fan of long, you could use QDateTime, then pass .time() for TimestampElement ctor and and use it in your sorting just as fine.

jammehcow: Thanks for the feedback. I believe I've resolved both of those concerns in my latest commits.

pajlada:
Overall looks good @jammehcow - I went ahead and extracted some of the changes to a separate PR with one small change: I kept the `toLocalTime()` in the irc helper
If you can take a peek at #3735 and approve it I'll get that merged in, and afterwards you can update your branch with the main branch and I'll review it further from there

src/widgets/helper/SearchPopup.cpp
```diff
@@ -126,24 +132,90 @@ void SearchPopup::updateWindowTitle()
...
+
     this->channelView_->setChannel(filter(this->searchInput_->text(),
-                                          this->channelName_, this->snapshot_,
-                                          this->channelFilters_));
+                                          this->channelName_, this->snapshot_));
+}
+
+LimitedQueueSnapshot<MessagePtr> SearchPopup::buildSnapshot()
+{
+    // no point in filtering/sorting if it's a single channel search
```
pajlada, line 153: Does this mean opening a search window in a channel that had filters applied will search outside of those filters after this PR?

pajlada (changes requested):

src/widgets/helper/SearchPopup.cpp
```diff
@@ -126,24 +132,90 @@ void SearchPopup::updateWindowTitle()
...
 }
 
+void SearchPopup::showEvent(QShowEvent *)
+{
+    this->search();
+}
+
 void SearchPopup::search()
 {
+    if (snapshot_.size() == 0)
```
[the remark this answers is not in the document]
pajlada: `LimitedQueueSnapshot` doesn't have an `empty()` function :-)

src/widgets/helper/SearchPopup.cpp
pajlada: Tested this myself and it works as expected, very nice

pajlada (approved):
Looks good to me :+1:

pajlada:
@jammehcow With your approval I'm ok with having this merged in as is - @Felanbird or I can help you out when you're happy!

jammehcow (commented):
@pajlada happy with that change, thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added multi-channel searching to search dialog via keyboard shortcut. (#3694)
```
```suggestion
- Major: Added multi-channel searching to search dialog via keyboard shortcut. [Ctrl+Shift+F by default] (#3694)
```
i'm sneaking this in

PR #3696, opened 2022-04-28
sort /vips output

badoge:
Pull request checklist:

# Description
Twitch returns display names when you use the /vips command so usernames that start with uppercase letters get placed before names that start with lowercase letters.


badoge:
https://twitch.uservoice.com/forums/310201-chat/suggestions/45092644-return-login-names-instead-of-display-names-when-u

surely they will reply and resolve this
image

Felanbird:
I'm fine with merging this change, I don't expect a better standardized `case-insensitivity` sorting to be implemented.

pajlada (approved):

Felanbird (approved):

PR #3711, opened 2022-05-06
fix: parse irc tags with consecutive escapes

iProdigy:
Pull request checklist:

# Description
When an irc tag has multiple escaped characters in a row, the current `parseTagString` function fails to see the 2nd `\` due to the extra `i++` (since the loop declaration already has `i++` and the string replacement eats two characters)

Example: `system-msg=DefectiveCloak\s\sgifted\sa\sTier\s1\ssub\sto\saliiscrying!` (weird case where their display name has a trailing space)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Minor: Added chatter count for each category in viewer list. (#3683)
 - Minor: Sorted usernames in /vips message to be case-insensitive. (#3696)
 - Minor: Added option to open a user's chat in a new tab from the usercard profile picture context menu. (#3625)
+- Minor: Fixed tag parsing for consecutive escaped characters. (#3711)
 - Bugfix: Fixed live notifications for usernames containing uppercase characters. (#3646)
 - Bugfix: Fixed live notifications not getting updated for closed streams going offline. (#3678)
 - Bugfix: Fixed certain settings dialogs appearing behind the main window, when `Always on top` was used. (#3679)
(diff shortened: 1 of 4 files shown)


Felanbird:
`@badges=subscriber/0,premium/1;color=#1E90FF;display-name=ItsVinc\s;emotes=80481:0-4;id=376529fd-31a8-4da9-9c0d-92a9470da2cd;login=itsvinc;mod=0;msg-id=resub;msg-param-months=2;msg-param-sub-plan-name=Dakotaz;msg-param-sub-plan=1000;room-id=11148817;subscriber=1;system-msg=ItsVinc\s\sjust\ssubscribed\swith\sa\sTier\s1\ssub.\sItsVinc\s\ssubscribed\sfor\s2\smonths\sin\sa\srow!;tmi-sent-ts=1528191154801;turbo=0;user-id=41174867;user-type= :tmi.twitch.tv USERNOTICE #pajlada :pajaW`

before change:
image

after change:
image

Took an existing debug resub message and inserted a user with a trailing space, seems the literal `\s` is replaced with a literal space, which while annoying to see, is likely better than seeing a literal `\s`.

(or maybe I made a mistake in this rawMSG and this change works exactly as intended, tune in to find out).

iProdigy:
yeah this PR doesn't add any trimming between words, just makes sure to convert all `\s` to spaces

Felanbird (approved):

pajlada (approved):
Added some tests and then reverted your fix, and saw the wrong behaviour you described :+1: Thanks for the fix!

PR #3713, opened 2022-05-06
fix: raise existing emote popups from behind other windows (macOS)

LosFarmosCTL:
Pull request checklist:

# Description

Chatterino tries to refocus an existing emote popup instead of opening a new one, which works fine until the existing emote popup is now behind some other application. In that case chatterino still focuses the existing emote popup, but doesn't actually bring it to the front, resulting in no visible window being focused (since the main application has now lost focus) and you having to search for the emote popup somewhere on your desktop.

It seems like the issue is macOS specific since on other platforms/WMs you apparently just can't hide the emote popup behind something else as long as chatterino is focused, so this change should not affect any of them.

[felanbird]: here's his `.mov` showing the issue that github doesn't like https://i.nuuls.com/PgiHt.mp4


pajlada (approved):
Doesn't change behaviour in i3, happy to merge in and see if it breaks on Windows later :-) should be fine

Felanbird (approved):
Fine on windows, wasn't an issue since the menu can't go behind the main window on windows.

PR #3715, opened 2022-05-07
Prevent user from entering incorrect characters in Live Notifications channels list

kornes:
Pull request checklist:

# Description
Live notifications for channels which are not open in tabs are queried by user supplied names, in single call to Helix /streams. 
Twitch will return error 400 Bad Request, when one of passed channel names will contain illegal characters.

That will cause chatterino to silently fail and not deliver any notifications, even for correct names on the list
```
chatterino.http: "GET 400 https://api.twitch.tv/helix/streams?user_login=_EsfandTV&user_login=georgehotz"
chatterino.notification: Failed to fetch live status for  ("_EsfandTV", "georgehotz")
```
This PR fixes that by adding input validation support for TableView in editor mode. 
User won't be allowed to enter characters not matching pattern of correct twitch username:
- no special characters at position 0
- following, only 0-9 A-Z a-z and _ 
- [min length of 1, max 25](https://discuss.dev.twitch.tv/t/max-length-for-user-names-and-display-names/21315/2) (although Helix currently doesn't care about this one)


src/util/Twitch.hpp
```diff
@@ -4,6 +4,8 @@
 
 namespace chatterino {
 
+QString const TWITCH_USERNAME_PATTERN("^[a-zA-Z0-9][a-zA-Z0-9_]{1,24}");
```
iProdigy, line 7:
```suggestion
QString const TWITCH_USERNAME_PATTERN("^[a-zA-Z0-9][a-zA-Z0-9_]{0,24}$");
```

To ensure length cannot exceed 25 chars & allows for one-character usernames (like [i](https://www.twitch.tv/i/about))

src/util/Twitch.hpp
```diff
@@ -4,6 +4,8 @@
 
 namespace chatterino {
 
+QString const TWITCH_USERNAME_PATTERN("^[a-zA-Z0-9][a-zA-Z0-9_]{0,24}$");
```
zneix, line 7:
```suggestion
QString const TWITCH_USERNAME_PATTERN("^[a-zA-Z0-9]\w{0,24}$");
```

pajlada (approved):

Issue #3133, opened 2021-08-06
Automod notifications get posted twice


badoge:
**Describe your issue**
Messages held by automod get posted twice sometimes. Approved message also don't get displayed by chatterino but they do get sent.
I can only reproduce the bug after switching accounts, so this is probably related to #1022 or maybe some recent change introduced this bug since I don't remember getting duplicate notifications before 2.3.4

**Screenshots**
![image](https://user-images.githubusercontent.com/18620902/128564133-c5e6baf9-2c08-48c6-b5ca-c2443f265287.png)

**OS and Chatterino Version**
Windows 10, Chatterino 2.3.4

Felanbird:
I have replicated this by accident during account switching.

> I don't remember getting duplicate notifications before 2.3.4

I do, but not specifically what commit I would have been on.

zneix:
This occurs most likely because PubSub connection is being created twice. I've seen this happen very recently (2.3.2/2.3.3). Gotta investigate when we should properly (UN)LISTEN from/to moderation topics to prevent this from happening.

Felanbird:
Likely? Fixed by #3643

badoge:
It now crashes when automod holds a message 

Chatterino Nightly 2.3.5 (commit [f97780d84](https://github.com/Chatterino/chatterino2/commit/f97780d84), built on ) running on Windows 10 Version 2009, kernel: 10.0.19044

badoge:
crash fixed by #3716 but double notifications are still there 
![image](https://user-images.githubusercontent.com/18620902/167267346-2a2417b7-7235-4e79-aac7-ac6e3563dbc1.png)

PR #3718, opened 2022-05-07
fix: unlisten to automod on user change

iProdigy:
Pull request checklist:

# Description
Calls `unlistenAutomod` on `currentUserChanged`

Edit: confirmed fixed by badoge, related to #3133


pajlada (approved):
Confirmed works :+1: Thanks for the fix!

PR #3722, opened 2022-05-10
Added support for Twitch's Chat Replies

### LINKED ISSUE #1912: Add support for reply threads

dnsge:
# Description
Adds message reply thread support (https://help.twitch.tv/s/article/chat-basics?language=en_US#replies)

 - Only supported for `TwitchChannel`s
 - Upon clicking the reply button, the user is presented with a reply dialog that shows the message thread and allows for typing responses with the same features as the standard input box (emote autocomplete, for example)
 - New replies are automatically added to the reply dialog similar to the user info dialog
 - `/reply <username> <message>` command sends a reply to the given user's most recent message

Design choices:
 - The message being replied to is rendered, but only the first line like in Twitch
 - The message being replied to does NOT render emotes and shows the raw text instead

## The reply button

I initially tried only rendering the reply button when hovering a message, but the message rendering pipeline makes this pretty much impossible. The tiny reply button was the best compromise to having an accessible but not too intrusive option.

The reply buttons are not visible by default but can be enabled in settings if a user wants a single-click way to reply to a message.

## Screenshots
<details>
  <summary>Expand</summary>

**Reply to message option**:
Screen Shot 2022-05-29 at 10 41 27 PM

**Inline message replying**
Screen Shot 2022-05-29 at 10 41 39 PM

**Message with reply**:
Screen Shot 2022-05-29 at 10 41 59 PM

**Message thread dialog**:
Screen Shot 2022-05-29 at 10 42 05 PM
Screen Shot 2022-05-29 at 10 42 12 PM

**Reply buttons enabled**:
Screen Shot 2022-05-29 at 10 44 29 PM

</details>
Closes #1912

# Pull request checklist:


src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -248,6 +275,12 @@ MessagePtr TwitchMessageBuilder::build()
     // words
     QStringList splits = this->originalMessage_.split(' ');
 
+    if (this->thread_ && !splits.isEmpty() && splits[0].startsWith('@'))
+    {
+        // If this is a reply, remove the first word which should be "@username"
+        splits.pop_front();
+    }
+
```
zneix, line 283: I'm 100% against modifying actual message content. I want to be able to get the exact content of the message, not get lied to. this only complicates things with bots, mentions, selecting whole message text... besides, if users would try to execute bot commands or send a banphrase message, they'd get confused why bot doesn't react because they'd think that a message starts with "foo", not "@username foo", but in reality the latter is the actual beginning of the message content which I believe would cause issues if changed

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,120 @@
...
+                        // same reply thread, add message
+                        virtualChannel->addMessage(message);
+                    }
+                }));
+}
+
+void ReplyThreadPopup::updateInputPlaceholder()
+{
+    auto channel = this->split_->getChannel();
+    if (!channel || !channel->isTwitchChannel())
```
zneix, line 98: you'd need to check for /whispers, /mentions, /live etc. as well, since these return true for isTwitchChannel

zneix (changes requested):

src/providers/twitch/TwitchMessageBuilder.cpp
dnsge: I agree with your reasoning. I also found that removing the username messes up Twitch emote rendering, so this will be removed.

kornes:
From 30mins testing: 
- icons visible at all times are no go, too intrusive
- icons don't scale well with zoom changes [screen](https://user-images.githubusercontent.com/28986062/167927806-1ec71aa2-e3ea-4dc9-8e70-e25169e1e6ee.png)
- window titles should include channel of origin
- allowing single instance of replay window per target username (or even per channel) would be cool to minimize mess
- we shouldn't be allowed to replay own messages
- full thread view (with replays from both sides) shouldn't include "Replaying to" rows img2
- i would make whole SingleLineTextElement clickable to open thread view for easier access

Overall looks good, but I think it would be better to remove icons/settings altogether from initial implementation in this PR. Context menu link to initiate replay and whole "Replying to @user" row clickable to answer, would be good in my book.

Mm2PL:
> - allowing single instance of reply window per target username (or even per channel) would be cool to minimize mess

There is no limit to usercards, search windows. I don't see why this should have a limit.

> - we shouldn't be allowed to reply own messages

Why? IRC doesn't care about this. Webchat does, but webchat isn't the one-and-only chat experience.

kornes:
i didn't argue its technically wrong, it will just provide better use experience if u don't have to close each replay window manually and just open new one to replace. You would talk with given person in 1 thread anyway, so whats the use of multiple instances? Its same reasoning emote picker has one instance. Agree limiting per channel is too harsh, edited my comment. 

About allowing starting threads with myself, while good for testing i don't see how that's useful for regular mentally stable user :D

src/widgets/dialogs/ReplyThreadPopup.cpp
dnsge: Resolved in 51e478d

src/providers/twitch/TwitchMessageBuilder.cpp
dnsge: 306f7a7

dnsge:
I've made several changes, including:
- > window titles should include channel of origin 
- > full thread view (with replies from both sides) shouldn't include "Replying to" rows
- > `SingleLineTextElement` clickable to open thread view for easier access
- autofocusing the input box
- `/reply` command
- option to automatically close the reply dialog on loss of focus, like with the user info dialog

The screenshots in the original post have been updated with the changes.

kornes:
`/r` is in conflict with https://github.com/Chatterino/chatterino2/blob/efbf03d05b83092abaf09f909c2d77c0c77f6ab3/src/widgets/splits/SplitInput.cpp#L695 but seems like twitch is trying to kill /w sooner than later, so probably could be removed

Felanbird:
> but seems like twitch is trying to kill /w sooner than later, so probably could be removed

No, we have no public indication that whispers are going anywhere, we are not going to action on that until then.

dnsge:
> `/r` is in conflict with
> 
> https://github.com/Chatterino/chatterino2/blob/efbf03d05b83092abaf09f909c2d77c0c77f6ab3/src/widgets/splits/SplitInput.cpp#L695
> 

Whoops, missed that. I've removed that alias. Hopefully having `/reply` for reply threads and `/r` for whispers doesn't lead to any confusion.

src/messages/Link.hpp
```diff
@@ -23,6 +23,8 @@ struct Link {
         JumpToChannel,
         Reconnect,
         CopyToClipboard,
+        ReplyToMessage,
+        ViewThread
```
23rd, line 27:
```suggestion
        ViewThread,
```

src/messages/MessageElement.cpp
```diff
@@ -132,6 +134,31 @@ void ImageElement::addToContainer(MessageLayoutContainer &container,
...
+{
+}
+
+void PrettyImageElement::addToContainer(MessageLayoutContainer &container,
+                                        MessageElementFlags flags)
+{
+    if (flags.hasAny(this->getFlags()))
+    {
+        auto size = QSize(this->image_->width() * container.getScale(),
+                          this->image_->height() * container.getScale());
```
23rd, line 153: Can be simplified.
`QSize(this->image_->width(), this->image_->height()) * container.getScale()`

src/messages/MessageElement.cpp
```diff
@@ -395,6 +422,140 @@ void TextElement::addToContainer(MessageLayoutContainer &container,
...
+                auto image =
+                    emote->images.getImageOrLoaded(container.getScale());
+                if (!image->isEmpty())
+                {
+                    auto emoteScale = getSettings()->emoteScale.getValue();
+
+                    auto size = QSize(
+                        int(container.getScale() * image->width() * emoteScale),
+                        int(container.getScale() * image->height() *
+                            emoteScale));
```
23rd, line 499: Can be simplified.
`auto size = QSize(mage->width(), image->height()) * emoteScale * container.getScale();`

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -346,6 +353,21 @@ MessageLayoutElement *MessageLayoutContainer::getElementAt(QPoint point)
     return nullptr;
 }
 
+FloatingMessageLayoutElement *MessageLayoutContainer::getFloatingElementAt(
+    QPoint point)
+{
+    for (std::unique_ptr<FloatingMessageLayoutElement> &element :
+         this->floatingElements_)
```
23rd, line 360:
```suggestion
    for (const auto &element : this->floatingElements_)
```

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -205,6 +245,42 @@ void ImageWithBackgroundLayoutElement::paint(QPainter &painter)
     }
 }
 
+PrettyFloatingImageLayoutElement::PrettyFloatingImageLayoutElement(
+    MessageElement &creator, ImagePtr image, const QSize &size, int padding,
+    QColor background)
+    : FloatingMessageLayoutElement(creator,
+                                   size + QSize(padding * 2, padding * 2))
```
23rd, line 252:
```suggestion
                                   size + QSize(padding, padding) * 2)
```

dnsge: Changed in 936ae06

src/messages/layouts/MessageLayoutContainer.cpp
dnsge: Changed in 936ae06

src/messages/MessageElement.cpp
dnsge: Changed in 936ae06

src/messages/MessageElement.cpp
dnsge: Changed in 936ae06

src/messages/Link.hpp
dnsge: Changed in 936ae06

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    if (channel->isWritable())
+    {
+        this->ui_.replyInput->show();
+    }
+    else
+    {
+        this->ui_.replyInput->hide();
+    }
```
23rd, line 125:
```suggestion
    this->ui_.replyInput->setVisible(channel->isWritable());
```

src/widgets/dialogs/ReplyThreadPopup.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+private:
+    void addMessagesFromThread();
+    void updateInputUI();
+
+    // The message reply thread
+    std::shared_ptr<const MessageThread> thread_;
+    // The channel that the reply thread is in
+    ChannelPtr channel_;
+    Split *split_;
```
23rd, line 37: Can be `const` I guess.
```suggestion
    const Split *split_;
```

src/widgets/splits/ReplyInput.cpp
```diff
@@ -0,0 +1,71 @@
...
+    if ((this->prevMsg_.isEmpty() || !this->prevMsg_.endsWith(message)) &&
+        !message.trimmed().isEmpty())
+    {
+        this->prevMsg_.append(message);
+    }
+    bool shouldClearInput = true;
+    if (arguments.size() != 0 && arguments.at(0) == "keepInput")
+    {
+        shouldClearInput = false;
+    }
```
23rd, line 55:
```suggestion
    auto shouldClearInput = (arguments.empty() || arguments.at(0) != "keepInput");
```

src/widgets/dialogs/ReplyThreadPopup.hpp
dnsge: `ReplyThreadPopup` uses a non-const member function of `Split` (`getChannel()`, though I'm not sure the method should be non-const).

src/widgets/splits/ReplyInput.cpp
dnsge: Changed in 464851b

src/widgets/dialogs/ReplyThreadPopup.cpp
dnsge: Changed in 47aa13a

23rd:
I think the `Reply to message` menu item should be hidden from the `Reply Thread` window.

dnsge:
> I think the Reply to message menu item should be hidden from the Reply Thread window.

Updated in 5a35495 👍

23rd:
Another thing: I noticed that nicknames are always displayed in lower case in reply-line.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -204,11 +204,11 @@ MessagePtr TwitchMessageBuilder::build()
                 MessageColor::System, FontStyle::ChatMediumSmall)
             ->setLink({Link::ViewThread, this->thread_->rootId()});
 
-        this->emplace<TextElement>("@" + threadRoot->loginName + ":",
+        this->emplace<TextElement>("@" + threadRoot->displayName + ":",
```
pajlada, line 207: Preferably we should use the user's display name setting here

pajlada (changes requested):
The images you've used contain the png files only, if you've made these from svg files it would be nice to include those as well so we can always rescale things in the future

1. I'd prefer it if threads looked more like discord, where there's an arrow-like indicator showing the original message
image
2. Replying to a message should preferably do it inline in the message box instead, as discord does it - I don't have a super clear idea of what it would look like in Chatterino since we don't generally do resizing
image

src/providers/twitch/TwitchMessageBuilder.cpp
dnsge: Updated in f20538f

dnsge:
Thanks for the feedback.

> 1. I'd prefer it if threads looked more like discord, where there's an arrow-like indicator showing the original message
>    image

I added a similar curved line to Discord's, and I think it makes the reply look a lot better (699b277). See the OP comment for updated screenshots.

> 2. Replying to a message should preferably do it inline in the message box instead, as discord does it - I don't have a super clear idea of what it would look like in Chatterino since we don't generally do resizing
>    image

I was able to implement this (4aef01f), and I'm mostly happy with the result. Styling the position and appearance of the cancel button through code is a pain but I think it came out decent. Again, see the updated screenshots in the OP.

src/autogenerated/ResourcesAutogen.cpp
```diff
@@ -44,6 +46,7 @@ Resources2::Resources2()
     this->buttons.vip = QPixmap(":/buttons/vip.png");
     this->error = QPixmap(":/error.png");
     this->icon = QPixmap(":/icon.png");
+    this->icons.downLeftReply = QPixmap(":/icons/downLeftReply.png");
```
23rd, line 49: Why it can't be drawn with QPainter?

zneix:
I've just tried building to test the PR and discovered that `src/messages/layouts/MessageLayoutElement.hpp` is missing `<QPen>` include. Because of that building without precompiled headers failed.
I'd push the change myself but I don't want to mess with your branch.

I've noticed that parsed messages (e.g. loaded from recent-messages) which are replies don't get rendered as these. 

Another issue is that dialog with the thread gets spawned in tiling mode on i3wm. Trying to get it to floating mode closes the window - most likely due to focus loss which idk if I really like that these close upon focus loss (and also not close upon Esc press). I think some flags that make them proper dialog windows, like streamlink quality picker or whatever usercard dialog uses would have to be applied to mitigate at least the tiling/floating issue and/or also focus loss, closing with Esc.
Might be unrelated but smooth scrolling also makes the first message scroll itself up, so it's not visible and second message in the thread is at the top.

I'll give more feedback after more testing, changes look promising.

dnsge:
We can definitely fix the issues with the popup behavior. I had decided to not bother implementing replies for historical messages to reduce the complexity (and I'm not sure how easy/fast it would be as I'm not sure how historical messages are processed) but if we want we can work on that.

As a sidenote, one issue that I've come across is a segfault after doing a very specific action: viewing a reply thread in the popup and then changing the channel in the parent split. See this video:

https://user-images.githubusercontent.com/24928223/171067837-8de8ff87-5bd4-4d9c-ad8d-652b1598ac46.mov

I've spent time trying to track down what exactly is happening and why opening a reply dialog (which only takes a raw `*Split` pointer) would cause this behavior. I'm worried it's related to some smart pointer strangeness or something with how Qt works. It would be great if anyone else could also take a look at this problem :)

kornes:
channelChanged signal invoked here 
https://github.com/Chatterino/chatterino2/blob/e9e3e5a25a3d54c6d90b8e9bab350188bcbff32c/src/widgets/splits/Split.cpp#L690
is handled and crashes here
https://github.com/Chatterino/chatterino2/blob/e9e3e5a25a3d54c6d90b8e9bab350188bcbff32c/src/widgets/splits/SplitInput.cpp#L43
because getChannel() is called on null split_, so destroying reply window when out of focus is causing it
https://github.com/Chatterino/chatterino2/blob/b060aa8600898da747104463a729a5ab88a88cb3/src/widgets/dialogs/ReplyThreadPopup.cpp#L21

dnsge:
> split destructor is never called when changing channel to different one

Yeah, this is the most confusing part about it to me. I don't understand why the pointer to the split goes bad when the split should not go anywhere when the channel is changed.

src/messages/Message.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "common/FlagsEnum.hpp"
+#include "messages/MessageThread.hpp"
```
23rd, line 4:
```suggestion
```

src/messages/MessageThread.cpp
```diff
@@ -0,0 +1,58 @@
+#include "MessageThread.hpp"
+
+#include "util/DebugCount.hpp"
```
23rd, line 3:
```suggestion
#include "messages/Message.hpp"
#include "util/DebugCount.hpp"
```

src/messages/MessageThread.hpp
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include "messages/Message.hpp"
+
```
23rd, line 4:
```suggestion
```

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -76,6 +76,13 @@ void MessageLayoutContainer::addElement(MessageLayoutElement *element)
     this->_addElement(element);
 }
 
+void MessageLayoutContainer::addFloatingElement(
+    FloatingMessageLayoutElement *element)
+{
+    this->floatingElements_.push_back(
+        std::unique_ptr<FloatingMessageLayoutElement>(element));
```
23rd, line 83:
```suggestion
    this->floatingElements_.emplace_back(element);
```

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -402,4 +477,74 @@ int TextIconLayoutElement::getXFromIndex(int index)
...
+ReplyCurveLayoutElement::ReplyCurveLayoutElement(MessageElement &creator,
+                                                 const QSize &size,
+                                                 float thickness,
+                                                 float neededMargin)
+    : MessageLayoutElement(creator, size)
+    , neededMargin_(neededMargin)
+{
+    this->pen_.setColor(QColor("#888"));
+    this->pen_.setWidthF(thickness);
+    this->pen_.setCapStyle(Qt::RoundCap);
```
23rd, line 489:
```suggestion
    , pen_(QColor("#888"), thickness, Qt::SolidLine, Qt::RoundCap)
    , neededMargin_(neededMargin)
{
```

src/messages/layouts/MessageLayoutElement.hpp
```diff
@@ -142,4 +183,24 @@ class TextIconLayoutElement : public MessageLayoutElement
...
+    void paintAnimated(QPainter &painter, int yOffset) override;
+    int getMouseOverIndex(const QPoint &abs) const override;
+    int getXFromIndex(int index) override;
+    void addCopyTextToString(QString &str, int from = 0,
+                             int to = INT_MAX) const override;
+    int getSelectionIndexCount() const override;
+
+private:
+    QPen pen_;
+    float neededMargin_;
```
23rd, line 203:
```suggestion
    const QPen pen_;
    const float neededMargin_;
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -395,6 +404,33 @@ void TwitchChannel::sendMessage(const QString &message)
...
+{
+    auto app = getApp();
+    if (!app->accounts->twitch.isLoggedIn())
+    {
+        if (message.isEmpty())
+        {
+            return;
+        }
+
+        this->showLoginMessage();
```
23rd, line 420:
```suggestion
        if (!message.isEmpty())
        {
            this->showLoginMessage();
        }

```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -405,6 +441,37 @@ void TwitchChannel::sendMessage(const QString &message)
...
```
23rd, line 454:
```suggestion
        if (!message.isEmpty())
        {
            this->showLoginMessage();
        }

```

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,147 @@
+#include "ReplyThreadPopup.hpp"
+
+#include "Application.hpp"
+#include "common/Channel.hpp"
+#include "util/LayoutCreator.hpp"
+#include "widgets/helper/ChannelView.hpp"
+#include "widgets/helper/ResizingTextEdit.hpp"
```
23rd, line 7:
```suggestion
#include "messages/MessageThread.hpp"
#include "util/LayoutCreator.hpp"
#include "widgets/helper/ChannelView.hpp"
#include "widgets/splits/ReplyInput.hpp"
#include "widgets/helper/ResizingTextEdit.hpp"
#include "widgets/splits/Split.hpp"
```

src/widgets/dialogs/ReplyThreadPopup.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include "messages/MessageThread.hpp"
+#include "widgets/BaseWindow.hpp"
+#include "widgets/splits/ReplyInput.hpp"
+#include "widgets/splits/Split.hpp"
```
23rd, line 6:
```suggestion
#include "widgets/BaseWindow.hpp"
#include "ForwardDecl.hpp"
```

src/widgets/dialogs/ReplyThreadPopup.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <boost/signals2.hpp>
+#include <pajlada/signals/scoped-connection.hpp>
+#include <pajlada/signals/signal.hpp>
+
+#include <memory>
+
+namespace chatterino {
+
```
23rd, line 15:
```suggestion

class MessageThread;
class ReplyInput;
class Split;
```

src/widgets/splits/ReplyInput.cpp
```diff
@@ -0,0 +1,56 @@
+#include "ReplyInput.hpp"
+
+#include "Application.hpp"
+#include "controllers/commands/CommandController.hpp"
+#include "controllers/hotkeys/HotkeyController.hpp"
+#include "widgets/helper/ResizingTextEdit.hpp"
+
```
23rd, line 7:
```suggestion
#include "messages/MessageThread.hpp"
#include "widgets/splits/Split.hpp"

```

src/widgets/splits/ReplyInput.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "messages/MessageThread.hpp"
+#include "widgets/splits/Split.hpp"
```
23rd, line 4:
```suggestion
```

src/widgets/splits/ReplyInput.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include "widgets/splits/SplitInput.hpp"
+
+#include <memory>
+
+namespace chatterino {
+
```
23rd, line 10:
```suggestion

class MessageThread;
class Split;

```

src/widgets/splits/Split.hpp
```diff
@@ -3,6 +3,7 @@
 #include "common/Aliases.hpp"
 #include "common/Channel.hpp"
 #include "common/NullablePtr.hpp"
+#include "messages/MessageThread.hpp"
```
23rd, line 6: Can be moved to `Split.cpp` and replaced with a forward declaration `class MessageThread;`.
```suggestion
```

src/widgets/splits/SplitInput.cpp
```diff
@@ -3,7 +3,6 @@
 #include "Application.hpp"
 #include "common/QLogging.hpp"
 #include "controllers/commands/CommandController.hpp"
-#include "controllers/hotkeys/HotkeyController.hpp"
 #include "messages/Link.hpp"
```
23rd, line 6:
```suggestion
#include "controllers/hotkeys/HotkeyController.hpp"
#include "messages/Link.hpp"
#include "messages/MessageThread.hpp"
```

src/widgets/splits/SplitInput.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "controllers/hotkeys/HotkeyController.hpp"
+#include "messages/MessageThread.hpp"
```
23rd, line 4: A forward declaration `class MessageThread;` should be placed below.
```suggestion
```

dnsge:
I've added viewing replies for historical messages in 5af3358. The only remaining possible problem with historical message reply rendering is when the very first message received in a channel is a reply to a message that hasn't been loaded from the historical messages API, but I don't think this should be a concern. It would be difficult to deal with anyways because messages are immutable after being built.

src/messages/MessageThread.cpp
```diff
@@ -0,0 +1,59 @@
+#include "MessageThread.hpp"
+
+#include "messages/Message.hpp"
+#include "util/DebugCount.hpp"
+
+namespace chatterino {
+
+MessageThread::MessageThread(const std::shared_ptr<const Message> &rootMessage)
+    : rootMessageId_(rootMessage->id)
+    , rootMessage_(rootMessage)
```
pajlada, line 10: Copy in and move
```suggestion
MessageThread::MessageThread(std::shared_ptr<const Message> rootMessage)
    : rootMessageId_(rootMessage->id)
    , rootMessage_(std::move(rootMessage))
```

src/messages/MessageThread.hpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    const std::vector<std::weak_ptr<const Message>> &replies() const
+    {
+        return replies_;
+    }
+
+private:
+    QString rootMessageId_;
+    std::shared_ptr<const Message> rootMessage_;
```
pajlada, line 43: Can be made const
```suggestion
    const QString rootMessageId_;
    const std::shared_ptr<const Message> rootMessage_;
```

pajlada (changes requested):
Overall looks really cool, happy that chat commands work in replies still.
Some small changes to `MessageThread`

Will take a look at the crash to see if I can make heads or tails of it

pajlada:
The crash you've noticed is fixed by https://github.com/Chatterino/chatterino2/pull/3799

src/messages/MessageThread.cpp
dnsge: Updated in 9df01c4

src/messages/MessageThread.hpp
dnsge: Updated in 9df01c4

Mm2PL:
> The only remaining possible problem with historical message reply rendering is when the very first message received in a channel is a reply to a message that hasn't been loaded from the historical messages API, but I don't think this should be a concern. It would be difficult to deal with anyways because messages are immutable after being built.

Is it not possible to use the IRC tags for this? Or maybe just show that the message is a reply but the original is unavailable?

> ### tags on incoming PRIVMSGs
> [...]
> - Optional tags: 
>  - [...]
>  - `reply-parent-display-name` Display name of the person who sent the message this is a reply to.
>  - `reply-parent-msg-body` Message body of the message this is a reply to.
>  - `reply-parent-msg-id` Message id of the message this is a reply to.
>  - `reply-parent-user-id` User ID of the author of the message this is a reply to.
>  - `reply-parent-user-login` Login of the author of that message.



<details>
<summary>Example reply with the tags</summary>

```
@badge-info=;badges=vip/1,glhf-pledge/1;color=#DAA520;display-name=Mm2PL;emotes=;first-msg=0;flags=;id=cf635d8c-67ab-425f-b83a-4f3c444a15be;mod=0;reply-parent-display-name=Mm_sUtilityBot;reply-parent-msg-body=@Mm2PL\sHi\smm2pl\s(Your\sID\sin\sbot\sdatabase:\suser\s#1,\saccount\s#1,\srole:\sadmin),\sthis\sis\smm_sutilitybot\sv3,\sa\sstupid\sgo\sbot.\sRunning\sfor\s54m51s.\sPing\sto\sTMI:\s157ms;reply-parent-msg-id=43e108b9-d3b7-4f60-bf04-751cede9e67c;reply-parent-user-id=442600612;reply-parent-user-login=mm_sutilitybot;room-id=31400525;subscriber=0;tmi-sent-ts=1654444863155;turbo=0;user-id=117691339;user-type= :mm2pl!mm2pl@mm2pl.tmi.twitch.tv PRIVMSG #supinic :@Mm_sUtilityBot -tags
```

</details>

dnsge:
> Is it not possible to use the IRC tags for this? Or maybe just show that the message is a reply but the original is unavailable?

I wasn't sure about how to go about this, but I ended up implementing it pretty quickly after you asked about it. Now, the message is rendered according to the display name and body tags sent along with the reply. 

Screen Shot 2022-06-05 at 2 38 28 PM

jammehcow:
The right click context menu of resub messages allows you to send a reply to the resub message even though replies to resub messages aren't possible. Message will send as a normal message.

pajlada:
To expand on jammehcow's message above

image
So the announcement "jammehcow subscribed at Tier 1" does not give options to reply, as expected.
However, the resub message "LULE i subed" shows the reply button/reply context menu option while it's not "repliable"

The resub message in this picture is this:
`@badge-info=subscriber/3;badges=subscriber/3,no_audio/1;color=#EBA2C0;display-name=jammehcow;emotes=;flags=;id=ff348bdf-c823-4fa3-994c-ef57e094b810;login=jammehcow;mod=0;msg-id=resub;msg-param-cumulative-months=3;msg-param-months=0;msg-param-multimonth-duration=0;msg-param-multimonth-tenure=0;msg-param-should-share-streak=0;msg-param-sub-plan-name=Hackerman\sClub\smembership;msg-param-sub-plan=1000;msg-param-was-gifted=false;room-id=31400525;subscriber=1;system-msg=jammehcow\ssubscribed\sat\sTier\s1.\sThey've\ssubscribed\sfor\s3\smonths!;tmi-sent-ts=1656757923655;user-id=82674227;user-type= :tmi.twitch.tv USERNOTICE #supinic :LULE i subed` so it's a `USERNOTICE` instead of a `PRIVMSG`

while this is an example of a normal PRIVMSG
`@badge-info=subscriber/46;badges=moderator/1,subscriber/12;color=#1E90FF;display-name=Supibot;emotes=305954156:0-7/1886493:92-101;first-msg=0;flags=;id=809d136a-b9ba-4010-b2d3-c54737223056;mod=1;returning-chatter=0;room-id=31400525;subscriber=1;tmi-sent-ts=1656757923910;turbo=0;user-id=68136884;user-type=mod :supibot!supibot@supibot.tmi.twitch.tv PRIVMSG #supinic :PogChamp Clap jammehcow just resubscribed for 3 months! Welcome back to the Hackerman Club! supiniHack`

pajlada:
Feel-wise of the PR, it's starting to feel pretty good! It felt like closing the popup using Escape was missing, so I added this functionality in.

I also run into an issue on i3 where the thread sometimes moves itself to the top left of my screen (the requested coords are -15, -15). This does not happen with UserInfoPopup but I assume it's some i3-specific bug so I don't really care to put too much time into this.

Only thing I'd like to have a second eye on is the look/feel of the reply button. It looks kind of ugly with its gray background but I'm not a designer.

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -40,9 +43,56 @@ ReplyThreadPopup::ReplyThreadPopup(bool closeAutomatically, QWidget *parent,
...
+             }
+             else
+             {
+                 qCWarning(chatterinoHotkeys) << "Unknown scroll direction";
+             }
+             return "";
+         }},
+        {"execModeratorAction", nullptr},
+
+        // these actions make no sense in the context of a usercard, so they aren't implemented
```
leon-richardt, line 87: This is not a user card

dnsge: Resolved in 30a1c8d

dnsge:
Thanks for adding those couple of things. I definitely didn't like how the reply button looked, so I redesigned it again:
New Reply Button

I think this looks a lot better, and the button will no longer barely overlap some content. I also fixed the button/context menu options being available where they shouldn't:
No Reply Button

pajlada (approved):
I'm happy with this feature as is. Polishing the looks can be done in future PRs. Any "bad code" I find is there because the structure is bad, so nothing can be done without refactoring a bunch of other things at the same time (thinking about message building)

Would like final opinions/reviews from more contributors before merging this in.

I'm personally looking forward to being able to use `/reply`!

dnsge:
Just pushed a commit that cleans up some old unused code that frankly was a bad idea to implement in the first place. Should be good to go now. 😃

leon-richardt (changes requested):
Overall, great feature! It feel really nice from a usability standpoint already.

I have a few comments, mostly relating to ownership semantics when passing `shared_ptr`s. (See mainly [this C++ Core Guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-owner) for that.) I usually requested changes in only the implementation so remember to also change the declaration accordingly as well.

src/widgets/splits/SplitInput.cpp
```diff
@@ -814,4 +1005,17 @@ void SplitInput::resizeEvent(QResizeEvent *)
...
+    this->ui_.textEdit->setFocus(reason);
+}
+
+void SplitInput::setReply(const std::shared_ptr<MessageThread> &reply)
+{
+    this->replyThread_ = reply;
+    this->ui_.textEdit->setPlainText("@" + reply->root()->displayName + " ");
+    this->ui_.textEdit->moveCursor(QTextCursor::EndOfBlock);
+    this->ui_.replyLabel->setText("Replying to @" + reply->root()->displayName);
+}
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    this->ui_.replyInput->setVisible(channel->isWritable());
+
+    auto user = getApp()->accounts->twitch.getCurrent();
+    QString placeholderText;
+
+    if (user->isAnon())
+    {
+        placeholderText = "Log in to send messages...";
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/messages/MessageElement.cpp
```diff
@@ -395,6 +422,138 @@ void TextElement::addToContainer(MessageLayoutContainer &container,
...
+                                             const MessageColor &color,
+                                             FontStyle style)
+    : MessageElement(flags)
+    , color_(color)
+    , style_(style)
+{
+    for (const auto &word : text.split(' '))
+    {
+        this->words_.push_back({word, -1});
+        // fourtf: add logic to store multiple spaces after message
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

I had left this comment in because I wasn't sure what fourtf was trying to do, but because a SingleLineTextElement doesn't have anything after it, I decided to just remove this comment.

src/controllers/commands/CommandController.cpp
```diff
@@ -1114,6 +1116,61 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                    else
+                    {
+                        thread = msg->replyThread;
+                    }
+
+                    QString reply;
+                    for (int n = 2; n < words.size(); ++n)
+                    {
+                        reply += words[n] + ' ';
+                    }
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/controllers/commands/CommandController.cpp
```diff
@@ -1114,6 +1116,61 @@ void CommandController::initialize(Settings &, Paths &paths)
         return "";
     });
 
+    this->registerCommand(
+        "/reply", [](const QStringList &words, ChannelPtr channel) {
+            auto *twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
+            if (twitchChannel == nullptr)
+            {
+                channel->addMessage(makeSystemMessage(
+                    "The /reply command only works in Twitch Channels"));
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -402,4 +441,72 @@ int TextIconLayoutElement::getXFromIndex(int index)
...
+
+int ReplyCurveLayoutElement::getXFromIndex(int index)
+{
+    if (index <= 0)
+    {
+        return this->getRect().left();
+    }
+    else if (index == 1)
+    {
+        // fourtf: remove space width
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/providers/twitch/TwitchMessageBuilder.hpp
```diff
@@ -45,6 +46,8 @@ class TwitchMessageBuilder : public SharedMessageBuilder
     void triggerHighlights() override;
     MessagePtr build() override;
 
+    void setThread(const std::shared_ptr<MessageThread> &thread);
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -242,6 +242,84 @@ void IrcMessageHandler::handlePrivMessage(Communi::IrcPrivateMessage *message,
...
+{
+    const auto &tags = message->tags();
+    if (const auto it = tags.find("reply-parent-msg-id"); it != tags.end())
+    {
+        const QString replyID = it.value().toString();
+        auto threadIt = channel->threads_.find(replyID);
+        if (threadIt != channel->threads_.end() && !threadIt->second.expired())
+        {
+            // Thread already exists (has a reply)
+            builder.setThread(threadIt->second.lock());
```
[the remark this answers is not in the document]
dnsge: Good catch, updated in e0cabce

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -242,6 +242,84 @@ void IrcMessageHandler::handlePrivMessage(Communi::IrcPrivateMessage *message,
...
+            // Linear search is justified by the infrequent use of replies
+            for (auto &otherMsg : otherLoaded)
+            {
+                if (otherMsg->id == replyID)
+                {
+                    // Found root reply message
+                    std::shared_ptr<MessageThread> newThread =
+                        std::make_shared<MessageThread>(otherMsg);
+
+                    // Store weak reference to thread
```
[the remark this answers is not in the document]
dnsge: Good point, comment was in a weird place. Updated in e0cabce

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -301,6 +379,32 @@ void IrcMessageHandler::addMessage(Communi::IrcMessage *_message,
...
+            {
+                // Found root reply message
+                std::shared_ptr<MessageThread> newThread =
+                    std::make_shared<MessageThread>(root);
+
+                // Store weak reference to thread
+                builder.setThread(newThread);
+                channel->addReplyThread(newThread);
+            }
+        }
```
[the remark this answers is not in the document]
dnsge: As above, updated in e0cabce

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -395,6 +404,32 @@ void TwitchChannel::sendMessage(const QString &message)
...
+        if (!message.isEmpty())
+        {
+            this->showLoginMessage();
+        }
+
+        return;
+    }
+
+    qCDebug(chatterinoTwitch)
+        << "[TwitchChannel" << this->getName() << "] Send message:" << message;
```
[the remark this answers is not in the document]
dnsge: `qDebug()` actually inserts spaces around the arguments that it logs, so these spaces aren't needed.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -405,6 +440,36 @@ void TwitchChannel::sendMessage(const QString &message)
...
+        if (!message.isEmpty())
+        {
+            this->showLoginMessage();
+        }
+
+        return;
+    }
+
+    qCDebug(chatterinoTwitch) << "[TwitchChannel" << this->getName()
+                              << "] Send reply message:" << message;
```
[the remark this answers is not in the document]
dnsge: See above, unneeded

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -311,46 +332,34 @@ boost::optional<ChannelPointReward> TwitchChannel::channelPointReward(
...
-
-    qCDebug(chatterinoTwitch)
-        << "[TwitchChannel" << this->getName() << "] Send message:" << message;
+    const auto linkColor = MessageColor(MessageColor::Link);
+    const auto accountsLink = Link(Link::OpenAccountsPage, QString());
+    const auto currentUser = getApp()->accounts->twitch.getCurrent();
+    const auto expirationText =
+        QString("You need to log in to send messages. You can link your "
+                "Twitch account");
+    const auto loginPromptText = QString("in the settings.");
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -773,6 +838,7 @@ void TwitchChannel::loadRecentMessages()
             auto &handler = IrcMessageHandler::instance();
 
             std::vector<MessagePtr> allBuiltMessages;
+            auto tc = dynamic_cast<TwitchChannel *>(shared.get());
```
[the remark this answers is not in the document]
dnsge: Whoops. Removed in e0cabce

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -118,6 +123,11 @@ class TwitchChannel : public Channel, public ChannelChatters
     // Cheers
     boost::optional<CheerEmote> cheerEmote(const QString &string);
 
+    // Replies
+    void addReplyThread(const std::shared_ptr<MessageThread> &thread);
```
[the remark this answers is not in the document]
dnsge: Comment added in e0cabce

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -164,13 +173,16 @@ class TwitchChannel : public Channel, public ChannelChatters
     const QString &getDisplayName() const override;
     const QString &getLocalizedName() const override;
 
+    QString prepareMessage(Application *app, const QString &message) const;
```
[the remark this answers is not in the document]
dnsge: Yeah, I had added the argument because all two callsites already had a pointer to the application instance and it felt weird calling `getApp` again inside `prepareMessage`, but I agree it makes more sense to do so given it's a singleton.

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -370,66 +375,91 @@ bool TwitchIrcServer::hasSeparateWriteConnection() const
...
-        std::lock_guard<std::mutex> guard(this->lastMessageMutex_);
+    std::lock_guard<std::mutex> guard(this->lastMessageMutex_);
 
-        //        std::queue<std::chrono::steady_clock::time_point>
-        auto &lastMessage = channel->hasHighRateLimit()
-                                ? this->lastMessageMod_
-                                : this->lastMessagePleb_;
-        size_t maxMessageCount = channel->hasHighRateLimit() ? 99 : 19;
-        auto minMessageOffset = (channel->hasHighRateLimit() ? 100ms : 1100ms);
+    //        std::queue<std::chrono::steady_clock::time_point>
```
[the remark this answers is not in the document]
dnsge: Removed in e0cabce

src/widgets/DraggablePopup.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+class DraggablePopup : public BaseWindow
+{
+    Q_OBJECT
+
+public:
+    /// DraggablePopup implements the automatic dragging behavior when clicking
+    /// anywhere in the window (that doesn't have some other widget).
+    ///
+    /// If closeAutomatically is set, the window will close when clicking off it,
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,193 @@
...
+const QString TEXT_TITLE("Reply Thread - @%1 in #%2");
+
+namespace chatterino {
+
+ReplyThreadPopup::ReplyThreadPopup(bool closeAutomatically, QWidget *parent,
+                                   Split *split)
+    : DraggablePopup(closeAutomatically, parent)
+    , split_(split)
+{
+    this->setWindowTitle("Reply Thread");
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+void ReplyThreadPopup::setThread(
+    const std::shared_ptr<const MessageThread> &thread)
+{
+    this->thread_ = thread;
+    this->ui_.replyInput->setThread(thread);
+    this->addMessagesFromThread();
+    this->updateInputUI();
+}
```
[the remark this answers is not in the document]
dnsge: Updated in e0cabce

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -405,6 +440,36 @@ void TwitchChannel::sendMessage(const QString &message)
...
+    // Do last message processing
+    QString parsedMessage = this->prepareMessage(app, message);
+
+    bool messageSent = false;
+    this->sendReplySignal.invoke(this->getName(), parsedMessage, replyId,
+                                 messageSent);
+
+    if (messageSent)
+    {
+        qCDebug(chatterinoTwitch) << "sent";
```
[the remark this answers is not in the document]
dnsge: Do you mean just the `qCDebug` call? I had kept it because we do the same logging in `TwitchChannel::sendMessage`, but we could remove both if we don't think we need it.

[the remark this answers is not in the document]
leon-richardt: Yes, I was referring to the debug output. Fair point regarding the use in `::sendMessage()`, I wasn't aware when making the comment. I think it could be removed in both places then but I would like to hear other opinions on that as well.

src/messages/layouts/MessageLayoutElement.hpp
```diff
@@ -93,6 +94,23 @@ class ImageWithBackgroundLayoutElement : public ImageLayoutElement
...
+                                           const QSize &imageSize, QColor color,
+                                           int padding);
+
+protected:
+    void paint(QPainter &painter) override;
+
+private:
+    QColor color_;
+    QSize imageSize_;
+    int padding_;
```
23rd, line 111: Guess can be safely const.
```suggestion
    const QColor color_;
    const QSize imageSize_;
    const int padding_;
```

src/widgets/DraggablePopup.cpp
```diff
@@ -0,0 +1,90 @@
...
+        BaseWindow::EnableCustomFrame, BaseWindow::Frameless,
+        BaseWindow::FramelessDraggable};
+#endif
+
+}  // namespace
+
+DraggablePopup::DraggablePopup(bool closeAutomatically, QWidget *parent)
+    : BaseWindow(closeAutomatically ? popupFlagsCloseAutomatically : popupFlags,
+                 parent)
+    , lifetimeHack_(new bool)
```
23rd, line 28: ?
```suggestion
    , lifetimeHack_(std::make_shared<bool>(false))
```

src/messages/layouts/MessageLayoutElement.hpp
dnsge: Updated in c3af3fe5

src/widgets/DraggablePopup.cpp
dnsge: Updated in c3af3fe5

Felanbird (approved):
Works and looks exactly as I'd expect it to.

Felanbird:
random note: I noticed because of this PR that the `Sending message as felanbird...` text was not centered, this PR happens to center it which will be slightly annoying at first for people who do notice it, but is likely correct.

src/widgets/splits/ReplyInput.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "Application.hpp"
+#include "controllers/commands/CommandController.hpp"
+#include "controllers/hotkeys/HotkeyController.hpp"
+#include "messages/MessageThread.hpp"
+#include "widgets/helper/ResizingTextEdit.hpp"
+#include "widgets/splits/Split.hpp"
+
+namespace chatterino {
+
+ReplyInput::ReplyInput(QWidget *parent, Split *split)
```
Mm2PL, line 12: Should this really be a subclass? It already uses functionality from SplitInput, it has a Split parent, just like SplitInput. There already is logic for replies in SplitInput itself. Why is it a subclass in the first place?

Mm2PL (changes requested):
For me another consideration would be the close-when-lost-focus setting. I think it should be either off by default or default to the value set in the usercard setting.

There is also the tab-completion bug: for me if I tab-complete an emote in the reply popup, it unfocuses the input. This used to be a rare bug in normal splits, but I think it was fixed. And then there is the user-card dislocation bug on i3 where the reply and user card popups move to the top left of the screen.
I also have some comments/questions about the code:

src/widgets/splits/ReplyInput.cpp
dnsge: You're right, it doesn't really make sense now. I had initially subclassed the SplitInput to keep the logic separate but by adding inline replying at some point I introduced reply logic to SplitInput anyways.

src/widgets/splits/SplitInput.cpp
```diff
@@ -217,6 +289,296 @@ void SplitInput::openEmotePopup()
...
+
+QString SplitInput::hotkeySelectWord()
+{
+    auto cursor = this->ui_.textEdit->textCursor();
+    cursor.select(QTextCursor::WordUnderCursor);
+    this->ui_.textEdit->setTextCursor(cursor);
+    return "";
+}
+
```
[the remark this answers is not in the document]
dnsge: Mostly reverted in 841b3d2 (kept send message logic out in function because it's long)

src/messages/MessageElement.cpp
```diff
@@ -502,4 +661,36 @@ void ScalingImageElement::addToContainer(MessageLayoutContainer &container,
...
+    , size_(size, size)
+{
+}
+
+ReplyCurveElement::ReplyCurveElement(int neededMargin, int width, int height)
+    : MessageElement(MessageElementFlag::RepliedMessage)
+    , neededMargin_(neededMargin)
+    , size_(width, height)
+{
+}
```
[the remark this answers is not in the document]
dnsge: Good point 👍 becadbd

dnsge:
> random note: I noticed because of this PR that the `Sending message as felanbird...` text was not centered, this PR happens to center it which will be slightly annoying at first for people who do notice it, but is likely correct.

This was actually a bug that I did not notice... The VBox that contains the "Replying to" label was still taking up space while its children were not visible. I've now fixed that problem so the message input box looks the same as master when not replying.

pajlada:
This is the patch required to fix the tab/autocomplete focus lose issue:
```patch
diff --git a/src/widgets/helper/ChannelView.cpp b/src/widgets/helper/ChannelView.cpp
index 0b3459b3..09696c5b 100644
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -163,7 +163,7 @@ ChannelView::ChannelView(BaseWidget *parent, Split *split, Context context)
     // and tabbing to it from another widget. I don't currently know
     // of any place where you can, or where it would make sense,
     // to tab to a ChannelVieChannelView
-    this->setFocusPolicy(Qt::FocusPolicy::StrongFocus);
+    this->setFocusPolicy(Qt::FocusPolicy::ClickFocus);
 }
 
 void ChannelView::initializeLayout()
```

I touched this place very recently and felt like something was wrong in the Focus Policy that was set for ChannelView. Would need to be tested a bit extra most likely but worked as expected for me on Arch Linux btw

src/singletons/Settings.hpp
```diff
@@ -158,6 +159,8 @@ class Settings : public ABSettings, public ConcurrentSettings
     FloatSetting mouseScrollMultiplier = {"/behaviour/mouseScrollMultiplier",
                                           1.0};
     BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", true};
+    BoolSetting autoCloseThreadPopup = {"/behaviour/autoCloseThreadPopup",
+                                        true};
```
Felanbird, line 163:
```suggestion
                                        false};
```

Felanbird:
Tab focus issue is resolved with the `ClickFocus` patch, focus setting should still be disabled by default as it does not make sense for an input box to be removed on loss of focus, as well as it's impossible to use the emote menu with that setting enabled.

src/singletons/Settings.hpp
dnsge: Done 👍

Felanbird (approved):

pajlada (approved):
Awesome work!

Mm2PL (approved):
Found only a tiny visual bug, I'm fine with merging as is and fixing this at a later date.
With closing reply threads on lost focus:


Without:

dnsge:
> Found only a tiny visual bug

This isn't actually a visual bug but an intentional change. When close-on focus loss is true, we use a frameless window. On non-linux platforms, we won't have any area to drag the popup around because the ChannelView and SplitInput take up the entire popup window. So instead, we introduce a margin to allow the user to drag the popup around.

https://github.com/dnsge/chatterino2/blob/2ea0d8c7c92836c84d363bc75275b647f4e8f7ab/src/widgets/dialogs/ReplyThreadPopup.cpp#L92-L93

We could guard this behavior in a macro because we don't do frameless windows on linux: 

https://github.com/dnsge/chatterino2/blob/2ea0d8c7c92836c84d363bc75275b647f4e8f7ab/src/widgets/DraggablePopup.cpp#L11-L21

But like you said, we can do this at a later time. With the amount of code that was touched in this change, I'm sure we'll find more issues to fix once this change is on nightly.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -48,6 +48,66 @@ const QSet<QString> zeroWidthEmotes{
...
+            default:
+            case UsernameDisplayMode::UsernameAndLocalizedName: {
+                if (hasLocalizedName)
+                {
+                    usernameText = username + "(" + localizedName + ")";
+                }
+                else
+                {
+                    usernameText = username;
+                }
```
23rd, line 91: Can be simplified with a ternary operator.

src/widgets/splits/SplitInput.hpp
```diff
@@ -109,6 +137,7 @@ private slots:
     void editTextChanged();
 
     friend class Split;
+    friend class ReplyThreadPopup;
```
23rd, line 140: Is it needed here?

src/providers/twitch/TwitchMessageBuilder.cpp
pajlada: 1) The code is just copy-pasted from another function since it needed to be reused, so not super in scope of this PR to fix it
2) I would argue that using ternary operators here wouldn't necessarily make the code more readable

src/widgets/splits/SplitInput.hpp
pajlada: This is necessary for `ReplyThreadPopup` to be able to call `giveFocus` on the SplitInput

src/common/Channel.cpp
```diff
@@ -278,6 +278,13 @@ bool Channel::canSendMessage() const
     return false;
 }
 
+bool Channel::isWritable() const
+{
+    using Type = Channel::Type;
+    auto type = this->getType();
+    return type != Type::TwitchMentions && type != Type::TwitchLive;
```
zneix, line 285: Why `TwitchMentions` in specific? What about `TwitchLive` or TwitchWhispers`? seems like some weird "magic" comparison to make.

zneix (changes requested):

src/common/Channel.cpp
leon-richardt: You can use the split input in the whispers split for `/r`/`/w` usage) as well as in the `TwitchWatching` channel

zneix:
thanks for caring about me at all

PR #3727, opened 2022-05-13
Added `/debug-env` command

zneix:
Pull request checklist:

# Description

Purely for debugging-related addition, similar to `/debug-args`.
This prints multiple messages, showing values of Env::instance's members. I wanted to make it print one message with values separated / joined by a newline, but that didn't work, so I decided to just print multiple messages.

Example:
output of /debug-args command

Since this is debug information with placeholder values, messages printed by /debug-env don't parse links.

P.S. I don't really like this code but I decided to push this anyway. Maybe someone has a better idea on how to make this.

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -455,7 +455,11 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         for (QString &str : debugMessages)
         {
-            channel->addMessage(makeSystemMessage(str, false));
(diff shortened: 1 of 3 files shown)


Mm2PL (approved):
Useful for debugging unexpected behavior of the environment. Can for example check if custom rm host works :) One flaw of this kind of command is that it needs to be updated when chatterino::Env is updated, but I don't think it's that bad and won't cause much issues even if not tended to.

PR #3730, opened 2022-05-14
Fix: ignore whitespaces pasted in EmotePopup search

kornes:
Pull request checklist:

# Description

QValidator used to check user's input will fully reject string from being pasted if one character is incorrect. Leading or trailing whitespaces are common offenders in copied text.

EmotePopup search and settings Live Notifications (RegExpItemDelegate) are currently affected by this resulting in poor UX.

This PR adds trimmed RegExp validator fixing above problem.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Minor: Fixed tag parsing for consecutive escaped characters. (#3711)
 - Minor: Prevent user from entering incorrect characters in Live Notifications channels list. (#3715)
 - Minor: Fixed automod caught message notice appearing twice for mods. (#3717)
+- Bugfix: Pasting to search in emote popup window will now ignore whitespaces. (#3730)
```
Felanbird, line 17:
```suggestion
- Minor: Prevent user from entering incorrect characters in Live Notifications channels list. (#3715, #3730)
- Minor: Fixed automod caught message notice appearing twice for mods. (#3717)
```
I'm not sure if this needs a separate changelog entry, while it does add an extremely useful functionality, it could be covered under the original(?)

Felanbird (approved):

All incorrect characters still cannot be inputted, along with space when it's in any position except trailing in a copy+paste.

src/widgets/helper/RegExpItemDelegate.cpp
```diff
@@ -1,4 +1,5 @@
 #include "widgets/helper/RegExpItemDelegate.hpp"
+#include "widgets/helper/TrimRegExpValidator.hpp"
```
pajlada, line 2:
```suggestion

#include "widgets/helper/TrimRegExpValidator.hpp"
```

PR #3735, opened 2022-05-14
Store serverReceivedTime in messages

pajlada:
Pull request checklist:

# Description

- feat(message): expose time via getter
- refactor(message): store time as millis with IRC conversion
- Make last changes from #3694

This PR is meant as a preamble to #3694 to make that PR contain less fluff


jammehcow (approved):
Haven't tested, but looks as I remember.

Felanbird (approved):
Seems fine - Worst case scenario we find an issue with the time changes even sooner.

PR #3740, opened 2022-05-16
Added more streaming software to streamer mode binaries list

badoge:
Pull request checklist:

- [x] update linux/macos binaries list

# Description

Added a couple binaries to the list.
Also renamed "DetectObs" to "DetectStreamingSoftware" since it will be detecting more than OBS now.

I searched for lists of streaming software and these seem to be the most popular ones.
Streamlabs Desktop on windows uses obs.exe but on macOS it needed to be added on its own.
PRISMLiveStudio is an obs fork but it does not use obs.exe like streamlabs.
XSplit and vMix are windows only and don't have mac/linux versions.


src/util/StreamerMode.hpp
```diff
@@ -4,7 +4,11 @@
 
 namespace chatterino {
 
-enum StreamerModeSetting { Disabled = 0, Enabled = 1, DetectObs = 2 };
+enum StreamerModeSetting {
+    Disabled = 0,
+    Enabled = 1,
+    DetectStreamingSoftware = 2
```
zneix, line 10: Using this enum literal will make git diffs of any potential future changes cleaner
```suggestion
    DetectStreamingSoftware = 2,
```

pajlada (approved):

PR #2905, opened 2021-06-20
Fix emotesets not loading properly

zneix:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

Initial work. It seems to work, but please do not make any reviews yet. WIP

Fixes #2904

PR #3743, opened 2022-05-16
Remove unused Env member for emoteset resolver url

zneix:
Pull request checklist:

# Description

This is a follow-up to GH-2905, this code should've been removed back then.

I forgot how to use git when I made GH-3742


Felanbird (approved):
sure why not - never heard of this env in my life

PR #3745, opened 2022-05-16
Added notes to usercards

### LINKED ISSUE #2413: Add user 'notes' to usercards
**What should be added?**
An extension to usercards, which could allow users to make their own notes on someone. Could be done in a very similar way of how it is done on Discord:
IF YOU READ THIS I VON

**Why should it be added?**
Many users, including me, could see it as a new cool feature, e.g. that would make it possible to remember more and more people, by adding little notes about where we met them or who they are.

**Additional Notes**
I am not sure whether we should use username-based or twitchid-based notes

MrAuro:
Pull request checklist:

# Description

Allows users to set notes for other people. The note is stored alongside the user's ID so a username change wont affect them. All clientside and doesn't have to do with mod comments. Works kind of like the discord notes. would close #2413

**Functionality is all there, just need to make the UI look better**

Example:

https://user-images.githubusercontent.com/35087590/168693634-fdd9c20d-a54a-4eee-b8ab-9f2067b0eb3c.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added usercard notes. (#3745)
 - Minor: Added `is:first-msg` search option. (#3700)
 - Minor: Added quotation marks in the permitted/blocked Automod messages for clarity. (#3654)
 - Minor: Adjust large stream thumbnail to 16:9 (#3655)
--- a/chatterino.pro
+++ b/chatterino.pro
@@ -169,6 +169,7 @@ SOURCES += \
     src/controllers/moderationactions/ModerationAction.cpp \
     src/controllers/moderationactions/ModerationActionModel.cpp \
     src/controllers/nicknames/NicknamesModel.cpp \
+    src/controllers/notes/NoteModel.cpp \
     src/controllers/notifications/NotificationController.cpp \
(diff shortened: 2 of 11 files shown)


xHeaveny (commented):
I would make the place for the notes a lot smaller because now  its j8ust too big in my opinion

MrAuro:
> I would make the place for the notes a lot smaller because now its j8ust too big in my opinion

Wanted to get something down first. Didn't know how the UI should be when making this PR. I'll work on kornes' suggestions later today/tomorrow

Felanbird:
I'm wondering if a better approach is displaying the note, and using a small edit button to bring up a prompt to set the note. It seems very correct to just have a seperate window for setting the note. 

similar to the PR adding a reply window
c2_3745

Mm2PL:
> using a small edit button to bring up a prompt to set the note

I'd dislike having a popup window spawning a popup window. I feel like that might get confusing. pajaDent

But a green or blue paint on the bikeshed doesn't really matter when there is no bikeshed build yet.

pajlada:
- [ ] Use a separate file for this rather than saving it in `settings.json`
- [ ] Improve UI (see Discord for inspiration)

MrAuro:
Lost interest in this, if anyone wants to pick this up go ahead.

PR #3749, opened 2022-05-17
Updated link to VCRedist in README

zneix:
Pull request checklist:

# Description

As per fabZeef's experience after Windows reinstall, both .exe's from links from current README didn't work and he had to install the 'all' version. I think it's better. The new permalink link points to "VCRedist 2015-2022", which ensures you get everything you might need.
The new permalink is taken from Micro$oft's downloads page: https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170

Left out link to 2013 version just in case, but maybe it's old enough we can remove it as well(?).
I considered replacing it with what's on downloads page (https://aka.ms/highdpimfc2013x64enu), but it seemed to provide an older version, so maybe it's better to not change that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Windows users can also install Chatterino [from Chocolatey](https://chocolatey.o
 
 You can download the latest Chatterino 2 build over [here](https://github.com/Chatterino/chatterino2/releases/tag/nightly-build)
 
-You might also need to install the [VC++ 2017 Redistributable](https://aka.ms/vs/15/release/vc_redist.x64.exe) from Microsoft if you do not have it installed already.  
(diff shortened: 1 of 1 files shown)


zneix:
I'm 90% positive this is a good change to be made, but I'd need an eye of some windows expert on this just in case.

I know fourtf ships Windows installer with VCRedist 2015 (or maybe 2017) and changing what we recommend here might have some impact on this (version difference between what's shipped with the installer).

kornes:
CI uses vc toolset 143 for windows builds, so user needs VCRedist 2022 or newer (assuming whats distributed via website comes from CI).
We dont need anything older, so 2013 link can be dropped - each new redist release is backward compatible too, so user only needs newest one.

pajlada (approved):

PR #3758, opened 2022-05-22
Make an IApplication interface

pajlada:
Pull request checklist:

# Description

Function is `getIApp` for now
This should be preferred where possible for new code to make mocking possible in the future


pajlada (commented):
Another great PR adding no actual functionality
image

PR #3760, opened 2022-05-22
Build GoogleTest & GoogleMock as submodules

pajlada:
Pull request checklist:

# Description

We need GMock which is part of the same library, but there are no premade CMake constructs to include it unless we require CMake 3.20+, which we realistically cannot do.
Because of this, I've added the `googletest` submodule which contains both projects, and a few lines for building it if `BUILD_TESTS` is checked.


pajlada:
Looks good to me!
image

PR #3766, opened 2022-05-23
Added opening a user's chat in a new tab/split/window with keybinds+click

Infinitay:

The following are the added keybinds for the modifiers required to be held when left clicking a username

<table>
<tr>
	<td>Modifier
	<td>Result
<tr>
	<td>SHIFT
	<td>Open chat in a new popup
<tr>
	<td>CTRL
	<td>Open chat in a new tab
<tr>
	<td>ALT
	<td>Open chat in a new split
</table>

Originally was for #2881 but after implementing this and re-reading the request I now realize that #3625 is what the user wanted. 

image

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -742,15 +742,7 @@ void CommandController::initialize(Settings &, Paths &paths)
             return "";
         }
 
-        auto *app = getApp();
-        Window &window = app->windows->createWindow(WindowType::Popup, true);
-
-        auto *split = new Split(static_cast<SplitContainer *>(
-            window.getNotebook().getOrAddSelectedPage()));
-
-        split->setChannel(app->twitch->getOrAddChannel(target));
-
-        window.getNotebook().getOrAddSelectedPage()->appendSplit(split);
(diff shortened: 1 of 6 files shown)


kornes (commented):
sidenote, not a fan of hardcoding such modifiers, because majority of ppl won't know these exist and there are quite a few already. Maybe adding cheatsheet somewhere in app wouldnt be bad idea :D

Infinitay:
Implemented your suggestions. Thanks for the clarification earlier once again. If someone would like to make further changes such as by creating a proper copy constructor for Split that would be appreciated in replace of the _hacky_ solution in `WindowManager#createChannelWindow(Split*)`

> sidenote, not a fan of hardcoding such modifiers, because majority of ppl won't know these exist and there are quite a few already. Maybe adding cheatsheet somewhere in app wouldnt be bad idea :D

I agree with you, that's why before I made this PR I was looking into our existing hotkeys but noticed it doesn't (yet) support mouse click bindings.

As for a cheatsheet, I was considering adding an unmodifiable entry into existing Hotkeys but not sure if that's how we want to go about things.

PR #3767, opened 2022-05-23
Fixed channels not being set as offline

zneix:
Pull request checklist:

# Description

After change to the stream fetch in bulk, only channels present in `getHelix()->fetchStreams(...)` response have had their stream status updated (to live obviously). After channels would go offline, they'd not be marked as offline.

For this to work I had to add a finally callback to `getHelix()->fetchStreams(...)` to ensure channels get set as offline even if our Helix request has failed.

Iterating over all channels in the finally callback makes it so we don't have to invoke too many unnecessary iterations by using `getChannelOrEmptyByID(...)` which is a O(n) operation.

Fixes a degradation introduced in GH-3757 

Side note: Ideally though, there should most likely be a proper way to get a channel through its ID (maybe even through some local cache map that can give us a userLogin from a provided userID and just simply call `getChannelOrEmpty(...)`).


src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -356,6 +361,27 @@ void TwitchIrcServer::bulkRefreshLiveStatus()
             },
             []() {
                 // failure
+            },
+            [this, offlineChannelIDs] {
+                // All the channels that were not present in fetchStreams response are assumed to be offline
+                // It is necessary to update their stream status in case they've went live -> offline
+                // Otherwise some of them will be marked as live forever
+
+                for (const auto &weak : this->getChannels())
```
iProdigy, line 370: nice $ O(n) $ approach

iProdigy (approved):

xel86:
Feels weird giving feedback on a fix for a bug that I caused, but we can probably avoid making a finally callback and doing more operations. I have to go to work but I made a quick alternative that I think could work instead

```
    QStringList userIDs;
    QSet<TwitchChannel *> channel_set;
    this->forEachChannel([&userIDs, &channel_set](ChannelPtr chan) {
        auto twitchChan = dynamic_cast<TwitchChannel *>(chan.get());
        if (!twitchChan->roomId().isEmpty())
        {
            userIDs.push_back(twitchChan->roomId());
            channel_set.insert(twitchChan);
        }
    });

    for (const auto &batch : getChannelsInBatches(userIDs))
    {
        getHelix()->fetchStreams(
            batch, QStringList(),
            [this, &channel_set](std::vector<HelixStream> streams) {
                for (const auto &stream : streams)
                {
                    auto chan = this->getChannelOrEmpty(stream.userLogin);
                    if (chan->getType() != Channel::Type::Twitch)
                        continue;

                    auto twitchChan = dynamic_cast<TwitchChannel *>(chan.get());
                    twitchChan->parseLiveStatus(true, stream);
                    channel_set.remove(twitchChan);
                }

                for (const auto &unsetChan : channel_set)
                {
                    unsetChan->parseLiveStatus(false, {});
                }
            },
            []() {
                // failure
            });
    }
```

We would just make a set of the `TwitchChannel` pointers that we already had and remove each that we found in the /streams payload, then iterate all of the channels that weren't found and set them offline. Again, sorry for missing this the first time around!

pajlada:
> We would just make a set of the `TwitchChannel` pointers that we already had and remove each that we found in the /streams payload, then iterate all of the channels that weren't found and set them offline. Again, sorry for missing this the first time around!

This is the logic I've used for stream-online checking in other products before, and it's been solid

zneix:
> we can probably avoid making a finally callback and doing more operations.
> 
> We would just make a set of the `TwitchChannel` pointers that we already had and remove each that we found in the /streams payload, then iterate all of the channels that weren't found and set them offline.
> 
> Edit: If the call fails just set all of the pointers offline in the failure callback.

We came up with mm2pl to use a finally callback as it is meant for code that should be ran regardless of the network request's result. I don't think it's a big deal with adding it to the Helix's class methods - it needs its own rework later on anyway.

Based on your comment and some random thoughts I had while trying to come up with an answer to your comment (and pajlada's feedback in chat, thanks btw) I've come up with changes present in above commit (2cb8bd1), optimizing amount of `dynamic_cast`'s we perform, also removing temporary lists we previously relied on.
One thing I'm not really sure about myself is the usasge of `->keys()` below, thinking there might be an even better way to get values of the `QHash` somehow, but I'm not sure:
https://github.com/Chatterino/chatterino2/blob/2cb8bd11001ff877b29c3f046c388932bf694466/src/providers/twitch/TwitchIrcServer.cpp#L345
Please check it out and let me know if there's any further questions!

xel86:
Best solution by far, thanks for fixing this! 😅

pajlada (changes requested):
Right now if we have enough channels that we reach a batching request, we'll be firing the finally call twice handling in the same map - while eventually the results would be correct anyway, it feels unneccesarily dirty. We discussed a method in chat that makes `getChannelsInBatches` return a vector of QHashes with max 100 channels that point the room id to the channel pointer directly so each finally request can handle their own 100 channels instead

Mm2PL:
I'll hold off on reviewing today. I tested with [this](https://kotmisia.pl/haste/he82Bp8YZp0CE9J) window-layout.json and some channels' live status appeared and disappeared. I wish I knew what went wrong.

zneix:
Thank you pajlada for spotting and pointing out that unexpected behaviour.

As we've discussed the solution in chat yesterday, I've came up with it and pushed it to the [`zneix/fix/bulk-stream-fetch-set-offline-alt1`](https://github.com/Chatterino/chatterino2/commit/0e208662057058ca15c9f68fda0fcaf04f1fe8fb) branch.

However, just as I was about to push that change here I've realized there's an easier solution involving just limitting our scope of operations in the `finally` callback to the elements from the current batch while just accessing QHash's elements.
Because of that, I've decided to split this and above change to separate branches, to decide which one we want to go with; second change was pushed to the [`zneix/fix/bulk-stream-fetch-set-offline-alt2`](https://github.com/Chatterino/chatterino2/commit/1b839cf2143352c39bfb0a406dd0c9d75fb128d3) branch - see the exhaustive commit message explaining the reasoning.
As stated there, I think that creating subsets of QHash'es is unnecessary while we already have a pointer containing all channels and if we can limit ourselves to operate just on channels for which we've fetched the live status (which is possible and exactly what fix on `...-alt2` branch does) which could turn out to be less complex and/or faster.

Also big thanks to mm2pl for spending his time and providing the accurate testing layout for this PR.
I've used it and after throughout testing I confirm that behaviour is 100% identical - it's just a matter of how we access channels in the `finally` callback.

Please have a look at both solutions and let me know which one seems easier - while we've agreed yesterday of going with `...-alt1`, I'm personally leaning towards `...-alt2`.

kornes:
2 more things after checking all versions:
I think using finally here is unnesesery overhead; extra coping and thread switching for no profit whatsoever. I would do said cleaning after NetworkRequest returns.

and lastly, getChannelsInBatches could be replaced with generic global util function (or template), taking iterable (twitchChans here), chunkSize and callback fn to execute on each chunk.

zneix:
> and lastly, getChannelsInBatches could be replaced with generic global util function (or template), taking iterable (twitchChans here), chunkSize and callback fn to execute on each chunk.

This was already discussed in #3757 and I've also addressed that in [`zneix/fix/bulk-stream-fetch-set-offline-alt2`](https://github.com/Chatterino/chatterino2/commit/1b839cf2143352c39bfb0a406dd0c9d75fb128d3).. In short: something to deal with in the future, since it's already copied from another place in the code, but it'd be unrelated to modify other code while only working on channel stream status.

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -332,30 +332,49 @@ namespace {
 
 void TwitchIrcServer::bulkRefreshLiveStatus()
 {
-    QStringList userIDs;
-    this->forEachChannel([&userIDs](ChannelPtr chan) {
+    auto twitchChans = std::make_shared<QHash<QString, TwitchChannel *>>();
```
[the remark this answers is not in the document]
zneix: When I was writing code for this PR I've noticed that without making QHash a pointer, it was copied in onSuccess and finally callbacks, making `twitchChans->remove(...);` call do nothing since it operated on a copied value.

kornes:
shit my bad, was my first look on NetworkPrivate and didnt notice its all async, edited my comment

about batching function, touched on that topic only because i saw some edits on alt1 and alt2, i guess it wouldn't take lot of changes beyond your edits, but i understand you want to make it properly in future 👍

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -332,30 +332,49 @@ namespace {
...
+                    // remaining channels will be used later to set their stream status as offline
+                    twitchChans->remove(stream.userId);
                 }
             },
             []() {
                 // failure
+            },
+            [this, twitchChans] {
+                // All the channels that were not present in fetchStreams response should be assumed to be offline
+                // It is necessary to update their stream status in case they've went live -> offline
```
Felanbird, line 367:
```suggestion
                // It is necessary to update their stream status in case they've gone live -> offline
```
minor english xd - `they went live` is also acceptable

zneix: Sure, titlechange_bot always posts `forsen has gone live`, not `forsen has went live` so I like this.

pajlada (approved):
:+1: Looks good to me! @Mm2PL once you've tested/approved this feel free to merge it in!

kornes:
tested c9a7e4d47b036f3d44597f3dbdaa3a04c1e9c850 for 25mins, seems to works stable and as expected.

Mm2PL (approved):
Works as expected, after latest changes doesn't introduce noticeable instabilities

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -302,60 +331,55 @@ std::shared_ptr<Channel> TwitchIrcServer::getChannelOrEmptyByID(
...
-            [this](std::vector<HelixStream> streams) {
+            batch, {},
+            [twitchChans](std::vector<HelixStream> streams) {
                 for (const auto &stream : streams)
                 {
-                    auto chan = this->getChannelOrEmpty(stream.userLogin);
-                    if (chan->getType() != Channel::Type::Twitch)
+                    // remaining channels will be used later to set their stream status as offline
+                    // so we use take(id) to remove it
+                    auto tc = twitchChans->take(stream.userId);
```
iProdigy, line 356: Aside: if a channel changes their username after roomId has been set, and goes live, we could detect the new `userLogin` here & possibly join the new channel name

zneix: Good to keep in mind for the future :+1: although at the moment we have 0 logic handling renames so while this is potentially useful, it'll have to be deferred until unforeseen future.

PR #3771, opened 2022-05-26
Corrected the way we parse comma-separated "list tags" in PRIVMSGs

zneix:
Pull request checklist:

- [x] Add tests from:
https://paste.ivr.fi/raw/marehysiwo
https://paste.ivr.fi/raw/ocidamuxyq
https://paste.ivr.fi/raw/opeponetyg
https://paste.ivr.fi/raw/icejijiqaq

# Description

### Why was the logic change in parsing "badges" and "badge-info" made/needed:

tl;dr: we should split by slash only its occurrence instead of every occurrence.

It's been discovered that elements of the comma-separated list (taglist) in "badge-info" tag can have slashes in their values.
For example, "foo/bar/baz,tri/hard" can be a valid badge tag
In that case, valid map content should be:
	`{"foo": "bar/baz", "tri": "hard"}`
With old logic of just splitting elements of the taglist by (every) slash it would be instead:
	`{"foo": "bar" "baz", "tri": "hard"}`
Here, the "foo" element of the taglist would be considered invalid and skipped (because the amount of substrings after `QString::split` operation was not equal to 2).
Now, the lack of slash can indicate that the tag is invalid and should be skipped instead of checking for amount of substring being non-equal to 2.
And to ensure we parse taglist's elements properly, a [`QString::section`](https://doc.qt.io/qt-5/qstring.html#section) is used which can return a substring of elements before and after a specified occurrence of the delimeter, similarly to GNU Coreutils' cut command:
```sh
zneix@uds ~$ echo "foo/bar/baz" | cut -d/ -f1
foo
zneix@uds ~$ echo "foo/bar/baz" | cut -d/ -f2-
bar/baz
```

### Actual noticable differences

With old logic, some edge-cases were wrongly parsed, resulting in `badge-info` taglist values being skipped (fallback text is being used):
old behaviour

New logic (element of `badge-info` is parsed correctly and used in the tooltip):
new behaviour

### Details about codebase changes:

In the process a lot of duplicate code was removed to have the logic responsible for this in one place only. For that to work, some type changes were required.
~~I've also decided in this scenario it's better to use `std::unordered_map` over `std::map` - perhaps it's convinient to use `QHash` at this point, but I'm not sure so I just left it as is.~~
nevermind, map/hash can't be used for badges - that breaks the order in which they're displayed (which btw, should follow the order in which badges came in to us from twitch).
Changed map for `badge-info` to an `std::unordered_map` though which seemed appropriate.

~~Changing the default constructor for `Badge` class in TwitchBadges.cpp made things easier to handle, making `SharedMessageBuilder::parseTagList` work for parsing both `badge-info` and `badges` tags in PRIVMSG messages, without having to duplicate the same function just for the sake of returning `std::vector<Badge>` instead of `std::(unordered_)map<QString, QString>`. This comes with a little compromise in TwitchMessageBuilder.cpp, where a helper vector is declared and set as `this->message().badges` - perhaps this can be mitigated somehow.~~
this isn't true anymore - while retaining badge order, it's less complex to parse `badges` and `badge-info` separetly.

### DIFF
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -79,13 +79,13 @@ void SharedMessageBuilder::parse()
 // In that case, valid map content should be 'split by slash' only once:
 // {"foo": "bar/baz", "tri": "hard"}
 std::pair<QString, QString> SharedMessageBuilder::slashKeyValue(
-    const QString &keyValueString)
+    const QString &kvStr)
 {
     return {
         // part before first slash (index 0 of section)
-        keyValueString.section('/', 0, 0),
+        kvStr.section('/', 0, 0),
         // part after first slash (index 1 of section)
-        keyValueString.section('/', 1, -1),
+        kvStr.section('/', 1, -1),
     };
 }
 
@@ -106,7 +106,7 @@ std::vector<Badge> SharedMessageBuilder::parseBadgeTag(const QVariantMap &tags)
             continue;
         }
 
-        auto pair = this->slashKeyValue(badge);
+        auto pair = SharedMessageBuilder::slashKeyValue(badge);
         b.emplace_back(Badge{pair.first, pair.second});
     }
 
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -91,7 +91,7 @@ std::pair<QString, QString> SharedMessageBuilder::slashKeyValue(
 
 std::vector<Badge> SharedMessageBuilder::parseBadgeTag(const QVariantMap &tags)
 {
-    auto b = std::vector<Badge>();
+    std::vector<Badge> b;
 
     auto badgesIt = tags.constFind("badges");
     if (badgesIt == tags.end())
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -340,7 +340,7 @@ void SharedMessageBuilder::parseHighlights()
     bool badgeHighlightSet = false;
     for (const HighlightBadge &highlight : *badgeHighlights)
     {
-        for (const auto &badge : badges)
+        for (const Badge &badge : badges)
         {
             if (!highlight.isMatch(badge))
             {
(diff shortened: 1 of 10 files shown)


src/messages/SharedMessageBuilder.cpp
```diff
@@ -103,6 +75,44 @@ void SharedMessageBuilder::parse()
...
+    };
+}
+
+std::vector<Badge> SharedMessageBuilder::parseBadgeTag(const QVariantMap &tags)
+{
+    std::vector<Badge> b;
+
+    auto badgesIt = tags.constFind("badges");
+    if (badgesIt == tags.end())
+        return b;
```
pajlada, line 98: Braces around the if please

src/messages/SharedMessageBuilder.hpp
```diff
@@ -32,6 +33,11 @@ class SharedMessageBuilder : public MessageBuilder
     virtual void triggerHighlights();
     virtual MessagePtr build() = 0;
 
+    static std::pair<QString, QString> slashKeyValue(const QString &str);
```
pajlada, line 36: Mismatching definition
```suggestion
    static std::pair<QString, QString> slashKeyValue(const QString &keyValueString);
```

will most likely need a reformat after the suggestion too

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1033,17 +985,36 @@ boost::optional<EmotePtr> TwitchMessageBuilder::getTwitchBadge(
     return boost::none;
 }
 
+std::unordered_map<QString, QString> TwitchMessageBuilder::parseBadgeInfoTag(
```
pajlada, line 988: Make static

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1033,17 +985,36 @@ boost::optional<EmotePtr> TwitchMessageBuilder::getTwitchBadge(
...
         return;
     }
 
-    auto badgeInfos = parseBadgeInfos(this->tags);
-    auto badges = parseBadges(this->tags);
+    auto badgeInfos = this->parseBadgeInfoTag(this->tags);
+    auto badges = this->parseBadgeTag(this->tags);
 
-    for (const auto &badge : badges)
+    for (const Badge &badge : badges)
```
pajlada, line 1017:
```suggestion
    for (const auto &badge : badges)
```

pajlada (changes requested):
Nitpicks, rest of the code looks great. thanks for adding tests :rocket:

pajlada (approved):
Looks good to me :+1: Merge in whenever you want

PR #3775, opened 2022-05-28
Simplify code for fake debug messages

zneix:
Pull request checklist:

# Description

this PR has been split into separate ones.

Nvim suggested I mark some functions as inline


pajlada (changes requested):
Good changes! You making these changes makes me think we should probably keep all our sample messages in a single file, and this PR would be a good time to do so!

zneix:
> Good changes! You making these changes makes me think we should probably keep all our sample messages in a single file, and this PR would be a good time to do so!

Definitely agree on this, I've moved all the hardcoded debug messages to `SampleData.hpp` - it keeps things in one nice pile of junk.

pajlada (approved):

pajlada (commented):
actually missing SampleData.cpp in `chatterino.pro` and `src/CMakeLists.txt`

PR #3776, opened 2022-05-29
Removed unused files IsBigEndian.hpp and rangealgorithm.hpp

zneix:
Pull request checklist:

# Description

Cherry picking from #3775


pajlada (changes requested):
Update `chatterino.pro` and remove references to the two removed files

Note that they were never included in the `src/CMakeLists.txt` file, which is an indication that this PR is overdue :feelsgood:

pajlada (approved):

Issue #3768, opened 2022-05-25
Odd error message when prompted for an update.


aknight2015:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

So, I get the "update" arrow next to the gear icon. I click it and get what's displayed in the screenshot. I'm using the most recent nightly build. I'm running on Linux by the way. Debian. I do wanna say that Chatterino has replaced all other Twitch chat applications. I hope you continue this and eventually make a better Twitch stream viewer.

![Screenshot 2022-05-24 8 06 00 PM](https://user-images.githubusercontent.com/47067592/170151761-45864e16-10fd-4d4b-a550-e07efe3d16e6.png)

### OS and Chatterino Version

Chatterino 2.3.5 (commit 81a62764c)

zneix:
To be honest I wouldn't label this as a bug really, it seems to be rather intended(?)
More context, taken from [Chatterino Discord Server](https://discord.gg/qq7DDxjste):
![no updares on beta](https://cdn.zneix.eu/6B6SJR6.png)

Mm2PL:
Can we please switch to github releases api from notitia? It is easier to just use a single release system

PR #3779, opened 2022-05-29
Use GitHub releases for automatic updates

dnsge:
This PR is a WIP/experiment to see if using GitHub releases for directly delivering updates would be preferable to the current solution. Please voice your opinions!

Pull request checklist:

# Description
1. Migrates installing updates from Notitia server to directly using GitHub releases. 
2. Adds automatic download and installation of Chatterino on macOS **(all two macOS users rejoice)**  
3. Adds the `CHATTERINO2_FORCE_UPDATE=true` environment variable to make testing the installation process easier. 

~Note: I have not tested this at all on Windows or Linux. It probably doesn't work on Windows.~ I have now tested this on Windows and got it to a satisfactory place. Still haven't done anything on Linux.

Update Process:
- Checks `latest` release
- Sees if tag version differs
- If so, looks through assets to find correct asset to download
- Follows almost exact same process from before for downloading and installing the update

TODO:
- [ ] Support nightly installations?
- [ ] Beta support? Do we still want betas if we can auto-update to nightly?

Inspired by this comment by Mm2PL https://github.com/Chatterino/chatterino2/issues/3768#issuecomment-1136992843


src/widgets/dialogs/UpdateDialog.cpp
```diff
@@ -27,7 +29,6 @@ UpdateDialog::UpdateDialog()
 
     QObject::connect(install, &QPushButton::clicked, this, [this] {
         Updates::instance().installUpdates();
-        this->close();
```
dnsge, line 30: `this->close()` would segfault on my machine, probably because the `installUpdates()` makes the Dialog go out of focus, destroying it and leading to a use-after-free. Removing this line doesn't seem to cause any issues, at least on macOS.

dnsge:
I finally managed to get a working Windows environment and was able to test out this PR and fix a couple of issues that I found, mostly around not being able to automatically launch the installer sometimes.

RapidHoglet:
I am for an option to switch to a Nightly Build. I saw that at Multimc [Wiki](https://github.com/MultiMC/Launcher/wiki/Switching-update-channels). There you can switch between Stable and Development Build. As soon as you press the update button, the changelog is displayed with the changes between the installed and the last version.
You can then decide whether the update should be carried out or not.
Implementing that in the same way or slightly differently would be great.

Felanbird:
> I am for an option to switch to a Nightly Build. I saw that at Multimc [Wiki](https://github.com/MultiMC/Launcher/wiki/Switching-update-channels). There you can switch between Stable and Development Build. As soon as you press the update button, the changelog is displayed with the changes between the installed and the last version. You can then decide whether the update should be carried out or not. Implementing that in the same way or slightly differently would be great.

image

RapidHoglet:
I only describe how it could be implemented and try to bring in my opinion/ideas. 
Maybe my wording is wrong or incomprehensible. 
Sorry for that. Have a great day.

fourtf:
I like the idea. However, I'm against using the GitHub API directly from the client. Adding an API to notitia to potentially be triggered by CI seems better to me. Perhaps some other release channels (e.g. weekly or monthly) could be added instead of beta.

dnsge:
Yeah, I would agree with that. Working with the GitHub API is a bit painful for things like getting nightly releases and figuring out whether the local version is actually behind the online version. And like you say, we'd have a lot more freedom with integrating with our CI scripts.

jammehcow:
Is there potentially a better solution out there for update notification/delivery that we can integrate with? This is a problem that hundreds of apps face and surely there's something already in existence that we can use. 

I'll have a look around to see if I can find anything.

PR #2139, opened 2020-10-26
Add Mm2PL's avatar to contributors list.

Mm2PL:
Pull request checklist:

- ~~[ ] `CHANGELOG.md` was updated, if applicable~~ N/A

very simple PR, title
See #2126

PR #3781, opened 2022-05-29
Organized version information

zneix:
Pull request checklist:

# Description

#### Organized version information shown in About page:
- moved Qt version information to the top (main motivation)
- organized and split compile-time and runtime information since it was getting a bit cluttered in there; diff there is stupid, just read new code
- removed ancient irrelevant comments
- added new `bool isModified()` member to `Version` class - used to determine whether the vcs tree was compiled or not at the time of building the app. Idea adopted from [Golang's source code](https://github.com/golang/go/blob/016d7552138077741a9c3fdadc73c0179f5d3ff7/src/cmd/go/internal/vcs/vcs.go#L303):

#### Added information about running in DEBUG mode to fullVersion:
At first, I wanted to only add this to AboutPage.cpp but I figured it might be good to be consistent in other places (result of `--version` and/or title window); however if this change is "too invasive" I will revert it and add this information to AboutPage.cpp only.

#### Previews:

built in debug mode with modified tree + no nightly:
debug modified no nightly

built in non-debug (release) mode with clean tree + nightly:
release clean nightly


kornes:
while on about page topic, maybe we could move contributors list to chatterino api (since it needs to be curated and we cant use github directly)? so there wont be need to bundle 50 avatars and txt file.

zneix:
> 

Firstly, I'm focusing on version-related changes here (not just in About page - it's not about reworking that page but reworking logic around version information).
Secondly, No. There is a good reason we include authors' information locally, but I won't go over that again here since this was already raised in https://github.com/Chatterino/chatterino2/pull/2139#issuecomment-716670625

src/common/Version.hpp
```diff
@@ -26,6 +26,7 @@ class Version
 
     const QString &version() const;
     const QString &commitHash() const;
+    const bool &isModified() const;
```
pajlada, line 29: Move the comment from `Version.cpp` to here instead, that way it's accessible whenever `isModified` is autocompleted

pajlada (changes requested):
Looks good! At some point we should just make two functions, one that generates "build information" and one that generates "runtime information" so diffs looks nicer in the future

zneix:
> The new wrapped around line of information cannot be copied.

Thanks for catching this!

Before my changes, the structure was something like:
`parent layout` -> `QGroupBox` -> `QLabel` with version
But after I wanted to add a line break in between I turned it to something like:
`parent layout` -> `QGroupBox` -> `QVBoxLayout` -> (`QLabel` , `QLabel`)
Even though `QVBoxLayout` renders its labels vertically, they're in fact completely "isolated" so that it's impossible to select text from multiple `QLabel`'s at once, even with right mouse interaction flags being set.

I've tried to find a reasonable solution to this and getting rid of second `QLabel` + putting a simple `<br>` between lines seems to be the only non-overengineered quick solution.
One thing I'm not so sure about this is that with only one `QLabel`, the `QVBoxLayout` is technically useless, however I've noticed that its presence makes some nice "padding" that pther `QGroupBox`'es in About page use, which makes Version's `QGroupBox` a bit more inline with the other ones (they all have a `QVBoxLayout` holding all the elements), look at the padding:
ocd satisfied
Like I said, even though `QVBoxLayout` is not necessary, the lack of it makes things look a bit more 'ugly':
reeeee

So I'm thinking I'll just keep the `QVBoxLayout` here for the sake of being consistent with the paddings, even if it's going to hold only 1 element. Such a minor thing and so much writeup... but alas I've chosen the modification visible on the 1st screenshot in 4a60365d1d14c2f5c0b5dda03e67a2004037c06b.

zneix:
I decided it'll be cleaner to determine the boolean at qmake/cmake step already, which should also take care of qmake compilation issues. While doing this I've decided it should be impossible to override the value of GIT_MODIFIED since then it'd be pointless to have this "variable" in the first place. And there's no need for any fallback through env variables since if git command fails, `Version::isModified()` would be set to false by default.

cmake/GIT.cmake
```diff
@@ -62,4 +77,4 @@ if (DEFINED ENV{GIT_RELEASE})
     set(GIT_RELEASE "$ENV{GIT_RELEASE}")
 endif ()
 
-message(STATUS "Injected git values: ${GIT_COMMIT} (${GIT_RELEASE}) ${GIT_HASH}")
+message(STATUS "Injected git values: ${GIT_COMMIT} (${GIT_RELEASE}) ${GIT_HASH} ${GIT_MODIFIED}")
```
Mm2PL, line 80: This `${GIT_MODIFIED}` returns `0` or `1`, i see this being confusing, maybe adding a string saying `modified: 0/1` or removing this value completely would help.

Mm2PL (changes requested):
nice PR, just small things

cmake/GIT.cmake
zneix: You mean just changing the output that's printed, right? Not changing the value of the variable itself. I figured that because there's a space between these it would be ok if you know what you're reading, but I suppose it's better to be more descriptive.
As per another option which'd be removal... not sure, I think it might be more profitable to keep this in case of any potential debugging in the future, so I guess I'll just add a little note like you proposed.

pajlada:
image
I removed a `;` before the `built` part
I uppercased the `running` string

Both strings are now independently generated once on startup by the Version class, and returned with `buildString()` and `runningString()`

pajlada (approved):
I'm happy with the code as it looks now, feel free to take a polishing step on it if you'd like, or come with feedback!

Mm2PL (approved):

PR #3782, opened 2022-05-29
Announcement message order & announcement colors

neil-lobo:
#Pull request checklist:

# Description
image

1. Changed the order of message parts being built and added to the channel. Originally, the "Announcement" message was added 2nd, which does not reflect what occurs in native chat.
image

2. Added the different announcement colors (blue, green, and orange) as well as new `MessageFlag`s
   - All announcement messages are given an `Announcement` flag as well as an `Announce[color]` flag


src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -707,12 +723,29 @@ void IrcMessageHandler::handleUserNoticeMessage(Communi::IrcMessage *message,
         else if (msgType == "announcement")
         {
             messageText = "Announcement";
+            QString announceCol = tags.value("msg-param-color").toString();
+            if (announceCol == "PRIMARY" || announceCol == "PURPLE" || announceCol == "")
+            {
+                announceFlag = MessageFlag::AnnouncePurple;
+            }
```
zneix, line 730: Keep in mind that `msg-param-color` tag is not present at all times - invoking `.announce foo` from IRC completely skips it.
Perhaps you should try to use an iterator/check if the tag exists in the message before trying to 'stringify' it.
Nitpick, but why not name `announceCol` just `announceColor`.

I also don't like how handling of "PRIMARY" and empty string is packed together with the same condition with "PURPLE" - you shouldn't assume that purple will always be primary colour. First two mentioned by me cases should be treated as sort of 'fallback' in my opinion.

As something that's not necessarily up to you to figure out (but would be nice) is that this whole if-else chain could probably make use of newly added magic_enum dependency - and

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -707,12 +723,29 @@ void IrcMessageHandler::handleUserNoticeMessage(Communi::IrcMessage *message,
...
+            {
+                announceFlag = MessageFlag::AnnounceOrange;
+            }
         }
 
         auto b = MessageBuilder(systemMessage, parseTagString(messageText),
                                 calculateMessageTime(message).time());
 
-        b->flags.set(MessageFlag::Subscription);
+        b->flags.set(announceFlag);
```
zneix, line 748: I believe that `MessageFlag::Subscription` is responsible for some other under-the-hood things like categorizing messages in search which perhaps shouldn't be touched.

zneix (changes requested):
As mentioned in one of my comments, colors in announcement USERNOTICEs are pretty much a webchat-only functionality - Twitch doesn't mention anything regarding how we should treat colours. My little concern here is that the way they handle this might be a subject to change.

As per changing the order of incoming messages...
The same behaviour of messages coming in order `user msg -> system msg` is for all of the USERNOTICEs in Chatterino. If the motivation for changing the order of these is just so that native chat does this for announcement messages then I'd be fine with it if that was announcement-specific. As you mention: `Originally, the "Announcement" message was added 2nd` - yes, but this is also a case for all system messages in this kind of messages. I'll admit with shame that we've always had this sort of quirk in Chatterino and it's something that should be taken care of on its own but not here, so I'd ask you to revert that.
On a different note, I've actually started working on this a couple days ago on a local branch; reworking both `IrcMessageHandler::handleUserNoticeMessage` and `IrcMessageHandler::parseUserNoticeMessage` so that they'll construct and add only 1 message with system message at the top, linebreak and then the user's message. Now I don't want to sound rude but your changes would only make that a bit more messy in general.

src/providers/colors/ColorProvider.hpp
```diff
@@ -13,6 +13,10 @@ enum class ColorType {
     Whisper,
     RedeemedHighlight,
     FirstMessageHighlight,
+    AnnouncePurple,
+    AnnounceBlue,
+    AnnounceGreen,
+    AnnounceOrange
```
zneix, line 19: Notice that before you added extra members to this enum, the very last member already had a trailing comma. Doing this in enum/list/string literals is allowed and considered a good practice here which usually makes formatting + potential future diffs cleaner.
```suggestion
    AnnounceOrange,
```

neil-lobo:
> I've actually started working on this a couple days ago on a local branch; reworking both IrcMessageHandler::handleUserNoticeMessage and IrcMessageHandler::parseUserNoticeMessage so that they'll construct and add only 1 message with system message at the top, linebreak and then the user's message.

In that case, I think it'd be best to wait for your single announcement message to be complete since announcement colours seem to be too closely related to have 2 branches working on it. I'll close this PR

PR #3785, opened 2022-05-30
Use `setPlainText` when finishing a completion

pajlada:
Pull request checklist:

# Description

Without it, you could "execute html" for yourself in the input box, which had no use

https://i.nuuls.com/sxpyU.png


pajlada:
Looks good to me!
image

PR #3788, opened 2022-05-31
Added chatrooms-user-v1 topic with initial parsing

zneix:
Pull request checklist:

# Description

Should let you track current user's "chatting status", that involves:
- whether you're able to chat or not
- get timeout/ban reasons
- get notified whenever you get untimedout/unbanned
- will work for all all chats
- implementing this will allow us to "count down" remaining timeout duration -> could perhaps rework the placeholder text and make it more useful.

WIP, but once it's done it might supress #3329


zneix:
nobody will like this pr anyway

PR #3798, opened 2022-06-05
Rewrite & optimize LimitedQueue

### LINKED ISSUE #1421: Crash if recent messages endpoint returns 1000 messages

dnsge:
# Description
Rewrites `LimitedQueue<T>` to use a circular buffer instead of allocating and moving around tons of `shared_ptrs`. Adds some additional features to the limited queue like `find`/`rfind` so that taking snapshots of the entire queue is less necessary. `LimitedQueueSnapshot<T>` has also been updated with convenience functions like `begin`/`end` for use in range-based for loops/libraries.

Attempting to profile the application on macOS with Instruments (lol) shows a minor speedup and a bit lower memory usage, but I'm not sure if I'm really using the profiler correctly. I added some benchmarks and tests, too.

This rewrite also seems to fix #1421, where Chatterino would crash on receiving a message after loading 1000 historical messages.

**Disclaimer: I was not 100% sure which methods in `LimitedQueue` did _not_ need to be guarded by the mutex, so please verify my decisions.**

<details>
<summary>Benchmark (w/ mutex locking)</summary>

This PR (5af89c3):
<pre>
Run on (8 X 24.1208 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x8)
Load Average: 4.86, 3.55, 2.54
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
BM_LimitedQueue_PushBack                     19.4 ns         19.4 ns     34152006
BM_LimitedQueue_PushFront_One                 670 ns          675 ns      1039594
BM_LimitedQueue_PushFront_Many               2834 ns         2834 ns       247147
BM_LimitedQueue_Replace                       261 ns          261 ns      2678934
BM_LimitedQueue_Snapshot                     1030 ns         1030 ns       679704
BM_LimitedQueue_Snapshot_ExpensiveCopy       3710 ns         3710 ns       188782
BM_LimitedQueue_Find                         1729 ns         1729 ns       405879
</pre>
</details>

## Pull request checklist:


pajlada (commented):
Really cool, I pushed some random small things that I noticed.
Once you're happy with this to receive a full review you can poke me and we can get this tested and merged in!

dnsge:
@pajlada your changes look great, thanks for doing the `shared_mutex` stuff!

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,245 @@
...
+        return this->buffer_.full();
+    }
 
-            // push back new chunk
-            auto newChunk = std::make_shared<Chunk>();
-            newChunk->resize(this->chunkSize_);
-            newVector->push_back(newChunk);
+    [[nodiscard]] size_t space() const
+    {
+        return this->limit() - this->size();
```
iProdigy, line 49: 👍 no locking needed here (would be handled by size call)

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,245 @@
...
+     *
+     * @param items the vector of items to push
+     * @return vector of elements that were pushed
+     */
+    std::vector<T> pushFront(const std::vector<T> &items)
+    {
+        std::unique_lock lock(this->mutex_);
 
-            this->chunks_ = newChunks;
+        size_t numToPush = std::min(items.size(), this->space());
```
iProdigy, line 130:
```suggestion
        size_t numToPush = std::min(items.size(), this->limit() - this->buffer_.size());
```

if a shared lock is applied to `this->size()`, calling `space()` here could be problematic without a recursive lock (as `space()` calls `this->size()`), so this should avoid a self-deadlock scenario

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,245 @@
...
+     * @brief Replace the item at index with the given item
+     *
+     * @param[in] index the index of the item to replace
+     * @param[in] replacement the item to put in place of the item at index
+     * @return true if a replacement took place
+     */
     bool replaceItem(size_t index, const T &replacement)
     {
-        std::lock_guard<std::mutex> lock(this->mutex_);
+        std::unique_lock lock(this->mutex_);
```
iProdigy, line 176: Commenting on this instance, but referring to all uses of `unique_lock`: given that there's only one mutex to be locked & no early unlocking that occurs during the block, can't we use the simpler `lock_guard` (avoiding a little overhead from the extra features of `unique_lock`)?

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,245 @@
...
-            this->lastChunkEnd_ = 0;
-            lastChunk = this->chunks_->back();
-        }
+    // Value Accessors
+    // copies of values are returned so that references aren't invalidated
 
-        lastChunk->at(this->lastChunkEnd_++) = item;
+    [[nodiscard]] T at(size_t index) const
+    {
+        assert(index < this->buffer_.size());
```
iProdigy, line 57:
```suggestion
        std::shared_lock lock(this->mutex_);
        assert(index < this->buffer_.size());
```

iProdigy (commented):
Nice work; added a few lock-related comments but I'm far from an expert on C++ concurrency

src/messages/LimitedQueue.hpp
pajlada: To me, it's mostly the semantics of `unique_lock` vs `shared_lock` that matter most - explicitly declaring whether or not we can share access to it

iProdigy:
Agree, can resolve the shared lock concerns if the methods were private (as the calling scope likely would've already acquired a lock)

dnsge:
I added most of the suggested locks. However, I'm still unsure of the necessity of locking for simple accessors like `empty()`  or `space()` that return integral types.

I was reading [this answer](https://stackoverflow.com/a/2252478) on StackOverflow which seems to suggest it might not be needed, but I'm not very well versed with multi-threaded programming so I still want more discussion around it.

iProdigy:
the safest approach is to lock or privatize

in the public + no lock scenario,
1. to avoid undefined behavior with a concurrent read/write, you'd want the int to be updated atomically. since `std::atomic` isn't explicitly being used, the atomicity depends on the underlying architecture (and I would prefer the variable marked as `volatile` to avoid compiler optimizations that could get in the way of atomicity, but that's not possible here)
2. further, possible memory barriers could make concurrent reads on different threads yield different values, even if there is no concurrent write (though this degree of correctness probably isn't strictly required for this application)

src/widgets/helper/ChannelView.cpp
```diff
@@ -945,11 +934,9 @@ void ChannelView::messageReplaced(size_t index, MessagePtr &replacement)
 
 void ChannelView::updateLastReadMessage()
 {
-    auto _snapshot = this->getMessagesSnapshot();
-
-    if (_snapshot.size() > 0)
+    if (!this->messages_.empty())
     {
-        this->lastReadMessage_ = _snapshot[_snapshot.size() - 1];
+        this->lastReadMessage_ = this->messages_.back();
```
iProdigy, line 939: queue is not guaranteed to be non-empty at the time of the `back` call

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,266 @@
...
     }
 
-private:
-    qsizetype space() const
+    // Actions
+
+    /**
+     * @brief Returns the first item matching a predicate, starting from the front
+     * 
+     * @param[in] pred predicate that will be applied to items
```
pajlada, line 219: improve comment

src/messages/LimitedQueue.hpp
```diff
@@ -2,296 +2,266 @@
...
 
-        return this->limit_ - totalSize;
+        return boost::none;
     }
 
-    bool deleteFirstItem(T &deleted)
+    /**
+     * @brief Returns the first item matching a predicate, starting from the back
+     * 
+     * @param[in] pred predicate that will be applied to items
```
pajlada, line 241: improve comment

src/messages/LimitedQueueSnapshot.hpp
```diff
@@ -1,60 +1,61 @@
 #pragma once
 
+#include <boost/circular_buffer.hpp>
 #include <cassert>
 #include <memory>
 #include <vector>
```
pajlada, line 6:
```suggestion
#include <boost/circular_buffer.hpp>

#include <cassert>
#include <memory>
#include <vector>
```

src/messages/LimitedQueue.hpp
dnsge: attempted in fd05420

src/messages/LimitedQueue.hpp
dnsge: attempted in fd05420

dnsge:
I added two more accessors, `first` and `last` which are the `get` equivalent of `front` and `back`. Adding the two made me wonder whether we should even bother supporting unchecked access of elements because it seems hard for callers to make guarantees about the size/emptiness of the queue.

We could just have `front`, `back`, and `at` which all return `boost::optional<T>` instead of having two for each type of access.

leon-richardt:
> Adding the two made me wonder whether we should even bother supporting unchecked access of elements because it seems hard for callers to make guarantees about the size/emptiness of the queue.

Please correct me if I'm wrong but in strictly single-threaded use, callers should be able to make such guarantees, right?

iProdigy:
> Please correct me if I'm wrong but in strictly single-threaded use, callers should be able to make such guarantees, right?

that's correct (no issues if all reads/writes done from the same thread)

pajlada (approved):

PR #3799, opened 2022-06-05
Fix crash that can occur when changing channels

pajlada:
Pull request checklist:

# Description

The signal that handled the splitinput completer never disconnected, so
if it closed it would access bad memory


zneix (approved):
LGTM

PR #3810, opened 2022-06-10
Add useful streamlink args and link to their documentation website

goldbattle:
Pull request checklist:

# Description

image

### DIFF
--- a/src/util/StreamLink.cpp
+++ b/src/util/StreamLink.cpp
@@ -272,7 +272,7 @@ void openStreamlinkForChannel(const QString &channel)
         args << "--stream-sorting-excludes" << exclude;
     }
 
-    // Append any extra options to to our streamlink command
+    // Append any extra options to our streamlink command
     if (getSettings()->streamlinkOptsLatency)
     {
         args << "--twitch-low-latency";
--- a/src/util/StreamLink.cpp
+++ b/src/util/StreamLink.cpp
@@ -272,6 +272,16 @@ void openStreamlinkForChannel(const QString &channel)
         args << "--stream-sorting-excludes" << exclude;
     }
 
+    // Append any extra options to to our streamlink command
+    if (getSettings()->streamlinkOptsLatency)
+    {
+        args << "--twitch-low-latency";
+    }
+    if (getSettings()->streamlinkOptsAds)
+    {
+        args << "--twitch-disable-ads";
+    }
+
     openStreamlink(channelURL, quality, args);
 }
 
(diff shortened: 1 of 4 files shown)


src/util/StreamLink.cpp
```diff
@@ -272,6 +272,16 @@ void openStreamlinkForChannel(const QString &channel)
...
+    // Append any extra options to our streamlink command
+    if (getSettings()->streamlinkOptsLatency)
+    {
+        args << "--twitch-low-latency";
+    }
+    if (getSettings()->streamlinkOptsAds)
+    {
+        args << "--twitch-disable-ads";
+    }
+
```
pajlada, line 284: We should not be suggesting options for use with streamlink

pajlada (changes requested):

src/util/StreamLink.cpp
goldbattle: Good point, I meant to default them to false / disabled. Should be fixed now.

Felanbird: > suggesting options

This still suggests options, he means remove the suggestions.

goldbattle: Oh, then I would disagree. There are only two options streamlink provides besides disabling of reruns. The user should be able to quickly toggle them if they want to. Each time I do a fresh install I need to find these arguments. I don't see why people would not want this.

Felanbird: > Oh, then I would disagree. There are only two options streamlink provides besides disabling of reruns. The user should be able to quickly toggle them if they want to. Each time I do a fresh install I need to find these arguments. I don't see why people would not want this.

Because one of your suggestions is very obviously something Twitch does not want to support, and we aren't going to bring attention to said frowned upon functions from another application, in this one. 
If we were going to spit in Twitch's face I'd much rather do it in better functional ways.

zneix (changes requested):
I really don't like adding even more and more (unnecessary) settings.
This doesn't do anything code/performance/behaviour-wise. The only thing changed over here is us changing discoverability of a third-party software's features. If that is the intention, which I believe it is, we're better off informing the user about optimal streamlink options/configuration which would be benefitial.

I'm against merging this and rather leaning towards closing.

pajlada:
Feel free to re-open a pull request that links to the documentation of the external application

goldbattle:
Sounds good, no worries.

goldbattle:
Link to docs was added in https://github.com/Chatterino/chatterino2/pull/4217

PR #3811, opened 2022-06-10
Add settings to increase split and usercard scrollback

goldbattle:
Pull request checklist:

# Description

Currently the problem is that for fast chats, the user popup log history doesn't provide much information since only 1000 messages are kept for each channel (. This of course makes the memory of the program very small, but basically only allows the logs to have the most recent message from a user (instead of messages in the past 5 minutes).

This adds two params to configure how many messages are displayed on-screen `ChannelView` and how many are kept in memory for use in the chat log / searching `Channel`.

image

Use of the shared pointer is not ideal, if there is a better way let me know. I didn't dig deep but it errors on me just re-constructing the variable in the constructor (is there a unique pointer someplace in the `LimitedQueue`? not sure..).


pajlada:
@goldbattle Have you tested this functionality with more than 1000 messages in a channel? I suspect it would cause crashes before #3798 has been merged in

Felanbird:
Seems interesting, just wondering how global search will like 8000+ message look-back in every channel.

> I suspect it would cause crashes

Seems to be fine, tested with 1150 messages being render-able, seems the original issue is just loading all of them at once, 800 message history+350 new messages (and 1150 new is fine also).

goldbattle:
I tested up to 5k messages for logs with 2k on screen. Works as expected, but the crash with inserting a lot of historical messages is still an issue in this PR. I saw https://github.com/Chatterino/chatterino2/issues/1421 a couple times when I ran the program.

pajlada (changes requested):
What this feature changes is the scrollback of channels. Update the changelog entry, settings keys, settings names etc to reflect this

The `messages_` member of `Channel` does not need to be a shared_ptr, you just need to initialize it in the constructor (see how `name_`, `type_`, `lastDate_` and `completionModel` are initialized in the constructor)

Get your branch up to date with the main branch since we've changed a lot of the LimitedQueue

src/singletons/Settings.hpp
```diff
@@ -375,6 +375,14 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/misc/twitch/messageHistoryLimit",
         800,
     };
+    IntSetting twitchMessageLogLimit = {
+        "/misc/twitch/messageLogLimit",
+        8000,
+    };
```
pajlada, line 381: The setting controls the amount of scrollback in the Usercard logs
The setting does not do anything Twitch-specific
Setting name should be changed to reflect this
Variable name should be changed to reflect this
Default value should be revererted back to the default 1000

src/singletons/Settings.hpp
```diff
@@ -375,6 +375,14 @@ class Settings : public ABSettings, public ConcurrentSettings
...
+    IntSetting twitchMessageOnScreenLimit = {
+        "/misc/twitch/messageOnScreenLimit",
+        1000,
+    };
```
pajlada, line 385: The setting controls the amount of scrollback in Splits
The setting does not do anything Twitch-specific
Setting name should be changed to reflect this
Variable name should be changed to reflect this

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -693,6 +693,12 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     // TODO: Change phrasing to use better english once we can tag settings, right now it's kept as history instead of historical so that the setting shows up when the user searches for history
     layout.addIntInput("Max number of history messages to load on connect",
                        s.twitchMessageHistoryLimit, 10, 800, 10);
+    layout.addIntInput(
+        "Max number of messages for user chat logs (requires restart) ",
+        s.twitchMessageLogLimit, 1000, 100000, 1000);
+    layout.addIntInput(
+        "Max number of on-screen messages in a channel (requires restart) ",
+        s.twitchMessageOnScreenLimit, 1000, 100000, 1000);
```
pajlada, line 701: Update these labels once the variable names have been changed, also remove the space suffix

pajlada (changes requested):

src/singletons/Settings.hpp
goldbattle: I think the default for the log history should be increased here as in faster chats 1000 isn't enough. For example at 10-20 message per second chat would allow for around 13-10 minutes worth of history with the 8000 limit. With 1000 it is less than a minute for user logs.

pajlada: This is the reason it's configurable - if you run into the issue where messages disappear from the buffer/scrollback too quickly you can change it. For the average user, this change wouldn't change anything except for the memory usage

goldbattle:
Any updates on what is blocking this?

pajlada:
> Any updates on what is blocking this?

Code looks fine, I need to test this out a bit

goldbattle:
Just want to re-ping on this one. Don't want it to get too far behind and have to keep it up to master. Let me know what I need to do.

pajlada (approved):
Changes look good to me now, I'd just like a final review from you @goldbattle to ensure it solves for your issues

My commits should be fairly well-organized to read through, but in short I have:
1) Split up & reworded the changelog entries
2) Ensured the ChannelView's limit is ONLY changed for `Split`s and `UserInfoPopup`s (i.e. usercards)
3) Renamed the settings & all phrasing around it to fit with our glossary


Once you've approved of my changes (or come with feedback) I'd be happy to merge this in.

goldbattle:
Looks good to me. I would still recommend that the default chat log is for usercard is increased Right now usercard log history isn't very helpful in streams > 10k viewers. But I am good either way, thanks for the cleanup.

PR #3813, opened 2022-06-11
Faster startup

goldbattle:
Pull request checklist:

# Description

Since on startup the stream id is found as each TwitchChannel needs to ask for their display name to id conversion. Thus instead of waiting for the roomid over IRC we can get from the api. This speeds up loading a lot (1sec now instead of maybe 10-20sec) and gets rid of cases were sometimes channels fail to load for me.

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -918,6 +918,10 @@ void TwitchChannel::fetchDisplayName()
             }
             channel->addRecentChatter(channel->getDisplayName());
             channel->displayNameChanged.invoke();
+            if (channel->roomId().isEmpty())
+            {
+                channel->setRoomId(user.id);
+            }
         },
         [] {});
 }


iProdigy:
Can you define what you mean by faster startup? Obtaining user id's via helix doesn't change the rate limited irc join task

goldbattle:
IRC is just used for getting messages from chat so you are correct that this does not change that speed.
But there are a lot of other things that rely on the roomId being known.
All these processes are very delayed, with the live and stream title status being of particular usefulness on startup.

Specifically:
```cpp
// room id loaded -> refresh live status
this->roomIdChanged.connect([this]() {
    this->refreshPubSub();
    this->refreshTitle();
    this->refreshLiveStatus();
    this->refreshBadges();
    this->refreshCheerEmotes();
    this->refreshFFZChannelEmotes(false);
    this->refreshBTTVChannelEmotes(false);
});
```

iProdigy:
Also on `setRoomId`, we call `loadRecentMessages` - does this change properly interact with the recent messages api? that is, are historical messages loaded & does this change result in a gap in chat history between the recent message api call & irc join event?

goldbattle:
> Also on `setRoomId`, we call `loadRecentMessages` - does this change properly interact with the recent messages api? that is, are historical messages loaded & does this change result in a gap in chat history between the recent message api call & irc join event?

Yes it does. The query to the recent api uses the displayname, thus doesn't depend on the roomId. The historical service is actually down for me right now, but it worked as intended when I tested it last night. Historical messages are always appended to the top (back of the message queue) above the IRC messages.

zneix:
> Yes it does. The query to the recent api uses the displayname

I don't like this approach, ideally login names should be used. Display names sound like a wrong solution, even if the "non-ascii names" are taken care off.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Minor: Added ability to execute commands on chat messages using the message context menu. (#3738, #3765)
 - Minor: Added `/copy` command. Usage: `/copy <text>`. Copies provided text to clipboard - can be useful with custom commands. (#3763)
 - Minor: Removed total views from the usercard, as Twitch no longer updates the number. (#3792)
+- Minor: Speed up program loading time GO FAST (#3813)
```
zneix, line 24: As per this changelog entry I don't think it's even needed at all. It actually sounds wrong because the whole speedup here is just placebo imo. Nothing speeds up, you just perform certain actions a bit sooner - not faster. Technically program won't finish "loading" until you have JOINed all the channels, which is sadly limitted by twitch to ~20 JOINs/10s so you'll have to wait regardless.

And in the case if I'm actually wrong and the changelog entry is to be kept, the last two words in caps don't make any sense, remove them and end the sentence with a comma (that suggestion is only if we decide to keep this, but my preference is to completely remove it).

zneix:
I'm not sure if this is the right way to do it, we'd load the room ID which is equivalent to us being JOINed the chatroom, but in reality that only happens once we receive either first ROOMSTATE or a JOIN for our own user.
There are be cases where our JOIN request might get lost (either due to somehow getting rate-limited or whatever) which results in us not being JOINed a chatroom yet we'd have its ID loaded anyway. Not sure if my rambling makes sense but tl;dr this'd not speed up anything and could lead to potential unhandled edge-cases.

pajlada:
I'm not sure what this does, feel free to create a new PR with a bit more information about what this is supposed to do, and how it does it.

Issue #2241, opened 2020-11-30
FrankerFaceZ Channel-Specific Badge Assignments


SirStendec:
**What should be added?**
Recently, we (that is FrankerFaceZ) started letting users assign bot badges to users in their channels. This has been exposed in the API for some time, but between poor documentation and the inability for users to take advantage of it, I don't blame anyone for not catching it and/or implementing support. The feature is intended to let users easily flag their own custom bots without any form of approval, while the global bot badge will be limited to bots in widespread use across Twitch.

All single room endpoints now include arrays of usernames and user Twitch IDs for assigning badges to users in those specific rooms. When possible, clients should always use the Twitch ID rather than a username as usernames may be out of sync with Twitch after an account name change.

Data from the API looks something like:
```json
{
	"room": {
		...
		"user_badges": {
			"2": ["quoteconut"]
		},
		"user_badge_ids": {
			"2": [74024612]
		}
	}
}
```

`user_badges` and `user_badge_ids` are objects with keys that are the badge ID as a string (only a string due to the nature of JSON, and can be treated as integers) and the values are arrays of usernames or Twitch IDs.

In the future, this may support badges other than the bot badge. However, for now, it's limited to the bot badge with ID 2.

Side note: We now have automatically generated documentation available at https://api.frankerfacez.com/docs/ describing rooms and all other endpoints.

**Why should it be added?**
I don't know if you'd interested in this sort of feature for Chatterino, but I thought I'd bring it up as something new esp. since Chatterino recently added support for FrankerFaceZ's badges in a global sense.

SputNikPlop:
@pajlada Does this still need to be done and are there any other considerations that need to be taken into account? I know this is a little bit of an older issue.

pajlada:
> @pajlada Does this still need to be done and are there any other considerations that need to be taken into account? I know this is a little bit of an older issue.

This is still not implemented, correct. Nothing special that needs to be taken into account - the API is straight-forward

PR #3818, opened 2022-06-13
Show all global FFZ badges of a user

Jaxkey:
Pull request checklist:

# Description

In Chatterino, a maximum of one global FFZ badge appears at a time.

Moobot has both the `Bot` & `FFZ Supporter` global FFZ Badges.

In Chatterino, the `Bot` badge is missing and only one FFZ badge appears.

Changed `badgeMap` to store a vector of ID's in its second position for each FFZ badge a user has (previously only stored a single ID).
Changed `appendFfzBadges()` to loop through each FFZ badge a user has.

Updated version:

### To Reproduce

Have a user with multiple global FFZ badges send a chat message


Felanbird:
I'm not sure if this is the correct approach, the FFZ Bot badges we load in Chatterino are a relic of the past and purely appear because they are in the same API as supporter badges. 
I'm wondering how this will effect any future implementation of #2241.

Infinitay:
Jumping on to mention IIRC one service _combines_ a bot badge with a mod badge if that user is defined as a bot of a channel. Not sure if it's specific to BTTV or FFZ who combines them.

pajlada:
This doesn't really conflict with the transition to v2 since I don't think anyone is working on that right now, so I'm happy to merge this feature in. Will push a little commit cleaning up the code at the same time as I solve the merge conflict, then let @Jaxkey  get the last review in

pajlada (approved):
Works as expected :+1: 
Happy to merge upon your approval of my changes @Jaxkey 

image

Jaxkey:
Looks good @pajlada. Merge it 👍

PR #3822, opened 2022-06-15
Make generic version of batcher function

zneix:
Pull request checklist:

# Description

Removed duplicated declarations of functions doing the same across the codebase - there's now 1 definition in `util/Helpers.hpp` which also has reworked logic which is meant to be more clear and more readable.
Updated tests accordingly.

## Notes

Because of how iterators specific to `QList<T>` work, this is the only possible type that the function can take. I tried moving "batch append" part to a separate function with different overloads for `QList<T>` and other generic types like `std::set` (which mostly implement `Type::emplace_back` and `it::value()` for their iterators), yet that didn't exactly work how I planned it. There was also an issue where `QStringList` wasn't properly understood by the helper function as `QList<QString>` - otherwise I'd simply declare the helper function as `std::vector<QList<T>> splitListIntoBatches(const QList<T> list, int batchSize = 100)`.

P.S.: Review from @leon-richardt would be very much welcome!


pajlada (commented):
Good to merge once YungLPR's suggestion has been implemented

pajlada (approved):

src/util/Helpers.hpp
```diff
@@ -37,4 +37,26 @@ QColor getRandomColor(const QString &userId);
 QString formatUserMention(const QString &userName, bool isFirstWord,
                           bool mentionUsersWithComma);
 
+template <typename T>
+std::vector<T> splitListIntoBatches(const T list, int batchSize = 100)
+{
+    std::vector<T> batches;
```
iProdigy, line 43:
```suggestion
    int batchCount = (list.size() / batchSize) + 1;
    std::vector<T> batches;
    batches.reserve(batchCount);
```

zneix: @iProdigy instead of just giving out suggestion please tell me why and/or what's the benefit here. I had that call to `reserve` function in the previous implementation but I couldn't find any reference or documentation on whether I can/could/should/have to do this. it doesn't seem to change anything and is just unnecessary imo. I can add this, I just want to know why...

iProdigy: Since `std::vector` has a dynamic size, it may repeatedly reallocate space as elements are added to ensure there is sufficient space. These allocations are not costless & can be optimized by simply indicating the target size

zneix: Thank you so much! It's very useful to know for the future, I'll add this fix right now.

src/util/Helpers.hpp
```diff
@@ -37,4 +39,28 @@ QColor getRandomColor(const QString &userId);
...
+    for (int j = 0; j < batchCount; j++)
+    {
+        T batch;
+
+        for (int i = 0; i < batchSize && it != list.end(); i++)
+        {
+            batch.append(*it);
+            it++;
+        }
+        batches.emplace_back(std::move(batch));
```
iProdigy, line 60: perhaps we can break early if batch is empty (in case of concurrent modifications making the list smaller); else emplace_back

src/util/Helpers.hpp
```diff
@@ -37,4 +39,28 @@ QColor getRandomColor(const QString &userId);
 QString formatUserMention(const QString &userName, bool isFirstWord,
                           bool mentionUsersWithComma);
 
+template <typename T>
+std::vector<T> splitListIntoBatches(const T &list, int batchSize = 100)
+{
+    std::vector<T> batches;
+    int batchCount = std::ceil(list.size() / batchSize);
```
Mm2PL, line 46:
```suggestion
    int batchCount = std::ceil(static_cast<double>(list.size()) / batchSize);
```

When not using floating point numbers here, this breaks and never rounds up.

iProdigy (approved):

Mm2PL (approved):

Issue #3827, opened 2022-06-21
Quick Switcher does not focus correct window when opened from a popup window


leon-richardt:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Description
When opening the Quick Switcher via a popup window and switching splits with it, focus is not transferred to the window the split resides in.

### Steps to Reproduce
1. Open a popup window (e.g., via the Ctrl + Shift + N popup).
2. Open the Quick Switcher (Ctrl + K).
3. Enter the name of an existing channel and confirm the "Switch to tab" item.

### Expected Behaviour
Focus is transferred to the window the selected split/tab resides in.

**Note:** I'm not sure if OS/WM policies allow applications to transfer focus between windows like that. But I would assume so.

### OS and Chatterino Version

Chatterino 2.3.5 (commit 57783c7) running on Arch Linux, kernel: 5.18.5-arch1-1

pajlada:
Fixed in https://github.com/Chatterino/chatterino2/pull/4819 
The highlight stays with the correct window, popup or main window

Issue #3826, opened 2022-06-21
Quick Switcher adds tabs to main window instead of popup window


leon-richardt:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Description
When opening the Quick Switcher via a popup window and adding channels with it, tabs are always added to the main window instead of the popup window.

### Steps to Reproduce
1. Open a popup window (e.g., via the <kbd>Ctrl + Shift + N</kbd> keybind).
2. Open the Quick Switcher (<kbd>Ctrl + K</kbd>).
3. Enter a channel name and confirm the "Open in new tab" item.

### Expected Behaviour
The channel is added to the popup window the Quick Switcher was opened from.

### OS and Chatterino Version

Chatterino 2.3.5 (commit 57783c747) running on Arch Linux, kernel: 5.18.5-arch1-1

pajlada:
Fixed in https://github.com/Chatterino/chatterino2/pull/4819

PR #3828, opened 2022-06-21
feat: add Quick Switcher item to open channel in a new popup window

leon-richardt:
Pull request checklist:

# Description

This patch adds the possibility to open a channel in a new popup via the Quick Switcher. I think this change is worth inclusion since I often briefly want to check out other channels without including them in my permanent tab layout.

As you can tell by the provided icon, I'm still pretty bad at design. Feel free to replace this icon with anything you see more fit.

## Related Issues
During implementation, the following issues were discovered:
- #3826 
- #3827


pajlada (approved):
Very nice, and good bug finds!

PR #3839, opened 2022-06-26
Remove official support for QMake

Mm2PL:
Pull request checklist:

# Description

Totally removes QMake from Chatterino.


Felanbird:
1 step closer to spellcheck 🙂

pajlada (changes requested):
The release asset uploads are defined twice now, like this
```yaml
      - name: Upload release asset (Windows)
        uses: actions/upload-release-asset@v1.0.2
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }}
          asset_path: ./windows/chatterino-windows-x86-64.zip
          asset_name: chatterino-windows-x86-64.zip
          asset_content_type: application/zip

      - name: Upload release asset (Windows)
        uses: actions/upload-release-asset@v1.0.2
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }}
          asset_path: ./windows/chatterino-windows-x86-64.zip
          asset_name: chatterino-windows-x86-64.zip
          asset_content_type: application/zip
```

Remove one of each

pajlada:
Would like confirmation from a Windows developer that the instructions make sense

zneix:
Checklist for all remaining references to qmake that we should revisit/remove:
- [ ] `.CI/CreateAppImage.sh`
- [ ] `.CI/CreateDMG.sh`
- [ ] `.CI/InstallQTStylePlugins.sh`
- [ ] `.docker/Dockerfile.build` NOTE: There's also a reference to Ubuntu 18.04, but we've recently came to a conclusion that Ubuntu 20.04 is required at minimum
- [ ] `conanfile.txt`
- [ ] `tools/build-breakpad-linux.sh`

Mm2PL:
@pajlada do you think it's okay to get rid of the Docker build thing introduced in e253208eb? it's not mentioned anywhere in our docs, nothing seems to use it, it was never updated. It uses qmake and old Ubuntu.

Felanbird (changes requested):
> Would like confirmation from a Windows developer that the instructions make sense

Due to the fact that the instructions itself were not changed, I took the approach of attempting to utilize `CMakeLists.txt` while already having completed the build steps prior for QMake.

The only issue I ran into was the import refusing to proceed unless I installed conan, once I did everything worked exactly as expected and after building, Chatterino started up.

Somewhere in the above build docs this should be mentioned (but I'm unsure specifically where), but if we keep the additional CMake section, it might to mention this part is only for running with QtCreator.

```
Install [conan](https://conan.io/downloads.html) and make sure it's in your `PATH` (default setting). Note: If conan does not automatically add itself to your `PATH` - a guide can be found [here](https://www.computerhope.com/issues/ch000549.htm#windows10).
```

pajlada:
> @pajlada do you think it's okay to get rid of the Docker build thing introduced in [e253208](https://github.com/Chatterino/chatterino2/commit/e253208ebdef7322b274667acf0890fca7c83aed)? it's not mentioned anywhere in our docs, nothing seems to use it, it was never updated. It uses qmake and old Ubuntu.

Fine by me

pajlada:
> Checklist for all remaining references to qmake that we should revisit/remove:
> 
>     * [x]  `.CI/CreateAppImage.sh` (Mm2PL: I think linuxdeployqt requires qmake to be present to find qt but it doesn't compile anything with it)
> 
>     * [x]  `.CI/CreateDMG.sh` (Mm2PL: There doesn't seem to be anything to change here?)
> 
>     * [x]  `.CI/InstallQTStylePlugins.sh` (Mm2PL: only travis used this, but since that's going, might as well delete it)
> 
>     * [ ]  `.docker/Dockerfile.build` NOTE: There's also a reference to Ubuntu 18.04, but we've recently came to a conclusion that Ubuntu 20.04 is required at minimum
> 
>     * [ ]  `tools/docker/build.sh`
> 
>     * [x]  `conanfile.txt`
> 
>     * [ ]  `tools/build-breakpad-linux.sh`
> 
>     * [x]  `appveyor.yml` NOTE: This file served same purpose as travis, should be removed altogether as well.

I've deleted the .docker directory, the tools/docker directory, and the tools/build-break-pad-linux.sh script

pajlada (approved):
The changes look good to me, I still want feedback from Windows developers to make sure we're not messing up their flow too much.

Felanbird:
> I still want feedback from Windows developers to make sure we're not messing up their flow too much.
> 
> EDIT: Specifically, it's the "simplest QtCreator build flow" I want to make sure works as expected

@brian6932 is going to look into this once he's back from vacation.

brian6932 (approved):
LGTM 👍, the things I noticed while building was that there are a lot of errors/warnings when you load the project into QtCreator (same issue happens in both preview and non-preview versions), however, it does still build.
QtCreator
QtCreatorPreview
I also had some weird issue with Ninja (ninja build stopped subcommand failed.) at one point, but after re-cloning, it went away.

Mm2PL:
Maybe if Conan isn't detected on windows we should just set the skip var shown in Brian's errors, but I think that's out of scope for this pr.

Felanbird:
I resolved the issues stated below with some git submodule syncing 
```
git clean -xdf
git submodule sync
git submodule deinit -f .
git submodule update --init --recursive
```

~~I've run into this ninja issue as well, and just to be sure I jumped backwards to commit f63f4de (right before pajlada's changes) and the issue resolved itself for a while, then swapping back around to test the issue caused the error to happen no matter what commit.~~

~~Swapping to a release build of f63f4de allowed building again, attempting to build release with all commits causes the issue again. If brian was able to resolve the issue with a re-clone maybe it's just another god of war only-during-PR-testing-issue like during the custom hotkeys PR 🤷~~

Felanbird (approved):

pajlada (approved):
I am going to merge this in

dnsge:
I'm a little late, but is there a reason that we didn't remove the `lib/*.pri` files? It seems that the only references to these files were in the `chatterino.pro` file which was deleted.

If we _do_ want to remove them now, I suppose we should find a place to insert the contents of `lib/warnings.pri`? I don't know much about configuring CMake so I'm not sure how we'd go about it but I just wanted to bring it up.

Mm2PL:
looks like i forgot those existed

PR #3877, opened 2022-07-24
Disable submodule-build of QtKeychain

pajlada:
Pull request checklist:

# Description

The original reason for using the fork was to support our qmake build method. The changes we had made specifically were:
 - Updated the .pri file to point to the right mac/apple-specific file (since added upstream)
 - Added #undef singals in libsecret.cpp to fix compilation. (no longer necessary since we build with cmake now)

However, it's not as easy as just swapping to upstream - the way the library is built makes it difficult to include as part of our build process. Because of this, we now use `FetchContent` to ensure the build happens completely separately, and it also provides us with proper build/source directories we can use to find the include files.

~**The PR currently fetches the project from the internet, however once the build is stable the submodule will be updated and we'll just point to the local directory instead.**~

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,9 +20,6 @@
 [submodule "lib/rapidjson"]
 	path = lib/rapidjson
 	url = https://github.com/Tencent/rapidjson
-[submodule "lib/qtkeychain"]
-	path = lib/qtkeychain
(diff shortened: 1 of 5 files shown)


pajlada:
Not very feasible atm, closing

PR #3878, opened 2022-07-25
Automatically load recent messages on reconnect

### LINKED ISSUE #1308: Add ability to trigger the recent-messages api without restarting the client.

When the connection is dropped and Chatterino reconnects. There's no way to see what messages were sent between the disconnect and reconnect.

I would like a menu option and/or a command that shows historical messages obtained using the recent-messages service using a popup similar to the search one.

 * Not doing it at all
 * Automatically filling in the messages that couldn't be received. If the API doesn't have a parameter in the request to select the time we want the messages from, there would be a need to filter them based on the timestamps.

**Additional context**

screenshot of the search popup
Currently there is no way to see the recent messages without using logs or restarting. Logs don't show every users' actions.

dnsge:
Pull request checklist:

# Description

When Chatterino drops the connection for whatever reason, messages can be missed before it gets a chance to reconnect. This PR adds functionality that automatically fetches the messages that might have been missed from the recent messages API.

<details>
<summary>Video in fast moving chat</summary>

https://user-images.githubusercontent.com/24928223/180701202-d95bb5c9-b7e6-405a-875b-931f6cdd1171.mov

</details>

Closes #1308

# Implementation Details

The recent messages are inserted by referencing the `Message::serverReceivedTime` field. Two methods were added to `LimitedQueue` to support this feature:
 - `insertBefore`, used to insert messages in time ascending order
 - `insertAfter`, also used to insert messages in time ascending order but when we can't find a message that comes later (typically in slow chats where no messages are sent between reconnecting and the recent messages API returning a response)

Also adds the `Channel::filledInMessages` signal that notifies the `ChannelView` that messages were filled in and allows the ChannelView to proxy the messages from the underlying channel that the fill occurred in to the (potentially) filtered channel used to display messages.

~As a side note, the implementation does not make the assumption that the recent messages API returns messages with their `serverReceivedTime` value in ascending order.~ Actually, it does make this assumption upon further reflection.


src/common/QLogging.cpp
```diff
@@ -40,3 +40,5 @@ Q_LOGGING_CATEGORY(chatterinoWidget, "chatterino.widget", logThreshold);
 Q_LOGGING_CATEGORY(chatterinoWindowmanager, "chatterino.windowmanager",
                    logThreshold);
 Q_LOGGING_CATEGORY(chatterinoHighlights, "chatterino.highlights", logThreshold);
+Q_LOGGING_CATEGORY(chatterinoRecentMessages, "chatterino.recentmessages",
+                   logThreshold);
```
leon-richardt, line 44: Originally, logging categories were kept in alphabetic order. @pajlada (supposedly unintentionally?) broke with this tradition in 7ccf60111d5e8ba54c070fe09d926bdff5290da5 but I would suggest we keep it up for new additions.

src/common/QLogging.hpp
```diff
@@ -31,3 +31,4 @@ Q_DECLARE_LOGGING_CATEGORY(chatterinoWebsocket);
 Q_DECLARE_LOGGING_CATEGORY(chatterinoWidget);
 Q_DECLARE_LOGGING_CATEGORY(chatterinoWindowmanager);
 Q_DECLARE_LOGGING_CATEGORY(chatterinoHighlights);
+Q_DECLARE_LOGGING_CATEGORY(chatterinoRecentMessages);
```
leon-richardt, line 34: Originally, logging categories were kept in alphabetic order. @pajlada (supposedly unintentionally?) broke with this tradition in 7ccf60111d5e8ba54c070fe09d926bdff5290da5 but I would suggest we keep it up for new additions.

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -163,6 +165,8 @@ class TwitchChannel : public Channel, public ChannelChatters
     void refreshBadges();
     void refreshCheerEmotes();
     void loadRecentMessages();
+    void loadRecentMessagesReconnect();
+    // void fillInMissingMessages(const std::vector<MessagePtr> &messages);
```
leon-richardt, line 169:
```suggestion
```

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -14,13 +14,15 @@
 #include "providers/twitch/api/Helix.hpp"
 #include "util/QStringHash.hpp"
 
+#include <IrcMessage>
```
leon-richardt, line 17:
```suggestion
```

Can be removed I think?

src/widgets/helper/ChannelView.cpp
```diff
@@ -940,6 +957,41 @@ void ChannelView::messageReplaced(size_t index, MessagePtr &replacement)
...
+    auto snapshot = this->channel_->getMessageSnapshot();
+
+    this->messages_.clear();
+    this->scrollBar_->clearHighlights();
+    this->lastMessageHasAlternateBackground_ = false;
+    this->lastMessageHasAlternateBackgroundReverse_ = true;
+
+    for (const auto &msg : snapshot)
+    {
+        auto messageLayout = new MessageLayout(msg);
```
leon-richardt, line 971: Why not:
```suggestion
        auto messageLayout = std:.make_shared<MessageLayout>(msg);
```
(And then moving the `shared_ptr` into `this->messages_.push_back()` later.)

I would feel safer that way

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Major: Added support for Twitch's Chat Replies. [Wiki Page](https://wiki.chatterino.com/Features/#message-replies) (#3722)
 - Major: Added multi-channel searching to search dialog via keyboard shortcut. (Ctrl+Shift+F by default) (#3694, #3875)
+- Minor: Load missing messages from Recent Messages API upon reconnecting (#3878)
```
leon-richardt, line 7: Is the changelog entry intentionally at the top of the "Minor" section? I was under the impression we usually place them at the bottom of the respective section

leon-richardt (changes requested):
A little bit of nitpicking but overall LGTM :+1:

src/common/Channel.cpp
```diff
@@ -231,6 +231,70 @@ void Channel::addMessagesAtStart(std::vector<MessagePtr> &_messages)
...
+        {
+            // We never found a message already in the channel that came after
+            // the current message. Put it at the end and make sure to update
+            // which message is considered "the end".
+            this->messages_.insertAfter(lastMsg, msg);
+            lastMsg = msg;
+        }
+    }
+
+    // We only invoke a signal once at the end of filling all messages to
```
pajlada, line 293: If none of the messages received from the API were inserted, we still invoke the `filledInMessages`.
Would it make sense to use a variable to keep track of whether _any_ message has been inserted? And if not, skip invoking the signal?

src/providers/RecentMessagesApi.hpp
```diff
@@ -0,0 +1,247 @@
...
+public:
+    /**
+     * @brief Loads recent messages for a channel using the Recent Messages API
+     * 
+     * @param channelName Name of Twitch channel
+     * @param channelPtr Weak pointer to Channel to use to build messages
+     * @param onLoaded Callback taking the built messages as a std::vector<MessagePtr>
+     * @param onError Callback called when the network request fails
+     */
+    template <typename OnLoaded, typename OnError>
```
pajlada, line 174: Since the callbacks are well-defined in their parameters and return value, I'd prefer if they were defined as `std::function`'s in the parameters.
That would also allow us to make the file a standard header/source file combo.

pajlada (changes requested):

src/common/QLogging.cpp
dnsge: Realphabetized in c0a20a0

src/providers/twitch/TwitchChannel.hpp
dnsge: Whoops, removed

src/widgets/helper/ChannelView.cpp
dnsge: I went ahead and made this change (and the change in other places in the file where similar code occurs). The only """issue""" is that `LimitedQueue` doesn't have a version of `pushBack` that takes an rvalue, so (I think) that means we can't `std::move` into the `LimitedQueue`. Not a big deal.

CHANGELOG.md
dnsge: I was under the impression we placed new entries at the top of the section 😅 But I went ahead and reordered it

src/common/Channel.cpp
dnsge: Good point 👍

src/providers/RecentMessagesApi.hpp
dnsge: Updated this. Was also thinking of doing this because it would probably have needed to be done anyways when #3901 is updated with this PR.

pajlada (approved):

src/widgets/helper/ChannelView.cpp
leon-richardt: Good catch, didn't even think about that. At least we're future-proof this way if anyone ever adds a rvalue version (which will surely happen ...)

leon-richardt (approved):
:+1:

PR #3884, opened 2022-07-29
Add ability to pin Usercards to stay open even if it loses focus

goldbattle:
Pull request checklist:

# Description

This adds a pin button so that user pin cards can be pinned and not closed after clicking off.
It is useful to be able to pin a specific user, but be able to still open and close others quickly.

https://user-images.githubusercontent.com/2222562/181842903-70e5acf7-230a-455e-b3f9-370a9720fc88.mp4


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -134,13 +134,14 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, QWidget *parent)
...
     , dragTimer_(this)
 {
     this->setWindowTitle("Usercard");
     this->setStayInScreenRect(true);
 
-    if (closeAutomatically)
+    if (this->closeAutomatically_)
         this->setActionOnFocusLoss(BaseWindow::Delete);
     else
         this->setAttribute(Qt::WA_DeleteOnClose);
```
pajlada, line 147: Move this section into its own private function, something like `updateFocusLoss` or something which can then be called from the "on pin change" connection

pajlada (changes requested):
If you have an SVG of the resources added, please include them for when we want to scale these features

goldbattle:
> If you have an SVG of the resources added, please include them for when we want to scale these features

Unfortunately I just edited a png in GIMP.

src/widgets/dialogs/UserInfoPopup.cpp
goldbattle: I left the case for `this->setAttribute(Qt::WA_DeleteOnClose);` but have both here and the button press call the update focus function.

resources/resources_autogenerated.qrc
```diff
@@ -1,122 +1,122 @@
...
-        <file>twitch/prime.png</file>
-        <file>twitch/staff.png</file>
-        <file>twitch/subscriber.png</file>
-        <file>twitch/turbo.png</file>
-        <file>twitch/verified.png</file>
-        <file>twitch/vip.png</file>
-        <file>buttons/pinDisabledDark.png</file>
-        <file>buttons/pinDisabledLight.png</file>
-        <file>buttons/pinEnabled.png</file>
-    </qresource>
```
goldbattle: This seems to be what happens when adding the resource through qtcreator. Didn't realize there was a script to do this.

src/widgets/dialogs/UserInfoPopup.cpp
goldbattle: Not a worry any more after merge.

jammehcow:
I've remade the pin assets (slightly different from originals) as PNGs & SVGs:

c2_pin_black c2_pin_black
c2_pin_orange c2_pin_orange
c2_pin_white c2_pin_white

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -867,6 +882,23 @@ void UserInfoPopup::updateUserData()
...
+        if (this->ui_.pinButton != nullptr)
+            this->ui_.pinButton->setPixmap(getApp()->themes->buttons.pin);
+    }
+    else
+    {
+        this->setActionOnFocusLoss(BaseWindow::Nothing);
+        if (this->ui_.pinButton != nullptr)
+            this->ui_.pinButton->setPixmap(getResources().buttons.pinEnabled);
+    }
+}
```
pajlada, line 900: `updateFocusLoss` should not be responsible for toggling the `this->closeAutomatically_` value, rather it should read it as you do on and act on it.

It must also be called in the constructor once to initiate the default value.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - Minor: Added `/copy` command. Usage: `/copy <text>`. Copies provided text to clipboard - can be useful with custom commands. (#3763)
 - Minor: Removed total views from the usercard, as Twitch no longer updates the number. (#3792)
 - Minor: Add Quick Switcher item to open a channel in a new popup window. (#3828)
+- Minor: Add ability to pin user popup dialog (#3884)
```
pajlada, line 28:
```suggestion
- Minor: Added ability to pin individual Usercards to stay open even if it loses focus. Only available if "Automatically close usercard when it loses focus" is enabled. (#3884)
```

pajlada (changes requested):
Another small request which isn't anything you've changed: Could you please change the label of the "Automatically close user popup when it loses focus" setting to "Automatically close usercard when it loses focus"?
Thank you!

goldbattle:
Any updates on what is blocking this? If nothing, I can fix the conflicts so we can merge.

goldbattle:
Just want to re-ping on this one. Don't want it to get too far behind and have to keep it up to master. Let me know what I need to do.

pajlada (approved):

pajlada:
Thank you @goldbattle and sorry for the long review wait :heart:

PR #3901, opened 2022-08-03
[WIP] Speed up loading of recent messages

leon-richardt:
Pull request checklist:

# Description

[WIP]

(See individual commit messages for context.)
~~(Currently includes commits from #3904 but this branch will need to be rebased multiple times anyway before merging.)~~


src/util/RatelimitBucket.hpp
```diff
@@ -6,13 +6,29 @@
...
+        : QObject(parent)
+        , budget_(budget)
+        , cooldown_(cooldown)
+        , callback_(callback)
+    {
+    }
 
-    void send(QString channel);
+    // QList::append() only takes const T& anyway
+    void handle(const T &argument)
```
leon-richardt, line 23: Note to self: rename to `submit()`

leon-richardt:
(I am aware of https://github.com/Chatterino/chatterino2/pull/3878 and would be fine to update my branch as soon as #3878 is merged into `maszer`.)

src/util/RatelimitBucket.hpp
```diff
@@ -25,16 +41,32 @@ class RatelimitBucket : public QObject
...
+            return;
+        }
+
+        auto item = this->queue_.takeFirst();
+
+        this->budget_--;
+        this->callback_(item);
+
+        QTimer::singleShot(cooldown_, this, [this] {
+            this->budget_++;
```
iProdigy, line 66: consider making budget atomic if `submit` could be called separately from the qt event loop

src/providers/recentmessages/RecentMessagesProvider.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+    void queueLoadFor(std::weak_ptr<Channel> channel);
+
+private:
+    RecentMessagesProvider();
+
+    // TODO(leon): Figure out reasonable values for these.
+    // TODO(leon): Potentially make these configurable for custom message providers
+    static constexpr int REQUEST_BUDGET{10};
+    static constexpr int REQUEST_COOLDOWN{1000};
```
dnsge, line 28: If we go the configurable route, which I don't see why not, these should be environment variables. A value of zero could disable rate limiting.

src/providers/recentmessages/RecentMessagesProvider.cpp
```diff
@@ -0,0 +1,218 @@
...
+                    {
+                        builtMessage->flags.set(MessageFlag::RecentMessage);
+                        allBuiltMessages.emplace_back(builtMessage);
+                    }
+                }
+
+                postToThread([ownedChannel, root,
+                              messages =
+                                  std::move(allBuiltMessages)]() mutable {
+                    ownedChannel->addMessagesAtStart(messages);
```
dnsge, line 166: I don't really like having the loading implementation doing the adding of messages to the channel -- I think it's better to keep that logic separate in `TwitchChannel` and just have this class handle the loading and parsing of the messages.

Somewhat related to how I implemented the #3878, so not a big deal right now.

leon-richardt: Fair point but as you said, I would like to wait for #3878 to be merged in before making any major design changes

src/util/RatelimitBucket.hpp
leon-richardt: For now, I opted to make the minimal required changes to `RatelimitBucket` but I will consider this

src/providers/recentmessages/RecentMessagesProvider.hpp
leon-richardt: I agree, I intend to follow the established format of the existing `CHATTERINO2_RECENT_MESSAGES_URL` here

src/util/RatelimitBucket.hpp
leon-richardt: (Resolved via rebase)

PR #3904, opened 2022-08-05
chore: warn when parsing environment variable fails

leon-richardt:
Pull request checklist:

## Description

Provides a "generic" function to warn when parsing an environment variable (e.g., to a `uint16_t`) fails. Currently, this is only applicable to the `CHATTERINO2_TWITCH_SERVER_PORT` variable but I intend to use it for #3901.

Screenshot of what it looks like:
image
image


pajlada (approved):

src/util/TypeName.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <string_view>
+
+namespace chatterino {
+
+// Adapted from: https://stackoverflow.com/a/56766138.
+// NOTE: Relies on the "magic" prefixes and suffixes. There are implementations
+// that attempt to manually detect these (linked in the SO answer above) but
+// they seemed too complex for the scope we have here.
+template <typename T>
+constexpr auto type_name()
```
Mm2PL, line 12: I think we can use typeid(t).name() here? 
https://en.cppreference.com/w/cpp/language/typeid

leon-richardt: https://godbolt.org/z/W65sv4deo

Issue #3062, opened 2021-07-23
Chatterino has a memory usage problem


TETYYS:
**Describe your issue**
Chatterino either leaks or retains too much memory for something. Memory usage seems to increase over time, but not linearly - for example my current chatterino instance uses 1.67GB private bytes for 3d3h runtime.

Applications like [MTuner](https://github.com/milostosic/MTuner) might help to find the cause.

**Screenshots**
![ProcessHacker_hkQ4qxOm6F](https://user-images.githubusercontent.com/3091645/126764261-eccc1866-7394-451c-a299-c9ced0db8250.png)

**OS and Chatterino Version**
Win10 20H2 19042.1052
Chatterino Nightly 2.3.2 (commit f9ddd5303, )

Mm2PL:
Maybe the debug popup (F10) can help.
cf. my one, after a day or so of running 
![2021-07-23_13-52](https://user-images.githubusercontent.com/25011746/126778077-542980d3-a6f6-45b6-bc17-72da9cde65aa.png)

TETYYS:
lots of message layout elements

![chatterino_nJs5uu4Kxd](https://user-images.githubusercontent.com/3091645/126778188-48ed8ca0-a373-421b-8fce-cd3755350f71.png)

Mm2PL:
I'm sure messages are leaked somewhere (I had no channels but there were messages) but I can't replicate it consistently.

TETYYS:
Yeah, seems like it. I deleted and recreated a couple of splits and managed to get it down to 700k, but I see the number going down sometimes, so there are situations where layout elements are deleted

KararTY:
Is it maybe related to the amount of loaded animated emotes?

In your screenshot you have 1800 animated emotes, and my own count of 837 animated emotes puts my mem usage at 900 MB.
![image](https://user-images.githubusercontent.com/13697809/126794497-c7f71688-fe1e-45de-8aa5-69aabf48eeec.png)

I test this by going into channels with a lot of animated BTTV emotes and opening up the emote window and my memory usage consistently goes up and stays that high until I close the tab: https://streamable.com/eqxqcn

ALazyMeme:
Yes I'm also experiencing this high ram usage. As noticed today, my ram usage was up to 1.6gb. After restarting, it dropped down to 600mb.

[...]

PR #3915, opened 2022-08-10
Periodically free memory from unused images

dnsge:

# Description

Attempts to reduce memory usage due to images.

1. Reduces copying of frame data (though this doesn't actually increase memory usage in the long term)
2. Periodically discards frame data of Images that haven't recently been painted

Images store the last time they were accessed via `Image::pixmapOrLoad`. Every **1 minute**, Chatterino will check every `Image` instance and check if the last accessed time was more than **10 minutes** ago. If so, it deletes the frame data and marks the image as needing to be loaded again. 

The intention is that Chatterino instances with many open splits but with only a few being used at any moment will no longer hold Image data for emotes that aren't used in the active splits. This has no effect on emotes shared between splits.

These duration values are specified here: https://github.com/dnsge/chatterino2/blob/b35071aee8511aa87155e70184254aec51b14013/src/messages/Image.cpp#L28-L31

Addresses #3062

# Memory Usage Graph
The following graph's lines indicate:
1. Opening the emote popup for a single channel
2. Closing the emote popup
4. The automatic expired image frame data cleanup occurring (at shorter expiration for testing)

Screen Shot 2022-08-09 at 11 27 11 PM

Utilized memory goes from 154.74 MB before cleanup to 88.57 MB after.


src/messages/Image.cpp
```diff
@@ -33,8 +38,8 @@ namespace detail {
         DebugCount::increase("images");
     }
 
-    Frames::Frames(const QVector<Frame<QPixmap>> &frames)
-        : items_(frames)
+    Frames::Frames(QVector<Frame<QPixmap>> &&frames)
+        : items_(std::move(frames))
```
pajlada, line 42: For reference, QVector is one of the `implicit shared` classes of Qt where the class itself doesn't store the data, but rather essentially a shared_ptr to data, so copying just adds a reference to the underlying data https://doc-snapshots.qt.io/qt5-5.15/implicit-sharing.html
Either way, I'm fine with this change

pajlada:
No big feedback yet, but:
The `allImages` fields contains every single emote and emoji, which means the iteration we do happens for all ~6k images. Could they be added only when they are actually loaded (i.e. from the `Image::load` function)? Could they also then be removed from the list again since they would be re-added whenever `Image::load` is called next?

Images also seem to load as soon as they are created, even if the split isn't visible. Probably not relevant for this PR, but this PR has made that fact very clear.

src/messages/Image.cpp
dnsge: I see, I was wondering why we were using QVector at all

pajlada:
I want to take another look at the code before I merge this in, but I've been running this branch for 2 days without any crashes/freezes

src/messages/Image.hpp
```diff
@@ -80,13 +85,57 @@ class Image : public std::enable_shared_from_this<Image>, boost::noncopyable
...
+private:
+    friend class Image;
+
+    ImageExpirationPool();
+    static ImageExpirationPool &instance();
+
+    /**
+     * @brief Stores a reference to the given Image in the ImagePool.
+     * 
+     * The caller is responsible that the pointer will be valid for as long as
```
fourtf, line 115: Please remove these doc comments. The description given about "The caller is responsible that the pointer" is already incorrect. The function itself is self explanatory.

src/messages/Image.cpp
```diff
@@ -459,6 +493,13 @@ void Image::actuallyLoad()
         .execute();
 }
 
+void Image::expireFrames()
+{
+    assertInGuiThread();
+    this->frames_->clear();
```
fourtf, line 499: Why add `Frames::clear` when we can reset the unique_ptr instead?

src/messages/Image.cpp
```diff
@@ -211,14 +238,8 @@ namespace detail {
...
             // put into stack
-            static std::queue<std::pair<Assign, QVector<Frame<QPixmap>>>>
-                queued;
-            static std::mutex mutex;
-
             std::lock_guard<std::mutex> lock(mutex);
-            queued.emplace(assign, frames);
-
-            static std::atomic_bool loadedEventQueued{false};
+            queued.push(std::make_pair(assign, std::move(frames)));
```
fourtf, line 242: why remove use of `emplace`?

src/messages/Image.cpp
```diff
@@ -476,4 +517,97 @@ bool Image::operator!=(const Image &other) const
...
+ImageExpirationPool &ImageExpirationPool::instance()
+{
+    static ImageExpirationPool instance;
+    return instance;
+}
+
+void ImageExpirationPool::addImagePtr(ImagePtr imgPtr)
+{
+    std::lock_guard<std::mutex> lock(this->mutex_);
+    this->allImages_.insert(
```
fourtf, line 549: Why not use `this->allImages_.emplace(
        imgPtr.get(), std::weak_ptr<Image>(imgPtr));`?

fourtf (dismissed):
Looks good to me apart from some small things

src/messages/Image.cpp
dnsge: I didn't want to call `reset` because `this->frames_` is always expected to be non-null in the case of non-empty images. I also didn't want to do `this->frames_ = std::make_unique<detail::Frames>()` because it does an unnecessary allocation that will just be overwritten when we load the image again.

pajlada:
Ignore clang-tidy for now :dagger:

There seems to be a "leak" in the `images` debugcount - you can reproduce this by opening Chatterino, press F10 to open the Debug popup, then type an emote in a split. It will increase the `images` counter by 1.
Type the same emote again, it will increase the `images` counter by 1 again. Every time the emote is written, the count is increased.

dnsge:
> There seems to be a "leak" in the images debugcount

I found the issue, it wasn't actually due to incorrectly keeping track of the number of images. Instead, every message would have the reply button image pixmap created and appended, regardless of whether the setting is enabled (this is correct behavior).

https://github.com/dnsge/chatterino2/blob/20c974fdab215c54297cc5a47ea7d1e03c6065be/src/providers/twitch/TwitchMessageBuilder.cpp#L346-L350

I added caching to creating `Image` instances from resource pixmaps so that we reuse the same `Image` for the same pixmap and scale.

dnsge:
By the way, commit 2dd96d2 adds the usage of `<boost/functional/hash.hpp>`. I have no clue about vcpkg, but I remember that bringing in `circular_buffer` required adding some vcpkg entry (#3941). It would be good if someone could check if this new include requires an entry too.

src/messages/Image.cpp
```diff
@@ -268,13 +304,37 @@ ImagePtr Image::fromUrl(const Url &url, qreal scale)
...
+        cache;
+    static std::mutex mutex;
+
+    std::lock_guard<std::mutex> lock(mutex);
+
+    auto it = cache.find({&pixmap, scale});
+    if (it != cache.end())
+    {
+        auto shared = it->second.lock();
+        if (shared)
```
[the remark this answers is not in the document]
Mm2PL: wat

pajlada (approved):

PR #3921, opened 2022-08-12
Fix usage of FrankerFaceZ global emote API

### LINKED ISSUE #1116: FFZ Global emote sets API parsed incorrectly
```
curl https://api.frankerfacez.com/v1/set/global
```

```json
{
  "default_sets": [
    3
  ],
  "sets": {
    "3": {
      "_type": 0,
      "css": null,
      "description": null,
      "emoticons": [
        {
          "css": null,
          "height": 30,
          "hidden": true,
          "id": 9,
          "margins": null,
          "modifier": false,
          "name": "ZrehplaR",
          "offset": null,
          "owner": {
            "_id": 2,
            "display_name": "dansalvato",
            "name": "dansalvato"
          },
          "public": false,
          "urls": {
            "1": "//cdn.frankerfacez.com/4556987ef91323110080b223f96d7400.png"
          },
          "width": 33
        },
        ...
      ],
      "icon": null,
      "id": 3,
      "title": "Global Emotes"
    },
    "4330": {
      "_type": 0,
      "css": null,
      "description": null,
      "emoticons": [
        ...
      ],
      "icon": null,
      "id": 4330,
      "title": ": Sten's Cheaty Emotes"
    }
  },
  "users": {
    "4330": [
      "sirstendec"
    ]
  }
}
```

The global API endpoint specifies a number of sets (3 and 4330 in this case) but only 3 is global as specified by the "global" array. For example this causes the AndKnuckles emote to be shown in chatterino when it really isn't a global emote and will not show up for web chat users.

pajlada:
Pull request checklist:

# Description

We no longer blindly parse all sets as global emotes, but rather match them against the `default_sets` as intended.

This means that some emotes will no longer be visible through Chatterino (e.g. AndKnuckles). This is more in line with how the FrankerFaceZ browser extension works.

Fixes #1116


Mm2PL:
Sadge . o 0 ( AndKnuckles )

pajlada:
Looks good to me!
image

PR #3923, opened 2022-08-14
Added minimal supported macOS version to CI builds.

23rd:
Pull request checklist:

# Description

Recently, nightly builds stopped working on macOS with outdated versions (I guess this is caused by the `macos-latest` being `macos-11.0`).

<details><summary>Popup</summary>
image
</details>

That makes me sad.
Let's support macOS Catalina for a while.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -197,6 +197,7 @@ jobs:
             cd build
             cmake \
                 -DCMAKE_BUILD_TYPE=Release \
+                -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15 \
                 -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl \
                 -DUSE_PRECOMPILED_HEADERS=${{ matrix.pch }} \
                 ..


pajlada:
Does this make the builds stop working on newer macos systems? If not then this is fine by me for now.
Long term, I suppose expanding the number of macos builds would be the way to go but I'd rather not do that until we've refactored the build workflows

jammehcow:
It should be noticed that macOS Catalina (10.5) is marked a deprecated on GitHub's runner page. I'd rather we didn't support deprecated images but I understand that a fair chunk of users still use 10.5. 

Source: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

23rd:
> Does this make the builds stop working on newer macos systems?

A result binary (from CI) works on 12.5 as well.

> It should be noticed that macOS Catalina (10.5) is marked a deprecated on GitHub's runner page.

My PR does not change a version of the macOS runner, it is still `macos-latest`.

jammehcow:
> My PR does not change a version of the macOS runner, it is still `macos-latest`.

Misread, sorry!

pajlada (approved):
Merging this in as in as a stop-gap solution

Would be nice with a support matrix for what versions of OSes we support, not sure if it makes sense to keep in the wiki or the repo.

like

| os|support|
|---|---|
| macos montgomery (12.5) | :heavy_check_mark: |
| macos scooby (11.7) | :heavy_check_mark:*<sup>requires manual installation of libXD</sup> |

PR #3926, opened 2022-08-15
Allow hiding moderation actions in streamer mode

### LINKED ISSUE #2729: Show moderation messages part of streamer settings
Seeing messages that get deleted by mods might not be something you want to see while streaming or showing on chatterino during stream so it would make sense to make this a streamer mode setting rather than normal setting.

(The template for enhancements is gone)

Mm2PL:
Pull request checklist:

# Description

This adds a new setting to streamer mode which will hide any moderation messages. This doesn't hide moderated messages, that could be another PR. This hides any messages which have the Timeout or Untimeout flags, exactly like the Show moderation messages setting. Also of note: this won't hide timeout responses for commands you run.
Closes #2729.


iProdigy (approved):

src/singletons/Settings.hpp
```diff
@@ -221,6 +221,8 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting streamerModeMuteMentions = {"/streamerMode/muteMentions", true};
     BoolSetting streamerModeSuppressLiveNotifications = {
         "/streamerMode/supressLiveNotifications", false};
+    BoolSetting streamerModeSuppressModActions = {
+        "/streamerMode/streamerModeSuppressModActions", true};
```
zneix, line 225: why is `streamerMode` twice there, where's consistency? Everything around these lines of code has it just as a category, not also as a prefix later on.

src/messages/layouts/MessageLayout.cpp
```diff
@@ -139,11 +140,15 @@ void MessageLayout::actuallyLayout(int width, MessageElementFlags flags)
...
+        if (this->message_->flags.has(MessageFlag::Timeout) ||
+            this->message_->flags.has(MessageFlag::Untimeout))
         {
-            continue;
+            if (getSettings()->hideModerationActions ||
+                (getSettings()->streamerModeSuppressModActions &&
+                 isInStreamerMode()))
+            {
+                continue;
+            }
```
zneix, line 151: is the change of prioritising either check for relevant settings or check for relevant message flags intended? either way I guess it's okay since checking flags seems less expensive than calling settings instance

zneix (dismissed):
I'm surprised someone actually bothered requesting my - review on anything... few last merges definitely show like you don't need me here

people just approve changes like this here without even paying attention to details, while also disregarding me or my opinion, then complain about quirks that exist in master because "somehow that mistake got merged in.." 🙄

src/messages/layouts/MessageLayout.cpp
iProdigy: this order seems preferable to potentially avoid an expensive `isInStreamerMode()` lookup (on a cache miss)

jammehcow:
> I'm surprised someone actually bothered requesting my - review on anything... few last merges definitely show like you don't need me here
> 
> people just approve changes like this here without even paying attention to details, while also disregarding me or my opinion, then complain about quirks that exist in master because "somehow that mistake got merged in.." 🙄

@zneix This comment has nothing to do with the review of this code. Mistakes happen, we fix them and move on - we're human. It's also inconsiderate to the other maintainers/developers of chatterino to claim they don't pay attention to the details when reviewing additions to the project. I'm sure they do in fact care about the quality of what's going in to the codebase as does the community because they care about the software they use. 

You chose to step away from the project for a bit which is understandable and respectable - but to come back and claim that your opinion isn't respected by the community or that they're actively disregarding your input is ignorant of the fact that you made that choice to back off. 

I personally find your skills and knowledge of chatterino's domain valuable. The drama however, is annoying and doesn't contribute anything positive to the developer community. I would encourage you to think about how you're impacting the community by exhibiting this sort of behaviour and grow from it.

src/messages/layouts/MessageLayout.cpp
Mm2PL: Yeah, I made this order to be somewhat readable and avoid calling isInStreamerMode if not needed. I'll add a comment explaining why this is.

src/singletons/Settings.hpp
Mm2PL: I think my completion changed this and I didn't notice. Fixed in [fbe9799](https://github.com/Chatterino/chatterino2/pull/3926/commits/fbe97999e96f3e24518bce31c73999c5b59c7450).

pajlada (approved):

src/singletons/Settings.hpp
```diff
@@ -247,6 +247,8 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/streamerMode/supressLiveNotifications", false};
     BoolSetting streamerModeSuppressInlineWhispers = {
         "/streamerMode/suppressInlineWhispers", true};
+    BoolSetting streamerModeSuppressModActions = {
+        "/streamerMode/suppressModActions", true};
```
zneix, line 251: where is consistency with the changelog & changes made to `GeneralPage.cpp`? I suggest to either keep suppress or hide everywhere. ~~really surprised nobody even cared/noticed~~
```suggestion
    BoolSetting streamerModeHideModActions = {
        "/streamerMode/hideModActions", true};
```

zneix (changes requested):

src/singletons/Settings.hpp
Mm2PL: Fair point, changed in f1c0d79.

zneix (approved):

PR #3932, opened 2022-08-20
Fix crash when adding recent messages to empty Channel

### LINKED ISSUE #3930: Crash when reloading message history
### Checklist

### Steps to Reproduce
1. Reconnect due to either an account switch or Twitch deploy
2. crash

stacktrace:
https://paste.ivr.fi/wigozyxoge.rs
Points to https://github.com/Chatterino/chatterino2/blob/8ea5b49/src/common/Channel.cpp#L262 `corrected by dnsge`

~~My assumption is that it's just not liking how many channels need to be reconnected to at once, as I was unable to replicate the issue when just using `-c pajlada` or a short list of channels.~~ see below comment

(note): This crash happened sometime after the first set of changes, as I had a user testing the PR as of 4a7d1cc and they had no problems or crashes, but switching to the recent nightly causes them to crash on reconnects as well.
PR: Chatterino/chatterino2/pull/3878

### Screenshots

image

### OS and Chatterino Version

Chatterino Nightly 2.3.5 (commit 8ea5b49c modified) built on 2022-08-14 with Qt 5.15.2, Windows SDK, MSVC 193231332 Running on Windows 10 Version 2009, kernel: 10.0.19043

dnsge:
Pull request checklist:

# Description

When loading recent messages upon reconnecting, we would blindly assume that the channel already had at least one message in it. However, in the case that it somehow doesn't, we would index out of bounds of our `snapshot` and crash. 

This PR adds checks for empty channels (and empty loaded recent messages).

Fixes #3930

(Note: I haven't verified this _actually_ fixes the crash because I can't reproduce it on my own. However, the stack trace provided by Felanbird strongly suggests this should fix the crash).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 - Minor: Removed total views from the usercard, as Twitch no longer updates the number. (#3792)
 - Minor: Add Quick Switcher item to open a channel in a new popup window. (#3828)
 - Minor: Warn when parsing an environment variable fails. (#3904)
-- Minor: Load missing messages from Recent Messages API upon reconnecting (#3878)
+- Minor: Load missing messages from Recent Messages API upon reconnecting (#3878, #3932)
 - Bugfix: Fix crash that can occur when closing and quickly reopening a split, then running a command. (#3852)
 - Bugfix: Connection to Twitch PubSub now recovers more reliably. (#3643, #3716)
 - Bugfix: Fix crash that can occur when changing channels. (#3799)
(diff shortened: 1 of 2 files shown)


iProdigy (approved):

Felanbird (commented):
Crash is fixed 👍 

I'm going to give it a quick once over on an artifact once it builds on my fork, as QtCreator just freezes during the account switch (but that happens often when lots of stuff is happening).

Felanbird (approved):

pajlada (approved):

PR #3943, opened 2022-09-03
Remove unused mutex from Emotes

pajlada:
Pull request checklist:

# Description

The mutex was initially used to limit access to the twitchEmotesCache_ member
but it's no longer necessary since it's been made a UniqueAccess type


pajlada:
Looks good to me
image

Issue #3861, opened 2022-07-14
Cannot search channel history when opening a duplicate tab of an existing channel


Felanbird:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Steps to Reproduce
1. Open a channel (or already have one open)
2. Open a duplicate tab/split of said channel
3. Open the search window

Only messages sent after you opened the tab will be searchable.

Note: Global search is not affected by this, which I personally had been using so I didn't run into this issue immediately.

### Screenshots

![qtcreator_2022-07-13_20-26-59](https://user-images.githubusercontent.com/41973452/178859305-678c8c9d-9430-40d8-a1ce-d97800890ca1.gif)

### OS and Chatterino Version

#3694 - Cannot replicate 1 commit back, issue still persists on current master 965b066

Felanbird:
This was fixed probably by #4742 or something idk glad it's fixed it sucked

Felanbird:
I lied this isn't fixed, I was testing it with channels that already got history populated due to being open on startup, not a dupe specifically.

PR #3953, opened 2022-09-07
feat: add `Go to message` action in various places

Nerixyz:
Pull request checklist:

# Description
When searching a message, one might want to see it in context to other messages in chat. That's why this PR adds a `Go to message` action to the search context menu.

## Places

- [x] Search window
  * [x] Regular
  * [x] Multiple channels
  * Should the search window close after clicking the action?
- [x] `/mentions`
- [x] User-Card
  * [x] From mentions
  * [x] From reply-thread
  * [x] From search
  * [x] From search in mentions
  * [x] From multisearch
- [x] Reply-Threads
  * [x] From mentions
  * [x] From user-card
  * [x] From search
  * [x] From search in mentions
  * [x] From multisearch
  * [x] From user-card in search in mentions (?)
- [x] Deleted Messages (link to original message)
  * [x] From search
- ~~If a channel has filters enabled, should there be a `Go to message` as well? For example: one view is the regular chat, one only contains the mod messages.~~


Felanbird:
Quick immediate thought(s) based on title alone:
- How will this handle multiple splits of the same channel? (global search)

  - Would the fixing of #3861 allow for the approach where the currently focused tab is prioritized, but not cause issues when the duplicate tab is freshly opened. 
 `e.g`: Using /mentions context (maybe noticing it after 5min), opening a new tab with a duplicate split of said channel. attempting to use that new split to search for the message and ending up with a blank search.

dnsge:
It would be great to add this functionality to user cards and reply threads, too. 

One small problem that I noticed: if you open a split with not enough messages to scroll, selecting "Go to message" will force scroll the channel view to scroll. Mousing over/doing something to update the channel view will then make it realize that it shouldn't be scrolled, making it jump back.

Nerixyz:
> * How will this handle multiple splits of the same channel? (global search)

All channels that contain this message will scroll to the message. `ChannelView::scrollToMessage` should probably include a check that early-outs if the message cannot possibly be in the channel to avoid searching the entire history.

> * Would the fixing of #3861 allow for the approach where the currently focused tab is prioritized, but not cause issues when the duplicate tab is freshly opened?

There could be a check to only go to a message if the tab is focused if I understand the question correctly.

> if you open a split with not enough messages to scroll, selecting "Go to message" will force scroll the channel view to scroll.

Fixed in 4a9ecb4a85a84fdf83f80d07ad84b2ed994de449.

> It would be great to add this functionality to user cards and reply threads, too.

I agree; added _reply-threads_ to my original comment. Though I don't know if that should happen in this PR.

CHANGELOG.md
```diff
@@ -22,7 +22,8 @@
 - Minor: Add scrollbar to `Select filters` dialog. (#3737)
 - Minor: Added `/requests` command. Usage: `/requests [channel]`. Opens the channel points requests queue for the provided channel or the current channel if no input is provided. (#3746)
 - Minor: Added ability to execute commands on chat messages using the message context menu. (#3738, #3765)
-- Minor: Added `/copy` command. Usage: `/copy <text>`. Copies provided text to clipboard - can be useful with custom commands. (#3763)
+- Minor: Added `/copy` command. Usage: `/copy <text>`. Copies provided text to clipboard - can be useful with custom
+  commands. (#3763)
```
Felanbird, line 26: 🤔

Nerixyz: Oops I accidentally hit enter .

src/widgets/helper/SearchPopup.cpp
```diff
@@ -106,6 +106,14 @@ void SearchPopup::addChannel(ChannelView &channel)
     this->updateWindowTitle();
 }
 
+void SearchPopup::goToMessage(const MessagePtr &message)
+{
+    for (const auto &channel : this->searchChannels_)
+    {
+        channel.get().scrollToMessage(message);
```
pajlada, line 113: Early out

src/widgets/helper/ChannelView.hpp
```diff
@@ -208,6 +227,12 @@ class ChannelView final : public BaseWidget
     void enableScrolling(const QPointF &scrollStart);
     void disableScrolling();
 
+    /**
+     * @param layout Must be from this channel.
+     * @param messageIdx Must be an index into this channel.
+     */
+    void scrollToMessage(MessageLayout *layout, size_t messageIdx);
```
pajlada, line 234: Rename to something more internally appropriate, maybe something like `scrollToMessageInLayout`

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
 - Minor: Add settings to toggle BTTV/FFZ global/channel emotes (#3935)
 - Minor: Add AutoMod message flag filter. (#3938)
 - Minor: Added whitespace trim to username field in nicknames (#3946)
+- Minor: Added `Go to message` context menu action to search popup, mentions, user-card and reply-threads (#3953)
+- Minor: Added link to deleted message (#3953)
```
pajlada, line 36: Explain the "deleted message" changelog entry a bit more verbosely

pajlada (changes requested):
Currently doesn't switch tabs if the message/channel is found in another tab

Other than that, animation and functionality seems great!

CHANGELOG.md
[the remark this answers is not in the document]
Nerixyz: Done in 20fd8ba2d08547c72b3087ad05df13ca4af4d9be.

src/messages/Link.hpp
```diff
@@ -25,6 +25,7 @@ struct Link {
         CopyToClipboard,
         ReplyToMessage,
         ViewThread,
+        MessageId,
```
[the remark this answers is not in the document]
Nerixyz: Done in d78503937fa0330c23f6326672ce2c64d008256c.

src/singletons/WindowManager.cpp
```diff
@@ -320,6 +320,11 @@ void WindowManager::select(SplitContainer *container)
     this->selectSplitContainer.invoke(container);
 }
 
+void WindowManager::selectAndScrollToMessage(const MessagePtr &message)
```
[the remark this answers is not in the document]
Nerixyz: Done in 51d231f5099ce8293f0fb169fad3966fcf0cf43b.

src/widgets/helper/SearchPopup.cpp
Nerixyz: Done in 4ed3937dc6e3351be95858a887720418147b5a32.

src/widgets/helper/ChannelView.hpp
Nerixyz: Done in c8325c32479116991f940b240be69706344abac0.

src/widgets/Notebook.cpp
```diff
@@ -1006,6 +1007,29 @@ SplitNotebook::SplitNotebook(Window *parent)
...
+            {
+                if (auto sc = dynamic_cast<SplitContainer *>(item.page))
+                {
+                    auto &&splits = sc->getSplits();
+                    auto split = std::find_if(
+                        splits.begin(), splits.end(), [message](Split *split) {
+                            return split->getChannel()->getType() !=
+                                       Channel::Type::TwitchMentions &&
+                                   split->getChannelView().scrollToMessage(
+                                       message);
```
pajlada, line 1024: Convert this to a for loop instead, remember you'll need to return to use `return` since `break` will only break the inner loop

src/widgets/helper/ChannelView.cpp
```diff
@@ -339,9 +359,26 @@ void ChannelView::themeChangedEvent()
...
+    {
+        this->highlightAnimation_.setStartValue(QColor(20, 20, 20, maxOpacity));
+        this->highlightAnimation_.setEndValue(QColor(20, 20, 20, 0));
+    }
+    else
+    {
+        this->highlightAnimation_.setStartValue(
+            QColor(230, 230, 230, maxOpacity));
+        this->highlightAnimation_.setEndValue(QColor(230, 230, 230, 0));
+    }
```
pajlada, line 379: Instead of checking for light theme, define two colors (`highlightAnimationStart` and `highlightAnimationEnd`) in the theme class that does the light theme checking

That way this class would not need a branch, and just reference the two new colors

src/widgets/helper/ChannelView.hpp
```diff
@@ -83,6 +84,17 @@ class ChannelView final : public BaseWidget
...
+    /**
+     * Attempts to scroll to a message in this channel.
+     * @return <code>true</code> iff the message was found and highlighted.
+     */
+    bool scrollToMessage(const MessagePtr &message);
+    /**
+     * Attempts to scroll to a message id in this channel.
+     * @return <code>true</code> iff the message was found and highlighted.
+     */
+    bool scrollToMessageId(const QString &id);
```
pajlada, line 96: Just use `if` instead of `iff` here

pajlada (changes requested):

src/widgets/Notebook.cpp
Nerixyz: Done in 4798ed67d9a77be42a2afd01c74bc3a58cacee37.

src/widgets/helper/ChannelView.cpp
Nerixyz: Done in aed9164c0b420ce7c937f21507db91e3bf7cef5f.

src/widgets/helper/ChannelView.hpp
Nerixyz: Done in df19120fe14c32ff7f125e48e8ceda399edd01d5.

src/widgets/Notebook.cpp
```diff
@@ -1015,17 +1015,14 @@ SplitNotebook::SplitNotebook(Window *parent)
...
-                        });
-                    if (split != splits.end())
+                    for (auto *split : sc->getSplits())
                     {
-                        break;
+                        if (split->getChannel()->getType() !=
+                            Channel::Type::TwitchMentions)
+                        {
+                            split->getChannelView().scrollToMessage(message);
+                            return;
```
pajlada, line 1024: Should only return if `scrollToMessage` returns true

pajlada (changes requested):

src/widgets/Notebook.cpp
Nerixyz: Done in 8ed335105a8d3c07791d3251281bb546b78dd2bb.

pajlada (approved):

PR #3958, opened 2022-09-08
Implement initial support for RTL languages

### LINKED ISSUE #720: Problem with Languages using the Arabic Script - Hebrew and Arabic are reversed

mohad12211:
Pull request checklist:

# Description

image

This is a fix #720 for RTL languages.

- [x] line breaks work.
- [x] Neutral words (words that contains numbers or non-alphabetical characters like `& $ # ` etc...) are ordered properly
- [x] punctuation marks are displayed properly.
- [x] mixing RTL and LTR works (the correct way, not Twitch's way)
- [x] ~reply messages are handled properly.~ will be another PR, we need to reorder words when creating the reply element due to #4067 
- [x] ~Selection do not work yet.~ won't fix for now, use (right click -> copy message) as a workaround.


src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -272,6 +272,55 @@
...
+            }
+        }
+    }
+
+    std::vector<int> swapped;
+    std::stack<int> rtlStack;
+
+    for (int i = startIndex; i <= endIndex; i++)
+    {
+        if (this->elements_[i]->getText().isRightToLeft())
```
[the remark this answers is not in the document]
Mm2PL: clang-tidy, what the fuck are you on

mohad12211:
The last commit should show proper mixing between RTL and LTR, Twitch is doing it the wrong way actually.
If the text starts with an RTL word (e.g. Arabic), with LTR in the middle, it will only swap the RTL sequences, which is wrong, and a frustrating issue on Twitch... Should be working here. And it makes emotes show in the correct way as well.
Example: 
image

Felanbird:
The artifacts to test these changes can be found here, incase any issues can be found by native RTL speakers.

image

EndlessJest:
thank you so much! Everything seems to work correctly so far.

Felanbird:
From a functionality standpoint, it appears these changes push the "space" when using large emotes to the bottom, instead of the top of the message.

before:
image

after:
image

mohad12211:
@Felanbird yes that was due to a weird thing I did, fixed after this refactor.

mohad12211:
I refactored the fix a bit.

What is done is basically this:
we call  `_addElement` normally, once we if we detect an RTL word in the container, we reorder the elements.
we call `addElement` on each element again, without adding the element to `this->elements_`, if the text starts with an RTL word, we render from right to left (so `currentX_` will decrease instead of increase), otherwise render normally since we are calling `_addElement` in the correct order. since we need the previous element for Compact-Emotes, I pass the index for the previous element, I pass -2 by default, so that in the first call we just do it the old way, the second call it will start from -1.
I think this is better than making two `_addElement` functions, one for RTL and one for LTR with 90% similar code.
let me know if I can do it in a better way.

Felanbird:
image
Top portion is this PR - bottom is normal chatterino incorrect rendering

mohad12211:
Everything should be working now except selection, copying the entire message works for now as an alternative to selection.
I will see what I can do about selection soon.
Please test this as much as possible, preferably with other RTL languages other than Arabic.
To download the app with this changes. go to Checks, at the top, then build, then at the bottom you will find binaries for all operating systems:
image
image

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -184,22 +222,138 @@ void MessageLayoutContainer::_addElement(MessageLayoutElement *element,
...
 void MessageLayoutContainer::breakLine()
 {
+    if (this->containsRTL)
+    {
+        if (this->elements_.size() > 0 &&
+            this->elements_[0]->getFlags().has(
+                MessageElementFlag::RepliedMessage) &&
+            this->line_ == 0)
+        {
+            this->reorderRTL(4);
```
pajlada, line 339: Add comment here for what the 4 means

src/messages/layouts/MessageLayoutContainer.hpp
```diff
@@ -86,7 +91,8 @@ struct MessageLayoutContainer {
     };
 
     // helpers
-    void _addElement(MessageLayoutElement *element, bool forceAdd = false);
+    void _addElement(MessageLayoutElement *element, bool forceAdd = false,
```
pajlada, line 94: Could you incorporate your comment from here https://github.com/Chatterino/chatterino2/pull/3958#issuecomment-1249849323 to this function?

src/messages/layouts/MessageLayoutContainer.hpp
```diff
@@ -60,6 +64,7 @@ struct MessageLayoutContainer {
     void breakLine();
     bool atStartOfLine();
     bool fitsInLine(int width_);
+    void reorderRTL(int firstTextIndex);
```
pajlada, line 67: Brief comment here how the function achieves its goal, and what `firstTextIndex` is used for

src/util/Helpers.cpp
```diff
@@ -123,6 +124,13 @@ bool startsWithOrContains(const QString &str1, const QString &str2,
     return str1.contains(str2, caseSensitivity);
 }
 
+bool isNeutral(const QString &s)
+{
+    const QRegularExpression re("\\p{L}");
```
pajlada, line 129: Can make this static

```suggestion
    static const QRegularExpression re("\\p{L}");
```

pajlada (changes requested):

mohad12211:
ready for final review & merge.

pajlada:
Looks like this for me (left = your PR, right = master)
image

Testing with this text: `a یک دو سه b`

pajlada:
When switching to Segoe UI it looks correct to me

image

pajlada:
I attribute the previous errors I ran into to bad fonts on Linux

Felanbird (approved):
As far as I believe I can test this as someone who does not speak/write an RTL language, I believe it works.

pajlada (approved):
Thank you!!

Felanbird:
@mohad12211 Thanks for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

To do so, open a new PR where you modify the resources/contributors.txt file and add yourself to the list. Make sure to read the comments at a top for instructions.

PR #3988, opened 2022-09-11
Migrate /color command to Helix API

### LINKED ISSUE #3963: Migrate /color command to Helix API
As part of Twitch's announced deprecation of IRC-based commands ([see here for more info](https://discuss.dev.twitch.tv/t/deprecation-of-chat-commands-through-irc/40486), the `/color` command needs to be migrated to use the relevant Helix API endpoint.

Helix API reference: https://dev.twitch.tv/docs/api/reference#update-user-chat-color

Split from #3956

pajlada:
Pull request checklist:

# Description

This PR migrates the `/color` chat command to use the Helix API.
This migration does not necessitate any time-gating as the functionality is essentially the same except:
 - ~~The pre-defined colors have changed names~~ This is no longer a concern to users as we accept both the old and new names. We will, however, only show the new names to users.
 - The user has to re-authenticate to get the required scope on their token.
 
In addition to migrate the command, I've also added some documentation for how to add support for new Helix endpoints in the `/providers/twitch/api/README.md` file

Fixes #3963

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -1238,8 +1238,8 @@ void CommandController::initialize(Settings &, Paths &paths)
                 {
                     case HelixUpdateUserChatColorError::UserMissingScope: {
                         errorMessage +=
-                            "Missing required scope. Reauthenticate with your "
-                            "user and try again.";
+                            "Missing required scope. Re-login with your "
+                            "account and try again.";
                     }
                     break;
 
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -2,6 +2,7 @@
 
 #include "Application.hpp"
 #include "common/Env.hpp"
+#include "common/QLogging.hpp"
 #include "common/SignalVector.hpp"
 #include "controllers/accounts/AccountController.hpp"
 #include "controllers/commands/Command.hpp"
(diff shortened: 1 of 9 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -1195,6 +1196,77 @@ void CommandController::initialize(Settings &, Paths &paths)
...
         });
+
+    this->registerCommand("/color", [](const QStringList &words, auto channel) {
+        auto user = getApp()->accounts->twitch.getCurrent();
+
+        // Avoid Helix calls without Client ID and/or OAuth Token
+        if (user->isAnon())
+        {
+            channel->addMessage(makeSystemMessage(
+                "You must be logged in to use the /color command"));
```
Felanbird, line 1207:
```suggestion
                "You must be logged in to use the /color command."));
```

Felanbird (approved):
Works as expected 👍 
image

iProdigy:
Could add a [mapping](https://github.com/twitch4j/twitch4j/blob/e3501746680a1aa6ae655b119ac7519f7d4248d5/twitch4j/src/main/java/com/github/twitch4j/internal/ChatCommandHelixForwarder.java#L187-L188) of the (lowercase) irc color names to the helix versions to maintain compatibility

Felanbird:
> Could add a [mapping](https://github.com/twitch4j/twitch4j/blob/e3501746680a1aa6ae655b119ac7519f7d4248d5/twitch4j/src/main/java/com/github/twitch4j/internal/ChatCommandHelixForwarder.java#L187-L188) of the (lowercase) irc color names to the helix versions to maintain compatibility

I wouldn't be against this, I didn't even realize this PR was actually replacing the functionality, I instinctively just commented out the `/color` IRC command listing before I started testing.

pajlada:
PR has been updated to map the old naming scheme and allow any case

so the following all work:
```
/color golden_rod
/color Golden_Rod
/color GoldenRod
/color GOLDENROD
```

Unit tests have been added for this behaviour

src/controllers/commands/CommandController.cpp
```diff
@@ -1195,6 +1196,87 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        };
+
+        auto colorString = words.value(1);
+
+        if (colorString.isEmpty())
+        {
+            channel->addMessage(makeSystemMessage(
+                QString("Usage: /color <color> - Color must be one of Twitch's "
+                        "supported colors (%1) or a hex code (#000000) if you "
+                        "have Turbo or Prime.")
```
iProdigy, line 1226:
```suggestion
                        "have Turbo.")
```

Despite what the twitch docs say, non-turbo prime users aren't able to set a custom hex code

iProdigy (approved):
Nice work on supporting the old naming 👍

src/controllers/commands/CommandController.cpp
iProdigy: I tested this incorrectly; the command *does* work for prime & this is now an advertised feature of prime https://www.amazongames.com/en-us/support/prime-gaming/articles/what-is-prime-gaming

https://github.com/Chatterino/chatterino2/pull/3988/commits/fd8ae3f1cf1cd8039afe7a59755b32e5e6e875d2 can be reverted

src/controllers/commands/CommandController.cpp
```diff
@@ -1195,6 +1196,87 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        // Avoid Helix calls without Client ID and/or OAuth Token
+        if (user->isAnon())
+        {
+            channel->addMessage(makeSystemMessage(
+                "You must be logged in to use the /color command"));
+            return "";
+        }
+
+        // Colors retreived from https://dev.twitch.tv/docs/api/reference#update-user-chat-color 2022-09-11
+        const QStringList validColors{
```
Mm2PL, line 1212: Why do we have these values stored twice? Other one is in `src/util/Twitch.cpp`. Would make sense to only have it once if twitch ever decides to ~~break~~ change something like they like to.

Mm2PL (changes requested):
One tiny thing.

Mm2PL (approved):

src/controllers/commands/CommandController.cpp
pajlada: The two containers serve unique purposes - for clarity I've added comments to both and moved the `validColors` container to `src/util/Twitch.cpp`

Felanbird (approved):
image

PR #3989, opened 2022-09-13
fix: retain text from input when replying

Nerixyz:
Pull request checklist:

# Description

When clicking `Reply to message` the input text is replaced with `@{displayName} `, this PR attempts to retain any input that was typed before (i.e. the text is replaced with `@{displayName} {input} `).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Minor: Added whitespace trim to username field in nicknames (#3946)
 - Minor: Added `Go to message` context menu action to search popup, mentions, usercard and reply threads. (#3953)
 - Minor: Added link back to original message that was deleted. (#3953)
+- Minor: Input text is retained when replying to a message. (#3989)
 - Bugfix: Fix crash that can occur when closing and quickly reopening a split, then running a command. (#3852)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 - Minor: Added whitespace trim to username field in nicknames (#3946)
 - Minor: Added `Go to message` context menu action to search popup, mentions, usercard and reply threads. (#3953)
 - Minor: Added link back to original message that was deleted. (#3953)
+- Minor: Input text is retained when replying to a message. (#3989)
```
dnsge, line 37: We could probably add this to the existing reply threads changelog entry with 3722.

Felanbird: Agreed

Nerixyz: True. I'd say this doesn't really need an entry at all since there isn't a release between this PR and the original one.

pajlada (approved):

CHANGELOG.md
pajlada: I went with adding this PR to the existing changelog entry

PR #3994, opened 2022-09-17
Migrate /clear command to Helix API

### LINKED ISSUE #3962: Migrate /clear command to Helix API
As part of Twitch's announced deprecation of IRC-based commands ([see here for more info](https://discuss.dev.twitch.tv/t/deprecation-of-chat-commands-through-irc/40486), the `/clear` command needs to be migrated to use the relevant Helix API endpoint.

Helix API reference: https://dev.twitch.tv/docs/api/reference#delete-chat-messages

Split from #3956

aidenwallis:
Pull request checklist:

# Description

- Closes #3962 

- Implemented a new Helix func `deleteChatMessages` - this can be used for both `/delete` and `/clear `commands as they call the same Twitch Helix endpoint under the hood.

- Adds command handler for `/clear`, defined as a lambda function as I will come back in a followup PR to hook up the same lambda to `/delete` too.

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -1270,7 +1270,7 @@ void CommandController::initialize(Settings &, Paths &paths)
         return "";
     });
 
-    auto deleteMessagesLambda = [](auto channel, const QString &messageID) {
+    auto deleteMessages = [](auto channel, const QString &messageID) {
         const auto *commandName = messageID.isEmpty() ? "/clear" : "/delete";
         auto *twitchChannel = dynamic_cast<TwitchChannel *>(channel.get());
         if (twitchChannel == nullptr)
@@ -1351,10 +1351,9 @@ void CommandController::initialize(Settings &, Paths &paths)
     };
 
     this->registerCommand(
-        "/clear",
-        [deleteMessagesLambda](const QStringList &words, auto channel) {
+        "/clear", [deleteMessages](const QStringList &words, auto channel) {
             (void)words;  // unused
-            deleteMessagesLambda(channel, QString());
+            deleteMessages(channel, QString());
             return "";
         });
 }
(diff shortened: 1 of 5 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -1269,6 +1269,93 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+
+                    case HelixDeleteChatMessagesError::MessageUnavailable: {
+                        // Override default message prefix to match with IRC message format
+                        errorMessage =
+                            QString(
+                                "The message %1 does not exist, was deleted, "
+                                "or is too old to be deleted.")
+                                .arg(messageID);
+                    }
+                    break;
```
aidenwallis, line 1328: This condition isn't met (yet), will be when I follow up with `/delete`, I could implement in that PR but it made sense to me to cover all possible enums in this one - and this enum should still exist in the API implementation as it's a possible response from the API endpoint (but should not happen with no messageID passed)

src/controllers/commands/CommandController.cpp
```diff
@@ -1269,6 +1269,93 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            },
+            [channel, messageID](auto error, auto message) {
+                QString errorMessage =
+                    QString("Failed to delete chat messages - ");
+
+                switch (error)
+                {
+                    case HelixDeleteChatMessagesError::UserMissingScope: {
+                        errorMessage +=
+                            "Missing required scope. Re-login with your "
```
aidenwallis, line 1309: Also double checked that client_login has the right scope for this functionality 👍

src/controllers/commands/CommandController.cpp
jammehcow: Introducing that enum value in this PR makes sense as it's a response that the API *could* return.

pajlada (approved):

pajlada:
Thanks for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

PR #3071, opened 2021-07-25
7TV Emote Provider Implementation

AnatoleAM:
Pull request checklist:

- [ ] `CHANGELOG.md` was updated, if applicable
- [x] Add an option to hide unlisted emotes
- [x] Adjust scaling of emotes
- [x] Support zero-width emotes
- [x] Animated Avatars ([chatterino7#69](https://github.com/SevenTV/chatterino7/pull/69))
- [x] Add the 4x CDN link in the right click context ([chatterino7#62](https://github.com/SevenTV/chatterino7/pull/62))
- [ ] Migrate to 7TV API v3
- [ ] Append instructions to building docs regarding WebP
- [ ] Use data-driven CDN URLs 
- [ ] Use the REST API instead of GQL for fetching emotes
- [ ] (Optional) Implement the [7TV EventAPI](https://github.com/SevenTV/EventAPI)
- [ ] (Optional) Implement Nametag Paints
- [ ] (Optional) Add support for images in AVIF format

# Description

Implementing 7TV (https://7tv.app) as a new Emote Provider in Chatterino2. The changes include adding a new provider to fetch global emotes, channel emotes and user badges from the 7TV API.

This pull request mirrors the [SevenTV/chatterino7](https://github.com/SevenTV/chatterino7) fork maintained by myself and 
@TroyKomodo, with contributions by @zneix, @goldbattle, @MrAuro, @mmattbtw and others, while backtracking a few changes (such as versioning and workflows) to upstream.

PR #3812, opened 2022-06-10
FFZ should explicit use 4x emote links, also fix scale factors

goldbattle:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

This fixes FFZ to have the correct scale for their 4x higher resolution emotes.
This is also in preparation for 7tv support which has higher resolution emotes.

Comments:
- Scale factor should be ration to the base size of 32px
- Example: 32 / 112 = 0.286 scale factor for 3x emotes
- Example: 32 / 128 = 0.250 scale factor for 4x emotes

PR #4002, opened 2022-09-19
feat: Add 7TV Emotes and Badges

Nerixyz:
Pull request checklist:

# Description

This PR adds 7TV emotes and badges to chatterino. It's mostly the code from SevenTv/Chatterino7 (7TV fork, see #3071) + an update to the REST v3 API.

Badges aren't available in v3 (see [comment](https://github.com/Chatterino/chatterino2/pull/4002#issuecomment-1251332743)) that's why we use the v2 REST API.

# TODO

-    ~~Figure out a good way to cache emotes (and their images). In contrast to BTTV and FFZ, 7TV is unique, because emotes can have aliases. The current approach is to cache all emotes that aren't aliased nor global but cache the images for these. (This needs more documentation)~~
- ~~Use REST v3 for Badges (not added yet)~~
- [x] Check global emotes once the endpoint is added
- [x] Is it `7TV`, `7Tv`, `SevenTV`, `SevenTv`, or `Seventv`?
-    ~~Should entire emote-sets be cached? I think we should observe the use of emote-sets and if we see many users use one emote-set, we can add caching later.~~
- [x] (the TODOs I added)
- [x] Proper message for channels without 7TV emotes

# Known Issues

- [ ] Zero Width emotes aren't actually zero-width (e.g. the gloabl `SteerR`). This isn't an issue with this PR but an issue with 7TV's REST API not setting the [`ActiveEmoteFlagModelZeroWidth`](https://github.com/SevenTV/API/blob/9069cc79c6c8896c1016947994a41e38657834cd/data/model/emote-set.model.go#L32) for active emotes.

# Goals for this PR

I want to stay as close to SevenTv/Chatterino7 as possible. That means settings should remain the same (to allow a smooth migration), and autocompletion should work the same.

I've copied a lot of code from the fork, and I'd like to include the contributors of these changes.

# What's next?

The 7TV fork has three more features: animated profile pictures, EventApi (automatic emote updates), and Paints (custom user colors/gradients/etc.).

I think, if we want users to switch to upstream chatterino, we should at least add the EventApi in a followup PR before the next release (I could do that since I've added the implementation in the fork). I noticed, in contrast to BTTV and FFZ, users tend to add/remove/update emotes more frequently and having used it I must say it's very nice not having to press F5 every minute in some channels. In addition, the [user-by-connection](https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/internal/rest/v3/routes/users/users.by-connection.go#L28) endpoint caches the response for 60s so F5 might not work as expected immediately.

I'm not sure about animated profile pictures but it shouldn't be a large change.


src/providers/seventv/SeventvEmotes.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+class SeventvEmotes final
+{
+public:
+    SeventvEmotes();
+
+    std::shared_ptr<const EmoteMap> globalEmotes() const;
+    boost::optional<EmotePtr> globalEmote(const EmoteName &name) const;
+    void loadGlobalEmotes();
+    static void loadChannelEmotes(std::weak_ptr<Channel> channel,
```
Nerixyz, line 64: I didn't go with the other emote providers' method names (`emotes(), emote(), loadEmotes(), loadChannel()`) since I think they're confusing if they're operating on global or channel emotes.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1021,24 +1023,44 @@ Outcome TwitchMessageBuilder::tryAppendEmote(const EmoteName &name)
...
     auto flags = MessageElementFlags();
     auto emote = boost::optional<EmotePtr>{};
 
     // Emote order:
     //  - FrankerFaceZ Channel
+    //  - 7TV Channel
     //  - BetterTTV Channel
+    //  - 7TV Global
     //  - FrankerFaceZ Global
     //  - BetterTTV Global
```
Nerixyz, line 1037: This order doesn't match the concept BTTV -> FFZ -> 7TV, but that's how it was implemented in the fork.

src/singletons/Settings.hpp
```diff
@@ -194,6 +195,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting enableEmoteImages = {"/emotes/enableEmoteImages", true};
     BoolSetting animateEmotes = {"/emotes/enableGifAnimations", true};
     FloatSetting emoteScale = {"/emotes/scale", 1.f};
+    BoolSetting showUnlistedEmotes = {"/emotes/showUnlistedEmotes", false};
```
Nerixyz, line 198: This setting doesn't mention 7TV although 7TV is the only emote-provider that has unlisted emotes. I kept it to allow a smooth migration of settings.

Felanbird:
> Is it 7TV, 7Tv, SevenTV, SevenTv, or Seventv?

7TV for things the user would actually see, whatever matches the code standard for code `sevenTV SevenTV sevenTv` (whatever fits.)
image
image
image
image

AnatoleAM:
> Use REST v3 for Badges (not added yet)

All other functionality is coming online, but there will not be a badges/cosmetics endpoint in v3; the way cosmetics will work moving forward is going to completely change, whereas instead of a large static endpoint it will be dynamic and event-driven.

You can continue to use `7tv.io/v2/cosmetics` safely for now, it's a special case which won't be deprecated until the new cosmetics system is ready.

> Regarding brand formatting

If something is user-facing it should be spelled as `7TV`, variation like `seventv` is only used internally, in code or legal documents.

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -0,0 +1,469 @@
...
+        .onError([](const NetworkResult &result) {
+            qCWarning(chatterinoSeventv)
+                << "Couldn't load 7TV global emotes" << result.getData();
+        })
+        .execute();
+}
+
+void SeventvEmotes::loadChannelEmotes(const std::weak_ptr<Channel> &channel,
+                                      const QString &channelId,
+                                      std::function<void(EmoteMap &&)> callback,
```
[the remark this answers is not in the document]
Nerixyz: no

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -196,6 +200,7 @@ class TwitchChannel : public Channel, public ChannelChatters
 protected:
     Atomic<std::shared_ptr<const EmoteMap>> bttvEmotes_;
     Atomic<std::shared_ptr<const EmoteMap>> ffzEmotes_;
+    Atomic<std::shared_ptr<const EmoteMap>> seventvEmotes_;
```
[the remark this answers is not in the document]
Nerixyz: There's no rule for protected memebers' suffix. Some classes use an underscore, some don't. This should be specified in `.clang-tidy` with [`readability-identifier-naming.ProtectedMemberSuffix`](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#cmdoption-arg-protectedmembersuffix).

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1021,24 +1023,44 @@ Outcome TwitchMessageBuilder::tryAppendEmote(const EmoteName &name)
...
     if (this->twitchChannel && (emote = this->twitchChannel->ffzEmote(name)))
     {
         flags = MessageElementFlag::FfzEmote;
     }
+    else if (this->twitchChannel != nullptr &&
+             (emote = this->twitchChannel->seventvEmote(name)))
+    {
```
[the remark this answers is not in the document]
Nerixyz: I don't think these cases should be merged even though they have the same body, since we explicitly define the order about the statements.

src/providers/twitch/TwitchMessageBuilder.cpp
```diff
@@ -1210,6 +1232,14 @@ void TwitchMessageBuilder::appendFfzBadges()
     }
 }
 
+void TwitchMessageBuilder::appendSeventvBadges()
```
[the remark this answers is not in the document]
Nerixyz: This seems wrong, we're calling `this->emplace` which isn't static.

src/providers/seventv/SeventvEmotes.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
+
+#include "SeventvEmoteCache.hpp"
```
Nerixyz, line 3: Should be
```suggestion
#include "providers/seventv/SeventvEmoteCache.hpp"
```
I'll do that next push.

src/providers/seventv/SeventvEmotes.hpp
```diff
@@ -0,0 +1,74 @@
...
+    static void loadChannelEmotes(const std::weak_ptr<Channel> &channel,
+                                  const QString &channelId,
+                                  std::function<void(EmoteMap &&)> callback,
+                                  bool manualRefresh);
+
+private:
+    Atomic<std::shared_ptr<const EmoteMap>> global_;
+};
+
+}  // namespace chatterino
```
pajlada, line 74: Missing newline :angry:

src/providers/seventv/SeventvEmoteCache.cpp
```diff
@@ -0,0 +1,33 @@
+#include "SeventvEmoteCache.hpp"
```
pajlada, line 1: Assuming we're not losing 99% performance or something, I'd prefer a simpler uncached version to start with, will make it easier to understand what the PR does and keeps it as minimal as possible

src/providers/twitch/TwitchMessageBuilder.cpp
pajlada: I propose the following emote order:
FrankerFaceZ Channel -> BetterTTV Channel -> 7TV Channel
FrankerFaceZ Global -> BetterTTV Global -> 7TV Global

The global order change will not make any difference for users, the channel order *should* not make any difference, unless a user has two emotes added with the same name.

src/singletons/Settings.hpp
pajlada: Let's at least rename the variable name - I'm fine with leaving the setting path the same.

`showUnlistedSevenTVEmotes`

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -314,7 +316,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addCheckbox("Enable emote auto-completion by typing :",
                        s.emoteCompletionWithColon);
     layout.addDropdown<float>(
-        "Size", {"0.5x", "0.75x", "Default", "1.25x", "1.5x", "2x"},
+        "Size", {"0.5x", "0.75x", "Default", "1.25x", "1.5x", "2x", "3x", "4x"},
```
pajlada, line 319: Do you know why this was added/necessary for this PR?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -328,6 +330,23 @@ void GeneralPage::initLayout(GeneralPageView &layout)
 
     layout.addCheckbox("Remove spaces between emotes",
                        s.removeSpacesBetweenEmotes);
+    layout.addCheckbox("Show unlisted / unapproved emotes (7TV only)",
```
pajlada, line 333: A little rename, matches documentation here: https://github.com/Chatterino/wiki/pull/235

```suggestion
    layout.addCheckbox("Show unlisted 7TV emotes",
```

src/singletons/Settings.hpp
```diff
@@ -149,6 +149,7 @@ class Settings : public ABSettings, public ConcurrentSettings
         "/appearance/badges/useCustomFfzModeratorBadges", true};
     BoolSetting useCustomFfzVipBadges = {
         "/appearance/badges/useCustomFfzVipBadges", true};
+    BoolSetting showBadgesSeventv = {"/appearance/badges/seventv", true};
```
pajlada, line 152:
```suggestion
    BoolSetting showBadgesSevenTV = {"/appearance/badges/seventv", true};
```

pajlada (changes requested):
Overall looks good, some logic changes left to figure out

src/messages/MessageElement.hpp
```diff
@@ -87,54 +95,64 @@ enum class MessageElementFlag : int64_t {
...
     // for links
-    LowercaseLink = (1LL << 29),
-    OriginalLink = (1LL << 30),
+    LowercaseLink = (1LL << 32),
+    OriginalLink = (1LL << 33),
 
     // ZeroWidthEmotes are emotes that are supposed to overlay over any pre-existing emotes
-    // e.g. BTTV's SoSnowy during christmas season
-    ZeroWidthEmote = (1LL << 31),
+    // e.g. BTTV's SoSnowy during christmas season or zerowidth 7TV emotes
```
Felanbird, line 144:
```suggestion
    // e.g. BTTV's SoSnowy during christmas season or 7TV's RainTime
```
"zerowidth emotes are 7tv zerowidth emotes" (yes I'm aware it's a chatterino7 comment but I've wanted it fixed for a while.)

src/widgets/settingspages/GeneralPage.cpp
Nerixyz: I suppose since we have emotes with larger sizes (#3812) we can display them bigger.

For reference:
`emoteScale` is used in
* https://github.com/Chatterino/chatterino2/blob/c692dd9b4461f486762f44cc007125c229d28592/src/messages/ImageSet.cpp#L64
* https://github.com/Chatterino/chatterino2/blob/c692dd9b4461f486762f44cc007125c229d28592/src/messages/MessageElement.cpp#L191
* https://github.com/Chatterino/chatterino2/blob/c692dd9b4461f486762f44cc007125c229d28592/src/messages/MessageElement.cpp#L493

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -0,0 +1,350 @@
...
+            auto ptr = cachedOrMake(std::move(result.emote), result.id);
+            emotes[result.name] = ptr;
+        }
+
+        return emotes;
+    }
+
+}  // namespace
+
+SeventvEmotes::SeventvEmotes()
```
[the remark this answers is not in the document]
Nerixyz: Shouldn't this be in the header file if any? This doesn't appear on clang-tidy 15.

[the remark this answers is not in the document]
dnsge: clang-tidy is delusional, many of its suggestions are nonsensical

ShowMeBillyJo:
I don't have anything technical to contribute, but I wanted to express my appreciation and gratitude to this whole team for working so hard on the 7TV integration. ❤️ Y'all are doing some fantastic work for the community. Thank you!

src/messages/ImageSet.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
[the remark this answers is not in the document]
Nerixyz: I've removed all 4x related changes, so this PR can be merged independently from #3812.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -636,6 +656,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                        s.useCustomFfzModeratorBadges);
     layout.addCheckbox("Use custom FrankerFaceZ VIP badges",
                        s.useCustomFfzVipBadges);
+    layout.addCheckbox("7TV", s.showBadgesSevenTV);
```
8thony, line 659: This setting should be above the split line, directly under FFZ Badge, in my opinion

Nerixyz: Thanks, I missed that!

src/providers/twitch/TwitchMessageBuilder.cpp
AnatoleAM: With what can be seen so far during this halloween season, it seems many streamers are taking advantage of emote sets and the fact 7TV overrides the other providers.

Perhaps worth re-thinking the proposed order for channel emotes?

Felanbird: > Perhaps worth re-thinking the proposed order for channel emotes?

Why don't they just.... you know.... not use emotes with the same name?

If they find it annoying to swap emotes out during seasonal events, then they should use 7TV as it has the best features for emote-set-swapping, but I don't see the argument for priority when it comes to something the user chooses to do.

Global emotes on the other hand would be a better argument.

LiveLaden: > Why don't they just.... you know.... not use emotes with the same name?

For people that upload emotes it sucks, I would have to delete emotes with a lot of channels using it from BTTV, since they have limited upload slots, and it auto-enables on the channel

Felanbird: > For people that upload emotes it sucks, I would have to delete emotes with a lot of channels using it from BTTV, since they have limited upload slots, and it auto-enables on the channel

Is this issue resolved by 7TV? So the same scenario but backwards won't happen to 7TV emote uploaders?

LosFarmosCTL: > Is this issue resolved by 7TV? So the same scenario but backwards won't happen to 7TV emote uploaders?

Yeah, it actually is, you don’t have to enable your own emotes in 7tv, you can just disable them but keep them uploaded.

LiveLaden: > Is this issue resolved by 7TV? So the same scenario but backwards won't happen to 7TV emote uploaders?

It won't happen, since I can create and upload emotes, and they don't take slots in my channel, BTTV after the change merged emotes created and added to the channel, it's really stupid

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * # References
+ *
+ * - EmoteSet: https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/data/model/emote-set.model.go#L8-L18
+ * - ActiveEmote: https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/data/model/emote-set.model.go#L20-L27
+ * - EmotePartial (emoteData): https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/data/model/emote.model.go#L24-L34
+ * - ImageHost: https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/data/model/model.go#L36-L39
+ * - ImageFile: https://github.com/SevenTV/API/blob/a84e884b5590dbb5d91a5c6b3548afabb228f385/data/model/model.go#L41-L48
+ */
+namespace {
```
pajlada, line 29: Move the anonymous namespace out of the chatterino namespace, but have `using namespace chatterino;` at the top to make the indentation a bit nicer - see IrcMessageHandler.cpp

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+            qCDebug(chatterinoSeventv)
+                << "Loaded" << emoteMap.size() << "7TV Channel Emotes for"
+                << channelId << "manual refresh:" << manualRefresh;
+
+            if (hasEmotes)
+            {
+                callback(std::move(emoteMap));
+            }
+            if (auto shared = channel.lock(); manualRefresh)
```
pajlada, line 306: I don't like this usage of `if (initialization; check)` where the initialization & check are completely different

I'd rather check manualRefresh first, then lock channel and output something if the lock succeeded

pajlada (changes requested):

src/providers/seventv/SeventvBadges.cpp
```diff
@@ -0,0 +1,72 @@
...
+        return emotes_[it->second];
+    }
+    return boost::none;
+}
+
+void SeventvBadges::loadSeventvBadges()
+{
+    // Cosmetics will work differently in v3, until this is ready
+    // we'll use this endpoint.
+    static QUrl url("https://api.7tv.app/v2/badges");
```
AnatoleAM, line 33: Can this be changed to `/v2/cosmetics`? Badges is an alias that has been deprecated for a while

AnatoleAM: Also, use `7tv.io` as base instead of `api.7tv.app`

Nerixyz: sure

pajlada (approved):

PR #4003, opened 2022-09-19
Migrate /announce command to Helix API.

Mm2PL:
Pull request checklist:

# Description


Wissididom:
Idk if it is important, but shouldn't `Helix::sendChatAnnoucement` rather be `Helix::sendChatAnnouncement` (missing n in Announcement)

Felanbird:
Can we implement colored announcements in this PR? I believe the only real disdain against them was the fact that IRC invoked announcements did not return color data, which is "resolved" by their deprecation.

Mm2PL:
> Idk if it is important, but shouldn't Helix::sendChatAnnoucement rather be Helix::sendChatAnnouncement (missing n in Announcement)

fixed

Mm2PL:
> Can we implement colored announcements in this PR (maybe just the commands)? I believe the only real disdain against them was the fact that IRC invoked announcements did not return color data, which is "resolved" by their deprecation.

I decided to introduce internal support for sending colored announcements as it was simply more effort to not do it. I didn't want to do commands because I wanted the PR to only move the existing command to helix, just fix the 'bug' and not introduce any new features at the same time. :)

pajlada (approved):
:+1: Looks good to me!
Letting you hit the Merge button so you have an opportunity to complain about the changes I've made @Mm2PL

Mm2PL:
ibana.pjg

PR #4028, opened 2022-10-01
fix: `smoothScrollingNewMessages` sometimes hiding messages

### LINKED ISSUE #779: "smoothScrollingNewMessages" hides initial messages in channels
If you start chatterino but you don't focus the tab /mentions, you get a highlight in that tab but you don't get the messages there until you focus it first.

To test, open chatterino, get pinged by a bot or something and see the /mentions tab highlight.
Click it and it has no message.

Nerixyz:
Pull request checklist:

# Description

This is an attempt at fixing #779.

To reproduce the issue, make sure chatterino doesn't start in the ChannelView you want to test (e.g. start it in `foo`).
I used the `/mentions` view to test since it's empty at the start, so make sure you have one before starting.

Then mention the current user from another instance in e.g. `bar` (to be safe, use a channel that hasn't been shown yet). Now go to the `/mentions` channel.

I suggest you to inspect the values:

<details><summary>Added logging</summary>

```diff
diff --git a/src/widgets/Scrollbar.cpp b/src/widgets/Scrollbar.cpp
index 13f1a08d..e9fa449b 100644
--- a/src/widgets/Scrollbar.cpp
+++ b/src/widgets/Scrollbar.cpp
@@ -156,10 +156,33 @@ void Scrollbar::setDesiredValue(qreal value, bool animated)
             else
             {
                 this->currentValueAnimation_.setEndValue(value);
+                if (auto *p = dynamic_cast<ChannelView *>(this->parent()))
+                {
+                    qCDebug(chatterinoWidget)
+                        << p->channel()->getName() << "[SET END]";
+                }
             }
         }
     }
 
+    if (auto *p = dynamic_cast<ChannelView *>(this->parent()))
+    {
+        bool hasChanged =
+            std::abs(this->currentValue_ + this->smoothScrollingOffset_ -
+                     value) > 0.0001;
+        qCDebug(chatterinoWidget)
+            << p->channel()->getName() << "{ hasChanged:" << hasChanged << "}"
+            << "desired:" << value << "startValue:"
+            << this->currentValueAnimation_.startValue().toReal()
+            << "endValue:" << this->currentValueAnimation_.endValue().toReal()
+            << "currentValue[A]:"
+            << this->currentValueAnimation_.currentValue().toReal()
+            << "currentValue[B]:" << this->currentValue_
+            << "state:" << this->currentValueAnimation_.state()
+            << "atBottom:" << this->atBottom_
+            << "smoothScrollingOffset:" << this->smoothScrollingOffset_;
+    }
+
     this->smoothScrollingOffset_ = 0;
     this->desiredValue_ = value;
 
```
</details>

Fixes #779.


src/widgets/Scrollbar.cpp
```diff
@@ -120,7 +120,7 @@ void Scrollbar::setDesiredValue(qreal value, bool animated)
     value = std::max(this->minimum_,
                      std::min(this->maximum_ - this->largeChange_, value));
 
-    if (std::abs(this->desiredValue_ + this->smoothScrollingOffset_ - value) >
+    if (std::abs(this->currentValue_ + this->smoothScrollingOffset_ - value) >
```
Nerixyz, line 123: Not sure if this is exactly needed but it seems reasonable to do, since the start value is set to `this->currentValue_ + this->smoothScrollingOffset_` and the end value is set to `value` and we want to find out if we need to animate.

If this is wrong then there should be a comment above why we need `this->desiredValue_`.

src/widgets/Scrollbar.cpp
```diff
@@ -153,6 +153,10 @@ void Scrollbar::setDesiredValue(qreal value, bool animated)
                     0.0001;
                 setCurrentValue(value);
             }
+            else
+            {
+                this->currentValueAnimation_.setEndValue(value);
```
Nerixyz, line 158: We're setting the end value here because we don't want to discard the value.

To see why it might be helpful to log all values here as well.

pajlada (approved):

PR #4035, opened 2022-10-02
Enable Helix timegating

pajlada:
Pull request checklist:

# Description

For commands affected by the timegating, they will continue to use their IRC command equivalent until **the 11th of February, 2023**.
This is one week before the actual migration is supposed to start. The wording of the date is shaky, so we start a bit before to be sure.
Any highly affected commands will have a temporary setting at the bottom of the General settings page to override the timegating functionality.

Any commands that are affected will also have their changelog entry updated to notify of the timegating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@
 - Minor: Migrated /unvip command to Helix API. (#4025)
 - Minor: Migrated /untimeout to Helix API. (#4026)
 - Minor: Migrated /unban to Helix API. (#4026)
-- Minor: Migrated /raid command to Helix API. (#4029)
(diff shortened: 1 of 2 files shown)


Mm2PL:
:(

pajlada:
How it's intended to be used can be seen here: https://github.com/Chatterino/chatterino2/blob/54129f7/src/controllers/commands/CommandController.cpp#L1988-L2009

pajlada:
> I also think we should link to the [deprecation announcement](https://discuss.dev.twitch.tv/t/deprecation-of-chat-commands-through-irc/40486) somewhere in the changelog, since it's referenced so often. Maybe this PR is not the right place for it though.
> 
> What do you think?

My idea was to have a wiki page or something for the deprecation with some information, and that will link to the deprecation announcement

PR #4041, opened 2022-10-03
fix: Double-space when using replies with an empty input box

### LINKED ISSUE #4031: Double-space when using replies with an empty input box
Due to the changes implemented in #3989 which allowed text in the input box to not be lost when executing a reply, there is now a double-space when executing a reply with an empty input box
image

This is purely a cosmetic minor-annoyance issue as double spaces in the input box are handled fine by Chatterino.

It will require a bit of refactoring as removing the space from this line, fixes the issue, but crashes when trying to execute a reply with text in the input box.
https://github.com/Chatterino/chatterino2/blob/ba586f01d0c015fd3111ae6d7e33db23c4c89631/src/widgets/splits/SplitInput.cpp#L980

Nerixyz:
Pull request checklist:

# Description

Fixes #4031


Felanbird (approved):
Fixed 👍

Issue #1163, opened 2019-07-27
Some channel-specific notices are shown in all channels.


RAnders00:
E.g:

```
user> /w forsen test
(in all channels): That user's settings prevent them from receiving this whisper.
```

That notice should only appear in the channel where the whisper was sent i think?

RAnders00:
Also affects the `This room is in subscribers only mode. To talk, purchase a channel subscription at https://www.twitch.tv/products/jtv/ticket?ref=subscriber_only_mode_chat` notice

RAnders00:
also this is a bug not an enhancement in my opinion since the underlying IRC command specifies a channel, even for the whisper notice originally reported

```
@msg-id=whisper_restricted_recipient :tmi.twitch.tv NOTICE #testchannel :That user's settings prevent them from receiving this whisper.
```

I suppose it gets harder when whispers are sent in the `/whispers` split, and re-routing the message from whatever channel it really goes to to the `/whispers` split would be an `enhancement`.

Mm2PL:
> Also affects the `This room is in subscribers only mode. To talk, purchase a channel subscription at https://www.twitch.tv/products/jtv/ticket?ref=subscriber_only_mode_chat` notice

This happens only with the `#jtv` channel. A comment in the code mentions that `NOTICE #jtv` is a global notice, is that still the case?

jupjohn:
Retested post #4052 merge, whisper failures now only show in the channel they were sent from. Subscriber mode is also per-channel. If there aren't any other cases we can think of, this issue could probably be closed :+1:

Felanbird:
Tested both whispers and subscribers and as mentioned the error is only printed in the correct channel :+1:

PR #4052, opened 2022-10-07
Migrate `/w` to Helix API

### LINKED ISSUE #3986: Migrate /w command to Helix API
As part of Twitch's announced deprecation of IRC-based commands ([see here for more info](https://discuss.dev.twitch.tv/t/deprecation-of-chat-commands-through-irc/40486), the `/w https` command needs to be migrated to use the relevant Helix API endpoint.

Helix API reference: https://dev.twitch.tv/docs/api/reference#send-whisper

Split from #3956

## Consequence
Users who might have been able to use `/w` to whisper users, may experience issues using the Helix API due to the two methods using the same constraints.

Nerixyz:
Pull request checklist:

# Description

This migrates _both_ `/w` and `.w` to the Helix API since `.w` wasn't sent directly to IRC.

Closes #3986.
Also some parts of #1163 will be fixed.

### DIFF
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -181,28 +181,6 @@ bool appendWhisperMessageWordsLocally(const QStringList &words)
     return true;
 }
 
-bool appendWhisperMessageStringLocally(const QString &textNoEmoji)
-{
-    QString text = getApp()->emotes->emojis.replaceShortCodes(textNoEmoji);
-    QStringList words = text.split(' ', Qt::SkipEmptyParts);
-
-    if (words.length() == 0)
-    {
-        return false;
-    }
-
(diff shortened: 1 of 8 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -181,28 +181,6 @@ bool appendWhisperMessageWordsLocally(const QStringList &words)
     return true;
 }
 
-bool appendWhisperMessageStringLocally(const QString &textNoEmoji)
```
Nerixyz, line 184: I'm not too confident in removing this, but I couldn't really see a reason for this to exist other than maybe some `dryRun` stuff.

Felanbird (changes requested):

pajlada (approved):

PR #4055, opened 2022-10-11
Fix rare reply mention crash

pajlada:
- Fix potential out-of-range access of at when stripping reply mention if the message contained nothing other than the username
- Update changelog entry

Pull request checklist:

# Description

Example crash scenario

Content: `@pajlada`
Tag `reply-parent-display-name`: `pajlada`

content.at(1+displayName.length()) would look for the character after the username `pajlada`, but since there was no more text it would throw an exception or something

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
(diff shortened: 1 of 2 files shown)


Felanbird:
still can't auto-merge PRs
image

PR #4066, opened 2022-10-16
Prevent copying in a couple places

zneix:
Pull request checklist:

# Description

Prevent copying QString type in iterations
Prevent copying while binding to a temporary

### DIFF
--- a/src/controllers/filters/parser/FilterParser.cpp
+++ b/src/controllers/filters/parser/FilterParser.cpp
@@ -54,7 +54,7 @@ ContextMap buildContextMap(const MessagePtr &m, chatterino::Channel *channel)
 
     bool subscribed = false;
     int subLength = 0;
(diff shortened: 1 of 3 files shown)


src/controllers/filters/parser/FilterParser.cpp
```diff
@@ -54,7 +54,7 @@ ContextMap buildContextMap(const MessagePtr &m, chatterino::Channel *channel)
 
     bool subscribed = false;
     int subLength = 0;
-    for (const QString &subBadge : {"subscriber", "founder"})
+    for (const char *const &subBadge : {"subscriber", "founder"})
```
pajlada, line 57: Try `const auto &` here instead, might look nicer

pajlada (changes requested):
nitpick

pajlada (approved):

PR #4070, opened 2022-10-19
Migrated `getSubage` to v2 version of the API

zneix:
Pull request checklist:

# Description

- Migrated `IvrApi::getSubage` function to use v2 version of its API.
- Updated link to the endpoint's documentation.
- Since this was the last endpoint from this provider to use a v1 route, base url for the API and its usages have been updated accordingly.
- Also applied some copying removal (not sure if it's correct though, pls correct me if that's a bad move).


pajlada (approved):
LGTM :+1:

PR #4087, opened 2022-10-29
feat: Basic PubSub Manager for Live Emote Updates

Nerixyz:
Pull request checklist:

# Description

To add live emote updates later, I'm adding a common PubSub manager that will be used for both BTTV and 7TV. 
This also adds a test that requires changes to twitch-pubsub-server-test (Chatterino/twitch-pubsub-server-test#27).

I'm not sure how much stuff I should implement here. My thinking is that I'll add message/emote-related changes when implementing support for a provider later.

See #4038.


src/providers/liveupdates/BasicPubSubClient.hpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    virtual ~BasicPubSubClient() = default;
+
+    BasicPubSubClient(const BasicPubSubClient &) = delete;
+    BasicPubSubClient(const BasicPubSubClient &&) = delete;
+    BasicPubSubClient &operator=(const BasicPubSubClient &) = delete;
+    BasicPubSubClient &operator=(const BasicPubSubClient &&) = delete;
+
+    virtual void start()
```
pajlada, line 51: If start/stop are not called by "users" of the class, it would be nice if they were not public.

For someone implementing a pubsubclient, it would be nicer if they could override `onConnectionEstablished` to perform on-connect actions

pajlada (approved):

PR #4088, opened 2022-10-29
Migrate /chatters commands to use Helix api

### LINKED ISSUE #3971: Migrate /mods command to Helix API
As part of Twitch's announced deprecation of IRC-based commands ([see here for more info](https://discuss.dev.twitch.tv/t/deprecation-of-chat-commands-through-irc/40486), the `/mods` command needs to be migrated to use the relevant Helix API endpoint.

Helix API reference: https://dev.twitch.tv/docs/api/reference#get-moderators

Split from #3956

## Consequence
**Only** broadcasters can now list moderators in their own channel

### LINKED ISSUE #4019: Migrate the Viewer List and the /chatters command to Helix API
As part of Twitch's announced deprecation of IRC ~~-based commands~~ (I don't remember where the announcement is for this specific one) the Viewer List and subsequently the /chatters command need to be migrated to use the relevant Helix API endpoint.

Helix API reference: https://dev.twitch.tv/docs/api/reference#get-chatters

It's mod only now so we'll need some form of error for `/chatters` and the viewer list for non-mods.

cbclemmer:
Pull request checklist:

# Description
closes #4019

## Features:
- Migrates /chatters command will use helix api or cache
	- If the user is neither mod or broadcaster the list is taken from the user list cache
	- If the user has mod rights the list is taken from the chat/chatters helix api command
- The old irc commands for /chatters have been removed based on comment in issue


pajlada (commented):
Thanks for working on this!

These changes are both pretty heavy, and while they rely on the same underlying functionality (the cursor-pages-fetching thing), it would be a lot easier to review if I could focus on a single functionality rather than two, since most feedback I have would apply to both.

So in short, could you undo the changes either to the /mods command or to the /chatters command (I have no preference) in this PR?

After that's done, I will go ahead and review the code and come with more proper/full feedback

cbclemmer:
@pajlada Code and pr message has been updated. I've got everything in another branch to merge in later. There were a few functions that will be shared for /chatters and /mods so some functions look like they shouldn't be their own thing, but they'll be useful later.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -621,7 +592,9 @@ const QString &TwitchChannel::popoutPlayerUrl()
 
 int TwitchChannel::chatterCount()
 {
-    return this->chatterCount_;
+    return this->hasModRights()
+        ? this->chatterCount_
+        : this->accessChatters()->getNumChatters();
```
pajlada, line 597: These two numbers are very different and used for different purposes, portraying them both as the same number is in my opinion confusing.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -135,10 +111,6 @@ TwitchChannel::TwitchChannel(const QString &name)
         }
     });
 
-    // timers
-    QObject::connect(&this->chattersListTimer_, &QTimer::timeout, [=] {
-        this->refreshChatters();
-    });
```
pajlada, line 141: Why do we want to disable this timer?

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -164,7 +136,6 @@ TwitchChannel::TwitchChannel(const QString &name)
 void TwitchChannel::initialize()
 {
     this->fetchDisplayName();
-    this->refreshChatters();
```
pajlada, line 167: Why do we no longer want to refresh chatters on startup?

src/controllers/commands/CommandController.cpp
```diff
@@ -889,9 +895,27 @@ void CommandController::initialize(Settings &, Paths &paths)
...
-                    .arg(localizeNumbers(twitchChannel->chatterCount()))));
+            auto addChatterCountMessage = [channel, twitchChannel]() -> void {
+                channel->addMessage(makeSystemMessage(
+                    QString("Chatter count: %1")
+                        .arg(localizeNumbers(twitchChannel->chatterCount()))));
+            };
+
+            // Fetch from chatter cache when user does not have mod rights
+            if (!twitchChannel->hasModRights()) {
+                addChatterCountMessage();
```
pajlada, line 906: Given my previous comment above, I would prefer to return an error message for non-mods rather than providing inaccurate information.

src/controllers/commands/CommandController.cpp
```diff
@@ -889,9 +895,27 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                return "";
+            }
+
+            // Use helix api when user has mod rights
+            twitchChannel->refreshChatters(
```
pajlada, line 911: This will end up making N requests fetching all viewers, it will probably be best to also provide a non-paginated version that this command can use directly

The way I envision it working is the user types `/chatters`, we make a request using `getHelix()->getChatters(channelsomething, /* how many users to fetch */ 1, /* pagination cursor */ boost::none, successCB, failCB)` where successCB only reads the total field and discards the rest of the payload

src/providers/twitch/api/Helix.hpp
```diff
@@ -329,6 +330,16 @@ struct HelixChatSettings {
     }
 };
 
+class HelixUserList {
+    public:
+        std::unordered_set<QString> users;
+        int total;
```
pajlada, line 336: Preferably `total` isn't included in this data structure to keep it as small/to-the-point as possible

src/providers/twitch/api/Helix.hpp
```diff
@@ -329,6 +330,16 @@ struct HelixChatSettings {
...
+
+        HelixUserList();
+
+        void AddUsersFromResponse(const QJsonObject &response);
```
pajlada, line 340: Casing should be `addUsersFromResponse`

Preferably the function is protected/private with only the Helix class having access to it so users don't get the idea that it can be called

src/providers/twitch/api/Helix.hpp
```diff
@@ -329,6 +330,16 @@ struct HelixChatSettings {
     }
 };
 
+class HelixUserList {
+    public:
+        std::unordered_set<QString> users;
+        int total;
+
+        HelixUserList();
```
pajlada, line 338: The constructor doesn't do anything special so it can be removed entirely

src/providers/twitch/api/Helix.cpp
```diff
@@ -8,6 +8,19 @@
...
+    users = std::unordered_set<QString>();
+}
+
+void HelixUserList::AddUsersFromResponse(const QJsonObject &response) {
+    for (const auto jsonStream : response.value("data").toArray()) 
+    {
+        auto user = jsonStream.toObject().value("user_login").toString();
+        users.insert(user);
+    }
+    users = std::move(users);
```
pajlada, line 21: This line doesn't do anything

src/providers/twitch/api/Helix.cpp
```diff
@@ -8,6 +8,19 @@
...
 
+HelixUserList::HelixUserList() {
+    users = std::unordered_set<QString>();
+}
+
+void HelixUserList::AddUsersFromResponse(const QJsonObject &response) {
+    for (const auto jsonStream : response.value("data").toArray()) 
+    {
+        auto user = jsonStream.toObject().value("user_login").toString();
+        users.insert(user);
```
pajlada, line 19: Use `this->` when accessing member variable

so `this->users.insert(user);` instead of `users.insert(user);`

src/providers/twitch/api/Helix.cpp
```diff
@@ -1991,6 +2004,157 @@ void Helix::sendWhisper(
...
+            errorMessage += "An unknown error has occurred.";
+        }
+        break;
+    }
+    return errorMessage;
+}
+
+// Recursive function with a maximun page of 50
+void Helix::getApiListRecursive(
+    HelixUserList *list,
```
pajlada, line 2046: Let's avoid a pointer here. You could pass it by reference, pass a `shared_ptr`, or move it from call to call.

pajlada (changes requested):

src/providers/twitch/TwitchChannel.cpp
cbclemmer: The chatter list can only be refreshed if you are a moderator now and it is automatically refreshed if a mod uses the /chatters command. So the only reason to update it is so that mods would have an updated @ list. I don't mind adding it back but I didn't know if it is worth it at that point.

src/providers/twitch/TwitchChannel.cpp
cbclemmer: Same as the timer essentially. If we need to make sure that mods have an updated chatter list it can be reverted.

src/providers/twitch/TwitchChannel.cpp
cbclemmer: `chatterCount_` has been removed because it cannot be used by non mods anyways. I've simplified this function down to just using `this->accessChatters()->getNumChatters()`

src/providers/twitch/TwitchChannel.cpp
cbclemmer: Mods should probably always have access to an accurate @ list. Reverted code.

src/controllers/commands/CommandController.cpp
cbclemmer: Added an error message.

src/controllers/commands/CommandController.cpp
cbclemmer: Good point, I forgot there is a total in the first returned request. I'll make a separate function to just get the total. The `Helix::GetChatters` function is still relevant because we need it for refreshing the @ list.

src/providers/twitch/api/Helix.hpp
cbclemmer: Ok, I don't think I need this class then, I'll just pass back the unordered set.

pajlada: Yeah that works well :+1:

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -621,7 +597,7 @@ const QString &TwitchChannel::popoutPlayerUrl()
 
 int TwitchChannel::chatterCount()
 {
-    return this->chatterCount_;
+    return this->accessChatters()->getNumChatters();
```
pajlada, line 600: We still can't use `getNumChatters` here since it relies on the chatters that are part of the `ChatterSet` and the `ChatterSet` a) doesn't remove stale entries b) doesn't go beyond 2000. So keeping and using the `chatterCount_` for this specific purpose still makes sense imo

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -917,31 +898,16 @@ void TwitchChannel::refreshChatters()
...
-                }
-
-                return pair.first;
-            })
-        .execute();
+    // Get chatter list via helix api
+    getHelix()->getChatters(
+        this->roomId(),
+        getApp()->accounts->twitch.getCurrent()->getUserId(),
+        [this](HelixUserList *chatterList) {
```
pajlada, line 905: We need to use the same `weakOf` 'hack' that the previous TMI version used to ensure the callback doesn't attempt to use a TwitchChannel that has been deleted

src/providers/twitch/TwitchChannel.cpp
cbclemmer: It now sets `chatterCount_` whenever `refreshChatters` is run and this function now uses `chatterCount_` again.

src/providers/twitch/TwitchChannel.cpp
cbclemmer: Not sure what you mean here. I tried to implement it based on the other functions and I think I got it...

cbclemmer:
@pajlada I changed the helper function for helix to be a more generalized version of `makeRequest` that can be paginated or not. I'll use this for the /mods pr too. It now can pass in a function that does what it needs to the response from the api and can recursively loop until the end of the list. The /chatters command now just gets one page of chat/chatters api and returns the the total property value. I also changed the error type to be a general error because there isn't any special error types with these requests.

src/providers/twitch/api/Helix.hpp
cbclemmer: Class deleted

src/providers/twitch/api/Helix.hpp
cbclemmer: Class deleted

src/providers/twitch/api/Helix.cpp
cbclemmer: class deleted

src/providers/twitch/api/Helix.cpp
cbclemmer: class deleted

src/providers/twitch/api/Helix.cpp
cbclemmer: Moved reference to list to the parent function referenced in a callback.

src/providers/twitch/api/Helix.hpp
```diff
@@ -519,6 +520,15 @@ enum class HelixWhisperError {  // /w
...
 };  // /w
 
+enum class HelixGeneralError {
+    Unknown,
+    UserMissingScope,
+    UserNotAuthorized,
+
+    // The error message is forwarded directly from the Twitch API
+    Forwarded,
+};
```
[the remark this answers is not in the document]
cbclemmer: Ah, ok that makes sense.

src/controllers/commands/CommandController.cpp
```diff
@@ -878,23 +878,79 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                break;
             }
+            return errorMessage;
+        };
 
+        if (words.size() != 1)
+        {
+            channel->addMessage(makeSystemMessage("Usage: /chatters"));
+            return "";
+        }
```
Felanbird, line 919: Why do we care about the input for this command?

Felanbird (changes requested):

pajlada:
@cbclemmer uno reverse card - I committed some stuff to your PR that I'd like you to review/approve. Most commits are small and do very little, they'd be easy to go through and it's easy to understand why. However, I made a big refactor PR for the pagination logic to change the pattern we use. This changes the pattern for paginated requests in this PR and for future paginated requests to need two functions; 1 public one that users (e.g. `CommandController` or `TwitchChannel`) can use which has a simple, single-use API and 1 private one that the public function will use & recursively call.
The big refactor commit can be found [here `a3aed80` (#4088)](https://github.com/Chatterino/chatterino2/pull/4088/commits/a3aed8084f02e110745de9a008c89a77c803d427) and has a commit message where I try to explain all of the changes I made and why.

Let me know if you have any questions!

cbclemmer:
@pajlada That all looks good. The reason I wanted to try to make a more general request was because the handling for the status codes and resulting errors were largely copied and pasted so I tried to make a framework to reduce that for the stuff I was writing. It does make a lot more sense to refactor everything later when everything is using a standardized way of handling errors like you commented before. I'm mainly a C# dev and garbage collection is usually an afterthought for me because of that, so thanks for ironing that out. Looks good to me 👍

src/controllers/commands/CommandController.cpp
cbclemmer: I was just following the convention of the other commands, but I guess those were using arguments and could lead being misused. I don't mind removing it.

pajlada (approved):

src/controllers/commands/CommandController.cpp
cbclemmer: nvm, it was already removed

Felanbird (approved):
I'm not 100% on the fact that we are deprecating the command very early, but `/chatters` usage is probably more common among stable users, so we could take another look at this around Feb when the deprecation of all other commands kicks in.

PR #4092, opened 2022-10-30
Add missing Text tag to reply message text

mohad12211:
# Description
Text Elements normally have the `MessageElementFlag::Text` flag, reply text does not.
This is needed in #3958 to properly order reply text.


pajlada (approved):
Good find :+1: Replies still work as expected after this

Thank you!

PR #4096, opened 2022-11-01
Include more error messaging for failed image uploads

pajlada:
Pull request checklist:

# Description

`parseJson` will return an empty object if the body is not a valid JSON string.

We currently try to read:
`.error` as an error string, of which we include a maximum of 300 characters.
`.code` as any type of value, of which we include a maximum of 20 characters.

Fixes #4095


pajlada:
Works
image

PR #4101, opened 2022-11-02
Fix Reply Text Showing In Reply Thread Popup

dnsge:
Pull request checklist:

There have been so many changes to the reply thread changelog entry (and this is a bugfix to a bugfix) that I really don't care if we don't add this PR to the changelog.

# Description

#4092 Introduced the following bug: 
image
The reply text is improperly rendered in the reply thread popup.

This PR fixes the bug. The bug itself is my fault because, during the (extensive) changes made to my original reply threads PR, I changed around how hiding the reply message elements worked. This change eliminates the `setRenderReplies` method which became unused at some point.

Also: sorry for the unrelated change of extracting the accessing of setting values. I just couldn't leave the repeated calls that could easily be reused.


pajlada (approved):
:+1: nice and readable

PR #4114, opened 2022-11-05
Fix chatters recursion not working

pajlada:
Pull request checklist:

# Description

Can be tested & verified with the `/test-chatters` command


pajlada:
Looks good to me

image

PR #4116, opened 2022-11-05
Move ChatSettings commands to their own file

pajlada:
Pull request checklist:

# Description

This is meant as an experimental refactor for CommandController to see if we can reduce the size of that class and make it easier to modify a single command without running the risk of having a huge merge conflict because someone else worked on a different command at the same time.

- Move ChatSettings commands to their own file
- reformat error message strings
- move ChatCommands together in CommandController.cpp


pajlada:
@xel86 I know you were interested in this if you don't mind taking a peek to see how this feels

Commands would be moved to `builtin/twitch` in logical groupings based on their functionality

We'd probably want  `/timeout`, `/ban/`, `/unban`, and `/untimeout` in the same file Moderation?

`/mods` would have its own file ListMods?
`/vips` would have its own file ListVips?

xel86:
> @xel86 I know you were interested in this if you don't mind taking a peek to see how this feels
> 
> Commands would be moved to `builtin/twitch` in logical groupings based on their functionality
> 
> We'd probably want `/timeout`, `/ban/`, `/unban`, and `/untimeout` in the same file Moderation?
> 
> `/mods` would have its own file ListMods? `/vips` would have its own file ListVips?

Yea these changes make things look way nicer and much more manageable.  Only a few nitpicks I noticed that clang-tidy already mentioned; mainly the unused `QStringList &words` in a few of the methods. You've voided some so may as well do all that apply.

As for next steps and the moderation commands; I agree with the ones you've mentioned but there are probably a few additional commands that would be suited for a `Moderation.hpp` like `/mod`, `/unmod` for instance.

xel86 (commented):
Yea these changes make things look way nicer and much more manageable. Only a few nitpicks I noticed that clang-tidy already mentioned; mainly the unused `QStringList &words` in a few of the methods. You've voided some so may as well do all that apply.

As for next steps and the moderation commands; I agree with the ones you've mentioned but there are probably a few additional commands that would be suited for a `Moderation.hpp`, like `/mod`, `/unmod`.

Or perhaps those should go into a unique `Broadcaster.hpp` for commands that only the broadcaster can perform (for Helix), which could be a suitable home for commands like `/raid`, `/unraid`, `/mod`, `/unmod`.

pajlada:
@xel86 With your approval, I'm happy to merge this in

src/controllers/commands/CommandController.hpp
```diff
@@ -46,10 +47,16 @@ class CommandController final : public Singleton
     using CommandFunction =
         std::function<QString(QStringList /*words*/, ChannelPtr /*channel*/)>;
 
-    void registerCommand(QString commandName, CommandFunction commandFunction);
+    using CommandFunctionWithContext = std::function<QString(CommandContext)>;
+
+    using CommandFunctionAlternatives =
```
xel86, line 53: Up to you but I feel like the name `CommandFunctionVariants` would make it easier to understand without looking at it's definition. Really means the same thing so not really a big deal either way.

xel86 (approved):

PR #4117, opened 2022-11-05
Migrate viewer list to helix with timegate setting

cbclemmer:
Pull request checklist:

# Description

This uses the new helix endpoints for mods, vips, and chatters to get the viewer list via the helix api instead of the tmi endpoint which *may* go out of date whenever the irc commands are deprecated. It basically future proofs the viewer list so that if the tmi endpoint is ever randomly deprecated the user can change a setting and it will work for mods and the broadcaster. This is **opt in only** and will still use the tmi endpoint by default using the same time gate functionality as we use for the irc commands. The setting is located with the other time gate settings at the bottom of the general settings tab. I intentionally copied and pasted most of the `Split::showViewerList` function instead of making a better integrated function for both use cases so that whenever it is deprecated the transition to delete the old code will be as painless as possible.

### DIFF
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -877,6 +877,19 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     helixTimegateModerators->setMinimumWidth(
         helixTimegateModerators->minimumSizeHint().width());
 
+
+    auto *helixTimegateViewerList =
+        layout.addDropdown<std::underlying_type<HelixTimegateOverride>::type>(
+            "Helix timegate viwer list behaviour",
+            // AlwaysUseIRC value is used to always use tmi endpoint
+            {"Timegate", "Always use tmi", "Always use Helix"},
+            s.helixTimegateViewerList,
+            helixTimegateGetValue,  //
+            helixTimegateSetValue,  //
+            false);
+    helixTimegateViewerList->setMinimumWidth(
+        helixTimegateViewerList->minimumSizeHint().width());
+
     layout.addStretch();
 
     // invisible element for width
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -956,6 +956,32 @@ void Split::openWithCustomScheme()
 
 void Split::showViewerList()
 {
+    switch (getSettings()->helixTimegateViewerList.getValue())
+    {
+        case HelixTimegateOverride::Timegate: {
+            if (CommandController::areIRCCommandsStillAvailable())
+            {
+                this->showViewerListTmi();
+            } else 
+            {
+                this->showViewerListHelix();
+            }
+        }
+        break;
+
+        case HelixTimegateOverride::AlwaysUseIRC: {
+            this->showViewerListTmi();
+        }
+        break;
+        case HelixTimegateOverride::AlwaysUseHelix: {
+            this->showViewerListHelix();
+        }
+        break;
+    }
+}
+
+void Split::showViewerListTmi()
+{
     auto viewerDock =
         new QDockWidget("Viewer List - " + this->getChannel()->getName(), this);
     viewerDock->setAllowedAreas(Qt::LeftDockWidgetArea);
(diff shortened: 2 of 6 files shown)


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -877,6 +877,18 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        layout.addDropdown<std::underlying_type<HelixTimegateOverride>::type>(
+            "Helix timegate viwer list behaviour",
+            // AlwaysUseIRC value is used to always use tmi endpoint
+            {"Timegate", "Always use tmi", "Always use Helix"},
+            s.helixTimegateViewerList,
+            helixTimegateGetValue,  //
+            helixTimegateSetValue,  //
+            false);
+    helixTimegateViewerList->setMinimumWidth(
+        helixTimegateViewerList->minimumSizeHint().width());
```
Felanbird, line 890: #### Don't feel any rush to action on this, I'd like some thoughts on my comment.

I would like this to use `AlwaysUseIRC` by default, because as of right now it seems the TMI viewer list will not be deprecated along with IRC commands.
`(It was announced separately and I do believe they said on stream the date would be announced "in the future", so it's unlikely)`

My main reason for this approach is due to the fact that new nightly users are generated daily, and this would allow them the most seamless continuation, and not require them to update, as this is the final IRC deprecation change.

If the deprecation does not happen in Feb, the list continues to work exactly as expected.

If the deprecation DOES happen, the following will happen:
- The viewer list will not work for non-mods, and there's nothing they can do about this.
- The viewer list will not work for mods or the broadcaster, but they can ratify this by swapping the setting to `Always Use Helix`


Comparing both scenarios, I believe less nightly users will be impacted taking the `AlwaysUseIRC` approach. (every single mod/broadcaster versus every single chatter)

cbclemmer: That makes sense. Should I just takeout the timed portion all together and just allow switching between helix and tmi with the tmi being the default?

Felanbird: You can probably take the time-gate out, but I'd wait for somebody else's opinion as they might disagree with me.

Mm2PL: I talked with pajlada about this. We basically came to the conclusion that using a timebomb for this is counterproductive and doesn't solve anything. I came up with the idea of adding an endpoint on the chatterino api and having a remote kill switch, when we know the TMI api is dead, we can just disable it on Chatterinos of willing users.

Felanbird (changes requested):
The approach you've gone with is fair, but it's not the intention of the viewer list, which is to show the people currently`(within reason)` in chat.

Due to the restrictions of the response we are given, this is the approach I have imagined:
1. Get the chatters list
2. Get the mods list
3. Get the vips list

`// for broadcasters`
If any of the users from the chatters response are also in the mods list, they are listed in the mods section
If any of the users from the chatters response are also in the vip list, they are listed in the vips section
Everyone else from the chatters response in the Viewers section

`// for mods`
If the broadcaster is in the Chatters response, place them at the top
The mods section should note that Chatterino can no longer check who is a moderator
The vips section should note Chatterino can no longer check who is a VIP
`(or remove these sections when it's a mod)`
Everyone else from the chatters response in the Viewers section

`// for plebs`
`Due to Twitch restrictions, this feature is only available for moderators. If you would like to see the Viewer list, you must use the Twitch website.`

image

```
{"data":[{"user_login":"felanbird"},{"user_login":"fossabot"},{"user_login":"kla
un85"},{"user_login":"ohbot"},{"user_login":"scriptorex"},{"user_login":"streame
lements"}],"pagination":{},"total":6}
```

src/widgets/settingspages/GeneralPage.cpp
pajlada: I think you should just remove the timeout section of this PR, yes.

If we want to look at some other solution to make it better for non-broadcaster users, it should be done in another PR imo.

cbclemmer: @pajlada So should there be a switch between the tmi and the helix requests? We can just keep the old function as is like I have it now.

cbclemmer:
@Felanbird New stuff:
- The viewer list will now only show mods and vips returned from the chatter list.
- The chatters section will not contain mods and vips.
- Warning text that you outlined are now visible.

cbclemmer:
I'm not a mod in any large channel so you probably want to test this yourself.

Felanbird:
> @Felanbird New stuff:
> 
>     * The viewer list will now only show mods and vips returned from the chatter list.
>     * The chatters section will not contain mods and vips.
:x:

image

>     * Warning text that you outlined are now visible.
✅ 
Implementation as of this comment loads fairly fast as a mod for 3,701 chatters

### Additional:
Mods should be lowercase, display name capitalization looks out of place with the rest of the list being lowercase.
Broadcaster should not be listed if they are not in the chatters response
All sections should be alphabetized

src/widgets/splits/Split.cpp
```diff
@@ -1110,6 +1137,375 @@ void Split::showViewerList()
...
+    else
+    {
+        chattersList->addItem(
+            formatListItemText("Due to Twitch restrictions, this feature is "
+                               "only available for moderators."));
+        chattersList->addItem(
+            formatListItemText("If you would like to see the Viewer list, you "
+                               "must use the Twitch website."));
+        loadingLabel->hide();
+    }
```
Felanbird, line 1451:
```suggestion
        chattersList->addItem(
            formatListItemText("Due to Twitch restrictions, this feature is "
                               "only \navailable for moderators."));
        chattersList->addItem(
            formatListItemText("If you would like to see the Viewer list, you "
                               "must \nuse the Twitch website."));
        loadingLabel->hide();
    }
```

I want to force a newline here as it would be better for the default window height.
before:
image
after:
image



It looks meh no matter what variation if you have Chatterino full-screened
https://user-images.githubusercontent.com/41973452/200187408-6b7b7cb0-8b40-4856-a70b-8505120884ce.png
https://user-images.githubusercontent.com/41973452/200187482-9dd79d95-e53f-4a4b-a25f-286da77d11ff.png

src/widgets/splits/Split.cpp
```diff
@@ -1110,6 +1137,375 @@ void Split::showViewerList()
...
+                auto errorMessage = formatModsError(error, message);
+                chattersList->addItem(formatListItemText(errorMessage));
+            });
+    }
+    else if (channel->hasModRights())
+    {
+        std::vector<QString> modList;
+        modList.push_back("Moderators cannot check who is a moderator");
+        std::vector<QString> vipList;
+        vipList.push_back("Moderators cannot check who is a vip");
```
Felanbird, line 1439:
```suggestion
        vipList.push_back("Moderators cannot check who is a VIP");
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -877,6 +877,18 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     helixTimegateModerators->setMinimumWidth(
         helixTimegateModerators->minimumSizeHint().width());
 
+    auto *helixTimegateViewerList =
+        layout.addDropdown<std::underlying_type<HelixTimegateOverride>::type>(
+            "Helix timegate viwer list behaviour",
```
Felanbird, line 882:
```suggestion
            "Helix timegate viewer list behaviour",
```

Felanbird (changes requested):

src/widgets/settingspages/GeneralPage.cpp
pajlada: I don't think we should keep any of the old "TMI Chatters" functionality. Remove the time gating that's been implemented in this PR and purely run requests of Helix and fail if the user does not have permission

cbclemmer:
@Felanbird I changed the vectors to QStringLists which had better functionality.
- All username should now be in lower case
- Everything should be sorted

cbclemmer:
> All sections should be alphabetized
@Felanbird I'm not sure what you mean

cbclemmer:
And broadcaster shouldn't be in list now if it wasn't returned from the chatters response.

Felanbird:
> > All sections should be alphabetized
> 
> @Felanbird I'm not sure what you mean

If possible the sections should be displayed 0-9 A-Z, to match the old TMI list
image

Felanbird:
- [x] All usernames are lowercase
- [x] Broadcaster should not be listed if they are not in the chatters response
- [x] All sections should be alphabetized
- [ ] The viewer list only shows mods and VIPs returned from the chatter list.
image

cbclemmer:
@Felanbird I missed the excluding vips and mods from the main viewer list. That should be fixed now. Also:
- I changed the "Viewers" section to "Chatters" to better reflect what is actually happening
- If a chatter is both a mod and a vip, they only show up in the mods list. I can change that if need be.

Felanbird:
> @Felanbird I missed the excluding vips and mods from the main viewer list. That should be fixed now. Also:

✔️ 
>     * I changed the "Viewers" section to "Chatters" to better reflect what is actually happening

I'm personally fine with that change but my opinion on it is not strong, so if we want to go back that's cool too
>     * If a chatter is both a mod and a vip, they only show up in the mods list. I can change that if need be.

That's fine, it shouldn't be possible to be both, so if that's happening it's a Twitch issue, mod seems more important to display.

cbclemmer:
@pajlada The time gate and setting stuff has been removed

Felanbird (approved):
I don't want to willingly give up the Chatters list this far before the actual deprecation `/chatters was a different story`, but I don't have a good argument why so I'll approve this in it's current state.

cbclemmer:
Is there anything still waiting to be done for this PR?

Felanbird:
the longer he doesn't remember to review this the longer we get to keep the viewer list 👺 

🙂

pajlada:
> Is there anything still waiting to be done for this PR?

No - everything in this PR seems good. I'm holding off on it a bit to do a proper review/test & evaluation

cbclemmer:
fwiw @Felanbird I agree with you lol

goldbattle:
Is there a reason that some moderators, vips, and chatters can't be directly filled in based on the people that have talked in the last x minutes in the chat? This is information that is already provided by just being in the chat room. Maybe displaying the `username (last chatted X minutes ago)` in the list would be really nice (maybe this is worth just doing a separate PR on?).

Felanbird:
> Is there a reason that some moderators, vips, and chatters can't be directly filled in based on the people that have talked in the last x minutes in the chat? This is information that is already provided by just being in the chat room. Maybe displaying the `username (last chatted X minutes ago)` in the list would be really nice (maybe this is worth just doing a separate PR on?).

It would definitely be a separate PR thing, if done.

Felanbird (approved):
Seems to still work as I would expect it to 👍

pajlada (approved):

PR #4122, opened 2022-11-07
Fix context menu not opening when username is right clicked from usercard/search/reply window

kornes:
Pull request checklist:

# Description
Right mouse click on chat username causes insert of mention to split's input, blocking opening of context menu.
This PR limits this behavior to default context only, so context menu will open normally in UserCard/Search/Reply view.


Felanbird (approved):
Right-clicking usernames in the pop-up windows open's the context menu.
image

It would be cool if right-clicking in the reply window would put the username into the input box, like in the main window, but that's not for this PR

pajlada (approved):

PR #4125, opened 2022-11-08
Switch to ncipollo/release-action for generating our Nightly releases

pajlada:
Pull request checklist:

# Description

Test release can be found here: https://github.com/pajlada/chatterino2/releases/tag/nightly-build

I also made the deb-creation script fall back to version `0.0.0-dev` in case no git tag was found - useful for forks that have no tags


Wissididom:
I noticed one thing:
In the current main Chatterino the deb file is called `Chatterino-x86_64.deb` and in the version of the PR source it is called `Chatterino.deb`

PR #4139, opened 2022-11-11
change unicode for better font support and fix some cases

BKSalman:
Pull request checklist:

# Description

changed ``0x200F`` unicode to ``0x202B`` for better [font support](https://github.com/Chatterino/chatterino2/pull/3958#issuecomment-1310792964),

and cases like ``الStyles`` (it's broken even in twitch, ال should be on the right and Styles on the left)
image

it has a minor issue with something like ``سلام 1-2``

thanks to @mohad12211 for the hard work

### DIFF
--- a/src/messages/layouts/MessageLayoutContainer.cpp
+++ b/src/messages/layouts/MessageLayoutContainer.cpp
@@ -278,12 +278,13 @@ void MessageLayoutContainer::reorderRTL(int firstTextIndex)
     // the second condition below covers the possible three cases:
     // 1 - if we are in RTL mode (first non-neutral word is RTL)
     // we render RTL, reversing LTR sequences,
-    // 2 - if we are in LTR mode (first non-neautral word is LTR or all wrods are neutral)
+    // 2 - if we are in LTR mode (first non-neautral word is LTR or all words are neutral)
     // we render LTR, reversing RTL sequences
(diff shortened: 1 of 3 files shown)


mohad12211 (approved):
can confirm this fixes the case when LTR is mixed with RTL word.
also it doesn't break punctuation marks.
I don't think the `2-1 سلام` thing can be solved easily. but that issue is still there with the `0x200F` character already.
its a very small thing anyways.

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -278,12 +278,13 @@ void MessageLayoutContainer::reorderRTL(int firstTextIndex)
     // the second condition below covers the possible three cases:
     // 1 - if we are in RTL mode (first non-neutral word is RTL)
     // we render RTL, reversing LTR sequences,
-    // 2 - if we are in LTR mode (first non-neautral word is LTR or all wrods are neutral)
+    // 2 - if we are in LTR mode (first non-neautral word is LTR or all words are neutral)
```
Felanbird, line 281:
```suggestion
    // 2 - if we are in LTR mode (first non-neutral word is LTR or all words are neutral)
```
sneak this in too

pajlada (approved):
Thank you!

PR #4172, opened 2022-11-20
Sort and force grouping of includes

pajlada:
Pull request checklist:

# Description

This change enforces strict include grouping using [IncludeCategories](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

In addition to adding this to the `.clang-format` file and applying it in the `tests/src` and `src` directories, I also did the following small changes:
- In `ChatterSet.hpp`, I changed lrucache to a <>include
- In `Irc2.hpp`, I change common/SignalVector.hpp to a "project-include"
- In `AttachedWindow.cpp`, `NativeMessaging.cpp`, `WindowsHelper.hpp`, `BaseWindow.cpp`, and `StreamerMode.cpp`, I disabled clang-format for the windows-includes
- In `WindowDescriptors.hpp`, I added the missing `vector` include. It was previously not needed because the include was handled by another file that was previously included first.
- `clang-format` minimum version has been bumped, so Ubuntu version used in the `check-formatting` job has been bumped to 22.04 (which is the latest LTS)


src/.clang-format
```diff
@@ -27,7 +27,28 @@ IndentCaseLabels: true
...
+IncludeBlocks: Regroup
+IncludeCategories:
+# Project includes
+  - Regex:  '^"[a-zA-Z\._-]+(/[a-zA-Z0-9\._-]+)*"$'
+    Priority: 1
+# Third party library includes
+  - Regex:  '<[[:alnum:].]+/[a-zA-Z0-9\._\/-]+>'
+    Priority: 3
+# Qt includes
+  - Regex:  '^<Q[a-zA-Z\._\/-]+>$'
```
kornes, line 40: should include numbers and i think can drop dot/dash 
https://doc.qt.io/qt-6/classes.html

src/.clang-format
```diff
@@ -27,7 +27,28 @@ IndentCaseLabels: true
...
+    Priority: 3
+    CaseSensitive: true
+# LibCommuni includes
+  - Regex:  '^<Irc[a-zA-Z]+>$'
+    Priority: 3
+# Misc libraries
+  - Regex:  '^<[a-zA-Z_]+\.h(pp)?>$'
```
kornes, line 47: unknown beforehand, so should include numbers too

kornes (commented):
looks good 👍

PR #4173, opened 2022-11-20
fix neutral elements order in multiple lines

mohad12211:
Pull request checklist:

# Description

If a text has RTL, and some line starts with neutral words (like emotes), it won't follow previous words (whether the previous was reversed or not) and that will result in wrong ordering across multiple lines.

before (wrong): 
image

after (correct): 
image

Test String: 
`واحد اثنين one two Kappa KKona three four `


pajlada:
The changelog entry is making changes to an already-released, be wary during merge & review.

pajlada (approved):

PR #4053, opened 2022-10-08
chore: migrate /vips command to Helix call

jammehcow:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

This PR aims to migrate the IRC server-side handling of the `/vips` command to a client-side Helix API call as part of the IRC command migration initiative. This feature is now broadcaster-only so care needs to be taken on how that's presented to the user :( 

## TODO: 
 - [x] actually display the VIPs in output
 - [x] sort VIPs alphabetically
 - [x] ensure all API errors are handled (at least the ones we want to share with the user)
 - [x] handle `0` VIPs
 - [x] verify how we want to explain to the user that `/vips` is broadcaster-only
 - [x] timegating (ty felan)

Closes #3984

PR #4174, opened 2022-11-21
Fix issue with VIPs who have Asian Language display names

### LINKED ISSUE #3524: `/vips` response changes
Twitch has changed the response Chatterino receives when using `/vips` to a case-sensitive variation of display names.

The change sort VIPs as follows:
1. 0-9
2. A-Z
3. a-z
4. multi-language characters [Japanese,Korean etc..]

List of things to do:
~1. Figure out if the case-sensitivity is intended (I've already asked a Staff member to see if the case-sensitivity can be changed.)~
	a) ~~If intended,~~ manually sort displaynames to be case-insensitive because it's stupid to be case-sensitive ✅ https://github.com/Chatterino/chatterino2/pull/3696
2. Figure out how the hell we are going to use asian language display names

~~or hopefully have to do nothing. 🙂~~

Felanbird:
Pull request checklist:

# Description

In a previous life, the IRC `/vips` response returned the `loginName` of said VIPs. One day we woke up and Twitch suddenly started sending us the `displayName`'s of those VIPs.
The problem with this was the fact that display names containing Asian language characters are of no use to us in Chatterino, as Twitch has no way of looking up a user, by these names.

Due to the migration of the `/vips` command to the Helix API, this fix will only help broadcasters (after the Timegate), but I do think it's better for them to be able to use the usercard feature in Chatterino, from the VIPs response.
Also suggested as a fix in https://github.com/Chatterino/chatterino2/pull/4053#discussion_r990955144

Fixes #3524

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
+
 ## 2.4.0
 
 - Major: Added support for emotes, badges, and live emote updates from [7TV](https://7tv.app). [Wiki Page](https://wiki.chatterino.com/Third_party_services/#7tv) (#4002, #4062, #4090)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
 
-## 2.4.0
+## 2.4.0-beta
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
```
Felanbird, line 5: This is the less-correct but more friendly to the average user, changelog.

I'd be willing to take the more-correct but less friendly approach as well:
```suggestion
- Bugfix: Prevented using display names for the `/vips` command, as Twitch does not allow looking up a user by their Asian language name. (#4174)
```

iProdigy (approved):

Felanbird (commented):
This works which is 🆒

pajlada (approved):

PR #4175, opened 2022-11-21
chore(msvc): replace /W3 with /W4

Nerixyz:
Pull request checklist:

# Description

If you looked through the build logs in CI, you might have noticed a warning on almost all files from `cl`: `Command line warning D9025 : overriding '/W3' with '/W4'`. This is because some project (I suspect websocketpp) is adding `/W3` to `CMAKE_CXX_FLAGS`.

We're doing the same fix here that [googletest does](https://github.com/google/googletest/blob/9c332145b71c36a5bad9688312c79184f98601ff/googletest/cmake/internal_utils.cmake#L49-L51) (just with `CMAKE_CXX_FLAGS`, as that's enough).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -783,6 +783,9 @@ endif ()
 
 # Configure compiler warnings
 if (MSVC)
+    # Someone adds /W3 before we add /W4.
+    # This makes sure, only /W4 is specified.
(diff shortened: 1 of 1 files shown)


kornes:
tested, warnings spam is gone, everything still works

Felanbird (approved):
I forgot what building Chatterino was like without an error on every single line

pajlada (approved):

PR #4181, opened 2022-11-21
fix: don't parent popup windows on Windows OS

### LINKED ISSUE #3863: Latest nightly doesn't display popup windows on the taskbar preview
### Checklist

### Describe your issue

New popup windows don't display on taskbar previews. It also seems like they aren't treated as separate windows since PowerToys FancyZones don't let them snap to layouts. 

This issue appears on the latest 7TV nightly versions too. 

### Screenshots

image
image

### OS and Chatterino Version

Chatterino Nightly 2.3.5 (commit 965b0667 modified) built on 2022-07-10 with Qt 5.15.2, Windows SDK, MSVC 193231332 Running on Windows 10 Version 2009, kernel: 10.0.22000

### LINKED ISSUE #4180: Minimizing popout windows places them in their own bar above the taskbar
Likely caused by #3836

### Discussed in https://github.com/Chatterino/chatterino2/discussions/4179

<sup>Originally posted by **Hotair10** November 21, 2022</sup>
In previous versions of Chatterino, when minimizing a popup chat, it would simply disappear into the taskbar. This paired perfectly with my "Always on Top" setting I had enabled, as it would be out of the way when I didn't want it blocking my screen, such as playing a game on that monitor.
In the newest update, it now minimizes into a little tab above my taskbar like this:
image

With the "Always on Top" setting enabled, this little tab is also always on top. While the tab can be moved around freely, it'd be nice if there was an option to disable the feature entirely and have it operate as it did in previous versions.

As a side note, it continues to add additional blips for each popup chat you minimize. No information on channel or anything, just a repeated list of the application title, making the feature even less useful than it could be.
image

tl;dr: Add a setting for old minimize behavior to still be an option.</div>

leon-richardt:
Pull request checklist:

# Description

Motivated by #4179, attempts to fix #3863 and fix #4180.

Need someone to test this as I don't have access to a Windows machine. This is not an ideal fix in my opinion. It would have been better if it was possible to parent windows on Windows and have them show up in the task bar regardless. However, I was not able to figure out how to do that or whether it is possible at all. Thus, aiming for this bandaid solution for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
 - Bugfix: Fixed the wrong right-click menu showing in the chat input box. (#4177)
-- Bugfix: Do not parent popup windows on Windows OS. (#4181)
+- Bugfix: Fixed popup windows not appearing/minimizing correctly on the Windows taskbar. (#4181)
 
 ## 2.4.0-beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
 - Bugfix: Fixed the wrong right-click menu showing in the chat input box. (#4177)
+- Bugfix: Do not parent popup windows on Windows OS. (#4181)
 
 ## 2.4.0-beta
 
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -262,12 +262,18 @@ Window &WindowManager::createWindow(WindowType type, bool show, QWidget *parent)
             return parent;
         }
 
+        // FIXME: On Windows, parenting popup windows causes unwanted behavior (see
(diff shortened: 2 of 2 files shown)


Felanbird (approved):
As expected both issues are resolved, I'm fine with this bandage approach as I doubt the user cares what approach we take, as long as it works.

Felanbird:
minimizing before this PR
image
taskbar when the popup window exists, before this PR
c2_4181_2

both minimizing & when a popup window exists, after this PR

c2_4181

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
 - Bugfix: Fixed the wrong right-click menu showing in the chat input box. (#4177)
+- Bugfix: Do not parent popup windows on Windows OS. (#4181)
```
Felanbird, line 7:
```suggestion
- Bugfix: Fixed popup windows not appearing/minimizing correctly on the Windows taskbar. (#4181)
```
more friendly to the user, I can apply before merge after this is tested by another windows user later.

leon-richardt: I agree, this is way more accessible. Feel free to apply whenever you see fit :+1:

pajlada (approved):

PR #4187, opened 2022-11-23
Fix crash happening when QuickSwitcher is used with popout window

kornes:
Pull request checklist:

## Crash reproduce
- rename any tab in main window to something unique like "asd"
- popout window (ctrl+n), can be any chat or empty tab
- open quick switch (ctrl+k) and search for "asd"
- click it to select and crash

## Why
QuickSwitcher has 2 types of search result: split (specific channel) or container (whole notebook page), tab rename above was used as simply way for container result which is focus of this crash.
Clicking on result invokes  `selectSplitContainer` signal:
https://github.com/Chatterino/chatterino2/blob/9d1b8b0a93a5d8df10ea6ea55159efbbe3bd65e1/src/singletons/WindowManager.cpp#L326
with callback declared here:
https://github.com/Chatterino/chatterino2/blob/9d1b8b0a93a5d8df10ea6ea55159efbbe3bd65e1/src/widgets/Notebook.cpp#L1071
Each poped out window adds own callback, so clicking on result in Quick Switcher fires selection in each window (even tho it doesnt actually work for now).
Crash is happening because `Notebook::select` called from handler, didn't check if passed page (aka splitContainer) is actually found in Notebook of given window.

## Fix in this PR
- `Notebook::findItem` will now return nullptr when requested page was not found, so `Notebook::select` can safely abort
- removed asserts since notebook select is now allowed to fail
- moved `page` null check earlier
- changed `setHidden(bool)` to equivalent `show()`, and call it only after page was found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Bugfix: Fixed crash happening when QuickSwitcher is used with popout window. (#4187)
+- Bugfix: Fixed crash happening when QuickSwitcher is used with a popout window. (#4187)
 - Bugfix: Fixed being unable to see the usercard of VIPs who have Asian language display names. (#4174)
 - Bugfix: Fixed the wrong right-click menu showing in the chat input box. (#4177)
 - Bugfix: Fixed popup windows not appearing/minimizing correctly on the Windows taskbar. (#4181)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Bugfix: Fixed crash happening when QuickSwitcher is used with popout window. (#4187)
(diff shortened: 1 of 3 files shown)


jammehcow:
Unrelated to the code changes, but I appreciate the amount of detail your put into your PRs; it really helps give context and aids in testing the change. Good stuff 👍

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Bugfix: Fixed crash happening when QuickSwitcher is used with popout window. (#4187)
```
```suggestion
- Bugfix: Fixed crash happening when QuickSwitcher is used with a popout window. (#4187)
```

Felanbird (approved):
Crash is resolved, as jammehcow has stated very useful explanation and crash-replication steps 👍

pajlada (approved):

PR #4193, opened 2022-11-27
perf: Speed Up Opening of Settings Dialog

### LINKED ISSUE #1266: Tabs are too wide after changing zoom
**Describe the bug**

After resetting the zoom some elements are misplaced and some others have size problems.

**How to Reproduce**

Zoom Chatterino as far as you can by holding control and scrolling up. Press Ctrl+0 to reset the zoom. Text boxes will be bigger than normal, tab headers will be bigger than needed, hints where the subs/notifs are will be moved towards the middle.

**Expected behavior**

The zoom will be reset like it is done with scrolling the wheel.

**Screenshots**

screenshot

**Chatterino version**

2.1.2 671c9ed
Edit: stupid mistake

**Additional context**

If you change the zoom by ctrl scrolling the wheel everything will be back to normal.

Nerixyz:
Pull request checklist:

# Description

I noticed that opening the settings dialog is really slow.

Two things jump out:

1. A lot of time is spent inside `updateScale`.
2. Calling `setWindowFlags` results in style recalculations ([Before](https://user-images.githubusercontent.com/19953266/204133202-6577709e-28a7-463d-ace9-ab454e0177bf.png), [After](https://user-images.githubusercontent.com/19953266/204133215-644bac40-770a-4cc4-b87f-3d204b65ba2c.png)).

I'm not sure if my fix to reduce calls to `updateScale` is correct. It would be nice if people could test this (e.g. with different zoom-levels and on high-dpi). If my fix turns out incorrect, I'd like to add a comment in the code as to why the change-callback must be run when creating the window (the second call to `updateScale` as added in d796517eb3f72604bb4cb0fee3fc014bfaa41fd6 to fix #1266).


kornes:
There is also this call
https://github.com/Chatterino/chatterino2/blob/f2b35fb7dc0b69ce60e740f2305f6ae510adbee3/src/widgets/BaseWindow.cpp#L787
triggered a LOT by `WM_SHOWWINDOW`, at minimum it could be changed to check if `nativeScale_` actually changed, but it could be also dropped from here i think, because `WM_DPICHANGED` handler is already calling `updateScale()`

another pr should properly fix #1266 without this double scalling fuckery :D

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Bugfix: Fixed the wrong right-click menu showing in the chat input box. (#4177)
 - Bugfix: Fixed popup windows not appearing/minimizing correctly on the Windows taskbar. (#4181)
 - Bugfix: Fixed white border appearing around maximized window on Windows. (#4190)
+- Bugfix: Improved time to open settings (#4193)
```
pajlada, line 12: The changelog entry is phrased as if it's in the Minor category rather than the Bugfix category

I'd prefer this to be something along the lines of "Fixed window scaling being applied too many times on startup, causing windows like Settings to be slow."

pajlada (changes requested):
Just changelog entry rewording, other than that the changes are solid :+1: 
thank you

Nerixyz:
> triggered a LOT by `WM_SHOWWINDOW`

I noticed that as well. I changed it to only call `updateScale` if the scale actually changed when this event is dispatched. This requires some further testing.

Nerixyz:
> triggered a LOT by `WM_SHOWWINDOW`

I noticed that as well.

> but it could be also dropped from here i think, because `WM_DPICHANGED` handler is already calling `updateScale()`

Probably, though, maybe `WM_DPICHANGED` doesn't get fired in some instances?

> another pr should properly fix #1266 without this double scalling fuckery :D

True, I think there might be more low-hanging fruits.

kornes:
> Probably, though, maybe `WM_DPICHANGED` doesn't get fired in some instances?
WM_DPICHANGED covers moving between screens and current screen change, so im 80% sure it could be removed from here, but i would leave it, once chatterino upgrades to qt6 it can be revisited with new dpi oriented apis

PR looks good to me 👍

pajlada (approved):

PR #4194, opened 2022-11-27
Clear up Highlight sound settings

### LINKED ISSUE #1540: Highlight sound settings are confusing
- "Play sound" and "Custom sound" are very far away. Columns need to be reordered to make more sense.
- "Cancel" in the file picker removes the custom sound. However there is no proper way to reset the custom sound appart from that.
- "Fallback sound" settings is hard to understand.
  - Why is it a checkbox? Can we remove the checkbox and call it "Default sound:"?

20200210_16h28m32s_grim

Nerixyz:
Pull request checklist:

# Description

As discussed in #1540, this clears up the highlight settings for sounds:

<details><summary>Without custom default sound</summary>

firefox_2022-11-27_18-14-59
</details>

Closes #1540.

### DIFF
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -22,11 +22,12 @@ namespace {
     QUrl getFallbackHighlightSound()
     {
         QString path = getSettings()->pathHighlightSound;
-        bool fileExists = QFileInfo::exists(path) && QFileInfo(path).isFile();
+        bool fileExists = !path.isEmpty() && QFileInfo::exists(path) &&
+                          QFileInfo(path).isFile();
 
-        // Use fallback sound when checkbox is not checked
-        // or custom file doesn't exist
-        if (getSettings()->customHighlightSound && fileExists)
+        // Use fallback sound when no custom sound is set
+        // or if the file doesn't exist.
+        if (fileExists)
         {
             return QUrl::fromLocalFile(path);
         }
--- a/src/messages/SharedMessageBuilder.cpp
+++ b/src/messages/SharedMessageBuilder.cpp
@@ -19,14 +19,16 @@ namespace chatterino {
 
 namespace {
 
+    /**
+     * Gets the default sound url if the user set one,
+     * or the chatterino default ping sound if no url is set.
+     */
     QUrl getFallbackHighlightSound()
     {
         QString path = getSettings()->pathHighlightSound;
         bool fileExists = !path.isEmpty() && QFileInfo::exists(path) &&
                           QFileInfo(path).isFile();
 
-        // Use fallback sound when no custom sound is set
-        // or if the file doesn't exist.
         if (fileExists)
         {
             return QUrl::fromLocalFile(path);
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -271,7 +271,6 @@ public:
 
     /// Highlighting
     //    BoolSetting enableHighlights = {"/highlighting/enabled", true};
-    //    BoolSetting customHighlightSound = {"/highlighting/useCustomSound", false};
 
     BoolSetting enableSelfHighlight = {
         "/highlighting/selfHighlight/nameIsHighlightKeyword", true};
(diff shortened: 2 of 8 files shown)


kornes:
suggestions:
- icon on clear button looks off, makes whole button kinda stands out too much. I would opt to remove it, or leave just icon, but then some variation of bin 🗑 would fit better i think.
- disable clear button when no custom sound is set
- maybe change current filename-only label, into input field with full file path, where user could paste/inspect path without clicking Change button. "Chatterino Ping" could be placeholder then.

8thony:
Things I noticed:
- If you have set a custom sound in the table before and then click to change the sound but hit the cancel button in the file browser, a ticked checkbox stays in the table. (Is this checkbox even necessary?)
- The table entry with no option to play a sound doesn't block the option to set a custom sound.

Nerixyz:
> * icon on clear button looks off, makes whole button kinda stands out too much. I would opt to remove it, or leave just icon, but then some variation of bin 🗑 would fit better i think.

Removed.

> * disable clear button when no custom sound is set

It's hidden now.

> * maybe change current filename-only label, into input field with full file path, where user could paste/inspect path without clicking Change button. "Chatterino Ping" could be placeholder then.

This doesn't really fit the other settings. A user can now click on the path to open a file.

> * If you have set a custom sound in the table before and then click to change the sound but hit the cancel button in the file browser, a ticked checkbox stays in the table. (Is this checkbox even necessary?)

I removed the checkbox since it wasn't used.

> * The table entry with no option to play a sound doesn't block the option to set a custom sound.

There were actually settings for the sound-urls. But they weren't used, so I removed them.

8thony:
If you set a sound with a long file name in the "Custom Sound" column and restart the program, the "Pattern" column becomes relatively narrow

If a sound with a long file name is set as "default sound", the settings window becomes wider

If you try this yourself, then make sure not to test the points above at the same time

src/singletons/Settings.hpp
```diff
@@ -271,7 +271,7 @@ class Settings : public ABSettings, public ConcurrentSettings
 
     /// Highlighting
     //    BoolSetting enableHighlights = {"/highlighting/enabled", true};
-    BoolSetting customHighlightSound = {"/highlighting/useCustomSound", false};
+    //    BoolSetting customHighlightSound = {"/highlighting/useCustomSound", false};
```
pajlada, line 274: Remove the setting entirely

```suggestion
```

src/messages/SharedMessageBuilder.cpp
```diff
@@ -22,11 +22,12 @@ namespace {
...
         QString path = getSettings()->pathHighlightSound;
-        bool fileExists = QFileInfo::exists(path) && QFileInfo(path).isFile();
+        bool fileExists = !path.isEmpty() && QFileInfo::exists(path) &&
+                          QFileInfo(path).isFile();
 
-        // Use fallback sound when checkbox is not checked
-        // or custom file doesn't exist
-        if (getSettings()->customHighlightSound && fileExists)
+        // Use fallback sound when no custom sound is set
+        // or if the file doesn't exist.
```
pajlada, line 29: Let's move this comment to above the function instead and describe what the function does

pajlada (changes requested):
Changing sounds doesn't update, it requires a restart

Nerixyz:
> If you set a sound with a long file name in the "Custom Sound" column and restart the program, the "Pattern" column becomes relatively narrow (Tested with 40 characters, including file extension)

Should be twice the width now, though this should really be changed after #3690.

> If a sound with a long file name is set as "default sound", the settings window becomes wider (Tested with 80 characters, including file extension)

Filenames are now truncated to 50 characters.

8thony:
> Should be twice the width now, though this should really be changed after #3690.

It doesn't change anything for me, but it seems to be a general issue that isn't unique to this settings page.

> Filenames are now truncated to 50 characters.

It's okay. Maybe it shouldn't be limited to a certain number but to the current size of the window. So making it wider reveals more of the name. I'll let you decide what is better.

pajlada (approved):

PR #4197, opened 2022-11-27
Fix unnecessary clears of split input selection

kornes:
# Description
Text selection in `ChannelView`, clears selection inside split input, and vice versa.
Currently due to a bug, handling of `selectionChanged` signal (which is fired each time new character selected) is always executing clear of split input. Bug is easily noticeable by resetting blink timer of input's cursor, while messages are being selected.

Fixed by proper checking if input has selection.

suggest no changelog


kornes:
while on this PR, i found out selection copy/clear is completely broken inside Reply window. Will fix it later today

pajlada (approved):

PR #4198, opened 2022-11-28
perf: ignore `WM_SHOWWINDOW` hide event

kornes:
Pull request checklist:

# Description

`handleSHOWWINDOW` is triggered by Windows for show and hide events; currently every tooltip (emotes, badges, links preview, split header) will run this function on mouse enter and leave.
Content of the function only cares about show event, so this PR introduce changes to minimze wasted cpu use:

- early exit when `msg->wParam` is false, meaning it was sent for hide event
- move `calcButtonsSizes()` (3 control buttons in custom frame title) to `shown_` block, so its called only once at first show event (resizing/scale change is covered elesewhere)
- early exit `calcButtonsSizes()` when window has no such buttons (frameless_)
- fix one call to `calcButtonsSizes()` outside of windows ifdef

tested:
- all mentioned tooltips show correctly
- calcButtonSizes works, resizing is not affected

suggest no changelog


src/widgets/BaseWindow.cpp
```diff
@@ -563,9 +563,9 @@ void BaseWindow::resizeEvent(QResizeEvent *)
             });
         });
     }
-#endif
 
     this->calcButtonsSizes();
+#endif
```
Nerixyz, line 568: Why was this moved? Shouldn't platforms other than Windows still recalculate the button sizes?

kornes: said buttons are part of custom frame, windows only. All other calls to calcButtonsSizes() are already ifdef'ed

pajlada (approved):

PR #4199, opened 2022-11-28
fix: iCCP - known incorrect sRGB profile

Nerixyz:
Pull request checklist:

# Description

When you start Chatterino, you get two warnings from `libpng`: `libpng warning: iCCP: known incorrect sRGB profile`.
This is because two resources shipped by Chatterino have this incorrect profile. This PR fixes that (essentially just `convert file.png file.png`).


fourtf:
Nice find! All this time i had no idea that it came from our own resources.

pajlada (approved):

PR #4201, opened 2022-11-28
fix: prevent unnecessary layout saves

kornes:
Pull request checklist:

# Description

`BaseWindow` is forcing delayed layout save
at every resizeEvent
https://github.com/Chatterino/chatterino2/blob/22bd78e3b3dd6721db5f75d3456e3876185f0f3d/src/widgets/BaseWindow.cpp#L543
and moveEvent
https://github.com/Chatterino/chatterino2/blob/22bd78e3b3dd6721db5f75d3456e3876185f0f3d/src/widgets/BaseWindow.cpp#L575

TooltipWidget and other subclasses unnecessarily triggers both of above.
PR adds new BaseWindow flag `DisableLayoutSave` to fix that

Tested movement and resizing of tooltips (after preview images load) and everything works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed unnecessary saving of windows layout. (#4201)
+
 ## 2.4.0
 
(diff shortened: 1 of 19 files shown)


kornes:
turned out to be quite a few windows/popups. 
not sure if its better to keep current Disable approach, or change it to Enable and only pick these we want to save (mostly splits/notebook stuff + emote popup)

pajlada (approved):

PR #4204, opened 2022-11-30
Fix red text color persisting from message overflow highlight

kornes:
Pull request checklist:

# Description
PR fixes problems with message overflow highlight, feature added in #3418

current issues:			
- selecting all (ctrl+a) on overflowed input and typing, will keep red color
- pressing up arrow on overflowed input, will keep red color on history messages
- removing characters at <500 position wont move red overflow marking, which can be useful when trying to edit message to fit
- hidden recursion in `setCharFormat()` re-emiting `QTextEdit::textChanged()`

PR introduced changes to fix above:
- always mark characters up to 500 with default color 
- add blocking of unwanted signals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed incorrect red coloring of message input. (#4204)
+
 ## 2.4.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed incorrect red coloring of message input. (#4204)
```
```suggestion
- Bugfix: Fixed message input showing as red after removing a message that was more than 500 characters. (#4204)
```

Felanbird (approved):
works as expected 👍

pajlada (approved):

PR #4207, opened 2022-12-01
feat: Allow negation of search predicates

### LINKED ISSUE #3998: Allow negation of search predicates
As described by @zneix in https://github.com/Chatterino/chatterino2/issues/458#issuecomment-655120637 (slightly abridged):

> option to just exclude some stuff but keep the rest - can be achieved like on GitHub, e.g.:
>   * `has:-link` would only show messages without links
>   * `from:-weeb123` would show everyone's messages but weeb123's

kornes:
Pull request checklist:

# Description

PR adds ability to negate search predicates, with both operators suggested in #3998 `-` and `!`

both of following will work the same:
```
badge:!no_audio
badge:-no_audio
```

comma separated lists are negated together as a group

closes #3998


pajlada (changes requested):
`isNegated` checking can be handled in the `MessagePredicate` base class instead, where we can also use human if-statements instead of robot xor :robot: 

`appliesTo` would then be protected, and `isNegated` private

src/messages/search/BadgePredicate.cpp
```diff
@@ -4,30 +4,28 @@
 
 namespace chatterino {
 
-BadgePredicate::BadgePredicate(const QStringList &badges)
+BadgePredicate::BadgePredicate(const QString &badges, bool negate)
+    : MessagePredicate(negate)
 {
     // Check if any comma-seperated values were passed and transform those
-    for (const auto &entry : badges)
```
kornes, line 10: this outer loop was dropped with change of `QStringList` param to `QString`. Previously searchPopup would bundle predicates of same type into one list, but since now each one can be separately negated, it can't be kept.
From user standpoint this changes nothing, comma separated lists typed into search will work the same.

src/messages/search/AuthorPredicate.cpp
```diff
@@ -4,16 +4,14 @@
...
 namespace chatterino {
 
-AuthorPredicate::AuthorPredicate(const QStringList &authors)
-    : authors_()
+AuthorPredicate::AuthorPredicate(const QString &authors, bool negate)
+    : MessagePredicate(negate)
+    , authors_()
 {
     // Check if any comma-seperated values were passed and transform those
-    for (const auto &entry : authors)
```
kornes, line 11: this outer loop was dropped with change of `QStringList` param to `QString`. Previously searchPopup would bundle predicates of same type into one list, but since now each one can be separately negated, it can't be kept.
From user standpoint this changes nothing, comma separated lists typed into search will work the same.

pajlada (changes requested):
I'm happy with actual implementation now, but I'd like to look at how the negation is applied

Right now, to search for people who are NOT T3 or T2 subbed, you would write `subtier:!3,2` which feels wrong, the `!` is placed in a way where it should apply to one of the subtiers only.

Should it be
a. `subtier:!3,2` (current implementation)
b. `!subtier:3,2` - Negation before the predicate name
c. `subtier:!3,!2` - Negation before each predicate parameter (would allow search for users with one badge but not another badge in the same "check" `badge:mod,!sub`)

I feel like b is clearest, but I'd like to hear other people's opinions if they disagree, or if they have any other alternatives

Felanbird:
Following the existing standard we have shown users, via filters, `B` seems the most correct to me.

pajlada (approved):

PR #4209, opened 2022-12-02
fix: Copy Shortcut Not Working In Reply Thread Popup

### LINKED ISSUE #4208: CTRL+C on text selection doesn't work in reply thread window consistently
### Checklist

### Describe your issue

This will be a bit difficult to show in screen recordings, but assume every time I click on the input field, I'm pressing CTRL+V or right-clicking and pasting via context menu. 

When doing manual selection or triple clicking, CTRL+C does nothing:
chatterino_fgmbnDpsPL

However, double-click on a single word works fine:
chatterino_TjHKxQLDN5

Making a selection then copying via context menu also works fine:
chatterino_k7zefQfzIo

Also possibly related - the input field happens to be unfocused when CTRL+C works, but I'd usually expect the input field to always be focused like it works in normal channel views.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.4.0 (commit [1dd06933](https://github.com/Chatterino/chatterino2/commit/1dd06933) modified) built on 2022-11-29 with Qt 5.15.2, Windows SDK, MSVC 193331631
Microsoft Windows 11 [Version 10.0.25247.1000]

Nerixyz:
Pull request checklist:

# Description

Previously, pressing CTRL + C would attempt to copy from the original split (→ and its channel view) instead. This PR fixes the bug.

Fixes #4208.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed CTRL + C not working in reply thread popups. (#4209)
+
 ## 2.4.0
 
 - Major: Added support for emotes, badges, and live emote updates from [7TV](https://7tv.app). [Wiki Page](https://wiki.chatterino.com/Third_party_services/#7tv) (#4002, #4062, #4090)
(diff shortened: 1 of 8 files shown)


Felanbird (approved):
works 👍

sando:
Tested, and this fixes #4208, thanks! ❤️

pajlada (approved):

PR #4216, opened 2022-12-05
feat: Automatically select newly added table rows

kornes:
Pull request checklist:

# Description
Pressing `Add` button above tables in settings window adds new row at the bottom. 
When table has scrollbar, new rows are not visible right away and its not clear if Add button even worked.

changes introduced in PR:
- new rows are automatically selected (and scrolled to when not visible) in all settings tables
- removed now redundant selections in Hotkeys (`KeyboardSettingsPage`)
- unnecessary `EditableModelView::selectRow(int row)` is removed and `QTableView::selectRow(int row)` is used internally instead. Functionality is still exposed via existing public api: `EditableModelView::getTableView()->selectRow()`

### DIFF
--- a/src/widgets/settingspages/KeyboardSettingsPage.cpp
+++ b/src/widgets/settingspages/KeyboardSettingsPage.cpp
@@ -40,13 +40,6 @@ KeyboardSettingsPage::KeyboardSettingsPage()
             auto newHotkey = dialog.data();
             int vectorIndex = getApp()->hotkeys->hotkeys_.append(newHotkey);
             getApp()->hotkeys->save();
-
-            // Select and scroll to newly added hotkey
-            auto modelRow = model->getModelIndexFromVectorIndex(vectorIndex);
-            auto modelIndex = model->index(modelRow, 0);
-            view->selectRow(modelRow);
-            view->getTableView()->scrollTo(modelIndex,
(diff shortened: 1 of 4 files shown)


src/widgets/settingspages/KeyboardSettingsPage.cpp
```diff
@@ -40,13 +40,6 @@ KeyboardSettingsPage::KeyboardSettingsPage()
             auto newHotkey = dialog.data();
             int vectorIndex = getApp()->hotkeys->hotkeys_.append(newHotkey);
             getApp()->hotkeys->save();
-
-            // Select and scroll to newly added hotkey
-            auto modelRow = model->getModelIndexFromVectorIndex(vectorIndex);
-            auto modelIndex = model->index(modelRow, 0);
-            view->selectRow(modelRow);
-            view->getTableView()->scrollTo(modelIndex,
-                                           QAbstractItemView::PositionAtCenter);
```
kornes, line 49: previously this add action would scroll further than needed to put new item at the center. Decided to not preserve this and keep all tables scroll-to-new behave the same. 
Both approaches are as good when it comes to visibility of newly added item.

pajlada (approved):
Nice clean fix, thank you!

PR #4218, opened 2022-12-06
Fix selection clearing not working in Reply window

kornes:
Pull request checklist:

# Description

Currently Reply window is missing behaviour of main chatterino window, where selection in splitInput and channelView cancels each other. 

PR changes:
- SplitInput will now expose `SplitInput::selectionChanged` signal, instead of doing any clearing locally 
- ~~moved `ChannelView::selectionChanged` signal to Qt for consistency~~
- use both signals in Reply window where it was missing

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,11 +17,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-20.04, macos-latest]
         qt-version: [5.15.2, 5.12.12]
         pch: [true]
         include:
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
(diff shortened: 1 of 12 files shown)


pajlada (changes requested):
Please undo the changes related to `moved ChannelView::selectionChanged signal to Qt for consistency` unless you have some other reasoning for it

kornes:
Main reasoning for this change was to have handling of both signals use same solution since they are next to each other, personally i prefer Qt because its connect() better show both sides of connection and improve readability.
With Qt overall progress and Qt6 bringing new nice things like one-shot connections, imo we should encourage using native solution in new code and fallback to alternative only when there is benefit.
And finally, while not really in scope, this change doesn't have side effects as changed signal is used just in this 1 place.

If my excuses didn't convince you, im fine with reverting it, $notify me or let me know here :D

pajlada:
Yeah please undo the change for now, can review at some other point

pajlada (approved):

PR #4224, opened 2022-12-07
feat: Add Setting to Hide Reply Context

Nerixyz:
Pull request checklist:

# Description

As discussed in #4220:

> Some people don't like reply threads or how they're displayed in Chatterino and want to "hide replies" - hide the reply context.

> My thinking was to add a setting that hides the reply context in the rendered messages, so these messages look like regular messages. I don't think adding a setting that completely ignores replies is good, as it will overcomplicate a lot of the code.

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -70,7 +70,7 @@ MessagePtr generateBannedMessage(bool confirmedBan)
 
 int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
 {
-    if (!getSettings()->stripReplyMention)
+    if (!getSettings()->stripReplyMention || !getSettings()->showReplyContext)
     {
         return 0;
     }
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1124,9 +1124,11 @@ MessageElementFlags ChannelView::getFlags() const
     if (this->sourceChannel_ == app->twitch->mentionsChannel)
         flags.set(MessageElementFlag::ChannelName);
 
-    if (this->context_ == Context::ReplyThread)
+    if (this->context_ == Context::ReplyThread ||
+        !getSettings()->showReplyContext)
     {
         // Don't show inline replies within the ReplyThreadPopup
+        // or if they're disabled
         flags.unset(MessageElementFlag::RepliedMessage);
     }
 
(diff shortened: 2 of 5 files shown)


src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -70,7 +70,7 @@ MessagePtr generateBannedMessage(bool confirmedBan)
 
 int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
 {
-    if (!getSettings()->stripReplyMention)
+    if (!getSettings()->stripReplyMention || !getSettings()->showReplyContext)
```
pajlada, line 73: I think this is more readable
```suggestion
    if (!getSettings()->stripReplyMention && getSettings()->showReplyContext)
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -1124,9 +1124,11 @@ MessageElementFlags ChannelView::getFlags() const
     if (this->sourceChannel_ == app->twitch->mentionsChannel)
         flags.set(MessageElementFlag::ChannelName);
 
-    if (this->context_ == Context::ReplyThread)
+    if (this->context_ == Context::ReplyThread ||
+        !getSettings()->showReplyContext)
```
pajlada, line 1128: Same thing here - I think this is more readable

```suggestion
    if (this->context_ == Context::ReplyThread &&
        getSettings()->showReplyContext)
```

pajlada (changes requested):

src/providers/twitch/IrcMessageHandler.cpp
Nerixyz: I suppose you mean `!(getSettings()->stripReplyMention && getSettings()->showReplyContext)`.

src/widgets/helper/ChannelView.cpp
Nerixyz: Same here: `!(this->context_ != Context::ReplyThread && getSettings()->showReplyContext)`. Though in this case I think this is less readable.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -843,10 +847,11 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
                        s.highlightMentions);
-    layout.addCheckbox("Strip leading mention in replies", s.stripReplyMention,
-                       true,
-                       "When disabled, messages sent in reply threads will "
-                       "include the @mention for the related thread");
+    layout.addCheckbox(
+        "Strip leading mention in replies", s.stripReplyMention, true,
+        "When disabled, messages sent in reply threads will "
+        "include the @mention for the related thread. If the rendering of the"
+        "reply context is turned off, these mentions will never be stripped.");
```
pajlada, line 854:
```suggestion
        "include the @mention for the related thread. If the rendering of the "
        "reply context is turned off, these mentions will never be stripped.");
```

pajlada (approved):

PR #4226, opened 2022-12-08
fix: don't run `push-aur.yml` on pull requests

Nerixyz:
Pull request checklist:

# Description

The action will always use the (upstream) master branch and never actually build the PR itself. 

It's not really clear from the name of the action what it will do - maybe it should be renamed?


pajlada (approved):

pajlada:
Idea was for it to push to AUR, it will probably do this in the future too (pushing an appimage)

PR #4045, opened 2022-10-04
Make menus and placeholders display appropriate custom key combos.

Mm2PL:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

Closes #4039.
Todo:
- [x] make `neededArguments` work
- [-] implement it everywhere:
  - [x] SplitHeader 

  - [-] Settings
    - [x] search placeholder
    - [x] Hide similar messages setting
    - [x] prefs button setting
  - [x] Notebook (hide tabs)
  - [ ] ~~SplitInput~~ that's a Qt menu.
  - [ ] something more?
- [ ] Consider showing multiple shortcuts when possible (?)
- [x] Update documentation

All the commits mentioning scuffed-ness need some kind of relayout or app restart to update the key bind description.

PR #4230, opened 2022-12-12
Add even more tooltips when hovering settings

Felanbird:
Pull request checklist:

# Description

I believe I remembered to test every setting, to make sure nothing like #4225 happens again 🦒

### DIFF
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -818,8 +818,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addCheckbox("Show Twitch whispers inline", s.inlineWhispers, false,
                        "Show whispers as messages in all splits instead "
                        "of just /whispers.");
-    layout.addCheckbox("Highlight received inline whispers",
-                       s.highlightInlineWhispers);
+    layout.addCheckbox(
+        "Highlight received inline whispers", s.highlightInlineWhispers, false,
+        "Highlight the whispers shown in all splits.\nIf \"Show Twitch "
+        "whispers inline\" is disabled, this setting will do nothing.");
     layout.addCheckbox("Load message history on connect",
                        s.loadTwitchMessageHistoryOnConnect);
     // TODO: Change phrasing to use better english once we can tag settings, right now it's kept as history instead of historical so that the setting shows up when the user searches for history
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -232,10 +232,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                            s.hidePreferencesButton, true);
         layout.addCheckbox("Show user button", s.hideUserButton, true);
     }
-    layout.addCheckbox("Show which channels are live in tabs", s.showTabLive,
-                       false,
+    layout.addCheckbox("Mark tabs with live channels", s.showTabLive, false,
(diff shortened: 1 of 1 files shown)


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -194,8 +194,12 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
         "This setting will only affect how messages are shown. You can reply "
         "to a message regardless of this setting.");
-    layout.addCheckbox("Show message reply button", s.showReplyButton);
-    layout.addCheckbox("Show tab close button", s.showTabCloseButton);
+    layout.addCheckbox("Show message reply button", s.showReplyButton, false,
+                       "Show a reply button next to every chat message");
+    layout.addCheckbox(
+        "Show tab close button", s.showTabCloseButton, false,
+        "When disabled, the x to close a tab will be hidden.\nTabs can still "
+        "be closed by right-clicking or CTRL+SHIFT+W");
```
Mm2PL, line 202: This tooltip should mention the defaultness of this or preferably use HotkeyController::getDisplaySequence introduced in #4045:

https://github.com/Chatterino/chatterino2/pull/4045/files#diff-d39bf5472005a5ba68ffdb09e75289913c7f1dd9b0285fd6acd44c799ac6771dR77-R86

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -318,7 +331,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addSeperator();
     layout.addCheckbox("Draw a line below the most recent message before "
                        "switching applications.",
-                       s.showLastMessageIndicator);
+                       s.showLastMessageIndicator, false,
+                       "Adds a red underline below the most recent message "
```
Mm2PL, line 335: Line color is configurable, having this mention red even when it's not certain might be confusing.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -779,7 +804,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
             [](auto args) {
                 return args.index + 1;
             },
-            false);
+            false,
+            "Customizes how you see Asian Language names.\ne.g. "
+            "testaccount_420(테스트계정420) -> tesaccount_420\n"
+            "testaccount_420(테스트계정420) -> 테스트계정420");
```
Mm2PL, line 810: Could you add the option value names to this? Not sure if users would know what a localized name is, maybe this could be better explained?

Mm2PL (changes requested):

pajlada (changes requested):
Small things

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -218,7 +232,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                            s.hidePreferencesButton, true);
         layout.addCheckbox("Show user button", s.hideUserButton, true);
     }
-    layout.addCheckbox("Show which channels are live in tabs", s.showTabLive);
+    layout.addCheckbox("Show which channels are live in tabs", s.showTabLive,
+                       false,
+                       "Shows a red dot in the top right corner of a tab to "
+                       "indicate they are live.");
```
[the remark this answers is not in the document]
Felanbird: image

This is the name I have but I don't like it, as this name sounds like it toggles on the entire live functionality in Chatterino

pajlada (approved):

PR #4246, opened 2022-12-18
Only try to extract images if the image uploader is enabled

### LINKED ISSUE #4245: Pasting cells from Excel misbehaving
Original conversation here: https://discord.com/channels/253897423439659008/459650160658481152/1054159287087276062

Original post:
> Excel has... weird... copy + paste shenanigans it does to try and maintain cell information. When I try to copy from Excel to Chatterino the paste just doesn't show up at all. My workaround for this is to use office.com which has a browser-based version of Excel that works fine. Google Sheets works fine as well. Also, copying from Excel to other stuff (say, Notepad) works fine. But I was curious if anyone knew of a way that I could use Excel and paste into Chatterino?

Single-cell & multi-cell copying fails with no error message.
Potentially due to us trying to upload it as an image but image uploading is disabled

OS: Windows

Application: Excel Version: Microsoft® Excel® for Microsoft 365 MSO (Version 2211 Build 16.0.15831.20098) 64-bit

Example file: [TestBook1.xlsx](https://github.com/Chatterino/chatterino2/files/10254877/TestBook1.xlsx)

pajlada:
Pull request checklist:

# Description

Tries to be a bit smarter for when to just paste regular old text instead of an image

Needs some thorough testing with mixed clipboards

Fixes #4245

To test:
1) Open excel/openoffice
2) Copy 1 or more cells
3) Paste into Chatterino with an image uploader configured. It should, atm, post an image of it
4) Disable your image uploader and paste it into Chatterino again. Before this PR it would just not paste anything, after this PR it should paste the text of the cells

Tested on Windows, macOS, Linux X11
Not tested on Wayland, someone can do that and make another PR if this breaks it (although I don't think it would)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Bugfix: Fixed unnecessary saving of windows layout. (#4201)
 - Bugfix: Fixed Reply window missing selection clear behaviour between chat and input box. (#4218)
 - Bugfix: Fixed crash that could occur when changing Tab layout and utilizing multiple windows. (#4248)
+- Bugfix: Fixed text sometimes not pasting properly when image uploader was disabled. (#4246)
 - Dev: Remove protocol from QApplication's Organization Domain (so changed from `https://www.chatterino.com` to `chatterino.com`). (#4256)
 - Dev: Ignore `WM_SHOWWINDOW` hide events, causing fewer attempted rescales. (#4198)
 - Dev: Migrated to C++ 20 (#4252, #4257)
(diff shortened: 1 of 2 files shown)


Mm2PL (approved):
Works as expected on Linux X11!

Felanbird (approved):
Fix works on 🅱️indows, able to paste excel-style text when the image uploader is disabled.

gempir:
tested on macOS Ventura 13.1

When disabled. Pastes Excel style text
image

When enabled fails with 
image

pajlada:
Going to merge this in as is - there's some more logic we could do to ensure text always pastes but that would fit better as a feature enhancement than a bug fix

leon-richardt:
Can confirm this works on sway v1.7 as described in the OP!

PR #4219, opened 2022-12-07
Change ubuntu-latest to ubuntu-20.04 in CI builds

Wissididom:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

For a consitently build. Chatted with pajlada in Discord [here](https://discord.com/channels/253897423439659008/459650160658481152/1050000444988072007)

PR #4252, opened 2022-12-23
Migrate to C++ 20 & switch to websocketpp develop branch

Nerixyz:
Pull request checklist:

# Description

All current compilers support a majority of the C++ 20 language and library features - [Implementation status](https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_features).

(Apple) Clang seems to be the odd one out by supporting less than GCC and MSVC. Specifically, ranges and concepts aren't fully implemented yet. 

There aren't many unavoidable migrations:

1. **Implicit capture of `this` in lambdas via `[=]` is deprecated** (see [Lambda Expressions/Lambda capture](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture)). This was the most common issue I found. In a lot of places, `this` was captured via `[=]`. I replaced it with `[this]` where only `this` was used, and `[=, this]` where additional variables were captured.
2. **The type of a `u8"..."` string literal is `const char8_t[N]`** (see [String Literal](https://en.cppreference.com/w/cpp/language/string_literal)). There is only one place where a UTF-8 string is used - in `TwitchChannel`.
3. Windows specific: The `min` and `max` macros are defined through `minwindef.h`, which is included though `Windows.h`. This breaks some usage of `std::min` and `std::max`. One could mitigate this by using `(std::min)(a, b)` over `std::min(a, b)`, but this isn't ideal, since you'd only get an error if `Windows.h` was included before - which isn't always the case. Following [this StackOverflow answer](https://stackoverflow.com/a/22744273/16300717), I added `NOMINMAX` as a definition on Windows.
3.1 MSVC didn't generate as many warnings as GCC and Clang. I went ahead and enabled a bit more warnings and added a reason as to why the disabled warnings are disabled.

As for the CI, I think `ubuntu-20.04` should be upgraded to `ubuntu-latest` - assuming the issue mentioned in #4219 is resolved.

I'm opening this as a draft to create a place for discussion about the adoption.


pajlada (changes requested):
Looks good, just a nitpick for the magic character.

Right now though, this doesn't build on GCC12 in C++20 mode, probably because it's more strict where template parameters can be put. Our websocketpp logger & the websocketpp logger/basic fails to compile for me due to the constructors being defined as
`basic<concurrency, names>(...)` rathern than just `basic(...)`.
This has been fixed in the `develop` branch of websocketpp so we'd have to migrate to that first

pajlada:
PR'd some fixes that were necessary for me https://github.com/Nerixyz/chatterino2/pull/8

pajlada (approved):

PR #4258, opened 2022-12-24
Enable LTO support for Chatterino builds

pajlada:
Pull request checklist:

# Description

This is disabled by default, and can be enabled by passing `-DCHATTERINO_LTO=On` to your cmake invocation.
I plan on enabling only for 1 of the PR builds, and for all PR builds on the main branch to not slow down CI too much


pajlada:
Looks good to me
image

zneix:
sorry for necro but for future reference, this was always disabled prior to the changes introduced in this PR, right?

PR #4261, opened 2022-12-24
SplitContainer refactor

pajlada:
Pull request checklist:

# Description

- Remove unused include util/Helpers.hpp
- SplitContainer::setTag fix parameter naming
- autofy/constify where possible
- Make SplitNode::Type an enum class
- Make SplitNode::Type an enum class
- Move QuickSwitcherPopup includes from header to source file
- Remove unused DropRegion code
- use empty() instead of size() == 0
- Add curly braces everywhere
- More const auto ptr magicifying
- Remove useless reinterpret_cast
- Clarify that the connect is QObject::connect
- SplitContainer::setSelected fix parameter naming
- Rename function variables to remove unneccesary underscore
- emplace_back where possible
- Name parameters
- Remove ineffective const from return type
- Make node getters const
- Flatten Node::releaseSplit
- Rename in-function variable to match code style
- [ACTUAL CODE CHANGE/MOVE] Move clamp logic to its own function
- name params
- applyFromDescriptorRecursively: rename node to baseNode
- Remove the many overloads for append/insertSplit

### DIFF
--- a/src/widgets/splits/DraggedSplit.hpp
+++ b/src/widgets/splits/DraggedSplit.hpp
@@ -1,0 +1,15 @@
+#pragma once
+
+#include "widgets/splits/Split.hpp"
+
+namespace chatterino {
+
+// Returns true if the user is currently dragging a split in this Chatterino instance
+// We need to keep track of this to ensure splits from other Chatterino instances aren't treated as memory we own
+[[nodiscard]] bool isDraggingSplit();
+
+// Set the current split dragging state
+// Controlled by Split's drag function
+void setDraggingSplit(bool isDraggingSplit);
+
+}  // namespace chatterino
--- a/src/widgets/splits/DraggedSplit.hpp
+++ b/src/widgets/splits/DraggedSplit.hpp
@@ -6,8 +6,12 @@ namespace chatterino {
 // We need to keep track of this to ensure splits from other Chatterino instances aren't treated as memory we own
 [[nodiscard]] bool isDraggingSplit();
 
-// Set the current split dragging state
-// Controlled by Split's drag function
-void setDraggingSplit(bool isDraggingSplit);
+// Set that a split is currently being dragged
+// Used by the Split::drag function when a drag is initiated
+void startDraggingSplit();
+
+// Set that a split is no longer being dragged
+// Used by the Split::drag function when a drag is finished
+void stopDraggingSplit();
 
 }  // namespace chatterino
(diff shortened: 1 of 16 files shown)


pajlada:
This has three actual code changing commits, the rest is just cosmetic changes.

Biggest change is changes to dropping splits, instead of storing the split being dragged in a pointer we now just read it from the drop event. I also moved that logic out of splitcontainer to its own file where I felt like I could document its purpose a bit more. I'm not super happy with the naming of those two functions but that can more easily be changed now that it has its own little home.

src/widgets/splits/DraggedSplit.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+namespace chatterino {
+
+// Returns true if the user is currently dragging a split in this Chatterino instance
+// We need to keep track of this to ensure splits from other Chatterino instances aren't treated as memory we own
+[[nodiscard]] bool isDraggingSplit();
+
+// Set the current split dragging state
+// Controlled by Split's drag function
+void setDraggingSplit(bool isDraggingSplit);
```
pajlada, line 13: Look into the naming of this function

src/widgets/splits/SplitContainer.hpp
```diff
@@ -110,6 +113,9 @@ class SplitContainer final : public BaseWidget
                     std::vector<DropRect> &dropRects_,
                     std::vector<ResizeRect> &resizeRects);
 
+        // Clamps the flex values ensuring they're never below 0
+        void clamp();
```
pajlada, line 117: Look to use `std::clamp` instead

src/widgets/splits/SplitContainer.cpp
```diff
@@ -136,41 +128,48 @@ Split *SplitContainer::appendNewSplit(bool openChannelNameDialog)
...
+    {
+        // options.relativeNode must not be set together with relativeSplit
+        assert(!options.relativeNode);
 
-    assertInGuiThread();
+        Node *node =
+            this->baseNode_.findNodeContainingSplit(options.relativeSplit);
+        assert(node != nullptr);
+
+        InsertOptions opts{};
```
pajlada, line 158: Should be removed, was used for testing

src/widgets/splits/SplitContainer.hpp
pajlada: Using `std::max` in the function instead

src/widgets/splits/DraggedSplit.hpp
pajlada: I split this up into two functions instead, it was the clearest alternative I could find

PR #4266, opened 2022-12-27
fix: Remove Unused Include Directives

Nerixyz:
Pull request checklist:

# Description

**This doesn't remove all unused include directives.** I only checked files that had imports that were only present on Windows (`#ifdef USEWINSDK`). Notably, this removes `gdiplus.h` from `BaseWindow.h` which sometimes relied on the `min` and `max` macros defined through `Windows.h`.

### DIFF
--- a/src/util/StreamerMode.cpp
+++ b/src/util/StreamerMode.cpp
@@ -13,6 +13,13 @@
 #    include <WtsApi32.h>
 // clang-format on
 #    pragma comment(lib, "Wtsapi32.lib")
+#else
+#    include "Application.hpp"
+#    include "messages/MessageBuilder.hpp"
+#    include "providers/twitch/TwitchIrcServer.hpp"
+#    include "widgets/Notebook.hpp"
+#    include "widgets/splits/Split.hpp"
+#    include "widgets/Window.hpp"
 #endif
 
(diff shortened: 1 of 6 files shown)


src/util/StreamerMode.cpp
```diff
@@ -1,28 +1,29 @@
...
+#    pragma comment(lib, "Wtsapi32.lib")
+#else
+#    include "Application.hpp"
+#    include "messages/MessageBuilder.hpp"
+#    include "providers/twitch/TwitchIrcServer.hpp"
+#    include "widgets/Notebook.hpp"
+#    include "widgets/splits/Split.hpp"
+#    include "widgets/Window.hpp"
+
+#    include <QProcess>
```
pajlada, line 24: I would rather not use an else-scope for includes like this unless strictly necessary as I believe it reduces maintainability/readability.

Essentially how I evaluate it is: If the include is only required on a specific platform but doesn't break the build of other platforms, just include them always in the main build.

pajlada (changes requested):
Small nitpick with the `#else` scope in StreamerMode.cpp, otherwise looks good

pajlada (approved):

PR #4268, opened 2022-12-28
ref: merge TooltipPreviewImage and TooltipWidget

kornes:
Pull request checklist:

# Description
TooltipPreviewImage serves little purpose and depends on TooltipWidget. Since both are signletones doing same thing and used together in codebase, this PR removes TooltipPreviewImage and moves image refreshing logic to TooltipWidget.

changes to merged TooltipWidget: 
- removed unused imports
- added missing parent to QLabels
- set `Qt::WindowStaysOnTopHint` attribute at construction, instead of Qt raise() and windows' HWND_TOPMOST
- use SignalHolder for 3 connections and change destructor to default one
- moved adjustSize() (needed for background sizing) to local showEvent()
- changed external `setImage(nullptr)` calls to `clearImage()`
- setImage() will exit early when passed image matches current one (happens when mouse move over same emote)
- rename refreshTooltipWidgetPixmap() to refreshPixmap()
- flattened refreshPixmap()

also found minor bug in current master, where non-animated images doesn't show up in tooltip from first hovered pixel and requires further mouse move. this PR fixes that.


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Dev: Migrated to C++ 20 (#4252, #4257)
 - Dev: Enable LTO for main branch builds. (#4258, #4260)
 - Dev: Removed unused include directives. (#4266)
+- Dev: Removed TooltipPreviewImage. (#4268)
```
pajlada, line 24: Make an entry for the bugfix

src/widgets/TooltipWidget.cpp
```diff
@@ -25,44 +19,52 @@ TooltipWidget::TooltipWidget(BaseWidget *parent)
...
     layout->addWidget(displayImage_);
     layout->addWidget(displayText_);
     this->setLayout(layout);
 
-    this->fontChangedConnection_ = getFonts()->fontChanged.connect([this] {
+    this->connections_.managedConnect(getFonts()->fontChanged, [this] {
         this->updateFont();
     });
-}
+    this->updateFont();
```
pajlada, line 47: Can you verify that `updateFont` isn't called twice here? `connect` functions to signals and settings should, by default, call the callback once unless another parameter is set to `false` (can't remember its name atm)

EDIT: I realize `updateFont` was already called like this here, you've just moved the call down so feel free to just add a comment here if you don't want to handle this in this PR

src/widgets/TooltipWidget.cpp
```diff
@@ -100,12 +102,64 @@ void TooltipWidget::setWordWrap(bool wrap)
...
+        this->displayImage_->setPixmap(pixmap->scaled(
+            this->customImgWidth, this->customImgHeight, Qt::KeepAspectRatio));
+    }
+    else
+    {
+        this->displayImage_->setPixmap(*pixmap);
+    }
     this->displayImage_->show();
-    this->displayImage_->setPixmap(image);
+    return true;
```
pajlada, line 162:
```suggestion

    return true;
```

Formatting nitpick

pajlada (changes requested):
Overall change looks good, good refactor. Just some nitpicks, then I'd like to take this for a spin with some real testing

src/widgets/TooltipWidget.cpp
kornes: confirmed its not called twice, only settings signals have autoInvoke from what i see, so will leave it in current form in this PR

pajlada (approved):
Changes look good, I assume you've tested that things work as expected on Windows, now I just want to ensure things work as expected on Linux X11 and then I'm happy to merge this in

kornes:
did tests on windows and wayland with always on top, tooltip worked as expected 
but would good if someone with macOS to test it too before merge

kornes:
pushed fix to bug i overlooked 
https://user-images.githubusercontent.com/28986062/209969031-4ec2618b-b560-4406-99a3-2bb20d9b4de3.mp4

Whats wrong
hide and show events wont trigger if 2 chat elements that are showing tooltip are next to eachother with no space between. 
When mouse move over to 2nd element, image was correctly replaced, but custom size did not reset (due to no hideEvent) and possible new size of 2nd element was not applied (due to pixmapRefresh being in showEvent)

Fix: 
1. moved pixmap refreshing from showEvent to `setImageScale()`, early exit if there is no change
2. always reset custom scale to 0x0 in `setImage()` with clearImage()

pajlada:
Works as expected for me so I'm merging this in, thank you!

PR #4271, opened 2022-12-29
Fix crash that would occur when performing certain actions after removing all tabs

### LINKED ISSUE #4269: Crash when all tabs are closed
### Steps to Reproduce
1. Open any amount of tabs, or utilize the single default tab
2. Close all tabs (CTRL+SHIFT+W)
3. Attempt to open a new tab (➕ or CTRL+SHIFT+T)
4. ⚰️ 

stacktrace(s):
https://paste.ivr.fi/catemuryvu // this one happens when pressing `➕ or CTRL+SHIFT+T`
https://paste.ivr.fi/sijuxunegu.r // this one happens when pressing `CTRL+T` in the same scenario

Explicitly tested on 2ba4da02aecf937c740ef9b0a9f7bd7b8c3ef256 but I doubt we've fixed this issue since last week

pajlada:
Pull request checklist:

# Description

Some timers would crash, and actions like CTRL+T that act on "the selected tab" would cause a crash.

This crash was introduced in https://github.com/Chatterino/chatterino2/pull/4187 - and because it has proper reproduction steps, I've tested those after this fix too to ensure we don't regress.

Fixes #4269

### DIFF
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -1172,7 +1172,7 @@ SplitContainer *SplitNotebook::getOrAddSelectedPage()
 
     if (selectedPage)
     {
-        return (SplitContainer *)(selectedPage);
+        return dynamic_cast<SplitContainer *>(selectedPage);
     }
 
     return this->addPage();
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -156,49 +156,50 @@ int Notebook::indexOf(QWidget *page) const
 
 void Notebook::select(QWidget *page, bool focusPage)
 {
-    if (!page)
-    {
-        return;
-    }
-
     if (page == this->selectedPage_)
     {
+        // Nothing has changed
         return;
     }
 
-    auto *item = this->findItem(page);
-    if (!item)
+    if (page)
     {
-        return;
-    }
-
-    page->show();
-
-    item->tab->setSelected(true);
-    item->tab->raise();
-
-    if (focusPage)
-    {
-        if (item->selectedWidget == nullptr)
+        // A new page has been selected, mark it as selected & focus one of its splits
+        auto *item = this->findItem(page);
+        if (!item)
         {
-            item->page->setFocus();
+            return;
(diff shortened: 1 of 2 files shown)


src/widgets/Notebook.cpp
```diff
@@ -1169,22 +1170,29 @@ SplitContainer *SplitNotebook::getOrAddSelectedPage()
 {
     auto *selectedPage = this->getSelectedPage();
 
-    return selectedPage != nullptr ? (SplitContainer *)selectedPage
-                                   : this->addPage();
+    if (selectedPage)
+    {
+        return (SplitContainer *)(selectedPage);
```
leon-richardt, line 1175:
```suggestion
        return static_cast<SplitContainer *>(selectedPage);
```

pajlada: We'd technically want this to be a dynamic cast, I'll look to see if messy it would be to chance the stored type from a `QWidget*` to a `SplitContainer*`

leon-richardt: Aren't we only ever storing `SplitContainer *`s anyway? (I think I recall that assumption being made in different places in the code as well.) If so, I believe  a `static_cast` would be correct

pajlada: It'll be quite a big change, basically merging the "Notebook" and "SplitNotebook" classes, so I'd prefer to defer this change to a refactor PR to keep this fix as small as possible. I only changed something here to make the code I was looking through readable

pajlada: nvm, Notebook is used in 3 places: 1. as a way to hold splitcontainers (the regular use). 2. controlling tabs in the "new channel" or "edit channel" popup. 3. the emote popup uses it. so the refactor would not merge the two, but rather look at what specializations make sense to keep in Notebook and what should be moved to SplitNotebook (which manages usecase 1.)

Felanbird (approved):
Crash is resolved, functionality of said various actions seems to be the same as before this crash was introduced. 👍

PR #4275, opened 2022-12-31
Remove Unnecessary Includes in Headers

Nerixyz:
Pull request checklist:

# Description

This PR removes unused includes in header files and moves unnecessary once to the `.cpp` files. "Unnecessary" is a broad term, so _here_ I've interpreted it as follows: everything that could be forward-defined.

This requires more includes in `.cpp` files. But this might make it easier for static tools to figure out if an include is unused.

Some changes might be too harsh, so I'm open to reverting some changes.

#### Further Work

* Improve/use `ForwardDecl.h` instead of manual declarations
* Static tools like [`include-what-you-use`](https://github.com/include-what-you-use/include-what-you-use) _could_ be considered, though, for me, it only reported something in `main.cpp`.


src/widgets/splits/SplitContainer.hpp
```diff
@@ -21,6 +21,10 @@ namespace chatterino {
 class Split;
 class NotebookTab;
 class Notebook;
+struct ContainerNodeDescriptor;
+struct SplitNodeDescriptor;
+using NodeDescriptor =
+    std::variant<ContainerNodeDescriptor, SplitNodeDescriptor>;
```
pajlada, line 27: Questionable, I think we're best off just including WindowDescriptors.hpp here since the file is so small & this leaks the implementation a bit

src/singletons/Logging.hpp
```diff
@@ -1,14 +1,16 @@
...
 
 #include <memory>
 
 namespace chatterino {
 
 class Paths;
+struct Message;
+using MessagePtr = std::shared_ptr<const Message>;
+class LoggingChannel;
+class LoggingChannel;
```
pajlada, line 13:
```suggestion
class LoggingChannel;
```

src/controllers/ignores/IgnorePhrase.cpp
```diff
@@ -112,4 +113,11 @@ bool IgnorePhrase::containsEmote() const
     return !this->emotes_.empty();
 }
 
+IgnorePhrase IgnorePhrase::createEmpty()
+{
+    return IgnorePhrase(
+        QString(), false, false,
+        ::chatterino::getSettings()->ignoredPhraseReplace.getValue(), true);
```
pajlada, line 120: The big namespace is no longer necessary since we're part of the namespace here

```suggestion
        getSettings()->ignoredPhraseReplace.getValue(), true);
```

src/controllers/filters/FilterSet.hpp
```diff
@@ -1,86 +1,30 @@
 #pragma once
```
pajlada, line 2: Let's include the very clearly used things here like QList and QMap to ensure we don't easily break things in the future

src/controllers/commands/builtin/twitch/ChatSettings.hpp
```diff
@@ -1,24 +1,27 @@
...
-QString followers(const CommandContext &ctx);
-QString followersOff(const CommandContext &ctx);
+    QString followers(const CommandContext &ctx);
+    QString followersOff(const CommandContext &ctx);
 
-QString uniqueChat(const CommandContext &ctx);
-QString uniqueChatOff(const CommandContext &ctx);
+    QString uniqueChat(const CommandContext &ctx);
+    QString uniqueChatOff(const CommandContext &ctx);
+}  // namespace commands
```
pajlada, line 25:
```suggestion
    QString uniqueChatOff(const CommandContext &ctx);

}  // namespace commands
```

pajlada (changes requested):
Small nitpicks, otherwise looks good. Thank you!

image

pajlada (approved):

PR #4276, opened 2022-12-31
Refactor SplitHeader class

pajlada:
Pull request checklist:

# Description

- Flatten static functions in anonymous namespace
- SplitHeader ctor: Rename param
- Header: Remove unnecessary `virtual`s
- auto ptr where possible
- Add curly braces
- Comment twitch room modes
- Treat roomModes->slowMode as an integer
- Remove unused `this` from lambdas
- Add `unsigned int` overload for localizeNumbers
- Move thumbnail max age to a constexpr static & set explicit types
- Explicitly use `QObject::connect`
- Use `empty()` instead of `size()`
- Name unused parameters
- Move moderation action refreshing logic from SplitHeader to Split

The only real code change here was moving the moderation action refresh logic to Split, and splitting that up into its own function for clarity. The previous `enable if enabled` was messy

### DIFF
--- a/src/providers/twitch/TwitchChannel.hpp
+++ b/src/providers/twitch/TwitchChannel.hpp
@@ -75,7 +75,21 @@ public:
         bool submode = false;
         bool r9k = false;
         bool emoteOnly = false;
+
+        /**
+         * @brief Number of minutes required for users to be followed before typing in chat
+         *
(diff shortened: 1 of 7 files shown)


pajlada:
Everything works as expected, looks good to me :+1: 

image

PR #4277, opened 2022-12-31
Fix crash that could occur when closing down splits

pajlada:
Pull request checklist:

# Description

Specifically, if a split was closed after the request for a thumbnail
had been made, but before the request had finished, we'd run into a
use-after-free issue


pajlada:
image

PR #4278, opened 2022-12-31
Fix crash that could occur when closing down the last of a channel when reloading emotes

pajlada:
Pull request checklist:

# Description

- Fix crash that could occur when a channel is closed
- Add changelog entry


pajlada:
image

Issue #2056, opened 2020-10-10
No Highlight Sound on Linux Mint


tintinmaster:
**Describe the bug**

No sound is played on a highlight, tested it on two different systems running Linux Mint 20 and deleted all Settings.
I also tested adding a custom sound.
It's not even displaying in "Applications playing" in the Sound Settings.
I don't remember when it started playing no sound, but I think at least since version 2.2.0 

**To reproduce**

Run Chatterino on Linux Mint 20 and get highlighted

**Screenshots**

**Chatterino version**

Chatterino 2.2.2 (commit 98762fdfa, 10.10.2020) using the AppImage

**Operating system**

OS: Linux Mint 20 x86_64
Kernel: 5.4.0-48-generic

**Additional information**

tintinmaster:
I get following console-output
```
(chatterino:8771): GStreamer-WARNING **: 23:52:42.430: Failed to load plugin '/usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgstcoreelements.so': /usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgstcoreelements.so: undefined symbol: gst_query_parse_bitrate

(chatterino:8771): GStreamer-WARNING **: 23:52:42.430: Failed to load plugin '/usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgstcoreelements.so': /usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgstcoreelements.so: undefined symbol: gst_query_parse_bitrate

(chatterino:8771): GLib-GObject-CRITICAL **: 23:52:42.430: g_object_set: assertion 'G_IS_OBJECT (object)' failed

(chatterino:8771): GStreamer-CRITICAL **: 23:52:42.430: gst_object_ref: assertion 'object != NULL' failed

(chatterino:8771): GStreamer-CRITICAL **: 23:52:42.430: gst_bin_add_many: assertion 'GST_IS_ELEMENT (element_1)' failed

(chatterino:8771): GStreamer-CRITICAL **: 23:52:42.430: gst_element_link_pads_full: assertion 'GST_IS_ELEMENT (src)' failed

(chatterino:8771): GStreamer-CRITICAL **: 23:52:42.430: gst_element_get_static_pad: assertion 'GST_IS_ELEMENT (element)' failed

[...]

PR #4285, opened 2023-01-04
Arch Linux ping sound issue fix/workaround

Greenlandicsmiley:
Pull request checklist:

# Description

This change fixes an issue at least on Arch Linux where after a few pings, Chatterino will no longer play the ping sound.

The issue was that the QT player loses the buffer for the media, or the media has been lost.
My change makes sure that if the media is not loading, loaded, stalled, buffering, or buffered, then the media for ping sound will be set.


src/controllers/notifications/NotificationController.cpp
```diff
@@ -98,20 +98,23 @@ void NotificationController::removeChannelNotification(
...
                   getSettings()->notificationPathSound.getValue())
             : QUrl("qrc:/sounds/ping2.wav");
 
-    if (currentPlayerUrl != highlightSoundUrl)
+    // set media if not loading, loaded, stalled, buffering, or fully buffered
+    if (!(player->mediaStatus() == QMediaPlayer::LoadingMedia ||
+          player->mediaStatus() == QMediaPlayer::LoadedMedia ||
+          player->mediaStatus() == QMediaPlayer::StalledMedia ||
+          player->mediaStatus() == QMediaPlayer::BufferingMedia ||
+          player->mediaStatus() == QMediaPlayer::BufferedMedia))
```
kornes, line 113: mediaStatus() will be either `QMediaPlayer::NoMedia` after QMediaPlayer was just created or `QMediaPlayer::EndOfMedia` at each successive call when ping sound finished [docs](https://doc.qt.io/qt-6.2/qmediaplayer.html#MediaStatus-enum), so both if's can be simplified.

also, setMedia() will be called at almost every ping now instead of only when audio file changed, would be cool to avoid that.

Greenlandicsmiley: You're right, I did some testing and found that when there's a ping, `mediaStatus()` will be `QMediaPlayer::NoMedia`, then when media is set it will be `QMediaPlayer::LoadingMedia`.

After a successful ping, `mediaStatus()` will be as you wrote `QMediaPlayer::EndOfMedia`.

After a few pings `mediaStatus()` becomes `QMediaPlayer::BufferedMedia` just right before `play()` is called and if media is not set, the ping sound will stop playing until media is set.

So the if statement should instead be:

```
    // set media if no media, or if media is buffered
    if (player->mediaStatus() == QMediaPlayer::NoMedia ||
        player->mediaStatus() == QMediaPlayer::BufferedMedia)
```

Right?

That should considerably reduce the amount of times `setMedia()` is called.

However, what I don't understand is why would the ping sound not play at all when `mediaStatus()` is to my understanding already buffered and ready to play again?
Or am I misunderstanding the role of `QMediaPlayer::BufferedMedia`?

kornes: tbh, hard for me to suggest good solution here because i dont know root cause of this bug and can't reproduce it on windows.
If your approach fixed ping sounds stopping, it would suggest calling `setMedia()` more often than needed is the key here, but its more of a workaround than addressing root cause. 
It could be many things, including qt bug or system problem, so would be best if more Arch users confirm issue and take a look into this.

pajlada: I'm fine if this solution is not a fix, but a stopgap solution. Qt6 has changed the multimedia API so we'll need to revisit this shortly anyway.
If this change doesn't break anything, only adding some more processing requirements on other platforms I'm ok with that tradeoff if it means pings will be more consistent

pajlada (approved):

pajlada:
:smiley: :mega: I am merging this in tomorrow

Thank you @Greenlandicsmiley

kornes:
it appears that `QMediaPlayer::BufferedMedia` state is only reached when 2nd notification start before previous finished playing, so current solution breaks runtime change of notification sound as it would require 2 overlapped notifications for `setMedia()` to happen and update sound.

Suggest to revert removal of currentPlayerUrl for now, and add check for BufferedMedia state
```
(currentPlayerUrl != highlightSoundUrl || player->mediaStatus() == QMediaPlayer::BufferedMedia)
```

as Pajlada said, qt6 heavily changed QMediaPlayer so it will be revisited soon ™️

Greenlandicsmiley:
> 
> Suggest to revert removal of currentPlayerUrl for now, and add check for BufferedMedia state
> 
> ```
> (currentPlayerUrl != highlightSoundUrl || player->mediaStatus() == QMediaPlayer::BufferedMedia)
> ```
> 

@kornes
Doesn't checking for `QMediaPlayer::NoMedia` replace `currentPlayerUrl`?
Because when `NotificationController::playSound()` and `SharedMessageBuilder::triggerHighlights()` are called for the first time, `currentPlayerUrl` will be empty, so therefore equivalent to `QMediaPlayer::NoMedia`, no?

Also @pajlada I forgot to implement the checking of when the ping sound is changed to a different audio file, so maybe hold the merge until there is a solution for that?
It might be a while though, because other than the changes i made, I do not have experience with C++ whatsoever.

pajlada:
All good, I'll hold

pajlada:
:smiley: :mega: I am merging this in this week - feedback time

Mm2PL:
Is this a fix for #2056?

Greenlandicsmiley:
Yeah so the title states Linux, which is very generic, I think this is more of an issue on Arch Linux, and seems like it is just Wayland.

I'm renaming the title to be more specific.

Felanbird:
@Greenlandicsmiley Thanks for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

To do so, open a new PR where you modify the resources/contributors.txt file and add yourself to the list. Make sure to read the comments at a top for instructions.

PR #4286, opened 2023-01-06
Use `libssl3` for Ubuntu builds

Felanbird:
Pull request checklist:

# Description

[SevenTV/chatterino7#172](https://github.com/SevenTV/chatterino7/issues/172#issuecomment-1372977457)

### DIFF
--- a/.CI/CreateUbuntuDeb.sh
+++ b/.CI/CreateUbuntuDeb.sh
@@ -28,7 +28,7 @@ Priority: optional
 Architecture: amd64
 Maintainer: Mm2PL <mm2pl@kotmisia.pl>
 Description: Testing out chatterino as a Ubuntu package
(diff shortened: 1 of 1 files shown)


Wissididom:
That would only work on 22.04 iirc because it is still 1.1 on 20.04 iirc. Will check later when I left bed.
Also I almost fixed the build and got it to work by bundling Qt and adjusting LD_LIBRARY_PATH. Just need to get the postinst script to work to set the library path variable automatically.

pajlada:
The correct approach to distributing `deb` files is to be specific which version we're targeting. Right now, we're targeting Ubuntu 20.04 with our deb, only changing the `libssl` version is not the right approach, we need to check all dependencies for a 22.04 release

Felanbird:
I'll be honest I was just taking a shot in the dark that this would resolve the issue, this isn't for me to fix as I do not have either env to test with. 🙂

PR #4287, opened 2023-01-06
Remove Deprecated `QDesktopWidget`

### LINKED ISSUE #1926: Emote picker is only appearing on the primary screen
**Describe the bug**
Whenever I open the emote picker it always opens up on my primary screen, even when I have previously opened, moved over and closed the emote picker on my secondary screen where I have chatterino open.

**To reproduce**
1. Have two or more screens
2. Open up the emote picker
3. Move the emote picker to a secondary screen
4. Close the emote picker
5. Open the emote picker again

It should open back up on the same spot you closed it, but it appears on the edge of the primary screen instead.

**Chatterino version**

**Operating system**
Windows 10 Pro (Version 2004, OS Build 19041.450)

**Additional information**
I've tried it with having the secondary screen on both the left and right sides of my primary screen and it does not seem to matter, neither does the fact that the primary and secondary screens have different resolutions (1440p and 1080p, tried having both on 1080p). The issue persists either way.

From a quick search around some the codebase and some issues and PRs, my initial guess would be that `desktop->availableGeometry(parent)` in this function (which ends up being called when the event picker tries to move to its previous position) does not consider my second monitor as available, somehow:
https://github.com/Chatterino/chatterino2/blob/07fc0c3ad7c2ea3495b3c3fafb97b16bff56fa36/src/widgets/BaseWindow.cpp#L587-L626

Nerixyz:
Pull request checklist:

# Description

[`QDesktopWidget` is marked as obsolete](https://doc.qt.io/qt-5/qdesktopwidget.html) in Qt 5 and removed in Qt 6. This PR migrates two (similar) usages.

Both used `QDesktopWidget` to check if something is inside the screen or not. [`QDesktopWidget::availableGeometry`](https://doc.qt.io/qt-5/qdesktopwidget.html#screenGeometry-1) used to return the available geometry of the screen the widget was in. Now, [`QGuiApplication::screenAt`](https://doc.qt.io/qt-5/qguiapplication.html#screenAt) is used.

Fixes #1926

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Dev: Removed unused include directives. (#4266, #4275)
 - Dev: Removed TooltipPreviewImage. (#4268)
 - Dev: Removed unused operators in `Image` (#4267)
+- Dev: Removed usage of deprecated `QDesktopWidget` (#4287)
 
 ## 2.4.0
 
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -26,7 +26,6 @@
 
 #include <boost/optional.hpp>
 #include <QDebug>
-#include <QDesktopWidget>
 #include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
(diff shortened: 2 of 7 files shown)


kornes:
this PR also fixes `moveIntoDesktopRect()` not working well with multiple monitors (where it would lock to primary screen only), both emote picker and tooltip widget appears on 2nd monitor when needed now, there are still improvements to be made but its lot better 👍

pajlada (approved):

PR #4300, opened 2023-01-11
Add CMake Install Support for Windows

Nerixyz:
Pull request checklist:

# Description

Chatterino had only [rudimentary support](https://github.com/Chatterino/chatterino2/blob/5ad8ca791c0498fb39e21d07b3c0e856b3e7677b/src/CMakeLists.txt#L640-L648) for `<build-tool> install` on Windows, because it only copied the `chatterino.exe`.

There were two main issues:

1. No shared libraries such as openssl were copied to `CMAKE_INSTALL_PREFIX`.
2. `windeployqt` never ran, because the command silently failed (cmake's [set](https://cmake.org/cmake/help/latest/command/set.html) joins [multiple arguments](https://github.com/Chatterino/chatterino2/blob/5ad8ca791c0498fb39e21d07b3c0e856b3e7677b/src/CMakeLists.txt#L642) as a semicolon-separated list).

This PR addresses these issues:

1. The [`RUNTIME_DEPENDENCIES`](https://cmake.org/cmake/help/latest/command/install.html#targets) are included. Windows DLLs are excluded, as done [here](https://discourse.cmake.org/t/runtime-dependencies-cannot-find-dll/3782).
2. `windeployqt` is run and aborts in case of any error ([`COMMAND_ERROR_IS_FATAL ANY`](https://cmake.org/cmake/help/latest/command/execute_process.html))

Furthermore, it reverts #4297 and adds awareness for `CMAKE_BUILD_TYPE`.

### How to build and install

**Using conan:**
```cmd
cmake .. -DUSE_CONAN=On -DCMAKE_PREFIX_PATH=<qt-path>\msvc2019_64\lib\cmake\Qt5 -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=<install-dir> -GNinja
ninja
ninja install
```

### Known issues

For `vcpkg`, `-DX_VCPKG_APPLOCAL_DEPS_INSTALL` is required.

### Documentation

I didn't add any documentation since this feature was never used. I want to rewrite the documentation for Windows (to include VS Code, Visual Studio, and vcpkg) so I'll add it there.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -649,17 +649,32 @@ if (BUILD_APP)
         )
 
     if (MSVC)
-        if (NOT VCPKG_INSTALLED_DIR)
+        if (NOT WINDEPLOYQT_PATH)
             get_target_property(Qt_Core_Location Qt${MAJOR_QT_VERSION}::Core LOCATION)
             get_filename_component(QT_BIN_DIR ${Qt_Core_Location} DIRECTORY)
-            set(WINDEPLOYQT_COMMAND "${QT_BIN_DIR}/windeployqt.exe" $<TARGET_FILE:${EXECUTABLE_PROJECT}> --release --no-compiler-runtime --no-translations --no-opengl-sw)
-
-            install(TARGETS ${EXECUTABLE_PROJECT}
-                    RUNTIME DESTINATION .
-                    )
-
-            install(CODE "execute_process(COMMAND ${WINDEPLOYQT_COMMAND} --dir \${CMAKE_INSTALL_PREFIX})")
+            string(APPEND WINDEPLOYQT_PATH ${QT_BIN_DIR} /windeployqt.exe)
+        else()
+            file(TO_CMAKE_PATH "${WINDEPLOYQT_PATH}" WINDEPLOYQT_PATH)
         endif()
+
+        if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+            set(WINDEPLOYQT_MODE --debug)
+        else()
+            set(WINDEPLOYQT_MODE --release)
+        endif()
+
+        set(WINDEPLOYQT_COMMAND_ARGV "${WINDEPLOYQT_PATH}" "$<TARGET_FILE:${EXECUTABLE_PROJECT}>" ${WINDEPLOYQT_MODE} --no-compiler-runtime --no-translations --no-opengl-sw)
+        string(REPLACE ";" " " WINDEPLOYQT_COMMAND "${WINDEPLOYQT_COMMAND_ARGV}")
+
+        install(TARGETS ${EXECUTABLE_PROJECT} 
+            RUNTIME_DEPENDENCIES 
+                PRE_EXCLUDE_REGEXES "api-ms-" "ext-ms-"
+                POST_EXCLUDE_REGEXES ".*system32/.*\\.dll" 
+                DIRECTORIES ${QT_BIN_DIR}
+            RUNTIME DESTINATION .)
+
+        install(CODE "message(\"Running: ${WINDEPLOYQT_COMMAND} --dir \\\"\${CMAKE_INSTALL_PREFIX}\\\"\")")
+        install(CODE "execute_process(COMMAND ${WINDEPLOYQT_COMMAND} --dir \"\${CMAKE_INSTALL_PREFIX}\" COMMAND_ERROR_IS_FATAL ANY)")
     elseif (APPLE)
         install(TARGETS ${EXECUTABLE_PROJECT}
                 RUNTIME DESTINATION bin
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -668,10 +668,6 @@ if (BUILD_APP)
 
         if (X_VCPKG_APPLOCAL_DEPS_INSTALL)
             install(TARGETS ${EXECUTABLE_PROJECT} RUNTIME DESTINATION .)
-            install(CODE "message(\"-- Flattening plugins/\")")
-            # vcpkg will add the plugin folders (like "iconengines" or "mediaformats") inside a `plugins` folder.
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -637,17 +637,32 @@ if (BUILD_APP)
         )
 
     if (MSVC)
-        if (NOT VCPKG_INSTALLED_DIR)
```
kornes, line 640: none of that should run when vcpkg toolchain is used, so `VCPKG_INSTALLED_DIR` check should be kept.

Nerixyz: Why? For me, it just leaves the folder empty.

kornes: take a look on this
https://github.com/microsoft/vcpkg/blob/036769f5d6da667391c9240bc7ca56e84cacd5b1/scripts/buildsystems/vcpkg.cmake#L48
and these 2 scripts, called in post-build stage
https://github.com/microsoft/vcpkg/blob/master/scripts/buildsystems/msbuild/applocal.ps1
https://github.com/microsoft/vcpkg/blob/master/ports/qt5-base/qtdeploy.ps1

maybe some problem with running powershell on your side causes it to fail

Nerixyz: > called in post-build stage

This PR is not about building chatterino, it's about installing chatterino. When building, the deployment script is run, but it needs to run at installation as well.

kornes: oh, i blame my lack of cmake knowledge then

it seems there is experimental switch for cmake install support, maybe it will help with your PR: https://github.com/microsoft/vcpkg/blob/036769f5d6da667391c9240bc7ca56e84cacd5b1/scripts/buildsystems/vcpkg.cmake#L52

anyways, my issue is resolved if it wont affect building

pajlada: Not sure how this works on Windows, but on Linux plugins are stored inside a `plugins` directory
e.g. if I install `extra/qt5-imageformats` it installs a directory under `/usr/lib/qt/plugins/` like this:
```
/usr/lib/qt/plugins
├── imageformats
│   ├── libqgif.so
│   ├── libqicns.so
│   ├── libqico.so
│   ├── libqjp2.so
│   ├── libqjpeg.so
│   ├── libqmng.so
│   ├── libqpdf.so
│   ├── libqsvg.so
│   ├── libqtga.so
│   ├── libqtiff.so
│   ├── libqwbmp.so
│   └── libqwebp.so
```

kornes: when building with vcpkg it also places all plugins inside /plugins. When Qt search for plugins it checks root executable folder and also Plugins path from qt.conf file, which defaults to /[plugins](https://doc.qt.io/qt-6/qt-conf.html#overriding-paths) when not set (like in chatterino case).
Imo to keep it consistent with vcpkg build, dont move plugins in this PR

Nerixyz: This seems like the problem described here https://github.com/microsoft/vcpkg/issues/7461#issuecomment-659612311. The linked fixes mostly manipulate [`qt.conf`](https://doc.qt.io/qt-5/qt-conf.html), which didn't really work for me.

pajlada:
Is this in a state both @Nerixyz and @kornes are happy with?

kornes:
i would revert moving plugins part from bfdbf1c2a6488a7bae72355923d292a1ad69b429, and if there is need for it (personally i dont mind it) address it in another PR for both building and install
other than that lgtm 👍

Nerixyz:
> and if there is need for it (personally i dont mind /plugins path) address it in another PR for both building and install

The reason I added this was because chatterino wouldn't launch without it. Does it launch for you after installing in a fresh directory?

pajlada:
I would expect an install step to make the binary run without further modification as long as the install step isn't destructive (i.e. it may override system files or w/e).

Normal CMake install:
1. Install executable file to `C:/forsen/chatterino.exe`
2. Install plugins to `C:/forsen/plugins/...`

Your extra step:
3. Move plugins from `C:/forsen/plugins/...` to `C:/forsen/...` right?

Nerixyz:
> I would expect an install step to make the binary run without further modification as long as the extra step isn't destructive (i.e. it may override system files or w/e).
> 
> Normal CMake install:
> 1. Install executable file to `C:/forsen/chatterino.exe`
> 2. Install plugins to `C:/forsen/plugins/...`
> 
> Your extra step:
> 3. Move plugins from `C:/forsen/plugins/...` to `C:/forsen/...` right?

Yes that's the planned procedure.

kornes:
> Does it launch for you after installing in a fresh directory?

chatterino ability to find qt plugins on this PR and master (win + vcpkg):
| plugins path | install (release) | install (debug) | build (release) | build (debug) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| `./`  | ✅ | ✅ | ✅ | ✅ | 
| `./plugins`  | ✅ | ⛔️ | ✅ | ⛔️ | 

so it's only debug vcpkg build acting up and not finding plugins in ./plugins folder (default hardcoded destination).
Suggested workaround is to specify path in `resources/qt.conf`, by adding:
```
[Paths]
Plugins = ./plugins
```
table then looks like this:
| plugins path | install (release) | install (debug) | build (release) | build (debug) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| `./`  | ✅  | ✅  | ✅  | ✅  | 
| `./plugins`  | ✅ | ✅ | ✅ | ✅ | 

so in theory this change shouldn't break other build systems (since `./` is checked first) and fix both install and build on vcpkg

pajlada (changes requested):
This breaks non-install builds on Linux

Nerixyz:
> This breaks non-install builds on Linux

Specifically, the changes in `qt.conf` break builds (or rather runs). My approach would be to revert the change to `qt.conf`, since it only affects vcpkg debug builds (that you wouldn't want to install anyway).

pajlada (approved):

Issue #1456, opened 2019-12-23
Add a filter for channels you want to save logs for


zneix:
**Is your feature request related to a problem? Please describe.**

I have a lot of channels joined in my Chatterino. I mod some of those, logging chat history is often really helpful, but I often end up having too many useless logs for big channels like forsen and I obviously don't need those, so that spam only takes more of my disk space.

---

**Describe the solution you'd like**

I thought of few things:
 - A table, may be a similar thing like the one with Highlights
 - Input field(s) that will let you type channel names separated by commas (e.g. "pajlada, zneix, normiestreamer123,")
 - In case you have **a lot** of those channels, then you can maybe include an option to select .txt file with channel names separated by newlines or commas

Selected channels would be logged, others not.
There could be also a tick box for enabling/disabling channel filtering in this feature.

---

**Describe alternatives you've considered**

There might be also an option to only "blacklist" certain channels but I think it is not *that* necessary to have.

---

**Additional context**
Not really, Have a good Christmas :)

leon-richardt:
If this is implemented, it might make sense to provide an interface for generic channel-specific settings. Options like "Notify when live" or "Mute highlight sound" could then be consolidated into a common place.
Designing such an interface might be quite a bit of work though.

apa420:
I don't think it would be that hard, you could get all open channels, create a table and have tabs with checkboxes for each of the settings

leon-richardt:
> I don't think it would be that hard, you could get all open channels, create a table and have tabs with checkboxes for each of the settings

The "quite a bit of work" was referring to generic channel-specific settings. This feature in isolation shouldn't be too hard, I agree.

apa420:
So you mean reworking the way the current implementations are done in a new common place for it? Just using the ones we have now and connecting the checkboxes would be easy.

leon-richardt:
This issue might not be the right place to have this discussion, but essentially yes.

For example, the channel-specific settings "Notify when live" or "Mute highlight sound" are currently implemented in the `NotificationController` and `PingController`, respectively.
I'm wondering whether it would make sense to introduce a new interface for managing channel-specific settings in a common place rather than spreading functionality over different classes. In that case, this feature request could be implemented there.

I'm open to discussion though.

apa420:
I do think implementations of issues is the right place to have discussions over it, where else would it be?

So yea that's what I was trying to get at.

It does make sense to introduce such an interface since it would also make it easier to implement some new channel-spefic feature.

[...]

PR #4302, opened 2023-01-14
Add feature to select channels to log

askepticaldreamer:
Pull request checklist:

# Description

Implemented feature requested in https://github.com/Chatterino/chatterino2/issues/1456 to allow custom selection of channels to log. Added a new setting on Moderation(Logs) page to enable this feature as well as a table to add channels and a checkbox for whether messages for each channel should be logged or not

### DIFF
--- a/src/controllers/logging/ChannelLog.hpp
+++ b/src/controllers/logging/ChannelLog.hpp
@@ -45,28 +45,28 @@ struct Deserialize<chatterino::ChannelLog> {
     static chatterino::ChannelLog get(const rapidjson::Value &value,
                                       bool *error = nullptr)
     {
-        chatterino::ChannelLog ChannelLog =
+        chatterino::ChannelLog channelLog =
             chatterino::ChannelLog::createEmpty();
 
         if (!value.IsObject())
         {
             PAJLADA_REPORT_ERROR(error);
-            return ChannelLog;
+            return channelLog;
         }
 
-        if (!chatterino::rj::getSafe(value, "channel", ChannelLog.channel))
+        if (!chatterino::rj::getSafe(value, "channel", channelLog.channel))
         {
             PAJLADA_REPORT_ERROR(error);
-            return ChannelLog;
+            return channelLog;
         }
         if (!chatterino::rj::getSafe(value, "loggingEnabled",
-                                     ChannelLog.loggingEnabled))
+                                     channelLog.loggingEnabled))
         {
             PAJLADA_REPORT_ERROR(error);
-            return ChannelLog;
+            return channelLog;
         }
 
-        return ChannelLog;
+        return channelLog;
     }
 };
 
--- a/src/controllers/logging/ChannelLog.hpp
+++ b/src/controllers/logging/ChannelLog.hpp
@@ -16,9 +16,9 @@ public:
 
     bool operator==(const ChannelLog &other) const;
 
-    QString toString() const;
+    [[nodiscard]] QString toString() const;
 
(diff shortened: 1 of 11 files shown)


src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -69,8 +70,23 @@ ModerationPage::ModerationPage()
 
     auto logs = tabs.appendTab(new QVBoxLayout, "Logs");
     {
-        logs.append(this->createCheckBox("Enable logging",
-                                         getSettings()->enableLogging));
+        QCheckBox *enableLogging = this->createCheckBox(
+            "Enable logging", getSettings()->enableLogging);
+        logs.append(enableLogging);
+        QCheckBox *customLogging =
+            this->createCheckBox("Enable logging for specific channels",
```
pajlada, line 77: Phrasing of the setting should be more along the lines of `Only log channels listed below`, and it should be moved to be just above the list of channels

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -140,6 +156,22 @@ ModerationPage::ModerationPage()
                              }));
                          });
 
+        EditableModelView *view =
+            logs.emplace<EditableModelView>(
+                    (new ChannelLoggingModel(nullptr))
+                        ->initialized(&getSettings()->loggedChannels))
+                .getElement();
+
+        view->setTitles({"Twitch channels", "Enable Logs"});
```
pajlada, line 165: Do we really need a checkbox for "Enable logs" here? I'd prefer if this was just a list, and if a channel is in the list, that means it's being logged.
I can't think of any use-cases where you'd need to disable logging for many channels, then enable it again briefly afterwards. If you just want to stop logging a channel for a while, it would make more sense in my opinion to just remove it.

src/singletons/Settings.hpp
```diff
@@ -364,6 +366,7 @@ class Settings : public ABSettings, public ConcurrentSettings
 
     /// Logging
     BoolSetting enableLogging = {"/logging/enabled", false};
+    BoolSetting enableCustomLogging = {"/logging/customLogging", false};
```
pajlada, line 369: Setting name & path should be adapted to something more along the lines of "onlyLogListedChannels" or w/e phrasing makes most sense

src/controllers/logging/ChannelLog.hpp
```diff
@@ -0,0 +1,73 @@
...
+        return ret;
+    }
+};
+
+template <>
+struct Deserialize<chatterino::ChannelLog> {
+    static chatterino::ChannelLog get(const rapidjson::Value &value,
+                                      bool *error = nullptr)
+    {
+        chatterino::ChannelLog ChannelLog =
```
pajlada, line 48:
```suggestion
        chatterino::ChannelLog channelLog =
```

pajlada (changes requested):
Thank you for looking at this feature!
Feature-wise this looks fine, just some nitpicks & UX things to consider.

src/widgets/settingspages/ModerationPage.cpp
askepticaldreamer: The "Enable logs" toggle was inspired by the discussion in the issue on having adding other toggleable channel specific settings in a table as well but I guess that wouldn't be on this page anyway. Removed the toggle.

pajlada (approved):
Works as expected

I've changed the JSON format a bit so it's `channelName` instead of just `channel` in case we move to IDs at some point.

PR #4303, opened 2023-01-14
Change the highlight order to prioritize Message highlights over User highlights

### LINKED ISSUE #4259: Message highlights taking priority over user highlights
### Discussed in https://github.com/Chatterino/chatterino2/discussions/3865

<sup>Originally posted by **LosFarmosCTL** July 17, 2022</sup>
Was there a specific reason to make user highlights take priority over message/phrase highlights?

IMO it would make more sense the other way around, highlight all messages from one user, but i.e. if that one user pings me I'd still want that to take priority.</div>

`felanbird:`
I'm moving this to an issue, in hopes it gets more traction. Since the release of 2.4.0, the most common deterrent to updating in certain areas of Twitch, is this functionality change. 
Due to the issues with 7.4.0's release, the discussion of updating happens way more often, and due to this, other chatters are able to give their opinion, which is to not update.
I'd say out of every 10 users, 4 of them update, 2 are too lazy, and the other 4 either experienced the change, or were deterred by another who did.

pajlada's comment:
image

pajlada:
Pull request checklist:

# Description

- Change the highlight order to prioritize Message highlights over User highlights
- Add changelog entry

Fixes #4259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Change the highlight order to prioritize Message highlights over User highlights. (#4303)
 - Minor: Added ability to negate search options by prefixing it with an exclamation mark (e.g. `!badge:mod` to search for messages where the author does not have the moderator badge). (#4207)
 - Minor: Search window input will automatically use currently selected text if present. (#4178)
 - Minor: Cleared up highlight sound settings (#4194)
--- a/src/controllers/highlights/HighlightController.cpp
+++ b/src/controllers/highlights/HighlightController.cpp
@@ -445,15 +445,15 @@ void HighlightController::rebuildChecks(Settings &settings)
     checks->clear();
 
     // CURRENT ORDER:
-    // Subscription -> Whisper -> User -> Message -> Reply Threads -> Badge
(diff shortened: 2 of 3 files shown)


Felanbird (approved):
See my section of #4259 for my opinion on why I'd like this.

Summary being `If a user you have highlighted, pings you, said message is not highlighted in the ping color, which causes confusion of who pinged you`

pajlada:
:sunglasses: :mega: I'm going to merge this in later today, leave your feedback if you have any

8thony:
I'm okay with message > user, but we should consider implementing a system so every user can set up the order of the highlights based on their personal preferences. 
for example my case (#4191) is impossible since v2.4.0.
Other users have ideas to make it more customizable like #3472. 
Let me know what you think. 
based on https://github.com/Chatterino/chatterino2/discussions/3872#discussioncomment-4684561

Felanbird:
Obviously it would be preferred to have the custom system, but as far as I'm aware nobody has it in the pipeline, and I would like as little users as possible to feel restricted on update, due to the loss of so much `/command` functionality if they do not update by Feb 18th.
So this approach is best needed as of now.

PR #4304, opened 2023-01-15
Format YAML and JSON Files With Prettier

Nerixyz:
Pull request checklist:

# Description

These files were previously ignored by prettier, but to keep a consistent style these should be formatted as well.

I moved the config from TOML to JSON, because the JSON file is easier to read (mainly because of the `overrides`).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,4 +22,4 @@ jobs:
         uses: actionsx/prettier@v2
         with:
           # prettier CLI arguments.
-          args: --check '**/*.{md,yml,yaml}' '.prettierrc' 'vcpkg.json'
+          args: --check .
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,7 +7,7 @@ on:
       - master
   pull_request:
 
-concurrency: 
+concurrency:
   group: lint-${{ github.ref }}
   cancel-in-progress: true
 
@@ -22,4 +22,4 @@ jobs:
         uses: actionsx/prettier@v2
         with:
           # prettier CLI arguments.
-          args: --check '**/*.md'
+          args: --check '**/*.{md,yml,yaml}' '.prettierrc' 'vcpkg.json'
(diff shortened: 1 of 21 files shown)


Felanbird (approved):

.prettierrc
```diff
@@ -0,0 +1,18 @@
...
+      "options": {
+        "proseWrap": "preserve"
+      }
+    },
+    {
+      "files": ["*.{yml,yaml}"],
+      "options": {
+        "singleQuote": true
+      }
+    }
```
pajlada, line 16: Setting `singleQuote` to true makes more changes than leaving it at its default value, so I'd argue this override should go away

.prettierignore
```diff
@@ -3,4 +3,4 @@ lib/*/
 conan-pkgs/*/
 cmake/sanitizers-cmake/
 
-.github/
+.clang-tidy
```
pajlada, line 6: If we ignore `.clang-tidy` (which I don't think we should do long-term, but I'm ok with that being made in a followup PR), we should also ignore `.clang-format`

.github/workflows/lint.yml
```diff
@@ -22,4 +22,4 @@ jobs:
         uses: actionsx/prettier@v2
         with:
           # prettier CLI arguments.
-          args: --check '**/*.md'
+          args: --check '**/*.{md,yml,yaml}' '.prettierrc' 'vcpkg.json'
```
pajlada, line 25: We can change this to just `.` if we're expanding it to check the entire project, it will go through all directories where it makes sense (skipping things in .prettierignore)

```suggestion
          args: --check .
```

.prettierrc
```diff
@@ -0,0 +1,18 @@
+{
```
pajlada, line 1: If we're changing the format of this file, I'd rather it be changed to YAML. the JSON format for overrides is imo not much prettier than TOML

pajlada (changes requested):
Small nitpicks

.prettierignore
Nerixyz: I included the `.clang-format` files in this PR.

pajlada (approved):

PR #4305, opened 2023-01-15
Fix the split "Search" menu action not opening the correct search window

pajlada:
Pull request checklist:

# Description


Felanbird (approved):
works 👍

PR #4159, opened 2022-11-17
:hammer: Automatically generate resources files with cmake

AnotherFoxGuy:
This PR replaces `resources/generate_resources.py` with a cmake script that will automatically generate the required resource files

Todo:
- [x] Test on Windows with QtCreator
- [x] Test on Windows with proper building
- [x] Test on Linux
- [x] Test on Mac
- [x] Maybe move the generated files to the build folder? (Needs to be discussed)
- [x] Update  `CHANGELOG.md` 

Fixes #3949

PR #4312, opened 2023-01-16
Add Wissididom to contributors list

Wissididom:
Pull request checklist:

# Description

Used pngcrush
Compiled and run
Checked About section.

One thing I noticed is that in the explanation inside `contributors.txt` there is the info to run `./generate_resources.py` which I didn't find, so I didn't run it.


Felanbird:
Script was removed in #4159 - I'll clean that comment up later, I'll look into when the resource should be generated

Felanbird (approved):
It just works due to the above mentioned PR

Issue #2497, opened 2021-03-05
It is possible to have two commands named the same


Mm2PL:
**Describe the bug**

You can have two commands named the same string of characters, which is confusing.

**To reproduce**

  1. Add a command (`/test123`)
  2. Add another one with the same name but different output
  3. Observe that the first command in the list will get executed and the second one won't.

**Screenshots**

![2021-03-05_14-31](https://user-images.githubusercontent.com/25011746/110122135-20efb680-7db7-11eb-8078-f745cfce075e.png)
![2021-03-05_14-31_1](https://user-images.githubusercontent.com/25011746/110122132-20572000-7db7-11eb-822a-90dba0f37ebf.png)

**Chatterino version**

Chatterino Nightly + Dankerino patches 2.2.3-beta2 (commit 3a589589) based on upstream b2aa3700

**Operating system**

Ubuntu 20.04.01

jammehcow:
Can reproduce, Chatterino 2.3.4 (commit [baeec59a](https://github.com/Chatterino/chatterino2/commit/baeec59a)). Will look into adding validation before a command is saved/cell is unfocused

PR #4322, opened 2023-01-21
feat: Warn for commands with duplicate triggers

askepticaldreamer:
Pull request checklist:

# Description

Fixes https://github.com/Chatterino/chatterino2/issues/2497 
Added warning tooltip for when user selected another cell and previous command name was a duplicate. Tried to force selection to the duplicate cell with the following

      int duplicateIndex = (i > previous.row()) ? i : previous.row();
      view->getTableView()->selectRow(duplicateIndex);
      view->getTableView()->selectColumn(0);
                    
  But didn't seem to work. Could leave as is with just tool tip warning or open to suggestions for methods to force user to change duplicate command name


pajlada:
I think the friendliest way to handle this atm will be to not stop the user from renaming or creating commands with specific names, but to mark all commands with duplicate entries. This can be done by making them slightly yellow, or adding a :warning: icon to the start of the row.
In addition, a warning text should appear above or below the list if any of the commands have duplicate names

pajlada:
Marked it as a draft until it solves it (or you think it solves the issue), feel free to poke me at any time before that if you want a review of the code or way you're trying to handle it

askepticaldreamer:
@pajlada what do you think of this approach? Am I overcomplicating things? The code is able to set text colors and add/remove the warning label as duplicates are added/removed but currently runs into an issue when it tries to set the color after removing an item or moving it up/down so not sure if I need a mutex somewhere or something.

src/widgets/settingspages/CommandPage.hpp
```diff
@@ -11,9 +12,11 @@ class CommandPage : public SettingsPage
 {
 public:
     CommandPage();
+    bool handleCommandDuplicates(EditableModelView *view);
```
pajlada, line 15: If we need to act on the model view, store it as a private variable instead so you can access it with `this->view`.
Then instead of a public `handle..` function, make it a private `checkCommandDuplicates` function that makes all changes necessary

src/widgets/settingspages/CommandPage.cpp
```diff
@@ -81,8 +99,61 @@ CommandPage::CommandPage()
...
+    }
+
     // ---- end of layout
     this->commandsEditTimer_.setSingleShot(true);
 }
 
+bool CommandPage::handleCommandDuplicates(EditableModelView *view)
+{
+    bool retval = false;
+    QMap<QString, QList<int>> map = *new QMap<QString, QList<int>>();
```
pajlada, line 122: a little memory leaking :tf: - don't use `new` here

pajlada (changes requested):
Looks alright, let's review in parts so you can work on these things first

src/widgets/settingspages/CommandPage.cpp
```diff
@@ -116,13 +116,13 @@ CommandPage::CommandPage()
     this->commandsEditTimer_.setSingleShot(true);
 }
 
-bool CommandPage::handleCommandDuplicates(EditableModelView *view)
+bool CommandPage::checkCommandDuplicates()
 {
     bool retval = false;
-    QMap<QString, QList<int>> map = *new QMap<QString, QList<int>>();
-    for (int i = 0; i < view->getModel()->rowCount(); i++)
+    QMap<QString, QList<int>> map = QMap<QString, QList<int>>();
```
pajlada, line 122:
```suggestion
    QMap<QString, QList<int>> map;
```

src/widgets/settingspages/CommandPage.cpp
```diff
@@ -38,18 +39,35 @@ CommandPage::CommandPage()
...
+        selectionModel, &QItemSelectionModel::currentChanged, this,
+        [this](const QModelIndex &current, const QModelIndex &previous) {
+            if (this->checkCommandDuplicates())
+            {
+                this->duplicateCommandWarning->show();
+            }
+            else
+            {
+                this->duplicateCommandWarning->hide();
+            }
```
pajlada, line 66: Instead of checking the return value, set the visibility of `duplicateCommandWarning` at the bottom of the `checkCommandDuplicates` command

pajlada (changes requested):

pajlada (changes requested):
With this PR I'm able to crash Chatterino by:
1. Removing the last row (maybe removing any row?)
2. Dragging one row onto another

pajlada:
I don't think a button for this is the right approach - if it cannot be figured out automatically then I don't think it provides much value for users

pajlada:
If it's just there for you while testing then my bad, bit fast with the comments :sunglasses:

pajlada:
Doesn't seem to be checking duplicates on row removal, and doesn't handle renaming from one duplicate to another duplicate well
https://user-images.githubusercontent.com/962989/218690818-94a129cd-91df-45b9-87c5-09937e4e6dfe.mp4

askepticaldreamer:
@pajlada that's interesting, both those cases work on my machine. I'll record a video when I get a chance. Should be fun to debug lol.

askepticaldreamer:
Was able to repro. Wasn't happening before since I was testing duplicate commands that weren't the one just added. Changed the signals and seems to work now

Mm2PL (approved):
lgtm

pajlada (approved):

PR #4350, opened 2023-01-31
Display message being replied to above input box

### LINKED ISSUE #3897: Show message content in replying to ... textbox
### Discussed in https://github.com/Chatterino/chatterino2/discussions/3896

<sup>Originally posted by **MrAuro** July 31, 2022</sup>
In webchat whenever you are replying to someone's message it shows the message content while in Chatterino ([ebc7852f](https://github.com/Chatterino/chatterino2/commit/ebc7852f)) it does not. This can cause confusion on what message from the author you are replying to:
Chatterino and webchat comparison
</div>

dnsge:
Pull request checklist:

# Description

When you begin replying to a message (by right-clicking a message and selecting "reply" or clicking the reply button), the message being replied to will now display above the input box. 

Example:
image

In the case of very long messages, the message will be truncated:
<details>
<summary>Expand for image</summary>
image
</details>

**The message displayed is uninteractable**: you can't click on usernames, highlight text, or see emote tooltips for example. The reply button is hidden and you won't see inline replies if the message you're replying to is also a reply.

The very long message truncation is done using a fixed height limit. This limit was chosen to be big enough for most messages, but we could make it larger/smaller.

Fixes #3897


src/widgets/splits/SplitInput.hpp
```diff
@@ -100,7 +101,7 @@ class SplitInput : public BaseWidget
     /// Clears the input box, clears reply thread if inline replies are enabled
     void clearInput();
 
-protected:
+private:
```
dnsge, line 104: This was changed at one point during my work on #3722 but was never changed back. Made it's way through code review. I noticed it while working on this PR and figured I'd change it back.

sheybey:
Looks fine on Windows for the most part, but the emote picker and close icons are broken. They still work as expected when I click on them. I'll try to debug that this afternoon.

Screenshot 2023-02-11 142614

dnsge:
I've updated this PR with the latest changes from master. I'm happy with the current state, so I'm marking it as ready for review. 

Something to decide: should showing the message being replied to be a setting? We could let the user toggle between what happens now (just the "replying to username" text) and this PR.

pajlada:
Looks like this now:
image

more like the "single line text" that's used in replies already

pajlada (approved):
Looks good to me, I'm happy with how this looks.
Do you want to have a look before I merge this in @dnsge?

dnsge:
@pajlada looks good! Less noisy when we have highlights. Thanks for updating it.

Felanbird (approved):

PR #4358, opened 2023-02-05
Add setting to only show tabs with live channels

### LINKED ISSUE #2334: Only open the Chats in Chatterino that are live
a

dnsge:
Pull request checklist:

# Description

Adds a setting to only show tabs that contain a channel that is live, i.e. has the "live" indicator. This "filtering" of tabs is implemented through a generic tab predicate which could be used for other kinds of conditional displaying of tabs in the future.

Adds the default key bind `Ctrl+Shift+L` to toggle live tabs only.

This PR is a replacement for #3658, which is apparently no longer being worked on. (felanbird): Closes #3658

Addresses #3656, #4353, and (possibly?) fixes #2334.


dnsge:
I've update this branch to `master` and have fixed the last few problems that I found with this feature.

I've also addressed the problems brought up in this review in #3658 here: https://github.com/Chatterino/chatterino2/pull/3658#pullrequestreview-937212400. 

1. When opening/viewing a tab that doesn't match the filter criteria (i.e. a not-live tab), the tab will be visible for as long as the user focuses it. The unfocusing is a bit jarring, but it seems to be the best solution.
2. When swapping to a hidden tab via the quick switcher, the tab will become visible
3. Dragging around tabs is supported when filtering tabs. Relative ordering is maintained.

**An idea open for discussion:**
Perhaps the "Live only tabs" filter should also include tabs where you have unread mentions? It would probably be weird to get notification sounds for mentions without being able to see the messages. This could also be its own visibility option in settings.

src/widgets/Notebook.cpp
```diff
@@ -262,45 +261,121 @@ void Notebook::selectIndex(int index, bool focusPage)
...
+            return;
+        }
+
+        this->select(this->items_[size - 1].page, focusPage);
         return;
     }
 
-    this->select(this->items_[size - 1].page, focusPage);
+    // find first tab permitted by filter starting from the end
+    for (auto it = this->items_.rbegin(); it != this->items_.rend(); ++it)
```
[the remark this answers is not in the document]
dnsge: surely c++20 ranges are supported across major compilers already? After all, it was completed in 2020

image

src/widgets/Notebook.hpp
```diff
@@ -35,6 +38,7 @@ class Notebook : public BaseWidget
     int indexOf(QWidget *page) const;
     virtual void select(QWidget *page, bool focusPage = true);
     void selectIndex(int index, bool focusPage = true);
```
pajlada, line 48: It would be nice if `selectIndex` was made private, but we use it in the SelectChannelPopup

src/widgets/Notebook.hpp
```diff
@@ -113,6 +125,8 @@ class Notebook : public BaseWidget
     NotebookTabLocation tabLocation_ = NotebookTabLocation::Top;
     QAction *lockNotebookLayoutAction_;
     QAction *showTabsAction_;
+
+    std::function<bool(const NotebookTab *)> tabFilter_ = nullptr;
```
pajlada: Maybe rename to `tabVisibilityFilter_`?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Minor: Add an icon showing when streamer mode is enabled (#4410)
 - Minor: Added `/shoutout <username>` commands to shoutout specified user. (#4638)
 - Minor: Improved editing hotkeys. (#4628)
+- Minor: Added setting to only show tabs with live channels. (#4358)
```
pajlada, line 9: Add information about its default hotkey

src/widgets/Notebook.hpp
```diff
@@ -19,6 +21,15 @@ class SplitContainer;
 
 enum NotebookTabLocation { Top = 0, Left = 1, Right = 2, Bottom = 3 };
 
+// Controls the visibility of tabs in this notebook
+enum NotebookTabVisibility : int {
+    // Show all tabs
+    Default = 0,
```
pajlada, line 27: Should probably be renamed to `AllTabs` - it just happens to be the default setting

pajlada (changes requested):
Only open question I have really is:
Can this be confusing if a user is not logged in, hence not getting live status for streams, causing all tabs to be shown as offline?

Functionally it works as expected :+1: Thank you!

pajlada:
As for your open discussion question, I feel like that should be a separate discussion/PR. I think this code is just abstract enough that it should fit in without any big issues.
Glad you made it an enum.

Felanbird:
I do agree that it is a problem, but it should be one that isn't interacted with _too_ often, since you have to be both logged out and having toggled the setting.

_If_ in the future we made it so when your token expires you fall-back to anon this could be an issue, but I'm doubtful we'll make that change.

pajlada (approved):
Thank you!

mjeveritt:
Thanks for this - makes a huge list to my ever-expanding tabs list.. !

One minor enhancement request .. would it be possible to 'exclude' (ie. mark always-live) channels such as Whispers, Live, Mentions (any channels prefixed by / (slash) possibly?) so that they are effectively 'tagged' when you select Live-Only?
I could have a look-see at the code myself, but I suspect it might be easier/quicker for you guys to patch it in, rather than me go hacking away... if that's not too much trouble?!

mjeveritt:
> Thanks for this - makes a huge list to my ever-expanding tabs list.. !
> 
> One minor enhancement request .. would it be possible to 'exclude' (ie. mark always-live) channels such as Whispers, Live, Mentions (any channels prefixed by / (slash) possibly?) so that they are effectively 'tagged' when you select Live-Only? I could have a look-see at the code myself, but I suspect it might be easier/quicker for you guys to patch it in, rather than me go hacking away... if that's not too much trouble?!

Conceivably, this could also be a toggle setting, in case there is a good use case for keeping the existing behaviour?! :smile:

pajlada:
@mjeveritt not sure how this would fit in easily with the current solution, what you can do as a temporary fix is add some always-online channel like `streamerhouse` and make it really small in that tab

mjeveritt:
> @mjeveritt not sure how this would fit in easily with the current solution, what you can do as a temporary fix is add some always-online channel like `streamerhouse` and make it really small in that tab

Thanks for your reply! I can see how to add splits, but it's not apparent how to resize those splits once created - they appear to be equal proportions to whatever horizontal/vertical 'edge' you bond them to .. is this as intended, or am I missing something!? There also does seem to be an absence of documentation (notably in the wiki) on how to create/adjust them, but I note there's a small animation in the submenu!

mjeveritt:
I was hoping there might be a simple/clever way to leverage the window/channel type vis https://github.com/Chatterino/chatterino2/blob/a045d3ee81338ff81abc5c11936e46703e3006bf/src/common/Channel.hpp#L33 but I can't quite see how non-channel windows/tabs are created in the code (ie. set those non-channel tabs to Live on creation/permanently) .. tips?!

pajlada:
The filter doesn't work on splits or channels, it uses the tab "isLive". You'd need to where the `setLive` call is made, and make it call `setLive(true)` if the notebook contains a split whose that is one of those tabs

mjeveritt:
> The filter doesn't work on splits or channels, it uses the tab "isLive". You'd need to where the `setLive` call is made, and make it call `setLive(true)` if the notebook contains a split whose that is one of those tabs

right .. I was trying to work out whether that could be done at 'tab creation'/initialisation, or needs to be done in a loop parsing notebook tabs/splits? perhaps I need to look deeper into the 'splits' aspect of Chatterino .. :)

mjeveritt:
ok, parsing and re-parsing what you wrote, I can see I've confused the 'setLive' functions, and a grep of the source code does confirm this.
What I need to do, is parse the channel type in https://github.com/Chatterino/chatterino2/blob/4b40b9a310ec44e46edff05834e5c4b8d387f2f4/src/widgets/splits/SplitContainer.cpp#L920 and, as you describe, call `setLive` there, if it's a Live/Whispers/Mentions/Watching channel-type.
Think I've got it now .. L931 logic needs changing a bit to contain something involving some
`if (this->getChannel()->getType() == Channel::Type::TwitchLive`
-type statements ..

Leave it with me .. :smiley:

mjeveritt:
> > @mjeveritt not sure how this would fit in easily with the current solution, what you can do as a temporary fix is add some always-online channel like `streamerhouse` and make it really small in that tab
> 
> Thanks for your reply! I can see how to add splits, but it's not apparent how to resize those splits once created - they appear to be equal proportions to whatever horizontal/vertical 'edge' you bond them to .. is this as intended, or am I missing something!? There also does seem to be an absence of documentation (notably in the wiki) on how to create/adjust them, but I note there's a small animation in the submenu!

Aha, for others curious, holding down Ctrl key allows you to change the divider position ... Split newbie!! :laughing:

Issue #3942, opened 2022-09-01
Channel Point crash


pajlada:
I had a crash happen earlier when a channel point reward was added through pubsub.

https://github.com/Chatterino/chatterino2/blob/46f43f3ce82fbf6be1f4035d75d5ee8d6685f08d/src/providers/twitch/TwitchChannel.cpp#L254 fires and is consumed here https://github.com/Chatterino/chatterino2/blob/46f43f3ce82fbf6be1f4035d75d5ee8d6685f08d/src/providers/twitch/IrcMessageHandler.cpp#L383 but the connection is never disconnected, nor does it contain any sort of shared pointer logic, so the function can go out of scope causing a segmentation fault

backtrace:
```
(gdb) bt
#0  0x00007f827d8a14dc in  () at /usr/lib/libc.so.6
#1  0x00007f827d851998 in raise () at /usr/lib/libc.so.6
#2  0x00007f827d83b53d in abort () at /usr/lib/libc.so.6
#3  0x00007f827da99833 in __gnu_cxx::__verbose_terminate_handler() () at /usr/src/debug/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
#4  0x00007f827daa5cfc in __cxxabiv1::__terminate(void (*)()) (handler=<optimized out>) at /usr/src/debug/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:48
#5  0x00007f827daa4cba in __cxa_call_terminate(_Unwind_Exception*) (ue_header=0x55aa14094c10) at /usr/src/debug/gcc/libstdc++-v3/libsupc++/eh_call.cc:54
#6  0x00007f827daa543a in __cxxabiv1::__gxx_personality_v0(int, _Unwind_Action, _Unwind_Exception_Class, _Unwind_Exception*, _Unwind_Context*)
    (version=<optimized out>, actions=6, exception_class=5138137972254386944, ue_header=<optimized out>, context=0x7ffd1adb9d10)
    at /usr/src/debug/gcc/libstdc++-v3/libsupc++/eh_personality.cc:688
#7  0x00007f827f286774 in _Unwind_RaiseException_Phase2 (exc=0x55aa14094c10, context=0x7ffd1adb9d10, frames_p=0x7ffd1adb9c18) at /usr/src/debug/gcc/libgcc/unwind.inc:64
[...]

PR #4360, opened 2023-02-08
Attempt to catch `std::bad_function_call` when adding a channel point reward

Nerixyz:
Pull request checklist:

# Description

Related to https://github.com/Chatterino/chatterino2/issues/3942#issuecomment-1421226657. This attempts to catch a [`std::bad_function_call`](https://en.cppreference.com/w/cpp/utility/functional/bad_function_call). **This doesn't resolve the underlying issue that causes this to be thrown.** Furthermore, there are probably more issues related to this, as some crash reports suggest that no exception is thrown.

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -315,7 +315,7 @@ void TwitchChannel::addChannelPointReward(const ChannelPointReward &reward)
         {
             qCWarning(chatterinoTwitch).nospace()
                 << "[TwitchChannel " << this->getName()
-                << "] Cought std::bad_function_call when adding channel point "
+                << "] Caught std::bad_function_call when adding channel point "
                    "reward ChannelPointReward{ id: "
                 << reward.id << ", title: " << reward.title << " }.";
         }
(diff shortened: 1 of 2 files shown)


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -304,7 +304,21 @@ void TwitchChannel::addChannelPointReward(const ChannelPointReward &reward)
...
+        // This only attempts to prevent a crash when invoking the signal.
+        try
+        {
+            this->channelPointRewardAdded.invoke(reward);
+        }
+        catch (const std::bad_function_call &)
+        {
+            qCWarning(chatterinoTwitch).nospace()
+                << "[TwitchChannel " << this->getName()
+                << "] Cought std::bad_function_call when adding channel point "
```
leon-richardt, line 318:
```suggestion
                << "] Caught std::bad_function_call when adding channel point "
```

:sunglasses:

Nerixyz: oops

pajlada (approved):

PR #4362, opened 2023-02-09
use AssocQueryString instead of directly querying the registry

sheybey:
Pull request checklist:

The existing getCommand() in IncognitoBrowser.cpp directly queries undocumented registry keys which could break if Windows decides to change the registry layout (as they have in the past.)

This PR changes getCommand() to instead call AssocQueryString, which smooths out differences between versions of Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - Bugfix: Fixed an issue on Windows when opening links in incognito mode that contained forward slashes in hash (#4307)
 - Bugfix: Fixed an issue where beta versions wouldn't update to stable versions correctly. (#4329)
 - Bugfix: Avoided crash that could occur when receiving channel point reward information. (#4360)
+- Bugfix: Don't rely on undocumented registry keys to find the default browser on Windows. (#4362)
 - Dev: Changed sound backend from Qt to miniaudio. (#4334)
 - Dev: Remove protocol from QApplication's Organization Domain (so changed from `https://www.chatterino.com` to `chatterino.com`). (#4256)
 - Dev: Ignore `WM_SHOWWINDOW` hide events, causing fewer attempted rescales. (#4198)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,6 @@
 - Bugfix: Fixed an issue on Windows when opening links in incognito mode that contained forward slashes in hash (#4307)
 - Bugfix: Fixed an issue where beta versions wouldn't update to stable versions correctly. (#4329)
 - Bugfix: Avoided crash that could occur when receiving channel point reward information. (#4360)
-- Bugfix: Don't rely on undocumented registry keys to find the default browser on Windows. (#4362)
 - Dev: Changed sound backend from Qt to miniaudio. (#4334)
 - Dev: Remove protocol from QApplication's Organization Domain (so changed from `https://www.chatterino.com` to `chatterino.com`). (#4256)
 - Dev: Ignore `WM_SHOWWINDOW` hide events, causing fewer attempted rescales. (#4198)
@@ -48,6 +47,7 @@
 - Dev: Added CMake Install Support on Windows. (#4300)
 - Dev: Changed conan generator to [`CMakeDeps`](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html) and [`CMakeToolchain`](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmaketoolchain.html). See PR for migration notes. (#4335)
 - Dev: Refactored 7TV EventAPI implementation. (#4342)
+- Dev: Don't rely on undocumented registry keys to find the default browser on Windows. (#4362)
 
 ## 2.4.0
 
--- a/src/util/IncognitoBrowser.cpp
+++ b/src/util/IncognitoBrowser.cpp
@@ -4,19 +4,15 @@
 #include <QRegularExpression>
 #include <QVariant>
 
-#if defined(Q_OS_WIN) and defined(USEWINSDK)
-#    include <Shlwapi.h>
-#    include <VersionHelpers.h>
-
-typedef HRESULT(CALLBACK *AssocQueryString_)(ASSOCF, ASSOCSTR, LPCWSTR, LPCWSTR,
-                                             LPWSTR, DWORD *);
+#ifdef USEWINSDK
+#    include "WindowsHelper.hpp"
 #endif
 
 namespace {
 
 using namespace chatterino;
 
-#if defined(Q_OS_WIN) and defined(USEWINSDK)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Bugfix: Fixed an issue on Windows when opening links in incognito mode that contained forward slashes in hash (#4307)
 - Bugfix: Fixed an issue where beta versions wouldn't update to stable versions correctly. (#4329)
 - Bugfix: Avoided crash that could occur when receiving channel point reward information. (#4360)
+- Bugfix: Don't rely on undocumented registry keys to find the default browser on Windows. (#4362)
```
Felanbird, line 33: If we Changelog this it should be `Dev:` and at the bottom

src/util/IncognitoBrowser.cpp
```diff
@@ -45,25 +52,71 @@ QString injectPrivateSwitch(QString command)
     return QString();
 }
 
+QString callAssocQueryString(ASSOCF flags, ASSOCSTR str, LPCWSTR pszAssoc,
```
pajlada, line 55: Could we move this function into `src/util/WindowsHelper.hpp/cpp` with an "as-minimal-as-possible" wrapper?
By as-minimal-as-possible I mean don't have the users of the function (`getCommand` in this case) pass in any arguments that we always set to null.
If possible, rename the function to something that describes its intent more clearly

pajlada (changes requested):
Overall the functionality of this change looks good, just some refactoring to leave the windows-specific implementation in its own file would be nice.

As a consequence of this change, Windows builds that do build without the Windows SDK enabled will no longer support incognito links. This is something I'm fine with.

src/util/IncognitoBrowser.cpp
sheybey: There is a flag that callers could pass in that would require them to pass in something to pszExtra, related to looking up programs by exe name. I tried to leave in the most flexibility possible for this wrapper, so I didn't prevent the caller from passing in that flag (and therefore left pszExtra to the caller.)

On second thought, it's pretty unlikely that chatterino would need to use that lookup, so I'll make the wrapper much more specific and only support protocol and file extension lookups.

sheybey:
> Overall the functionality of this change looks good, just some refactoring to leave the windows-specific implementation in its own file would be nice.

I also plan to implement linux support (well, freedesktop support) for this in a separate PR. Should I create a new XDGHelper or LinuxHelper file for that?

pajlada:
> > Overall the functionality of this change looks good, just some refactoring to leave the windows-specific implementation in its own file would be nice.
> 
> I also plan to implement linux support (well, freedesktop support) for this in a separate PR. Should I create a new XDGHelper or LinuxHelper file for that?

Yes, for that sort of functionality additional OS-specific helper files in the same directory is fine (i.e. LinuxHelpers.cpp/hpp)

pajlada (changes requested):
Nitpicking as it's code I'm mostly unfamiliar with Windows APIs, thank you!

src/util/WindowsHelper.cpp
```diff
@@ -83,6 +88,54 @@ void setRegisteredForStartup(bool isRegistered)
...
+    {
+        // ASSOCF_IS_PROTOCOL was introduced in Windows 8
+        if (IsWindows8OrGreater())
+        {
+            flags |= ASSOCF_IS_PROTOCOL;
+        }
+        else
+        {
+            return QString();
+        }
```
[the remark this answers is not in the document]
sheybey: The documentation doesn't specify, it just says this flag was introduced in Windows 8. I assumed it was possible that it would be ignored, causing the protocol to be looked up as a file extension.

Theoretically the lookup would always fail even in that case since it doesn't have a period in front of it, but I figured it would be more correct and efficient to just error out early. COM calls aren't especially cheap.

src/util/WindowsHelper.cpp
```diff
@@ -83,6 +88,54 @@ void setRegisteredForStartup(bool isRegistered)
...
+                          : GetProcAddress(shlwapi, "AssocQueryStringW"));
+
+    if (assocQueryString == NULL)
+    {
+        return QString();
+    }
+
+    // always error out instead of returning a truncated string when the
+    // buffer is too small - avoids race condition when the user changes their
+    // default browser between calls to AssocQueryString
```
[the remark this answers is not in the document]
sheybey: Documentation says it allows the function to access the disk to verify info it gets from the registry. It gives verifying the app friendly name by reading the EXE header as an example, but doesn't specify exactly what it allows.

It doesn't seem to make any difference with the restricted set of parameters I'm passing, so I'll remove it.

src/util/WindowsHelper.cpp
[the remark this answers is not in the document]
pajlada: Alright, that's ok by me!

src/util/WindowsHelper.cpp
```diff
@@ -119,14 +119,15 @@ QString getAssociatedCommand(ASSOCIATION_QUERY_TYPE queryType, LPCWSTR query)
     }
 
     DWORD resultSize = 0;
-    assocQueryString(flags, ASSOCSTR_COMMAND, query, NULL, NULL, &resultSize);
-    if (resultSize == 0)
+    if (!SUCCEEDED(assocQueryString(flags, ASSOCSTR_COMMAND, query, NULL, NULL,
```
pajlada, line 122: Last nitpick: Let's use `FAILED(` instead of `!SUCCEEDED` here

pajlada (approved):
One last nitpick, otherwise looks good to me. @kornes if you have time to take a look that would be appreciated :heart:

src/util/WindowsHelper.cpp
```diff
@@ -83,6 +88,55 @@ void setRegisteredForStartup(bool isRegistered)
     }
 }
 
+QString getAssociatedCommand(AssociationQueryType queryType, LPCWSTR query)
+{
+    static HINSTANCE shlwapi = LoadLibrary(L"shlwapi");
+    static auto assocQueryString = AssocQueryString_(
+        (shlwapi == NULL) ? NULL
+                          : GetProcAddress(shlwapi, "AssocQueryStringW"));
```
kornes, line 96: would be more readable to return early when LoadLibrary fails instead of ternary use here

kornes (commented):
good change, no issues just 1 nitpick :+1:
tested on win7 vm and incognito opens correctly, it fallbacks to .html assoc as expected.

pajlada:
@sheybey I pushed two commits, are you ok with a merge after the changes I made?

sheybey:
> @sheybey I pushed two commits, are you ok with a merge after the changes I made?

Looks good to me

Felanbird (approved):
works on windows 10 but I doubt anyone thought it wouldn't

PR #4364, opened 2023-02-11
Implement type checking/validation for filters

### LINKED ISSUE #2466: Filters are not working
**Describe the bug**
I'm currently trying to test either of the below filters:

`(author.name) != ("user1" || "user2")`
or
`(author.name) != (r"(user1|user2)")`
or
`(author.name) != "user1" || "user2"`
or
`author.name != r"(user1|user2)"`

None of which are filtering out the messages from either user.

**To reproduce**
1. Create a filter as mentioned above
2. Set the filter
3. Notice it doesn't work

**Screenshots**
image

image

image

image

**Chatterino version**
Nightly 2.2.3-beta2 (commit 7d9f4c2b0, )

**Operating system**
Windows 10

dnsge:
Pull request checklist:

# Description

This PR adds type-checking to filters. For example, the filter `channel.name == "forsen" && author.badges contains "moderator"` can be deduced to evaluate to the type `Bool`. This allows us to tell the user why their filter is not working. Before, attempting to perform invalid operations like `String - Int` would silently fail while the filter would purport validity.

This PR was primarily motivated by witnessing conversations in Discord where users were confused about why apparently-valid filters weren't working. Additionally, this PR separates the parsing and execution of filters so the filtering language can be easily used in other parts of Chatterino (if another use-case comes up, like advanced highlighting conditions).

This PR also adjusts some (useless) semantics (for example, `5 + "2"` no longer `== 7` nor `== "7"`), so the wiki will require some minor updates.

_Most of the code change comes from splitting Expressions into multiple files. `synthesizeType()` is the primary addition in that realm._

## Example of detecting type errors

Consider the filter `message.content contains ("hello" || "goodbye")`. Someone could conceivably believe this filter allows messages that contain either "hello" or "goodbye" (in fact, this example was taken from Discord). Of course, we can't compute the logical or of two strings, so this filter would silently fail. 

Now, the filters page reports that there was an issue with the filter, and opening the detailed output gives:
image

Closes #2466 (in my opinion)


src/controllers/filters/lang/Filter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        {"flags.reply", m->flags.has(MessageFlag::ReplyMessage)},
+        {"flags.automod", m->flags.has(MessageFlag::AutoMod)},
+
+        {"message.content", m->messageText},
+        {"message.length", m->messageText.length()},
+    };
+    {
+        using namespace chatterino;
+        auto *tc = dynamic_cast<TwitchChannel *>(channel);
+        if (channel && !channel->isEmpty() && tc)
```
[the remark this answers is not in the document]
dnsge: wow! clang-tidy knows that `tc->isLive()` always evaluates to `false`!

src/controllers/filters/lang/Filter.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+class Channel;
+struct Message;
+using MessagePtr = std::shared_ptr<const Message>;
+
+}  // namespace chatterino
+
+namespace chatterino::filters {
+
+static const QMap<QString, Type> messageTypingContext = {
```
pajlada, line 21: I'd prefer if this is `MESSAGE_TYPING_CONTEXT` too (rare clang-tidy agree), alongside a small comment explaining this

src/controllers/filters/lang/Filter.hpp
```diff
@@ -0,0 +1,72 @@
...
+    {"flags.system_message", Type::Bool},
+    {"flags.reward_message", Type::Bool},
+    {"flags.first_message", Type::Bool},
+    {"flags.elevated_message", Type::Bool},
+    {"flags.cheer_message", Type::Bool},
+    {"flags.whisper", Type::Bool},
+    {"flags.reply", Type::Bool},
+    {"flags.automod", Type::Bool},
+    {"message.content", Type::String},
+    {"message.length", Type::Int}};
```
pajlada, line 43:
```suggestion
    {"message.length", Type::Int},
};
```

Comma at the end for easier diffing in the future

src/controllers/filters/lang/Filter.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+ContextMap buildContextMap(const MessagePtr &m, chatterino::Channel *channel);
+
+class Filter;
+struct FilterError {
+    QString message;
+};
+
+using FilterResult = std::variant<Filter, FilterError>;
```
pajlada, line 52: I'm fine with this :+1:

pajlada (changes requested):
I'm not super fond of how the `PossibleType` class is set up, could we utilize a variant there instead? I think the variant `operator==` forwards it to the contained class so it should still be easy to make the API easy to use.

Thank you for the inclusion of the tests - this looks good overall, there's most likely going to be another round or two of small nitpicks that are found as I dig more into this feature.

src/controllers/filters/lang/Filter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+        else
+        {
+            vars["channel.live"] = false;
+        }
+    }
+    return vars;
+}
+
+FilterResult Filter::fromString(const QString &str)
```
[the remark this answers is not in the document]
dnsge: literally invalid to put static here but thanks

`error: 'static' can only be specified inside the class definition`

dnsge:
> I'm not super fond of how the PossibleType class is set up, could we utilize a variant there instead? I think the variant operator== forwards it to the contained class so it should still be easy to make the API easy to use.

I went ahead and reimplemented `PossibleType` to make use of `std::variant` internally (which ends up being much cleaner) in [5474c93](https://github.com/Chatterino/chatterino2/pull/4364/commits/5474c93e2e4b99c40101d3d969ed091d25a0e039). 

However, I feel like your comment was more suggesting replacing `PossibleType` altogether, e.g. 
```cpp
using PossibleType = std::variant<Type, IllTyped>;
``` 
This of course would be possible, but I thought that having a wrapper class would still be useful to have utility methods like `string()` and even `well()`, given the lovely syntax for checking variants (`std::holds_alternative<Type>(variant)`).

If you'd like it to make use of only a variant instead of a wrapper, I can implement that.

pajlada:
I'm fine with both `IllTyped` & `PossibleType` being classes as they will definitely have some member functions that help them be used, my thinking was more along the lines of:
```c++
using PossibleType = std::variant<TypeClass, IllTyped>;

class TypeClass {
  public:
    QString string() const;
    Type unwrap() const;

    bool operator==(Type) const;
    bool operator==(TypeClass) const;
    bool operator==(IllTyped) const;

  private:
  Type type_;
};

class IllTyped {
  public:
    QString description() const;

    bool operator==(Type) const;
    bool operator==(TypeClass) const;
    bool operator==(IllTyped) const;
};
```

Then in places where we've previously checked `if !possibleType.well() { print(possibleType.illTypedDescription()); }` we'd instead first unwrap it to either a `TypeClass` or `IllTyped`

Happy to hear your thoughts on that sort of solution - I can't see a downside to it but it might become apparent when it's actually done & used :dancers:

dnsge:
Updated to use `std::variant` directly :+1:

pajlada (approved):
Looks good to me :+1:

@dnsge if you can take a look at the changes I've made, we can get this merged in

dnsge:
@pajlada Everything looks good 👍

PR #4374, opened 2023-02-13
Automatically update `nightly-build` tag every nightly release

### LINKED ISSUE #3443: Fix Nightly build 'commits since this release'
Basically just a reminder issue, github changed something so nightly builds appear to be behind on commits despite having the most recent changes.

```
[2021-12-26 00:24:07] #pajlada pajlada: Error cleaning up old releases: HttpError: Resource not accessible by integration
[2021-12-26 00:24:17] #pajlada pajlada: github api change
[2021-12-26 00:24:37] #pajlada pajlada: probably solvable by giving it another type of token or something
```

Wissididom:
Pull request checklist:

# Description

This PR makes the nightly build workflow update the tag to have up-to-date source code options on the Nightly Release.

References/Proof:
https://github.com/ncipollo/release-action/issues/31
https://i.imgur.com/zkKPrem.png

(felanbird): Fixes #3443

### DIFF
--- a/.CI/CreateUbuntuDeb.sh
+++ b/.CI/CreateUbuntuDeb.sh
@@ -48,7 +48,7 @@ elif [ "$chatterino_version" = "ightly-build" ]; then
     chatterino_version=$(git describe --abbrev=0 --tags `git rev-list --tags --skip=1 --max-count=1` 2>/dev/null | cut -c 2-) || true
     echo "Found Chatterino nightly build via git. Falling back to the second to last tag"
     if [ -z "$chatterino_version" ]; then
-        # Fall back to this in case the build happened in a git repo without tags
+        # Fall back to this in case the build happened in a git repo with just the nightly-build tag
         chatterino_version="0.0.0-dev"
     fi
 fi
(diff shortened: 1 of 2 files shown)


.CI/CreateUbuntuDeb.sh
```diff
@@ -42,8 +42,15 @@ fi
 
 chatterino_version=$(git describe 2>/dev/null | cut -c 2-) || true
 if [ -z "$chatterino_version" ]; then
-    # Fall back to this in case the build happened outside of a git repo
+    # Fall back to this in case the build happened outside of a git repo or a repo without tags
     chatterino_version="0.0.0-dev"
+elif [ "$chatterino_version" = "ightly-build" ]; then
```
8thony, line 47:
```suggestion
elif [ "$chatterino_version" = "nightly-build" ]; then
```

8thony (changes requested):

.CI/CreateUbuntuDeb.sh
Wissididom: That is because the earlier call uses `cut -c 2-` which removes the first char, so it is intentional that the `n` is missing.

pajlada: I don't think we need to handle the case where `chatterino-version` is `nightly-build`, it's only ever relevant for forks that have no tags.

If you want to improve this, I'd rather it be done in a separate PR so we can nitpick over it without blocking the main purpose of the PR

pajlada (approved):

PR #4375, opened 2023-02-13
Handle non-versioned annotated tags gracefully when building a Ubuntu .deb package

Wissididom:
Pull request checklist:

# Description

pajlada: "If you want to improve this, I'd rather it be done in a separate PR so we can nitpick over it without blocking the main purpose of the PR" from https://github.com/Chatterino/chatterino2/pull/4374

### DIFF
--- a/.CI/CreateUbuntuDeb.sh
+++ b/.CI/CreateUbuntuDeb.sh
@@ -41,8 +41,8 @@ if [ ! -f ./bin/chatterino ] || [ ! -x ./bin/chatterino ]; then
 fi
 
 chatterino_version=$(git describe 2>/dev/null) || true
-if [[ ${chatterino_version:0:1} == 'v' ]]; then
-    chatterino_version=${chatterino_version:1}
+if [ "$(echo "$chatterino_version" | cut -c1-1)" = 'v' ]; then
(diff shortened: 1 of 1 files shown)


Felanbird:
I mean there's no world where we are doing this and calling it `ightly-build` right? 
If we are going to make this change we have to refactor this area to strip where we need it and not strip where we don't.

pajlada:
I think a better approach is to see if the tag we found starts with v, and if it does, cut the first character and use that as the version. if the tag we found does not start with v, just use `0.0.0-dev`

Wissididom:
The initial thought why I implemented it was that nightly-build became the latest release after force pushing so 2.4.0 wasn't the latest tag anymore

Wissididom:
Tested latest commit with
```
chatterino_version="v2.4.0"
#chatterino_version="nightly-build"
#chatterino_version=""
if [[ ${chatterino_version:0:1} == 'v' ]]; then
    chatterino_version=${chatterino_version:1}
else
    chatterino_version="0.0.0-dev"
fi
echo $chatterino_version
```
on my local machine while changing the first three line each try

pajlada (approved):

PR #4393, opened 2023-02-19
Fix Qt6 building

pajlada:
Pull request checklist:

# Description

- Add Qt Core5Compat dependency if building with Qt6
- Add specialization to `localizeNumber` that accepts a `qsizetype`
- Add missing `QFile` include to `SoundController.cpp`
- Add missing `QFile` include to `SettingsDialog.cpp`
- Add missing `QFile` include to `AboutPage.cpp`
- Format IRC username-randomizer in IrcServer using QString arg
- Comment out usage of `setCodec` in `LinkParser.cpp` for Qt6
- Comment out usage of `setCodec` in `AboutPage.cpp` for Qt6
- Store max spaces as a `qsizetype` in `CommandController`
- Use QStringView.split instead of QString splitRef
- `int`ify some usages of `size`
- Replace QStringRef with QStringView in TwitchMessageBuilder
- Replace `QStringRef` with `QStringView` in ResizingTextEdit
- Clarify QPoint constructor in SplitInput.cpp
- Use `initFrom` instead of `init` in QStyleOption
- Call `result` on QConcurrent runs in ModerationPage
- Add github workflow for building on Qt6

### DIFF
--- a/src/providers/twitch/TwitchIrcServer.cpp
+++ b/src/providers/twitch/TwitchIrcServer.cpp
@@ -325,7 +325,7 @@ std::shared_ptr<Channel> TwitchIrcServer::getChannelOrEmptyByID(
             continue;
 
         if (twitchChannel->roomId() == channelId &&
-            twitchChannel->getName().splitRef(":").size() < 3)
+            QStringView{twitchChannel->getName()}.split(':').size() < 3)
         {
             return twitchChannel;
         }
--- a/src/providers/twitch/TwitchIrcServer.cpp
+++ b/src/providers/twitch/TwitchIrcServer.cpp
@@ -325,7 +325,7 @@ std::shared_ptr<Channel> TwitchIrcServer::getChannelOrEmptyByID(
             continue;
 
         if (twitchChannel->roomId() == channelId &&
-            QStringView{twitchChannel->getName()}.split(':').size() < 3)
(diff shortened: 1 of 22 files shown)


src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -325,7 +325,7 @@ std::shared_ptr<Channel> TwitchIrcServer::getChannelOrEmptyByID(
             continue;
 
         if (twitchChannel->roomId() == channelId &&
-            twitchChannel->getName().splitRef(":").size() < 3)
+            QStringView{twitchChannel->getName()}.split(':').size() < 3)
```
Nerixyz, line 328:
```suggestion
            twitchChannel->getName().count(':') < 2)
```
No need to split.

Nerixyz:
This PR should also remove the following line:
https://github.com/Chatterino/chatterino2/blob/d3499e814e3ecb303faf5610652b0965e2f3d047/src/PrecompiledHeader.hpp#L108

This header moved to `QtGui` but it's included above either way:
https://github.com/Chatterino/chatterino2/blob/d3499e814e3ecb303faf5610652b0965e2f3d047/src/PrecompiledHeader.hpp#L16

So it can be removed. CI doesn't catch this, since the Qt 6 runner has PCH disabled.

PR #4402, opened 2023-02-22
Use `unique_lock` when loading 7TV badges

Nerixyz:
Pull request checklist:

# Description

While looking through the code I found that `SeventvBadges` used a `shared_lock` while writing to the locked data. This wasn't an issue, because the badges are only ever loaded once at the start.


pajlada (changes requested):
Remove the addition of the `const` modifier, otherwise looks good to me

pajlada (approved):

PR #4405, opened 2023-02-24
fix: update user-facing emote picker labels to use "7TV" naming

jammehcow:
Pull request checklist:

# Description

This PR aims to update the label for 7TV emotes that is shown to users when searching for emotes in the emote picker. When browsing emotes, the section's label uses the `7TV` name; however, when searching for emotes in the same popup window, the name `SevenTV` was used. 

As suggested by the 7TV crew (I believe Anatole was the one who we had this convo with), any user-facing branding should use the name `7TV`. Only code should reference the name `SevenTV`

Thanks for finding it Malachi :)


jammehcow:
I don't feel like this needs a changelog entry. Nobody will get excited about a label change.

Felanbird (approved):

PR #4411, opened 2023-02-25
Add ban user by id command `/banid`

iProdigy:
Pull request checklist:

# Description

With the `/ban` command migrated to Helix, we are no longer able to ban suspended or deactivated accounts because helix `getUsers` is not able to resolve the `user_id` associated with the login name. However, if a moderator already knows the `user_id`, helix `banUser` *is* able to ban the account, even if suspended or deactivated. Thus, this PR adds a `/banid` command to allow banning users by id.

Note: similar concerns apply for timeout/unban, but are not covered by this PR (and are not as important imo)

chatterino_W6yakM3S6Z

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
-- Minor: Added `/banid` command that allows banning by userID. (#4411)
+- Minor: Added `/banid` command that allows banning by user ID. (#4411)
 - Bugfix: Fixed uploaded AppImage not being able most web requests. (#4400)
 - Bugfix: Fixed a potential race condition due to using the wrong lock when loading 7TV badges. (#4402)
 - Dev: Add capability to build Chatterino with Qt6. (#4393)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
+- Minor: Added `/banid` command that allows banning by user ID. (#4411)
 - Bugfix: Fixed uploaded AppImage not being able most web requests. (#4400)
 - Bugfix: Fixed a potential race condition due to using the wrong lock when loading 7TV badges. (#4402)
 - Dev: Add capability to build Chatterino with Qt6. (#4393)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
-- Minor: Added `/banid` command that allows banning by user ID. (#4411)
+- Minor: Added `/banid` command that allows banning by userID. (#4411)
 - Bugfix: Fixed uploaded AppImage not being able most web requests. (#4400)
 - Bugfix: Fixed a potential race condition due to using the wrong lock when loading 7TV badges. (#4402)
 - Dev: Add capability to build Chatterino with Qt6. (#4393)
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -2842,7 +2842,7 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         const auto *usageStr =
             "Usage: \"/banid <userID> [reason]\" - Permanently prevent a user "
-            "from chatting via their userID. Reason is optional and will be "
+            "from chatting via their user ID. Reason is optional and will be "
             "shown to the target user and other moderators. Use \"/unban "
             "<username>\" to remove a ban.";
         if (words.size() < 2)
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -2841,10 +2841,10 @@ void CommandController::initialize(Settings &, Paths &paths)
         }
 
         const auto *usageStr =
-            "Usage: \"/banid <userid> [reason]\" - Permanently prevent a user "
-            "ID from chatting. Reason is optional and will be shown to the "
-            "target user and other moderators. Use \"/unban <username>\" to "
-            "remove a ban.";
+            "Usage: \"/banid <userID> [reason]\" - Permanently prevent a user "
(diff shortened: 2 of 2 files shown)


src/controllers/commands/CommandController.cpp
```diff
@@ -2829,6 +2829,54 @@ void CommandController::initialize(Settings &, Paths &paths)
         return "";
     });
 
+    this->registerCommand("/banid", [formatBanTimeoutError](
```
iProdigy, line 2832: Open to other suggestions for the command name

src/controllers/commands/CommandController.cpp
```diff
@@ -2829,6 +2829,54 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        getHelix()->banUser(
+            twitchChannel->roomId(), currentUser->getUserId(), target,
+            boost::none, reason,
+            [] {
+                // No response for bans, they're emitted over pubsub/IRC instead
+            },
+            [channel, target, formatBanTimeoutError](auto error, auto message) {
+                auto errorMessage =
+                    formatBanTimeoutError("ban", error, message, "#" + target);
+                channel->addMessage(makeSystemMessage(errorMessage));
```
iProdigy, line 2874: Unfortunately, Twitch doesn't send us an informative error message when a user tries to pass a login name instead of ID to the command:

```json
{
  "data": [],
  "error": "Internal Server Error",
  "status": 500
}
```

So this results in: `Failed to ban user - An unknown error has occurred.`

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
+- Minor: Added `/banid` command that allows banning by user ID. (#4411)
```
```suggestion
- Minor: Added `/banid` command that allows banning by userID. (#4411)
```
I have a vague memory of some argument why mentions of userids should be `userID` and while I don't remember what we we decided on or even if it was for chatterino/pajbot, but I think no space is more correct

src/controllers/commands/CommandController.cpp
```diff
@@ -2829,6 +2829,54 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            channel->addMessage(makeSystemMessage(
+                QString("The /banid command only works in Twitch channels")));
+            return "";
+        }
+
+        const auto *usageStr =
+            "Usage: \"/banid <userid> [reason]\" - Permanently prevent a user "
+            "ID from chatting. Reason is optional and will be shown to the "
+            "target user and other moderators. Use \"/unban <username>\" to "
+            "remove a ban.";
```
Felanbird, line 2847:
```suggestion
            "Usage: \"/banid <userID> [reason]\" - Permanently prevent a user "
            "from chatting via their userID. Reason is optional and will be "
            "shown to the target user and other moderators. Use \"/unban "
            "<username>\" to remove a ban.";
```

src/controllers/commands/CommandController.cpp
Felanbird: this name is fine for me

src/controllers/commands/CommandController.cpp
Felanbird: I'm not _against_ validating this against a `\d+` regex as I'm doubtful Twitch will ever send a correct error for this, but we could also just do nothing

Felanbird (commented):
Functionality seems to work as I would expect it to 👍

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
+- Minor: Added `/banid` command that allows banning by userID. (#4411)
```
pajlada, line 6:
```suggestion
- Minor: Added `/banid` command that allows banning by user ID. (#4411)
```

src/controllers/commands/CommandController.cpp
```diff
@@ -2829,6 +2829,54 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+        if (twitchChannel == nullptr)
+        {
+            channel->addMessage(makeSystemMessage(
+                QString("The /banid command only works in Twitch channels")));
+            return "";
+        }
+
+        const auto *usageStr =
+            "Usage: \"/banid <userID> [reason]\" - Permanently prevent a user "
+            "from chatting via their userID. Reason is optional and will be "
```
pajlada, line 2845:
```suggestion
            "from chatting via their user ID. Reason is optional and will be "
```

src/controllers/commands/CommandController.cpp
```diff
@@ -2829,6 +2829,54 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+            "shown to the target user and other moderators. Use \"/unban "
```
pajlada, line 2846: Does `/unban` work for usernames that are already banned? If not, we should not mention `/unban` here imo, not unless we have an `/unbanid` equivalent

src/controllers/commands/CommandController.cpp
pajlada: I feel like this is a bit of an advanced command, it shouldn't be necessary to put too much effort into parsing weird user errors.

pajlada (changes requested):
Small nitpicks, haven't tested functionality yet but it looks simple enough :+1:

CHANGELOG.md
Felanbird: none of the antonyms for `concatenate` are very exciting 😢

Mm2PL:
> Note: similar concerns apply for timeout/unban, but are not covered by this PR (and are not as important imo)

Maybe special syntax could be made so all `/(un)timeout`, `/(un)ban` commands simply work from the user's point of view, for example: `/ban ["id"] (name) (reason)` or `/ban #(id) (reason)`.

pajlada:
> > Note: similar concerns apply for timeout/unban, but are not covered by this PR (and are not as important imo)
> 
> Maybe special syntax could be made so all `/(un)timeout`, `/(un)ban` commands simply work from the user's point of view, for example: `/ban ["id"] (name) (reason)` or `/ban #(id) (reason)`.

If relevant use-cases come up in the future, we'll figure out a uniform solution for this that can be applied to all commands but for now we'll go with this simple one that has an actively request use-case.

pajlada (approved):

PR #4432, opened 2023-03-05
Fix Handling of FFZ CDN URLs with https already appended

### LINKED ISSUE #4430: FFZ emotes no longer work
FFZ used to send us API responses as follows `//cdn.frankerfacez.com/xddddd` without the necessary `https:`, so we manually added the `https:` so we could correctly display emotes.

As of a few hours ago, FFZ has resolved this, but unfortunately we did not want this resolved, and as a result chatterino now sends requests to `https:https://cdn.frankerfacez.com/xddddd` and said requests fail.

https://github.com/Chatterino/chatterino2/blob/1dc2bcc445d50378ed6c6664cf009373cbf7b0a7/src/providers/ffz/FfzBadges.cpp#L70-L77
https://github.com/Chatterino/chatterino2/blob/1dc2bcc445d50378ed6c6664cf009373cbf7b0a7/src/providers/ffz/FfzEmotes.cpp#L19-L29

dnsge:
Pull request checklist:

# Description
 
Fixes #4430

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
 - Minor: Added `/banid` command that allows banning by user ID. (#4411)
+- Bugfix: Fixed FrankerFaceZ emotes/badges not loading due to API change. (#4432)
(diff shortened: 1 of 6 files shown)


zneix (approved):

Felanbird (approved):
also my review that it works 👍 
image

PR #4434, opened 2023-03-05
Support Animated FFZ Emotes and Authors for Global Emotes

Nerixyz:
Pull request checklist:

# Description

This PR adds support for animated FFZ emotes and shows authors for global emotes:

https://user-images.githubusercontent.com/19953266/222955005-3c5dfc09-2162-4745-ad9f-9b355262e89a.mp4

---

I refactored the code a bit to reduce duplicated parsing of channel/global emotes and to make it easier to read (by replacing [`QJsonObject::value`](https://doc.qt.io/qt-5/qjsonobject.html#value) with [`QJsonObject::operator[]`](https://doc.qt.io/qt-5/qjsonobject.html#operator-5b-5d)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
 - Minor: Added `/banid` command that allows banning by user ID. (#4411)
+- Minor: Added support for FrankerFaceZ animated emotes. (#4434)
 - Bugfix: Fixed FrankerFaceZ emotes/badges not loading due to API change. (#4432)
 - Bugfix: Fixed uploaded AppImage not being able most web requests. (#4400)
 - Bugfix: Fixed a potential race condition due to using the wrong lock when loading 7TV badges. (#4402)
--- a/src/providers/ffz/FfzEmotes.cpp
+++ b/src/providers/ffz/FfzEmotes.cpp
@@ -69,6 +69,11 @@ namespace {
             auto author =
                 EmoteAuthor{emoteJson["owner"]["display_name"].toString()};
             auto urls = emoteJson["urls"].toObject();
+            if (emoteJson["animated"].isObject())
+            {
+                // prefer animated images if available
+                urls = emoteJson["animated"].toObject();
+            }
 
             Emote emote;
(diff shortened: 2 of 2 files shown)


src/providers/ffz/FfzEmotes.cpp
```diff
@@ -47,68 +52,94 @@ namespace {
...
+            case FfzEmoteKind::Channel:
+                return QStringLiteral("Channel");
+            case FfzEmoteKind::Global:
+                return QStringLiteral("Global");
+            default:
+                return QStringLiteral("(invalid kind)");
+        }
+    }
+
+    void parseEmoteSetInto(const QJsonObject &emoteSet, FfzEmoteKind kind,
```
pajlada, line 77: Let's accept the emote kind as a string instead of an enum, since all we do with it is convert it to a string

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Minor: Delete all but the last 5 crashdumps on application start. (#4392)
 - Minor: Added `/banid` command that allows banning by user ID. (#4411)
+- Minor: FrankerFazeZ animated emotes and authors for global emotes are now shown. (#4434)
```
pajlada, line 7: Something like "Added support for FrankerFaceZ animated emotes.", and if we want a changelog entry for the author then just put that as a separate changelog entry as its value is a lot lower

pajlada (changes requested):
For simplicity in reviewing, could you rebase the animated URL check into a separate commit from the refactors? Moving code (into a function in this case) & changing something in that function hides the functional changes that were made

not feedback, just notes for myself:
dev: This changes a bunch of `toObject().value(...)` calls to `[...]`. For QJsonValue's they do the same thing
dev: This changes a few `value(...)` calls to `[...]`. For QJsonObject's, they do the same thing

CHANGELOG.md
Nerixyz: The author addition just comes from the unification of parsing - I don't think it's that important to mention.

Felanbird (approved):
chatterino_2023-03-06_14-42-05

pajlada (approved):

Issue #4325, opened 2023-01-22
[Feature request] Segmentation of chat output


4164696f73:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Right now, every user message is basically a one liner, meaning everything is cluttered one after another and there's no way to change the layout of the chat. I wanted to request a segmentation of that, where instead of having one line per user, we have message parts:
- timestamp
- moderation buttons
- user name
- user message

and we can split the user output however we want, so for example we put moderation buttons on line1, timestamp and user on line2 and user's message in line3.

Examples:
![Example1:](https://i.imgur.com/neNWpjt.png)
![Example2:](https://i.imgur.com/BbyttvB.png)
![Example3:](https://i.imgur.com/YuNtJ7U.png)

I don't have a good idea on how to implement it, though, but what I thought about was either a textbox where you specify, e.g. "{timestamp} {moderation}\n{user.name}: {user.message}" etc (notice the new line indication) or a small visualization of a message, which by default says "[timestamp] [moderation] [user.name] [user.message]" and you can drag those boxes around however you like. Can even include lines, so if there are 4 fragments of a message (just like in my example), then there's 4 lines to be filled:
"line1:
line2:
line3: [timestamp] [moderation] [user.name] [user.message]
line4:"
And you can adjust it to whatever you like by dragging each [part] into a different line.

### OS and Chatterino Version

Chatterino 2.4.0 (commit 2ba4da0 modified) built with Qt 5.15.7 Running on KDE Flatpak runtime, kernel: 5.15.85-1-MANJARO, running from Flatpak

pajlada:
This functionality would need to come with a heavy lifting UX for configuring this. It's not something I'm comfortable accepting into Chatterino & continuing to maintain without more of an ask from users, so I'm closing this out for now.
You can of course keep running this in your own build of Chatterino

PR #4435, opened 2023-03-06
Add Setting to Change Message Layout (#4325)

atheridis:
Pull request checklist:

# Description

Implement Feature (https://github.com/Chatterino/chatterino2/issues/4325).

The change should probably use enums rather hardcoding QStrings, but my knowledge of C++ isn't great and I was running into some issues having `layout.addDropdown` return a vector.

The change as it stands allows for hardcoded text to also be applied into the layout (if anyone would want that), but due to the method I chose to parse it (splitting the `QString` on `"\\"`), a backslash can't be outputted.

The default setting should change nothing visually for the user.

### DIFF
--- a/resources/contributors.txt
+++ b/resources/contributors.txt
@@ -57,6 +57,7 @@ Jaxkey | https://github.com/Jaxkey | :/avatars/jaxkey.png | Contributor
 Explooosion | https://github.com/Explooosion-code | :/avatars/explooosion_code.png | Contributor
 mohad12211 | https://github.com/mohad12211 | :/avatars/mohad12211.png | Contributor
 Wissididom | https://github.com/Wissididom | :/avatars/wissididom.png | Contributor
+atheridis | https://github.com/atheridis | | Contributor
 
 # If you are a contributor add yourself above this line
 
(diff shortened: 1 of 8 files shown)


resources/contributors.txt
```diff
@@ -57,6 +57,7 @@ Jaxkey | https://github.com/Jaxkey | :/avatars/jaxkey.png | Contributor
 Explooosion | https://github.com/Explooosion-code | :/avatars/explooosion_code.png | Contributor
 mohad12211 | https://github.com/mohad12211 | :/avatars/mohad12211.png | Contributor
 Wissididom | https://github.com/Wissididom | :/avatars/wissididom.png | Contributor
+atheridis | https://github.com/atheridis | | Contributor
```
Felanbird, line 60: this can be done in a seperate PR once this PR is merged

8thony:
With this pr, I can crash the application when I double-click on a text or an emote.
I think it's due to the double-click/triple-click text selection mechanic.

pajlada:
Closing this reason for the same reason as the original issue #4325

PR #4442, opened 2023-03-10
Only log debug messages when NDEBUG is not defined

### LINKED ISSUE #4440: no way to set DEBUG_OFF in QLogging.cpp
### Checklist

### Describe your issue

Chatterino litters the hell out of your logs with mundane messages (like fetching emotes, or connecting to websocket) and there's no way to turn it off

[This](https://github.com/Chatterino/chatterino2/blob/master/src/common/QLogging.cpp#L3) exists but there's no way to set it since [QMake was removed](https://github.com/Chatterino/chatterino2/commit/044dd8a616e0b712eaf6cb954cd7b999aed18b06) 

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 7.4.2 (commit GIT-REPOSITORY-NOT-FOUND) built with Qt 5.15.8

llyyr:
Regression from https://github.com/Chatterino/chatterino2/commit/044dd8a616e0b712eaf6cb954cd7b999aed18b06

Closes #4440

Pull request checklist:

### DIFF
--- a/src/common/QLogging.cpp
+++ b/src/common/QLogging.cpp
@@ -1,6 +1,6 @@
 #include "common/QLogging.hpp"
 
-#ifdef DEBUG_OFF
+#ifdef NDEBUG
 static constexpr QtMsgType logThreshold = QtWarningMsg;
 #else
 static constexpr QtMsgType logThreshold = QtDebugMsg;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Dev: Only log debug messages when NDEBUG is not defined. (#4442)
+
 ## 2.4.2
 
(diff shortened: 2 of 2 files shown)


src/common/QLogging.cpp
```diff
@@ -1,9 +1,9 @@
 #include "common/QLogging.hpp"
 
-#ifdef DEBUG_OFF
-static constexpr QtMsgType logThreshold = QtWarningMsg;
-#else
+#ifndef NDEBUG
```
pajlada, line 3: Nitpick: if you keep this as `ifdef NDEBUG` this turns into a 1-line diff

pajlada (changes requested):

src/common/QLogging.cpp
llyyr: Done

pajlada (approved):

PR #4455, opened 2023-03-17
Fix dot commands (ie .ban) not being recognized

LittleBigBug:
Pull request checklist:

# Description

Allows you to use commands with . as a prefix again

Using .ban then /ban:

Before (current 2.4.2)

image

With fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Dev: Only log debug messages when NDEBUG is not defined. (#4442)
+- Minor: Fixed dot commands not being recognized. (#4455)
 
 ## 2.4.2
 
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -3194,6 +3194,11 @@ QString CommandController::execCommand(const QString &textNoEmoji,
 
     QString commandName = words[0];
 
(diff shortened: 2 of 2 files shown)


pajlada:
Commands with a dot prefix are meant as a way to bypass a chat client's handling of commands and send them straight to a server, which is the standard for IRC commands. The behaviour you're seeing prior to this PR is expected, since `.ban` is not a valid Twitch IRC command any longer.
Is there any scenario where using `/ban` instead of `.ban` doesn't work for you?

LittleBigBug:
That makes sense, this was just for parity with the twitch web chat client and to get familiar with the project.

> Is there any scenario where using /ban instead of .ban doesn't work for you?

Nope, lol I just created a habit of using . for whatever reason and noticed it stopped working.
If this shouldn't be supported then maybe I can just change the "unrecognized command" message to include the period?

LittleBigBug:
Well, other than this breaking user commands that use a period prefix, is there any case where sending a command directly to the twitch IRC would be useful?
That is, would one of Mm2PL's suggestions be acceptable?
I'm not super set on getting the dot commands to work again so if you definitely prefer not to support this at all that's just fine with me

pajlada:
There are very no `.`-commands that are useful to send to Twitch atm as far as I can tell.
I'm very set on not supporting `.`-commands as an alias to `/`-commands yeah

PR #4463, opened 2023-03-19
Add a local backup of the Twitch Badges API

pajlada:
Pull request checklist:

# Description

In case the Twitch Badges API fails, we now load a snapshot from our resources to ensure badges will continue to work.


pajlada:
image

src/providers/twitch/TwitchBadges.cpp
```diff
@@ -36,58 +37,71 @@ void TwitchBadges::loadTwitchBadges()
...
+
             this->loaded();
             return Success;
         })
         .onError([this](auto res) {
-            qCDebug(chatterinoTwitch)
-                << "Error loading Twitch Badges:" << res.status();
-            // Despite erroring out, we still want to reach the same point
-            // Loaded should still be set to true to not build up an endless queue, and the quuee should still be flushed.
+            qCWarning(chatterinoTwitch)
```
Mm2PL, line 48: nit pick: maybe this could print some system message to chat?

Issue #2237, opened 2020-11-28
Custom user themes


zurrty:
**What should be added?**
Customizable user themes in addition to the four pre-existing themes already present in the program. Depending on which of these is easier to implement, these themes could either be written in a format like CSS, or in a way similar to how [Blender's themes](https://docs.blender.org/manual/en/dev/editors/preferences/themes.html) are managed.
**Why should it be added?**
To give users more freedom when customizing their Chatterino user interface.

Mm2PL:
Related to #1041 and #1837

pajlada:
Themes (beta) are implemented since v2.4.6 - you can read more about how it works here https://wiki.chatterino.com/Themes/

For further feedback on Themes you can make a new issue

PR #4471, opened 2023-03-23
Store Themes as JSON files

Nerixyz:
Pull request checklist:

# Description

This PR removes the manual theme creation and adds theme parsing from JSON (as mentioned in https://github.com/Chatterino/chatterino2/discussions/4447#discussioncomment-5315581, JSON is probably the best format for this). I converted the built-in themes to JSON files that are included in the final executable. The colors are ported 1:1, so there shouldn't be any visual difference. To keep the parsing code small, I'm using a macro that ensures the JSON keys are the same as the C++ property names.

This is working towards #2237. You can already specify custom themes in the settings, but there's neither UI nor auto-reloading nor a wiki entry for it.


pajlada (approved):
Dark theme differences


| old | new | 
| -- | -- |
| old | new |


This has to do with some ancient ifdefs https://github.com/Chatterino/chatterino2/blame/8064f8a49ebf72e9332c0208e3b5b9d54bf3ec4c/lib/appbase/BaseTheme.cpp#L65-L69

I'm ok to concede the window background color so it's the same on Windows & MacOS, but changed on Linux (for now).

LGTM :+1: 
Thank you

pajlada:
I'm also not super fond of the macro solution but I understand there would be some string duplication without it, so I'm fine to leave that to some potential cleanup PR in the future.

PR #4479, opened 2023-03-26
CI: Make prettier print the diff on error

Mm2PL:
Pull request checklist:

- [ ]~~ `CHANGELOG.md` was updated, if applicable~~

# Description

Prettier CI prints the diff now. It annoyed me too much

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Reformat with Prettier
+      - name: Check formatting with Prettier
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -18,8 +18,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Check formatting with Prettier
+      - name: Reformat with Prettier
```
Felanbird, line 21:
```suggestion
      - name: Check formatting with Prettier
```

We are still using this to check formatting, even if we use a convoluted approach now.

Mm2PL: This step now reformats the code and then git diff actually checks if it changed the files.

PR #4498, opened 2023-04-01
Ensure tests have default-initialized settings

pajlada:
- Assert that Settings is initialized when `instance` is called
- Rework HighlightController directory creation/saving to ensure the test directory is written to & cleaned up appropriately
- Ensure settings are default-initialized before any tests are run
- Add changelog entry

Pull request checklist:

# Description


pajlada (commented):
image

PR #4501, opened 2023-04-01
Fix smooth scrolling when ChannelView is full

Pizzabelly:
Pull request checklist:

# Description

This fixes visual glitches related to smooth scrolling when ChannelView is full (removing a message for every message added). The idea is to emulate an infinitely growing list of messages from the perspective of a Scrollbar. This makes it so an offset in it's value has a consistent meaning for the whole animation. Then, reset it at an opportune time (scrolling finished, offset with no animation) to not overflow.


Pizzabelly:
I realized I didn't explain the issue, it's the same issue as #2084. When ChannelView is full, the value of the Scrollbar needs to be changed on every new message because a message was also removed. Seen here https://github.com/Chatterino/chatterino2/blob/master/src/widgets/helper/ChannelView.cpp#L908. This causes the view to jump around and flicker during the smooth scrolling animation. Here is an example of what is looks like without the work-around from #2084, a fast chat, and trackball scrolling (worse case). 

https://user-images.githubusercontent.com/8285260/229935190-d7c7f508-56a5-492d-939f-25bcf009da4f.mp4

With the work-around, the issue is much lessened but the chat will still occasionally jump around and more rarely still flicker. Here is the same situation with the changes from this PR.

https://user-images.githubusercontent.com/8285260/229935248-60930931-62b3-4b98-947c-4f4345f40f27.mp4

To expand on the solution a bit more, the main effect is that the value of a Scrollbar doesn't need to change to keep your position when you are scrolled up because maximum is always growing. But, we still only have a limited number of messages in view so minimum will also grow to keep that in check. If you leave the chat scrolled up, eventually minimum will grow over the Scrollbar value and the chat will continue scrolling because you are at the top. I don't know if overflow is realistically an issue but I accounted for it.

pajlada:
When starting Chatterino with this pull request, the chat is frozen & not loading until the user scrolls once
image

This is with having both `Smooth scrolling` & `Smooth scrolling on new messages` enabled

Pizzabelly:
The issue was updating maximum before offsetting the Scrollbar's value in messageAddedAtStart() was causing the Scrollbar to actually scroll by the amount of messages added. This would end up being less than the bottom because of the connected message and largeChange not being set, causing the chat to not layout correctly. Updating maximum after brings the behavior mostly inline with how it was, except that the Scrollbar's value won't actually be correct until the layout at the end of messageAddedAtStart() instead of the layout triggered by currentValueChanged after calling offset(), but this seems to not make a difference.

Pizzabelly:
I'm pretty sure smoothScrollingOffset can be removed because it doesn't get used after the changes from this pr.

pajlada:
Old behaviour (before this PR): Smooth scrolling on new messages didn't work at all in splits that were full
New behaviour (with this PR): Smooth scrolling **mostly** works on new messages in splits that are full, but sometimes doesn't

With the most recent merge from the main branch, you can add test channels that add consistent messages
See https://github.com/Chatterino/chatterino2/pull/4655 description

Scrolling with this PR seems fine, but the inconsistency in new message smooth scroll is offputting, even though the old behaviour was wrong, at least it was wrong consistently

Pizzabelly:
Smooth scrolling on new messages was inconsistent because of the call to scrollToBottom() with no animation (regardless of setting) on messageAppended() if a message was removed. This also happened before the changes in this PR but it was much more infrequent/unlikely. The issue was that by the end of the layout called by messageAppended() the Scrollbar value for the "bottom" could be different, triggering another scrollToBottom() but with the animation. This could then be extended by the no animation scrollToBottom() in messageAppended(), if it happened quick enough, causing it to look like smooth scrolling was working for a second. I assume the scrollToBottom() in messageAppended() was there because with the old behavior, once the ChannelView was full, when a message was added the Scrollbar value to stay at the bottom would only be adjusted instead of actually increasing and it didn't look right to animate that. With the changes here, if we just remove that scrollToBottom() and let it be handled in the layout, smooth scrolling on new messages seems to work as expected even if ChannelView is full.

I've also changed the behavior of updateScrollbar() to not attempt to scrollToBottom() if causedByScrollbar is true. This guards against using the Scrollbar in an in-between state, like what could happen in messageAddedToStart() when offset() triggers a scrollToBottom() before offsetMaximum() is called. This would cause the new largeValue to be used before maximum was updated, resulting in scrollToBottom() requesting a negative value from the Scrollbar. It also stops scrollToBottom() from repeatedly being called for every new message because it gets triggered by currentValueChanged which will fire many times before the animation is finished. This means the animation would be incrementally stopped and started down it's path. This did, however, change the way the animation looked because it caused the duration and curve to be reset every time it was called. I think the new behavior is more correct, especially considering the old behavior was inconsistent and possibly inefficient, but the old behavior did seem "smoother". I think if that's a concern a better approach would be to dynamically adjust the duration of the animation.

pajlada:
Message flow in as expected with the "Smooth scrolling on new messages" with all chat velocities :+1:
Scrolling with my mouse wheel works great at all chat velocities :+1: 
Scrolling by pressing the mousewheel does not work with fast chat velocities, `$$$$$` and above. See https://clips.twitch.tv/RacyTangentialParrotCeilingCat-2YuTnEoT071NSI2x for my clip LUL showing it off
I have not tested laptop trackpad behaviour, but I imagine it should be fine as they emulate mouse wheel events

Pizzabelly:
Mousewheel press scrolling was broken in that way because it explicitly doesn't use the smooth scrolling animation. So, when starting the scroll if the chat was moving very fast like $$$$$ there was a very high chance that when calling setDesiredValue() with no animation it would set the end value of the currently running animation instead of setting currentValue. In this path, atBottom wasn't updated, so after trying to scroll up it would jump back to the bottom. I don't see a problem with updating atBottom in setDesiredValue() any time the value is changed, which fixes that issue.

pajlada (changes requested):
Awesome work, scrolling works as expected in every combination of settings I've tried!

One last nitpick: Could you document the new methods you've added? (I know you haven't had that fortune :grimacing:)

Once that's done I'm happy to get this merged in!

pajlada (approved):

PR #4502, opened 2023-04-01
Fix: check reply-parent-user-id for blocked user

### LINKED ISSUE #4481: Reply context doesn't respect the block list
### Checklist

### Describe your issue

When a chatter replies to someone you have blocked it will show the message that the blocked person said.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 7.4.2 (commit c140c88) on Windows 10

2547techno:
Pull request checklist:

# Description

Closes #4481 assuming option 2. from https://github.com/Chatterino/chatterino2/issues/4481#issuecomment-1493143358 .

Reply context message is censored if from blocked user


kornes:
we shouldn't hide messages from non blocked users, 
personally i would go with censoring of reply context and replace it with some generic _"Blocked message"_

2547techno:
> we shouldn't hide messages from non blocked users, personally i would go with censoring of reply context and replace it with some generic _"Blocked message"_

Make's sense, it now replaces the message content with \*blocked message\*.

Another option may be to treat the message as if the "show reply context" setting was disabled which probably follows closer to what blocking a user achieves normally. This made more sense after hearing: https://www.twitch.tv/videos/1782533448?t=01h10m21s

I'll leave the PR as draft for now.

kornes:
I think leaving reply context visible is better solution, because it let user know given message responds to blocked account and can be ignored right away. 
Question if username should be also removed from context, i would leave it but not sure if everyone would agree with that, so maybe axe it too?

2547techno:
image
Reply context looks like this now.

Any other thoughts are welcome. Marking ready for review :)

Felanbird:
I'm putting in my vote for `[Blocked Message]` instead of `*Blocked Message*`

2547techno:
image

Felanbird (approved):
Seems to work as expected, also tested replying to a block user just to make sure nothing crazy would happen, and that worked as expected as well.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Bugfix: Fixed blocked user list sticking around when switching from a logged in user to being logged out. (#4437)
 - Bugfix: Fixed search popup ignoring setting for message scrollback limit. (#4496)
 - Bugfix: Fixed a memory leak that occurred when loading message history. This was mostly noticeable with unstable internet connections where reconnections were frequent or long-running instances of Chatterino. (#4499)
+- Bugfix: Fixed reply context showing blocked user's name & message (#4502)
```
pajlada, line 20: Nitpicking, but this is not a bugfix. There's no expectation set for replies to a blocked user to be hidden or censored. Make this a Minor entry instead, and make sure there's a period at the end of the changelog entry (e.g. `- Minor: asd asd. (#4502)`

pajlada (changes requested):
Only a nitpick on the changelog entry, implementation looks good to me :+1:

Felanbird (approved):
LGTM - only issue I would have would be making sure our native `show-blocked-messages-if-you're-a-mod` works, but I have a vague memory of testing that myself.

Felanbird:
@2547techno Thanks for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

To do so, open a new PR where you modify the contributors.txt file and add yourself to the list. Make sure to read the comments at a top for instructions.

PR #4512, opened 2023-04-04
Use New 7TV Cosmetics System

Nerixyz:
Pull request checklist:

# Description

This PR implements the new cosmetics system for 7TV, where cosmetics are updated using the EventAPI. This is essentially the same code that's already in [chatterino7](https://github.com/SevenTV/chatterino7) ([relevant PR](https://github.com/SevenTV/chatterino7/pull/184)). Adding and updating a badge for a user only requires taking one lock.

Once this is merged, I'm planning to add support for the similar [BTTV system](https://github.com/Chatterino/chatterino2/discussions/4323) as well.


Nerixyz:
```
--- /dev/fd/63	2023-04-04 14:15:21.963666976 +0000
+++ /dev/fd/62	2023-04-04 14:15:21.963666976 +0000
@@ -1264,8 +1264,8 @@
     this->threads_[thread->rootId()] = thread;
 }
 
-const std::unordered_map<QString, std::weak_ptr<MessageThread>> &
-    TwitchChannel::threads() const
+const std::unordered_map<QString, std::weak_ptr<MessageThread>>
+    &TwitchChannel::threads() const
 {
     return this->threads_;
 }
src/providers/twitch/TwitchChannel.cpp differs!!!!!!!
At least one file is poorly formatted - check the output above
```

`clang-format` should be updated :)

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -1594,4 +1603,70 @@ boost::optional<CheerEmote> TwitchChannel::cheerEmote(const QString &string)
...
+        .header("Content-Type", "application/json")
+        .payload(QJsonDocument(payload).toJson(QJsonDocument::Compact))
+        .onSuccess([chan = weakOf<Channel>(this)](const auto &response) {
+            const auto self =
+                std::dynamic_pointer_cast<TwitchChannel>(chan.lock());
+            if (!self)
+            {
+                return Success;
+            }
+            const auto json = response.parseJson();
```
pajlada, line 1654: Remove unused response parsing

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -1594,4 +1603,70 @@ boost::optional<CheerEmote> TwitchChannel::cheerEmote(const QString &string)
...
+            self->nextSeventvActivity_ =
+                QDateTime::currentDateTimeUtc().addSecs(10);
+            return Success;
+        })
+        .concurrent()
+        .execute();
```
pajlada, line 1660: Check thread safety

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -444,4 +445,35 @@ void TwitchAccount::autoModDeny(const QString msgID, ChannelPtr channel)
...
+            const auto json = response.parseJson();
+            const auto id = json["user"]["id"].toString();
+            if (!id.isEmpty())
+            {
+                this->seventvUserID_ = id;
+            }
+            return Success;
+        })
+        .concurrent()
+        .execute();
```
pajlada, line 476: Validate thread safety

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -444,4 +445,35 @@ void TwitchAccount::autoModDeny(const QString msgID, ChannelPtr channel)
...
+{
+    if (!this->seventvUserID_.isEmpty() || this->isAnon())
+    {
+        return;
+    }
+
+    static const QString seventvUserInfoUrl =
+        QStringLiteral("https://7tv.io/v3/users/twitch/%1");
+
+    NetworkRequest(seventvUserInfoUrl.arg(this->getUserId()),
```
pajlada, line 463: Should probably be part of a seventvapi file since we already implement this API in SeventvEmotes.cpp

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -444,4 +445,35 @@ void TwitchAccount::autoModDeny(const QString msgID, ChannelPtr channel)
...
+{
+    return this->seventvUserID_;
+}
+
+void TwitchAccount::loadSeventvUserID()
+{
+    if (!this->seventvUserID_.isEmpty() || this->isAnon())
+    {
+        return;
+    }
```
pajlada, line 458: Split this up into two separate if-statements

src/providers/seventv/eventapi/Dispatch.cpp
```diff
@@ -91,4 +93,46 @@ bool UserConnectionUpdateDispatch::validate() const
...
+bool CosmeticCreateDispatch::validate() const
+{
+    return !this->data.empty() && this->kind != CosmeticKind::INVALID;
+}
+
+EntitlementCreateDeleteDispatch::EntitlementCreateDeleteDispatch(
+    const Dispatch &dispatch)
+{
+    const auto obj = dispatch.body["object"].toObject();
+    this->userID = QString();
```
pajlada, line 113: why do we need this?

src/providers/seventv/SeventvEventAPI.cpp
```diff
@@ -53,6 +70,19 @@ void SeventvEventAPI::unsubscribeUser(const QString &id)
...
+{
+    if (this->subscribedTwitchChannels_.erase(id) > 0)
+    {
+        this->unsubscribe(
+            {ChannelCondition{id}, SubscriptionType::CreateCosmetic});
+        this->unsubscribe(
+            {ChannelCondition{id}, SubscriptionType::CreateEntitlement});
+        this->unsubscribe(
+            {ChannelCondition{id}, SubscriptionType::DeleteEntitlement});
+    }
```
pajlada, line 83: cosmetic nitpick: add a `,` after the last parameter in the unsubscribe call so it takes more lines and looks nicer

src/providers/seventv/SeventvEventAPI.cpp
```diff
@@ -35,6 +39,19 @@ void SeventvEventAPI::subscribeUser(const QString &userID,
...
+void SeventvEventAPI::subscribeTwitchChannel(const QString &id)
+{
+    if (this->subscribedTwitchChannels_.insert(id).second)
+    {
+        this->subscribe(
+            {ChannelCondition{id}, SubscriptionType::CreateCosmetic});
+        this->subscribe(
+            {ChannelCondition{id}, SubscriptionType::CreateEntitlement});
+        this->subscribe(
+            {ChannelCondition{id}, SubscriptionType::DeleteEntitlement});
```
pajlada, line 51: cosmetic nitpick: add a `,` after the last parameter in the unsubscribe call so it takes more lines and looks nicer

src/providers/seventv/SeventvEventAPI.cpp
```diff
@@ -261,4 +331,53 @@ void SeventvEventAPI::onUserUpdate(const Dispatch &dispatch)
     }
 }
 
+// NOLINTBEGIN(readability-convert-member-functions-to-static)
+
+// We're using `Application::instance`, because we're not in the GUI thread.
+// `seventvBadges` does its own locking.
```
pajlada, line 337: Move this comment to inside each function where Application::instance is used

Maybe something like

```c++
// We're using `Application::instance` instead of getApp(), because we're not in the GUI thread.
// `seventvBadges` does its own locking.
auto *seventvBadges = Application::instance->seventvBadges;
...
```

include `getApp()` in the comment to make it easier to find

pajlada (changes requested):

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
     using chatterino::seventv::eventapi::SubscriptionType;
     switch (value)
     {
+        case SubscriptionType::AnyEmoteSet:
+            return "emote_set.*";
+        case SubscriptionType::CreateEmoteSet:
+            return "emote_set.create";
```
pajlada, line 101: Where are these used?

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
...
         case SubscriptionType::UpdateEmoteSet:
             return "emote_set.update";
         case SubscriptionType::UpdateUser:
             return "user.update";
+        case SubscriptionType::AnyCosmetic:
+            return "cosmetic.*";
```
pajlada, line 107: Where is this used?

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
...
+        case SubscriptionType::CreateCosmetic:
+            return "cosmetic.create";
+        case SubscriptionType::UpdateCosmetic:
+            return "cosmetic.update";
```
pajlada, line 111: Where is this used?

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
...
+        case SubscriptionType::DeleteCosmetic:
+            return "cosmetic.delete";
```
pajlada, line 113: Where is this used?

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
...
+        case SubscriptionType::AnyEntitlement:
+            return "entitlement.*";
```
pajlada, line 115: Where is this used?

src/providers/seventv/eventapi/Subscription.hpp
```diff
@@ -70,10 +95,30 @@ constexpr magic_enum::customize::customize_t magic_enum::customize::enum_name<
...
+        case SubscriptionType::CreateEntitlement:
+            return "entitlement.create";
+        case SubscriptionType::UpdateEntitlement:
+            return "entitlement.update";
```
pajlada, line 119: Where is this used?

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -226,6 +226,8 @@ class TwitchChannel : public Channel, public ChannelChatters
     void showLoginMessage();
     /** Joins (subscribes to) a Twitch channel for updates on BTTV. */
     void joinBttvChannel() const;
+    void updateSevenTVActivity();
```
pajlada, line 229: A comment describing what this is for/what it's used for/when it's used would be nice

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -1594,4 +1604,60 @@ boost::optional<CheerEmote> TwitchChannel::cheerEmote(const QString &string)
...
+        this->roomId(), currentSeventvUserID,
+        [chan = weakOf<Channel>(this)]() {
+            const auto self =
+                std::dynamic_pointer_cast<TwitchChannel>(chan.lock());
+            if (!self)
+            {
+                return Success;
+            }
+            self->nextSeventvActivity_ =
+                QDateTime::currentDateTimeUtc().addSecs(10);
```
pajlada, line 1645: I'd prefer this interval to be increased to 60s or something like that - I assume the only downside to this is that it would take at most 50 seconds longer for your new badge to show up for yourself and others

src/singletons/Settings.hpp
```diff
@@ -268,6 +268,7 @@ class Settings : public ABSettings, public ConcurrentSettings
     BoolSetting enableSevenTVGlobalEmotes = {"/emotes/seventv/global", true};
     BoolSetting enableSevenTVChannelEmotes = {"/emotes/seventv/channel", true};
     BoolSetting enableSevenTVEventAPI = {"/emotes/seventv/eventapi", true};
+    BoolSetting sendSevenTVActivity = {"/emotes/seventv/sendActivity", true};
```
pajlada, line 271: The negative impact of disabling this by default is that users will *never* see your 7TV badge?
Or does it mean that the badge is just more seldomly updated

pajlada (changes requested):

src/providers/seventv/SeventvBadges.cpp
```diff
@@ -17,61 +18,102 @@ void SeventvBadges::initialize(Settings & /*settings*/, Paths & /*paths*/)
...
-    urlQuery.addQueryItem("user_identifier", "twitch_id");
+    const auto it = this->badgeMap_.find(userID.string);
+    if (it != this->badgeMap_.end() && it->second->id.string == badgeID)
+    {
+        this->badgeMap_.erase(userID.string);
+    }
+}
 
-    url.setQuery(urlQuery);
+void SeventvBadges::addBadge(const QJsonObject &badgeJson)
```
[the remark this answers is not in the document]
Nerixyz: No, the JSON schema is slightly different (mainly the images are different).

src/providers/seventv/eventapi/Subscription.hpp
Nerixyz: These aren't used, they're only here for completeness. I can remove them if desired.

src/singletons/Settings.hpp
Nerixyz: > users will _never_ see your 7TV badge

Yes, if the EventAPI is used exclusively.

> Or does it mean that the badge is just more seldomly updated

This is only the case with the v2 endpoint (only personal emotes won't be updated there). That's why I'd like to keep it as a backup.

AnatoleAM: note: V2 endpoint sunsets on November 1st, from then on it will return a `410 Gone` with the standard JSON error response, you should check this doesn't pose any issues with the current implementation. Otherwise, it's safe to keep while waiting for users to upgrade their clients.

src/providers/seventv/eventapi/Subscription.hpp
pajlada: Make it look like it does in the fork then for as few as possible merge conflicts

src/singletons/Settings.hpp
Nerixyz: Currently, all errors (non 2xx status, network, ...) are ignored when loading badges, so that should be fine.

pajlada: So am I correct in assuming that if I as a user never send my activity, others who exclusively use eventapi for cosmetics will never see any of my 7TV badges?

pajlada: > note: V2 endpoint sunsets on November 1st, from then on it will return a `410 Gone` with the standard JSON error response, you should check this doesn't pose any issues with the current implementation. Otherwise, it's safe to keep while waiting for users to upgrade their clients.

Could you provide an example of what that payload would look like?

AnatoleAM: > others who exclusively use eventapi for cosmetics will never see any of my 7TV badges

This is correct, yes.

> Could you provide an example of what that payload would look like?

Looks like this, error code may be different, there's no error code for a removed endpoint yet
```json
{
  "status_code": 410,
  "status": "Gone",
  "error": "Unknown Route",
  "error_code": 70498,
  "details": { "message": "The API endpoint requested does not exist" }
}
```

src/providers/seventv/SeventvBadges.cpp
```diff
@@ -17,61 +18,102 @@ void SeventvBadges::initialize(Settings & /*settings*/, Paths & /*paths*/)
...
+    if (emote.images.getImage1()->isEmpty())
+    {
+        return;  // Bad images
+    }
+
+    this->knownBadges_[badgeID] =
+        std::make_shared<const Emote>(std::move(emote));
+}
+
+void SeventvBadges::loadSeventvBadges()
```
pajlada, line 85: Given that the endpoint is being removed in November, I would rather just rip it out here and be done with it

pajlada (changes requested):

tests/src/main.cpp
```diff
@@ -38,7 +38,9 @@ int main(int argc, char **argv)
         chatterino::NetworkManager::deinit();
 
         settingsDir.remove();
-        QApplication::exit(res);
+        // This should be QApplication::exit(res);
+        // but using this will deadlock in ~QHostInfoLookupManager (if the twitch account was changed)
+        _exit(res);
```
Nerixyz, line 43: This is actually really sad, but I don't see a way to avoid this without refactoring network-request internals.

Nerixyz: Ok, I kinda lied. There is one way around it: calling `NetworkManager::workerThread.terminate()` after quitting the thread, which is probably a bit less ugly but still ugly.

Nerixyz: Hmm, the `terminate` doesn't seem to be that reliable. Delaying the loading of the 7TV user-id by 500ms seems to work. It's not 100% ideal, but fine since the ID is only needed once a message is sent.

benchmarks/src/main.cpp
```diff
@@ -22,7 +22,9 @@ int main(int argc, char **argv)
         ::benchmark::RunSpecifiedBenchmarks();
 
         settingsDir.remove();
-        QApplication::exit(0);
+        // This should be QApplication::exit(0);
+        // but using this will deadlock in ~QHostInfoLookupManager (if the twitch account was changed)
+        _exit(0);
```
pajlada, line 27: Would be nice to have a complete solution for this

pajlada (commented):
I'm happy to merge this in when the network issue thing is resolved

pajlada (approved):
Thank you for putting in the effort!

PR #4518, opened 2023-04-07
Fix build on latest Fedora

GUIpsp:
Note: I am not a c++ developer at all. I don't know if these changes are sound. But they let me build the project, and it seems to work.

# Description

Fixes issues of this form: 
```
/home/guipsp/local/chatterino2/src/widgets/splits/SplitHeader.cpp:531:46: error: converting to ‘const std::optional<std::vector<QString> >’ from initializer list would use explicit constructor ‘constexpr std::optional<_Tp>::optional(std::in_place_t, _Args&& ...) [with _Args = {}; typename std::enable_if<__and_v<std::is_constructible<_Tp, _Args ...> >, bool>::type <anonymous> = false; _Tp = std::vector<QString>]’
  531 |             notifySeq = h->getDisplaySequence(HotkeyCategory::Split,
      |                         ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
  532 |                                               "setChannelNotification", {{}});
```

### DIFF
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -391,7 +391,7 @@ void Notebook::setShowTabs(bool value)
     if (!value && getSettings()->informOnTabVisibilityToggle.getValue())
     {
         auto unhideSeq = getApp()->hotkeys->getDisplaySequence(
-            HotkeyCategory::Window, "setTabVisibility", {{}});
+            HotkeyCategory::Window, "setTabVisibility", {});
         if (unhideSeq.isEmpty())
         {
             unhideSeq = getApp()->hotkeys->getDisplaySequence(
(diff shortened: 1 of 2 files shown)


Mm2PL:
This changes behavior of menu key combo displays which have toggle options attached to them (for example hiding tabs).

The code snippet below looks up a hotkey like:

- Category: Window
- Action: `setTabVisibility`
- Arguments: *[empty]*

https://github.com/Chatterino/chatterino2/blob/master/src/widgets/Notebook.cpp#L393-L394

While this next one searches for one like:

https://github.com/GUIpsp/chatterino2/blob/master/src/widgets/Notebook.cpp#L393-L394

- Category: Window
- Action: `setTabVisibility`
- Arguments: *[does not matter]*

The intended meaning of those lines is the first one. @pajlada said he'd look into swapping the `{{}}` syntax for something else.

pajlada:
Could you verify that this builds as expected for you?

GUIpsp:
This builds fine, thanks!

For reverence, this is my `gcc` version:
```
❯ gcc --version
gcc (GCC) 13.0.1 20230401 (Red Hat 13.0.1-0)

```

pajlada (approved):

pajlada:
Thanks for your contribution @GUIpsp! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the application.

PR #4519, opened 2023-04-07
Add Fuzzy Matching to emote input completion

kklixd:
Pull request checklist:

# Description

This change adds Fuzzy Matching to the emote popup when colon (:) is typed in chat.

Example:

Before:
Before

After:
After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Minor: Added support for FrankerFaceZ animated emotes. (#4434)
 - Minor: Added a local backup of the Twitch Badges API in case the request fails. (#4463)
 - Minor: Added the ability to reply to a message by `Shift + Right Click`ing the username. (#4424)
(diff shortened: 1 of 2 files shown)


Felanbird:
We have a long discussion about bettering the colon emote menu in #4414 - but I feel like fuzzy search is the exact opposite of what people want.

LosFarmosCTL:
> but I feel like fuzzy search is the exact opposite of what people want.

I think fuzzy matching would be fine as long as it's lower priority than any of the other discussed options.

So an exact match has to always beat fuzzy search, prioritized by the criteria outlined in that discussion and only be at the top when there is no good match. But I think that in general it could be beneficial i.e. if you mistype something.

kklixd:
Sorry for the late reply, but hopefully I have fixed it.

pajlada:
With #4639 merged in, this PR will look pretty different. Please open a new PR if you are still interested in implementing fuzzy matching.

PR #4520, opened 2023-04-08
Remove unused variables, Refactor in MessageLayout

dnsge:
Pull request checklist:

No changelog entry needed in my opinion

# Description

* Moves `layoutCount_` and `bufferUpdatedCount_` debug counters to an `ifdef` block in `MessageLayout`
* Removes unused variable `MessageLayout::collapsedHeight_`
* Replaces `std::shared_ptr<MessageLayoutContainer> container_` with `MessageLayoutContainer container_` in `MessageLayout`
	- The container never escapes nor can be null so it made no sense to be in a `shared_ptr`
* Replaces `std::shared_ptr<QPixmap> buffer_` with `std::unique_ptr<QPixmap> buffer_` in `MessageLayout`
	- The buffer never escapes so I replaced it with a `unique_ptr` instead
* Extracts buffer creation logic to `MessageLayout::ensureBuffer`

Why? While scrolling on a debug build, my Chatterino crashed with the following stack trace:
```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0xff76443500e54f18 -> 0x0000443500e54f18 (possible pointer authentication failure)
Exception Codes:       0x0000000000000001, 0xff76443500e54f18
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [4395]

VM Region Info: 0x443500e54f18 is not in any region.  Bytes after previous region: 74513402646297  Bytes before following region: 30558677283048
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      commpage (reserved)        1000000000-7000000000   [384.0G] ---/--- SM=NUL  ...(unallocated)
--->  GAP OF 0x5f9000000000 BYTES
      MALLOC_NANO              600000000000-600008000000 [128.0M] rw-/rwx SM=PRV  

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   chatterino                    	       0x1012b0ba4 std::__1::shared_ptr<chatterino::MessageLayoutContainer>::operator->() const + 12 (memory:2872)
1   chatterino                    	       0x1012b0b88 chatterino::MessageLayout::getHeight() const + 28 (MessageLayout.cpp:70)
2   chatterino                    	       0x101b8f87c chatterino::ChannelView::wheelEvent(QWheelEvent*) + 376 (ChannelView.cpp:1376)
3   QtWidgets                     	       0x1096c4404 QWidget::event(QEvent*) + 128
4   QtWidgets                     	       0x10968e740 QApplicationPrivate::notify_helper(QObject*, QEvent*) + 292
5   QtWidgets                     	       0x109690de4 QApplication::notify(QObject*, QEvent*) + 5424
6   QtCore                        	       0x10aead6e4 QCoreApplication::notifyInternal2(QObject*, QEvent*) + 208
```

Ostensibly caused by the `MessageLayoutContainer` in the `MessageLayout` being null. I suppose this could be caused by a race condition in which the destructor was in progress and the layout container had been deleted, but not the layout? Not sure. Regardless, while investigating, I noticed that the `MessageLayoutContainer` didn't need to be a pointer at all. I also noticed the other few things that I included in this PR.

Feel free to say that this doesn't matter/shouldn't be changed. Though, I think extracting the debug counters to an `ifdef` and removing the unused variable should be a change made anyway.


src/messages/layouts/MessageLayout.hpp
```diff
@@ -69,27 +70,30 @@ class MessageLayout : boost::noncopyable
...
+
+    // Create new buffer if required, returning the buffer
+    QPixmap *ensureBuffer(QPainter &painter, int width);
+
     // variables
     MessagePtr message_;
-    std::shared_ptr<MessageLayoutContainer> container_;
-    std::shared_ptr<QPixmap> buffer_{};
+    MessageLayoutContainer container_;
+    std::unique_ptr<QPixmap> buffer_{};
```
Nerixyz, line 83: You might be able to do `QPixmap buffer_;` here and use [`QPixmap::isNull`](https://doc.qt.io/qt-6/qpixmap.html#isNull). This would remove another indirection, because a `QPixmap` is [essentially another shared-pointer](https://github.com/qt/qtbase/blob/7fac7b5b5011a3db9ba778e56f2b34021f6bd148/src/gui/image/qpixmap.h#L126) and constructing a null-pixmap [is cheap](https://github.com/qt/qtbase/blob/7fac7b5b5011a3db9ba778e56f2b34021f6bd148/src/gui/image/qpixmap.cpp#L64-L83). You then need to avoid using the `QPixmap`'s copy-constructor, since this [potentially copies the pixmap-data](https://github.com/qt/qtbase/blob/7fac7b5b5011a3db9ba778e56f2b34021f6bd148/src/gui/image/qpixmap.cpp#L183-L195).

pajlada (approved):
Good refactor, easy to follow :+1:

src/messages/layouts/MessageLayout.hpp
pajlada: follow-up pr, but can be explored if someone wants

PR #4525, opened 2023-04-08
Update BUILDING_ON_MAC.md

killertofus:
# Description
This PR adds instructions to build on m1 macs if you get Code Signature Invalid


Felanbird:
I don't think this is critical-enough information that we need to rush to implement it into the docs, especially when the person who resolved the issue with this command, stated they would update the docs to mention it.

pajlada:
Closing this for now until we have a more clear indication of what needs to be done as a developer self-building Chatterino on mac, and what versions this might be relevant for. Feel free to make a Discussion for this to figure these questions out

PR #4542, opened 2023-04-14
feat: add /lowtrust command

iProdigy:
# Description
Adds `/lowtrust` command to open `https://www.twitch.tv/popout/moderator/{channelLogin}/low-trust-users`


Mm2PL:
Why should this be a builtin command and can't just be an /openurl custom one?

iProdigy:
> Why should this be a builtin command and can't just be an /openurl custom one?

Fair point - I'd argue this is a helpful shortcut for many moderators that use chatterino (especially since we currently don't display messages from restricted users & there is no official api to unrestrict/unmonitor users)

leon-richardt:
I think it's worth having as a built-in.

pajlada (approved):

PR #4546, opened 2023-04-15
Add message for empty mod list.

2547techno:
# Description

Replace the system message when the mod list is empty to be consistent with the message shown for an empty VIP list ( `This channel does not have any VIPs.` )

Before: 
image

After: 
image


2547techno:
Wasn't sure if its a minor or a bugfix

Mm2PL:
I'd say it's a bugfix, but it can stay as a `Minor` entry. We're splitting hairs at this point.

pajlada (approved):

Felanbird (approved):

PR #4549, opened 2023-04-16
Restart the sound device if it's been stopped

pajlada:
# Description

Fixes a bug where we don't check if the sound device is active before trying to play a sound to it

We fix this by checking its state before every sound we try to play, and if the device is not in the started state we try to restart it.


pajlada:
image

PR #4551, opened 2023-04-16
Mark experimental builds at the start of their filename

pajlada:
# Description

also manually mark the builds as experimental instead of using a god of
war sed replace

The test for if this works can only be checked in the nightly release once this has been merged into the main branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -427,7 +427,10 @@ jobs:
 
       - name: Mark experimental
         run: |
-          for file in *; do mv -n "$file" "$(echo $file | sed 's/\(6\(\.[[:digit:]]\)\{2\}\)/\1-EXPERIMENTAL/g')"; done
+          # Mark all Qt6 builds as EXPERIMENTAL
+          mv chatterino-macos-Qt-6.5.0.dmg EXPERIMENTAL-chatterino-macos-Qt-6.5.0-dmg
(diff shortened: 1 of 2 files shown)


Mm2PL (approved):
Pepega :mega: EXPERIMENTAL

Felanbird (approved):

Mm2PL:
@pajlada I don't think it worked https://github.com/Chatterino/chatterino2/actions/runs/4715589021/jobs/8362890140

Felanbird:
For some reason the `Chatterino-ubuntu-22.04-Qt-6.2.4.deb` file is just not downloading. 
Throwing `ls` into the mix returns all the artifacts _except_ that one 
https://github.com/Felanbird/chatterino2/actions/runs/4716725597/jobs/8365024218

(was also trying setting the path manually, encase `working-directory` wasn't working)

Wissididom:
I assume it could also be that it simply overwrites the file. When I look into Felanbirds Workflow I see that those Ubuntu files do not contain anything that specifies the Qt version and since 22.04 is built for 5.15.2 and 6.2.4 the 5.15.2 version might just have been overwritten by the 6.2.4 version when downloading (when looking at the step order), because both are named `Chatterino-ubuntu-22.04-x86_64.deb`. So I see 3 options being either to rename the file inside the `CreateUbuntuDeb.sh` or somewhere prior to uploading it or to download it and after downloading the 5.15.2 to rename it and then downloading the 6.2.4 and rename that.

Their filenames currently are `Chatterino-ubuntu-20.04-x86_64.deb` and `Chatterino-ubuntu-22.04-x86_64.deb` according to Felanbirds Workflow.

PR #4570, opened 2023-04-21
Add the ability to select custom themes in the settings dialog

pajlada:
# Description

This loads themes (any .json file) from the <chatterino appdata>/Themes directory on startup and lists those in the settings page.
The custom themes in that directory will have a _Custom:_ prefix to show that they're custom.
In the settingsfile, we save the name of the theme (e.g. 'forsen', no more info).

When the theme is updated or on Chatterino startup, we find the best matching theme with that name, whether it's a built in theme or a custom theme.

There's currently no mechanism for making a theme that has the same name as a built in theme, and they still show up in the menu. IMO we should just not allow duplicate names for simplicity's sake.

There's a few things left to do, and a few things to discuss before taking this step, even though I still deem it to be "pre-production" even when this PR is merged in.

The changelog entry is left as a Dev entry on purpose since I don't think this is a feature that should be fully announced until there's more testing & better UX for users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unversioned
 
 - Dev: Added command to set Qt's logging filter/rules at runtime (`/c2-set-logging-rules`). (#4637)
+- Dev: Added the ability to see & load custom themes from the Themes directory. No stable promises are made of this feature, changes might be made that breaks custom themes without notice. (#4570)
 
 ## 2.4.4
 
--- a/src/singletons/Paths.cpp
+++ b/src/singletons/Paths.cpp
@@ -142,6 +142,7 @@ void Paths::initSubDirectories()
     this->miscDirectory = makePath("Misc");
     this->twitchProfileAvatars = makePath("ProfileAvatars");
     this->pluginsDirectory = makePath("Plugins");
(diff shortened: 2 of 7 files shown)


dnsge:
It'd be great if there was hot reloading of the current theme when the JSON file is modified. This would make creating themes and trying out different colors easy (compared to what I imagine is having to restart Chatterino after every change). 

There appears to be a Qt class that would make detecting the file change easy (in theory, barring os-specific annoyances): https://doc.qt.io/qt-6/qfilesystemwatcher.html. I haven't dug through the Theme code that much though, so it's possible that this is more difficult than it sounds.

pajlada:
> It'd be great if there was hot reloading of the current theme when the JSON file is modified. This would make creating themes and trying out different colors easy (compared to what I imagine is having to restart Chatterino after every change).

I would like that type of functionality in a V2/V3 kind of thing instead of in the initial release.
Initially I'd like this in a nightly release without any commitments to how/where the themes are located so it can be tested, then we look into specifics for how themes are worked on by users.

I could see a world where theme creation isn't tested in Chatterino itself, but rather a web UI. Or there's a `/reload-theme` command to do the reloading.

pajlada:
image

PR #4576, opened 2023-04-23
Refactor `LinkParser`

Nerixyz:
# Description

This PR is the continuation of #4507 and #4436.

These are the main changes:

- Use custom parsing for IPv4 addresses over a regex.
- Drop IPv6 support (was disabled by default).
- Use `QStringView` over `QStringRef`
- Removed usage of `goto`.

In my tests (`chatterino-benchmark --benchmark_filter=LinkParsing --benchmark_repetitions=100`), this version is about 6x faster:

### Before

| Benchmark | Time| CPU | Iterations|
|--------|--------|--------|--------|
|BM_LinkParsing    |         33159 ns    |    33692 ns   |     21333 |
|... |
|BM_LinkParsing_mean    |    33569 ns     |   33523 ns         | 100 |
|BM_LinkParsing_median  |    33178 ns     |   32959 ns         | 100 |
|BM_LinkParsing_stddev    |   1520 ns     |    1457 ns         | 100 |
|BM_LinkParsing_cv       |    4.53 %      |    4.35 %          | 100 |

flamegraph

### After
| Benchmark | Time| CPU | Iterations|
|--------|--------|--------|--------|
|BM_LinkParsing            |  5137 ns      |   5022 ns     |  112000|
|...|
|BM_LinkParsing_mean       |  5369 ns      |   5352 ns     |     100|
|BM_LinkParsing_median     |  5324 ns      |   5301 ns     |     100|
|BM_LinkParsing_stddev     |   273 ns      |    254 ns     |     100|
|BM_LinkParsing_cv         |  5.08 %       |   4.75 %      |     100|

flamegraph

As you can see, I didn't change the `isValidTld` it's now the bottleneck here. The main reason is that a `QString` needs to be created and case-converted. Ideally, it could use a `QStringView` and have case-insensitive search functions.


src/common/LinkParser.cpp
```diff
@@ -1,206 +1,231 @@
...
+    // We don't care about the actual value,
+    // we only want to verify the ip.
 
-    bool isValidIpv4(QStringRef &host)
+    char16_t sectionValue = 0;  // 0..256
+    uint8_t octetNumber = 0;    // 0..4
+    uint8_t sectionDigits = 0;  // 0..3
+    bool lastWasDot = true;
+
+    for (auto i : host)
```
pajlada, line 59: nit: naming, `i` makes me thing this is an index or iterator, maybe just `c`?

src/common/LinkParser.cpp
```diff
@@ -1,206 +1,231 @@
...
+            lastWasDot = true;
+            octetNumber++;
+            sectionValue = 0;
+            sectionDigits = 0;
+            continue;
+        }
+        lastWasDot = false;
 
-        return exp.match(host).hasMatch();
+        if (current > u'9' || current < u'0')
```
pajlada, line 76: [QChar::isDigit](https://doc.qt.io/qt-6/qchar.html#isDigit) implementation seems to handle unicode stuff, your solution seems good to me

src/common/LinkParser.cpp
```diff
@@ -1,206 +1,231 @@
...
+        }
+
+        sectionValue = sectionValue * 10 + (current - u'0');
+        sectionDigits++;
+        if (sectionValue >= 256 || sectionDigits > 3)
+        {
+            return false;
+        }
     }
+    return octetNumber == 3 && !lastWasDot;
```
pajlada, line 88: nit: let it breathe
```suggestion
    }

    return octetNumber == 3 && !lastWasDot;
```

pajlada (approved):

pajlada:
Code looks good, some small nitpicks that you can change if you feel like it's reasonable

src/common/LinkParser.cpp
Nerixyz: Right, `isDigit` also checks if the character is the "Nd" unicode category.

pajlada (approved):

PR #4578, opened 2023-04-24
Add tools to help debug image GC

Mm2PL:
# Description

Add:
- `/force-image-gc` which calls the image gc code normally, just as if its timer were to fire,
- `/force-image-unload` which clears all the images tracked by the gc,
- `image bytes`, `image bytes (ever {un,}loaded)` counters which try to visualize memory usage,
- `images tracked by gc` counter which shows how many total images GC has (including empty ones),
- `last image gc eligible` counter which shows how many images were eligible for GC,
- `last image gc expired` counter which shows how many images were evicted from memory.


src/controllers/commands/CommandController.cpp
```diff
@@ -3207,6 +3209,28 @@ void CommandController::initialize(Settings &, Paths &paths)
 
         return "";
     });
+
+    this->registerCommand(
+        "/force-image-gc",
+        [](const QStringList & /*words*/, auto /*channel*/) -> QString {
+            postToThread([]() {
```
Nerixyz, line 3216: It is extremely unlikely that this runs outside the gui-thread (e.g. in `execCommand` there's a call to `getApp`), so this should use `runInGuiThread`.

src/controllers/commands/CommandController.cpp
```diff
@@ -3207,6 +3209,28 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                auto &iep = ImageExpirationPool::instance();
+                iep.freeOld();
+            });
+            return "";
+        });
+
+    this->registerCommand(
+        "/force-image-unload",
+        [](const QStringList & /*words*/, auto /*channel*/) -> QString {
+            postToThread([]() {
```
Nerixyz, line 3227: Same here.

src/util/DebugCount.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "common/UniqueAccess.hpp"
+#include "util/Helpers.hpp"
```
pajlada, line 4: shouldn't need this include

pajlada (commented):
looks sane to me

dnsge (approved):

pajlada (approved):

PR #4582, opened 2023-04-29
Allow for local macOS signed builds

pajlada:
# Description

Split up the CreateDMG.sh macos script to two scripts:
MacDeploy.sh - this calls macdeployqt on the built app
CreateDMG.sh - this calls dmgbuild on the built & deployed app

Add a `SKIP_VENV` environment variable to CreateDMG.sh, this can be used to use the system version of dmgbuild
Add the ability to codesign the created dmg and its contents using the `MACOS_CODESIGN_CERTIFICATE` environment variable
Moved the output name logic from CreateDMG to the `OUTPUT_DMG_PATH` environment variable

The nightly release create job also doesn't remove artifacts now, it only replaces artifacts that are conflicting.
The downside to this is that if we change the name of an artifact, we need to manually delete the old artifact.
The upside to this is that we can now upload artifacts that are not handled in the same CI job.

### DIFF
--- a/.CI/CreateDMG.sh
+++ b/.CI/CreateDMG.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -eo pipefail
 
 if [ -d bin/chatterino.app ] && [ ! -d chatterino.app ]; then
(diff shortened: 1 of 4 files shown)


pajlada:
looks good to me
image

PR #4595, opened 2023-05-02
Fix the menu position on macOS when using Qt 6.5

pajlada:
# Description

Reported in Discord by Trummler - see video: https://cdn.discordapp.com/attachments/459650160658481152/1103008038576721921/Bildschirmaufnahme_2023-05-02_um_19.19.36.mov

This happened because we called popup then move on the menu, and the position was only correct when the move call was called because the size had not yet been evaluated in the menu.
Instead, when figuring out the position, we now use the menu's sizehint which should be correct even before the menu is ever shown.

### DIFF
--- a/src/widgets/helper/Button.cpp
+++ b/src/widgets/helper/Button.cpp
@@ -348,27 +348,24 @@ void Button::showMenu()
     if (!this->menu_)
         return;
 
-    auto point = [this] {
-        auto point = this->mapToGlobal(
-            QPoint(this->width() - this->menu_->width(), this->height()));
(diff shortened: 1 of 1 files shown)


pajlada:
Confirmed working as expected on macOS Qt 6.5, macOS Qt 5.15, Arch Linux Qt 6.2.4, Arch Linux Qt 5.15

Felanbird (approved):
Seems fine on 🅱️ indows, which the original issue did not effect - tested both 6.5 and 5.15

PR #4607, opened 2023-05-07
Add a Send button that can be used to send messages

### LINKED ISSUE #2279: Adding A "Send Message" Button
**What should be added?**
Please Add a "Send Message" Button like in the browser version of Twitch chat 

**Why should it be added?**
Tbh Im old and have multiple uses for this feature, since i grew up with it. 

But to completly honest, its mostly me being to lazy to press enter or hanging out on the couch only with the mouse.

pajlada:
# Description

This button is off by default, and can be enabled with the "Show send message button" setting

Fixes #2279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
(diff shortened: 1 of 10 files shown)


pajlada (commented):
LGTM :+1: 

image

PR #2819, opened 2021-05-23
Bump nightly appimage package after successful arch linux build

pajlada:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

PR #4610, opened 2023-05-08
Remove `push-aur` job

pajlada:
# Description

This never built the PR job, and doesn't handle pushing to AUR

The new plan is to have a separate AUR-specific repo that handles events
& rebuilding on dependency-change etc

Closes #2819


pajlada:
image
Great job, looks good to me!

PR #4616, opened 2023-05-09
Add 'joined channel' system message.

Mm2PL:
# Description

This is a very simple implementation of something like #4615. This doesn't check if it's a reconnect, just shows the message. There is no setting either, as I don't see why we should add one.


Felanbird (approved):
image

pajlada (approved):

PR #4618, opened 2023-05-10
Changed all window titles to use the full app version string

### LINKED ISSUE #4458: Changelog Window Title is Hard-Coded
### Checklist

### Describe your issue

The title of the changelog-window is hard-coded to `Chatterino 2`:
https://github.com/Chatterino/chatterino2/blob/0acbc0d2c346fe190e451c8ae6f81d4a395245f1/src/Application.cpp#L107

This might confuse users of Chatterino forks like Chatterino 7 or Dankerino.

[`Version::fullVersion()`](https://github.com/Chatterino/chatterino2/blob/0acbc0d2c346fe190e451c8ae6f81d4a395245f1/src/common/Version.hpp#L55) should be used to set the title.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 7.4.2 DEBUG (commit b84fde0e modified) built with Qt 5.15.2, Windows SDK, MSVC 193532215, Crashpad

Mm2PL:
# Description

Replace `"Chatterino"` with `Version::instance().fullVersion()` in various window titles.
Fixes #4458.


src/singletons/Updates.cpp
```diff
@@ -81,17 +81,19 @@ void Updates::installUpdates()
         assert(false);
         return;
     }
+    auto windowName =
+        QString("%1 - Update").arg(Version::instance().fullVersion());
```
Felanbird, line 85: is this a good chance to call it `Updater`

Mm2PL: I didn't change it because that's how it was called. You are free to apply a change to this (or another) branch if you feel it's necessary Okayeg

8thony:
4 other positions that might need to be changed
https://github.com/Chatterino/chatterino2/blob/168f346c81e0760ea2d5049ae154839073c8565b/src/common/Args.cpp#L22

https://github.com/Chatterino/chatterino2/blob/168f346c81e0760ea2d5049ae154839073c8565b/src/widgets/dialogs/ColorPickerDialog.cpp#L284

https://github.com/Chatterino/chatterino2/blob/168f346c81e0760ea2d5049ae154839073c8565b/src/widgets/Notebook.cpp#L500

https://github.com/Chatterino/chatterino2/blob/168f346c81e0760ea2d5049ae154839073c8565b/src/providers/liveupdates/BasicPubSubManager.hpp#L90-L94

Mm2PL:
I'm leaving the Args one and http User-Agent ones.

8thony:
In-app these changes look very bad (window titles are too long and sometimes truncated). 

My suggestion is to completely remove the version from the title as it is already present in the <code>add new account</code> window or in the <code>Searching in channel</code> window. The settings window itself looks fine.

Example: Instead of <code>Chatterino Nightly 2.4.4 - streamlink not found</code> just use <code>streamlink not found</code> as title.

What is your opinion?

Nerixyz:
> Example: Instead of `Chatterino Nightly 2.4.4 - streamlink not found` just use `streamlink not found` as title.

Just noticed that a lot of software puts the product/version at the end of the title, so this should be `Streamlink Not Found - Chatterino Nightly 2.4.4`. Also, the titles should be in title-case.

Mm2PL:
I will make it the other way around to make truncation lose less important information

8thony:
What's the point of cluttering every window title with the full version? 
Simple short titles are much better, in my opinion, like <code>add new account</code>, that's what the user interests.
Besides, most windows are not big enough to display the full title anyway.

PR #4424, opened 2023-03-04
Add Shift+Right Click Shortcut to Reply to a Message

dnsge:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

Shift+Right Clicking the `username` part of a message will start a reply to that message. This is one click faster than having to right-click and select "Reply to Message". 

Addresses #3934

Video: 

https://user-images.githubusercontent.com/24928223/222857256-18ff586f-caa5-4120-9242-870f33614067.mov

[felanbird]: Lets see if you can close discussions from PR bodies now that it's possible to close them
Closes #3934
Closes #4183

PR #4622, opened 2023-05-10
Allow for customizing the behavior of `Right Click`ing of usernames.

Mm2PL:
# Description

A follow up to https://github.com/Chatterino/chatterino2/pull/4424 and https://github.com/Chatterino/chatterino2/discussions/4183


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -297,6 +297,27 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         false,
         "Specify how Chatterino will handle messages that exceed Twitch "
         "message limits");
+    layout.addDropdown<std::underlying_type<UsernameRightClickBehavior>::type>(
```
pajlada, line 300: The settings are confusing to me since they don't really say what right-click behaviour actually does (specifically the Reply with modifier & Mention with modifier). They only say what action is taken with the modifier.
It might be better to split this into two settings:
Username right-click behaviour: ... reply/mention
Username modifier right-click behaviour: ... reply/mention

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -297,6 +297,27 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        [](auto index) {
+            return index;
+        },
+        [](auto args) {
+            return static_cast<UsernameRightClickBehavior>(args.index);
+        },
+        false,
+        "Specify how Chatterino will handle right-clicking a username in "
+        "chat.");
+    addKeyboardModifierSetting(layout,
```
pajlada, line 318: This adds `None` as a possible modifier, I don't think it makes sense to have None as an option here.
If anything, the "Right-click behaviour" and "Modifier right-click behaviour" settings should have "None" as options instead

pajlada (changes requested):

src/widgets/settingspages/GeneralPage.cpp
Mm2PL: I've split this into two settings as you said. I also added an option to do nothing in both cases.

src/widgets/settingspages/GeneralPage.cpp
Mm2PL: I've removed the `None` option for the modifier in the GUI, it's still possible to write an invalid config manually. If modifier is `None` and a username is right-clicked, Chatterino will now log a critical message saying that the config is invalid and bail.

pajlada (approved):

PR #4626, opened 2023-05-11
Reduce the size of the update prompt

### LINKED ISSUE #2123: Update prompt displays outside the current safe zone
**Describe the bug**
Can't click the install button when chatterino is full screen

**To reproduce**
1. Fullscreen chatterino
2. Click install button
3. Notice that it appears off the current monitor

**Screenshots**
image

**Chatterino version**

**Operating system**
Windows 10

Felanbird:
# Description

I just randomly cut the width in half and it worked perfectly

I had to resize chatterino to the absolute minimum possible and put it on the bottom of my screen to be once again unable to interact with install/dismiss buttons while using this resolution.

Default sized window:
image

Full-screen: 
image

Minimum possible width:
image

Fixes #2123


Felanbird:
It dawned on me that the `your version of chatterino is newer than the current stable` prompt is a little long, forcing some new lines into the text does re-fit it to this small window but it does look a little cluttered, but it doesn't seem _that_ bad
image

pajlada (approved):

PR #4628, opened 2023-05-12
Improve editing of hotkeys 

Mm2PL:
# Description

The UI will now show a simple picker on what a hotkey can do when there is such an option available.
For example `HotkeyCategory::Split -> "focus"` action has only 4 directions available, so it's possible to add this. However `HotkeyCategory::Window -> "openTab"` can have a number, so this approach does not extend to it.

Here is how the new UI looks for actions that support it:
screenshot of Edit Hotkey dialog with focus neighboring split action and a new Direction picker
For actions that don't support it, nothing will change.


src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -21,6 +38,19 @@ struct ActionDefinition {
 
     // maxCountArguments is the maximum amount of arguments the action accepts
     uint8_t maxCountArguments = minCountArguments;
+
+    // possibleArguments is empty or contains all possible argument values,
+    // left string from the pair is the thing the action will see,
+    // right one is what the user sees
+    std::vector<std::pair<std::vector<QString>, QString>> possibleArguments =
+        {};
```
pajlada, line 46: As discussed in chat, would be clearer if this was an `std::map<QString (what the user sees), std::vector<QString> (what arguments will be sent to the hotkey)>`

maybe vector of pairs, but the order needs to be the order they were added in at least

pajlada (changes requested):

src/controllers/hotkeys/ActionNames.hpp
Mm2PL: I swapped the order of the pairs in [1400424](https://github.com/Chatterino/chatterino2/pull/4628/commits/1400424042cdc1ee2e8be6aa5fbaff5df6af13af)

pajlada (approved):

Issue #4632, opened 2023-05-14
Remove QObjectRef in favor of QPointer


fourtf:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

We have a class called QObjectRef which is a pointer to a QObject which is reset when the referenced object is destroyed. It looks, however, like Qt itself has an equivalent class called QPointer. https://doc.qt.io/qt-6/qpointer.html

We can replace the QObjectRef usages and remove the class.

### OS and Chatterino Version

Chatterino 2.4.4

PR #4633, opened 2023-05-14
Refactor/Cleanup NetworkRequest and Related Code

Nerixyz:
# Description

This PR does many small changes - each commit can be reviewed on its own.

* Cleanup unused code (Twitch v5)
* Add `json` methods that simplify sending JSON. This also sets the `Accepts` header, since here, when JSON is sent, only JSON is accepted as a response (this is only done in Helix).
* `Helix`'s `makeRequest` now accepts an HTTP method.
* Cleaned some clang-tidy suggestions in `Network{Request,Private}`. The only suggestions left there are (1) a [`const-cast`](https://github.com/Chatterino/chatterino2/blob/4fa2cc26c9bffd5a6c770380c8a8bceb7368edbf/src/common/NetworkRequest.cpp#LL58C31-L58C41) and (2-5) unused `QFuture`'s from `QtConcurrent::run` calls. (1) should be resolved when #4632 is closed. (2-5) can be resolved by using [`QThreadPool::start`](https://doc.qt.io/qt-6/qthreadpool.html#start), since we don't need the future.
* Moved `asserts` to [`Q_ASSERT_X`](https://doc.qt.io/qt-6/qtassert-proxy.html#Q_ASSERT_X), which provides better error messages (mainly with a reasoning in the message).

### DIFF
--- a/src/common/NetworkPrivate.cpp
+++ b/src/common/NetworkPrivate.cpp
@@ -77,7 +77,7 @@ void loadUncached(std::shared_ptr<NetworkData> &&data)
     DebugCount::increase("http request started");
 
     NetworkRequester requester;
-    NetworkWorker *worker = new NetworkWorker;
+    auto *worker = new NetworkWorker;
 
     worker->moveToThread(&NetworkManager::workerThread);
 
@@ -245,12 +245,16 @@ void loadUncached(std::shared_ptr<NetworkData> &&data)
             if (data->onSuccess_)
             {
                 if (data->executeConcurrently_)
+                {
                     QtConcurrent::run([onSuccess = std::move(data->onSuccess_),
                                        result = std::move(result)] {
                         onSuccess(result);
                     });
+                }
                 else
+                {
                     data->onSuccess_(result);
+                }
             }
             // log("finished {}", data->request_.url().toString());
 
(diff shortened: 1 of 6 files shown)


src/providers/twitch/api/Helix.hpp
```diff
@@ -1361,7 +1361,9 @@ class Helix final : public IHelix
         FailureCallback<HelixGetModeratorsError, QString> failureCallback);
 
 private:
-    NetworkRequest makeRequest(QString url, QUrlQuery urlQuery);
+    NetworkRequest makeRequest(
```
pajlada, line 1364: I would rather see us be explicit in the method with separate functions, e.g. `makeGet`, `makePost`, `makeDelete`, `makePatch` if we want to refactor this. Otherwise, this doesn't do much than move the parameter from the builder to the function itself

src/common/NetworkPrivate.cpp
```diff
@@ -119,7 +122,10 @@ void loadUncached(const std::shared_ptr<NetworkData> &data)
                 case NetworkRequestType::Patch:
                     if (data->multiPartPayload_)
                     {
-                        assert(data->payload_.isNull());
+                        Q_ASSERT_X(
```
pajlada, line 125: The same message can be achieved with normal `assert` and doesn't add extra complexity (e.g. asserts firing because qt version was built with force-asserts, or the message not coming through because it uses qlogging)

e.g. `assert((data->payload_.isNull()) && "bla bla");`

Depending on who the assert is for, a comment would also be equally useful

src/common/NetworkPrivate.cpp
```diff
@@ -106,7 +106,10 @@ void loadUncached(const std::shared_ptr<NetworkData> &data)
                 case NetworkRequestType::Post:
                     if (data->multiPartPayload_)
                     {
-                        assert(data->payload_.isNull());
+                        Q_ASSERT_X(
```
pajlada, line 109: same assert comment

src/common/NetworkRequest.cpp
```diff
@@ -53,7 +42,8 @@ NetworkRequest NetworkRequest::caller(const QObject *caller) &&
     if (caller)
     {
         // Caller must be in gui thread
-        assert(caller->thread() == qApp->thread());
+        Q_ASSERT_X(caller->thread() == qApp->thread(), "NetworkRequest::caller",
```
pajlada, line 45: same assert comment

src/common/NetworkRequest.hpp
```diff
@@ -54,23 +58,26 @@ class NetworkRequest final
...
     NetworkRequest concurrent() &&;
-    NetworkRequest authorizeTwitchV5(const QString &clientID,
-                                     const QString &oauthToken = QString()) &&;
     NetworkRequest multiPart(QHttpMultiPart *payload) &&;
     /**
      * This will change `RedirectPolicyAttribute`.
      * `QNetworkRequest`'s defaults are used by default (Qt 5: no-follow, Qt 6: follow).
      */
     NetworkRequest followRedirects(bool on) &&;
+    NetworkRequest json(const QJsonObject &root, bool acceptJson = true) &&;
```
pajlada, line 73: I would remove the `acceptJson` parameter and just always assume we accept json in a method that sends json

pajlada (changes requested):
Small nitpicks, otherwise looks good

Nerixyz:
Constructing the url-query and json in Helix could be simplified by using the constructors taking initializer-lists.

src/common/NetworkPrivate.cpp
pajlada: My thinking boils down more to: Let's not replace 3 of 152 asserts when there's no need for the feature you're looking for.
Adding the extra information could just as well be done through a comment or through our standard `assert`

src/common/NetworkRequest.cpp
```diff
@@ -37,10 +29,7 @@ NetworkRequest::NetworkRequest(QUrl url, NetworkRequestType requestType)
     this->initializeDefaultValues();
 }
 
-NetworkRequest::~NetworkRequest()
-{
-    //assert(!this->data || this->executed_);
-}
+NetworkRequest::~NetworkRequest() = default;
```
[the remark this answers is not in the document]
pajlada: If it's in the header, there are more constraints for what needs to be included in the file. I imagine that's why they've chosen to keep this in the source file

pajlada (approved):
All changes look good to me :+1: 
If you want to re-add the assert comments in any way feel free, will merge this in later

PR #4637, opened 2023-05-15
Add Command to Set Logging/Filter Rules at Runtime

Nerixyz:
# Description

This PR adds the `/c2:loggingrules` command. You invoke it like this: `/c2:loggingrules chatterino.*.debug=true chatterino.http.debug=false`. This will apply both rules. I'm not too sure about the command name/prefix/namespace (my thinking was to have this behind a namespace/prefix, so it's clear that this is a Chatterino command).

This command is mainly for debugging release builds without restarting - so that you get more logs when you attach a debugger/console or even a dedicated program capturing logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Bugfix: Fixed Ctrl+Backspace not working after Select All in chat search popup. (#4461)
 - Bugfix: Fixed crash when scrolling up really fast. (#4621)
 - Dev: Added the ability to control the `followRedirect` mode for requests. (#4594)
+- Dev: Added command to set Qt's logging filter/rules at runtime (`/c2:loggingrules`). (#4637)
 
 ## 2.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Dev: Added command to set Qt's logging filter/rules at runtime (`/c2-set-logging-rules`). (#4637)
+
 ## 2.4.4
 
 - Minor: Added a Send button in the input box so you can click to send a message. This is disabled by default and can be enabled with the "Show send message button" setting. (#4607)
@@ -15,7 +17,6 @@
 - Bugfix: Fixed Ctrl+Backspace not working after Select All in chat search popup. (#4461)
 - Bugfix: Fixed crash when scrolling up really fast. (#4621)
 - Dev: Added the ability to control the `followRedirect` mode for requests. (#4594)
-- Dev: Added command to set Qt's logging filter/rules at runtime (`/c2:loggingrules`). (#4637)
 
 ## 2.4.3
 
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -3213,7 +3213,7 @@ void CommandController::initialize(Settings &, Paths &paths)
     this->registerCommand("/shield", &commands::shieldModeOn);
     this->registerCommand("/shieldoff", &commands::shieldModeOff);
 
-    this->registerCommand("/c2:loggingrules", &commands::setLoggingRules);
+    this->registerCommand("/c2-set-logging-rules", &commands::setLoggingRules);
 }
 
 void CommandController::save()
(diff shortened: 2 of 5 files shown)


src/controllers/commands/builtin/chatterino/Debugging.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+struct CommandContext;
+
+namespace commands {
+
+    QString setLoggingRules(const CommandContext &ctx);
+
+}  // namespace commands
+
+}  // namespace chatterino
```
pajlada, line 15: I would rather the forward-declaration is done in a separate namespace definition - see ShieldMode.hpp

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -0,0 +1,44 @@
...
+            "the format on the rules, see "
+            "https://doc.qt.io/qt-6/"
+            "qloggingcategory.html#configuring-categories"));
+        return {};
+    }
+
+    QLoggingCategory::setFilterRules(
+        QList(ctx.words.begin() + 1, ctx.words.end()).join('\n'));
+
+    auto message = QStringLiteral("Updated filter rules.");
```
pajlada, line 29: Break the final filter rules out into a local variable & add it to the message

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -0,0 +1,44 @@
...
+    if (ctx.words.size() < 2)
+    {
+        ctx.channel->addMessage(makeSystemMessage(
+            "Usage: /c2:loggingrules <rules...>. To enable debug logging for "
+            "all "
+            "categories from chatterino, use 'chatterino.*.debug=true'. For "
+            "the format on the rules, see "
+            "https://doc.qt.io/qt-6/"
+            "qloggingcategory.html#configuring-categories"));
+        return {};
```
pajlada, line 23: We return `""` at the end of the function - I'm fine with either but we should use the same in both places

src/controllers/commands/CommandController.cpp
```diff
@@ -3211,6 +3212,8 @@ void CommandController::initialize(Settings &, Paths &paths)
 
     this->registerCommand("/shield", &commands::shieldModeOn);
     this->registerCommand("/shieldoff", &commands::shieldModeOff);
+
+    this->registerCommand("/c2:loggingrules", &commands::setLoggingRules);
```
pajlada, line 3216: Naming is the hardest part - I think I would prefer `/c2-set-logging-rules` but happy to discuss it further before merging this in

pajlada (changes requested):
Small nitpicks, otherwise looks good :+1: Thank you

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Bugfix: Fixed Ctrl+Backspace not working after Select All in chat search popup. (#4461)
 - Bugfix: Fixed crash when scrolling up really fast. (#4621)
 - Dev: Added the ability to control the `followRedirect` mode for requests. (#4594)
+- Dev: Added command to set Qt's logging filter/rules at runtime (`/c2:loggingrules`). (#4637)
```
Wissididom, line 18: I think that should be in the `Unversioned` section, because 2.4.4 already is released.

Nerixyz: Thank you, totally overlooked that.

src/controllers/commands/CommandController.cpp
Nerixyz: Sounds like a good name to me.

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+    auto filterRules = QList(ctx.words.begin() + 1, ctx.words.end()).join('\n');
+#else
+    auto filterRules = [&]() {
+        auto tmpList = ctx.words;
+        tmpList.pop_front();
+        return tmpList.join('\n');
+    }();
+#endif
```
pajlada, line 33: Would this be a portable alternative? [qt5](https://doc.qt.io/qt-5/qlist.html#mid) [qt6](https://doc.qt.io/qt-6/qlist.html#mid)

```suggestion
    auto filterRules = ctx.words.mid(1).join('\n');
```

pajlada (changes requested):

pajlada (approved):

pajlada:
Thank you!

PR #4638, opened 2023-05-15
add "/shoutout" command

olafyang:
# Description

This pr adds a new "/shoutout" command which calls the helix api and shouts out the specified user.

Add:
- `/shoutout <username>` calls `chat/shoutouts` of helix api and gives specified user a twitch shoutout.

It seems like twitch is returning a 500 for all errors excluding 429 rate limit . Trying to shoutout an user who's not a mod gives back to following in the log. Therefore I can not provide more detailed error message for now.

```
chatterino.twitch: Helix send shoutout, unhandled error data: 500 "{\"error\":\"Internal Server Error\",\"status\":500,\"message\":\"\"}" QJsonObject({"error":"Internal Server Error","message":"","status":500})
```


src/providers/twitch/api/Helix.hpp
```diff
@@ -1318,6 +1333,12 @@ class Helix final : public IHelix
                           FailureCallback<HelixUpdateShieldModeError, QString>
                               failureCallback) final;
 
+    // https://dev.twitch.tv/docs/api/reference/#send-a-shoutout
+    void sendShoutout(
+        QString fromBroadCasterID, QString toBroadcasterID, QString moderatorID,
```
Felanbird, line 1338:
```suggestion
        QString fromBroadcasterID, QString toBroadcasterID, QString moderatorID,
```

tests/src/HighlightController.cpp
```diff
@@ -434,6 +434,14 @@ class MockHelix : public IHelix
                       failureCallback)),
                 (override));
 
+    // /shoutout
+    MOCK_METHOD(
+        void, sendShoutout,
+        (QString fromBroadCasterID, QString toBroadcasterID,
```
Felanbird, line 440:
```suggestion
        (QString fromBroadcasterID, QString toBroadcasterID,
```

src/providers/twitch/api/Helix.cpp
```diff
@@ -2631,6 +2631,66 @@ void Helix::updateShieldMode(
         .execute();
 }
 
+// https://dev.twitch.tv/docs/api/reference/#send-a-shoutout
+void Helix::sendShoutout(
+    QString fromBroadCasterID, QString toBroadcasterID, QString moderatorID,
```
Felanbird, line 2636:
```suggestion
    QString fromBroadcasterID, QString toBroadcasterID, QString moderatorID,
```

src/providers/twitch/api/Helix.cpp
```diff
@@ -2631,6 +2631,66 @@ void Helix::updateShieldMode(
...
+    ResultCallback<> successCallback,
+    FailureCallback<HelixSendShoutOutError, QString> failureCallback)
+{
+    using Error = HelixSendShoutOutError;
+
+    QUrlQuery urlQuery;
+    urlQuery.addQueryItem("from_broadcaster_id", fromBroadCasterID);
```
Felanbird, line 2643:
```suggestion
    urlQuery.addQueryItem("from_broadcaster_id", fromBroadcasterID);
```

Felanbird:
> It seems like twitch is returning a 500 for all errors excluding 429 rate limit 

I'm only able to replicate this issue when the shoutout sender is not a mod, I receive back a correct 401 when the `moderator_id` does not match the relevant oauth token, as well as a 400 when `from_broadcaster_id` is not live/has no viewers.

Felanbird (changes requested):

olafyang:
> > It seems like twitch is returning a 500 for all errors excluding 429 rate limit
> 
> I'm only able to replicate this issue when the shoutout sender is not a mod, I receive back a correct 401 when the `moderator_id` does not match the relevant oauth token, as well as a 400 when `from_broadcaster_id` is not live/has no viewers.
> 
> [twitchdev/issues#779](https://github.com/twitchdev/issues/issues/779)

i'll check on my end too and make changes, thanks!!

olafyang:
Typos are fixed in e52f8bba6e6b20785b63c4ced287059aa4f7d736.

I've also added some error handling for the api.

src/controllers/commands/CommandController.cpp
```diff
@@ -3211,6 +3211,111 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                                    "You are being ratelimited by Twitch. "
+                                    "Try again in a few seconds.";
+                            }
+                            break;
+
+                            case Error::UserNotMod: {
+                                errorMessage +=
+                                    "You are not a moderator of this channel.";
+                            }
+                            break;
```
pajlada, line 3287: I would think the "UserNotAuthorized" error covers this case already, so just remove the UserNotMod error case & make it UserNotAuthorized instead

src/providers/twitch/api/Helix.cpp
```diff
@@ -2631,6 +2631,96 @@ void Helix::updateShieldMode(
...
+                }
+                case 429: {
+                    failureCallback(Error::Ratelimited, message);
+                    break;
+                }
+                    // Helix returns 500 when user is not mod
+                case 500: {
+                    failureCallback(Error::UserNotMod,
+                                    "You are not a moderator.");
+                    break;
```
pajlada, line 2711: I would prefer to treat 500s as unhandled errors - just forward the message we get from twitch

src/controllers/commands/CommandController.cpp
```diff
@@ -3211,6 +3211,111 @@ void CommandController::initialize(Settings &, Paths &paths)
...
+                                    "The broadcaster is not streaming live or "
+                                    "does not have one or more viewers.";
+                            }
+                            break;
+
+                            case Error::Unknown: {
+                                errorMessage +=
+                                    "An unknown error has occurred.";
+                            }
+                            break;
```
pajlada, line 3306: I would be ok with including the `message` portion in unhandled errors here, so the user would see something like "Internal server error" or w/e if we don't handle one of the error cases.

src/controllers/commands/CommandController.cpp
```diff
@@ -3211,6 +3211,111 @@ void CommandController::initialize(Settings &, Paths &paths)
 
     this->registerCommand("/shield", &commands::shieldModeOn);
     this->registerCommand("/shieldoff", &commands::shieldModeOff);
+
+    this->registerCommand("/shoutout", [](const QStringList &words,
+                                          auto channel) {
```
pajlada, line 3216: I would prefer if this command was moved to its own file under `src/controllers/commands/builtin/twitch` (see `ShieldMode.hpp`/`ShieldMode.cpp`)

pajlada (changes requested):
Looks good overall, some nitpicks with error handling & where the code is located

src/providers/twitch/api/Helix.cpp
Mm2PL: Twitch likes to give an empty message on 500s so there needs to be a genetic "twitch fucked up" error case and one that forwards a non-empty message.

olafyang:
oops i forgot to return at the anon check and the argument check, i'll fix them in the evening once i get home

pajlada:
Currently fixing the merge conflicts, will solve the returns in the same batch

pajlada (approved):
Looks good to me, works as expected in the few scenarios I tested it

When you've taken a look at the changes @olafyang we can go ahead and merge this in

olafyang:
yep looks fantastic!! please go ahead and merge it. i learned a lot from this, thanks ❤️❤️ would love to contribute some more in the future

src/providers/twitch/api/Helix.cpp
```diff
@@ -2597,6 +2597,106 @@ void Helix::updateShieldMode(
...
+                case 429: {
+                    failureCallback(Error::Ratelimited, message);
+                }
+                break;
+
+                case 500: {
+                    // Helix returns 500 when user is not mod,
+                    if (message.isEmpty())
+                    {
+                        failureCallback(Error::Unknown, "Empty message");
```
Mm2PL, line 2680: Can we maybe add a thing to the message that the api fails when not mod?

pajlada: There was originally, but having that error for a generic error 500 will lead to outdated or wrong errors.
I would rather incorrectly trust Twitch's error handling than spit out bad errors ourselves

Felanbird (approved):

src/providers/twitch/api/Helix.cpp
pajlada: message updated from "Empty message" to "Twitch internal server error" to give users a bit more of a clear idea of what's going on

pajlada:

Felanbird:
Also you'll want to do it on a branch after you clean up your master branch, as creating this PR from your master branch has made a mess for you 😁

PR #4639, opened 2023-05-16
Consolidate input completion code in preparation for advanced completion strategies

dnsge:
This PR looks to consolidate the input completion code from `InputCompletionPopup` and `CompletionModel` into a single, extendible framework for sourcing and applying completions. 

This PR was [made in preparation](https://github.com/Chatterino/chatterino2/discussions/4414#discussioncomment-5115302) for implementing ideas from #4414. Additionally, because the framework accepts arbitrary strategies for generating autocomplete suggestions, #4519 can also be incorporated into this framework.

## Core Concepts
- `AutocompleteSource`: Represents a source for querying autocomplete suggestions. 
- `AutocompleteStrategy`: Represents a strategy for filtering/ordering items from an `AutocompleteSource`
  - Classic autocompletion strategies have been migrated
  - New strategies such as those discussed in #4414 or #4519 can be implemented and plugged in without changing other code
- `AutocompleteModel`: A `GenericListModel` which accepts an arbitrary `AutocompleteSource` to provide suggestions
- `TabAutocompleteModel`: A `QStringListModel` which automatically selects an appropriate `AutocompleteSource` based on the current query

## Classic Behavior
All tab completion and popup completion behavior has been replicated exactly (according to my testing). After we implement more advanced completion strategies, users who have built up muscle memory and don't want to use the new strategies could use the classic strategies by changing a setting.

## Testable
In theory, this completion framework is testable. However, I don't really know how to mock `IApplication` to correctly provide emotes, users, commands, etc. during testing.

## Example Strategy

Below is an example strategy that mostly implements the ideas presented in #4414. It doesn't handle edge cases like matching `:)` correctly and has no form of tiebreak. However, the `CoverScoreAutocompleteEmoteStrategy` can be plugged into the existing framework by changing one line of code (either in `InputCompletionPopup::getSource` or `TabAutocompleteModel::buildSource`) 

```c++
float scoreCompletionEmote(const CompletionEmote &emote, const QString &query)
{
    return (float)query.length() / emote.searchName.length();
}

void CoverScoreAutocompleteEmoteStrategy::apply(
    const std::vector<CompletionEmote> &items,
    std::vector<CompletionEmote> &output, const QString &query) const
{
    QString normalizedQuery = query;
    if (normalizedQuery.startsWith(':'))
    {
        normalizedQuery = normalizedQuery.mid(1);
    }

    for (const auto &item : items)
    {
        if (item.searchName.contains(normalizedQuery, Qt::CaseInsensitive))
        {
            output.push_back(item);
        }
    }

    std::sort(output.begin(), output.end(),
              [&](const CompletionEmote &a, const CompletionEmote &b) {
                  return scoreCompletionEmote(a, normalizedQuery) >
                         scoreCompletionEmote(b, normalizedQuery);
              });
}
```


pajlada (commented):
With a big change like this, tests and benchmarks are going to be extremely valuable.
Not just tests & benchmarks for the new code, but also for the old code. I think the approach for a big refactor like this, is going to be to modify the existing code to make it testable & benchmarkable, and then make the refactor, make sure all tests complete as expected (so the current behaviour **definitely** doesn't change), and that we don't slow things down by too much.

Abstraction wise, I would prefer to see less levels of abstraction. For AutocompleteSource for example, the `AutocompleteGenericSource` is something I'd like to see go away.
If this means sources like `AutocompleteEmoteSource` has to have an implementation that just calls a helper function, that's fine. I'd rather have a bit more c-v-paste than extra abstraction layers where possible

When it comes to tests, look at HighlightController to see how it provides highlights.
The potential lift with testing is that you would need to convert certain things to interfaces so they can be mockable in the first place. Emotes.hpp contains a mockable Emotes interface, so you can see how that's done for badges/commands etc

Overall the approach looks fine, will be easier to give proper feedback when we're aligned on code style/testing/benchmarks etc

dnsge:
After integrating the test cases from #4644, I discovered a couple discrepancies with the classic completion strategies which has been fixed in 85939eb.

Additionally, I realized that the tab completion was not respecting the `userCompletionOnlyWithAt` setting, which 
3f7a976 addressed.

With these being fixed, I feel the PR is now ready for a proper review.

src/providers/autocomplete/AutocompleteModel.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// @brief Represents the kind of completion occurring
+enum class CompletionKind { Emote, User };
+
+/// @brief AutocompleteModel is a GenericListModel intended to provide completion
+/// suggestions to an InputCompletionPopup. The popup can determine the appropriate
+/// source based on the current input and the user's preferences.
+class AutocompleteModel : public GenericListModel
+{
+public:
+    AutocompleteModel(QObject *parent = nullptr);
```
pajlada, line 23: Remove the default parameter, force users to specify nullptr if they really don't want to be the parent to this model

Make the constructor explicit if possible 
```suggestion
    AutocompleteModel(QObject *parent = nullptr);
```

src/providers/autocomplete/AutocompleteSources.cpp
```diff
@@ -0,0 +1,430 @@
+#include "providers/autocomplete/AutocompleteSources.hpp"
```
pajlada, line 1: Make a new directory for the CompletionSources and split each up into their own file

so we'd have:
 - `src/controllers/completion/sources/unified.hpp` + its cpp file
 - `src/controllers/completion/sources/commands.hpp`
 - `src/controllers/completion/sources/users.hpp`
 - etc...

src/providers/autocomplete/AutocompleteStrategy.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
pajlada, line 1: Same as the sources, I'd like a new directory where each strategy has its own files

src/providers/autocomplete/AutocompleteSources.hpp
```diff
@@ -0,0 +1,182 @@
...
+public:
+    using ActionCallback = std::function<void(const QString &)>;
+    using AutocompleteEmoteStrategy = AutocompleteStrategy<CompletionEmote>;
+
+    /// @brief Initializes a source for CompletionEmotes from the given channel
+    /// @param channel Channel to initialize emotes from
+    /// @param strategy AutocompleteStrategy to apply
+    /// @param callback ActionCallback to invoke upon InputCompletionItem selection.
+    /// See InputCompletionItem::action(). Can be nullptr.
+    AutocompleteEmoteSource(const Channel &channel,
```
pajlada, line 44: This should just take a `const Channel *channel` - it currently goes from a shared_ptr<Channel>, gets deferenced to this reference, then internally gets referenced to a pointer

pajlada (changes requested):
Some formatting / structural things before I get into the meaty parts of the review

pajlada:
Overall, btw, this looks good and works as expected :+1: I don't expect there to be any big changes necessary

pajlada (approved):
Let's test this in nightly :+1: 
Thanks for putting in the effort!

PR #4644, opened 2023-05-20
Add input completion test suite

dnsge:
# Description

Adds tests for `InputCompletionPopup` and `CompletionModel` behavior as a precursor to merging #4639. While developing these tests I identified two discrepancies in behavior from #4639 which I will fix (eventually).

This PR needed to make some changes to `BttvEmotes` (among others) to allow directly setting loaded emotes for mocking. It looks a bit weird but it seems necessary. 

## Benchmarking

I've all but finished writing benchmarks for the existing input completion code. However, I can't figure out a smart way to access `tests/src/mocks` from both tests and benchmarks. The code should obviously be shared. I tried moving it to `src` but that would require linking `gtest` into `chatterino-lib` (I think), which I don't understand the full implications of (nor would be I able to modify the cmake config to do so).

If anyone (read: pajlada) can think of a way to share the mocks between tests and benchmarks and how to configure `CMakeLists.txt` (as I also can't figure that out when simply copying `tests/src/mocks` to `benchmarks/src/mocks`), I'd be very grateful :)


dnsge:
I'll note that this PR does not include completion tests for username nor commands. Testing these would require a fully mockable `TwitchChannel` instance because the completion code relies on a dynamic cast. 

This is one of the advantages of the new system in the pending PR as strategies are testable outside of sources.

src/providers/seventv/SeventvEmotes.hpp
```diff
@@ -75,6 +75,7 @@ class SeventvEmotes final
     std::shared_ptr<const EmoteMap> globalEmotes() const;
     boost::optional<EmotePtr> globalEmote(const EmoteName &name) const;
     void loadGlobalEmotes();
+    void setEmotes(std::shared_ptr<const EmoteMap> emotes);
```
Nerixyz, line 78: I think this should be `setGlobalEmotes` (since the corresponding load is called `loadGlobalEmotes`).

```suggestion
    void setGlobalEmotes(std::shared_ptr<const EmoteMap> emotes);
```

tests/src/InputCompletion.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+})!";
+
+class InputCompletionTest : public ::testing::Test
+{
+protected:
+    void SetUp() override
+    {
+        // Write default settings to the mock settings json file
+        ASSERT_TRUE(QDir().mkpath("/tmp/c2-tests"));
```
Nerixyz, line 125: It would be awesome if the tests could run on all supported platforms. Consider using [`QTemporaryDir`](https://doc.qt.io/qt-6/qtemporarydir.html) to create a temporary directory.

benchmarks/src/Emojis.cpp
```diff
@@ -8,6 +10,8 @@ using namespace chatterino;
 
 static void BM_ShortcodeParsing(benchmark::State &state)
 {
+    chatterino::Settings settings("/tmp/c2-mock");
```
Nerixyz, line 13: Same comment as below regarding platform independence.

Nerixyz (commented):
Small things I noticed while looking at this.

tests/src/InputCompletion.cpp
pajlada: yes, but should not be done in this PR

benchmarks/src/Emojis.cpp
pajlada: Same response as before - yes tests should be cross-platform but not as part of this PR

pajlada:
I made the mock change to make them accessible for both tests & benchmarks here https://github.com/Chatterino/chatterino2/pull/4645

src/common/CompletionModel.hpp
```diff
@@ -56,6 +56,7 @@ class CompletionModel : public QAbstractListModel
     int rowCount(const QModelIndex &parent) const override;
 
     void refresh(const QString &prefix, bool isFirstWord = false);
+    std::vector<QString> allItems() const;
```
pajlada, line 59: Make private & add `InputCompletionTest` as a friend

src/common/CompletionModel.cpp
```diff
@@ -256,6 +258,19 @@ void CompletionModel::refresh(const QString &prefix, bool isFirstWord)
     }
 }
 
+std::vector<QString> CompletionModel::allItems() const
+{
+    std::unique_lock lock(this->itemsMutex_);
```
pajlada, line 263: This can be a shared lock since we only read from items_

src/widgets/splits/InputCompletionPopup.cpp
```diff
@@ -55,97 +50,112 @@ void addEmojis(std::vector<CompletionEmote> &out, const EmojiMap &map,
...
-
-    QObject::connect(&this->redrawTimer_, &QTimer::timeout, this, [this] {
-        if (this->isVisible())
-        {
-            this->ui_.listView->doItemsLayout();
-        }
-    });
-    this->redrawTimer_.setInterval(33);
-}
+namespace detail {
```
pajlada, line 53: Make this not indent by moving it to its own separate namespace block

```c++
namespace chatterino::detail {
...
}

namespace chatterino {
// old code
}
```

pajlada (changes requested):
Small nitpicks, haven't done a thorough review yet

pajlada (approved):
:+1: Thank you!

PR #4645, opened 2023-05-20
Move mocks to their own interface

pajlada:
# Description

Code mostly taken from https://github.com/Chatterino/chatterino2/pull/4644/files

This allows us to share mocks between multiple projects (i.e. tests & benchmarks)


pajlada:
LGTM 
image

PR #4650, opened 2023-05-22
Make tests more platform agnostic

Nerixyz:
# Description

* Windows systems (one of the supported operating systems) doesn't have a `/tmp` directory the same way, e.g. Linux has it. So instead, the tests are now using `QTemporaryDir` - which is cross-platform.
* Running Docker on Windows or macOS effectively runs an entire VM, which is a lot of overhead to _just_ get an HTTP echo server. I added a CMake option to use the public (but less reliable) httpbin server (`CHATTERINO_TEST_LOCAL_HTTPBIN=Off`).
* MSVC (apparently) has a different implementation of a `std::unordered_map` than the one used when writing the input completion tests. Since `unordered_map` doesn't guarantee an ordering (the name explicitly says `unordered`), one can't rely on values being in order in such a map. This map is the underlying data-type for an `EmoteMap`. For the input completions, this is a bit more tricky, since they do guarantee some amount of ordering (regarding the emote/emoji provider). So, for an emote provider, the completions are now sorted.
* For some reason, connecting to the PubSub server always took more than 50ms for me, so I increased the delay.

Now, the only program one needs to run next to the tests is the PubSub test-server.

_Note:_ The tests will still fail when using Qt 6 (which is still marked as experimental and thus not tested). I'll make another PR to fix that and add tests there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,6 @@ option(BUILD_TRANSLATIONS "" OFF)
 option(BUILD_SHARED_LIBS "" OFF)
 option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable EXPERIMENTAL plugin support in Chatterino" OFF)
-option(CHATTERINO_TEST_LOCAL_HTTPBIN "Use local httpbin docker image" ON)
 
 if(CHATTERINO_LTO)
     include(CheckIPOSupported)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ option(BUILD_TRANSLATIONS "" OFF)
 option(BUILD_SHARED_LIBS "" OFF)
 option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable EXPERIMENTAL plugin support in Chatterino" OFF)
+option(CHATTERINO_TEST_LOCAL_HTTPBIN "Use local httpbin docker image" ON)
 
 if(CHATTERINO_LTO)
     include(CheckIPOSupported)
--- a/tests/src/InputCompletion.cpp
+++ b/tests/src/InputCompletion.cpp
@@ -179,8 +179,8 @@ private:
     {
         auto bttvEmotes = std::make_shared<EmoteMap>();
         addEmote(*bttvEmotes, "FeelsGoodMan");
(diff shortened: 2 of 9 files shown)


CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ option(BUILD_TRANSLATIONS "" OFF)
 option(BUILD_SHARED_LIBS "" OFF)
 option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable EXPERIMENTAL plugin support in Chatterino" OFF)
+option(CHATTERINO_TEST_LOCAL_HTTPBIN "Use local httpbin docker image" ON)
```
pajlada, line 29: Define this option in the `tests/CMAKELISTS.txt` file instead (as far as I remember this should be possible)

Also inverse the option so it's something like `CHATTERINO_TEST_USE_PUBLIC_HTTPBIN` where the default is OFF

tests/src/InputCompletion.cpp
```diff
@@ -222,15 +226,25 @@ class InputCompletionTest : public ::testing::Test
...
+    const auto cmpCompletion = [](const auto &a, const auto &b) {
+        return a.displayName < b.displayName;
+    };
+
     auto completion = queryEmoteCompletion(":feels");
     ASSERT_EQ(completion.size(), 3);
-    ASSERT_EQ(completion[0].displayName, "FeelsBirthdayMan");
-    ASSERT_EQ(completion[1].displayName, "FeelsBadMan");
+    // all these matches are BTTV global emotes
+    std::sort(completion.begin(), completion.end(), cmpCompletion);
```
pajlada, line 238: I'm hesitant about this change - I'd rather leave this change out as part of the platform-agnostic PR & discuss it in a separate PR/issue.
Most likely I'd like to see us just change the EmoteMap to be consistently ordered instead

pajlada (changes requested):

tests/src/InputCompletion.cpp
Nerixyz: I reverted the change, but added a note for future visitors.

> Most likely I'd like to see us just change the EmoteMap to be consistently ordered instead

Not sure if it's that easy. It should probably be benchmarked to see if it makes a meaningful difference, since working with emotes and checking for emotes in a word is an integral part.

pajlada (approved):
Looks good, thank you!
One step closer to running tests on all platforms

PR #4663, opened 2023-05-31
Consider nicknames when searching

chrrs:
image

With this PR, searching now also finds results using nicknames instead of usernames.

The way I've done this is by prepending the stylized username to the search text, as it seems that nickname matching is based on the stylized username, instead of only the login name / display name. This isn't the best solution, as it does lead to a bit of duplication in the search text (for example, for the username `chrrrs`, it would lead to something like `chrrrs chrrrs: message`). If anyone has a better solution here, I'd be up for changing things around.

### DIFF
--- a/src/controllers/nicknames/Nickname.hpp
+++ b/src/controllers/nicknames/Nickname.hpp
@@ -58,25 +58,24 @@ public:
         return this->isCaseSensitive_;
     }
 
-    [[nodiscard]] bool match(QString &usernameText) const
+    [[nodiscard]] boost::optional<QString> match(const QString &usernameText) const
     {
         if (this->isRegex())
         {
             if (!this->regex_.isValid())
             {
-                return false;
+                return boost::none;
             }
             if (this->name().isEmpty())
             {
-                return false;
+                return boost::none;
             }
 
             auto workingCopy = usernameText;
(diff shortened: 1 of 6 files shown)


Felanbird:
https://github.com/Chatterino/chatterino2/blob/e9432d3b6590090961b869fdc260d3ec462ada39/src/widgets/settingspages/NicknamesPage.cpp#L20-L23

This will need to be updated also

chrrs:
Ah yeah, I updated it now to remove the search part from the text

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Consider nicknames when searching for messages (#4663)
```
pajlada, line 5: Small rewording suggestion

```suggestion
- Minor: Nicknames are now taken into consideration when searching for messages. (#4663)
```

pajlada (approved):
Thank you!

pajlada:
@chrrs As a first-time contributor, you can add yourself to the contributors list that's shown inside the About page in Chatterino.

If you want this, wait for this PR to be merged in, then open a new PR where you modify the `resources/contributors.txt` file and add yourself to the list. Make sure to read the comments at the top for instructions.

PR #4666, opened 2023-06-04
Remove QObjectRef in favor of QPointer

### LINKED ISSUE #4632: Remove QObjectRef in favor of QPointer

4rneee:
# Description
This PR resolves #4632


src/common/NetworkPrivate.cpp
```diff
@@ -174,7 +174,7 @@ void loadUncached(std::shared_ptr<NetworkData> &&data)
         }
 
         auto handleReply = [data, reply]() mutable {
-            if (data->hasCaller_ && !data->caller_.get())
+            if (data->hasCaller_ && !data->caller_.data())
```
pajlada, line 177: Since we are not using the returned pointer, use QPointers [isNull](https://doc.qt.io/qt-6/qpointer.html#isNull) function instead

Same goes for the rest of the data() checks in this file

pajlada (changes requested):
Small nitpick for clarity (thanks @Nerixyz )

pajlada (approved):

PR #4680, opened 2023-06-11
Add option to subscribe to and pin reply threads

Nerixyz:
# Description

There are two changes in this PR. First, the option to change the subscription status of a reply thread is added. Secondly, a pin-button is added, like on user cards:

example

This adds the concept of a subscribed thread. If one is subscribed to a thread, new messages will be highlighted (requires the _Subscribed Reply Threads_ highlight option). One can subscribe to threads through the popup and by participating in a thread (requires _Subscribe to participated reply threads_).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Minor: Improved editing hotkeys. (#4628)
 - Minor: The input completion and quick switcher are now styled to match your theme. (#4671)
 - Minor: Added setting to only show tabs with live channels (default toggle hotkey: Ctrl+Shift+L). (#4358)
+- Minor: Added option to subscribe to and unsubscribe from reply threads. (#4680)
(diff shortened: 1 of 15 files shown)


Nerixyz:
_Regarding https://github.com/Chatterino/chatterino2/pull/4678:_

```
C:\Users\runneradmin\.cargo\bin\sccache.exe -s
  Compile requests                    474
  Compile requests executed           474
  Cache hits                           67
  Cache hits (C/C++)                   67
  Cache misses                        407
  Cache misses (C/C++)                407
```

I think the git output is causing the misses, since it adds different definitions every time - 67 hits are from the `crashpad_handler`.

Felanbird (approved):
Seems to work exactly as I would expect it to, pin properly only shows up when it's related `auto hide` setting is enabled, same with the `subscribe` feature only appearing if you have thread highlights enabled. :+1:

image

pajlada (approved):

PR #4681, opened 2023-06-11
Define Git and date constants in executables only

Nerixyz:
# Description

This moves the definitions `CHATTERINO_{GIT_{HASH, RELEASE, COMMIT, MODIFIED}, CMAKE_GEN_DATE}` into the executable only. This removes them from the `chatterino-lib` target, which in turn makes it possible to cache, since the definitions will be the same across many runs (mentioned in https://github.com/Chatterino/chatterino2/pull/4680#issuecomment-1586303503).

**Local compile times will also be improved (e.g. when recompiling on a new day).**

To test this, you should rerun the Windows CI after this day (to get a new generator date) and check the _Post Setup sccache (Windows)_ step for the amount of cache-hits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Minor: Improved editing hotkeys. (#4628)
 - Minor: The input completion and quick switcher are now styled to match your theme. (#4671)
 - Minor: Added setting to only show tabs with live channels (default toggle hotkey: Ctrl+Shift+L). (#4358)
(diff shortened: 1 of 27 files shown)


pajlada (changes requested):
As we discussed in chat, having the version header & source file as part of a CMake `add_library(... OBJECT)` would make this a bit cleaner, overall approach looks good

pajlada:
Each executable has to also link to the VERSION_PROJECT. objects don't have a link step so it can't be included in sub-dependencies

pajlada (approved):

PR #4689, opened 2023-06-17
Make sanitizers truly optional

pajlada:
# Description

This allows the sanitizers submodule to be omitted if it's not necessary for the developer


pajlada:
image

PR #4693, opened 2023-06-18
Extend viewer list, /mods & /vips with new provider

lvoegl:
# Description

Implements [Chatters API](https://github.com/lvoegl/chatters-api) provider for loading viewer list, VIPs and moderators without moderator/broadcaster rank if authorized. Helix API is preferred if the user is broadcaster, otherwise Chatters API is queried.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Extend viewer list with [`Chatters API`](https://github.com/lvoegl/chatters-api).
 - Minor: Nicknames are now taken into consideration when searching for messages. (#4663)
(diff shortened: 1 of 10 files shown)


pajlada:
My initial thought is that I don't want to accept this into Chatterino.

Pros:
 - Users can use `/mods` to know who's a moderator
 - Users can see what moderators are in the chat with the chatter list

Cons:
 - Added complexity
 - Yet another third party API

For both of the Pros, we can provide what I think is good-enough functionality for it once we build the chatter list locally, similar to what Chatty does.

Happy to discuss it more, either here or in our Discord

lvoegl:
I just find it sad to see a blank viewer list on basically every stream, but I do see the cons and that this is not a perfect solution

PR #4700, opened 2023-06-22
Refactor tests and benchmarks

Nerixyz:
# Description

This PR makes a few changes:

* Tests and benchmarks run on the main thread
  * This caused network tests to fail, since they rely on the Qt event-loop. These were adjusted to use [`qWaitFor`](https://doc.qt.io/qt-6/qtest.html#qWaitFor) which keeps the event-loop spinning. Since this is actually [declared in QtCore](https://github.com/qt/qtbase/blob/1a9ba8ee7d6d7cd624f5a299141bf4069a7cdeab/src/corelib/kernel/qtestsupport_core.h#L17-L52), no dependency to the Test component is necessary.
* Running benchmarks/tests related to emojis would segfault (at least for me with MSVC), because the `ImageExpirationPool` would be destroyed before the `Image` from `getEmpty` (the destructor calls `removeImagePtr`).

  The fix for this is a bit more involved, as it's [not trivial](https://stackoverflow.com/a/469613/16300717) to control the order of destruction of static inside functions, so the statics are moved inside an anonymous namespace.
  Creating a `ImageExpirationPool` when starting up will cause the `freeTimer_` to not work, as the main thread isn't running yet. That's why it's created lazily.

  For the emoji tests/benchmarks, this will cause `QObject::startTimer: Timers can only be used with threads started with QThread` to be printed after they're finished, because that's the first time the `ImageExpirationPool` is accessed (when destroying the empty image). This _can be_ fixed by e.g. calling `ImageExpirationPool::instance()` somewhere before/during the tests/benchmarks, but it's not really an issue **here**, since we're about to destroy the pool regardless (thus the timer would've never fired).
* ~~The tests for filters failed on Qt 6 (filters are probably not fully working in Qt 6, so this could be a separate PR).~~
* https://httpbin.org was really slow for me at times, and sometimes it even timed out. I found https://postman-echo.com, which has the same API except for the special 402 status response.

Now the following is possible:

* Running _all_ tests in debug mode (this means that assertions will fire!).
* Using [`ctest`](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html).

### DIFF
--- a/src/controllers/filters/lang/expressions/BinaryOperation.cpp
+++ b/src/controllers/filters/lang/expressions/BinaryOperation.cpp
@@ -2,43 +2,6 @@
 
 #include <QRegularExpression>
 
-namespace {
-
-/// Loosely compares `lhs` with `rhs`.
-/// This attempts to convert both variants to a common type if they're not equal.
-bool looselyCompareVariants(QVariant lhs, QVariant rhs)
-{
-#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-    // Qt 6 and later don't convert types as much as Qt 5 did when comparing.
-    //
-    // Based on QVariant::cmp from Qt 5.15
-    // https://github.com/qt/qtbase/blob/29400a683f96867133b28299c0d0bd6bcf40df35/src/corelib/kernel/qvariant.cpp#L4039-L4071
-    if (lhs.metaType() != rhs.metaType())
-    {
-        if (rhs.canConvert(lhs.metaType()))
-        {
-            if (!rhs.convert(lhs.metaType()))
-            {
-                return false;
-            }
-        }
-        else
-        {
-            // try the opposite conversion, it might work
-            qSwap(lhs, rhs);
-            if (!rhs.convert(lhs.metaType()))
-            {
-                return false;
-            }
-        }
-    }
-#endif
-
-    return lhs == rhs;
-}
-
-}  // namespace
-
 namespace chatterino::filters {
 
 BinaryOperation::BinaryOperation(TokenType op, ExpressionPtr left,
@@ -97,14 +60,14 @@ QVariant BinaryOperation::execute(const ContextMap &context) const
                 return left.toString().compare(right.toString(),
                                                Qt::CaseInsensitive) == 0;
             }
-            return looselyCompareVariants(left, right);
+            return left == right;
         case NEQ:
             if (variantTypesMatch(left, right, QMetaType::QString))
             {
                 return left.toString().compare(right.toString(),
                                                Qt::CaseInsensitive) != 0;
             }
-            return !looselyCompareVariants(left, right);
+            return left != right;
         case LT:
             if (convertVariantTypes(left, right, QMetaType::Int))
                 return left.toInt() < right.toInt();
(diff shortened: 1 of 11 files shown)


Nerixyz (commented):
> and use ctest 

It doesn't actually use `ctest` - but it could: `ctest -j$(nproc) --output-on-failure` (optionally with `-T test`).

src/controllers/filters/lang/expressions/BinaryOperation.cpp
```diff
@@ -2,6 +2,43 @@
 
 #include <QRegularExpression>
 
+namespace {
+
+/// Loosely compares `lhs` with `rhs`.
+/// This attempts to convert both variants to a common type if they're not equal.
+bool looselyCompareVariants(QVariant lhs, QVariant rhs)
```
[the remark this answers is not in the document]
Nerixyz: I could do something like this (it would just use references, because of the `qSwap` call).

tests/src/NetworkRequest.cpp
```diff
@@ -183,30 +164,21 @@ TEST(NetworkRequest, FinallyCallbackOnError)
...
+            .finally([&requestDone, &finallyCalled] {
+                finallyCalled = true;
+                requestDone = true;
+            })
             .execute();
 
         // Wait for the request to finish
-        std::unique_lock lck(mut);
-        requestDoneCondition.wait(lck, [&requestDone] {
+        EXPECT_TRUE(QTest::qWaitFor([&requestDone] {
```
pajlada, line 179: Keep condition variable, use wait_for and manually call https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qtestsupport_core.h#L33-L40

can make helper function if it feels cleaner

tests/src/NetworkRequest.cpp
```diff
@@ -183,30 +164,21 @@ TEST(NetworkRequest, FinallyCallbackOnError)
     for (const auto code : codes)
     {
         auto url = getStatusURL(code);
-        std::mutex mut;
```
pajlada, line 186: rename this variable

tests/src/NetworkRequest.cpp
```diff
@@ -3,19 +3,24 @@
...
 using namespace chatterino;
 
 namespace {
 
 #ifdef CHATTERINO_TEST_USE_PUBLIC_HTTPBIN
-const char *const HTTPBIN_BASE_URL = "http://httpbin.org";
+// Not using httpbin.org, since it can be really slow and cause timeouts.
+// postman-echo has the same API. They only differ for the 402 response.
+const char *const HTTPBIN_BASE_URL = "https://postman-echo.com";
+constexpr const bool CUSTOM_402_RESPONSE = false;
```
pajlada, line 20: We can remove our test for the 402 code instead, and that should let us only replace the HTTPBIN base url

src/messages/Image.cpp
```diff
@@ -31,6 +31,46 @@ const auto IMAGE_POOL_CLEANUP_INTERVAL = std::chrono::minutes(1);
...
+
+}  // namespace chatterino
+
+namespace {
+using namespace chatterino;
+
+// These need to be in this order to ensure the empty image gets deleted before the expiration pool is deleted.
+//
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+LazyExpirationPool globalImageExpirationPool;
```
pajlada, line 68: I'd prefer if the ImageExpirationPool is just never free'd instead

src/controllers/filters/lang/expressions/BinaryOperation.cpp
```diff
@@ -2,6 +2,43 @@
 
 #include <QRegularExpression>
 
+namespace {
```
pajlada, line 5: I'd prefer the qt6 & these fixes to be a separate PR

pajlada (changes requested):

pajlada (approved):

PR #4704, opened 2023-06-25
Improve network error messages

Nerixyz:
# Description

Previously, some network errors would print `0` as their status code. This was because no status code was set for the reply. Instead, the error was in [`QNetworkReply::error`](https://doc.qt.io/qt-6/qnetworkreply.html#error) which wasn't used in error reporting. It's now used - if available, the HTTP status is still preferred.

I couldn't really come up with a good test case for `NetworkRequest` to test that the error is actually set.

### DIFF
--- a/src/common/NetworkResult.hpp
+++ b/src/common/NetworkResult.hpp
@@ -34,7 +34,7 @@ public:
     }
 
     /// The HTTP status code if a request was made.
-    std::optional<uint16_t> status() const
+    std::optional<int> status() const
     {
         return this->status_;
     }
@@ -48,7 +48,7 @@ private:
     QByteArray data_;
 
     Error error_;
-    std::optional<uint16_t> status_;
+    std::optional<int> status_;
 };
 
 }  // namespace chatterino
--- a/src/common/NetworkResult.hpp
+++ b/src/common/NetworkResult.hpp
@@ -12,9 +12,10 @@ namespace chatterino {
 class NetworkResult
 {
 public:
-    using Error = QNetworkReply::NetworkError;
+    using NetworkError = QNetworkReply::NetworkError;
 
-    NetworkResult(Error error, const QVariant &httpStatusCode, QByteArray data);
+    NetworkResult(NetworkError error, const QVariant &httpStatusCode,
+                  QByteArray data);
 
     /// Parses the result as json and returns the root as an object.
     /// Returns empty object if parsing failed.
@@ -28,26 +29,25 @@ public:
 
     /// The error code of the reply.
     /// In case of a successful reply, this will be NoError (0)
-    Error error() const
+    NetworkError error() const
     {
         return this->error_;
     }
 
-    /// The HTTP status code if a request was made.
+    /// The HTTP status code if a response was received.
     std::optional<int> status() const
     {
         return this->status_;
     }
 
-    /// Formats the error:
-    /// without HTTP status: [Name]
-    /// with HTTP status: [HTTP status]
+    /// Formats the error.
+    /// If a reply is received, returns the HTTP status otherwise, the network error.
     QString formatError() const;
 
 private:
     QByteArray data_;
 
-    Error error_;
+    NetworkError error_;
     std::optional<int> status_;
 };
 
--- a/src/common/NetworkResult.hpp
+++ b/src/common/NetworkResult.hpp
@@ -2,14 +2,19 @@
 
 #include <QJsonArray>
 #include <QJsonObject>
+#include <QNetworkReply>
 #include <rapidjson/document.h>
+
+#include <optional>
 
 namespace chatterino {
 
 class NetworkResult
 {
 public:
-    NetworkResult(const QByteArray &data, int status);
+    using Error = QNetworkReply::NetworkError;
+
+    NetworkResult(Error error, const QVariant &httpStatusCode, QByteArray data);
 
     /// Parses the result as json and returns the root as an object.
     /// Returns empty object if parsing failed.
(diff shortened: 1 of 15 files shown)


src/common/NetworkResult.hpp
```diff
@@ -20,13 +25,30 @@ class NetworkResult
...
+    /// without HTTP status: [Name]
+    /// with HTTP status: [HTTP status]
+    QString formatError() const;
 
 private:
     QByteArray data_;
-    int status_;
+
+    Error error_;
+    std::optional<uint16_t> status_;
```
pajlada, line 51: Just leave it as an optional int instead

src/common/NetworkResult.hpp
```diff
@@ -20,13 +25,30 @@ class NetworkResult
...
 
-    static constexpr int timedoutStatus = -2;
+    /// The error code of the reply.
+    /// In case of a successful reply, this will be NoError (0)
+    Error error() const
+    {
+        return this->error_;
+    }
+
+    /// The HTTP status code if a request was made.
```
pajlada, line 36: if a response was received*

src/common/NetworkResult.hpp
```diff
@@ -2,14 +2,19 @@
...
 
+#include <optional>
+
 namespace chatterino {
 
 class NetworkResult
 {
 public:
-    NetworkResult(const QByteArray &data, int status);
+    using Error = QNetworkReply::NetworkError;
```
pajlada, line 15: We can use qt's phrasing here and call it NetworkError instead, I think that's a bit clearer as to what its purpose is

src/common/NetworkResult.hpp
```diff
@@ -20,13 +25,30 @@ class NetworkResult
...
+
+    /// The HTTP status code if a request was made.
+    std::optional<uint16_t> status() const
+    {
+        return this->status_;
+    }
+
+    /// Formats the error:
+    /// without HTTP status: [Name]
+    /// with HTTP status: [HTTP status]
```
pajlada, line 44: ...if a successful reply was received, or something, returns the http status
otherwise, returns the network error

src/singletons/Updates.cpp
```diff
@@ -128,8 +128,8 @@ void Updates::installUpdates()
                     auto *box = new QMessageBox(
                         QMessageBox::Information, "Chatterino Update",
                         QStringLiteral("The update couldn't be downloaded "
-                                       "(HTTP status %1).")
-                            .arg(result.status()));
+                                       "(%1).")
```
pajlada, line 131: add Error: prefix in

src/singletons/Updates.cpp
```diff
@@ -189,8 +189,8 @@ void Updates::installUpdates()
...
```
pajlada, line 192: add Error: prefix in

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -386,23 +386,17 @@ void SeventvEmotes::loadChannelEmotes(
...
                     qCWarning(chatterinoSeventv)
                         << "Error fetching 7TV emotes for channel" << channelId
-                        << ", error" << result.status();
-                    shared->addMessage(
-                        makeSystemMessage("Failed to fetch 7TV channel "
-                                          "emotes. (unknown error)"));
+                        << ", error" << errorString;
+                    shared->addMessage(makeSystemMessage(
+                        QStringLiteral("Failed to fetch 7TV channel "
+                                       "emotes. (%1)")
```
pajlada, line 398: add Error: prefix

src/providers/ffz/FfzEmotes.cpp
```diff
@@ -273,24 +273,17 @@ void FfzEmotes::loadChannel(
...
                 qCWarning(chatterinoFfzemotes)
                     << "Error fetching FFZ emotes for channel" << channelID
-                    << ", error" << result.status();
-                shared->addMessage(
-                    makeSystemMessage("Failed to fetch FrankerFaceZ channel "
-                                      "emotes. (unknown error)"));
+                    << ", error" << errorString;
+                shared->addMessage(makeSystemMessage(
+                    QStringLiteral("Failed to fetch FrankerFaceZ channel "
+                                   "emotes. (%1)")
```
pajlada, line 285: add Error: prefix

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -259,23 +259,17 @@ void BttvEmotes::loadChannel(std::weak_ptr<Channel> channel,
...
                 qCWarning(chatterinoBttv)
                     << "Error fetching BTTV emotes for channel" << channelId
-                    << ", error" << result.status();
-                shared->addMessage(
-                    makeSystemMessage("Failed to fetch BetterTTV channel "
-                                      "emotes. (unknown error)"));
+                    << ", error" << errorString;
+                shared->addMessage(makeSystemMessage(
+                    QStringLiteral("Failed to fetch BetterTTV channel "
+                                   "emotes. (%1)")
```
pajlada, line 271: add Error: prefix

src/providers/RecentMessagesApi.cpp
```diff
@@ -217,17 +217,19 @@ void RecentMessagesApi::loadRecentMessages(const QString &channelName,
...
                 return;
+            }
 
             qCDebug(chatterinoRecentMessages)
                 << "Failed to load recent messages for" << shared->getName();
 
             shared->addMessage(makeSystemMessage(
-                QString("Message history service unavailable (Error %1)")
-                    .arg(result.status())));
+                QStringLiteral("Message history service unavailable (%1)")
```
pajlada, line 231: add Error: prefix

pajlada (changes requested):

pajlada (approved):

PR #4711, opened 2023-07-01
Fix crash that could occur if closing the usercard quickly after blocking

pajlada:
# Description

- Fix crash that could occurr if closing the usercard quickly after blocking or unblocking
- Add unrelated `-DUSE_ALTERNATE_LINKER` cmake option


pajlada:
image

PR #4712, opened 2023-07-01
Add clang-tidy script that runs on all files in parallel

pajlada:
# Description

Extra options are forwarded to clang-tidy

For example: `./tools/check-clang-tidy.sh --checks '-*,readability-braces-around-statements' --fix`


pajlada:
Closing in favour of #5083 which adds the same-ish script & runs it with some fixes

PR #4713, opened 2023-07-01
Ensure live status requests are always batched

pajlada:
# Description

This makes sure live status requests are always batched, including the first one

Titles are also now always batched, and updated frequently rather than just randomly when we hover over the split header.

The display name of a channel is now also updated, if necessary, alongside the livestatus changes. Technically unrelated to the live status-ness but since we have that data in the same request, we might as well update it.


pajlada:
image

PR #4718, opened 2023-07-07
Add command to automatically reload your theme

Nerixyz:
# Description

Adds an option to watch the custom theme and reload it on changes. The option is only active for the current session (until Chatterino is restarted) to reduce load.

This PR does a few more things. I wasn't sure if I should include this here. I could split the commits up until 8f4770aa0ffc41e199d3ea29dc551efc1499d733 into a separate PR.


pajlada (changes requested):
Should be toggleable as a command rather than a UI thing imo, since it's an extremely small amount of people it's relevant for

Instead of using a FileWatcher, I would prefer the simplicity of a QTimer that runs every X ms (maybe every 500ms?) and reloads the current theme - while less performant, I think it's simpler to reason about

pajlada (approved):
:+1:

PR #4719, opened 2023-07-11
focus message input when selecting emotes in emote popup

LoneCoder21:
# Description
When selecting emotes in emote popup menu, it doesn't automatically focus the message input. This is slightly annoying as I would like to type something immediately afterwards.


pajlada (approved):
This gets a bit funky on tiling window managers on Linux, but it still works well.

Does anyone dislike this feature? If not, I will just merge this in next weekend

Felanbird (approved):
This seems fine, I'm not a emote menu user but it's not like we are taking focus from the emote menu search to implement this.

pajlada:

PR #4720, opened 2023-07-11
Change emotepopup channel view layout to be faster to navigate

LoneCoder21:
# Description

Currently, the emotepopup menu layout is a bit slow to navigate. This is especially true with global and channel emote views as they have all the bttv, ffz, 7tv emotes packed in and hence scrolling becomes long. I refactored it to so that it's more distinctive and scrolling is easier. Let me know.

Updated Look
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unversioned
-
+- Minor: Emote Popup Menu is now easier to navigate. (#4720)
 - Minor: Nicknames are now taken into consideration when searching for messages. (#4663)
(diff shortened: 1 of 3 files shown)


pajlada:
I don't think this is an upgrade in browsability, at least not in every case. It's going to depend on user preference in the end, and I think the search bar + grouping by channel/global emotes is a better choice for most users.

LoneCoder21:
> I don't think this is an upgrade in browsability, at least not in every case. It's going to depend on user preference in the end, and I think the search bar + grouping by channel/global emotes is a better choice for most users.

You do have a great point about this. Most users use search bar anyways. I will close this PR now.

PR #4721, opened 2023-07-11
Increase max number of blocked users loaded from 100 to 1,000

Nerixyz:
# Description

This PR does the following things:

* Remove `SharedAccessGuard` - all accesses to the sets are on the GUI thread.
* Move storage to a hash-set (`QSet`)
* Add generic Helix pagination API
* Use pagination when loading blocks

Currently, all blocks are loaded. In https://github.com/Chatterino/chatterino2/issues/2767 it was mentioned, that only a certain amount of pages should be loaded. This is easy through the new cancellation token. I left this unimplemented, as there's no clear value for the limit.

If you want to test this and don't have more than 100 blocked users, you can set the `first` to a value lower than 100 to emulate loading multiple pages.

Closes https://github.com/Chatterino/chatterino2/issues/2767.


src/providers/twitch/TwitchAccount.cpp
```diff
@@ -100,79 +101,69 @@ bool TwitchAccount::isAnon() const
...
         },
         std::move(onFailure));
 }
 
-SharedAccessGuard<const std::set<TwitchUser>> TwitchAccount::accessBlocks()
-    const
+const QSet<TwitchUser> &TwitchAccount::blocks() const
 {
-    return this->ignores_.accessConst();
+    return this->ignores_;
```
pajlada, line 161: Could you add a ThreadGuard to TwitchAccount.hpp called "blockThreadGuard" and call its `guard()` function here and in `blockedUserIds()`?

src/providers/twitch/TwitchAccount.hpp
```diff
@@ -105,10 +107,11 @@ class TwitchAccount : public Account
...
     Atomic<QColor> color_;
 
-    mutable std::mutex ignoresMutex_;
     QStringList userstateEmoteSets_;
-    UniqueAccess<std::set<TwitchUser>> ignores_;
-    UniqueAccess<std::set<QString>> ignoresUserIds_;
+
+    ScopedCancellationToken blockToken_;
+    QSet<TwitchUser> ignores_;
+    QSet<QString> ignoresUserIds_;
```
pajlada, line 114: Could we make these `std::set/std::unordered_set` instead? That way we don't have to mess with extra hashing for TwitchUser

src/providers/twitch/TwitchUser.hpp
```diff
@@ -3,13 +3,20 @@
...
 
 #include <cassert>
 
 namespace chatterino {
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+#    define QHashValue uint
+#else
+#    define QHashValue size_t
+#endif
```
pajlada, line 18: should be removable once the last comment has been resolved (switching out QSet for std::set/std::unordered_set)

src/util/CancellationToken.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    void cancel()
+    {
+        if (this->isCanceled_ != nullptr)
+        {
+            this->isCanceled_->store(true, std::memory_order_release);
+        }
+    }
+
+    bool isCanceled() const
```
pajlada, line 36: Should be `isCancelled` to be consistent with US/Bri'ish spelling - same goes for the variable

src/providers/twitch/api/Helix.cpp
```diff
@@ -544,40 +548,30 @@ void Helix::createStreamMarker(
...
-            if (!data.isArray())
-            {
-                failureCallback();
-                return Failure;
-            }
+    this->paginate(
+        u"users/blocks"_s,
+        {{u"broadcaster_id"_s, userId}, {u"first"_s, u"100"_s}},
+        [pageCallback](const QJsonObject &json) {
+            const auto data = json["data"_L1].toArray();
```
pajlada, line 559: If the array is empty (i.e. it wasn't a valid array, or it was actually an empty array), make this function return something to tell paginate to stop paginating

src/providers/twitch/api/Helix.cpp
```diff
@@ -2931,6 +2925,47 @@ NetworkRequest Helix::makePatch(const QString &url, const QUrlQuery &urlQuery)
...
+        const auto json = res.parseJson();
+        onPage(json);
+
+        const auto pagination = json["pagination"_L1].toObject();
+        if (pagination.empty() || !pagination.contains("cursor"_L1))
+        {
+            return Success;
+        }
+        query.removeAllQueryItems(u"after"_s);
+        query.addQueryItem(u"after"_s, pagination["cursor"_L1].toString());
```
pajlada, line 2953: Just fetch the cursor, and return Success if cursor is not a string - maybe not as performant but more readable imo

src/util/CancellationToken.hpp
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+#include <atomic>
+#include <memory>
+
+namespace chatterino {
+
+class CancellationToken
```
pajlada, line 8: some mini comment
```suggestion
/// The CancellationToken is a thread-safe way for worker(s) to know if the task they want to continue doing should be cancelled
class CancellationToken
```

src/util/CancellationToken.hpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return this->isCanceled_ == nullptr ||
+               this->isCanceled_->load(std::memory_order_acquire);
+    }
+
+private:
+    std::shared_ptr<std::atomic<bool>> isCanceled_;
+};
+
+class ScopedCancellationToken
```
pajlada, line 46: some mini comment
```suggestion
/// The ScopedCancellationToken is a way to automatically cancel a CancellationToken when it goes out of scope
class ScopedCancellationToken
```

pajlada (changes requested):
Small nitpicks, overall looks good :+1: The paginate function will be useful for future Helix requests

pajlada (approved):
Nice!

PR #4727, opened 2023-07-19
Limit sccache caches and improve PCH

Nerixyz:
# Description

* Limits creation of caches created by `sccache` to the default branch only. PRs should use that cache when running their builds.
* Add `signals2` (used in various places) and `circular_buffer` (used in LimitedQueue) to PCH (found through `vcperf`)
* Remove duplicate includes from PCH (ones with `QtWidgets/...` or `QtCore/...`)
* Remove barely used includes (found through grepping for them)

This should hopefully speed up builds a bit.


.github/workflows/build.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         plugins: [false]
         skip-artifact: [false]
         skip-crashpad: [false]
+        skip-pch: [false]
```
pajlada, line 35: I'd rather this be `enable-pch` with the default value being `false` since we only want PCH enabled for 2 of the many jobs

pajlada (changes requested):

pajlada (approved):

PR #4731, opened 2023-07-23
Fix crash that could occur when closing a split before its display name was updated

pajlada:
# Description

- Disconnect the `displayNameChanged` signal when the split closes
- Store the "raw" display name sent from Twitch


pajlada (commented):
Looks good to me :+1: 
image

Issue #3101, opened 2021-08-01
Emote/user tab completion completes using wrong word


jupjohn:
**Describe your issue**
Tab completion of emotes/usernames can also complete when moving to prior word.

*Steps to reproduce*
 1. Type a single word
 2. Type another word and tab complete to an emote/user
 3. Press left arrow until you are at the last character of the first word
 4. Press tab again

*Expected behaviour*
Previous word tries to complete using the word

*Observed behaviour*
Previous word thinks it's the second word so tried to complete using those characters

**Screenshots**
https://user-images.githubusercontent.com/4962764/127768320-e14afc52-4b84-4ff3-ab03-c604c6f248e0.mp4

**OS and Chatterino Version**
Arch Linux (N/A), Chatterino 2.3.3 (commit ae9f92de)

This is a hard one to explain (and title), hope the vid does it justice.

Mm2PL:
Can reproduce on ae9f92de

xHeaveny:
can also reproduce on `3cb1e51`

M4tthewDE:
I did some investigation and it looks like the completer doesn't respect changing the cursor.
This happens as well if the cursor is changed with the mouse or the entire content of the textbox with up-arrow/down-arrow.
This issue could probably be solved by adding a check inside of `ResizingTextEdit::keyPressEvent` for those specific events.

PR #4735, opened 2023-07-27
Fixed tab completion rarely completing the wrong word.

Mm2PL:
# Description

Now moving the cursor in any way will result in the completion being cancelled.

Fixes: #3101

### DIFF
--- a/src/widgets/helper/ResizingTextEdit.cpp
+++ b/src/widgets/helper/ResizingTextEdit.cpp
@@ -7,6 +7,7 @@
 
 #include <QMimeData>
 #include <QMimeDatabase>
+#include <QObject>
 
 namespace chatterino {
 
@@ -20,8 +21,12 @@ ResizingTextEdit::ResizingTextEdit()
 
     QObject::connect(this, &QTextEdit::textChanged, this,
                      &QWidget::updateGeometry);
+
     QObject::connect(this, &QTextEdit::cursorPositionChanged, [this]() {
-        if (this->updatingText_ || !this->completionInProgress_)
+        // If tab was pressed and we're completing/replacing the current word,
+        // this code will not even be called, see ResizingTextEdit::keyPressEvent
+
+        if (!this->completionInProgress_)
         {
             return;
         }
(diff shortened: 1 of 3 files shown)


src/widgets/helper/ResizingTextEdit.hpp
```diff
@@ -66,6 +66,21 @@ class ResizingTextEdit : public QTextEdit
      */
     bool completionInProgress_ = false;
 
+    /**
+     * This is set whenever the tab completion is triggered.
+     * When this variable is set, moving the cursor will not reset tab completion.
```
Nerixyz, line 71: I find this comment confusing, because there's `completionInProgress_` which seems to do a similar thing (from the comments' perspective). It should include that this is immediately unset and used to ignore cursor-position-change-events ("moving the cursor" suggests user-input).

src/widgets/helper/ResizingTextEdit.cpp
```diff
@@ -158,6 +169,7 @@ void ResizingTextEdit::keyPressEvent(QKeyEvent *event)
             this->completionInProgress_ = true;
             this->completer_->setCompletionPrefix(currentCompletionPrefix);
             this->completer_->complete();
+            this->updatingText_ = false;
```
Nerixyz, line 172: This could use a signal blocker:

```suggestion
            {
                // block cursor position events
                QSignalBlocker b(this);
                this->completer_->complete();
            }
```

src/widgets/helper/ResizingTextEdit.cpp
```diff
@@ -183,6 +195,7 @@ void ResizingTextEdit::keyPressEvent(QKeyEvent *event)
         }
 
         this->completer_->complete();
+        this->updatingText_ = false;
```
Nerixyz, line 198: Same here.

src/widgets/helper/ResizingTextEdit.hpp
Mm2PL: Said comment has been removed for QSignalBlocker

Nerixyz (approved):

pajlada (approved):

Issue #4146, opened 2022-11-13
`/watching` Tab Parts Channels Even Though Browser Tab Is Still Open


Nerixyz:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When opening two (or more) streams in the browser and switching between them, the channel used in the `/watching` tab is updated. If one of the channels isn't open in any split, Chatterino needs to rejoin the channel and load recent-messages and emotes. This can take some time and puts load on recent-messages and the emote-providers, especially when switching quickly. It would be nice if the switch was quicker. 

My idea is that Chatterino could avoid parting a channel for some time if it was displayed in `/watching` or even sync the open tabs between the extension and Chatterino.

_Non-Windows users should be able to test this as well._

### OS and Chatterino Version

Chatterino Nightly 7.3.5 (commit f2f18ba5 modified) built on 2022-10-16 with Qt 5.15.2, Windows SDK, MSVC 193331630

LosFarmosCTL:
I don't really have experience with web extensions, but it should be possible to get a list of open tabs, right?
So maybe it would make sense to keep a chat loaded as long as the corresponding browser tab is still opened. Not sure if there are maybe events for closing a tab, or how that would be implemented but seems like a sensible way to do this if possible.

PR #4738, opened 2023-07-28
Refactor Native Messages

Nerixyz:
# Description

In order to implement https://github.com/Chatterino/chatterino2/issues/4146, I wanted to first refactor the native-message-handler. The commits can be reviewed on their own. Some high level changes:

* Moved `boost::interprocess` interactions to new class.
* Created dedicated handlers for message types.
* Resolved clang-tidy warnings.
* Added CMake option to debug native messages.

### DIFF
--- a/src/util/IpcQueue.cpp
+++ b/src/util/IpcQueue.cpp
@@ -39,7 +39,6 @@ public:
 
 IpcQueue::IpcQueue(IpcQueuePrivate *priv)
     : private_(priv){};
-IpcQueue::IpcQueue(IpcQueue &&) = default;
 IpcQueue::~IpcQueue() = default;
 
 std::variant<IpcQueue, QString> IpcQueue::tryReplaceOrCreate(
--- a/src/util/IpcQueue.cpp
+++ b/src/util/IpcQueue.cpp
@@ -37,22 +37,18 @@ public:
     boost_ipc::message_queue queue;
 };
 
-IpcQueue::IpcQueue() = default;
+IpcQueue::IpcQueue(IpcQueuePrivate *priv)
+    : private_(priv){};
+IpcQueue::IpcQueue(IpcQueue &&) = default;
 IpcQueue::~IpcQueue() = default;
 
-std::optional<QString> IpcQueue::tryReplaceOrCreate(const char *name,
-                                                    size_t maxMessages,
-                                                    size_t maxMessageSize)
+std::variant<IpcQueue, QString> IpcQueue::tryReplaceOrCreate(
+    const char *name, size_t maxMessages, size_t maxMessageSize)
 {
     try
     {
-        Q_ASSERT_X(this->private_ == nullptr, "IpcQueue::tryReplaceOrCreate",
-                   "The function can be called at most once.");
-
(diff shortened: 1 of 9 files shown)


src/util/IpcQueue.cpp
```diff
@@ -0,0 +1,95 @@
...
+class IpcQueuePrivate
+{
+    IpcQueuePrivate(IpcQueue *q_ptr, const char *name, size_t maxMessages,
+                    size_t maxMessageSize)
+        : q_ptr(q_ptr)
+        , queue_(boost_ipc::open_or_create, name, maxMessages, maxMessageSize)
+    {
+    }
+
+    Q_DECLARE_PUBLIC(IpcQueue)
```
pajlada, line 38: Replace with `friend class IpcQueue`

src/util/IpcQueue.cpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return QString::fromLatin1(ex.what());
+    }
+}
+
+QByteArray IpcQueue::receive()
+{
+    try
+    {
+        Q_D(IpcQueue);
```
pajlada, line 71: Replace with `auto *d = ...`

src/util/IpcQueue.hpp
```diff
@@ -0,0 +1,35 @@
...
+                                              size_t maxMessageSize);
+
+    // TODO: use std::expected
+    /// Try to receive a message.
+    /// `tryReplaceOrCreate` must have been called before.
+    /// In the case of an error, the buffer is empty.
+    QByteArray receive();
+
+private:
+    QScopedPointer<IpcQueuePrivate> d_ptr;
```
pajlada, line 31: replace with a unique pointer & rename it or fix the casing. maybe just call it `impl`?

src/util/IpcQueue.hpp
```diff
@@ -0,0 +1,35 @@
...
+    Q_DECLARE_PRIVATE(IpcQueue)
```
pajlada, line 32: remove this, as it was only necessary for the `Q_D` macro

pajlada (changes requested):
Just nitpicks

src/util/IpcQueue.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+class QByteArray;
+class QString;
+
+namespace chatterino::ipc {
+
+void sendMessage(const char *name, const QByteArray &data);
+
+class IpcQueuePrivate;
+class IpcQueue
```
pajlada, line 14: It would be better if IpcQueue could only have a successful state.
We can achieve this by not exposing the constructor, but only a "creator" like the `tryReplaceOrCreate` function, that returns a unique_ptr to an IpcQueue. It will just need to return a nullptr if it failed.

As we talked about in chat, maybe

```cpp
std::pair<std::unique_ptr<IpcQueue>, QString>
```

so it can be used like this:

```c++
auto [messageQueue, messageQueueError] = IpcQueue::create(...);
if (!messageQueue) {
  // handle error and print the error or something
}
```

pajlada (approved):

PR #4741, opened 2023-07-31
Sync channels with browser

### LINKED ISSUE #4146: `/watching` Tab Parts Channels Even Though Browser Tab Is Still Open

Nerixyz:
# Description

In conjunction with https://github.com/Chatterino/chatterino-browser-ext/pull/95.

This PR warms channels that are open in the browser, which results in a better experience when switching tabs in the browser (described in more detail in #4146). Channels that were already open will only have their reference-count incremented.

Fixes #4146.


src/singletons/NativeMessaging.cpp
```diff
@@ -263,6 +279,39 @@ void NativeMessagingServer::ReceiverThread::handleDetach(
 }
 // NOLINTEND(readability-convert-member-functions-to-static)
 
+void NativeMessagingServer::ReceiverThread::handleSync(const QJsonObject &root)
+{
+    // Structure:
+    // { action: 'sync', twitch?: string[] }
```
pajlada, line 285: Let's be explicit with what's being synced, rename the `twitch` key to `twitchChannels` or `twitch_channels` (whichever standard is used in the other json fields)

src/singletons/NativeMessaging.hpp
```diff
@@ -26,21 +31,38 @@ namespace nm::client {
...
+        void handleSync(const QJsonObject &root);
+
+        NativeMessagingServer &parent_;
     };
 
+    void syncChannels(const QJsonArray &twitchChannels);
+
     ReceiverThread thread;
+
+    std::vector<ChannelPtr> channelWarmer_;
```
pajlada, line 63: Small comment explaining that this joins channels the user of the extension has open

pajlada (changes requested):
Small nitpicks, will require changes in the extension PR too

pajlada (approved):

PR #4745, opened 2023-08-02
Add support for opening links in incognito mode on Linux & BSD

sheybey:
# Description

This PR adds support for opening links in private/incognito mode under Linux/BSD when using a desktop environment that follows the freedesktop specification. See https://github.com/Chatterino/chatterino2/discussions/4366

## List of technical changes

- Add private switches for firefox-esr and chromium
- Get only the browser executable name instead of the full command line on windows (to match new behavior under linux)
- Implement XDG specs for reading desktop files and finding default applications
- ~~ego check (add myself to contributors.txt)~~

## Discussion points

I originally thought that you could read desktop files, which are essentially .ini files, with QSettings. However, the default app database (which is a desktop file) uses mimetypes as key names, which is incompatible with QSettings as the key may include `/`. I wrote a parser for them which didn't end up bigger than the code consuming it, so I feel comfortable still not adding another dependency for this. That being said, it should probably get as many eyes on it as possible since it is still a home-rolled parser.

Some other points I would appreciate opinions on:
- XDGDesktopFile: is returning a static empty group for non-existent groups the right move here? std::optional instead?
- IncognitoBrowser.cpp: I didn't change too much of this implementation, which means that `getDefaultBrowserExecutable` is called twice per settings open. Is a big search like this PR adds worth caching?
- The XDG spec allows for multiple defaults listed by preference. This implementation just returns the first one, and if chatterino doesn't know how to open it in private mode, it gives up immediately. Could try to fall back to the next default if there is one, but this might end up seemingly ignoring whatever the user set in their DE if we don't support it.

I've tested this under Windows (for the minor behavior change) and under Debian testing with gnome. Throw your hairiest setups at this and let me know if anything breaks.


resources/contributors.txt
```diff
@@ -65,6 +65,7 @@ ZonianMidian | https://github.com/ZonianMidian | :/avatars/zonianmidian.png | Co
 olafyang | https://github.com/olafyang | | Contributor
 chrrs | https://github.com/chrrs | | Contributor
 4rneee | https://github.com/4rneee | | Contributor
+sheybey | https://github.com/sheybey | | Contributor
```
Felanbird, line 68: separate PR

sheybey: ok, I will remove this commit when I rebase on current master

pajlada (commented):
Looked through everything except XDGHelper so far, other than some nitpicks that I've fixed myself it looks fine :+1:

pajlada (approved):
Works as expected for me now. If you can think of some other tests, please add them.
Please also read through the commits I've done and see if there's anything I've misunderstood

When you're happy to have this merged in, let us know and we'll hit the merge button

sheybey:
> Works as expected for me now. If you can think of some other tests, please add them. Please also read through the commits I've done and see if there's anything I've misunderstood

Done and done. Looks good to me

Issue #1844, opened 2020-08-08
Add the ability to control the popup chat font size separately.


Danis-Enikeev:
**What should be added?**
A setting that would allow resizing the font/emote size in "Popup" chat window without affecting the main chatterino window.

**Why should it be added?**
Right now in order to have a window of chatterino with different font size you have to open a second instance of chatterino and change font size there like shown on the screenshot below:

![image](https://user-images.githubusercontent.com/43012453/89711827-5a815000-d995-11ea-9b76-3b74661bd831.png)
It would be much easier to do if the font size of the popup window could be controlled separately. This can be very useful with a dual monitor setup, when monitors have different DPI values.

PR #4746, opened 2023-08-02
Add transparent overlay window

### LINKED ISSUE #1618: Transparent background on program
I'd like to get a transparent background on chat so I can have vlc windowed or borderless with streamlink and have the chat be on top of vlc. (Imagine chat with theatre mode on phone)
I'm wondering if it's possible, thanks

Nerixyz:
# Description

This PR adds a hotkey to open a transparent overlay channel-view. The colors of the overlay can be customized in the theme. To move, resize, and close the overlay, you have to press the <kbd>SHIFT</kbd> key.

This is an early feature and requires more PRs in the future to support more stuff like scaling.

### TODO

- [x] Wait for #4868
- [x] Add dialog explaining functionality after first time opening.
- [x] Test `/watching`.
- [x] Fix split lifetime (https://github.com/Chatterino/chatterino2/pull/4747)
- [x] Check theme colors
- [x] Fix cursor
- ~~Add minimize button~~ If that's desired, it should be done in a future PR.
- [x] Maybe use `NCHITTEST`?
- [x] Draw background behind buttons when <kbd>SHIFT</kbd> is held.
- [x] Support click-through
- [x] Support multiple click-through windows
- [x] Add [optional] shadow (maybe future PR?)
- [ ] High DPI tests - _I don't think these are really needed, as Qt handles that for us now._
- ~~Update theme migrations (wiki)~~ Not needed as they will use a fallback
- Support scaling (independent - see also: https://github.com/Chatterino/chatterino2/issues/1844) - _Future PR_
- [x] Check margins (and interaction ~~overlay~~ underlay)
- Maybe some transparency changes can be in a separate PR?
- Support blur behind/acrylic/... [mica?] → In another PR - https://github.com/tauri-apps/window-vibrancy/ provides a good foundation

https://github.com/Chatterino/chatterino2/assets/19953266/3e798169-a82f-4123-b834-7ce55daf0fc1

Link to the video in case the embed is broken: `https://github.com/Chatterino/chatterino2/assets/19953266/3e798169-a82f-4123-b834-7ce55daf0fc1`

Closes #1618.


RangerFrank:
I've been using https://github.com/baffler/Transparent-Twitch-Chat-Overlay for the functionality that this PR is going to add to Chatterino, very glad to see it finally happening. One major feature that is an oversight for many chat overlays is that it needs to have toggleable click-through, so that the widget can be sized and moved, but when toggled won't be able to be clicked or interfere with mouse pointer tracking in games and other programs.

pajlada:
Looks great, I'd like to see the interactability toggleable as part of a Chatterino hotkey instead of a Global Shortcut.

If the need arises for speed, we can re-investigate Global Shortcuts

src/singletons/WindowManager.hpp
```diff
@@ -128,6 +128,9 @@ class WindowManager final
     // again
     void queueSave();
 
+    /// Toggles the inertia in all open popup windows
```
pajlada, line 131:
```suggestion
    /// Toggles the inertia in all open overlay windows
```

src/widgets/splits/Split.cpp
```diff
@@ -765,6 +766,47 @@ void Split::addShortcuts()
...
+                 {
+                     getApp()->getWindows()->toggleAllPopupInertia();
+                 }
+                 return {};
+             }
+             if (arg == "all")
+             {
+                 getApp()->getWindows()->toggleAllPopupInertia();
+                 return {};
+             }
```
pajlada, line 807: I think the default keybind action should be `all`, and I think `toggleAllPopupIntertia` should not toggle each overlay individually, rather think of all of them as a group.
If any overlay is interactable, make all overlays inert. Otherwise, make all overlays interactable.

src/widgets/helper/ChannelView.cpp
```diff
@@ -601,6 +606,24 @@ void ChannelView::themeChangedEvent()
     this->setupHighlightAnimationColors();
     this->queueLayout();
     this->messageColors_.applyTheme(getTheme());
+    if (this->colorVisitor_)
+    {
+        this->colorVisitor_(this->messageColors_, getTheme());
+    }
```
pajlada, line 612: Could `applyTheme` on MessageColors accept the `overlayBackgroundOpacity` plus a boolean whether this is an overlay enough, and let it do the logic for how to apply the colors instead?

so `MessageColors::applyTheme(Theme*)` would instead be `MessageColors::applyTheme(Theme*, bool isInOverlay, int backgroundOpacity`) - that should let us get rid of the visitor pattern

pajlada (commented):
Some small nits

Nerixyz:
75e3dbfea5ff2fd8e7e78c1db76b5360c34baa09 is a bit bigger, as I didn't use the correct colors from the theme for the text (it used the "regular" ones instead of the ones for the overlay). Light themes don't look as good as dark themes, as the text colors are "normalized". We should do that a bit differently, but that's for a future PR imo.

src/widgets/OverlayWindow.cpp
```diff
@@ -0,0 +1,529 @@
...
+            {
+                this->channelView_.setGraphicsEffect(nullptr);
+                this->dropShadow_ = nullptr;  // deleted by setGraphicsEffect
+            }
+            this->applyTheme();
+        },
+        this->holder_);
+    settings->overlayBackgroundOpacity.connect(
+        [this] {
+            this->channelView_.updateColorTheme();
```
pajlada, line 151: Missing a call to `update` (or something similar) to ensure the opacity updates as it's being changed
```suggestion
            this->channelView_.updateColorTheme();
            this->update();
```

pajlada (changes requested):
Just found one small nit - not committing it myself in case there's a better way to do it

pajlada (approved):

PR #4747, opened 2023-08-03
Allow `ChannelView`s without a `Split`

Nerixyz:
# Description

This PR changes the references to splits kept by `ChannelView`s, `ReplyThreadPopup`s, and `UserInfoPopup`s to be a `QPointer<Split>`. This way, they can't contain dangling pointers to the original `Split`s. In turn, the following UB and potential crashes are prevented:

- Using `Go to Message` in a user card/reply thread popup after the original split has been deleted. 
- Sending messages in a reply thread after the original split has been deleted. (Try replacing the `Split*` in `SplitInput` with a `QPointer<Split>` if there's no crash)

This in turn makes it possible to create user cards and reply thread popups _without_ a split. Also required for https://github.com/Chatterino/chatterino2/pull/4746.


src/widgets/splits/SplitInput.cpp
```diff
@@ -76,6 +76,13 @@ SplitInput::SplitInput(QWidget *parent, Split *_chatWidget,
                                        });
 }
 
+SplitInput::~SplitInput()
+{
+    // shortcuts are registered on the parent widget,
+    // but capture this widget.
+    this->clearShortcuts();
```
pajlada, line 83: Is this change still relevant?

src/widgets/helper/ChannelView.hpp
```diff
@@ -75,7 +76,23 @@ class ChannelView final : public BaseWidget
         Search,
     };
 
-    explicit ChannelView(BaseWidget *parent = nullptr, Split *split = nullptr,
+    /// Creates a channel view without a split.
+    /// In such a channel, usercards and reply-threads can't be opened.
```
pajlada, line 80:
```suggestion
    /// In such a view, usercards and reply-threads can't be opened.
```

src/widgets/helper/ChannelView.hpp
```diff
@@ -75,7 +76,23 @@ class ChannelView final : public BaseWidget
...
+    /// @param parent The parent of this widget. Can be `nullptr`.
+    /// @param context The context in which this view is shown (e.g. as a usercard).
+    /// @param messagesLimit The maximum amount of messages this view will display.
+    explicit ChannelView(QWidget *parent, Context context = Context::None,
+                         size_t messagesLimit = 1000);
+
+    /// Creates a channel view in a split.
+    ///
+    /// @param parent The parent of this widget.
+    /// @param split The split contianing this widget.
```
pajlada, line 91:
```suggestion
    /// @param split The split containing this widget.
```

src/widgets/helper/ChannelView.hpp
```diff
@@ -75,7 +76,23 @@ class ChannelView final : public BaseWidget
...
+    ///              @a split must be in the widget tree of @a parent.
```
pajlada, line 92: what are the `@`'s for here? some docstring magic?

src/widgets/helper/ChannelView.hpp
```diff
@@ -12,6 +12,7 @@
 #include <pajlada/signals/signal.hpp>
 #include <QMenu>
 #include <QPaintEvent>
+#include <QPointer>
```
pajlada, line 15: Remove unused include
```suggestion
```

src/widgets/helper/ChannelView.hpp
```diff
@@ -192,6 +209,12 @@ class ChannelView final : public BaseWidget
                          QPoint &relativePos, int &index);
 
 private:
+    struct InternalCtor {
+    };
+
+    ChannelView(InternalCtor tag, QWidget *parent, Split *split,
+                Context context = Context::None, size_t messagesLimit = 1000);
```
pajlada, line 216: We can remove all the default values here
```suggestion
    ChannelView(InternalCtor tag, QWidget *parent, Split *split,
                Context context, size_t messagesLimit);
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -273,12 +273,30 @@ std::pair<int, int> getWordBounds(MessageLayout *layout,
...
 
-ChannelView::ChannelView(BaseWidget *parent, Split *split, Context context,
+ChannelView::ChannelView(QWidget *parent, Context context, size_t messagesLimit)
+    : ChannelView(InternalCtor{}, parent, nullptr, context, messagesLimit)
+{
+}
+
+ChannelView::ChannelView(QWidget *parent, Split *split, Context context,
                          size_t messagesLimit)
+    : ChannelView(InternalCtor{}, requireNotNull(parent), requireNotNull(split),
```
pajlada, line 293: I would prefer if the asserts for these params are made in the constructor itself, without the `requireNotNull` function

Technically they're asserting too late, but since both values are allowed to be null by the internal ctor, that should be fine, and the assert will fire with a little bit better info

pajlada (commented):
Just some small nitpicks, overall looks good :+1:

src/widgets/helper/ChannelView.hpp
Nerixyz: It emphasizes the word ([docs](https://www.doxygen.nl/manual/commands.html#cmda); clangd doesn't display that right now).

src/widgets/splits/SplitInput.cpp
Nerixyz: It's still wrong, but a `SplitInput` is never destroyed without the parent (reply-thread or split).

pajlada (approved):

PR #4475, opened 2023-03-25
Populate mentions tab with message history

2547techno:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

An implementation of #4474 

Ended up just adding to mentions tab when message history is loaded from joining a new channel rather than just on startup.
Messages preserve their chronological order in the mentions tab, so opening a channel that you have a mention in will not necessary place that message at the bottom of the split.

PR #4752, opened 2023-08-05
Highlights loaded from message history will now correctly appear in the /mentions channel

pajlada:
# Description

Originally https://github.com/Chatterino/chatterino2/pull/4475


pajlada:
image

PR #4758, opened 2023-08-07
Minor: AutoMod term management messages are now hidden in Streamer Mode.

Mm2PL:
# Description

If you add or remove a blocked term, it no longer gets shown in streamer mode.


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -596,8 +596,9 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addCheckbox(
         "Hide viewer count and stream length while hovering over split header",
         s.streamerModeHideViewerCountAndDuration);
-    layout.addCheckbox("Hide moderation actions", s.streamerModeHideModActions,
-                       false, "Hide bans & timeouts from appearing in chat.");
+    layout.addCheckbox(
+        "Hide moderation actions", s.streamerModeHideModActions, false,
+        "Hide bans, timeouts and automod messages from appearing in chat.");
```
Felanbird, line 601: This is not technically a good description of what this PR did, BUT this setting already did hide automod messages so now the tooltip is more correct than it was before, and automod messages I believe is blanket enough to cover what we changed here.

Felanbird (approved):
works 👍

src/widgets/settingspages/GeneralPage.cpp
Felanbird:
```suggestion
        "Hide bans, timeouts, and automod messages from appearing in chat.");
```
actually oxford would be mad at us if we didn't

pajlada (approved):
The change itself isn't major, but it may be a very visible change for streamers so I've elaborated on the changelog entry and moved it to be a Major change.
Alternatively, it could be a Minor change but left at the top instead.

I'm happy to have this merged in whenever someone else confirms my change

PR #4760, opened 2023-08-10
Show historic timeouts and bans in usercard

### LINKED ISSUE #3030: Usernames in historic bans / timeouts aren't clickable.
**Describe your issue**

Bans / timeouts loaded from recent-messages aren't processed the same way as from chat so they don't have usercard links added.

**Expected behavior**

You can click usernames from bans in historic messages.

**Actual behavior**

You can't

**OS and Chatterino Version**

Chatterino + Dankerino patches 2.3.3 4cc402af9027d0d9e4889673fa5bbcc0f24ce3c4 upstream faafbdaa

Nerixyz:
# Description

This PR fixes timeouts and bans of users not being shown in their usercard and timeouts not stacking. Previously, `CLEARCHAT`s were converted to `NOTICE`s. This was removed and `CLEARCHAT`s are handled.

The fix is a bit more involved, as the `addOrReplaceMessage` logic now adapts to any buffer. I wasn't very sure where to put the shared logic, so I chose `ChannelHelpers` (analogous to `IrcHelpers`).

Fixes #3030.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - Bugfix: Fixed crash that could occurr when closing the usercard too quickly after blocking or unblocking a user. (#4711)
 - Bugfix: Fixed highlights sometimes not working after changing sound device, or switching users in your operating system. (#4729)
 - Bugfix: Fixed key bindings not showing in context menus on Mac. (#4722)
+- Bugfix: Fixed timeouts from history messages not behaving consistently. (#4760)
 - Bugfix: Fixed tab completion rarely completing the wrong word. (#4735)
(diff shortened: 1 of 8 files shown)


pajlada (approved):
This change is fine to me - it was not difficult to test.

One quirk it has that I documented in ChannelHelpers is that timeouts can be merged into historic message timeouts even though they technically wouldn't if that timeout was a normal message, since we rely on the parse time of a message to know if we should merge into it.

I'm OK with the implementation as-is since it would require a pretty heavy lift to fix this issue.

Let me know if you are comfortable having this PR merged in

pajlada (approved):

PR #4763, opened 2023-08-12
Refactor recent messages API

pajlada:
# Description

This exposes internal functions for testing by splitting the implementation & internal API into separate files


pajlada:
image

PR #4715, opened 2023-07-02
Add subtitle to Hype Chats

Nerixyz:
# Description

This PR adds a subtitle to Hype Chats. It contains the level, the duration, and the amount of money spent. The money is currently formatted according to the system locale.

**Example:**
chatterino_2023-07-02_17-52-03

PR #4766, opened 2023-08-13
Change `Elevated Messages` to `Hype Chats`

Felanbird:
# Description
Updated user-facing reference & added relevant search/filter options under this new name.

Relies on #4765 as I'll be changing the #4715 changelog
```
- - Minor: Added better support for Twitch's Hype Chat feature. (#4715)
+ - Minor: Added currency & duration to Hype Chat messages. (#4715)
+ - Minor: Added `is:hype-chat` search option. (#4766)
+ - Minor: Added `flags.hype_chat` filter variable. (#4766)
```

image
image


src/controllers/filters/lang/Tokenizer.hpp
```diff
@@ -25,6 +25,7 @@ static const QMap<QString, QString> validIdentifiersMap = {
     {"flags.reward_message", "channel point reward message?"},
     {"flags.first_message", "first message?"},
     {"flags.elevated_message", "elevated message?"},
+    {"flags.hype_chat", "elevated message?"},
```
Nerixyz, line 28: Shouldn't this be `hype chat message?` (and the `flags.elevated_message` as well)?

Felanbird: probably, I just figured it copied the existing information

Nerixyz: Ideally the values should be unique, but I don't really see a nice way here. Since currently, the decision whether `flags.hype_chat` or `flags.elevated_message` is used in `ChannelFilterEditorDialog::ValueSpecifier::expressionText` depends on `QMap::key` (`std::map` layout in Qt 6 and internal implementation in Qt 5).

pajlada: @Felanbird if you could add a comment mentioning ^ above the hype_chat entry that'd be nice

pajlada (approved):
Resolve the PR conversation, fix the changelog entry, then you should be good to merge :+1: 
LGTM

PR #4771, opened 2023-08-19
Fix data race in PubSub

pajlada:
# Description
This ensures the close call is posted to the io_service executor
This fixes a data race with the get_con_from_hdl function


pajlada:
image

PR #4776, opened 2023-08-20
Remove boost::noncopyable use & boost::random dependency

pajlada:
# Description

This also removes the move ctor/assignment operator where applicable


pajlada:
image

PR #4791, opened 2023-08-26
Encode `modes` file in UTF8

Nerixyz:
# Description

I noticed that PowerShell 5 (which is the version used in CI) encodes files using UTF16-LE when doing `echo foo > xd`. This doesn't happen in PowerShell 7 which I used for local testing. For the built installer, that shouldn't change anything, since it ignores lines that contain junk and the release `modes` file is empty.

Since using `Out-File` with `-Encoding Utf8` will result in a file encoded in UTF8 with BOM, I used `ASCII` (which is fine here).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Bugfix: Fixed UTF16 encoding of `modes` file for the installer. (#4791)
+- Dev: Fixed UTF16 encoding of `modes` file for the installer. (#4791)
 
 ## 2.4.5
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed UTF16 encoding of `modes` file for the installer. (#4791)
```
pajlada, line 5: Let's make this a dev entry instead

pajlada (changes requested):

pajlada (approved):

PR #4795, opened 2023-08-27
Add (invisible) resize handle to frameless usercards and reply threads

Nerixyz:
# Description

Frameless usercards and reply threads can't be resized currently. This PR adds a resize handle in the bottom right corner.

This might only be an issue on Windows, so I'm fine with having this only added there.


pajlada (approved):
Works as expected for me and looks good to me :+1: 
Will merge in manually later with a changelog entry

PR #4807, opened 2023-09-07
Use `QFontMetrics::elidedText` over manual implementation

Nerixyz:
# Description

There is a major performance issue when displaying replies to a message with certain Arabic characters (measuring the width for them seems to be a slow path). This PR attempts to resolve parts of that issue. 

<details><summary>Reproduction</summary>

If you're running a debug build, do the following:

1. Open `!!!` as a new split
2. Run the following command a few times:
   ```
   /fakemsg @reply-parent-user-login=medicwho;id=162f1550-4686-40a9-bec1-66aa5cf4fcc0;subscriber=1;reply-parent-msg-id=72733dad-a857-408d-ab16-e96834aadf73;first-msg=0;reply-parent-display-name=MedicWho;reply-thread-parent-msg-id=72733dad-a857-408d-ab16-e96834aadf73;badges=subscriber/24,no_audio/1;reply-thread-parent-user-login=medicwho;room-id=11148817;user-id=117691339;turbo=0;color=#DAA520;reply-parent-user-id=71651832;display-name=Mm2PL;badge-info=subscriber/27;reply-parent-msg-body=ࢿࣃࣄࣁࢾࣂࣄࣁࢿࣃࣄࢿࢾࣂࣃࢿࣄࣃࣁࢿࣆࣀࣃࣄࢿࣃࣆࣃࣀࣂࢾࣅࣀࣃࢿࣆࣄࢿࣂࣂࢾࣆࣁࣂࢾࣅࢿࢾࣁࣄࢿࣀࣁࣆࣁࣁࢿࣂࣆࣆࣄࣃࢿࣀࣁࣁࣃࣀࣅࣃࣃࣀࣂࣂࣀࢿࢾࣅࢾࣃࢿࢾࣅࣀࣂࣀࣁࣃࣃࣃࣀࢾࣄࣃࣆࢾࣃࣆࣄࣃࣅࣃࣂࣄࣁࣆࢾࣄࢿࣃࢾࢿࣅࣆࣃࢾࣄࣄࣂࢾࣂࣂࣁࣅࣅࢿࢾࣀࢾࣄࣅࣅࢿࣆࣄࣁࣄࣂࣁࣄࢾࣁࣅࣆࣂࢿࣄࣅࣆࢿࢿࣄࣀࣄࣅࣃࣄࢿࣀࣄࢿࣅࣆࣆࣆࢾࣀࢾࢿࣄࣂࣁࣆࣁࣅࣀࢿࣃࣀࣅࢿࣆࣃࣅࣂࢿࢿࢾࣀࣆࣆࣁࣂࣃࣄࣁࣀࣁࣆࢾࢾࣁࢾࣆࢿࣁࣆࣂࣃࢾࣄࢿࣂࣂࢿࣂࣃࣃࣄࣃࣂࣆࢿࢿࣄࣂࢾࣄࢾࢿࣃࢿࣁࣅࣆࣅࣁࢾࣀࣆࢾࣂࣂࣅࣃࣀࣂࢾࣁࢿࣅࣃࢿࣂࣄࢾࣆࢿࣂࣄࣁࣁࢿࣂࢿࣄࣆࣃࢾࣆࣀࣀࣀࢾࣃࣃࣀࣁࣃࣄࣀࣀࣆࣅࣁࣁࣂࣁࣅࣅࢾࣀࢿࣂࣀࣀࢿࢿࣃࣆࣆࣃࣄࣆࣆࣁࣁࣆࢾࢾࣃࣄࢿࢿࢾࣆࣆࣄࢿࣄࢾࣅࣂࣆࣀࣀࣅࣂࣁࣄࣅࢾࢾࣆࣀࣆࣆࣅࣀࣁࣅࣅࢿࣆࣀࣆࣂࢾࢿࢿࣃࢿࣃࢿࢾࢾࣄࣁࣁࢾࣁࣀࣅࣅࢾࣆࣄࣂࣄࣅࢿࢾࣄࣅࣄࣁࣅࣆࣄࢾࣀࣃࣄࢿࣂࣀࣁࣆࣆࣃࣁࣃࣆࣄࣆࢿࣁࢾࣂࣃࣃࣂࣆࣆࣂࣃࢿࣁࣆࣃࣆࣀࣅࣅࣃ;mod=0;returning-chatter=0;emotes=;flags=;tmi-sent-ts=1694024399198;user-type= :mm2pl!mm2pl@mm2pl.tmi.twitch.tv PRIVMSG #!!! :@MedicWho _unicode
   ```

---

When running a release build:

> [!WARNING]
> Do this in a private channel! This will impact others!

1. Send 
   ```
   ࢿࣃࣄࣁࢾࣂࣄࣁࢿࣃࣄࢿࢾࣂࣃࢿࣄࣃࣁࢿࣆࣀࣃࣄࢿࣃࣆࣃࣀࣂࢾࣅࣀࣃࢿࣆࣄࢿࣂࣂࢾࣆࣁࣂࢾࣅࢿࢾࣁࣄࢿࣀࣁࣆࣁࣁࢿࣂࣆࣆࣄࣃࢿࣀࣁࣁࣃࣀࣅࣃࣃࣀࣂࣂࣀࢿࢾࣅࢾࣃࢿࢾࣅࣀࣂࣀࣁࣃࣃࣃࣀࢾࣄࣃࣆࢾࣃࣆࣄࣃࣅࣃࣂࣄࣁࣆࢾࣄࢿࣃࢾࢿࣅࣆࣃࢾࣄࣄࣂࢾࣂࣂࣁࣅࣅࢿࢾࣀࢾࣄࣅࣅࢿࣆࣄࣁࣄࣂࣁࣄࢾࣁࣅࣆࣂࢿࣄࣅࣆࢿࢿࣄࣀࣄࣅࣃࣄࢿࣀࣄࢿࣅࣆࣆࣆࢾࣀࢾࢿࣄࣂࣁࣆࣁࣅࣀࢿࣃࣀࣅࢿࣆࣃࣅࣂࢿࢿࢾࣀࣆࣆࣁࣂࣃࣄࣁࣀࣁࣆࢾࢾࣁࢾࣆࢿࣁࣆࣂࣃࢾࣄࢿࣂࣂࢿࣂࣃࣃࣄࣃࣂࣆࢿࢿࣄࣂࢾࣄࢾࢿࣃࢿࣁࣅࣆࣅࣁࢾࣀࣆࢾࣂࣂࣅࣃࣀࣂࢾࣁࢿࣅࣃࢿࣂࣄࢾࣆࢿࣂࣄࣁࣁࢿࣂࢿࣄࣆࣃࢾࣆࣀࣀࣀࢾࣃࣃࣀࣁࣃࣄࣀࣀࣆࣅࣁࣁࣂࣁࣅࣅࢾࣀࢿࣂࣀࣀࢿࢿࣃࣆࣆࣃࣄࣆࣆࣁࣁࣆࢾࢾࣃࣄࢿࢿࢾࣆࣆࣄࢿࣄࢾࣅࣂࣆࣀࣀࣅࣂࣁࣄࣅࢾࢾࣆࣀࣆࣆࣅࣀࣁࣅࣅࢿࣆࣀࣆࣂࢾࢿࢿࣃࢿࣃࢿࢾࢾࣄࣁࣁࢾࣁࣀࣅࣅࢾࣆࣄࣂࣄࣅࢿࢾࣄࣅࣄࣁࣅࣆࣄࢾࣀࣃࣄࢿࣂࣀࣁࣆࣆࣃࣁࣃࣆࣄࣆࢿࣁࢾࣂࣃࣃࣂࣆࣆࣂࣃࢿࣁࣆࣃࣆࣀࣅࣅࣃ
   ```
4. Reply to the message

</details> 

The manual text elision is really slow here, because it removes one character of the last overflowing word at a time. Qt has a native text elision algorithm, that seems to be faster ([`QFontMetrics::elidedText`](https://doc.qt.io/qt-6/qfontmetrics.html#elidedText)).

_This isn't the exact same workload (as I couldn't interact with Chatterino before) but in both cases, the same channel was shown._ 
| Before | After |
|---| ---|
| vtune-gui_2023-09-07_13-03-18  | vtune-gui_2023-09-07_13-03-01 |

### Further Improvements

Even with this implementation, Chatterino will lag if many channels are open and one or more channels contain a reply to a bad message. This is because there are too many forced layouts. When many channels are open and an image gets loaded, all `ChannelViews` are forcefully laid out. Images are loaded when laying out messages (see https://github.com/Chatterino/chatterino2/issues/3929 and https://github.com/Chatterino/chatterino2/issues/3929). 

A `ChannelView` that's not visible shouldn't need to be laid out. I attempted a fix in [`perf/layout-visible`](https://github.com/Nerixyz/chatterino2/tree/perf/layout-visible). This does have some consequences (e.g. switching channels isn't _as smooth_ but not really noticeable - could be behind a setting [lazy vs. eager layout]). On this upside, this has the same effect as https://github.com/Chatterino/chatterino2/issues/3929 in that images won't be loaded if they're not needed, thus reducing memory usage and reducing lag when loading recent messages at the start for all open channels. 

Another optimization one can make is to swap `LimitedQueue` for message layouts inside a `ChannelView` with a raw `boost::circular_buffer`, since it's only ever accessed by the GUI thread. I did that in [`perf/no-limited-queue`](https://github.com/Nerixyz/chatterino2/tree/perf/no-limited-queue). This will reduce a lot of unnecessary copies (for all the snapshots) resulting in less memory use and lower rendering times (copying the shared pointers requires incrementing the reference count).

Yet another optimization could be to remember in a `MessageLayout` if it contained any unloaded images. `MessageLayout::layout` could then ignore layout requests if an image was loaded, but the current element doesn't contain any unloaded image.


Nerixyz:
Just realized that `SingleLineTextElement` doesn't need to split the message into words. `Emoji::parse` can handle spaces. This would also reduce the amount of measurements when the text contains spaces. Furthermore, once Qt 5.12 support is dropped, `Emoji::parse` can return `QStringView`s into the original string.

pajlada (approved):

PR #4811, opened 2023-09-10
Only layout visible `ChannelView`s

Nerixyz:
# Description

I mentioned this in https://github.com/Chatterino/chatterino2/pull/4807#issue-1885699481. Before, all `ChannelView`s are laid out when requested. In active channels, there can be a lot of requests for this (every message -> one layout request). When a `ChannelView` isn't shown it shouldn't need to be laid out, but only once it becomes shown. Parts of the scrollbar implementation relied on the previous behavior, so a lot of testing is needed.

As mentioned in https://github.com/Chatterino/chatterino2/pull/4807#issue-1885699481, this behavior could be a setting.


pajlada:
Have you tested this with smooth scrolling on new messages enabled?

Nerixyz:
> Have you tested this with smooth scrolling on new messages enabled?

A bit, but not much. I have it disabled on my main instance (where I'm using the changes).

pajlada (approved):
Code looks good, this should be a performance win for all but I want to try this out in nightly to see if we get any complaints about behaviour changes

PR #4812, opened 2023-09-11
Fixed too much text being copied when copying chat messages

### LINKED ISSUE #2449: Copying selection adds an extra space at the end

Nerixyz:
# Description

This is a continuation of #3080. The same approach is used. 

Because of the reply button being in every message, setting `hasTrailingSpace` on the last element (even the last text element) when building as mentioned in https://github.com/Chatterino/chatterino2/pull/3080#issuecomment-893014909 doesn't work. 
Instead, `hasTrailingSpace` is set on the last _layout_ element. With the button enabled, a trailing space is copied (as that space is visually selected). Without the button, no trailing space is copied. When selecting multiple lines, having no delimiter after a message is bad, so a newline is added (over a space to distinguish lines).

Fixes #2449.

Additionally, this fixes the issue of copying emotes when they're not selected (e.g. in `a 🅰` with `a ` selected, Chatterino would copy `a 🅰 `).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Bugfix: Fixed Usercard popup not floating on tiling WMs on Linux when "Automatically close user popup when it loses focus" setting is enabled. (#3511)
 - Bugfix: Fixed selection of tabs after closing a tab when using "Live Tabs Only". (#4770)
 - Bugfix: Fixed input in reply thread popup losing focus when dragging. (#4815)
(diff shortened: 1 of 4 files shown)


pajlada (approved):
LGTM :+1: Will do a manual merge when another PR has been merged in

PR #4816, opened 2023-09-14
Add `WindowManager::getLastSelectedWindow()` to replace `getMainWindow()`

Mm2PL:
# Description

This is needed to fix `/clearmessages`. I will replace all usages of `getMainWindow()` in another PR.

Here is a simple fix for `/clearmesages` which can be used to test this.

Fixes #3353. Required to finally close issue #2406. 

<details>
<summary>Patch</summary>

```patch
diff --git a/src/controllers/commands/CommandController.cpp b/src/controllers/commands/CommandController.cpp
index 9ce784ba..68eb2939 100644
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -1377,18 +1377,25 @@ void CommandController::initialize(Settings &, Paths &paths)
         return "";
     });
 
-    this->registerCommand("/clearmessages", [](const auto & /*words*/,
-                                               ChannelPtr channel) {
-        auto *currentPage = dynamic_cast<SplitContainer *>(
-            getApp()->windows->getMainWindow().getNotebook().getSelectedPage());
+    this->registerCommand(
+        "/clearmessages",
+        [](const auto & /*words*/, ChannelPtr channel) -> QString {
+            auto *win = getApp()->windows->getLastSelectedWindow();
+            if (win == nullptr)
+            {
+                return "";
+            }
 
-        if (auto split = currentPage->getSelectedSplit())
-        {
-            split->getChannelView().clearMessages();
-        }
+            auto *currentPage = dynamic_cast<SplitContainer *>(
+                win->getNotebook().getSelectedPage());
 
-        return "";
-    });
+            if (auto *split = currentPage->getSelectedSplit())
+            {
+                split->getChannelView().clearMessages();
+            }
+
+            return "";
+        });
 
     this->registerCommand("/settitle", [](const QStringList &words,
                                           ChannelPtr channel) {
```

</details>

PR #4819, opened 2023-09-16
fix: QuickSwitcherPopup now pops up in the selected window

pajlada:
# Description

- fix: QuickSwitcherPopup now pops up in the selected window
- QuickSwitcherPopup: reduce indentation anon namespace
- QuickSwitcherPopup: constexpr the minimum size
- QuickSwitcherPopup: remove redundant virtual from `themeChangedEvent`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Bugfix: Fixed `/shoutout` command not working with usernames starting with @'s (e.g. `/shoutout @forsen`). (#4800)
 - Bugfix: Fixed selection of tabs after closing a tab when using "Live Tabs Only". (#4770)
 - Bugfix: Fixed input in reply thread popup losing focus when dragging. (#4815)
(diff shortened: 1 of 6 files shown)


pajlada:
@Mm2PL I removed one of the questionable usages of `getMainWindow`, could you take a look at this PR? Only the first commit makes actual behavioral changes

Related to https://github.com/Chatterino/chatterino2/pull/4816

Mm2PL (changes requested):
Opening a new channel in a tab still uses the main window. Can you fix that? It is located in `src/widgets/dialogs/switcher/NewTabItem.cpp`.

pajlada:
> Opening a new channel in a tab still uses the main window. Can you fix that? It is located in `src/widgets/dialogs/switcher/NewTabItem.cpp`.

[`2b75961` (#4819)](https://github.com/Chatterino/chatterino2/pull/4819/commits/2b75961d44656b1bf990ac961173876e0eea81ec) should fix this. I have tested it but would be good with a sanity check. Good catch!

Mm2PL (approved):
Yep works as expected now.

PR #4821, opened 2023-09-16
Add the Changelog Category Check workflow

pajlada:
This ensures the changelog entry added, if any, is under the Unreleased category

This action can be skipped with the `skip-changelog-check` label

# Description


pajlada:
image

Issue #4825, opened 2023-09-17
Reduce Compilation Warnings


Nerixyz:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When compiling Chatterino, there are a few warnings. These depend on the compiler and configuration used, so I'd like to collect them all here and have some kind of tracking issue.

If you're fixing a warning, please do one category (e.g. unused variables) or subcategory or file at a time (like https://github.com/Chatterino/chatterino2/pull/4818).

I'll start with Qt 6 on Windows - I included at most one warning kind per file, meaning there might be multiple warnings in one file. Note that [C4505](https://learn.microsoft.com/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4505), [C4100](https://learn.microsoft.com/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100), and [C4267](https://learn.microsoft.com/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4267) are explicitly disabled in CMake.

## Signed/unsigned mismatch (C4018, `sign-compare`)

- [MessageLayoutContainer.cpp](https://github.com/Chatterino/chatterino2/blob/3265df76619e689e9a738517732fe930fbe2b590/src/messages/layouts/MessageLayoutContainer.cpp#L586-L626): warning C4018: '<=': signed/unsigned mismatch (THERE ARE A LOT OF THESE)
- [ChannelView.cpp](https://github.com/Chatterino/chatterino2/blob/3265df76619e689e9a738517732fe930fbe2b590/src/widgets/helper/ChannelView.cpp#L1559-L1678): warning C4018: '<': signed/unsigned mismatch (THERE ARE A LOT OF THESE)

## Unused variables (C4189, C4101, `unused-variable`)

- [PubSubManager.cpp](https://github.com/Chatterino/chatterino2/blob/3265df76619e689e9a738517732fe930fbe2b590/src/providers/twitch/PubSubManager.cpp#L213): warning C4101: 'ok': unreferenced local variable
[...]

PR #4830, opened 2023-09-18
Fix selection rendering

Nerixyz:
# Description

After 3265df76619e689e9a738517732fe930fbe2b590, the selection rendering was off. Selecting the trailing newline of a message (upwards) would render the whole last line of a message to be selected. However, when copying the text, only the trailing newline of the message would be included.

This PR refactors the rendering of the selection to be a bit more readable and to reduce `sign-compare` warnings in the code (see https://github.com/Chatterino/chatterino2/issues/4825). In addition, a special case for the newline is added to be consistent with rendering the selection of the "trailing newline" (downwards).

Even though the selection is now correctly rendered, this might not be the exact behavior one would expect. Not including the trailing newline might be desired. Feel free to discuss this here and I can potentially adjust the behavior.


pajlada:
I'm unable to reproduce the reported issue, could you provide some more reproduction instructions?

pajlada (approved):
image

Great job @Nerixyz!

PR #4834, opened 2023-09-22
streamlink: trim custom paths

brian6932:
Windows paths can't have trailing whitespace, currently having trailing whitespace in the path will result in the path not being found


pajlada:
I don't think there would be any issue with trimming this path for other platforms as well, can anyone else come up with any scenarios where that would be unwanted?

iProdigy:
Technically other platforms do allow whitespace at the end of folder names

brian6932:
> Technically other platforms do allow whitespace at the end of folder names

Yea this is why I didn't want to do that, just in the case that someone had such an edge csae

Felanbird (approved):

pajlada (approved):

PR #4840, opened 2023-09-24
Fix emoji unified/non-qualified version for sending & parsing

### LINKED ISSUE #4792: Emoji Inconsistancies
When it comes to how emojis are handled on Chatterino, we have no issues displaying the emojis we send and are sent from Twitch, but there are problems on Twitch with the emojis we send.

In certain cases such as 🏴󠁧󠁢󠁳󠁣󠁴󠁿 - this is the fault of Twitch and documented via https://togithub.com/twitchdev/issues/issues/307

In other cases, it falls to https://github.com/iamcal/emoji-data not using `U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}`
###### I'm unsure if this is an error on their end, or just something Twitch expects that this db didn't automatically provide.

Some examples of this are as follows:

| Emoji     | Twitch outcome | Chatterino outcome |
| ------------- |:-------------:| -----:|
| ❤️ | `❤◌️` | `❤` |
| 🛏️ | `🛏◌️`| `🛏`|
| 🗑 | `🗑◌️` | `🗑` |
| ☄️ | `☄◌️` | `☄` |
| 🏖️ | `🏖◌️` | `🏖` |
| 🛩 | | 

<details>
<summary>Raw unicode</summary>

```
❤️
Twitch:
U+2764 : HEAVY BLACK HEART
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino:
U+2764 : HEAVY BLACK HEART
```
```
🛏
Twitch:
U+1F6CF : BED
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino:
U+1F6CF : BED
```
```
🗑️
Twitch:
U+1F5D1 : WASTEBASKET {trashcan}
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino:
U+1F5D1 : WASTEBASKET {trashcan}
```
```
☄️ 
Twitch:
U+2604 : COMET
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino: 
U+2604 : COMET
```
```
🏖️ 
Twitch: 
U+1F3D6 : BEACH WITH UMBRELLA
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino: 
U+1F3D6 : BEACH WITH UMBRELLA
```
```
🛩
Twitch: 
U+1F6E9 : SMALL AIRPLANE
U+FE0F : VARIATION SELECTOR-16 [VS16] {emoji variation selector}

Chatterino: 
U+1F6E9 : SMALL AIRPLANE

```

</details>

Options:

1. Fix and PR to https://github.com/iamcal/emoji-data
2. Fork and modify for our usage where necessary
3. Do nothing and continue to annoy browser chatters

I'm leaning towards option 2 for the following reasons:
- The list of problematic emojis is basically unknown from our PoV, it relies on users reporting the problem. Whether or not this is the correct standard is also unknown, but necessary for the purpose of arguing the change.
- Forking it ourselves seems like less work with the exception of learning _how_ to fix this problem, from what I can gather we'd probably just need to make a file similar to https://github.com/iamcal/emoji-data/blob/master/build/unicode/emoji-zwj-sequences.txt for our applicable situation.
- Doing nothing is a little toxic, I've been gaslighting moonmoon's mods about it but it does seem to be a problem outside of Twitch's normal emoji issues

pajlada:
# Description

- fix: use unified emoji for sending, & use both unified & nonQualified for parsing
- refactor: rename parsed shortCodes to shortNames to avoid name clashes
- refactor: mark emojis name as unused in the emoji set loading
- refactor: use contains(..) instead of count(..)
- refactor: const auto the emoji
- refactor: move anon namespace out to flatten in
- refactor: constify & rename TONE_NAMES
- refactor: parseEmoji make shortCode const ref
- move map include

Fixes #4792

This has gone an extra step to make sure that this version of Chatterino can parse emojis that are sent from old versions of chatterino & from users of web chat

this means slightly higher memory usage which can be resolved by splitting EmojiData into an Emoji and EmojiData, where we mostly only store Emoji (with its value, nonUnified, and Emote values).
We'd just need to figure out how to change emoji sets


pajlada:
this breaks some emojis, gotta test this more

Felanbird:
notes:

as of this comment all known broken emojis _send_ correctly
image

heart and comet _render_ correctly, the others do not
image

other emotes such as 😂 render incorrectly now
image

src/providers/emoji/Emojis.cpp
```diff
@@ -296,31 +319,61 @@ std::vector<boost::variant<EmotePtr, QString>> Emojis::parse(
...
-                    match = false;
+                    matchedEmoji = emoji;
+                    matchedEmojiLength = emoji->value.length();
 
                     break;
                 }
             }
-
-            if (match)
+            else if (remainingCharacters >= emojiNonQualifiedExtraCharacters)
```
Nerixyz, line 348: This `else if` should be an `if` otherwise nonqualified emojis will never be matched in the middle of strings.

src/providers/emoji/Emojis.cpp
```diff
@@ -296,31 +319,61 @@ std::vector<boost::variant<EmotePtr, QString>> Emojis::parse(
...
+
+                for (int j = 1; j < emoji->nonQualified.length(); ++j)
+                {
+                    if (text.at(i + j) != emoji->nonQualified.at(j))
+                    {
+                        match = false;
+
+                        break;
+                    }
+                }
```
Nerixyz, line 362: To make this more readable, this can be written as 

```cpp
bool match = QStringView{emoji->nonQualified}.mid(1) == 
             QStringView{text}.mid(i + 1, emojiNonQualifiedExtraCharacters);
```

Though it should be noted that this will introduce a call to `QStringView operator==` which can't be inlined and this code is in a hot loop (up to about 1600 iterations per detected emoji).

src/providers/emoji/Emojis.cpp
```diff
@@ -296,31 +319,61 @@ std::vector<boost::variant<EmotePtr, QString>> Emojis::parse(
...
+
+                if (match)
+                {
+                    matchedEmoji = emoji;
+                    matchedEmojiLength = emoji->nonQualified.length();
 
-                break;
+                    break;
+                }
+                // look in emoji->nonQualified
```
Nerixyz, line 371: This should probably be moved to the top of the block to match the block above.

src/providers/emoji/Emojis.cpp
```diff
@@ -296,31 +319,61 @@ std::vector<boost::variant<EmotePtr, QString>> Emojis::parse(
 
         for (const std::shared_ptr<EmojiData> &emoji : possibleEmojis)
         {
+            int emojiNonQualifiedExtraCharacters =
```
Nerixyz, line 322: This and all other `int`s here should be `QString::size_type` (can often use `auto`) for compatibility with both Qt 5 and Qt 6.

src/providers/emoji/Emojis.cpp
```diff
@@ -12,123 +12,144 @@
...
-        else
-        {
-            unicodeCharacters = emojiData->unifiedCode.toLower().split('-');
-        }
-        if (unicodeCharacters.length() < 1)
+    QStringList nonQualifiedCharacters =
+        emojiData->nonQualifiedCode.toLower().split('-');
+    QStringList unicodeCharacters = emojiData->unifiedCode.toLower().split('-');
+
+    assert(unicodeCharacters.length() >= 1);
```
Nerixyz, line 83: Isn't this always the case? Even `QString().split("-").length()` is `1`.

src/providers/emoji/Emojis.cpp
```diff
@@ -12,123 +12,144 @@
...
         }
+    }
 
-        int numUnicodeBytes = 0;
+    // We can safely do a narrowing static cast since unicodeBytes will never be a large number
+    emojiData->value = QString::fromUcs4(unicodeBytes.data(),
+                                         static_cast<int>(unicodeBytes.size()));
 
-        for (const QString &unicodeCharacter : unicodeCharacters)
+    if (nonQualifiedCharacters.length() > 0)
```
Nerixyz, line 101:
```suggestion
    if (!emojiData->nonQualifiedCode.isEmpty())
```

As noted above, there will always be one item, even in empty strings.

src/providers/emoji/Emojis.cpp
Nerixyz: And it should include `!emoji->nonQualified.isNull()`.

pajlada: Fixed in [`bc8e731` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/bc8e7317a50217f1053dcd56d1008f6290909f27)

src/providers/emoji/Emojis.cpp
pajlada: [`bc8e731` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/bc8e7317a50217f1053dcd56d1008f6290909f27)

src/providers/emoji/Emojis.cpp
pajlada: This assert just makes sure there is a `unifiedCode` for the emoji. Instead of asserting at this stage I'll change this to a check right as `unifiedCode` is parsed from the json value

pajlada: Should be cleared up in [`027dcf6` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/027dcf6674647d5bcc676530c222af5a87378570)

pajlada: I see what you mean now, since split keeps empty parts

src/providers/emoji/Emojis.cpp
pajlada: Fixed in [`5198a28` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/5198a287325719b7481357bd7d75b4140536e6b2)

src/providers/emoji/Emojis.cpp
pajlada: Working on benchmarks and then I'll look at this

pajlada: Benchmark showed the two implementations were equal, see [`92133d8` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/92133d81daa4d15dcacd9915766e93bf44b1f01b)

src/providers/emoji/Emojis.cpp
pajlada: Fixed in [`1213173` (#4840)](https://github.com/Chatterino/chatterino2/pull/4840/commits/1213173a88fc62a174c42253c3efcabc971ff50b)

pajlada:
image

PR #4841, opened 2023-09-27
Add `author.id` variable for filters

Foretack:
# Description

This PR adds the `author.id` variable for use in filters. The use of this variable would be especially useful to deal with name changes

### DIFF
--- a/src/controllers/filters/lang/Filter.cpp
+++ b/src/controllers/filters/lang/Filter.cpp
@@ -26,6 +26,7 @@ ContextMap buildContextMap(const MessagePtr &m, chatterino::Channel *channel)
      * author.badges
      * author.color
      * author.name
+     * author.id
      * author.no_color
      * author.subbed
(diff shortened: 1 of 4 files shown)


src/controllers/filters/lang/Filter.cpp
```diff
@@ -81,6 +82,7 @@ ContextMap buildContextMap(const MessagePtr &m, chatterino::Channel *channel)
         {"author.badges", std::move(badges)},
         {"author.color", m->usernameColor},
         {"author.name", m->displayName},
+        {"author.id", m->id},
```
Nerixyz, line 85: This isn't the author's ID, it's the messages' ID:

https://github.com/Chatterino/chatterino2/blob/783c7530f3aedcdf493e508549b4082b13610cf9/src/providers/twitch/TwitchMessageBuilder.cpp#L587-L595

PR #4847, opened 2023-09-29
Check VCRT and show installed/shipped version

Nerixyz:
# Description

* If the installed VCRT is old enough, the task to install it will be hidden.
* If the version is too low, the installer now shows the installed and shipped version.

For testing, you can change `Bld` to a lower number in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64` and/or rename it (or `Version`) to simulate the case where it's not installed.

---

Aside: The installer currently doesn't have the option to specify the installation directory. I'm not sure if that's intentional. `DisableDirPage=false` can be added to show the page.

### DIFF
--- a/.CI/chatterino-installer.iss
+++ b/.CI/chatterino-installer.iss
@@ -19,7 +19,7 @@
 #endif
 ; Set to the string representation of the VCRT version
 #ifndef SHIPPED_VCRT_VERSION
-#define SHIPPED_VCRT_VERSION '?'
+#define SHIPPED_VCRT_VERSION ?
 #endif
 
 [Setup]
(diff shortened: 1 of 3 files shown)


.CI/chatterino-installer.iss
```diff
@@ -56,7 +65,8 @@ SetupWindowTitle=Setup - %1 (Nightly)
 #endif
 
 [Tasks]
-Name: "vcredist"; Description: "Install the required Visual C++ 2015/2017/2019/2022 Redistributable";
+; Only show this option if the VCRT can be updated.
+Name: "vcredist"; Description: "Install the required Visual C++ 2022 Redistributable ({code:VCRTDescription})"; Check: NeedsNewVCRT();
```
kornes, line 69: maybe instead of hardcoding `Visual C++ 2022 Redistributable` we could use VersionInfo.Product here?
it resolves to `Microsoft Visual C++ 2015-2022 Redistributable (x64) - 14.36.32532`, would save some maintenance in future

Nerixyz: Thank you! I was actually looking for something like this.

Nerixyz: Hmm, `VersionInfo.Product` doesn't seem to return anything for me (even though it's shown when stringifying the version info), so I used `VersionInfo.FileDescription`.

kornes: both Product and FileDescription have same content on all versions i had on hand, so should be fine

pajlada (approved):

PR #4848, opened 2023-09-30
Fix to interpret logger channel names as lowercase

GongBingWong:
# Description

Fixes the issue (https://github.com/Chatterino/chatterino2/issues/4835) where logging channels with capitalization wouldn't work, by making channel names lowercase.

I tested the original exe file without my changes and then tested an exe file with my changes and it works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Minor: The account switcher is now styled to match your theme. (#4817)
 - Minor: Add an invisible resize handle to the bottom of frameless user info popups and reply thread popups. (#4795)
 - Minor: The installer now checks for the VC Runtime version and shows more info when it's outdated. (#4847)
-- Bugfix: Fixed capitalized channel names in log inclusion list not being logged. (#4835)
+- Bugfix: Fixed capitalized channel names in log inclusion list not being logged. (#4848)
 - Bugfix: Trimmed custom streamlink paths on all platforms making sure you don't accidentally add spaces at the beginning or end of its path. (#4834)
 - Bugfix: Fixed a performance issue when displaying replies to certain messages. (#4807)
 - Bugfix: Fixed a data race when disconnecting from Twitch PubSub. (#4771)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Minor: Migrate to the new Get Channel Followers Helix endpoint, fixing follower count not showing up in usercards. (#4809)
 - Minor: The account switcher is now styled to match your theme. (#4817)
 - Minor: Add an invisible resize handle to the bottom of frameless user info popups and reply thread popups. (#4795)
+- Bugfix: Fixed capitalized channel names in log inclusion list not logged. (#4835)
 - Bugfix: Trimmed custom streamlink paths on all platforms making sure you don't accidentally add spaces at the beginning or end of its path. (#4834)
 - Bugfix: Fixed a performance issue when displaying replies to certain messages. (#4807)
 - Bugfix: Fixed a data race when disconnecting from Twitch PubSub. (#4771)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Minor: Migrate to the new Get Channel Followers Helix endpoint, fixing follower count not showing up in usercards. (#4809)
 - Minor: The account switcher is now styled to match your theme. (#4817)
 - Minor: Add an invisible resize handle to the bottom of frameless user info popups and reply thread popups. (#4795)
+- Bugfix: Fixed capitalized channel names in log inclusion list not logged. (#4835)
```
Felanbird, line 8:
```suggestion
- Bugfix: Fixed capitalized channel names in log inclusion list not being logged. (#4835)
```

src/controllers/logging/ChannelLog.cpp
```diff
@@ -9,7 +9,8 @@ ChannelLog::ChannelLog(QString channelName)
 
 QString ChannelLog::channelName() const
 {
-    return this->channelName_;
+    QString lowercaseChannelName_ = this->channelName_.toLower();
+    return lowercaseChannelName_;
```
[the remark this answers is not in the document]
GongBingWong: Since the "lowercaseChannelName_" variable stored the "channelName_" variable as lowercase, I added a "_" at the end of it. Is "lowercaseChannelName" better than "lowerCaseChannelName_"?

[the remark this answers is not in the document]
pajlada: Local variables should not use an underscore at the end, no

pajlada, line 13: nit: You don't have to make a temporary variable here, you can just `return this->channelName_.toLower();` since [QString::toLower](https://doc.qt.io/qt-5/qstring.html#toLower) returns a copy of the string

pajlada (changes requested):
Works as expected, I sent one suggestion as a stylistic thing but otherwise looks good.
Thank you!

src/controllers/logging/ChannelLog.cpp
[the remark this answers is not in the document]
GongBingWong: Okay, thanks. I'll commit the bot's suggestion then.

src/controllers/logging/ChannelLog.cpp
```diff
@@ -9,7 +9,8 @@ ChannelLog::ChannelLog(QString channelName)
 
 QString ChannelLog::channelName() const
 {
-    return this->channelName_;
+    QString lowercaseChannelName = this->channelName_.toLower();
+    return lowercaseChannelName;
```
pajlada, line 13: nit: You don't have to make a temporary variable here, you can just return this->channelName_.toLower(); since [QString::toLower](https://doc.qt.io/qt-5/qstring.html#toLower) returns a copy of the string

pajlada (changes requested):
Works as expected, I sent one suggestion as a stylistic thing but otherwise looks good.
Thank you!

src/controllers/logging/ChannelLog.cpp
GongBingWong: Okay, thank you for the suggestion. Its an improvement over the current code.  Should I 1) leave the current code as is, 2) commit the line you suggested, or 3) wait for you to do a "suggested change" within the file?

GongBingWong: I decided to just do a commit since I saw it was a "requested change".

pajlada: That's the right option in my book, thank you!

pajlada (approved):
Thanks for your contribution!

In our changelog style, the number behind the changelog entry should be the number of the PR rather than the issue, since we think it gives a bit more context for the whole fix process. Other than that, this works as expected and is the correct solution to this :+1: 

Only other feedback I have for future contributions is that it's best to use a branch other than `master` or `main` or `mainline` for your PRs, make a specific change branch like `fix/logger-lowercase`, that way it'll be easier to manage your workflow if you have multiple PRs active to the same repo, and it'll be easier on my end since your branch name doesn't conflict with our main branch name

pajlada:

PR #4851, opened 2023-09-30
Fix flickering when running with Direct2D

Nerixyz:
# Description

When running with Direct2D on Windows, Chatterino would flicker and print warnings that `EndDraw` failed with `D2DERR_POP_CALL_DID_NOT_MATCH_PUSH`. This was because derived buttons (titlebar & notebook) would call `Button::paintEvent` in their `paintEvent` - creating two painters for a paint device at a time (not permitted - [see `QPainter::begin`](https://doc.qt.io/qt-6/qpainter.html#begin)). This PR fixes that.

The only other "issues" I had with Direct2D was [an exception leaking through the driver](https://forums.developer.nvidia.com/t/poco-notfoundexception-thrown-in-driver-version-531-18/245285) and that text rendering isn't as crisp.


pajlada (approved):
:+1:

Issue #4796, opened 2023-08-27
Add a compile-time flag to disable the updater


pajlada:
For builds that are updated through other means (e.g. brew or AUR) there shouldn't be a need for the update to prompt that it's out of date, at least not as loudly

libklein:
I believe that we can close this issue with #4854 being merged.

Felanbird:
Implemented in #4854

PR #4854, opened 2023-10-01
Implement compile-time flag to disable automatic update checks.

libklein:
# Description

This PR adds a compile-time flag to disable automatic updates (cf. #4796). Specifically, it prevents the update button from being shown/updated if the Updater detects that an update is necessary. A potential drawback of this approach is that it also does not show the button if the user "requests" this, e.g., by enabling beta updates. 

An alternative implementation could also simply disable the automatic update check on application startup.

I believe that the former approach is more desirable as it ensures that the Updater class is in it's expected state regardless of whether updates are disabled or not. We could, for instance, silently log that the application is out of state.


Mm2PL (changes requested):
Chatterino never updates automatically without the user's explicit action. It only checks for updates and displays a prompt (which I believe to be an important enough distinction to mention).
Please don't just hide the button. This approach results in wasted requests to the update server. Instead please look at `src/singletons/Updates` (notably `Updates::checkForUpdates`) and disable updates there.

src/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ add_compile_definitions(QT_DISABLE_DEPRECATED_BEFORE=0x050F00)
 
 # registers the native messageing host
 option(CHATTERINO_DEBUG_NATIVE_MESSAGES "Debug native messages" OFF)
+option(DISABLE_AUTOMATIC_UPDATES "Disable automatic updates" OFF)
```
pajlada, line 8: I'd prefer if this was named `CHATTERINO_DISABLE_UPDATER` and defined in the root CMakeLists.txt, underneath the `CHATTERINO_PLUGIN` option

pajlada (changes requested):
+ what mm2pl mentioned, I think I prefer the network request not being run in the first place

libklein:
Sounds good!

I'd personally prefer disabling the updates by removing the call in https://github.com/Chatterino/chatterino2/blob/master/src/RunGui.cpp#L273 instead of modifying `Updater.cpp` directly. This way the class itself remains functional, it'll just not be triggered.

I'd also suggest disabling https://github.com/Chatterino/chatterino2/blob/master/src/Application.cpp#L190 to stay consistent - otherwise we'd have to trigger an update check on startup based on the betaUpdates setting. I actually think that hiding the beta checkbox/ignoring the setting altogether is probably in line with the PR, given that it's motiviation is to stop annoying users that source chatterino from a package repository with update notifications. I do see that this is debatable tough.

Happy to hear your opinion on this!

pajlada:
Instead of ifdefing out the calls to `checkForUpdates`, I'd rather have this check just early return in the `checkForUpdates` itself

pajlada (approved):

Mm2PL (approved):
lgtm

CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ option(BUILD_TRANSLATIONS "" OFF)
 option(BUILD_SHARED_LIBS "" OFF)
 option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable EXPERIMENTAL plugin support in Chatterino" OFF)
+option(CHATTERINO_DISABLE_UPDATER "Disable update checks" OFF)
```
fourtf, line 29: This appears to be the only flag that disables something instead of enabling something. Should this be flipped?

pajlada: I considered this, but the cmake flag `CHATTERINO_DISABLE_UPDATER` is more closely aligned what I'd like the code in `Updates.cpp` to look like

fourtf: Seems like a matter of replacing `#ifdef` with `#ifndef`, or am I missing something?

pajlada: You're right, it's a question of

```c++
#ifdef CHATTERINO_DISABLE_UPDATER
   return;
#endif
```

which I read as "If the updater is disabled"

vs

```c++
#ifndef CHATTERINO_ENABLE_UPDATER
   return;
#endif
```

which I read as "If the updater is not enabled"

and to me, the upper option more clearly describes the intent

100% nitpicking and I would be fine with either, especially since it has such a small scope, only being used in one place in one file

fourtf: 1) I believe you would usually wrap `#ifdef CHATTERINO_ENABLE_UPDATER` around the entire body of the function (after `{` and before `}`) in order to avoid warnings of unreachable code (since nothing after the return can be executed).

2) The issue I have with an outlying inverted boolean like ...DISABLE.. this is that it's not logical when you look at it in the list of options. I'd wager that the `#ifdef` is not be read as often as the name of the flag in either the source, cmake output (if that ever shows it) or potentional CI scripts. The implementation would therefore be the better place to handle this logical oddity. But if 1) was applied then there would be no negation anyways.

Felanbird: If the argument is namesake is there not a world where we just call it `CHATTERINO_UPDATER` and have it ON? OFF being no update checks

pajlada: > 1. I believe you would usually wrap #ifdef CHATTERINO_ENABLE_UPDATER around the entire body of the function (after { and before }) in order to avoid warnings of unreachable code (since nothing after the return can be executed).

That's fair, the code should be updated so it's not just an early out

> 2. The issue I have with an outlying inverted boolean like ...DISABLE.. this is that it's not logical when you look at it in the list of options.

True, the option should be marked as advanced so it doesn't show up in GUIs unless the user knows what they're doing

The latest commit means the builds that don't want to use the built-in updater would write
`cmake -DCHATTERINO_UPDATER=OFF ..`

If you open it in gui like `cmake-gui` the option won't show up as to not confuse people.

If `CHATTERINO_UPDATER` is set to *OFF* or *NO* or *0* by the user compiling Chatterino, the compile definition `CHATTERINO_DISABLE_UPDATER` is added (as to not add a compile definition for 99.9% of compiles)

The full `checkForUpdates` code-block is wrapped in `#ifndef CHATTERINO_DISABLE_UPDATER` to make sure there's no unreachable code warnings

pajlada:
Thank you @libklein ! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the About page in Chatterino.

PR #4853, opened 2023-09-30
fix: Fix username tab completion without @

dnsge:
## Description

Fixes usernames not being tab-completed without the `@` prefix (and the "Only search for username autocompletion with an @" setting disabled). 

The completion items being returned by the user source always started with an `@`. The candidate for completion of the query `icel` would be `@icelys, ` which is incompatible due to the extra "@".

Fixes #4850

Issue #3440, opened 2021-12-30
"Emote autocompletion at the start of emote names" setting messes with command autocompletion


Felanbird:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

1. Attempt to use tab auto-completion for a command from the middle of the command 
`e.g. /trihard` `hard`+`tab` - This will not work
2. Toggle off the setting `Only search for emote autocompletion at the start of emote names`
3. Restart Chatterino OR close and re-open the tab with `CTRL+W` & `CTRL+G`
4. Repeat Step 1 - This will now work, but not intentionally

Noticed the issue by accident, I had toggled off the setting for comparison on my main Chatterino 703f3717e for #3437 and while typing a word I hit tab by accident and it changed the word into an unrelated command. 

Also tested on 98e83ad, 2 "commits" behind master and the changes implemented in #3144 did NOT effect this issue. (at least not on the surface)

I see three options this time:

1. Change the title of the setting* to reflect commands & emotes
2. Split this feature into a second setting for command auto-completion
3. Remove this functionality for commands

### OS and Chatterino Version

Chatterino Nightly 2.3.4 98e83ad & 703f3717e | Windows 10

jupjohn:
Cannot reproduce (anymore.)

_Chatterino 2.5.4 (commit [117d06b92](https://github.com/Chatterino/chatterino2/commit/117d06b92)) built with Qt 6.9.2
Running on Arch Linux, kernel: 6.16.8-zen3-1-zen_

PR #4855, opened 2023-10-01
fix: Don't use QCompleter prefix filter

dnsge:
## Description

Toggling the "Only search for emote autocompletion at the start of emote names" would require a user to close and reopen a tab for the change to take effect.

The overhaul put the onus of determining completions completely on the Strategy selected. We don't want to limit ourselves to only completing prefixes.

### DIFF
--- a/.CI/chatterino-installer.iss
+++ b/.CI/chatterino-installer.iss
@@ -2,7 +2,7 @@
 ; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
 
 #define MyAppName "Chatterino"
-#define MyAppVersion "2.4.5"
+#define MyAppVersion "2.4.6"
 #define MyAppPublisher "Chatterino Team"
 #define MyAppURL "https://www.chatterino.com"
 #define MyAppExeName "chatterino.exe"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - master
+      - "bugfix-release/*"
+      - "release/*"
   pull_request:
   workflow_dispatch:
   merge_group:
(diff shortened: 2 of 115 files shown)


Felanbird:
```
Bugfix: - Fixed an issue where the setting `Only search for emote autocompletion at the start of emote names` wouldn't disable if it was enabled when the client started. (#4855)
```

issue could have bee resolved on a per-channel basis without a client restart, but seems like this better explains the bug

Felanbird:
@pajlada current problem is as follows:

When the client is started with the setting `Only search for emote autocompletion at the start of emote names` enabled(default), disabling it does not update the completion model, the tab itself must be closed & re-opened or the entire client restarted.
When the client is started with this setting _disabled_ keeping it disabled and enabling it both work exactly as expected

The changes in this PR resolve the first problem. <br><br>

but, this setting also effects commands see #3440


on 06eb30a5 (before any input completion overhaul):
Enabled on startup = `pains` -> `pains` `popo` -> `popo`
Enabled on startup & disabled | `pains` -> `pains` `popo` -> `popo`
Disabled on startup =  `pains` -> `Pajapains` `popo` -> `/popout`
Disabled on startup & re-enabled = both emotes & commands are properly limited `pains` -> `pains` | `popo` -> `popo`

on 1474ee12 (current commit from https://github.com/Chatterino/chatterino2/pull/4853):
Enabled on startup = `pains` -> `pains` | `popo` -> `popo`
Enabled on startup & disabled | `pains` -> `pains` | `popo` -> `popo`
Disabled on startup =  `pains` -> `Pajapains` `popo` -> `/popout`
Disabled on startup & re-enabled = emote completion respects the setting, commands do not `pains` -> `pains` | `popo` -> `/popout`

on this PR : 
Enabled on startup = `pains` -> `pains` | `popo` -> `/popout`
Enabled on startup & disabled = `pains` -> `Pajapains` | `popo` -> `/popout`
Disabled on startup = `pains` -> `Pajapains` | `popo` -> `/popout`
Disabled on startup & enabled = `pains` -> `pains` | `popo` -> `/popout`



So this PR fixes the issue with commands, but does not fix the problem with commands not respecting the setting, though only partially caused by this PR

brian6932 (approved):
fixes #4892

pajlada (approved):

PR #4856, opened 2023-10-01
Update Nightly to 2.4.6 and get qt6-related-changes from 2.4.6

pajlada:
# Description

This cherry-picks changes from v2.4.6 (the security release), making sure nightly build users also get the same security fixes.
This means nightly will also be for windows 10 and above users now - we'll continue distributing an older build but it is unsafe.


pajlada:
image

PR #4857, opened 2023-10-01
Don't download non-existant qt 5.15 debug symbols for Windows

pajlada:
# Description

thanks @Wissididom


Wissididom:
If wanted, I think you could also like "if out" the `Upload artifact (Windows - symbols)` job, because that produces a warning:
Screenshot from 2023-10-01 14-50-07

PR #4863, opened 2023-10-03
Fix qtkeychain include for Qt6 users

pajlada:
# Description

AUR used to patch this with https://aur.archlinux.org/cgit/aur.git/tree/0001-qt6-qtkeychain.patch?h=chatterino2-git but with this PR it should no longer be necessary


pajlada:
image

PR #4869, opened 2023-10-04
Fix websocketpp depending on boost::random 

kornes:
After boost::random dependency was dropped #4776, vcpkg build started to fail because websocketpp still required it.

Websocketpp fallback to boost polyfill if it cant detect compiler support for C++11 or newer 
https://github.com/zaphoyd/websocketpp/blob/b9aeec6eaf3d5610503439b4fae3581d9aff08e8/websocketpp/common/cpp11.hpp#L48

Detection is based on `__cplusplus` macro, which reports correct values for gcc and clang, but [static and old value for msvc](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170).

This PR adds `/Zc:__cplusplus` compiler option for MSVC, which makes it report actual C++ version instead of `199711L`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1007,6 +1007,7 @@ if (MSVC)
         /wd4505
         /wd4100
         /wd4267
+        # Enable updated '__cplusplus' macro - workaround for CMake#18837
         /Zc:__cplusplus
     )
     # Disable min/max macros from Windows.h
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1007,6 +1007,7 @@ if (MSVC)
         /wd4505
         /wd4100
         /wd4267
+        /Zc:__cplusplus
     )
     # Disable min/max macros from Windows.h
     target_compile_definitions(${LIBRARY_PROJECT} PUBLIC NOMINMAX)


pajlada (approved):
Any thoughts on this @Nerixyz ?

src/CMakeLists.txt
```diff
@@ -1007,6 +1007,7 @@ if (MSVC)
         /wd4505
         /wd4100
         /wd4267
+        /Zc:__cplusplus
```
Nerixyz, line 1010: Nit: add comment as to why we're setting this.

```suggestion
        # Enable updated '__cplusplus' macro - workaround for CMake#18837
        /Zc:__cplusplus
```

Nerixyz (changes requested):
LGTM.

There's a [CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/18837) for this, which will require a policy if this will ever be added.

Arguably, `websocketpp` should set the flag. They [do some compiler detection](https://github.com/zaphoyd/websocketpp/blob/b9aeec6eaf3d5610503439b4fae3581d9aff08e8/CMakeLists.txt#L104-L113), but apparently not enough.

pajlada (approved):

PR #4872, opened 2023-10-06
Update vcpkg to Qt6

kornes:
changes:
- Qt dependencies are trimmed down to whats actually used
- `gtest` and `benchmark` are now togglable features 
- `project()` had to be moved down, because `VCPKG_MANIFEST_FEATURES` needs to be set before cmake enters toolchain file, which happens in project() call
- because vcpkg `windeployqt.exe` is placed in /tools folder instead of /bin, i changed its discovery to find_program(), hopefully it will work in all scenarios :D
- removed `X_VCPKG_APPLOCAL_DEPS_INSTALL` as its not used anymore in qt6 ports
- Qt plugins have runtime dependencies which are not properly deployed by windeployqt:

> imageformats/qjpeg.dll
> └── jpeg62.dll
> 	
> imageformats/qwebp.dll
> ├── libwebpdemux.dll
> ├── libwebpmux.dll
> └── libwebp.dll
>      └── libsharpyuv.dll
> 
> tls/qopensslbackend.dll
> ├── libssl-3-x64.dll
> └── libcrypto-3-x64.dll

OpenSSL dlls are being already installed by cmake `RUNTIME_DEPENDENCIES`, I added remaining 5 files as optional installs.
- vcpkg puts release and debug artifacts in different location, which required extra steps to make Debug installs work
> ├── .../bin
> ├── .../Qt6/plugins
> ├── .../debug/bin
> └── .../debug/Qt6/plugins
1. to make windeployqt use correct folder, debug installs use dedicated script [`windeployqt.debug.bat`](https://github.com/microsoft/vcpkg/blob/cc97b4536ae749ec0e4f643488b600b217540fb3/ports/qtbase/windeployqt.debug.bat) provided by qt6 port
2. to copy bin dependencies from correct folder, `LOCATION_DEBUG` property is used to get path for debug /bin
- baseline updated up until Qt 6.5.2


pajlada (commented):
Looks reasonable to me, let me know when you want a full review.

As an aside, I'm open to adding vcpkg as a builder in CI, even if we don't publish its artifact, just to make sure things keep building as expected

Nerixyz (approved):
👍 Works for me with conan and a regular Qt build.

> i changed its discovery to find_program(), hopefully it will work in all scenarios :D

Seems to work without having Qt's `bin` directory in `PATH`.

---

I found some possible improvements when running `cmake --install .`. Not sure if these should be done here or in another PR. Technically, we could also use `cmake --install . --prefix Chatterino2` in CI (would save a line in `conanfile.py` but invalidate the cache).

kornes:
thanks for testing, @Nerixyz could you please test Debug installs after changes in 0972f14c59b13872d859016aa1c6e85dc56e8924.

Nerixyz (approved):
👍 Still works fine in debug mode.

pajlada (approved):

PR #4875, opened 2023-10-07
Display all parsed elements when parsing emojis in replies

### LINKED ISSUE #4874: Replying to a message with an emoji followed by any text, without any spaces, will cause the text following the emoji to be eaten in the reply
### Checklist

### Describe your issue

```
▪pajaDank
▪ pajaDank test with a space
pajaDank test with no emoji
```

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit 75282579) built on 2023-10-06 with Qt 6.5.0, MSVC 193532217 Running on Windows 10 Version 21H2, kernel: 10.0.19044

Nerixyz:
# Description

Previously, only the first parsed element was used. This also adds a helper for Boost's and `std`'s `apply_visitor`/`visit`, which ensures all alternative types from the variants are handled.

Fixes #4874.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Bugfix: Fixed too much text being copied when copying chat messages. (#4812, #4830, #4839)
 - Bugfix: Fixed empty page being added when showing out of bounds dialog. (#4849)
 - Bugfix: Fixed issue on Windows preventing the title bar from being dragged in the top left corner. (#4873)
-- Bugfix: Fixed emojis getting lost in replies. (#4875)
+- Bugfix: Fixed an issue where reply context didn't render correctly if an emoji was touching text. (#4875)
 - Dev: Fixed UTF16 encoding of `modes` file for the installer. (#4791)
 - Dev: Temporarily disable High DPI scaling on Qt6 builds on Windows. (#4767)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Bugfix: Fixed too much text being copied when copying chat messages. (#4812, #4830, #4839)
 - Bugfix: Fixed empty page being added when showing out of bounds dialog. (#4849)
 - Bugfix: Fixed issue on Windows preventing the title bar from being dragged in the top left corner. (#4873)
+- Bugfix: Fixed emojis getting lost in replies. (#4875)
```
Felanbird, line 21:
```suggestion
- Bugfix: Fixed an issue where reply context didn't render correctly if an emoji was touching text. (#4875)
```

Felanbird (approved):
works on 🅱️ indows, unlucky for the other ones that didn't build

Nerixyz:
Seems like we _do_ need the deduction guidelines even though we're on C++ 20.

pajlada (changes requested):
You can keep the Overloaded type, but since we have another big PR making changes here it'd be best if you just add the loop to fix the issue

pajlada (approved):

PR #4877, opened 2023-10-08
Replace boost::optional with std::optional

pajlada:
# Description

The commits are split up into two parts:
1. The actual replacing of boost::optional with std::optional
2. The small refactors I've done around the code as the replacing took place - there weren't many things I changed, mostly just


pajlada (commented):
Looks good to me

image

PR #4879, opened 2023-10-09
perf: prevent unnecessary layouts at leaveEvent

kornes:
Currently `ChannelView::leaveEvent` triggers 2 layouts every time mouse leaves. 

- `ChannelView::unpause` now returns if channelview was not paused 
- removed unnecessary `queueLayout`, i didnt find practical use to force it here, tested for 1h with no issues


pajlada (approved):
Seems to work as expected, can let nightly users test it some more

PR #4889, opened 2023-10-11
perf: reduce repaints amount caused by selection

kornes:
PR reduces amount of painting while not changing behavior/visuals of text selection
- removed `selecting_` bool, only used by triple click selection, which forced layout on each click release.
- fixed broken `SelectionItem` != operator which worked like ==, only used by already removed `selecting_` so no side effects
- moved triple click selection due to above changes
- changed `ChannelView::setSelection` to repaint/emit signal only if selection was updated

tested:
- all consumers of `selectionChanged` signal
- single / double / triple click selection


src/widgets/helper/ChannelView.cpp
```diff
@@ -1980,6 +1972,15 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
             {
                 return;
             }
+
+            // Triple-clicking a message selects the whole message
+            if (this->clickTimer_->isActive() &&
+                (fabsf(distanceBetweenPoints(this->lastDClickPosition_,
+                                             event->screenPos())) < 10.f))
+            {
+                this->selectWholeMessage(layout.get(), messageIndex);
```
Nerixyz, line 1981: This can segfault if you double-click on the bottom of the last message, then move the cursor below the last message (distance must be less than 10px). Probably needs a check if `foundElement` is true.

src/messages/Selection.hpp
```diff
@@ -38,7 +38,7 @@ struct SelectionItem {
 
     bool operator!=(const SelectionItem &b) const
     {
-        return this->operator==(b);
+        return !this->operator==(b);
     }
```
Nerixyz, line 42: As we're on C++ 20, we could remove the `operator!=` as it's synthesized, but not sure if all our compilers support this yet 🙃.

Nerixyz (changes requested):
Aside from the segfault, this PR works for me. You really notice the absence of re-layouts when selecting in debug builds 👍.

Nit: This PR reduces the amount of (re-)layouts performed when selecting, the widget is still repainted (about) 50 times per second.

src/widgets/helper/ChannelView.cpp
kornes: fixed in 662c7ab88a48254d65eb47fad1c6a7edf91bf767, thanks!

src/messages/Selection.hpp
kornes: by quick look 22 files define own `!=`, something nice to explore for sure

pajlada: I made this into an issue by itself https://github.com/Chatterino/chatterino2/issues/4895 with a new "Refactor" label - if you see any other small things like that feel free to make issues with that label

pajlada (approved):

PR #4897, opened 2023-10-13
Add some more filter tests

pajlada:
# Description

This changes the `make coverage` function to use `gcovr` instead of `lcov`, and to have it generate an html file directly at `coverage/index.html` under the build directory

The only thing this changes, other than adding tests, is making the `Expression` class pure virtual. Every derived class should implement each of the functions

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -679,15 +679,15 @@ endif()
 if (CHATTERINO_GENERATE_COVERAGE)
     include(CodeCoverage)
     append_coverage_compiler_flags_to_target(${LIBRARY_PROJECT})
-    target_link_libraries(${LIBRARY_PROJECT} PUBLIC gcov)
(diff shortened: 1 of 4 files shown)


pajlada (commented):
image

src/CMakeLists.txt
```diff
@@ -679,15 +679,15 @@ endif()
...
+    setup_target_for_coverage_gcovr_html(
         NAME coverage
-        EXECUTABLE ./bin/chatterino-test
-        BASE_DIRECTORY ${PROJECT_SOURCE_DIR}/src
+        EXECUTABLE ctest
         EXCLUDE "/usr/include/*"
         EXCLUDE "build-*/*"
         EXCLUDE "lib/*"
+        EXCLUDE "*/ui_*.h"
+        EXCLUDE "*/moc_*.cpp"
```
Nerixyz, line 690: Should also exclude `*/*.moc`.

PR #4898, opened 2023-10-13
refactor: simplify double click selection

kornes:
Selection has 2 modes, single click starts single character matching at cursor move and double click starts whole word matching at cursor move. This PR refactors double click selection to use simpler approach: creates unified selection between hovered word and starting word.

1a0b611b3c23de9d382162c9589f89c137b5194d also fixes bug present on master, where double-click selection would move at each new message due to lack of index shifting. This could be separated PR but included it here for easier testing.


src/messages/Selection.hpp
```diff
@@ -72,6 +73,13 @@ struct Selection {
         return !this->operator==(b);
     }
 
+    //union of both selections
+    Selection operator|=(const Selection &b) const
```
Nerixyz, line 77: This should probably be `operator|`, since it doesn't assign.

```suggestion
    Selection operator|(const Selection &b) const
```

Note that this messes with `start` and `end` a bit, which denote the start/end point of the selection. When double-clicking, however, we have two starting positions (word start & end), so these become meaningless, but there's nothing using them for double-click-selection right now.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1512,15 +1518,30 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
         this->currentMousePosition_ = event->screenPos();
     }
 
-    // is selecting
+    // check for word underneath cursor
+    const MessageLayoutElement *hoverLayoutElement =
+        layout->getElementAt(relativePos);
+
+    //selecting single characters
     if (this->isLeftMouseDown_)
```
Nerixyz, line 1526: We don't need to do this work if we're double-clicking and `isDoubleClick_` ⇒ `isLeftMouseDown_`.

```suggestion
    if (this->isLeftMouseDown_ && !this->isDoubleClick_)
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -1512,15 +1518,30 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
...
     }
 
+    //selecting whole words
+    if (this->isDoubleClick_ && hoverLayoutElement)
+    {
+        auto [wordStart, wordEnd] =
+            this->getWordBounds(layout.get(), hoverLayoutElement, relativePos);
+        auto wordSelection = Selection{SelectionItem(messageIndex, wordStart),
+                                       SelectionItem(messageIndex, wordEnd)};
+        auto selectUnion = this->doubleClickSelection_ |= wordSelection;
```
Nerixyz, line 1540:
```suggestion
        auto selectUnion = this->doubleClickSelection_ | wordSelection;
```

Maybe a comment here would be helpful, because `doubleClickSelection_` never changes when moving the mouse (and no messages are added). Something along the lines of _"doubleClickSelection marks the initially selected word. selectUnion spans the selection from the initial word to the current word."_.

src/widgets/helper/ChannelView.hpp
```diff
@@ -210,10 +210,11 @@ class ChannelView final : public BaseWidget
 
     void drawMessages(QPainter &painter);
     void setSelection(const SelectionItem &start, const SelectionItem &end);
+    void setSelection(const Selection &newSelection);
     void selectWholeMessage(MessageLayout *layout, int &messageIndex);
-    void getWordBounds(MessageLayout *layout,
-                       const MessageLayoutElement *element,
-                       const QPoint &relativePos, int &wordStart, int &wordEnd);
+    std::pair<size_t, size_t> getWordBounds(MessageLayout *layout,
```
Nerixyz, line 215: Some doc comment might be useful.
```suggestion
    /// @returns [wordStart, wordEnd]
    std::pair<size_t, size_t> getWordBounds(MessageLayout *layout,
```

src/widgets/helper/ChannelView.hpp
```diff
@@ -307,7 +308,7 @@ class ChannelView final : public BaseWidget
     bool isLeftMouseDown_ = false;
     bool isRightMouseDown_ = false;
     bool isDoubleClick_ = false;
-    DoubleClickSelection doubleClickSelection_;
+    Selection doubleClickSelection_{};
```
Nerixyz, line 311: It's already default initialized. Maybe it can be moved to `selection_`?

```suggestion
    Selection doubleClickSelection_;
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -2892,17 +2754,17 @@ void ChannelView::selectWholeMessage(MessageLayout *layout, int &messageIndex)
...
-    const int mouseInWordIndex = element->getMouseOverIndex(relativePos);
-    wordStart = layout->getSelectionIndex(relativePos) - mouseInWordIndex;
+    const auto wordStart = layout->getSelectionIndex(relativePos) -
+                           element->getMouseOverIndex(relativePos);
     const int selectionLength = element->getSelectionIndexCount();
     const int length =
         element->hasTrailingSpace() ? selectionLength - 1 : selectionLength;
-    wordEnd = wordStart + length;
+
+    return {wordStart, wordStart + length};
```
Nerixyz, line 2767: I wonder if we can use `size_t` for everything here without running into overflow problems.

Nerixyz (commented):
Works as intended 👍

src/widgets/helper/ChannelView.cpp
kornes: everything will use size_t now ce10b811b024fe625c0f7b8c3f997b7c3aff659f

src/widgets/helper/ChannelView.hpp
kornes: done a034cef8b03b6399dcc2d1f259247d875b1c0122

src/widgets/helper/ChannelView.hpp
kornes: comment added ce10b811b024fe625c0f7b8c3f997b7c3aff659f

src/widgets/helper/ChannelView.cpp
kornes: added comment and renamed wordSelection to show intent better 3126d1dc2aa561075cd96545e52218ca06a0f384

src/widgets/helper/ChannelView.cpp
kornes: i didnt add it originally because both bools appears to be mutually exclusive, when tracing `isLeftMouseDown_` in `mouseDoubleClickEvent` it never appears to be true. Should we add this to possibly future proof?

src/messages/Selection.hpp
kornes: op type changed 008fa5b5ad36782c835b2451f43d9b22cc602099, 
can you elaborate on issue with start/end? would be best to have it work in all scenarios, even if currently unused

src/widgets/helper/ChannelView.cpp
Nerixyz: If they're mutually exclusive, it's fine.

pajlada:
Could you record how to make the bug appear on master? I can't reproduce it

kornes:
tested on release 2.4.6 (https://github.com/Chatterino/chatterino2/commit/eb8c7f2d)

steps:
1. start selection from double click on word
1. left mouse click needs to be down (selecting) 

- moving chat bug: 
https://kappa.lol/VKEIq (file too big for github)
change in https://github.com/Chatterino/chatterino2/commit/1a0b611b3c23de9d382162c9589f89c137b5194d fixes this problem in current PR

- static chat bug, unrelated to above: 
https://github.com/Chatterino/chatterino2/assets/28986062/916ead56-4880-4c01-8cb3-b140add80991
seems like hovering same word in other message resets selection.
This problem is not present in this PR, fixed by refactor.

pajlada (approved):

PR #4899, opened 2023-10-14
refactor: Toast

Nerixyz:
# Description

* Fixes a bug where avatars weren't loaded on fresh installations of Chatterino.
* Avatars now update every two weeks.
* Removes misleading `DownlaodManager` (now part of `Toasts.cpp`).
* Refactors usage of WinToast to be easier to read.
* Added version to AUMI.
* Removes manual `QString` → `std::wstring` conversions.
* Removes uses of implicit ASCII casts in `Toasts.cpp`, meaning it can be compiled with `QT_NO_CAST_FROM_ASCII`.


src/singletons/Toasts.cpp
```diff
@@ -49,24 +97,18 @@ bool Toasts::isEnabled()
...
-    auto iterator = Toasts::reactionToString.find(reaction);
-    if (iterator != Toasts::reactionToString.end())
-    {
-        return iterator->second;
-    }
-    else
-    {
-        return DONT_OPEN;
-    }
+    return reactionToString(reaction);
```
pajlada, line 100: Just inline the `reactionToString` function in here

src/singletons/Toasts.cpp
```diff
@@ -131,95 +172,143 @@ class CustomHandler : public WinToastLib::IWinToastHandler
...
+        if (this->file_.isOpen())
+        {
+            this->file_.close();
+        }
+        emit downloadComplete();
+        this->deleteLater();
+    });
+}
+
+#include "Toasts.moc"
```
pajlada, line 312: What is this?

pajlada (changes requested):

src/singletons/Toasts.cpp
Nerixyz: This is from the `AvatarDownloader : QObject`.

pajlada: yuck but I assume this is because it's a class QObject in a source file

pajlada (approved):

PR #4904, opened 2023-10-19
Add selection tests

kornes:
Adds 100% cover for Selection.hpp

### DIFF
--- a/tests/src/Selection.cpp
+++ b/tests/src/Selection.cpp
@@ -127,8 +127,10 @@ TEST(Selection, ShiftMessageIndex)
 
     a.shiftMessageIndex(10);
     ASSERT_EQ(a.selectionMin, a.start);
+    ASSERT_EQ(a.selectionMin.charIndex, 1);
     ASSERT_EQ(a.selectionMin.messageIndex, 90);
     ASSERT_EQ(a.selectionMax, a.end);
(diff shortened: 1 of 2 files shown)


tests/src/Selection.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    {
+        auto a = Selection{{3, 3}, {2, 1}};
+        auto b = Selection{{1, 0}, {1, 10}};
+        auto ab = a | b;
+
+        ASSERT_EQ(ab.start, b.start);
+        ASSERT_EQ(ab.end, a.start);
+    }
+}
```
Nerixyz, line 113: This should test that `(a | b) == (b | a)`.

Maybe it could test for `selectionMin/Max` as well?

Nerixyz (commented):
There are quite a few tests now, maybe some directories should be added to group tests?

tests/src/Selection.cpp
kornes: added commutative check 88bff8895a72091b8e8a7749fac494cb4ea5787a
I moved select min/max to its own test in 77b135a3e6ae85aaa4df301c58b11343805dac39, if the other test pass there is nothing else to test here.

pajlada (approved):

PR #4906, opened 2023-10-19
Remove direct dependency on Qt 5 compatibility module

Nerixyz:
# Description

This removes the last use of `QStringRef` in the code by ifdef-ing it. I wasn't sure if I could use a function in the tests to create a `QStringRef` as opposed to a macro, but it should be removed soon-ish anyway.


pajlada:
Tests can be less clean too, that's ok

pajlada (approved):

PR #4914, opened 2023-10-23
Use custom frame for main window only

kornes:
We use custom frame to draw extra controls in title bar on Windows. Currently few windows beside main window also use `BaseWindow::EnableCustomFrame` flag. This PR removes use of custom frame outside of `Window.cpp`, so only main chatterino window and popup chats are left.

Since Qt6.4 system theme and user preferred colors are used for title bars by default. 

other changes: 
- `BasePopup` is always enabling [Qt::Dialog](https://doc.qt.io/qt-6/qt.html#WindowType-enum) flag, which removes minimize/maximize buttons. This is undesired for `EmotePopup`, so i added bool to disable that behavior for that single window. This is possibly breaking change for Linux WMs, so needs test, didn't want to ifdef right away.

### DIFF
--- a/src/widgets/BasePopup.cpp
+++ b/src/widgets/BasePopup.cpp
@@ -6,9 +6,9 @@
 
 namespace chatterino {
 
-BasePopup::BasePopup(FlagsEnum<Flags> _flags, QWidget *parent,
+BasePopup::BasePopup(FlagsEnum<Flags> flags_, QWidget *parent,
                      bool addDialogFlag)
-    : BaseWindow(addDialogFlag ? _flags | Dialog : _flags, parent)
+    : BaseWindow(addDialogFlag ? flags_ | Dialog : flags_, parent)
 {
 }
 
--- a/src/widgets/BasePopup.cpp
+++ b/src/widgets/BasePopup.cpp
@@ -6,8 +6,9 @@
 
 namespace chatterino {
 
-BasePopup::BasePopup(FlagsEnum<Flags> _flags, QWidget *parent)
-    : BaseWindow(_flags | Dialog, parent)
+BasePopup::BasePopup(FlagsEnum<Flags> _flags, QWidget *parent,
+                     bool addDialogFlag)
+    : BaseWindow(addDialogFlag ? _flags | Dialog : _flags, parent)
 {
 }
 
--- a/src/widgets/BasePopup.hpp
+++ b/src/widgets/BasePopup.hpp
@@ -11,7 +11,7 @@ class BasePopup : public BaseWindow
 {
 public:
     explicit BasePopup(FlagsEnum<BaseWindow::Flags> flags_ = None,
-                       QWidget *parent = nullptr);
+                       QWidget *parent = nullptr, bool addDialogFlag = true);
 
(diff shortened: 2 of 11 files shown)


Nerixyz (commented):
At least for the draggable popups, this does make it look a bit weird:



The background for the _Subscribe to thread_ looks a bit off now, as the window frame doesn't have the same color anymore.

Otherwise, this looks good.

fourtf:
This seems to make the frames of secondary windows look inconsistent with the main window. What's the reasoning behind this change?

kornes:
windows on master already using native frame: settings, search, filters select
this PR changes frame for:
- emote popup
- thread/user popup (if not set to close at lost focus)
- select channel dialog
- color picker (general settings page)
- update dialog
- license window (about page)
- tutorial video window (split -> how to)

not changed: 
- main window, secondary windows with chats
- any window using `BaseWindow::Frameless` which overrides customframe flag

**Motivation:**
- main purpose of custom frame is to draw things which are not possible on native titlebar, windows affected by this PR doesn't draw custom stuff on their titlebars.
- visually, native window will also use dark color by default:

| Before  | After |
| ------------- | ------------- |
| b1  |a1 |
| a1 | a2|

unless user specifically setup their system to use accent colors for title bars, then it's best to respect the color settings specified by the user.
- letting OS draw title bars makes resizing more fluent

| Before  | After |
| ------------- | ------------- |
| https://i.nuuls.com/yxNYT.mp4 | https://i.nuuls.com/RiyYk.mp4 |
|notice tearing/shadowing and occasional appearing of win7 frame | |
- title bar height will be consistent with rest of OS
- double click to maximize doesn't work in emote popup, while it works in main window. From what i recall issue was connected with calling `QWidget::winId()` and results it have on custom frame handling
- moving native window between monitors doesn't cause resize/redraw we have in custom frame
- this PR was spawned by my research for adding support for [win11 snap layouts](https://support.microsoft.com/en-us/windows/snap-your-windows-885a9b1e-a983-a3b1-16cd-c531795e6241#WindowsVersion=Windows_11), custom frame will need quite a few changes to make it work, because it doesn't hint proper system buttons on hover/hit test.

tl;dr why sacrifice superior native feel, if window doesn't draw custom buttons. Future OS updates will introduce more maintenance on custom frame, so i wanted to minimize its use to just bare minimum we need.

Nerixyz:
> this PR was spawned by my research for adding support for win11 snap layouts, custom frame will need quite a few changes to make it work, because it doesn't hint proper system buttons on hover/hit test.

I wonder if we could fix this by returning `HTMAXBUTTON` (and others) in `NCHITTEST`.

kornes:
yeah obviously its fixable and we gonna have it working on main window either by your or mine effort. 
I gave it as example of issues connected with own custom frame - its sacrifice you need to take for custom buttons in title, basically reinvent the wheel and 50 edge cases. But there is little reason to use it outside of main window.

pajlada:
Currently I don't think I want to touch this, until there's better understanding of what all of our various windows/popups/dialogs/tooltips are, and whether custom framing of them makes sense

PR #4915, opened 2023-10-24
refactor: Message

pajlada:
# Description

this PR was cherry-picked into smaller PRs

 - [x] Virtual-override-related changes: https://github.com/Chatterino/chatterino2/pull/4917
 - [x] FilterParser-changes: https://github.com/Chatterino/chatterino2/pull/4925
 - [x] ChannelView changes: https://github.com/Chatterino/chatterino2/pull/4926
 - [x] IrcMessageHandler changes: https://github.com/Chatterino/chatterino2/pull/4927

With the last remaining changes to `Message` & some other small changes staying here

Message: Remove empty anon namespace
Message: Remove else after return
Message: Avoid repeating type in return
Message: Remove unused includes (there were a few)
Helix: Remove unneccesary `static` for already-static variables
AttachedWindow: Remove unused ForwardDecl.hpp include

### DIFF
--- a/src/messages/Message.cpp
+++ b/src/messages/Message.cpp
@@ -33,26 +33,30 @@ SBHighlight Message::getScrollBarHighlight() const
     {
         return SBHighlight(this->highlightColor);
     }
-    else if (this->flags.has(MessageFlag::Subscription) &&
-             getSettings()->enableSubHighlight)
+
+    if (this->flags.has(MessageFlag::Subscription) &&
+        getSettings()->enableSubHighlight)
     {
         return SBHighlight(
             ColorProvider::instance().color(ColorType::Subscription));
     }
-    else if (this->flags.has(MessageFlag::RedeemedHighlight) ||
-             this->flags.has(MessageFlag::RedeemedChannelPointReward))
+
+    if (this->flags.has(MessageFlag::RedeemedHighlight) ||
+        this->flags.has(MessageFlag::RedeemedChannelPointReward))
     {
         return SBHighlight(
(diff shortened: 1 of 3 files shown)


Mm2PL:
image
lgtm

src/messages/Message.cpp
```diff
@@ -26,45 +19,57 @@ Message::~Message()
...
+    if (this->flags.has(MessageFlag::FirstMessage))
     {
-        return SBHighlight(
+        return {
             ColorProvider::instance().color(ColorType::FirstMessageHighlight),
-            SBHighlight::Default, false, true);
+            ScrollbarHighlight::Default,
+            false,
+            true,
+        };
```
Nerixyz, line 69: Not from this PR: The `ScrollbarHighlight` needs some refactoring. The arguments `isRedeemedHighlight`,
`isFirstMessageHighlight`, and `isElevatedMessageHighlight` are mutually exclusive - shouldn't this be refactored to an enum? Even if they're not mutually exclusive, a bit field would be more readable.

Nerixyz (commented):
For a future refactoring: There are a few `#pragma comment(lib, "my.lib")` in the code (e.g. in _AttachedWindow.cpp_). These can be replaced by `target_link_libraries(<target> PRIVATE my.lib)` (maybe there's even a better way of doing this in CMake).

src/messages/Message.cpp
pajlada: Definitely should be enums yeah

pajlada (commented):
image

PR #4917, opened 2023-10-24
refactor: Use override instead of virtual where possible

pajlada:
# Description

This PR fixes most warnings from the clang-tidy `cppcoreguidelines-explicit-virtualfunctions` check

1. For base classes implementing virtual functions, just use virtual
2. For derived classes overriding a base class virtual function, just use override since override implies virtual

Part of #4915

## Unrelated change(s)

Irc2: Add some newlines to let the anon namespace breathe
LambdaRunnable: Initialize action_ in member initializer list
IrcChannel: Made the class final
TwitchChannel: Made the class final

## How to review

Most changes are just removing virtual & the reformatting that comes with it.
TwitchChannel had 3 virtual functions that were never overwritten, `refresh{BTTV,FFZ,SevenTV}ChannelEmotes`. I removed virtual complete there.
Made `LambdaRunnable`'s `run` function override - it was not override or virtual before, but run is most likely meant to override `QRunnable`'s `run` function.
Made `GeneralPage`'s `filterElements` function override
`ComboBoxItemDelegate` now uses override instead of virtual, since they override functions from `QStyledItemDelegate`

### DIFF
--- a/src/common/SignalVectorModel.hpp
+++ b/src/common/SignalVectorModel.hpp
@@ -101,7 +101,7 @@ public:
         return this;
     }
 
-    virtual ~SignalVectorModel()
+    ~SignalVectorModel() override
     {
         for (Row &row : this->rows_)
         {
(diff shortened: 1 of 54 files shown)


pajlada:
I have looked through all changes in this PR and they look good to me - I'll leave it up for a bit to give people time to look through it before I merge it in.

image

PR #4919, opened 2023-10-25
feat: `/reply` command now replies to the latest message of the user

iProdigy:
Keep thread continuity when using `/reply` command by replying directly to the message id rather than the root id.

Since Twitch [updated](https://twitter.com/TwitchSupport/status/1687193841240059905) their reply logic, this approach shows the correct parent username/body now

Screenshot_20231024_213424


Nerixyz (commented):
Not really an issue from this PR, but it might be confusing that messages from nested threads don't get added to the original thread:

Felanbird (approved):
As noted by neri visually it would be nice if threads were nested, but this behavior is not any different than what Chatterino users currently experience from 7tv browser chatters, and also not for this PR.

pajlada (approved):

PR #4281, opened 2023-01-01
Show follower emotes outside of channel if you're a subscriber

pajlada:
Pull request checklist:

- [x] `CHANGELOG.md` was updated, if applicable

# Description

PR #4922, opened 2023-10-28
Handle follower emotes differently if subscribed

### LINKED ISSUE #4233: Update Follower Emote Handling
Twitch has announced an update to how follower emotes work. https://twitter.com/TwitchSupport/status/1603125305010364417

You are now be able to post follower emotes anywhere, if you are currently subscribed to said channel.
image

In Chatterino we originally made sure follower emotes only appeared in the channel they came from, as that channel was the only channel they worked in. We will now need to update them to respect this new handling.
image

baines:
If we also have a subscriber emote set for a channel, then don't treat a follower emote set as local to that channel.

# Description

On Twitch it's possible to use a channel's follower emotes everywhere if also subscribed to the followed channel. 
But Chatterino did not show the emotes as available in other channels or allow them to be auto-completed in this case.

This PR makes a small change so that if we also have a subscriber emote set for a channel, then the follower emote set is not treated as `local`.

I know there's another PR to implement this behaviour (https://github.com/Chatterino/chatterino2/pull/4281) but I'm not confident enough with the changes in there to pick that one up.

This code doesn't attempt to convert an existing local set into a non-local one or vice versa (when subscription is gained/lost). I don't know if twitch sends another GLOBALUSERSTATE with the emote sets when that happens or what.

It's also making an assumption that `loadUserstateEmotes` is called only once after all the emote sets are known. If they can come in "piecemeal" then it might not find the subscriber set if it's in a later batch than the follower one.

(felanbird): Closes #4233


Felanbird:
Those suggestions are all shit you can go ahead and revert them, the bot doesn't always get it right

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -264,6 +264,25 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
...
                 auto localEmoteData = this->localEmotes_.access();
+
+                std::set<QString> subscriberChannelIDs;
+                for (auto emoteSet : emoteSetArray)
+                {
+                    IvrEmoteSet ivrEmoteSet(emoteSet.toObject());
+                    if (!ivrEmoteSet.tier.isNull())
+                    {
+                        subscriberChannelIDs.insert(ivrEmoteSet.channelId);
+                    }
```
iProdigy, line 275: This ends up deserializing every emote set twice (once on L271 & once on 290).

Computing all `subscriberChannelIDs` upfront is reasonable if you're concerned about follower emote sets appearing before the corresponding subscriber emote set. In reality, `emote-sets` in `GLOBALUSERSTATE` has the UUIDs are the end, so this is not a concern. (Also the IVR endpoint response has the emote sets in the same order as the request, so misordering wouldn't occur from here either)

Thus, you could move the [L272-L275](https://github.com/Chatterino/chatterino2/blob/6e0c9d611dfbf9d45acb0e94a904c94477a2f16e/src/providers/twitch/TwitchAccount.cpp#L272-L275) logic after [L290](https://github.com/Chatterino/chatterino2/blob/6e0c9d611dfbf9d45acb0e94a904c94477a2f16e/src/providers/twitch/TwitchAccount.cpp#L290) so deserialization only occurs once.

Alternatively, if we're afraid of twitch no longer sorting `emote-sets`, we should consider deserializing once and storing to a vector for reuse

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -285,8 +304,13 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
                         continue;
                     }
 
+                    emoteSet->channelID = ivrEmoteSet.channelId;
                     emoteSet->channelName = ivrEmoteSet.login;
                     emoteSet->text = ivrEmoteSet.displayName;
+                    emoteSet->subscriber = !ivrEmoteSet.tier.isNull();
+
+                    bool haveSubscriberSetForChannel =
+                        subscriberChannelIDs.contains(ivrEmoteSet.channelId);
```
iProdigy, line 313: Edge case: if user is subscribed to a channel that offers follower emotes but no sub emotes, `haveSubscriberSetForChannel` would be false, leading to `local = true`

that said, this edge case is rare, the PR is a significant improvement, and fixing this edge case is likely not worth the complexity/additional IVR requests

iProdigy (commented):
nice work

iProdigy:
Also, we could hook on subscription event to update `local` status for the relevant follower emotes (but this can happen in a separate PR)

Nerixyz (commented):
Seems good. Just a note: Twitch announced a _Get User Emotes_ endpoint in their [Standard Output](https://www.twitch.tv/videos/1953435059?t=00h29m36s) stream. However, it isn't clear how this will handle follower emotes.

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -264,6 +264,25 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
             [this, weakChannel](QJsonArray emoteSetArray) {
                 auto emoteData = this->emotes_.access();
                 auto localEmoteData = this->localEmotes_.access();
+
+                std::set<QString> subscriberChannelIDs;
```
Nerixyz, line 268: This could be an unordered set.

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -264,6 +264,25 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
...
+                for (auto emoteSet : emoteSetArray)
+                {
+                    IvrEmoteSet ivrEmoteSet(emoteSet.toObject());
+                    if (!ivrEmoteSet.tier.isNull())
+                    {
+                        subscriberChannelIDs.insert(ivrEmoteSet.channelId);
+                    }
+                }
+
+                for (auto emoteSet : emoteData->emoteSets)
```
Nerixyz, line 278: As clang-tidy said, this should be a const reference.

Nerixyz (approved):

src/providers/twitch/TwitchAccount.cpp
Felanbird: I wasn't sure if that one was good, usually const suggestions are poo but I did notice that one referenced a different rule

Nerixyz: It's a bit confusing. Clang-tidy warns when local variables aren't declared `const` which we don't want (I'm pretty sure there was a PR that fixed this - [`misc-const-correctness.AnalyzeValues`](https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html#cmdoption-arg-AnalyzeValues)).

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Allow use of Twitch follower emotes in other channels if subscribed. (#4922)
```
pajlada, line 5:
```suggestion
- Major: Allow use of Twitch follower emotes in other channels if subscribed. (#4922)
```

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -302,7 +326,8 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
                             getApp()->emotes->twitch.getOrCreateEmote(id, code);
 
                         // Follower emotes can be only used in their origin channel
```
Felanbird, line 328:
```suggestion
                        // Follower emotes can be only used in their origin channel
                        // unless the user is subscribed, then they can be used anywhere.
```

src/providers/twitch/TwitchAccount.cpp
baines: Originally I tried storing the `IvrEmoteSet`s in a `std::vector` then sorting it so that the subscriber ones came first. But i was hitting errors about `operator=(&&)` being deleted that I didn't fully understand, so just deserialised it twice instead.

iProdigy: When storing to a vector you don't need to sort it (even if we assume twitch doesn't send it already sorted), just iterate twice (as $O(2n) = O(n)$): first iteration does deserialization & populates `subscriberChannelIDs`, second iteration has the remaining logic

Felanbird (approved):
I'm happy with this change, prevent the double serialization on merge if anyone really cares (they probably do) but I'm glad to have a working solution to this.

src/providers/twitch/TwitchAccount.cpp
```diff
@@ -264,11 +264,31 @@ void TwitchAccount::loadUserstateEmotes(std::weak_ptr<Channel> weakChannel)
             [this, weakChannel](QJsonArray emoteSetArray) {
                 auto emoteData = this->emotes_.access();
                 auto localEmoteData = this->localEmotes_.access();
-                for (auto emoteSet_ : emoteSetArray)
+
+                std::unordered_set<QString> subscriberChannelIDs;
+                std::vector<IvrEmoteSet> ivrEmoteSets;
```
iProdigy, line 269:
```suggestion
                std::vector<IvrEmoteSet> ivrEmoteSets;
                ivrEmoteSets.reserve(emoteSetArray.size());
```

optimization to avoid expensive resizing

iProdigy (approved):
Great first contribution!

pajlada (approved):

pajlada:
@Mm2PL want to take a peek?

Mm2PL (approved):
LGTM, works as expected. Thank you for your contribution.

pajlada:
Thank you @baines! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the About page in Chatterino.

PR #4926, opened 2023-10-29
refactor: ChannelView

pajlada:
# Description

- comment getElementAt function
- Use relative-to-src include
- Remove unused wordFlagsChanged slot
- Refactor padding variables
- Make scrollUpdateRequest a normal function instead of a slot
- add missing parameter names
- Make clickTimer_ a value instead of a pointer
- Add const to auto & where possible
- move static functions to the better namespace
- Make getWordBounds return a pair of the bounds instead of accepting it by reference
- Clean up pause timer function
- Make addImageContextMenuItems & addLinkContextMenuItems static
- Make addHiddenContextMenuItems static
- auto * where possible
- Remove const in return value
- add braces where possible
- Use capital F when making a float literal
- use uint32_t instead of uint in pauses
- No else after return
- Remove pessimistic move
- Make setChannel const &sharedptr
- Be more explicit about scroll events we ignore
- Refactor menu item add functions
- JumpToMessage logic
- Refactor thread creation

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -2069,7 +2069,8 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
 
             return;
         }
-        else if (foundElement)
+
+        if (foundElement)
         {
             const MessageLayoutElement *hoverLayoutElement =
                 layout->getElementAt(relativePos);
@@ -2078,14 +2079,14 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
             {
                 return;
             }
-            else if (hoverLayoutElement->getFlags().has(
-                         MessageElementFlag::Username))
+            if (hoverLayoutElement->getFlags().has(
+                    MessageElementFlag::Username))
             {
                 openTwitchUsercard(this->channel_->getName(),
                                    hoverLayoutElement->getLink().value);
                 return;
             }
-            else if (hoverLayoutElement->getLink().isUrl() == false)
+            if (hoverLayoutElement->getLink().isUrl() == false)
             {
                 return;
             }
@@ -2889,12 +2890,10 @@ void ChannelView::setInputReply(const MessagePtr &message)
             {
                 return threadIt->second.lock();
             }
-            else
-            {
-                auto thread = std::make_shared<MessageThread>(message);
-                tc->addReplyThread(thread);
-                return thread;
-            }
+
+            auto thread = std::make_shared<MessageThread>(message);
+            tc->addReplyThread(thread);
+            return thread;
         };
 
         if (auto tc =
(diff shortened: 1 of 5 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -160,9 +298,18 @@ ChannelView::ChannelView(BaseWidget *parent, Split *split, Context context,
...
+        {
+            if (it->second)
+            {
+                it = this->pauses_.erase(it);
+            }
+            else
+            {
+                it = ++it;
+            }
+        }
```
Nerixyz, line 312: Seems like `std::remove_if(..., [](const auto &p) { return p.has_value(); })` (or just `p`).

src/widgets/helper/ChannelView.cpp
```diff
@@ -1898,14 +2082,14 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
...
-                         MessageElementFlag::Username))
+            if (hoverLayoutElement->getFlags().has(
+                    MessageElementFlag::Username))
             {
                 openTwitchUsercard(this->channel_->getName(),
                                    hoverLayoutElement->getLink().value);
                 return;
             }
-            else if (hoverLayoutElement->getLink().isUrl() == false)
+            if (hoverLayoutElement->getLink().isUrl() == false)
```
Nerixyz, line 2092:
```suggestion
            if (!hoverLayoutElement->getLink().isUrl())
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -1368,13 +1530,15 @@ void ChannelView::drawMessages(QPainter &painter)
...
 {
-    if (!event->angleDelta().y())
+    if (event->angleDelta().y() == 0)
     {
+        // Ignore any scrolls where no vertical scrolling has taken place
         return;
     }
 
-    if (event->modifiers() & Qt::ControlModifier)
+    if ((event->modifiers() & Qt::ControlModifier) != 0)
```
Nerixyz, line 1539:
```suggestion
    if (event->modifiers().testFlag(Qt::ControlModifier))
```

src/widgets/helper/ChannelView.cpp
pajlada: I'm not sure I prefer this, I'd rather have a helper function here, or a local variable - the negation is too far away from the verb

Nerixyz: Didn't see clang-tidy.

src/widgets/helper/ChannelView.cpp
pajlada: Fixed in [`b9b7137` (#4926)](https://github.com/Chatterino/chatterino2/pull/4926/commits/b9b7137fa8971413f4237468546129455ede6d30)

Mm2PL (approved):
looks good to me

PR #4928, opened 2023-10-31
Fix log files unlocking when closing chat

kornes:
### Issue
Currently log files are never unlocked, to clear logs of closed chat user needs to close chatterino.

### Cause

`LogginChannel` [destructor](https://github.com/Chatterino/chatterino2/blob/master/src/singletons/helper/LoggingChannel.cpp#L49-L53) is never called, because channels are currently never removed from map in `Logging` singleton 

### Fix
- add `Logging::removeChannel` method and call it when `TwitchChannel` or `IrcChannel2` are destructed, which happens when last split of given channel is closed

### Other changes
- move channelPlatform QString creation away from `Channel::addMessage` (frequent code path) to constructor made `logFolderName_`
- add virtual `Channel::logFolderName()` and override it for IRC channels
- move map inserting logic from `Logging::addMessage` to its own `Logging::addChannel`, and call it once per `Channel` instance
- remove unused imports 
- change `loggingChannels_` from map to unordered_map

### DIFF
--- a/src/common/Channel.hpp
+++ b/src/common/Channel.hpp
@@ -104,6 +104,7 @@ public:
     virtual bool shouldIgnoreHighlights() const;
     virtual bool canReconnect() const;
     virtual void reconnect();
+    virtual QString &logFolderName();
 
     static std::shared_ptr<Channel> getEmpty();
 
@@ -118,6 +119,8 @@ private:
     const QString name_;
     LimitedQueue<MessagePtr> messages_;
     Type type_;
+    bool isLogInitialized_{false};
+    QString logFolderName_;
     QTimer clearCompletionModelTimer_;
 };
 
--- a/src/common/Channel.hpp
+++ b/src/common/Channel.hpp
@@ -104,7 +104,7 @@ public:
     virtual bool shouldIgnoreHighlights() const;
     virtual bool canReconnect() const;
     virtual void reconnect();
-    virtual QString &logFolderName();
+    virtual const QString &logFolderName();
 
     static std::shared_ptr<Channel> getEmpty();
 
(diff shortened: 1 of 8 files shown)


src/common/Channel.hpp
```diff
@@ -104,6 +104,7 @@ class Channel : public std::enable_shared_from_this<Channel>
     virtual bool shouldIgnoreHighlights() const;
     virtual bool canReconnect() const;
     virtual void reconnect();
+    virtual QString &logFolderName();
```
Nerixyz, line 107:
```suggestion
    virtual const QString &logFolderName();
```

Looks like it's only used as a getter.

src/singletons/Logging.cpp
```diff
@@ -49,27 +46,36 @@ void Logging::addMessage(const QString &channelName, MessagePtr message,
         }
     }
 
-    auto platIt = this->loggingChannels_.find(platformName);
-    if (platIt == this->loggingChannels_.end())
+    this->loggingChannels_.at(platformName)
+        .at(channelName)
+        ->addMessage(message);
```
Nerixyz, line 51: This feels dangerous. Although all channels initialize themselves, it would be nice if this didn't throw if the platform or the channel can't be found.

Nerixyz (commented):
Chatterino now creates log files for all open channels, even when logging is disabled.

src/singletons/Logging.cpp
kornes: throw should never happen, because addMessage is called only after loggingChannels_ was populated with same set of args, i would argue it failing instantly is better than hiding potential problems

kornes:
taken over by @Nerixyz

PR #4929, opened 2023-10-31
ci: Use clang-format 16

pajlada:
# Description

We now use https://github.com/DoozyX/clang-format-lint-action for running clang-format instead, since it allows us to change which clang-format version to use easier.


pajlada:
image

PR #4934, opened 2023-11-02
Add support for opening usercards by ID

Mm2PL:
# Description

Syntax is: `/usercard #117691339`.

Technically maybe enables `@#117691339` to be a link to my user if it ever matches.


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -792,6 +792,14 @@ void UserInfoPopup::updateUserData()
             return;
         }
 
+        // Correct for when being opened with "id:"
+        if (this->userName_ != user.login)
+        {
+            this->userName_ = user.login;
+
+            // Ensure recent messages are shown
+            this->updateLatestMessages();
```
kornes, line 801: `updateLatestMessages()` will run twice now, once with #id then with correct username from helix.
Would be best to avoid that, maybe prefix detection could be done sooner in `setData()` and just pass bool to `updateUserData()`

kornes (commented):
one issue, but works good 👍

src/widgets/dialogs/UserInfoPopup.cpp
Mm2PL: I considered two options:
 - rewrite the message filter to allow for checking for IDs
 - just call `updateLatestMessages()` when the information is available, which seemed fine until I just saw that it creates a new signal connection.

I'll make it skip the first call.

Mm2PL: Updated in [ba4151b](https://github.com/Chatterino/chatterino2/pull/4934/commits/ba4151b774a73424980ab4e918076fa2fb40b2e7)

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -701,7 +701,18 @@ void UserInfoPopup::setData(const QString &name,
                             const ChannelPtr &contextChannel,
                             const ChannelPtr &openingChannel)
 {
-    this->userName_ = name;
+    const QString idPrefix = "#";
```
Nerixyz, line 704: This (most likely) doesn't do what you want - it will always create a new `QString`. You can use `const QStringView ID_PREFIX = u"#"` (this is constexpr).

Nerixyz (approved):
When entering a nonexistent ID, the ID will display in the name with the hash. But this isn't any different from before, so not an issue. :+1:

src/widgets/dialogs/UserInfoPopup.cpp
Mm2PL:
```suggestion
    const QStringView idPrefix = u"#";
```

pajlada:
I would rather see this be a separate command, similar to `/banid`

pajlada:
Second attempt: I dislike the `#` prefix - it's already used for channels for IRC. Can we be explicit and use `id:`? (e.g. `/usercard id:11148817`)

Felanbird:
Yeaaaaaaaaaah, it's unfortunate that IRC uses # since it's seems obviously correct for #USERID

Mm2PL:
I could actually allow both prefixes and not complicate the code that much. As for it being a separate command, that could work although it's not something I'd like. As felanbird said, it shouldn't need to be a separate command.

pajlada:
https://github.com/Chatterino/chatterino2/pull/4945 implements this for `/ban` and `/timeout` using `id:` prefix, with unit tests so it can be re-used for further commands (like this one) in the future.
I don't think allowing both `id:` and `#` is a good idea, so just `id:` as done in my PR above is my preferred way forward

pajlada (approved):

PR #4939, opened 2023-11-04
chore: Add a 10m timeout to the clang-tidy CI job

pajlada:
# Description

Add a 10m timeout to the clang-tidy job


Nerixyz:
It would be nice if this PR could disable `misc-const-correctness.AnalyzeValues` as well (since it does _something_ with clang-tidy).

PR #4943, opened 2023-11-05
fix: prevent channelPointRewardAdded invocation while connecting

iProdigy:
Follow-up PR 1/2 for #4942 - implements the mutex outlined in https://github.com/Chatterino/chatterino2/issues/3942#issuecomment-1248397260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Bugfix: Fixed tooltips appearing too large and/or away from the cursor. (#4920)
 - Bugfix: Fixed a crash when clicking `More messages below` button in a usercard and closing it quickly. (#4933)
 - Bugfix: Fixed occasional crash for channel point redemptions with text input. (#4942)
+- Bugfix: Fixed first redemption of a channel points reward not appearing in rare cases. (#4943)
 - Dev: Change clang-format from v14 to v16. (#4929)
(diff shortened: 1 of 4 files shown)


Nerixyz (commented):
How does this "Fix first redemption of a channel points reward not appearing in rare cases"?

Both `TwitchChannel::addChannelPointReward` and `IrcMessageHandler::addMessage` [must] run on the GUI thread - they can't run concurrently. I can only see this introducing a deadlock condition where the `addMessage` call in the callback would result in another callback being added.

I feel like the whole reward handling should be refactored, and the signal business should be removed. Maybe with a `multimap`/`unordered_multimap` of reward-id → pending messages?

iProdigy:
> Both `TwitchChannel::addChannelPointReward` and `IrcMessageHandler::addMessage` [must] run on the GUI thread - they can't run concurrently.

Ah, I assumed the signals weren't guaranteed to run on the same thread but this is wrong, closing.


> I can only see this introducing a deadlock condition where the `addMessage` call in the callback would result in another callback being added.

That deadlock shouldn't happen since `channelPointRewards_` would contain the reward, so another callback would not be created

> I feel like the whole reward handling should be refactored, and the signal business should be removed. Maybe with a `multimap`/`unordered_multimap` of reward-id → pending messages?

I'm not opposed

PR #4957, opened 2023-11-10
fix: /banid and /ban id: performing duplicate bans

pajlada:
# Description

- fix: duplicate ban when banning by id
- update changelog entry


pajlada (commented):
image

Felanbird (approved):

PR #4958, opened 2023-11-11
fix: Split input sometimes not accepting focus

pajlada:
# Description

When clicking between the Split Input & the Split chat, there was a pixel or two where focus was not being granted to the split. This PR fixes that by adding a `giveFocus` call to the `mousePress` event.


pajlada:
image

PR #4959, opened 2023-11-11
refactor: Remove `Outcome` from network requests

### LINKED ISSUE #4909: Remove `Outcome` from network callbacks
It's never used.

Nerixyz:
# Description

As explained in the issue, the return value was never used.

Closes #4909.

### DIFF
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -2,6 +2,7 @@
 
 #include "common/NetworkRequest.hpp"
 #include "common/NetworkResult.hpp"
+#include "common/Outcome.hpp"
 #include "common/QLogging.hpp"
 #include "messages/Emote.hpp"
 #include "messages/Image.hpp"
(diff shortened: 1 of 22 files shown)


src/providers/bttv/BttvEmotes.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "common/NetworkRequest.hpp"
 #include "common/NetworkResult.hpp"
+#include "common/Outcome.hpp"
```
pajlada, line 5: remove include

pajlada (commented):
can also just remove the outcome.hpp file & its cmakelists entry
EDIT: nvm it's still used in places

fourtf:
RIP Outcome, it was my favorite

src/providers/bttv/BttvEmotes.cpp
pajlada: nvm, it's actually used here

pajlada (approved):

PR #4970, opened 2023-11-17
feat: Run tests on Windows & macOS in CI

### LINKED ISSUE #4780: Run tests in CI on Windows & macOS
Currently, tests are only run on Ubuntu 22.04.
It would be good to run tests on Windows & macOS too to ensure we don't break the test experience for our Windows & macOS developers.

The tests for Windows & macOS should be split up into separate files (i.e. `.github/workflows/test-windows.yml` and `.github/workflows/test-macos.yml`)

Relevant PR that unbroke Windows & macOS tests #4779

Wissididom:
# Description

- Windows: Docker can only run Windows Containers, so the PubSub Server and httpbin needed to be run manually in the background. First I tried using the repository that was used inside the docker image of the ubuntu test workflow, but that wasn't easy and pajlada suggested using httpbin-rs. That just didn't have the `$HOST/status/{code}` endpoints, so I searched for alternatives and found httpbox, which I then used.
- macOS: Using Colima to make Docker work. Without it, I get `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`.
- macOS: Use beta macos-13 to make the C++20 code work.
- Ubuntu: Add `--output-on-failure` to the workflow to see where the problem is if a test fails. Initially done to find a problem during playing around with it, but I think I can be helpful for later, so I kept it.

# Occasional Problem
- Sometimes the macOS Qt6 workflow fails with error `Subprocess aborted` at different cases but I don't know why that might be. For me, it was always fixed after re-running the workflow.

# Thought
To not have to use Docker and Colima on macOS, I could build the pubsub server and the httpbox tool on macOS in background too I think, but that probably won't make a difference I think.

Closes #4780


.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,189 @@
...
+        with:
+          repository: "kevinastone/httpbox"
+          path: "httpbox"
+
+      - name: Test
+        if: startsWith(matrix.os, 'windows')
+        timeout-minutes: 30
+        run: |
+          cd ..\httpbox
+          cargo build
```
Nerixyz, line 174: You should be able to use `cargo install --git https://github.com/kevinastone/httpbox` (pretty sure this could accept a commit/tag as well). This should be in its own step to make sure it doesn't contribute to the 30 min. Maybe we should fork the project and add releases (with `cargo-dist` for example). Then we could use [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall)

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,189 @@
...
+          cargo run -- --port 9051 &
+          cd ..\build-test
+          cd ..\pubsub-server-test
+          Invoke-WebRequest -Uri "https://github.com/Chatterino/twitch-pubsub-server-test/raw/master/cmd/server/server.crt" -outfile "server.crt"
+          Invoke-WebRequest -Uri "https://github.com/Chatterino/twitch-pubsub-server-test/raw/master/cmd/server/server.key" -outfile "server.key"
```
Nerixyz, line 179: This could use the same tag that _Download and extract Twitch PubSub Server Test_ uses, which could be defined at the top like `QT_QPA_PLATFORM`.

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,189 @@
...
+  cancel-in-progress: true
+
+jobs:
+  test-windows:
+    name: "Test ${{ matrix.os }}, Qt ${{ matrix.qt-version }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest]
+        qt-version: [5.15.2, 6.2.4]
```
Nerixyz, line 26: Shouldn't we test on 6.5 or 6.6?

.github/workflows/test-macos.yml
```diff
@@ -0,0 +1,111 @@
...
+  cancel-in-progress: true
+
+jobs:
+  test-macos:
+    name: "Test ${{ matrix.os }}, Qt ${{ matrix.qt-version }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-13]
+        qt-version: [5.15.2, 6.2.4]
```
Nerixyz, line 24: Shouldn't we test on 6.5 or 6.6?

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Setup sccache
+        # sccache v0.5.3
+        uses: nerixyz/ccache-action@9a7e8d00116ede600ee7717350c6594b8af6aaa5
+        if: startsWith(matrix.os, 'windows')
+        with:
+          variant: sccache
+          # only save on the default (master) branch
+          save: ${{ github.event_name == 'push' }}
+          key: sccache-test-${{ matrix.os }}-${{ matrix.qt-version }}-${{ matrix.skip-crashpad }}
```
Nerixyz, line 93: I'm not 100% sure: Will these use the same cache as the ones in the `build.yml`.

.github/workflows/test-windows.yml
Wissididom: Iirc I just took the version I found in the Ubuntu Test, where there was 6.2.4. Haven't really thought about what versions to use

.github/workflows/test-windows.yml
Wissididom: I don't know I more or less just copied it from the `build.yml`

.github/workflows/test-windows.yml
pajlada: Yeah, whichever version is used to build on windows/macos are the ones that should be used.
6.2.4 is used in Ubuntu 22.04 because that's what's available in the system repo

.github/workflows/test-windows.yml
Wissididom: Just tested the following command locally: `cargo install --git https://github.com/kevinastone/httpbox --rev 89b971f` and it seemed to have worked. If a tag existed, I could use `--tag` instead of `--rev` according to `https://doc.rust-lang.org/cargo/commands/cargo-install.html`.
Regarding the separate step, I just did it in the same step because the docker containers in the ubuntu test were started in the Test step too.

.github/workflows/test-windows.yml
Wissididom: Looking into the build.yml 6.5.0 is used for Windows and MacOS is not getting built in GitHub Actions for Qt6 (Just for Qt5), so would change it to 6.5.0 in `test-windows.yml` if I understand it correctly, but I wouldn't know if I should change it to something else on MacOS.

pajlada: The universal macOS build uses 6.5.0, so that would be the good version to use :+1:

.github/workflows/test-macos.yml
```diff
@@ -0,0 +1,111 @@
...
+
+jobs:
+  test-macos:
+    name: "Test ${{ matrix.os }}, Qt ${{ matrix.qt-version }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-13]
+        qt-version: [5.15.2, 6.5.0]
+        force-lto: [false]
```
pajlada, line 25: Remove everything related to LTO in this file

.github/workflows/test-macos.yml
```diff
@@ -0,0 +1,111 @@
...
+        plugins: [false]
+        clang-tidy-review: [false]
```
pajlada, line 27: Remove everything related to clang-tidy-review in this file

.github/workflows/test-macos.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v3.3.0
+        with:
+          cache: true
+          cache-key-prefix: ${{ runner.os }}-QtCache-${{ matrix.qt-version }}-v2
+          modules: ${{ env.QT_MODULES }}
+          version: ${{ matrix.qt-version }}
+
+      - name: Install dependencies
+        if: startsWith(matrix.os, 'macos')
```
pajlada, line 66: Remove all `if: startsWith(matrix.os, 'macos')` in this file

.github/workflows/test-macos.yml
```diff
@@ -0,0 +1,111 @@
...
+        timeout-minutes: 30
+        run: |
+          docker pull kennethreitz/httpbin
+          docker pull ${{ env.TWITCH_PUBSUB_SERVER_IMAGE }}
+          docker run --network=host --detach ${{ env.TWITCH_PUBSUB_SERVER_IMAGE }}
+          docker run -p 9051:80 --detach kennethreitz/httpbin
+          ctest --repeat until-pass:4 --output-on-failure --exclude-regex ClassicEmoteNameFiltering
+        working-directory: build-test
+        shell: bash
```
pajlada, line 104: Can remove all `shell: bash` in this file since it's the default in macOS

pajlada (changes requested):
Some nitpicks for the macOS file - overall this looks good :+1:

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,158 @@
...
+          # Enable PCH on Windows, later we can have it depend on the run/Qt Version like in the build workflow, I just want to make it work
+          C2_WINDOWS_USE_PCH: True
+        run: |
+          cmake `
+              -G"NMake Makefiles" `
+              -DCMAKE_BUILD_TYPE=RelWithDebInfo `
+              -DBUILD_TESTS=On `
+              -DBUILD_APP=OFF `
+              -DCMAKE_TOOLCHAIN_FILE="conan_toolchain.cmake" `
+              -DUSE_PRECOMPILED_HEADERS=${{ env.C2_WINDOWS_USE_PCH }} `
```
pajlada, line 123: You can just set this to `On` here - no need to use an environment variable

pajlada (approved):
This looks good from my perspective, I imagine the cache using the same as the build script should be ok - @Nerixyz is is double-checking before this is merged in

Nerixyz (approved):
LGTM 👍

PR #4972, opened 2023-11-19
Hide the Usercard button in the User Info Popup in special channels

### LINKED ISSUE #550: Channel-specific user info popup actions don't work in special channels

Mm2PL:
# Description

Closes #550.


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -739,6 +742,13 @@ void UserInfoPopup::setData(const QString &name,
         this->updateLatestMessages();
     }
     // If we're opening by ID, this will be called as soon as we get the information from twitch
+
+    if (auto *tc = dynamic_cast<TwitchChannel *>(this->channel_.get());
+        tc == nullptr)
+    {
```
kornes, line 748: this will also hide button from /mentions or search where it works as expected,
could be changed to check if channel type is `TwitchWhispers` or `TwitchLive`

src/widgets/dialogs/UserInfoPopup.hpp
```diff
@@ -80,6 +81,8 @@ class UserInfoPopup final : public DraggablePopup
 
         Label *noMessagesLabel = nullptr;
         ChannelView *latestMessages = nullptr;
+
+        EffectLabel2 *usercard = nullptr;
```
kornes, line 85:
```suggestion
        EffectLabel2 *usercardLabel = nullptr;
```

kornes (approved):

pajlada (approved):
I've tested this & made sure it opened as expected in:
 - Normal Twitch channels - it correctly shows the button
 - `/mentions` - it correctly shows the button & uses the channel the message was originally sent in
 - `/whispers` - it correctly hides the button, previously it would show the button & take me to `https://www.twitch.tv/popout//whispers/viewercard/testaccount_420`
 - `/automod` - it currectly shows the button & uses the channel the offending message was sent in
 - `/live` - it correctly hides the button, previously  it would show the button & take me to `https://www.twitch.tv/popout//live/viewercard/pajlada`

I have not tried `/watching` since I'm on Linux btw

PR #4977, opened 2023-11-25
fix: Only display spaces between words in reply context

Nerixyz:
# Description

There were some bugs regarding spaces and emojis in the reply context (a missing one before an emoji in `abc 😂 def` and an invalid one in `abc😂def`).

This PR fixes the bugs and adds tests for emoji parsing.

---

- Emoji parsing still uses a `boost::variant`, is that still necessary?
- The `EmojiMap emojis` in `Emojis` can be a `std::vector`, since the key is only used in tests and benchmarks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 - Bugfix: Fixed an issue where the setting `Only search for emote autocompletion at the start of emote names` wouldn't disable if it was enabled when the client started. (#4855)
 - Bugfix: Fixed empty page being added when showing out of bounds dialog. (#4849)
 - Bugfix: Fixed issue on Windows preventing the title bar from being dragged in the top left corner. (#4873)
-- Bugfix: Fixed an issue where reply context didn't render correctly if an emoji was touching text. (#4875)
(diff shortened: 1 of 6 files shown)


Nerixyz:
Not sure what the Windows 5.15 issue is (I suspect it's my [telegram's] qstringliteral).

Nerixyz:
> Not sure what the Windows 5.15 issue is (I suspect it's my [telegram's] qstringliteral).

I should've read the logs.

Nerixyz:
Seems like the highlight controller test is broken.

Wissididom:
> Seems like the highlight controller test is broken.

Yeah, we noticed it when creating the tests. With Qt 6.2.4 it seemed to have worked but with Qt 6.5.0 it times out. I played around with it here: https://github.com/Wissididom/chatterino2/pull/19

Nerixyz:
> Yeah, we noticed it when creating the tests. With Qt 6.2.4 it seemed to have worked but with Qt 6.5.0 it times out. I played around with it here: [Wissididom#19](https://github.com/Wissididom/chatterino2/pull/19)

Off-topic: We probably want to run tests in debug mode on macOS and Windows as well.

pajlada (approved):

PR #4978, opened 2023-11-26
Add support for sound backends & some miniaudio changes

pajlada:
# Description

Miniaudio now runs everything in a separate audio thread - this uses `boost::asio`'s `io_context`.
Our miniaudio implementation is now also much simplified - it does not use its own resource manager or device. This might end up being stupid if sounds don't work after changing output device or locking or w/e
I've made the sound controller into an interface, meaning we can support multiple sound backends in Chatterino. I've added a Null sound backend that disables all sound. A QMediaPlayer or QSoundEffect or Qt backend could be added.
Miniaudio might idle & disable the device now too, not sure

I've added some unrelated changes in the form of a new setting type, and a new setting page helper function for it, which will hopefully make adding new enum settings easier in the future.
This setting stores its value as a string instead of an int, and uses magic_enum to convert between that string value and its enum value.


src/controllers/sound/MiniaudioBackend.cpp
```diff
@@ -0,0 +1,300 @@
...
+    , workGuard(boost::asio::make_work_guard(this->ioContext))
+    , sleepTimer(this->ioContext)
+{
+    qCInfo(chatterinoSound) << "Initializing miniaudio sound backend";
+}
+
+MiniaudioBackend::~MiniaudioBackend()
+{
+    // NOTE: This destructor is never called because the `runGui` function calls _exit before that happens
+    // I have manually called the destructor prior to _exit being called to ensure this logic is sound
```
Wissididom, line 206: Is it intetional, that it ends in `sound` here?

pajlada: Yes :sunglasses:

PR #4981, opened 2023-11-26
fix: deadlock and use-after-free in tests

Nerixyz:
# Description

* Fixes use-after-free in Settings (when custom settings get destroyed, `getSettings()` would still point to the old instance).
* Fixes a deadlock (or similar?) as loading twitch accounts would init a request to 7TV.


pajlada (approved):
Good fixes, thank you! :+1: :rocket: :fire:

PR #4986, opened 2023-11-27
feat: add channel for messages caught by AutoMod

iProdigy:
Implements a split that shows automod caught messages across all channels the user moderates. Useful for folks that moderate multiple channels at once.


Nerixyz:
This feels like something in the direction of https://github.com/Chatterino/chatterino2/discussions/4931 - i.e. it could _in the future_ contain more types of messages, for example some follower, prediction, poll, or similar feed.

src/messages/MessageBuilder.cpp
```diff
@@ -193,6 +193,12 @@ std::pair<MessagePtr, MessagePtr> makeAutomodMessage(
 
     //
     // Builder for offender's message
+    builder2.message().channelName = channelName;
+    builder2
+        .emplace<TextElement>("#" + channelName,
+                              MessageElementFlag::ChannelName,
+                              MessageColor::System)
+        ->setLink({Link::JumpToChannel, channelName});
```
Nerixyz, line 201: Wouldn't this cause the channel name to be shown in the message added to the regular channel?

iProdigy: no

Nerixyz: yea nevermind

pajlada (commented):
Code overall looks good to me, two things
1) In the AutoMod logs, the message from automod doesn't have a channel attached to it
```
# Start logging at 2023-12-02 12:43:36 CET
# [12:43:36] automod: AutoMod: Held a message for reason: sexwords level 4. Allow will post it in chat. Allow Deny
#testaccount_420 [12:43:36] pajlada: pajlada: ...
```
2) Global search shows messages from the automod channel - as you already outlined in the description

iProdigy:
Fixed global search for automod, but couldn't figure out how messages from mentionsChannel don't appear in global search without such a conditional...

PR #4987, opened 2023-11-27
Make emote completion a lot smarter

### LINKED ISSUE #4944: Emote sorting, what I type, is not what I get
### Checklist

### Describe your issue

Emote sorting, do sort, but with all these emotes, they don't sort casesensitive

- I can't type the emote that I want, and get the emote that I want.
- _"That what I type, is not what I get"_ Sadge, definitely not a CS2 reference.

image

As example, when I type this. It suggests me ``rime``, while there is an emote named Rime
image

It's a bit of a question. But definitely the emote that you type and matches an emote **should atleast** display on the emote suggestion.

image

whether emotes should be sorted by casesensitivity, or display by casesensitivity, might be a thing of _preference_ and could be implemented but toggled in the settings

image

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.4.5 (commit 8fe3af35) built on 2023-09-16 with Qt 5.15.2, MSVC 193532217 Running on Windows 10 Version 2009, kernel: 10.0.19045

Mm2PL:
# Description

I've updated emote completion to be smarter.

Changes:
 - It now has vim-like smartcase. That is given `pajaW` and `PAJAW`, if you search for `PA` you'll only get `PAJAW`.
 - If smartcase search fails, the fallback is suggesting more uppercase emotes. That is if you search `Pa`, you'll get `PAJAW`, then `pajaW`.
 - If all lowercase: you will get case insensitive matching, preferring lowercase emotes
 - The emote sorting order is now well-defined: count missing characters then case difference then ASCII-betically.
 - Sorting with :-completion and tab-completion is now identical.

To test you will need to actually toggle it on.
Closes #4944.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Minor: Add menu actions to reply directly to a message or the original thread root. (#4923)
 - Minor: The `/reply` command now replies to the latest message of the user. (#4919)
 - Minor: All sound capabilities can now be disabled by setting your "Sound backend" setting to "Null" and restarting Chatterino. (#4978)
+- Minor: Add an option to use new experimental smarter emote completion. (#4987)
 - Bugfix: Fixed an issue where certain emojis did not send to Twitch chat correctly. (#4840)
 - Bugfix: Fixed capitalized channel names in log inclusion list not being logged. (#4848)
 - Bugfix: Trimmed custom streamlink paths on all platforms making sure you don't accidentally add spaces at the beginning or end of its path. (#4834)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -123,7 +123,7 @@ set(SOURCE_FILES
         controllers/completion/sources/UnifiedSource.hpp
         controllers/completion/sources/UserSource.cpp
         controllers/completion/sources/UserSource.hpp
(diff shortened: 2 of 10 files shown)


karl-police:
It works.

But there's something questionable that I don't know.


By default, the original version does this when starting the message with a ``:``

image


while the smarter completion one does this:
image


otherwise, the feature seems to work


image


happy smiley gets turned into the sad one Hmmmmmmmm

pajlada (approved):
LGTM :+1: 
Merging this in as-is without really testing its behaviour to allow more people to try it out

karl-police:
This is fixed now.

image

along with this
image

still wondering though how the original one prioritized the emotes when just typing ``:``, it feels like that it gave a more interesting selection of emotes

karl-police:
What's also a bit to experiment with, is

should it show Rime or rime

when one types ``rIme`` 🤷

karl-police:
Then how much should it even change?


Comparison:

Old one:
image





New one:
image

Mm2PL:
With my logic, as outlined in the description it should be, {"Rime", "rime"} when you search for "rIme" because:
 - it tries a case sensitive search (query has an uppercase letter)
 - no results
 - it tries case insensitive and since the query has an uppercase letter it prioritizes more uppercase emotes but returns all results

PR #4993, opened 2023-12-01
Cleanly exit Chatterino instead of force exiting

Mm2PL:
# Description

Joining the thread stops us from crashing.

Known issues relating to this:
 - if you start and close Chatterino after the recent-messages request was sent but before the response was received, Chatterino will crash. This didn't use to be the case because we'd just throw the process into `_exit`.


pajlada:
Needs some more testing, specifically on:
 - macOS
 - Windows
 - Windows with browser extension

Nerixyz:
This currently causes a call to std::terminate on Windows, the following fixed it:

```diff
diff --git a/src/providers/liveupdates/BasicPubSubManager.hpp b/src/providers/liveupdates/BasicPubSubManager.hpp
index f82f70363..3a61ee8fe 100644
--- a/src/providers/liveupdates/BasicPubSubManager.hpp
+++ b/src/providers/liveupdates/BasicPubSubManager.hpp
@@ -94,7 +94,10 @@ public:
                 .toStdString());
     }
 
-    virtual ~BasicPubSubManager() = default;
+    virtual ~BasicPubSubManager()
+    {
+        this->stop();
+    }
 
     BasicPubSubManager(const BasicPubSubManager &) = delete;
     BasicPubSubManager(const BasicPubSubManager &&) = delete;
```

We should make sure the app always eventually exits (not like in the tests some time ago, where the Qt event-loop wouldn't spin).

src/RunGui.cpp
```diff
@@ -311,6 +311,7 @@ void runGui(QApplication &a, Paths &paths, Settings &settings)
         pajlada::Settings::SettingManager::gSave();
     }
 
+    thread.join();
```
pajlada, line 314:
```suggestion
    if (thread.joinable()) {
        thread.join();
    }
```

src/RunGui.cpp
```diff
@@ -308,13 +311,14 @@ void runGui(QApplication &a, Paths &paths, Settings &settings)
         pajlada::Settings::SettingManager::gSave();
     }
 
+    if (thread.joinable()) {
+        thread.join();
+    }
```
Mm2PL, line 316:
```suggestion
    if (thread.joinable())
    {
        thread.join();
    }
```

brian6932:
Tested closing on Windows with the browser extension attached, and had no issues

pajlada:
Latest commit tested on i3 & macOS - works as expected & shuts down pretty fast

Nerixyz:
This was broken by #5058 and #5059 will break this on Windows.

pajlada:
I think I would prefer to do smaller steps towards removing `_exit`, essentially lean into the `fakeDtor` we have now and start cleaning up one part of Chatterino at a time, until fakeDtor essentially runs the full destructor

pajlada (approved):
image

Works as expected on Linux & macOS - @nerixyz and @brian6932 are testing on Windows

pajlada:
Closing in favour of https://github.com/Chatterino/chatterino2/pull/5537

PR #4868, opened 2023-10-04
Use Qt's High-DPI scaling on Windows

Nerixyz:
# Description

This switches from manual DPI handling to Qt's DPI handling, as Qt 6 disabled the option to opt-out of high DPI scaling through application attributes. Although this PR mainly fixes Windows bugs, it should be tested on all platforms, as the changes will affect everyone.

Windows now mostly uses the display scaling other platforms used.

You can test the scaling for yourself by setting `QT_SCALE_FACTOR` to something other than 1.

~~I'm waiting for https://github.com/Chatterino/chatterino2/pull/3690 to be merged, as that will fix some font-bugs (although I didn't see any on this branch).~~ This PR should be merged first.

- Fixes #4552.
- Fixes #4862.
- Fixes #5048. 
- I'm pretty sure #2348 will be fixed as well.
- #3754 on Windows
- #754 on Windows
- #752 on Windows
- #750 on Windows
- #749 on Windows
- #5072 should be fixed (?)

I left out "fixes" for some issues, as they might persist on other platforms.

**Stuff to test**

One major thing to look out for is the size of the window reported to the OS. You can use something like [ShareX](https://getsharex.com/) to measure the size. Make sure the size is _exact_ (not too large, not too small). Previously, the window was 1px too tall. Some of the testing might need to be done on Qt 5.15 or 6.7 (or later).

- [x] Check borders on Windows 11 (are they needed?, is the window cut off?)
  - There seems to be a 1-2px border, however, due to [QTBUG-124250](https://bugreports.qt.io/browse/QTBUG-124250) we can't adjust the border (yet... 🙏)
- [x] Check resizing on Windows 11 (is it janky?)
- [ ] Dragging a window across screens
- [ ] Saving the position on high-dpi screens (#2348)

Issue #1967, opened 2020-09-14
Part of the title bar can't be dragged


hesoyam5:
**Describe the bug**
Part of the title bar can't be dragged, by title bar i mean the top bar that you hold-click and drag around the screen, the one with minimize, maximize and close buttons.
**To reproduce**
I click-hold to the left of 'tt' in Chatterino and i can't drag the chatterino window around, i click to the right of 'tt' in Chatterino and i can drag it around just fine.

**Chatterino version**
Chatterino 2.2.2 (commit 07fc0c3ad), 14.09.2020

**Operating system**
Windows 10 Pro, Version 2004, OS Build 19041.508

**Additional information**
Also, can you add the ability to hide the account manager from the title bar? or at least make it possible to drag the window by hold-clicking on the account manager

Felanbird:
Can replicate

Windows 10 | 214afc8fc

TranRed:
Can reproduce with Windows 10 | 7eabba959

Could be default behaviour, because it seems to be roughly the place where the context menu normally is.
Example settings pop-up:
![grafik](https://user-images.githubusercontent.com/16636423/93250839-ac3aa880-f793-11ea-9b4c-e95a15c51e54.png)

alazymeme:
Can reproduce. I agree that it may be the above, but other applications handle this just fine (eg discord, firefox, etc)

jose-seabra:
This is a bigger issue if you have a long username. You have to resize the window to be able to drag it

![chatterinoi](https://user-images.githubusercontent.com/55154249/115960023-0f00d900-a507-11eb-9560-e2b5b52082ed.png)

PR #4994, opened 2023-12-01
Return correct hit-test values for title bar buttons on Windows

### LINKED ISSUE #4991: Return correct hit-test values for title bar buttons on Windows
### Checklist

### Describe your issue

From https://github.com/Chatterino/chatterino2/pull/4914#issuecomment-1777449080, we should return `HTMINBUTTON`, `HTMAXBUTTON` and `HTCLOSE` for the title bar buttons on Windows when handling [`WM_NCHITTEST`](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-nchittest) to allow the OS to display e.g. snap layouts.

This isn't as easy as _just_ returning the correct value, but we will need to at lease handle [`WM_NCMOUSEMOVE`](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-ncmousemove). Possibly, we need to handle the L/R clicks as well. Most of these should synthesize events to Qt (or native events that Qt will handle afterward) as these events will cause the "regular" events (like `WM_MOUSEMOVE`) to stop regardless of the return value in the event handler.

### Screenshots

_No response_

### OS and Chatterino Version

-

Nerixyz:
# Description

This Pr returns the correct hit-test values for the title bar buttons (minimize, maximize/restore, and close). This will allow Windows 11 snap layouts to work with Chatterino (and possibly other OS features).

I adapted the implementation from https://github.com/desktop-app/lib_ui/blob/44161f183c255dc0dac7ebe9558a1ca48f5e5258/ui/platform/win/ui_window_win.cpp#L755-L810.

* In `WM_NCMOUSE{MOVE, HOVER}`, we simulate move events for the title bar buttons, as Qt doesn't generate those.
* In `WM_NCMOUSELEAVE` (and the first `WM_MOUSEMOVE` after a move in the non-client area), we clear/leave all hovered buttons.
* In `WM_NCLBUTTON{UP, DOWN}`, we simulate a click on a button.

From my testing, it seems that we don't need to handle `WM_NC{L, M, R, X}BUTTONDBLCLK` and `WM_NC{M, R, X}BUTTON{DOWN, UP}`.
_For some future PR (maybe):_ `QWidget::nativeEvent` always returns `false` and does nothing else - we don't need to call it.

When testing, please test as many native gestures/tricks/…, like double-clicking the title bar, as you can.

Closes #4991.
Addresses #1967 (as the window can be dragged in the top left corner now - not 100% sure why).

### TODO

- [x] Explain this in more detail
- [x] Add comments for the messages
- [x] Test on multiple monitors
- [x] Test on high-dpi
- [x] Handle default window frames
-  ~~Check what the tooltips are about~~ Seems like they're from the OS
- [ ] Test on Windows 11 (I can't do that 😭)

<sub>As for high-dpi, this needs to be changed if #4868 ever lands.</sub>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,6 @@ jobs:
             plugins: false
             skip-artifact: false
             skip-crashpad: false
-            clang-tidy-review: false
           # Ubuntu 22.04, Qt 5.15
           - os: ubuntu-22.04
             qt-version: 5.15.2
@@ -45,7 +44,6 @@ jobs:
             plugins: false
             skip-artifact: false
             skip-crashpad: false
-            clang-tidy-review: true
           # Ubuntu 22.04, Qt 6.2.4 - tests LTO & plugins
           - os: ubuntu-22.04
             qt-version: 6.2.4
@@ -53,7 +51,6 @@ jobs:
             plugins: true
             skip-artifact: false
             skip-crashpad: false
-            clang-tidy-review: false
           # macOS
           - os: macos-latest
             qt-version: 5.15.2
@@ -61,7 +58,6 @@ jobs:
             plugins: false
             skip-artifact: false
             skip-crashpad: false
-            clang-tidy-review: false
           # Windows
           - os: windows-latest
             qt-version: 6.5.0
@@ -69,7 +65,6 @@ jobs:
             plugins: false
             skip-artifact: false
             skip-crashpad: false
-            clang-tidy-review: false
           # Windows 7/8
           - os: windows-latest
             qt-version: 5.15.2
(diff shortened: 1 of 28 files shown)


pajlada (commented):
I'm not a huge fan of Q_ASSERT*, can you confirm Q_ASSERT_X doesn't fire in our CI-built windows artefacts?
I've confirmed Q_ASSERT* works as expected on my machine on Linux by just adding a Q_ASSERT at the top of main.cpp

Once you've confirmed it works, I'm fine to leave it here and would be more comfortable using it for future GUI code

I will test this on Windows11 now

PR #4995, opened 2023-12-02
Add a UI for uploaded images

Mm2PL:
# Description

### This includes migrations which shouldn't be but might turn out being buggy, use at your own risk.
Ensure that you won't lose anything that would be in your `Logs/ImageUploader.json` file.

This PR adds a UI that lists all logged images. It also adds a button that forgets an image from the log.


Infinitay:
**Personally**, I'm not a fan of the image preview given size constraints, so I have a few suggestions instead if you're up for them:

1. The most obvious, remove the preview
2. Instead, add the ability to bring up a on-hover pop-up when hovering over the image path. Similar to when hovering over emotes and links in chat.
    - Or replace the image preview with the text "Hover to Preview" and add it there instead 
3. Increase the height of all rows to better view a preview and/or give the ability for users to dynamically change the cell size to their own liking.

Mm2PL:
> Would it be possible to add the ability to bring up a on-hover pop-up when hovering over the image preview for a better view? Similar to when hovering over emotes and links in chat.

I'm not sure I like this solution. It requires interaction to just look at an image, where there is no other good 'index' to go by.

Mm2PL:
> Found 2 issues while testing on [097eb7b](https://github.com/Chatterino/chatterino2/pull/4995/commits/097eb7ba79295ca60264b8ea52e0a6d1896ccd49):

@jupjohn I fixed lazy loading causing issues and migrations.

Mm2PL:
Note to self: add channel which the image was intended to be sent to

Mm2PL:
There were complexity complaints related to this PR. They mostly included removing the image preview. However UI-wise an implementation that doesn't load previews might as well just be your text editor. Searching for any image in upload history (for example to repost or delete) is already tedious enough, especially if it's older than a couple of days. IMO this idea is not worth doing without the preview. I'd like to hear feedback on continuing development of this.
@pajlada

pajlada:
To necrobump this since I didn't response for more than a year:
I don't think I want to support much more advanced use cases for the image uploader, so I'd prefer if we didn't add this.

From my standpoint, Chatterino should excel at uploading images (and maybe gifs or mp4s) and posting them to chat. Anything else is extra that I'd need a very good reason to want to support. Keeping a log of uploaded images is already extra, and would allow something like this to be built as a separate application or website.

Issue #3501, opened 2022-01-13
Tooltips spawn as separate windows rather than popups on Wayland when the Chatterino window isn't focused


ilya-zlobintsev:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

While this is somewhat a duplicate of #1647 , that issue has been piled up with other wayland-related problems so I feel like it would be appropriate to open an issue for each separate problem.

When running Chatterino on a wayland environment and using the wayland-native QT backend (this is not always the default depending on the distro), hovering over anything that shows a tooltip while the Chatterino window isn't focused spawns the tooltip in a separate window, rather than showing a tooltip. It works properly when the chatterino window is focused.

This seems to be the behaviour on all wayland compositors (Gnome/KDE/Sway), but it's most noticeable on tiling WMs such as sway since the newly spawned "tooltip window" gets tiled, leading to pretty much an unusable experience (look at the recording for an example).

Steps to reproduce:
- Run a Wayland desktop environment
- Start Chatterino using the wayland-native QT backend: `QT_QPA_PLATFORM=wayland chatterino` (requires installing `qt5-wayland` on Arch Linux. On some distros QT defaults to Wayland.) You can check if an application is running on Wayland or XWayland(Xorg compatibility mode) by trying to use `xprop`: if you can click on the window and it shows its properties, the application is using XWayland, otherwise xprop doesn't work on it.
- Focus a window other than Chatterino
- Hover over anything that shows a tooltip in Chatterino, such as an emote.

Potential causes and fixes:
This could be caused by the tooltip not having a proper parent set, as I believe wayland enforces this.
This seems to be a useful guide on fixing some wayland issues in QT applications: https://community.kde.org/Guidelines_and_HOWTOs/Wayland_Porting_Notes

### Screenshots

https://streamable.com/yuem12 (you can see which window is currently focused by the blue border or the text at the top)

### OS and Chatterino Version

Linux, any Wayland compositor, any Chatterino version running natively on Wayland

ilya-zlobintsev:
Everything works properly after #4998

PR #4998, opened 2023-12-03
Fix missing parent of tooltips for Wayland

crazysmc:
In response to issue #3501

Each widget that uses TooltipWidget gets its own instance of it with itself as a parent set.

That fixes the issue for me on Wayland. Although I'm not using a tiling WM, the parent-less tooltips appeared in the wrong place before.


ilya-zlobintsev:
This seems to solve the issue for me on Arch Linux using KDE 5.27. Tested on both QT5 and QT6.

ilya-zlobintsev:
I've also tested sway - the behaviour is slightly different, the previews don't show up at all when hovering over something when the window is unfocused, but IMO that is fine as well

Nerixyz:
Seems to work on Windows 10 with Qt 6. Works when the window is unfocused. Only thing I noticed is that tooltips will stay when the window is resized, but shouldn't be an issue.

pajlada (approved):
Thank you! Works as expected on Arch Linux w/ i3

PR #5000, opened 2023-12-03
Add a new completion API for experimental plugins feature.

Mm2PL:
# Description

I remade it for the new completion system. Currently it only allows tab completion.
Example plugin using this api: https://github.com/Mm2PL/supibot-completion-plugin
I have updated https://github.com/Mm2PL/chatterino-test-plugin/ to also include a completion example

I might refactor out [Unrelated: Add error tracking on load](https://github.com/Chatterino/chatterino2/pull/5000/commits/1a188f73082aacae477924fd924f589c89ade3bd) to be a separate PR.

Happy 5k
FeelsBirthdayMan


src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -54,18 +65,111 @@ StackIdx push(lua_State *L, const std::string &str);
...
 bool peek(lua_State *L, std::string *out, StackIdx idx = -1);
+bool peek(lua_State *L, api::CompletionList *out, StackIdx idx = -1);
 
 /**
  * @brief Converts Lua object at stack index idx to a string.
  */
 QString toString(lua_State *L, StackIdx idx = -1);
 
+// Special
+class BalanceKepper
```
pajlada, line 81: Add a comment to this, and probably rename to `BalanceKeeper`, maybe something including `Stack`?

src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -54,18 +65,111 @@ StackIdx push(lua_State *L, const std::string &str);
...
+{
+    int expected;
+    lua_State *L;
+
+public:
+    BalanceKepper(lua_State *L)
```
pajlada, line 87: Comment the two constructors

src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -54,18 +65,111 @@ StackIdx push(lua_State *L, const std::string &str);
...
+        {
+            stackDump(this->L, "BalanceKepper check tripped");
+            // clang-format off
+            // clang format likes to insert a new line which means that some builds won't show this message fully
+            assert(false && "internal error: lua stack was not in an expected state");
+            // clang-format on
+        }
+    }
+    BalanceKepper operator=(BalanceKepper &) = delete;
+    BalanceKepper &operator=(BalanceKepper &&) = delete;
```
pajlada, line 112: Delete the copy/move ctors too

src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -150,6 +254,7 @@ StackIdx push(lua_State *L, T inp)
 template <typename T>
 bool pop(lua_State *L, T *out, StackIdx idx = -1)
 {
+    BalanceKepper _(L, -1);
```
pajlada, line 257: Let's give this & uses like it names in the future - for mutex guards.
For mutex guards, we'd still do something like
```c++
std::unique_lock lock(this->mutex);
```

src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -186,6 +291,55 @@ StackIdx pushEnumTable(lua_State *L)
...
+    CallbackFunction(lua_State *L, StackIdx stackid)
+        : stackidx(stackid)
+        , L(L)
+    {
+    }
+
+    // this type owns the stackidx, it must not be trivially copiable
+    CallbackFunction operator=(CallbackFunction &) = delete;
+
+    CallbackFunction &operator=(CallbackFunction &&) = default;
```
pajlada, line 311: Delete the copy & move ctors too

src/controllers/plugins/LuaUtilities.hpp
```diff
@@ -186,6 +291,55 @@ StackIdx pushEnumTable(lua_State *L)
...
 
+// Represents a Lua function on the stack
+template <typename ReturnType, typename... Args>
+class CallbackFunction
+{
+    StackIdx stackidx;
+    lua_State *L;
+
+public:
+    CallbackFunction(lua_State *L, StackIdx stackid)
```
pajlada, line 302: naming/casing of member `stackidx` & paramter `stackid` should be consistent, probably `stackIdx` or `stackIndex`

src/controllers/plugins/PluginController.cpp
```diff
@@ -306,6 +313,46 @@ const std::map<QString, std::unique_ptr<Plugin>> &PluginController::plugins()
...
+    const QString &query, const QString &fullTextContent, int cursorPosition,
+    bool isFirstWord)
+{
+    QStringList results;
+    for (const auto &[name, pl] : getApp()->plugins->plugins())
+    {
+        if (!pl->error().isNull())
+        {
+            continue;
+        }
```
pajlada, line 326: This feels like an awkward API - having to check if the plugin has an error in it. Should maybe be handled by the plugin controller instead.
Not something that needs to be fixed in this PR, but maybe some comment should be created for it at least.

src/controllers/plugins/PluginController.cpp
```diff
@@ -306,6 +313,46 @@ const std::map<QString, std::unique_ptr<Plugin>> &PluginController::plugins()
...
+        lua::BalanceKepper _(pl->state_);
+        qCDebug(chatterinoLua)
+            << "Processing custom completions from plugin" << name;
```
pajlada, line 329: This log entry should be moved into the `if (opt)` part, or it would print for every plugin instead of every plugin that has a completion callback registered

src/controllers/plugins/Plugin.hpp
```diff
@@ -85,10 +87,50 @@ class Plugin
         return this->loadDirectory_;
     }
 
+    // Note: The CallbackFunction object's destructor will remove the function from the lua stack
+    std::optional<lua::CallbackFunction<lua::api::CompletionList, QString,
+                                        QString, int, bool>>
+        getCompletionCallback()
+    {
```
pajlada, line 94: Move this function to the source file, and let's maybe add a `using LuaCompletionCallback = ...`; to make this look a bit more sane

pajlada (commented):
Some nitpicks, would like to see an example plugin written in lua that utilizes the API

pajlada:
The current completion example doesn't work as expected i think - is this an issue in the test plugin or in the c++ code?

e.g. when I type `/foo` and press <TAB> it'll remove the `/foo` and start autocomplete `barbazA` and `barbazB`
Same when I type `/foo ` with a space - it'll remove the `/foo`

pajlada (approved):

Issue #5004, opened 2023-12-06
Ctrl+Backspace not working properly in Emote Search and Settings after Select All


21mtd:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

This problem happened and fixed in the past (#4461) but still occurred in Emote Menu and Settings. The last keyword remain undeleted after Select All then Ctrl + Backspace.

***Reproduction steps:***

- Open a Emote Menu or Settings
- Type random letters in input box
- Ctrl+A to select all then Ctrl+Backspace

### Screenshots

https://github.com/Chatterino/chatterino2/assets/75530836/1e2c1f50-2073-4f18-b98c-35ffb12b351c

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit c3d3903b) built on 2023-12-05 with Qt 6.5.0, MSVC 193732826 Running on Windows 10 Version 22H2, kernel: 10.0.19045

Mm2PL:
Can confirm. Similar to #4461. Probably will need a patch similar to #4536.

Mm2PL:
This was fixed in #5013.

PR #5013, opened 2023-12-10
Fixes to ctrl+backspace behavior

gremble0:
# Description
First contribution to chatterino, so hello :)

This PR addresses issue #5004 + fixes some issues related to that issue

## Changes
Added ctrl+backspace functionality while there is an active selection to `EmotePopup`, `SettingsDialog` and `SelectChannelDialog` classes, as well as a small change (fix) to ctrl+backspace behavior in and `SearchPopup`, which is also the behavior implemented in the other classes. 

Previously the behaviour for ctrl+backspace with an active selection was to clear all characters from the end of the selection to start of line (or the behavior described in the issue for the classes with it not implemented). This has been changed to ctrl+backspace deleting only the selected text, which is the behavior present in all other applications i tested it in.

Old behavior:
https://github.com/Chatterino/chatterino2/assets/45577341/160dd1e4-f46e-43aa-826e-a827dc264908

New behavior:
https://github.com/Chatterino/chatterino2/assets/45577341/8e2a4e4f-92e1-4f67-a54d-3d0544e1c7d9

It may be possible make a more generic solution to this issue than to add these `eventFilter()` methods wherever there is a input field, but as of yet i'm not familiar enough with QT and the chatterino codebase to do that.


src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -448,6 +449,22 @@ void EmotePopup::loadChannel(ChannelPtr channel)
...
 }
 
+bool EmotePopup::eventFilter(QObject *object, QEvent *event)
+{
+    if (object == this->search_ && event->type() == QEvent::KeyPress)
+    {
+        QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
+        if (keyEvent->key() == Qt::Key_Backspace &&
+            keyEvent->modifiers() == Qt::ControlModifier &&
+            this->search_->selectionLength() > 0)
```
Nerixyz, line 459:
```suggestion
        if (keyEvent == QKeySequence::DeleteStartOfWord &&
            this->search_->selectionLength() > 0)
```

We should be platform-agnostic.

src/widgets/dialogs/SelectChannelDialog.cpp
```diff
@@ -497,6 +497,13 @@ bool SelectChannelDialog::EventFilter::eventFilter(QObject *watched,
             widget->previousInFocusChain()->setFocus();
             return true;
         }
+        else if (event_key->key() == Qt::Key_Backspace &&
+                 event_key->modifiers() == Qt::ControlModifier &&
+                 this->dialog->ui_.twitch.channelName->selectionLength() > 0)
```
Nerixyz, line 502:
```suggestion
        else if (event_key == QKeySequence::DeleteStartOfWord &&
                 this->dialog->ui_.twitch.channelName->selectionLength() > 0)
```

Same here.

src/widgets/dialogs/SettingsDialog.cpp
```diff
@@ -206,6 +207,22 @@ void SettingsDialog::setElementFilter(const QString &query)
...
 }
 
+bool SettingsDialog::eventFilter(QObject *object, QEvent *event)
+{
+    if (object == this->ui_.search && event->type() == QEvent::KeyPress)
+    {
+        QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
+        if (keyEvent->key() == Qt::Key_Backspace &&
+            keyEvent->modifiers() == Qt::ControlModifier &&
+            this->ui_.search->selectionLength() > 0)
```
Nerixyz, line 217:
```suggestion
        if (keyEvent == QKeySequence::DeleteStartOfWord &&
            this->ui_.search->selectionLength() > 0)
```

And here.

src/widgets/helper/SearchPopup.cpp
```diff
@@ -204,9 +204,9 @@ bool SearchPopup::eventFilter(QObject *object, QEvent *event)
         QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
         if (keyEvent->key() == Qt::Key_Backspace &&
             keyEvent->modifiers() == Qt::ControlModifier &&
```
Nerixyz, line 206:
```suggestion
        if (keyEvent == QKeySequence::DeleteStartOfWord &&
```

And here.

Nerixyz (commented):
This is probably a bug in Qt. I couldn't find a bugreport after a quick search.

The bug is in [`QWidgetLineControl::processKeyEvent`](https://github.com/qt/qtbase/blob/dev/src/widgets/widgets/qwidgetlinecontrol.cpp#L1816-L1821). This calls `cursorWordBackward`, which removes the selected word from the selection. Calling `del` will only delete the remaining selection.

gremble0:
Yeah, i supposed this sounds like something chatterino probably shouldn't have to deal with. Anyways I've added your suggestions which will fix the behavior until it potentially gets fixed upstream. Until then i suppose it's up to you what you want to do in chatterino.

pajlada:
@Nerixyz Does it make sense to merge this in the interim, while waiting for your fix to Qt to be released?
I'm guessing yes since this fix is small enough that it would be easy to revert once we're on a fixed Qt release

Nerixyz:
> @Nerixyz Does it make sense to merge this in the interim, while waiting for your fix to Qt to be released? I'm guessing yes since this fix is small enough that it would be easy to revert once we're on a fixed Qt release

Yes, we should merge this. The fix should be in 6.6.2 (to be released on 17.01.2024). But, at least on Windows, we're still waiting on [509236](https://codereview.qt-project.org/c/qt/qtbase/+/509236).

pajlada (approved):

pajlada:

PR #5026, opened 2023-12-15
feat: add sound and flash alert for automod caught messages

iProdigy:
When automod split isn't present in the current tab (or chatterino is minimized), moderators may not realize an automod caught message has been received. This PR allows a sound notification and/or taskbar flash upon receiving a caught message


pajlada (commented):
I think I would rather see this be a new "default" bar in the Highlights settings section, and it functioning more closely to a highlight.

Let me know if I'm missing some context that makes that stupid

iProdigy:
While the new default highlight row has been added, this PR is still a draft since automod messages don't work through SharedMessageBuilder, so `HighlightController::check` is never called - will follow-up with a refactor that addresses this

iProdigy:
Yeah I'm not tied to having any scrollbar color (old caught messages are likely to be unactionable anyways), so we can remove it for now (and changing the message background could look strange)

pajlada (approved):

PR #5033, opened 2023-12-16
perf: Only update ChannelView for GIFs if visible

dnsge:
## Description

On every `gifRepaintRequested` invocation (e.g. every `GIFTimer` tick), we currently update *every* ChannelView, even those in tabs that aren't active. The number of repaints every 20ms scaled with the number of open tabs. 

This PR only calls `update` on a ChannelView if the ChannelView is visible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@
 - Dev: Load less message history upon reconnects. (#5001)
 - Dev: BREAKING: Replace custom `import()` with normal Lua `require()`. (#5014)
 - Dev: Fixed most compiler warnings. (#5028)
+- Dev: Only update visible ChannelViews for GIF repaints. (#5033)
 
 ## 2.4.6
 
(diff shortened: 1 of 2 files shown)


pajlada:
An update does not necessarily mean it gets repainted - I suspect Qt is smart enough to not repaint if a widget is not visible
The CPU usage on my system looks the same with this PR, so I'd like to test this a bit more before approving it

dnsge:
> An update does not necessarily mean it gets repainted - I suspect Qt is smart enough to not repaint if a widget is not visible The CPU usage on my system looks the same with this PR, so I'd like to test this a bit more before approving it

I will say that I added print debug statements to `queueUpdate` and `paint` and saw that both log statements were triggered every time for every open channel. But definitely look more into this because as you say Qt should be smart.

Nerixyz:
> I suspect Qt is smart enough to not repaint if a widget is not visible

Yes, it is: https://github.com/qt/qtbase/blob/64a283a98894705fe51c9423812bbeb634138850/src/widgets/kernel/qwidget.cpp#L11205-L11206.

dnsge:
Looks like you're right, I must have been logging in the wrong place.

PR #5036, opened 2023-12-17
fix: ReplyThreadPopup now requires a split as its parent

pajlada:
# Description

Fixes the problem described in #4747 in a different way


pajlada:
image

PR #5037, opened 2023-12-17
chore: Unify `.clang-format` file

pajlada:
# Description

- Move all `.clang-format` files to the root directory
  Since they're all the same, there's no need to copy paste the same file multiple times.
  The only downside this has it that when opening a project under `lib` that does not have its own `.clang-format` file, your editor might try to format it based on our config.
- Update `scripts/check-format.sh` to include all source dirs


pajlada:
image

PR #5040, opened 2023-12-25
fix: Tooltip parenting on Windows

### LINKED ISSUE #5019: Moving over different windows' popups automatically switches focus
### Checklist

### Describe your issue

**Describe the bug**
When multiple Chatterino windows are opened and overlaid on top of each other, moving the cursor over another one's popups (hovers) will automatically move the focus over to the other window.

**To reproduce**
- Open two windows of Chatterino 
- Overlay them on top of each other partially
- Mouse over different popups (e.g. stream title, link preview, emote detail, badge detail, etc.)
- Observe windows switching focus
- Likely only observable on Windows

### Screenshots

**Video preview**
- In this video preview, I do not press any keys or any mouse buttons

https://github.com/Chatterino/chatterino2/assets/7684273/c7934cb3-607b-4e17-88ef-25888b31e47e

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit fd4cac2c) built on 2023-12-10 with Qt 6.5.0, MSVC 193732826 Running on Windows 10 Version 22H2, kernel: 10.0.19045

Nerixyz:
This PR sets the parent on tooltips to be `nullptr` on Windows and deletes the tooltip when appropriate.

Fixes #5019.

### DIFF
--- a/src/widgets/TooltipWidget.hpp
+++ b/src/widgets/TooltipWidget.hpp
@@ -29,11 +29,6 @@ class TooltipWidget : public BaseWindow
     Q_OBJECT
 
 public:
-    /// @param parent The parent of this widget.
-    ///               If this widget is a window, use tooltipParentFor().
-    ///               Remember to delete the widget!
-    ///
-    /// @see tooltipParentFor()
     TooltipWidget(BaseWidget *parent);
     ~TooltipWidget() override = default;
 
@@ -77,19 +72,5 @@ private:
 
     pajlada::Signals::SignalHolder connections_;
 };
-
-#ifdef Q_OS_WIN
-template <typename T>
-inline constexpr T *tooltipParentFor(T * /*desiredParent*/)
-{
-    return nullptr;
-}
-#else
-template <typename T>
-inline constexpr T *tooltipParentFor(T *desiredParent)
-{
-    return desiredParent;
-}
-#endif
 
 }  // namespace chatterino
--- a/src/widgets/TooltipWidget.hpp
+++ b/src/widgets/TooltipWidget.hpp
@@ -29,6 +29,11 @@ class TooltipWidget : public BaseWindow
     Q_OBJECT
 
 public:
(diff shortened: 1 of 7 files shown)


src/widgets/TooltipWidget.hpp
```diff
@@ -29,6 +29,11 @@ class TooltipWidget : public BaseWindow
     Q_OBJECT
 
 public:
+    /// @param parent The parent of this widget.
+    ///               If this widget is a window, use tooltipParentFor().
```
pajlada, line 33: Is this more accurate?

```suggestion
    ///               If the parent is a window, use tooltipParentFor().
```

pajlada (approved):
Looks good to me, just a question about the documentation before I'll merge this in

src/widgets/TooltipWidget.hpp
Nerixyz: No. If we take `ChannelView` for example: `ChannelView` is not a window, but `Tooltip` is used as a window there. See [`QWidget::isWindow`](https://doc.qt.io/qt-6/qwidget.html#isWindow).

pajlada: In what scenario would we not want to use `tooltipParentFor` then?
Right now we use it both times the tooltip widget is created (`ChannelView` & `SplitHeader`)

Nerixyz: If the tooltip widget is embedded into another widget (or through multiple layers) that is then shown as a window.

pajlada: Since that's not a scenario that we run into now, let's just simplify this & expand to support those use cases as they come up.
Either just `#ifdef` in the constructor itself, or make `tooltipParentFor` a static function in the source file & use it in the constructor

pajlada (changes requested):
Pending small change from discussion above

pajlada (approved):

PR #5041, opened 2023-12-25
refactor: Make `Args` less of a singleton

pajlada:
This means it's no longer a singleton, and its lifetime is bound to our application.
This felt like a good small experiment to see how its changes would look
if we did this.
As a shortcut, `getApp` that is already a mega singleton keeps a
reference to Args, this means places that are a bit more difficult to
inject into call `getApp()->getArgs()` just like other things are
accessed.


pajlada:
I plan to merge this in in the next few days - I'm open to feedback in the meanwhile. Where feasible/practical I will take this approach further

pajlada:
image

PR #5042, opened 2023-12-25
perf: skip update from GIF timer if no animated elements are shown

Nerixyz:
Currently, when animated elements are turned on, a tick from the GIF timer (50x/s) will always trigger a repaint. This isn't needed.

This PR lays the foundation for more work on selective updates from the GIF timer by tracking elements that contain animations.
After this PR, the GIF timer will only trigger an update if at least one element was animated.

### DIFF
--- a/src/messages/layouts/MessageLayoutContainer.cpp
+++ b/src/messages/layouts/MessageLayoutContainer.cpp
@@ -238,12 +238,12 @@ void MessageLayoutContainer::paintElements(QPainter &painter,
 bool MessageLayoutContainer::paintAnimatedElements(QPainter &painter,
                                                    int yOffset) const
 {
-    bool anyElement = false;
+    bool anyAnimatedElement = false;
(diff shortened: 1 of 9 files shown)


src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -235,13 +235,15 @@ void MessageLayoutContainer::paintElements(QPainter &painter,
     }
 }
 
-void MessageLayoutContainer::paintAnimatedElements(QPainter &painter,
+bool MessageLayoutContainer::paintAnimatedElements(QPainter &painter,
                                                    int yOffset) const
 {
+    bool anyElement = false;
```
pajlada, line 241: nit: naming, should include something about animated (maybe anyAnimatedElement?)

pajlada (approved):
Looks good to me :+1: Just a nit on a variable naming and I'm happy to merge this in

PR #5045, opened 2023-12-27
fix: Add check for tall messages

### LINKED ISSUE #5044: Messages sometimes don't render, or flicker, if they are only partially visible with an animated emote
Most likely due to https://github.com/Chatterino/chatterino2/pull/5043

https://github.com/Chatterino/chatterino2/assets/962989/d828ae82-1640-409a-866d-c2b40d4de49e

Nerixyz:
Previously, only the bottom and the top were checked. This PR adds a check for tall messages (taller than the channel-view).

Fixes #5044.


pajlada (approved):
Looks like it fixed the issue I reported - thank you for the quick turnaround!

PR #5047, opened 2023-12-28
Improve docs/supplemental files for plugins

Nerixyz:
While trying to add docs for plugins on the wiki, I noticed some issues in the plugin docs:

- The JSON schema didn't allow `$schema` as a key.
- The TypeScript types contained the wrong event type.
- `register_callback` wasn't documented at all.
- No Lua metadata for plugins was present. [LuaLS](https://luals.github.io/) seems to be the most common language server.


pajlada (approved):
Looks good to me :+1: 
Would be cool if the `docs/plugin-meta.lua` file could be auto-generated since it's a bit of a duplicate of `LuaAPI.hpp`

Mm2PL:
>  Would be cool if the `docs/plugin-meta.lua` file could be auto-generated since it's a bit of a duplicate of `LuaAPI.hpp`

I can work on that. In the mean time this PR is a good step towards that.

Mm2PL (approved):

PR #5051, opened 2023-12-29
fix: non-native drag on Windows

### LINKED ISSUE #5050: Usercard regressions
### Checklist

### Describe your issue

aa11a241639094818f43898c9ad8910e689c1d8b makes dragging the usercard laggy (a bit hard to display, my monitor's 165Hz and the recording's only set to 60fps, may not look like it, but IRL it's very noticeable, can try recording at a higher framerate if necessary), and no longer responds to the context of the display's bounds.

### Screenshots

Left is before the commit, right is after (The recording was done at the top of my monitor/display)

Using a wm that doesn't respect the Qt limitations, like AltSnap, doesn't exhibit the lag, feels about the same as the left side/before the commit.

### OS and Chatterino Version

built on 2023-12-28 with Qt 6.6.0, MSVC 193833133 Running on Windows 10 Version 21H2, kernel: 10.0.19044

Nerixyz:
Setting the parent to a `Split` on the usercards and reply threads caused the [mouse-tracking code](https://github.com/Chatterino/chatterino2/blob/04a46f60dc26bc0ef56a2318624fae754cff4765/src/widgets/BaseWindow.cpp#L379-L392) in `BaseWindow` to check the `Split` and `SplitContainer` (both have mouse tracking).

Fixes #5050.

I noticed the reply thread popup having a similar bug in the header where the checkbox has mouse tracking but doesn't handle events if it's stretched, so I inserted a dummy widget that takes the remaining space.


CHANGELOG.md
```diff
@@ -57,6 +57,7 @@
 - Bugfix: Fixed a bug on Wayland where tooltips would spawn as separate windows instead of behaving like tooltips. (#4998, #5040)
 - Bugfix: Fixes to section deletion in text input fields. (#5013)
 - Bugfix: Show user text input within watch streak notices. (#5029)
+- Bugfix: Fixed draggable popups falling back to a non-native drag implementation on Windows. (#5051)
```
pajlada, line 60: If this bugfix doesn't fix something from the last stable release, it should just add its number to the changelog that caused it

pajlada (approved):

PR #5052, opened 2023-12-29
fix: some buttons triggering when releasing mouse outside

Nerixyz:
`Button::clicked` was emitted when the mouse was released outside the area of the button. This isn't expected. The avatar in usercards and the moderation button both used this.


pajlada (approved):
I can confirm I could reproduce the listed issues before this PR, and not with this PR

Thank you!

PR #5057, opened 2023-12-30
Improve color selection and display

Nerixyz:
This PR improves color selection and display by doing the following changes:

- A checkerboard is displayed where the color isn't opaque
- Switched to saturation-brightness canvas
- Added an alpha slider
- The CSS input now shows alpha and accepts `#rgb`, `#rrggbb`, and `#rrggbbaa` (this is the CSS one, not the Qt one)

| Before | After |
|---| ---|
| firefox_2023-12-30_16-29-55 | chatterino_2023-12-30_16-27-01 |
| firefox_2023-12-30_16-31-08 | firefox_2023-12-30_16-28-57 <br> same for users and badges |

I'm still not sure about the rounding of the widgets and the circle selector, so feedback is welcome!


pajlada:
Some quick feedback
1) At the current size, It looks like we should have room for at least one more row of recently used colors. Maybe we should add more recently used colors, or we should reduce the height of the dialog overall
1) I don't mind the roundedness of the recently used/default colors/selected color blobs
1) I don't mind the roundedness of the hue or alpha sliders
1) I like the new color widget thing in the highlight dialog
1) I miss the "Selected" label for the current color. I think it would be nice to have a similar box under it/next to it with the old color
1) The circle cursor would be good if it had a solid color of the actually selected color (e.g. how it's done here: https://coloris.js.org/examples.html). I understand that might be a bit more challenging to implement, so I'm fine with using a simple `+` cursor too

Nerixyz:
> At the current size, It looks like we should have room for at least one more row of recently used colors. Maybe we should add more recently used colors, or we should reduce the height of the dialog overall

The recent colors are just the colors of the message and user highlights. So if you have more entries with different colors, you'll have more recently used colors.

>     5. I miss the "Selected" label for the current color. I think it would be nice to have a similar box under it/next to it with the old color

Added.

> 6. The circle cursor would be good if it had a solid color of the actually selected color (e.g. how it's done here: [coloris.js.org/examples.html](https://coloris.js.org/examples.html)). I understand that might be a bit more challenging to implement, so I'm fine with using a simple `+` cursor too

I added the opaque color, but it doesn't really make much of a difference. At least in the alpha slider, I think it's better to be transparent. Doing it the way it's done on the web isn't as easy, as we need to add another mask when painting the checkerboard (which seems overkill).

pajlada:
Can revert the circle opaque color in that case, it was probably fine

src/widgets/helper/color/ColorSelect.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include <QBoxLayout>
+#include <QWidget>
+
+namespace chatterino {
+
+class ColorSelect : public QWidget
+{
+    Q_OBJECT
+    Q_PROPERTY(QColor color READ color WRITE setColor NOTIFY colorChanged)
```
pajlada, line 15: Remove `Q_PROPERTY` usage here & in other files

src/widgets/helper/color/ColorSelect.hpp
```diff
@@ -0,0 +1,46 @@
...
+    SBCanvas sbCanvas_;
+    HueSlider hueSlider_;
+    AlphaSlider alphaSlider_;
+    QVBoxLayout rows_;
+
+    void connectWidgets();
+
+signals:
+    void pickerColorChanged(QColor color);
+    void detailsColorChanged(QColor color);
```
pajlada, line 43: Remove these unused signals

src/widgets/helper/color/SBCanvas.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <QWidget>
+
+namespace chatterino {
+
+class SBCanvas : public QWidget
```
pajlada, line 7: Comment on the class what it is since "SB" isn't a known acronym to me, or just rename it to its full name

src/widgets/helper/color/SBCanvas.hpp
```diff
@@ -0,0 +1,55 @@
...
+    void setHue(int hue);
+    QSize sizeHint() const override;
+
+signals:
+    void saturationChanged(int saturation) const;
+    void brightnessChanged(int brightness) const;
+
+public slots:
+    int saturation() const;
+    int brightness() const;
```
pajlada, line 26: These should not be slots

src/widgets/helper/color/ColorDetails.cpp
```diff
@@ -0,0 +1,160 @@
...
+        return color.blue();
+    });
+    updateComponent(this->alpha_, [&]() {
+        return color.alpha();
+    });
+
+    this->updateCss();
+}
+
+void ColorDetails::updateCss()
```
pajlada, line 136: Call it Hex instead as discussed in chat

src/widgets/helper/color/ColorDetails.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <QGridLayout>
+#include <QLabel>
+#include <QLineEdit>
+#include <QSpinBox>
+#include <QWidget>
+
+namespace chatterino {
+
+class ColorDetails : public QWidget
```
pajlada, line 11: Class naming is iffy to me, maybe ColorInputs?

src/widgets/helper/color/ColorSelect.cpp
```diff
@@ -0,0 +1,86 @@
...
+    this->rows_.addWidget(&this->alphaSlider_);
+    this->setLayout(&this->rows_);
+    this->connectWidgets();
+}
+
+void ColorSelect::connectWidgets()
+{
+    const auto applyColor = [this]() {
+        emit colorChanged(this->currentColor_);
+    };
```
pajlada, line 26: Can we get rid of this lambda since it's a single-line anyway?

pajlada (commented):
If it's not the biggest hassle, I'd like to get rid of the `ColorSelect` "widget" (and maybe even the ColorDetails) and just add them straight to the ColorPickerDialog - going through the anatomy of the dialog has been a few too many classes deep imo, maybe that will make it a bit simpler

src/widgets/helper/color/ColorButton.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+QSize ColorButton::sizeHint() const
+{
+    return {50, 30};
+}
+
+void ColorButton::setColor(const QColor &color)
+{
+    this->currentColor_ = color;
+    this->update();
```
pajlada, line 25: Should we check to only call update if currentColor_ and color are not the same? or is that overengineering this use case

pajlada (approved):
Looks good to me! Just one quick review question & I'm happy to merge this in!

src/widgets/helper/color/ColorButton.cpp
Nerixyz: Yes we should check this.

pajlada (approved):

PR #5063, opened 2024-01-02
refactor: NetworkPrivate

### LINKED ISSUE #2949: NetworkRequest timeout timer starts before network request is sent
**Describe your issue**
All network requests done via `NetworkRequest` can timeout when the network request hasn't been sent/has just begun.

Discussion took place on #2900 

**OS and Chatterino Version**
N/A

Nerixyz:
Refactors network requests to be more extensible and less indented.

- Refactor `loadUncached` into a class (similar to `AvatarDownloader` in `Toasts.cpp`)
- The timeout timer is now started when the request is sent
- **All** event handlers now respect the concurrency 
- The network request type is now logged in PascalCase (due to the use of `magic_enum`; not sure about this one)
- Hashes can be computed concurrently (no need for a global mutex)
- Chatterino now warns about slow HTTP handlers ~~in the GUI thread~~ (enable with `CHATTERINO_WARN_SLOW_HTTP` in release mode). The threshold is rather low, as the number of requests often scales with the number of open channels.
- Removed custom boundary - this is done automatically for multipart requests

Fixes #2949.


src/common/NetworkPrivate.cpp
```diff
@@ -54,360 +149,311 @@ QString NetworkData::getHash()
...
+                {
+                    qCWarning(chatterinoHTTP)
+                        << "Slow HTTP success handler for" << url.toString()
+                        << timer.elapsed()
+                        << "ms (threshold:" << SLOW_HTTP_THRESHOLD << "ms)";
+                }
+            });
+        return;
+    }
+#endif
```
pajlada, line 236: Why is this all a separate runCallback call and not just inside the runCallback below?

pajlada (commented):
Haven't looked super much, but this stood out

I'd really really prefer not to add a compile flag or env variable for this, especially with special logic.
Can we not just always measure the time the CB takes & log if it takes too long?

src/common/NetworkPrivate.cpp
```diff
@@ -54,360 +149,311 @@ QString NetworkData::getHash()
...
+}
+
+void NetworkData::emitSuccess(NetworkResult &&result)
+{
+    if (!this->onSuccess_)
+    {
+        return;
+    }
+
+#ifdef CHATTERINO_WARN_SLOW_HTTP
```
Mm2PL, line 211: The execution path here is needlessly complex. The response timer can be put into the same lambda as the non-timer code path. I prefer the non-timer code path to be fully eliminated. The timer impact shouldn't be a lot more than calling `time()` twice.

src/common/NetworkPrivate.cpp
```diff
@@ -54,360 +149,311 @@ QString NetworkData::getHash()
...
+}
 
-        if (data->onReplyCreated_)
-        {
-            data->onReplyCreated_(reply);
-        }
+QLatin1String NetworkData::typeString() const
+{
+    auto view = magic_enum::enum_name<NetworkRequestType>(this->requestType_);
+    return QLatin1String{view.data(),
```
Mm2PL, line 291: Why is this latin1? Do compilers actually define typenames to be latin1, not ASCII or any of the popular Unicode encodings like UTF-8?

src/common/NetworkPrivate.cpp
```diff
@@ -3,16 +3,115 @@
...
+#    else
+constexpr qsizetype SLOW_HTTP_THRESHOLD = 90;
+#    endif
+#endif
+
+namespace {
+
+using namespace chatterino;
+
+/// Guard to call `deleteLater` on a QObject when destroyed.
```
Mm2PL, line 34: Maybe this should be moved somewhere else? It feels weird lying next to network internals.

src/common/NetworkPrivate.cpp
```diff
@@ -3,16 +3,115 @@
...
+
+#ifdef CHATTERINO_WARN_SLOW_HTTP
+#    include <QElapsedTimer>
+
+#    ifdef NDEBUG
+constexpr qsizetype SLOW_HTTP_THRESHOLD = 30;
+#    else
+constexpr qsizetype SLOW_HTTP_THRESHOLD = 90;
+#    endif
+#endif
```
Mm2PL, line 28: I don't like having different behaviors on debug and release builds when it can be avoided. I believe this should be axed to only have a single threshold and not depend on `NDEBUG`.

src/common/NetworkPrivate.cpp
```diff
@@ -54,360 +149,311 @@ QString NetworkData::getHash()
...
+void loadCached(std::shared_ptr<NetworkData> &&data)
+{
+    QFile cachedFile(getPaths()->cacheDirectory() + "/" + data->getHash());
 
-                case NetworkRequestType::Delete:
-                    return NetworkManager::accessManager.deleteResource(
-                        data->request_);
+    if (!cachedFile.exists() || !cachedFile.open(QIODevice::ReadOnly))
+    {
+        // File didn't exist OR File could not be opened
```
Mm2PL, line 174: This comment repeats the code above it. I don't think it helps, removing it is probably best.
```suggestion
```

src/common/NetworkPrivate.cpp
```diff
@@ -54,360 +149,311 @@ QString NetworkData::getHash()
...
-            if (data->hasCaller_ && data->caller_.isNull())
-            {
-                return;
-            }
-            data->onSuccess_(result);
+            // Operation cancelled, most likely timed out
+            qCDebug(chatterinoHTTP).noquote()
+                << this->data_->typeString() << "[cancelled]"
+                << this->data_->request_.url().toString();
+            return;
```
Mm2PL, line 402: This `if` can be moved outside of the NoError check.

Mm2PL (changes requested):
Just a couple things from looking around the code.

src/common/NetworkPrivate.cpp
Nerixyz: It was there because the handlers should only be checked when running in the GUI thread (as mentioned in the description). But as you don't see the use, I removed it.

src/common/NetworkPrivate.cpp
Nerixyz: > Why is this latin1?

Because it's the only way of printing a string-view over 8bit character data in Qt <6.5 (with 6.5 you can print `std::basic_string_view`). Otherwise, we'd need to create a `QString`.

> Do compilers actually define typenames to be latin1, not ASCII or any of the popular Unicode encodings like UTF-8?

It's a subset of Unicode (see [P1949R7](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1949r7.html)).

We can't use `QUtf8StringView`, because Qt 5 is supported. But realistically, Latin-1 is totally fine. The names we use for enum values are in the range `A-Z`, `a-z`, `0-9` (and possibly `_`) - these are all in US-ASCII and thus Latin-1.

src/common/NetworkPrivate.cpp
Nerixyz: It's not ideal, but I think it's better to have separate thresholds, as debug builds are usually slower - often by more than 2x. Furthermore, you can expect less performance from debug builds.

pajlada: I think this is acceptable since it's a best-effort thing, so we're trying to get useful data where possible to just look into things. We can easily adjust these values going forward

pajlada (approved):
I'm happy with this refactor :+1:

Mm2PL (approved):
The code looks good. Looks like everything works as expected. While I don't love the slow http threshold being different on debug and release builds, it can stay.

PR #5070, opened 2024-01-05
Add some tests for NotebookTab

pajlada:
- EmptyApplication: Add asserts to rest of getters (except for getSeventvAPI)
- Theme: make getTheme call getIApp()->getThemes() instead
- Use getIApp() instead of getApp() in a few places
- Add some NotebookTab tests

Related to https://github.com/Chatterino/chatterino2/pull/5068


pajlada (commented):
image

Issue #5073, opened 2024-01-08
Setting path to streamlink's executable AppImage only works without filename extension


SchimmelGelenk:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

After an excruciating amount of time trying to get streamlink work I realized the problem as stated in the title. So setting custom streamlink path in Chatterino lead to behaviour:

First:
"/home/schim/AppImages/streamlink.AppImage" -> chatterino.streamlink: Got error code 1

Then renaming the AppImage:
"/home/schim/AppImages/streamlink" -> works as intended 

### OS and Chatterino Version

Chatterino 2.4.6 (commit eb8c7f2 modified) built with Qt 6.5.3 Running on KDE Flatpak runtime, kernel: 6.6.6-76060606-generic, running from Flatpak

SchimmelGelenk:
While this can be expected behaviour, it did not occur to me as Streamlink Twitch GUI recognizes the executable regardless of the filename extension. Maybe adding a note in above the custom path box would suffice as the user is not aware how Chatterino handles the executable

pajlada:
The StreamLink errors/messaging can mention what the default binary path is, and what it expects the binary's name to be :+1: This would be a good first issue for someone looking to contribute

pajlada:
I'd say this is resolved by the improved documentation in #5076

PR #5076, opened 2024-01-10
Improve Streamlink documentation

fraxxio:
Solving #5073
Added new QLabel in ExternalToolsPage.cpp.

Note

### DIFF
--- a/src/widgets/settingspages/ExternalToolsPage.cpp
+++ b/src/widgets/settingspages/ExternalToolsPage.cpp
@@ -29,7 +29,7 @@ ExternalToolsPage::ExternalToolsPage()
         auto group = layout.emplace<QGroupBox>("Streamlink");
         auto groupLayout = group.setLayoutType<QFormLayout>();
 
-        auto description = new QLabel(
+        auto *description = new QLabel(
             "Streamlink is a command-line utility that pipes video streams "
             "from various "
             "services into a video player, such as VLC. Make sure to edit the "
@@ -37,7 +37,7 @@ ExternalToolsPage::ExternalToolsPage()
         description->setWordWrap(true);
         description->setStyleSheet("color: #bbb");
 
-        auto links = new QLabel(
+        auto *links = new QLabel(
             formatRichNamedLink("https://streamlink.github.io/", "Website") +
             " " +
             formatRichNamedLink(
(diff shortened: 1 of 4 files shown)


src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -60,6 +60,12 @@ ExternalToolsPage::ExternalToolsPage()
                                  getSettings()->streamlinkUseCustomPath);
         groupLayout->setWidget(2, QFormLayout::SpanningRole, customPathCb);
 
+        auto note =
+            new QLabel("The custom path may not need the file extension.");
+        note->setWordWrap(true);
+        note->setStyleSheet("color: #bbb");
+        groupLayout->setWidget(3, QFormLayout::SpanningRole, note);
```
Mm2PL, line 67: This is not accurate on Linux. To create a new process on Linux you *have* to have the full file name, extensions are not special.

Mm2PL (changes requested):
Doesn't apply to Linux and I don't think it applies to Mac either. Should probably be if-defed to Windows only.

src/widgets/settingspages/ExternalToolsPage.cpp
fraxxio: I misunderstood the issue, the problem is not with the custom path but with the name of the executable. I was able to recreate the issue #5073 on linux, where renaming the executable name from `streamlink.AppImage` to `streamlink` fixed the problem. 
So the note message should probably be `If Chatterino is not able to find Streamlink executable, renaming the executable to not have a file extension might help.`, or something like that. Also this message probably should only appear on Linux.

pajlada: It's probably best to say what Chatterino looks for instead, something like "Chatterino expects the binary to be called X" where X should be the name of the streamlink binary depending on the platform

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -60,6 +69,12 @@ ExternalToolsPage::ExternalToolsPage()
                                  getSettings()->streamlinkUseCustomPath);
         groupLayout->setWidget(2, QFormLayout::SpanningRole, customPathCb);
 
+        auto note = new QLabel("Chatterino expects the binary to be called ");
+        note->setText(note->text().append(getBinaryName()));
```
Nerixyz, line 73:
```suggestion
        auto note = new QLabel(QStringLiteral("Chatterino expects the binary to be called \"%1\".").arg(STREAMLINK_BINARY_NAME);
```

We can set the text correctly from the start.

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -14,6 +14,15 @@
...
 namespace chatterino {
 
+const char *getBinaryName()
+{
+#ifdef _WIN32
+    return "streamlink.exe";
+#else
+    return "streamlink";
+#endif
+}
```
Nerixyz, line 24: This should be in an anonymous namespace (`namespace { }`) above. Or, even better imo: create a `constexpr QStringView STREAMLINK_BINARY_NAME = u"abc";`

pajlada: @Nerixyz Could you give your opinion on how it's done here [`4e19758` (#5076)](https://github.com/Chatterino/chatterino2/pull/5076/commits/4e19758c047e368690a1d17ce05c2b895427da3b) ?

pajlada (approved):
Looks good to me, would like approval from the previous reviewers before merging this in

I took the opportunity to do some minor refactors at the same time, thank you for your contribution!

Mm2PL (approved):

Nerixyz (approved):

PR #5079, opened 2024-01-12
Refactor/remove operator!=

### LINKED ISSUE #4895: C++20 refactor: Investigate removing operator!=
See https://github.com/Chatterino/chatterino2/pull/4889#discussion_r1355622462 for more context

4rneee:
This is a simple refactor which removes the `operator!=` overloads since it is not needed in C++20 anymore.

Resolves https://github.com/Chatterino/chatterino2/issues/4895

According to [this](https://en.cppreference.com/w/cpp/compiler_support) our compilers should (hopefully) support this feature.


Nerixyz:
Imo, it's fine to use [gcc-10](https://packages.ubuntu.com/focal/gcc-10) for **building**. We could also think about using clang 17 (not sure if this works, but seems like it's [relatively easy to install](https://github.com/marketplace/actions/install-llvm-and-clang)).

pajlada:
Closing this for now as it'll be approached separately post-#5113

4rneee:
I just found this old branch and think that this should work now. I already force-pushed to the branch to clean it up.

Can you reopen this PR or should I create a new one?

Nerixyz:
Probably easier to open a new PR for this, since a lot has changed since then.

pajlada:
@4rneee  I can't re-open it  - iirc it has to do with you force pushing to your branch before the PR is opened so just create a new PR and link this one in the description

PR #5095, opened 2024-01-16
Use the same input padding between light & dark themes

pajlada:
For context, before this change there was a difference in the margin of a split input depending on if your theme was dark or light.
This PR makes the margin the same regardless of theme.

Dark mode:
dark-mode

Light mode before this PR:
light-mode-before

Light mode with this PR:
light-mode-after

Thoughts @treuks ?


treuks:
i cant really see what you mean by the margin being the same, there's a pretty huge difference

image psd(3)
image psd(4)

pajlada:
there's probably some other margin then being theme-specific then, the margin/padding/spacing this fixes is "inside" the text input (especially noticeable if you focus on the emote button)

PR #5107, opened 2024-01-19
deps(conan): use `boost` in header-only mode

Nerixyz:
Seems like we can use boost in header-only mode. This saves some time when running `conan install`. However, the OpenSSL dependency still slows the build down. This can be solved by moving the WebSocket clients to boost-beast and using boost-wintls on Windows, which uses the native SChannel (first, all WebSocket clients would need some kind of common wrapper to make the transition easier).

<sub>Yes, `boost::json` works.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,7 @@
 - Dev: Removed duplicate scale in settings dialog. (#5069)
 - Dev: Fix `NotebookTab` emitting updates for every message. (#5068)
 - Dev: Added benchmark for parsing and building recent messages. (#5071)
+- Dev: Boost is depended on as a header-only library when using conan. (#5107)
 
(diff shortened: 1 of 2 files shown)


pajlada (approved):

CHANGELOG.md
```diff
@@ -126,6 +126,7 @@
 - Dev: Removed duplicate scale in settings dialog. (#5069)
 - Dev: Fix `NotebookTab` emitting updates for every message. (#5068)
 - Dev: Added benchmark for parsing and building recent messages. (#5071)
+- Dev: Boost is depended on as a header-only library when using conan. (#5107)
```
Wissididom, line 129:
```suggestion
- Dev: Boost is dependant on as a header-only library when using conan. (#5107)
```
?

PR #5111, opened 2024-01-20
feat: add command line argument to select/add tab with a channel

### LINKED ISSUE #2498: Command line argument to launch chatterino with a specified tab selected.
There has been some talk about this already, in #2484 and #2460, as well as streamlink/streamlink-twitch-gui#607.

**What should be added?**
I'm suggesting a command line argument, maybe something like `-a, --active <channel>`, that launches chatterino with that specified tab selected. If there is no such tab in the layout, create it.

**Why should it be added?**
Right now when chatterino is launched, the user likely has a number of tabs open. Whichever tab was active (selected) when the user closed chatterino will be active now. This means that the user will likely have to do manual labor by selecting the tab (or opening a new tab) that corresponds to the stream that was just launched alongside chatterino.

Basically it would be a useful addition for scripting and for using chatterino in conjunction with something like streamlink.

I realise that I'm asking for functionality that automates a 1 second mouse move and click, but it would be nice to have this one day.

Nerixyz:
Adds a `-a <channel>`/`--activate <channel>` command line option to activate a tab containing a channel or adding one (the feature request proposed `--active`).

- Only Twitch channels are supported (special channels are not supported)
- Only the main window is searched for tabs.
- If no tab/main window with the channel is found, a new one is added. _When adding a new window, it will show an out-of-bounds warning, but it should be fairly rare that no main window is present (in that case you might as well specify a window-layout with `-c`)._
- Tabs with fewer splits are preferred.
- Splits without filters are preferred.

Fixes #2498.

### DIFF
--- a/src/common/Args.hpp
+++ b/src/common/Args.hpp
@@ -36,7 +36,7 @@ class Args
 public:
     struct Channel {
         ProviderId provider;
-        QString spec;
+        QString name;
     };
 
     Args() = default;
--- a/src/common/Args.hpp
+++ b/src/common/Args.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "common/ProviderId.hpp"
 #include "common/WindowDescriptors.hpp"
 
 #include <QApplication>
@@ -26,12 +27,18 @@ class Paths;
 /// -v, --verbose
 /// -V, --version
 /// -c, --channels=t:channel1;t:channel2;...
+/// -a, --activate=t:channel
 ///     --safe-mode
 ///
 /// See documentation on `QGuiApplication` for documentation on Qt arguments like -platform.
 class Args
 {
 public:
+    struct Channel {
+        ProviderId provider;
+        QString spec;
+    };
+
     Args() = default;
     Args(const QApplication &app, const Paths &paths);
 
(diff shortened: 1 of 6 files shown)


src/common/WindowDescriptors.cpp
```diff
@@ -229,4 +229,110 @@ WindowLayout WindowLayout::loadFromFile(const QString &path)
...
+                            .channelName_ = spec,
+                        }},
+                    },
+                },
+        });
+        return;
+    }
+
+    TabDescriptor *bestTab = nullptr;
+    // The tab score is calculted as follows:
```
pajlada, line 268:
```suggestion
    // The tab score is calculated as follows:
```

src/common/WindowDescriptors.cpp
```diff
@@ -229,4 +229,110 @@ WindowLayout WindowLayout::loadFromFile(const QString &path)
...
+    // +4 for every split
+    // +2 for the desired channel
```
pajlada, line 270: Should be possible to remove this score +2

src/common/Args.hpp
```diff
@@ -26,12 +27,18 @@ class Paths;
...
+/// -a, --activate=t:channel
 ///     --safe-mode
 ///
 /// See documentation on `QGuiApplication` for documentation on Qt arguments like -platform.
 class Args
 {
 public:
+    struct Channel {
+        ProviderId provider;
+        QString spec;
```
pajlada, line 39: Rename to `name`

pajlada (approved):

pajlada (approved):

PR #5113, opened 2024-01-20
dev: drop Qt 5.12 support and builds

Nerixyz:
As explained in https://github.com/Chatterino/chatterino2/discussions/4292, requiring Qt 5.15.2 and a better C++ 20 support is a big step.

**What changed?**

- Qt 5.15.2 is now a minimum requirement to build Chatterino
- GCC 10 is now a minimum requirement to build Chatterino (see #5079; I could split this up)
- The `.deb` file for Ubuntu 20.04 is no longer available (use the AppImage or Flatpak)

**What didn't change?**

- AppImage builds remain (and use Qt 5.15 now, since it's bundled)
- Qt 5 is still supported (seems like we're stuck with this for some time 🙃)

Setting `QT_DISABLE_DEPRECATED_BEFORE` to 6.0.0 isn't possible due to Qt 5.15.2 not having the non-deprecated position accessors for [`QMouseEvent`](https://doc.qt.io/qt-6/qmouseevent.html) (moving the filters away from non-deprecated methods is possible).

### DIFF
--- a/.docker/Dockerfile-ubuntu-20.04-base
+++ b/.docker/Dockerfile-ubuntu-20.04-base
@@ -21,7 +21,12 @@ RUN apt-get update && apt-get -y install --no-install-recommends \
         libxcb-image0 \
         libxcb-keysyms1 \
         libxcb-render-util0 \
-        libxcb-xinerama0
+        libxcb-xinerama0 \
+        gcc-10
+
+RUN update-alternatives \
+        --install /usr/bin/gcc gcc /usr/bin/gcc-10 60 \
+        --slave /usr/bin/g++ g++ /usr/bin/g++-10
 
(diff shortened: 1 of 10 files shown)


pajlada:
I'm not willing to drop Ubuntu 20.04 deb support yet. I'm looking into alternatives

Nerixyz:
> I'm looking into alternatives

I was wondering if we could ship Qt 5.15.2 with a deb file, but I don't know deb files enough.

pajlada:
Yeah that's essentially what I'm looking into. If it's possible, it should be possible to just ship Qt6 which would let us get rid of Qt5 altogether.
Currently installing openSUSE image to check out their build system

Nerixyz:
Superseded by #5170.

PR #5118, opened 2024-01-21
feat: add system theme on Qt 6.5 and up

### LINKED ISSUE #2452: Option to respect windows internal light and dark mode and switch accordingly
Enable a mode to automatically switch from Light to Dark Mode according to the system setting. I have a neat little app called Auto Dark Mode which automatically switches the Windows setting to Dark mode I use it in conjecture with Dark Reader to automatically enable Dark Mode on the websites I visit, but it doesn't affect the Chatterino Browser Extension and I have to manually switch the theme every time.

Nerixyz:
As explained in https://github.com/Chatterino/chatterino2/issues/2452#issuecomment-1452464154, Qt 6.5 added a signal for the system color scheme. The system theme I added is a dummy theme that resolves to the correct theme once `update` is called.

https://github.com/Chatterino/chatterino2/assets/19953266/4a13c346-3682-4077-91a7-a1fd40597fc5

Closes #2452.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Minor: Normalized the input padding between light & dark themes. (#5095)
 - Minor: Add `--activate <channel>` (or `-a`) command line option to activate or add a Twitch channel. (#5111)
 - Minor: Chatters from recent-messages are now added to autocompletion. (#5116)
+- Minor: Add _System_ theme that updates according to the system's color scheme (requires Qt 6.5). (#5118)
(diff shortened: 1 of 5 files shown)


pajlada:
I would like to see this use a few more settings in the backend to control
1) a boolean or enum option to choose to follow the system theme
2) a setting option (just like the theme selector now) where you can choose which theme to use when your system is using light mode
3) a setting option (just like the theme selector now) where you can choose which theme to use when your system is using dark mode

The UX might be a bit tricky to get right, so let's keep it simple in this first iteration & just add the checkbox/selector straight up to the settings page.

pajlada (approved):

PR #5119, opened 2024-01-21
feat: Show FrankerFaceZ channel badges

### LINKED ISSUE #2241: FrankerFaceZ Channel-Specific Badge Assignments

pajlada:
We already make the request, so this just parses more of the payload.
One thing of note, for the API parsing: FrankerFaceZ gives us Twitch User IDs as integers, while we always treat them as strings. For this, I've made sure to handle a potential change in the API if they change to strings.

For thread safety, I've added a thread guard which will assert in debug mode if the channel badge map is accessed or written to in different threads. On Arch Linux with Qt 6.6.3 this works as expected. This should be tested on macOS & Windows builds.

Fixes #2241


SputNikPlop:
@pajlada Tested on macOS and it looks good

pajlada (commented):
image

PR #5123, opened 2024-01-23
perf: add signal to invalidate buffers without doing layout

Nerixyz:
I wanted to fix a [bug in the 7TV fork](https://github.com/SevenTV/chatterino7/issues/259#issuecomment-1906731105) related to paints and noticed that I needed a way to just invalidate the paint buffers but not do layout (if it's not needed).

This PR adds a signal similar to `layoutRequested` that tells channel-views to discard the paint buffers. This is done by setting a flag when doing (re-)layout.

I noticed two settings that benefit from this: _Alternate background color_ and _Separate with lines_. You can test this PR by toggling the settings.

### DIFF
--- a/src/messages/layouts/MessageLayout.cpp
+++ b/src/messages/layouts/MessageLayout.cpp
@@ -74,7 +74,8 @@ int MessageLayout::getWidth() const
 
 // Layout
 // return true if redraw is required
-bool MessageLayout::layout(int width, float scale, MessageElementFlags flags)
+bool MessageLayout::layout(int width, float scale, MessageElementFlags flags,
+                           bool deleteBuffer)
 {
     //    BenchmarkGuard benchmark("MessageLayout::layout()");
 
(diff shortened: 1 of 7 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -425,6 +425,15 @@ void ChannelView::initializeSignals()
             }
         });
 
+    this->signalHolder_.managedConnect(
+        getIApp()->getWindows()->deleteBuffersRequested, [&](Channel *channel) {
```
pajlada, line 429: nit: I'd prefer us to be explicit in what we capture
```suggestion
        getIApp()->getWindows()->deleteBuffersRequested, [this](Channel *channel) {
```

src/messages/layouts/MessageLayout.cpp
```diff
@@ -108,6 +109,11 @@ bool MessageLayout::layout(int width, float scale, MessageElementFlags flags)
 
     if (!layoutRequired)
     {
+        if (deleteBuffer)
+        {
+            this->deleteBuffer();
+            return true;
+        }
```
pajlada, line 116: Should we check deleteBuffer on line 122 as well, where we check the width?

Nerixyz:
I noticed that invalidating is better than deleting, since we know the width of the message didn't change. This way we can re-use the old buffer.

pajlada (approved):

PR #5135, opened 2024-01-28
feat: add option to change the top-most status of a window

Nerixyz:
Adds a pin button next to the title bar buttons on Windows and in the Notebook on other platforms. If the _Always on top_ setting is on, this allows one to disable the top-most state of the popup. Not sure if this should be behind a setting.

Closes #3360.

https://github.com/Chatterino/chatterino2/assets/19953266/4c2c7419-e725-40ac-83a2-a3311997892a


pajlada (commented):
I'd prefer if this was a right-click option in the notebook/tabs in popups-only, rather than a button. The option should be whether or not to 'override' the current "always on top" setting.
If always on top is enabled, the right-click option would allow you to "un-always-on-top" the popup.
If always on top is disabled, the right-click option would allow you to "always-on-top" the popup

The setting shouldn't save/load, but rather just persist for as long as Chatterino runs.

If the "always on top" setting is changed, the override should be reset.
So if I have "always on top" enabled, and I right-click a popup & set it to not be always-on-top, then change "always on top" from enabled to disabled, then disabled to enabled, all windows, including my popup, should be "always-on-topped"

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Minor: Added icons for newer versions of macOS. (#5148)
 - Minor: Added the `--incognito/--no-incognito` options to the `/openurl` command, allowing you to override the "Open links in incognito/private mode" setting. (#5149)
 - Minor: Added support for the `{input.text}` placeholder in the **Split** -> **Run a command** hotkey. (#5130)
+- Minor: Added the ability to pin a popup window. (#5135)
```
pajlada, line 33: Should be updated to reflect the phrasing "top-most", and that it can be used to un-top-most a window. I don't know how to phrase it :1st_place_medal: . PR title should be the same as this entry basically

pajlada (commented):
Just a nitpick about the phrasing, then I'm happy to merge :+1: Thank you!

pajlada (approved):

PR #5137, opened 2024-01-29
fix: more MSVC warnings

Nerixyz:
Further working towards #4825.

Despite some rather simple fixes, the following was done:

- Defined `_WIN32_WINNT` to Windows 10 - this is what the inclusion of `Windows.h` defines. I'm not 100% sure if this is okay (e.g. looking at https://github.com/Chatterino/chatterino2/discussions/4852)
- Re-enabled [C4505](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4505) ('_function_': unreferenced local function has been removed), as we only need to declare `static` declarations (i.e. declarations for functions local to a translation unit) in headers as either `inline` or in a source file.
- Disabled [C4458](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4458) (declaration of '_identifier_' hides class member) as we have a rule of exclusively using `this->` to access members.

### DIFF
--- a/src/providers/twitch/api/Helix.cpp
+++ b/src/providers/twitch/api/Helix.cpp
@@ -2704,6 +2704,7 @@ void Helix::updateShieldMode(
                     {
                         failureCallback(Error::UserMissingScope, message);
                     }
+                    [[fallthrough]];
                 }
                 case 401: {
                     failureCallback(Error::Forwarded, message);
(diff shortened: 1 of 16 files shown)


src/providers/twitch/api/Helix.cpp
```diff
@@ -2704,6 +2704,7 @@ void Helix::updateShieldMode(
                     {
                         failureCallback(Error::UserMissingScope, message);
                     }
+                    [[fallthrough]];
```
pajlada, line 2707: I don't think we mean to fallthrough here - this feels like a bug.
Should add a break here & add an else case the "Missing scope" check & call `failureCallback(Error::Forwarded, message);`

pajlada (commented):
Small note on the passthrough thing, which is a good catch, otherwise looks good :+1:

pajlada (approved):

PR #5141, opened 2024-02-01
Add a new Channel API for experimental plugins feature

Mm2PL:
Example usage:
```lua
local pajs = c2.Channel.by_name("pajlada", c2.Platform.Twitch)
pajs:add_system_message("Hello there")
```

When Chatterino deletes the Channel, all lua ChannelRefs to it expire and using them will produce an error. It is possible to check for this before using those functions with `Channel:is_valid()`. I expect that other weak objects will have identical methods, I created IWeakResource interface class which defines this method.

## Breaking
This PR removes `c2.system_msg` and `c2.send_msg`.


docs/chatterino.d.ts
```diff
@@ -12,13 +12,43 @@ declare module c2 {
...
+  class Channel implements IWeakResource {
+    is_valid(): boolean;
+    get_name(): string;
+    get_type(): ChannelType;
+    get_display_name(): string;
+    is_twitch_channel(): boolean;
+
+    static by_name(name: string, platform: Platform): null | Channel;
+    static by_twitch_id(id: string): null | Channel;
+  }
```
Wissididom, line 45: Isn't there `send_message` and `add_system_message` missing, because `send_msg` and `system_msg` were removed a few lines below this class definition and in the `plugin-meta.lua` there are those functions defined as member of `Channel`?

Mm2PL: They are missing, thank you for reminding me of that

pajlada (commented):
LGTM, just add the missing documentation to `class Channel`

pajlada (approved):

4rneee:
I saw this PR just now and noticed two things:
1. With the removal of `c2.system_msg` and `c2.send_msg` we now need  the channel name and platform to send a (system-)message. The `CommandContext` should therefore be extended to also contain the platform so that replying to a command is possible without assuming that the command was triggered in a twitch channel.
2. Since this changes quit a lot how the API works it would be great if [docs/wip-plugins.md](https://github.com/Chatterino/chatterino2/tree/master/docs/wip-plugins.md) would be changed to stay consistent with the current implementation.

pajlada:
@4rneee Could you create an issue to track that? Super helpful feedback

PR #5144, opened 2024-02-02
fix: Twitch global emotes not loading

pajlada:
The attempt to load Twitch global badges failed because it was made before initialize had been called on AccountController, causing Helix to be in anonymous mode, causing the request to fail.
This was a bit harder to notice since we have a list of backup badges in case a user is not logged in, but it was reported in Discord just now.
Alongside the fix, I've updated our list of Twitch global badges.


pajlada:
image

PR #5155, opened 2024-02-05
chore: Build with Qt6 by default on Windows

pajlada:
Reported by @Leppunen - thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,7 +116,7 @@
 - Dev: Removed `Outcome` from network requests. (#4959)
 - Dev: Added Tests for Windows and MacOS in CI. (#4970, #5032)
 - Dev: Added "Copy message as JSON" option when shift-right-clicking a message. (#5150)
-- Dev: Windows build now build with Qt6 by default. (#5155)
+- Dev: Windows now builds with Qt6 by default. (#5155)
 - Dev: Move `clang-tidy` checker to its own CI job. (#4996)
 - Dev: Refactored the Image Uploader feature. (#4971)
 - Dev: Refactored the SplitOverlay code. (#5082)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,7 +116,7 @@
 - Dev: Removed `Outcome` from network requests. (#4959)
 - Dev: Added Tests for Windows and MacOS in CI. (#4970, #5032)
 - Dev: Added "Copy message as JSON" option when shift-right-clicking a message. (#5150)
-- Dev: Fixed the Windows build flags by adding the BUILD_WITH_QT6 flag, since it's the only supported build for Windows. (#5155)
+- Dev: Windows build now build with Qt6 by default. (#5155)
 - Dev: Move `clang-tidy` checker to its own CI job. (#4996)
 - Dev: Refactored the Image Uploader feature. (#4971)
 - Dev: Refactored the SplitOverlay code. (#5082)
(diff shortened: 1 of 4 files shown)


pajlada:
@Nerixyz could you sanity-check that I haven't missed something obvious here?

pajlada:
I ended up adding this as a "shared step" instead of updating every single build method section

Nerixyz (commented):
I think if we want this to be build documentation for new developers, we should include `-DBUILD_WITH_QT6=On` in the commands (_Building > Using CMake > Build_, _Developing with CLion_, and in the vcpkg docs). Sadly, there isn't a single character that both CMD and PowerShell support for multiline commands (CMD: ^, PowerShell: \`).

hemirt:
For future reference:

Using Qt Creator, using Qt 6.5.[] and 6.6.[] (the only two versions available)

Cmake (auto-)executed inside Qt Creator.
No extra flags changed.

Issue 1:

conan just failed setting up openssl 1.1.1
- error that did not make much sense - settings.compiler.runtime missing value
(yet in conan profile show, the value was there)
(Temp. fix - setting skip conan variable inside cmakelists.txt)

Issue 2:
It tried searching for Qt 5
(Temp. fix - using Qt 5 kit, it worked)

Issue 3:
After skipping conan, tries using vcpkg.
Vcpkg building qt_base_everywhere fails (might just be oom kill, though i still had plenty of free ram, plus swap space)
(Temp. fix - setting skip vcpkg variable inside cmakelists.txt)

pajlada:
Yeah I think we only support Qt6 on Windows now, so if you're doing a Qt5 build on Windows you're a bit on your own

Maybe it's time to split up each build method into its own file & have an owner/maintainer of each method.

hemirt:
> Yeah I think we only support Qt6 on Windows now, so if you're doing a Qt5 build on Windows you're a bit on your own
> 
> Maybe it's time to split up each build method into its own file & have an owner/maintainer of each method.

Yeah, but it's not an issue of Qt 5, it's an issue of Qt 6. The cmake file is not properly set up to use Qt 6.

pajlada:
> Yeah, but it's not an issue of Qt 5, it's an issue of Qt 6. The cmake file is not properly set up to use Qt 6.

Things build in CI, which uses conan & qt6 so it should still work

hemirt:
> > Yeah, but it's not an issue of Qt 5, it's an issue of Qt 6. The cmake file is not properly set up to use Qt 6.
> 
> Things build in CI, which uses conan & qt6 so it should still work

The issue is not that the build does not work (it seemed to work fine using command line), the issue is that the configuration step inside Qt Creator does not work. The cmake script is defaulting to wrong values. Hopefully it's clear what I mean now.

pajlada:
aaaaaaaaaaaaah ok yeah, issue 2 will be solved by this PR at least

hemirt:
I guess the conanfile should also be updated to include the latest (and I've heard, it's the last) patch to openssl1.1.1

CHANGELOG.md
```diff
@@ -116,6 +116,7 @@
 - Dev: Removed `Outcome` from network requests. (#4959)
 - Dev: Added Tests for Windows and MacOS in CI. (#4970, #5032)
 - Dev: Added "Copy message as JSON" option when shift-right-clicking a message. (#5150)
+- Dev: Windows build now build with Qt6 by default. (#5155)
```
Nerixyz, line 119:
```suggestion
- Dev: Windows now builds with Qt6 by default. (#5155)
```

hemirt:
If the Qt 5 is not in support, the PR checks for it could be removed.

pajlada:
For developers it's not supported at least, we can keep it around as artifacts & in CI as long as it doesn't use up any of our time

PR #5159, opened 2024-02-05
chore: Use OpenSSL 3 by default in conan

pajlada:
From https://github.com/Chatterino/chatterino2/pull/5155#issuecomment-1926689191


pajlada:
image

PR #5172, opened 2024-02-12
fix: resolve /live channel liveness using their channel ID

### LINKED ISSUE #5158: `/live` channels never resolve when the channel was populated before displayname casing could be obtained
The `/live` tab uses display name casing, which is fair since we use it to populate the tab name as well.

The problem is the `/live` tab will sometimes populate the channels before the displayname casing is loaded, and due to this the channel will never gray out.

image
These are the `/live` tabs for me from 2024-1-28, the ones that did not get displayname casing never resolved themselves.

Probably could be fixed by just accepting case-insensitive resolving, if we're watching `bobross` but `BobRoss` goes offline, that's good enough.

KleberPF:
This fix for #5158 uses the `TwitchChannel` `roomId()` as the message id when sending live messages (`TwitchMessageBuilder::liveMessage` and `TwitchMessageBuilder::liveSystemMessage`). I wasn't able to replicate the same logic for live messages sent through `NotificationController` (I couldn't find a way to access the `TwitchChannel` from there), so I went with the other option of using a case insensitive string compare. I accept any ideas on how I could do this better, especially the `NotificationController` bit.

(felanbird): Fixes #5158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 - Bugfix: Fixed double-click text selection moving its position with each new message. (#4898)
 - Bugfix: Fixed an issue where notifications on Windows would contain no or an old avatar. (#4899)
 - Bugfix: Fixed headers of tables in the settings switching to bold text when selected. (#4913)
+- Bugfix: Fixed an issue in the `/live` split that caused some channels to not get grayed-out when they went offline. (#5172)
 - Bugfix: Fixed tooltips appearing too large and/or away from the cursor. (#4920)
 - Bugfix: Fixed a crash when clicking `More messages below` button in a usercard and closing it quickly. (#4933)
 - Bugfix: Fixed thread popup window missing messages for nested threads. (#4923)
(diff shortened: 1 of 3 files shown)


Felanbird:
changelog.md entry
```- Bugfix: Fixed an issue where channels who were live before you opened Chatterino didn't always get greyedout in the `/live` tab when they went offline (#5172)```

Felanbird (approved):
This seems to work as expected, so far 3 of my channels loaded before displayName capitalization was obtained, have correctly been greyed-out in the `/live` tab when they went offline

KleberPF:
Thanks for the review. Should I add a commit with the changelog entry?

Felanbird:
> Thanks for the review. Should I add a commit with the changelog entry?

if you want, otherwise pajlada will do it on the weekend (and maybe move my comma)

pajlada (approved):
Little bit hard to test but it seems to work as expected - thank you!

pajlada:

PR #5173, opened 2024-02-13
fix: "reply to message" now takes usernames with @'s & commas into consideration

### LINKED ISSUE #5171: Reply does not work if you already have @name, inside chatbox
### Checklist

### Describe your issue

Right click on a name of person to insert @name, (or just type @name, ) and then when you try right click and reply to a message of his, it does not switch to the reply message/context, and the message is sent as normal message instead of a reply.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit fe1e498a) built on 2024-02-11 with Qt 6.5.0, MSVC 193732826 Running on Windows 11 Version 23H2, kernel: 10.0.22631

KleberPF:
This fixes not being able to reply to an user if you are already mentioning them before the reply action. Here are the possibilities:

### Before reply action
image
### After reply action
image

### Before reply action
image
### After reply action
image

### Before reply action
image
### After reply action
image

### Before reply action
image
### After reply action
image

### Before reply action
image
### After reply action
image

Fixes #5171


Felanbird:
```- Bugfix: Fixed an issue where you had to click the `reply` button twice if you already had that users @ in your input box. (#5173)```

I like this entry because I noticed it when testing the original report, and it feels possible that this is the problem users saw

pajlada (approved):
Thank you!
I added some unit tests, if you can think of any additional unit tests that should or should not work, feel free to add those

PR #5176, opened 2024-02-15
Live streams that are marked as reruns now mark a tab as yellow instead of red

### LINKED ISSUE #2648: Show an indicator in tab for channels casting reruns
#  **Describe the bug:**  #
**Reruns should show up in Chatterino like being live does. When a channel is live it says “(live)” next to the channel name in the title bar, whereas on a Rerun it should not show the channel as being offline, and rather it should say “(rerun)” next to the channel name.**

#  **To reproduce:**  #
**Start or have a rerun going on a channel and go to said channel on Chatterino.**

#  **Screenshots:**  #

Chatterino Rerun Issue 1

Chatterino Rerun Issue 2

#  **Chatterino version:**  #
**Chatterino 2.3.0 (commit 2db140d5a)**

#  **Operating system:**  #
**macOS Big Sur Version 11.2.3**

macOS Big Sur Version 11.2.3

#  **Additional information:**  #
**View above notes on this issue. Thank you.**

Mm2PL:
They removed official rerun field in Helix, so we have to rely on a hack now, but it works. Fixes #2648.
image

This was really simple and easy to make surprisingly.


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -797,6 +806,11 @@ bool TwitchChannel::isLive() const
     return this->streamStatus_.accessConst()->live;
 }
 
+bool TwitchChannel::isRerun() const
+{
+    return this->streamStatus_.accessConst()->live;
```
pajlada, line 811: This feels wrong

Mm2PL: Oh lmao, that's the worst copypaste-o I've done in a while

pajlada:
The yellow blip is not fully yellow, other than that this works as expected
image


It would also be nice if you add a comment to the red & yellow blip to say that they should be themeable

Mm2PL:
I agree on adding theming but I believe this is out of scope of this PR.

pajlada:
> I agree on adding theming but I believe this is out of scope of this PR.

Agreed - don't implement theming here but since we have multiple dots now it would be nice to have a comment regarding it for future adventurers

pajlada (approved):

PR #5177, opened 2024-02-15
fix: searching redemptions that require user input

KleberPF:
This fix stops discarding the previous `searchText` when building a message in `TwitchMessageBuilder::appendChannelPointRewardMessage`.

image

Let me know if any more tests are needed. From what I've seen, regular messages have an empty `searchText` by that point, so the search string doesn't become redundant.

Fixes: #4804


Felanbird:
`- Bugfix: Fixed an issue that prevented searching a redemption by it's title. (#5117)`
or
`- Bugfix: Fixed an issue preventing searching a redemption by it's title when the redemption contained text input. (#5117)`

pajlada (approved):

Issue #2007, opened 2020-09-28
Emote tooltip window can go out of screen bounds on first draw


jupjohn:
When the mouse moves ever an emote in a message for the first time (the first pixel into that emote) the tooltip window does not display the emote preview (#2006). When the mouse is moved 1 pixel further into the bounding box the image loads and the tooltip (sometimes) expands over the boundaries of the display. 

**Steps to reproduce:**
 1. Place window near bottom right corner of screen
 2. Have a chat window open with a message containing an emote at the bottom of the screen (so that a tooltip with origin at top left of the tooltip would go offscreen)
 3. Move the mouse 1 pixel into the emote's bounding box
 4. Move mouse a further 1 pixel into the box (image draws onto the tooltip causing the tooltip to resize & go off screen)

Any further mouse movements will correct the tooltip placement.

**Evidence:**
![image](https://user-images.githubusercontent.com/4962764/94391636-86bf7e80-0145-11eb-9402-ca5c004cf964.png)
Mouse is on the red arrow.

**Chatterino version**
`Chatterino 2.2.2 (commit 07fc0c3ad)`

**Operating system**
Windows 10 Enterprise Build 18362

Relates to #2006 (a precursor to this).

PR #5178, opened 2024-02-15
refactor: Turn link-info into its own element and class

Nerixyz:

- **Links (to web URLs) now have their own `MessageElement` - `LinkElement`**
  This element holds the data (tooltip, image, ...) about the link and both its lowercase and original URL. It returns the potentially unshortened URL when calling `getLink`.
- **`TextElement` now only stores the list of words**
  This makes it easy to swap the words (as they're implicitly shared).
- **`MessageLayoutElement::getLink` now uses the link of its creator if no override is set**
  Before, most elements created called `setLink(getLink())` this is now unnecessary. Only the twitch moderation element needs to set an override.
- **Added `LinkInfo` to represent a potentially resolved link**
  This class manages the lifecycle of the info (created → loading → { resolved, errored }).
- **`TooltipWidget` now takes a `LinkInfo`**
  This makes it possible to update the tooltip once the info has been resolved. However, this doesn't fix https://github.com/Chatterino/chatterino2/issues/2007 (I'll do that in another PR)


pajlada (commented):
Rename Lifecycle to State in LinkInfo
Have ChannelView own the "creation" of TooltipEntry's instead of TooltipWidget knowing what a LinkInfo is
Make LinkResolver a non-static class that lives in App (doesn't need to inherit from Singleton) that contains the logic for actually loading a link & updating a LinkInfo. This means instead of LinkInfo having an "ensureLoadingStarted", the LinkResolver class would have a "load" function that accepts a LinkInfo pointer/reference & updates it if necessary
Preferably no friend-classing to make this easier to mock, so if LinkInfo contains some public APIs to set state & what-not that's ok as long as it's documented

src/widgets/TooltipWidget.cpp
```diff
@@ -3,7 +3,10 @@
 #include "Application.hpp"
 #include "messages/Image.hpp"
 #include "singletons/Fonts.hpp"
+#include "singletons/Resources.hpp"
+#include "singletons/Settings.hpp"
 #include "singletons/WindowManager.hpp"
+#include "util/StreamerMode.hpp"
```
pajlada, line 9: Can these includes be removed?

src/providers/links/LinkResolver.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+class ILinkResolver
+{
+public:
+    ILinkResolver() = default;
+    virtual ~ILinkResolver() = default;
+    ILinkResolver(const ILinkResolver &) = default;
+    ILinkResolver(ILinkResolver &&) = delete;
+    ILinkResolver &operator=(const ILinkResolver &) = default;
+    ILinkResolver &operator=(ILinkResolver &&) = delete;
```
pajlada, line 15: delete instead of default copy ctor/operator

tests/src/SignalSply.hpp
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
pajlada, line 1: nit: Rename to SignalSpy

pajlada (approved):

PR #5179, opened 2024-02-17
feat: add the ability to unit test some aspects of SplitInput

pajlada:
- Make getLastUserThatWhisperedMe mockable in TwitchIrcServer
- Use getIApp in SplitInput
- Add test-only:-)-method to set a SplitInput's input text
- Add SplitInput tests

### DIFF
--- a/tests/src/SplitInput.cpp
+++ b/tests/src/SplitInput.cpp
@@ -1,0 +1,109 @@
+#include "widgets/splits/SplitInput.hpp"
+
+#include "common/Literals.hpp"
+#include "controllers/accounts/AccountController.hpp"
+#include "controllers/commands/Command.hpp"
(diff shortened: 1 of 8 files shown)


pajlada:
image

tests/src/SplitInput.cpp
```diff
@@ -0,0 +1,109 @@
...
+    AccountController accounts;
+    CommandController commands;
+    Emotes emotes;
+};
+
+class SplitInputFixture : public ::testing::Test
+{
+protected:
+    SplitInputFixture()
+        : split(new Split(nullptr))
```
Nerixyz, line 83: This split is never deleted, is that intentional?

pajlada: Not intentional, no - but yolo since it's just in tests

PR #5184, opened 2024-02-18
Make Lua CommandContext use a ChannelRef

### LINKED ISSUE #5181: Extend CommandContext of the Plugin-API to support different platforms
### Checklist

### Describe your issue

https://github.com/Chatterino/chatterino2/pull/5141 removed `c2.system_msg` and `c2.send_msg` from the plugin API.  
Now to be able to send (system-)messages we first have to create a `Channel` object with `c2.Channel.by_name` which expects the channel name and platform as parameters.  
`CommandContext` however only gives the developer the channel name and should be extended to also contain the platform so that replying to a command is possible without assuming that the command was triggered in a twitch channel. Alternatively `CommandContext` could directly contain a `Channel` object instead of a channel name as this would probably reduce the boilerplate code of first getting the channel by name in every command handler.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.4.6 (commit 4a5d4ef3) built with Qt 6.6.2 Running on Arch Linux, kernel: 6.7.4-arch1-1

Mm2PL:
Fixes #5181


pajlada (approved):
LGTM - @4rneee can you confirm this works as you expect?

4rneee:
> @4rneee can you confirm this works as you expect?

yes, works as expected

PR #5186, opened 2024-02-18
fix: tooltips showing out-of-bounds after loading images

### LINKED ISSUE #2007: Emote tooltip window can go out of screen bounds on first draw

Nerixyz:
This fixes #2007 (I guess "doesn't fix #X" counts as fixing).

When calling `(showAnd)moveTo` on `BaseWindow`, the last arguments are now saved and can be reused to recheck the bounds in `applyLastBoundsCheck` (no-op if `moveTo` was never called or called without performing a bounds-check). This is called when loading an image in the tooltip and when updating the link-tooltip after it changed.


pajlada (approved):
Works as expected on i3 on Arch Linux btw

PR #5188, opened 2024-02-18
Allow theming of tab live and rerun indicators

Mm2PL:
Nothing should change unless you are using a custom theme.


pajlada (approved):

pajlada:
Custom themes that were created before this now have black live indicators - this will be fixed in a follow-up PR

PR #5192, opened 2024-02-21
Estimate size of images to avoid layout shifts

Nerixyz:
Currently, every Image is estimated to be 16x16 before it's loaded. This can result in unexpected layout shifts when it takes some time to load an image.
Most of the providers we use for images either give a size hint in their response or provide estimated size values in the documentation (also mentioned in https://github.com/Chatterino/chatterino2/pull/3950/files#diff-24f7e36cd830c753f94fd3d004d9e0f441ea4f8728f02dad6c75904deb67daf6R507-R509). This PR makes use of these hints and reports the size of images not loaded yet based on these values. While testing, most of the mismatches I've seen were from Twitch emotes such as `:)` (you can add a debug log in `Image::height()` and disable the network cache).

I've also heard reports about image elements not resizing after loading, this PR would "fix" this (not the primary intention as it's not really a fix).

While this PR visually fixes some layout shifts, there will still be re-layouts when loading images. Can we fix this?

A tiny list of stuff that could be done regarding images:

- The 3x entry should be renamed to 4x and 7TV should provide the 4x image by default (everyone _except_ 7TV has 1x, 2x and 4x) (see also: https://github.com/Chatterino/chatterino2/pull/3812)
- The loading of images can be done in the worker thread we get by doing a concurrent request. I think we can get rid of the whole `assignDelayed` business and push updates immediately (only one per event-loop iteration, though).


src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -432,7 +432,7 @@ ImageSet SeventvEmotes::createImageSet(const QJsonObject &emoteData)
     auto baseUrl = host["url"].toString();
     auto files = host["files"].toArray();
 
-    std::array<ImagePtr, 3> sizes;
+    std::array<ImagePtr, 4> sizes;
```
pajlada, line 435: As discussed in chat, remove from this PR

pajlada (approved):
Looks good, the FFZ channel badges has been merged in now so might be an extra image from url to add

I'd prefer to get rid of the default parameters in the fromUrl function so we don't accidentally miss any expected sizes

pajlada (approved):

Issue #5190, opened 2024-02-21
@ing tab autocomplete to streamer username


bmundt:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

I cannot @ tab autocomplete the streamer. I see from other issues that the @ autocomplete has issues pulling from the whole chat and only pulls recent chatters. But, I feel like there should be an exception to this for the streamer, so you can @ them easily. Maybe even a shortcut to go to the streamer, since often people like to mimic the name of the streamer.

### Screenshots

![image](https://github.com/Chatterino/chatterino2/assets/11317863/c0acc327-1668-4b9f-ba33-7395e66d525b)

![image](https://github.com/Chatterino/chatterino2/assets/11317863/5ae17d0c-b781-4b25-b293-def63b0b9577)

### OS and Chatterino Version

Chatterino 2.4.6 (commit eb8c7f2d) built with Qt 6.5.0, MSVC 193532217 Running on Windows 10 Version 22H2, kernel: 10.0.19045

Felanbird:
Non-mods can't access the viewerlist anymore so we cannot confirm the streamer has the chat open, Hasan may be an exception to this logic due to the fact that he often has Chatterino on screen, but we can't magically use that information for every channel.

I'm not sure if I like the idea of faking a broadcaster into the `@mention` list

bmundt:
Or perhaps just having it at the top of the mention list like in the normal twitch website.

![image](https://github.com/Chatterino/chatterino2/assets/11317863/4c351302-3a65-4fd4-a396-29b729855c5e)

Felanbird:
He's at the top of the mention list because you are still allowed to view the viewerlist in browser chat, that suggestion doesn't change my previous answer

KleberPF:
On the browser chat, even when the streamer is offline, they are the first option when typing @. I agree that it would be nice to also do this on Chatterino.

pajlada:
I am fine with adding the streamer to the user autocomplete list, even if they're not in the chatters list or among the recent chatters.
When added, there should be a setting that can be disable this behaviour.

If my memory isn't failing me, I believe we did this before - prior to the chatter list changes

[...]

PR #5193, opened 2024-02-21
Add option always include broadcaster in user completions

KleberPF:
Settings text probably needs to be improved. 

## Setting off:
image
image

## Setting on:
image
image

I noticed we got some unit tests for ChatterSet, so let me know if I should add any.

Fixes: #5190


resources/contributors.txt
```diff
@@ -68,6 +68,7 @@ chrrs | https://github.com/chrrs | | Contributor
 crazysmc | https://github.com/crazysmc | :/avatars/crazysmc.png | Contributor
 SputNikPlop | https://github.com/SputNikPlop | | Contributor
 fraxx | https://github.com/fraxxio | :/avatars/fraxx.png | Contributor
+KleberPF | https://github.com/KleberPF | | Contributor
```
Felanbird, line 71: seperate PR

src/common/ChatterSet.cpp
```diff
@@ -62,6 +62,22 @@ std::vector<QString> ChatterSet::filterByPrefix(const QString &prefix) const
     return result;
 }
 
+std::vector<std::pair<QString, QString>> ChatterSet::allExcept(
+    const QString &userName) const
+{
+    std::vector<std::pair<QString, QString>> result;
```
Nerixyz, line 68: This should be resized to `this->items.size()` to prevent reallocations, since we know an upper bound of the size.

src/common/ChatterSet.cpp
```diff
@@ -62,6 +62,22 @@ std::vector<QString> ChatterSet::filterByPrefix(const QString &prefix) const
...
+
+    for (auto &&item : this->items)
```
Nerixyz, line 70:
```suggestion
    for (const auto &item : this->items)
```

We can't move out of the item, since the method is `const` (I know the other methods do the same).

src/common/ChatterSet.hpp
```diff
@@ -39,6 +39,12 @@ class ChatterSet
     /// are in mixed case if available.
     std::vector<QString> filterByPrefix(const QString &prefix) const;
 
+    /// Get all recent chatters, except the one with the matched username.
+    /// The first pair element contains the username in lowercase,
+    /// while the second pair element is the original case.
```
Nerixyz, line 44: This should mention that the username is the lowercase name (or: the login name).

Nerixyz (commented):
Looks good! Maybe the setting name could be a bit shorter with some tooltip text, but I can't think of anything good right now 🙃

If you want, you could [add a test case](https://github.com/Chatterino/chatterino2/blob/ffa63f0dc96fb1054947833617cc070914a889b0/tests/src/ChatterSet.cpp) (maybe testing the case-sensitivity).

src/controllers/completion/sources/UserSource.cpp
```diff
@@ -58,7 +58,16 @@ void UserSource::initializeFromChannel(const Channel *channel)
...
+    if (getSettings()->showBroadcasterUsernameFirst)
+    {
+        this->items_ = tc->accessChatters()->allExcept(tc->getName());
+        this->items_.insert(this->items_.begin(),
+                            {tc->getName(), tc->getDisplayName()});
+    }
+    else
+    {
+        this->items_ = tc->accessChatters()->all();
+    }
```
pajlada, line 70: Would it be feasible to do the remove/move broadcaster logic in this function entirely? thus removing the `allExcept` function from ChatterSet

pajlada:
The "at the top" isn't important to me, just that the streamer is included. So for the variable names & the setting labels that should be reflected (e.g. "always include broadcaster in user completions" or something)

src/controllers/completion/sources/UserSource.cpp
KleberPF: I guess so. Is there any way to do it other than store a copy, erase and then insert the copy? Is this more about performance or code quality?

KleberPF:
Changed the setting name (might be too big now but I feel like it's reads well)

src/common/ChatterSet.cpp
KleberPF: I'll wait on your suggestions in case we keep `allExcept`

src/controllers/completion/sources/UserSource.cpp
pajlada: More about code quality than performance yeah, the `allExcept` function sticks out a bit. It's already not going to be the most performant code, so I'd prefer it be more readable/implemented closer to where it's used

src/controllers/completion/sources/UserSource.cpp
```diff
@@ -58,7 +58,16 @@ void UserSource::initializeFromChannel(const Channel *channel)
...
+    if (getSettings()->alwaysIncludeBroadcasterInMentions)
+    {
+        this->items_ = tc->accessChatters()->allExcept(tc->getName());
+        this->items_.insert(this->items_.begin(),
+                            {tc->getName(), tc->getDisplayName()});
+    }
+    else
+    {
+        this->items_ = tc->accessChatters()->all();
+    }
```
KleberPF, line 70: @pajlada I wasn't able to add the suggestion to the other comment. How does this look?

```suggestion
    if (getSettings()->alwaysIncludeBroadcasterInMentions)
    {
        auto broadcaster = tc->getName();
        auto it = std::find_if(this->items_.begin(), this->items_.end(),
                               [&broadcaster](const UserItem &user) {
                                   return user.first == broadcaster;
                               });
        if (it != this->items_.end())
        {
            std::rotate(it, it + 1, this->items_.begin());
        }
        else
        {
            this->items_.insert(this->items_.begin(),
                                {tc->getName(), tc->getDisplayName()});
        }
    }
```

pajlada: That looks reasonable at first glance to me, commit that & I will test & review fully this weekend

src/controllers/completion/sources/UserSource.cpp
KleberPF: Replied in a new comment

pajlada: The proposed solution in the new comment seems good to me :+1:

src/controllers/completion/sources/UserSource.cpp
Nerixyz: The `std::rotate` will break the order of the completions (the last chatter won't be first). Not sure if people rely on that.

KleberPF: The rotate logic also throws when the streamer is already at the last position in the vector. I'll fix this logic (maybe will have to erase/insert) and add some unit tests later

Nerixyz: > [a, b, c, streamer, e] would become [streamer, a, b, c, e]

That would become [streamer, e, a, b, c].

[Cppreference](https://en.cppreference.com/w/cpp/algorithm/rotate) has further explanation on this.

You can do this in one loop using `emplace` and `erase` I believe:

```cpp
// auto items = ...;
// auto streamer = {streamerLogin, streamerName};
auto it = items.emplace(items.begin(), streamer);
while(it != items.end()) {
  if (it->first == streamerLogin) {
    it = items.erase(it);
  } else {
    it++;
  }
}
```

KleberPF:
@pajlada ended up using `find` and `erase`, everything seems to be working. It doesn't seeem too hard to add InputCompletion unit tests, just mock `accessChatters` and create a `ChatterSet` using preset names, but I'm a Visual Studio guy and CMake is kicking my ass. I wasn't able to figure out how to compile the tests, so if you could give me some pointers on how to do that I'd appreciate it a lot (of course, feel free to add the tests yourself if you want)

pajlada (approved):

PR #5198, opened 2024-02-24
feat: Add a fallback theme to custom themes

pajlada:
If no fallback theme is specified by the custom theme, the fallback theme will be "Dark"


docs/ChatterinoTheme.schema.json
```diff
@@ -390,6 +390,11 @@
           "$comment": "Determines which icons to use. 'dark' will use dark icons (best for a light theme). 'light' will use light icons.",
           "enum": ["light", "dark"],
           "default": "light"
+        },
+        "fallbackTheme": {
+          "$comment": "Determined which built-in Chatterino theme to use as a fallback in case a color isn't configured.",
```
Nerixyz, line 395: The comment above uses "Determines", this uses "Determined".

src/singletons/Theme.cpp
```diff
@@ -25,15 +25,21 @@ namespace {
...
 {
-    const auto &jsonValue = obj[key];
+    auto jsonValue = obj[key];
     if (!jsonValue.isString()) [[unlikely]]
     {
-        qCWarning(chatterinoTheme) << key
-                                   << "was expected but not found in the "
-                                      "current theme - using previous value.";
-        return;
+        jsonValue = fallbackObj[key];
```
Nerixyz, line 34: I can't comment there, but there's also a case where the color is invalid (below) which should be handled too.

src/singletons/Theme.cpp
```diff
@@ -422,13 +463,29 @@ std::optional<ThemeDescriptor> Theme::findThemeByKey(const QString &key)
...
-
     this->isLight_ =
         root["metadata"_L1]["iconTheme"_L1].toString() == u"dark"_s;
 
+    std::optional<QJsonObject> fallbackTheme;
+    if (isCustomTheme)
+    {
+        // Only attempt to load a fallback theme if the theme we're loading is a custom theme
+        auto fallbackThemeName =
+            root["metadata"_L1]["fallbackTheme"_L1].toString("Dark");
```
Nerixyz, line 476: This should probably fall back to "Light" if `this->isLight_` and "Dark" otherwise.

src/singletons/Theme.cpp
pajlada: Good catch - resolved by [`e748cf8` (#5198)](https://github.com/Chatterino/chatterino2/pull/5198/commits/e748cf84be5fe12812eecd3408f1a7bca3c6ef33)

src/singletons/Theme.cpp
pajlada: Should be resolved by [`e7d98e6` (#5198)](https://github.com/Chatterino/chatterino2/pull/5198/commits/e7d98e6397400d48b1c949d9dc88115b1c627bcc)

The reason I did not implement this in the first place is that themes don't have a concept of "dark" or "light" themes really, just which color icons they use, which is the opposite of dark/light theme which felt spaghetti. But I'm fine with this solution

docs/ChatterinoTheme.schema.json
pajlada: Fixed in a force push - I never made this typo this must've been a bug in GitHub's interface :)

pajlada:
image

PR #4962, opened 2023-11-12
feat: Add initial EventSub Helix support

pajlada:
# Description

- Add support for the Create EventSub Subscription Helix API
- Add debug command to create an eventsub subscription

This PR is expected to fail in CI - to build locally you need to have the `https://github.com/pajlada/beast-websocket-client` repo cloned in the same dir as your c2 repo is cloned.
I will include this as a submodule once this PR is ready to be merged in.

PR #5200, opened 2024-02-25
feat: Send messages using Helix API

### LINKED ISSUE #5127: Implement "Send Chat Message" over Helix
https://dev.twitch.tv/docs/api/reference/#send-chat-message

Nerixyz:
This changes the default way messages are sent to use the Helix API instead of IRC. There's still an option to use IRC.

I decided to use a struct as the parameters to the Helix call to make help callers get the order right (using `{.property = value}`), since it would be four `QString` values in a row.

This augments #4962.
Fixes #5127.

### DIFF
--- a/src/providers/twitch/api/Helix.cpp
+++ b/src/providers/twitch/api/Helix.cpp
@@ -2840,7 +2840,7 @@ void Helix::sendShoutout(
         .execute();
 }
 
-// https://dev.twitch.tv/docs/api/reference/#send-a-shoutout
+// https://dev.twitch.tv/docs/api/reference/#send-chat-message
 void Helix::sendChatMessage(
     HelixSendMessageArgs args, ResultCallback<HelixSentMessage> successCallback,
     FailureCallback<HelixSendMessageError, QString> failureCallback)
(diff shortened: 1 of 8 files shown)


src/providers/twitch/api/Helix.cpp
```diff
@@ -2840,6 +2840,94 @@ void Helix::sendShoutout(
         .execute();
 }
 
+// https://dev.twitch.tv/docs/api/reference/#send-a-shoutout
```
Wissididom, line 2843:
```suggestion
// https://dev.twitch.tv/docs/api/reference/#send-chat-message
```

pajlada, line 2843:
```suggestion
// https://dev.twitch.tv/docs/api/reference/#send-chat-message
```

src/singletons/Settings.hpp
```diff
@@ -524,6 +524,8 @@ class Settings
         HelixTimegateOverride::Timegate,
     };
 
+    BoolSetting enableHelixChatSend = {"/misc/helixChatSend", true};
```
pajlada, line 527: I would prefer if this is an enum setting with the values:
 - Default (this should be the default value)
 - IRC
 - Helix

pajlada (commented):
and as mentioned in chat, the default should be IRC for now

pajlada (approved):

PR #5201, opened 2024-02-25
fix: show correct scale factor of images

Nerixyz:
Before this PR, images had the static 1x, 2x, and 3x labels. 1x and 2x were correct, while 3x was incorrect (except for 7TV), since these images were four times larger than the original one. This PR updates the labels for the items to be the factor based on the smallest image.

I'll update the 7TV code to use 4x (when available) in another PR (could also do it here if you want).


pajlada (commented):
Not really a bugfix, just a difference in what 1x/2x/3x/4x represents.
Changelog entry could be Minor & something along the lines of "Image links now reflect the scale of the image instead of its internal label" or "Image links now reflect the scale of the image instead of the image host's label"

PR #5212, opened 2024-02-27
fix(eventsub): compilation issues

Nerixyz:
Compiling the library as an object library causes some object files to not be linked to the final executable if it depends on the library transitively. Compiling the library as a static library (like other libraries) solves this. I don't know why the default is an object library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,7 @@ else()
     add_subdirectory("${CMAKE_SOURCE_DIR}/lib/settings" EXCLUDE_FROM_ALL)
 endif()
 
+set(TWITCH_EVENTSUB_WS_LIBRARY_TYPE STATIC)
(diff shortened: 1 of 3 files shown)


pajlada:
> At least Boost 1.83 is required for `boost::json::try_value_to( value const&, Context const&)` (Ubuntu uses 1.81)

Once https://github.com/Chatterino/docker/commit/ea62e3c891762d21dda801bfdccb33d57d81d303 CI is done, Ubuntu builds will use Boost 1.84

Nerixyz:
> Once [Chatterino/docker@ea62e3c](https://github.com/Chatterino/docker/commit/ea62e3c891762d21dda801bfdccb33d57d81d303) CI is done, Ubuntu builds will use Boost 1.84

The docker image [doesn't appear to be pushed](https://github.com/orgs/Chatterino/packages).

pajlada:
hopefully will be fixed by https://github.com/Chatterino/docker/pull/8 xd

Issue #5206, opened 2024-02-26
Make "AutoMod Caught Messages" highlight color & "show in mentions" configurable


djinnet:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

I cannot change the highlight color to any other colors in highlight page under messages tab. 
Its color picker is disabled for no reasons and it makes it harder to spot monitored or restricted users when it is blended with the background.  

### Screenshots

![image](https://github.com/Chatterino/chatterino2/assets/9974608/3a871e6a-4993-40f1-adaa-6a56dfe09306)

As you can see here; A has a color picker that allowed them to have different colors.
However B is disabled and has the same color as the darkmode's background and if the user of the program did not pay attention to this behavior, they can missed the message entirely. 

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit 0f42df78) built on 2024-02-26 with Qt 6.5.0, MSVC 193833135 Running on Windows 10 Version 22H2, kernel: 10.0.19045

pajlada:
I saw 3 separate feature requests here so I split your issue up
1) Make "AutoMod Caught Messages" configurable (highlight color & show in mentions)
2) Make "Restricted users" similary configurable #5207
3) Add support for "Monitored users" #5208

In the interim for this issue, there's a separate split called "AutoMod" that shows auto mod messages from every channel you have open that might be helpful

KleberPF:
Should the allow/deny message also show up in /mentions? If so, should the added/removed term messages also appear?

pajlada:
> Should the allow/deny message also show up in /mentions?

I think yes

> If so, should the added/removed term messages also appear?

I don't think this is important, no

Do you have any thoughts on this @djinnet?

djinnet:
I 100% agreed on the first one. @pajlada That could make life easier for moderation purpose in a massive community.

But for second one; I honestly dont know. 
That depends on if it is easy to implemented, but perhaps that could be a choice for the end-user to have that appear or not.
After all many people are using chatterino with different setups and perhaps some may need it. 
But that is a "nice-to-have" from my point of view if that makes sense? I would love to hear more thoughts on that.

[...]

PR #5215, opened 2024-02-28
Add highlight color and show in mentions to automod messages

KleberPF:
For now, the added/removed term messages don't show up in /mentions, but I'll add a commit doing that if we decide to do it.

## Previous messages change color when the user changes settings
image
image

## Scrollbar color
image

## Scrollbar color also changes when settings are changed
image

## Message showing up in mentions
image

## Message doesn't show up in mentions if setting is disabled
image
image

I was hitting an assert when allowing/denying a message, but I tested in master and it also happens there. It seems to be caused by the request created in `Helix::manageAutoModMessages` not being concurrent, but I didn't look too much into it. This was the call stack when the assert triggered

image

Fixes: #5206


Nerixyz:
You didn't cause the assertion issue: `isInStreamerMode` is called in a PubSub handler: https://github.com/Chatterino/chatterino2/blob/ea19c5c989dbea36b114bd524d50457adcb27206/src/Application.cpp#L710
The PubSub handler is running in it's own thread - not the GUI thread. `isInStreamerMode` isn't annotated, but has to run on a single thread as it accesses `cache`. This thread must be the GUI thread as it emits `streamerModeChanged`.

KleberPF:
Oh, nice. I think I mixed some things up. I put a breakpoint in `isGuiThread` for when the result was false, but this function isn't called just inside `assertInGuiThread`, and it ends up begin called by some functions in `NetworkData`. Ended up posting the correct call stack

src/messages/layouts/MessageLayout.cpp
```diff
@@ -376,11 +376,22 @@ void MessageLayout::updateBuffer(QPixmap *buffer,
              this->message_->flags.has(MessageFlag::LowTrustUsers))
     {
         backgroundColor = QColor("#404040");
+        if (this->message_->flags.has(MessageFlag::AutoModOffendingMessage) &&
+            ctx.preferences.enableAutomodHighlight)
+        {
+            backgroundColor =
+                *ctx.colorProvider.color(ColorType::AutomodHighlight);
+        }
```
KleberPF, line 384: Not sure if I should blend colors here

src/messages/layouts/MessageLayout.cpp
```diff
@@ -376,11 +376,22 @@ void MessageLayout::updateBuffer(QPixmap *buffer,
...
     }
     else if (this->message_->flags.has(MessageFlag::Debug))
     {
         backgroundColor = QColor("#4A273D");
     }
+    else if (this->message_->flags.has(MessageFlag::AutoModOffendingMessage) &&
+             ctx.preferences.enableAutomodHighlight)
+    {
+        backgroundColor = *ctx.colorProvider.color(ColorType::AutomodHighlight);
+    }
```
KleberPF, line 394: Forgot to remove, commenting to remind me

src/messages/layouts/MessageLayout.cpp
Nerixyz: Yes, they should be blended, but with the original background color - not the automod color. Previously, they weren't blended, as `#404040` is fully opaque. Now, users can set a color with transparency which would blend with the default color of a `QPixmap` - the backing buffer for a message (that one is actually undefined but usually some light gray with MSVC).

You could put the assignment to the default color in the `else` branch.

KleberPF: Got it, I think it's fixed now.

pajlada (commented):
I think the "header", i.e. message that contains the Allow & Deny buttons, should be colored based on the AutomodHighlight

src/providers/colors/ColorProvider.cpp
```diff
@@ -128,6 +128,9 @@ void ColorProvider::initTypeColorMap()
     initColor(ColorType::ThreadMessageHighlight,
               getSettings()->threadHighlightColor,
               HighlightPhrase::FALLBACK_THREAD_HIGHLIGHT_COLOR);
+
+    initColor(ColorType::AutomodHighlight, getSettings()->automodHighlightColor,
+              HighlightPhrase::FALLBACK_HIGHLIGHT_COLOR);
```
pajlada, line 133: This should probably have its own fallback color `#404040`, which was the default automod highlight color before

KleberPF:
> I think the "header", i.e. message that contains the Allow & Deny buttons, should be colored based on the AutomodHighlight

Is it ok to set `MessageFlag::AutoModOffendingMessage` for that message or should I make a new flag?

src/providers/colors/ColorProvider.cpp
KleberPF: Added `HighlightPhrase::FALLBACK_AUTOMOD_HIGHLIGHT_COLOR` with color `QColor(64, 64, 64);`

Nerixyz:
> Is it ok to set `MessageFlag::AutoModOffendingMessage` for that message or should I make a new flag?

The issue with reusing the flag is that you'd create two highlights. You can avoid creating a new flag by checking for `MessageFlag::Timeout`, which is only set for offending messages, and the "header" in combination with `AutoMod`.

KleberPF:
> The issue with reusing the flag is that you'd create two highlights. You can avoid creating a new flag by checking for `MessageFlag::Timeout`, which is only set for offending messages, and the "header" in combination with `AutoMod`.

Something like this?

```
if (ctx.preferences.enableAutomodHighlight && 
            (this->message_->flags.has(MessageFlag::AutoModOffendingMessage) || 
             this->message_->flags.has(MessageFlag::Timeout))
```

> checking for `MessageFlag::Timeout`, which is only set for offending messages

Is it a good idea to rely on this? I feel like it could break something in the future

Nerixyz:
> Is it a good idea to rely on this? I feel like it could break something in the future (I guess it's unlikely because we would the same combination of flags)

Yea, it's probably better to have a new flag.

KleberPF:
Added coloring for the message header
image
in mentions
image

Mm2PL:
Is it intended that AutoMod's "Hey! Your message is being checked" is not customizable with your changes?

KleberPF:
> Is it intended that AutoMod's "Hey! Your message is being checked" is not customizable with your changes?

Good question. I interpreted the issue as "highlight messages captured by automod", not "highlight that automod caught your message", but I'm not sure about the original intent.

Mm2PL:
They look really similar right now on `master` and are from the same source. I am of the opinion that they should be thrown into the same bag as mod-side automod messages. I don't think the issue author even considered their existence. Interestingly enough currently these messages don't show up in `/automod`.

pajlada:
I interpreted this as a moderator/broadcaster tool, so that's what I think is important

Mm2PL (approved):

pajlada (approved):

pajlada (approved):
Thank you @KleberPF !
Please don't forget to make a separate PR adding yourself into the About page :)

PR #5216, opened 2024-02-29
refactor: turn `StreamerMode` into a singleton(-like thing)

Nerixyz:
StreamerMode, as noticed in https://github.com/Chatterino/chatterino2/pull/5215, has some issues:

- It must be called from the GUI thread, but it doesn't really enforce that
- Calling `isInStreamerMode` can cause a process to be spawned on Unix (it should be fast though)
- Calling `isInStreamerMode` on Windows with a broadcasting binary open causes a memory leak
- It can't be mocked

`(I)StreamerMode` is now a singleton(-like) class that can be used from any thread. `isEnabled` is an atomic load, and connecting to `changed` from different threads is handled by `Qt::AutoConnection` (default).


src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+            getApp()->twitch->addGlobalSystemMessage(
+                "Streamer Mode is set to Automatic, but pgrep is missing. "
+                "Install it to fix the issue or set Streamer Mode to "
+                "Enabled or Disabled in the Settings.");
+        });
+    }
+
+    qCWarning(chatterinoStreamerMode) << "pgrep execution timed out!";
+    return false;
+#elif defined(Q_OS_WIN)  // TODO(C++ 23): use elifdef
```
pajlada, line 80: Remove todo

src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+void StreamerModePrivate::setEnabled(bool enabled)
+{
+    if (enabled == this->enabled_.load(std::memory_order::relaxed))
+    {
+        return;
+    }
+
+    this->enabled_.store(enabled, std::memory_order::relaxed);
+    this->parent_->updated();
```
pajlada, line 216: Pass the `enabled` value through here

src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+    [[nodiscard]] bool isEnabled() const;
+
+private:
+    void settingChanged(StreamerModeSetting value);
+    void setEnabled(bool enabled);
+
+    void check();
+
+    StreamerMode *parent_;
+    SingleConnection settingConnection_;
```
pajlada, line 149: Use a `SignalHolder`

src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+        case StreamerModeSetting::Enabled:
+            this->setEnabled(true);
+            this->timer_.stop();
+        case StreamerModeSetting::DetectStreamingSoftware: {
+            if (!this->timer_.isActive())
+            {
+                this->timer_.start(20s);
+                this->check();
+            }
+        }
```
pajlada, line 241: Duplicate the code instead of falling through like this

src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+    StreamerMode *parent_;
+    SingleConnection settingConnection_;
+
+    QTimer timer_;
+    QThread thread_;
+
+    std::atomic_bool enabled_ = false;
+    mutable std::atomic_uint8_t timeouts_ = 0;
+    StreamerModeSetting prevValue_ = StreamerModeSetting::Disabled;
```
pajlada, line 156: called it Setting instead of Value
Calling it `prev` also implies that this will contain the setting as-it-were prior to the current value, which I don't think is what it stores

src/singletons/StreamerMode.cpp
```diff
@@ -0,0 +1,253 @@
...
+    void check();
+
+    StreamerMode *parent_;
+    SingleConnection settingConnection_;
+
+    QTimer timer_;
+    QThread thread_;
+
+    std::atomic_bool enabled_ = false;
+    mutable std::atomic_uint8_t timeouts_ = 0;
```
pajlada, line 155: Can these not be specializations of atomic? (i.e. std::atomic<bool> & std::atomic<uint8_t>)

pajlada (changes requested):

pajlada (approved):
Works as expected for me - thanks for the refactor!

PR #5223, opened 2024-03-02
fix: correctly load twitch badges in highlights page

pajlada:
Broken state:
image

Fixed state after this PR:
image


pajlada:
image

kornes:
in what way https://github.com/Chatterino/chatterino2/pull/3736 broke it? 
`imageCount() == 0` check was present before my PR, so unless it reports -1 now (which is error value we probably shouldn't ignore), it was broken before

pajlada:
> in what way #3736 broke it? `imageCount() == 0` check was present before my PR, so unless it reports -1 now (which is error value we probably shouldn't ignore), it was broken before

My bad, I checked the diff too lazily then

PR #5224, opened 2024-03-02
fix: the font switcher now remembers your previous font

### LINKED ISSUE #2112: Font options aren't saved
**Describe the bug**
When trying to re-adjust your font, any custom option that may be set reverts back to another font choice

**To reproduce**
1. Choose custom font
2. Click ok and ok again
3. Go back into the custom font settings
4. Notice that it's changed

**Screenshots**
lol fonts broken

**Chatterino version**
Chatterino Nightly 2.2.3-beta (commit 5f54b1b49, 23.10.2020)

**Operating system**
Windows 10

pajlada:
Fixes #2112

I would appreciate someone other than me testing this out to make sure I've covered everything

You should be able to:
 - select one of the predefined fonts and the chat fonts should update
 - select "choose" which opens a font dialog & select a custom font
 - select "choose" which opens a font dialog & select cancel (which should revert back to your previously chosen font)

### DIFF
--- a/src/widgets/settingspages/GeneralPageView.hpp
+++ b/src/widgets/settingspages/GeneralPageView.hpp
@@ -197,29 +197,26 @@ public:
             },
             this->managedConnections_);
 
+        auto updateSetting = [combo, &setting, setValue = std::move(setValue)](
+                                 const int newIndex) {
+            setting = setValue(DropdownArgs{
+                .value = combo->itemText(newIndex),
+                .index = combo->currentIndex(),
+                .combobox = combo,
+            });
+            getIApp()->getWindows()->forceLayoutChannelViews();
+        };
+
         if (listenToActivated)
         {
-            QObject::connect(
-                combo, &QComboBox::activated,
-                [combo, &setting,
-                 setValue = std::move(setValue)](const int index) {
-                    setting = setValue(DropdownArgs{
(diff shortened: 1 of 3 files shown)


src/widgets/settingspages/GeneralPageView.hpp
```diff
@@ -197,14 +197,30 @@ class GeneralPageView : public QWidget
...
+            QObject::connect(
+                combo,
+                QOverload<const int>::of(&QComboBox::currentIndexChanged),
+                [combo, &setting,
+                 setValue = std::move(setValue)](const int newIndex) {
+                    setting =
+                        setValue(DropdownArgs{combo->itemText(newIndex),
+                                              combo->currentIndex(), combo});
+                    getIApp()->getWindows()->forceLayoutChannelViews();
+                });
```
Nerixyz, line 222: Aren't these the same callbacks? I think it would be easier to read if the callback was only declared once, like `updateValue` or `applyValue`.

pajlada: Fixed in [`8787e31` (#5224)](https://github.com/Chatterino/chatterino2/pull/5224/commits/8787e3161fc31ee8a7c8df6ad60e66eac499647f)

pajlada:
j
image

PR #5229, opened 2024-03-03
fix: settings "Cancel" button doesn't work first time

pajlada:
Mostly unrelated changes in this, the actual change is just ensuring saveSnapshot is *always* called when the settings dialog is opened
Benchmarks on my machine are 0.02ms for saving, 2ms for restoring


pajlada (commented):
image

PR #5230, opened 2024-03-03
fix(windows): show split tooltip before move

### LINKED ISSUE #5189: Split header tooltips sometimes show up in the wrong position
### Checklist

### Describe your issue

Hovering over the split header, the tooltip appears, sometimes off centered or in the wrong position.
This only happens on the first hover of the channel split header (if you move your mouse off and back on, the tooltip is rendered correctly)
With a bigger thumbnail size or longer stream title, it's better to see, but it's not necessary.
I tested this on ffa63f0, but I think it is since fec8f8c. On eb8c7f2 and 3ad2e4b, i am not having his bug.

Steps to reproduce:
- put c2 to right monitor border
- make c2 window width like normal twitch chat or even smaller
- click a tab with a channel and hover over the channel split header
- repeat last step until the tooltip moves out of the main window or looks off centered (different channel each time)

### Screenshots

image

### OS and Chatterino Version

Chatterino Nightly 2.4.6 (commit ffa63f0) Windows 10

Nerixyz:
I believe https://github.com/Chatterino/chatterino2/issues/5189 is a Windows only bug. This PR fixes tooltips of splits appearing in the wrong position [after the first hover(?)] on Windows. The issue might also occur when a streamer goes offline and the thumbnail isn't shown, but I didn't test that.
While debugging, I noticed that even though we call `adjustSize` which, according to the documentation, adjusts the area its children occupy, the tooltip still has a larger size than expected. It's only adjusted when calling `show()` <sub>(if you want to reproduce this, try placing logpoints in `QLayoutPrivate::doResize`, `QWidget::setGeometry`, and `QWidgetPrivate::setGeometry_sys`)</sub>. 

For non-Windows platforms, nothing changes.
On Windows, the tooltip is now shown before it's moved. This is fine on Windows, as the new window doesn't become visible right after calling `show()` (only after processing some more events). Since a lot of the resizing/activating happens in `show()`, the window has the correct size afterward, and `moveTo`-ing it will place it at the desired position.
I don't know why we don't need to do this for the tooltips in `ChannelView`.

Fixes #5189. @8thony could you test the change? Make sure you place the window on some side of your screen.

I have a fix for the bug I mentioned in https://github.com/Chatterino/chatterino2/issues/5189#issuecomment-1953158885 (it's one line), but I suppose it's better to do that separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@
 - Bugfix: Fixed link info not updating without moving the cursor. (#5178)
 - Bugfix: Fixed an upload sometimes failing when copying an image from a browser if it contained extra properties. (#5156)
 - Bugfix: Fixed tooltips getting out of bounds when loading images. (#5186)
+- Bugfix: Fixed split header tooltips showing in the wrong position on Windows. (#5230)
 - Dev: Run miniaudio in a separate thread, and simplify it to not manage the device ourselves. There's a chance the simplification is a bad idea. (#4978)
 - Dev: Change clang-format from v14 to v16. (#4929)
 - Dev: Fixed UTF16 encoding of `modes` file for the installer. (#4791)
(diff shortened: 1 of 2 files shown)


8thony (approved):
Seems to work the same as on 2.4.6, thanks

PR #5231, opened 2024-03-03
Add plugin permissions and IO API

### LINKED ISSUE #4620: Add a plugin permission system
Originally discussed in https://github.com/Chatterino/chatterino2/discussions/4352

### Checklist

### Describe your issue

Only some plugins should be able to more sensitive things like read/write files. We need a permission system even if sensitive functionality isn't implemented yet.

My proposal for such a system would look something like this:
 - A new plugin would *statically* declare a bunch of permissions which contain some filters on what it can do. Chatterino would convert them to user readable strings like: `Read and write files in plugin directory`, `Start "firefox" with arbitrary arguments` , et cetera.  The user would decide if wish to enable the plugin.
 - If the plugin permissions were to change when Chatterino isn't running, Chatterino should warn the user about new or changed permissions.

### Screenshots

N/A

### OS and Chatterino Version

Chatterino 2.4.3 DEBUG (commit 1f0aa808 modified) built with Qt 5.15.9 Running on Arch Linux, kernel: 6.3.1-arch1-1

Mm2PL:
This PR adds:
 - Wrappers for Lua's native `io` library. If the proper permissions are set, they are mostly transparent to the user and behave like Lua's functions. 
 - Support for declaring permissions in `info.json`
 - Two permission types: `FilesystemRead` and `FilesystemWrite`
Open questions:
 - ~~How should we go about globs in permissions?~~ We don't, per-path globs aren't a thing.
 - ~~Should we ensure that plugins don't overwrite their code?~~ Restrict everything to its data directory.
 - ~~How should we ensure that plugins suddenly can't overwrite their `info.json` to add permissions?~~ Same as above.

Closes #4620


iProdigy:
could we solve some of the open questions by restricting the path to something like `./plugins/plugin_name/` instead of allowing arbitrary paths to be defined in the json? (i'm apprehensive to allow more general file access unless there's a compelling use case)

Mm2PL:
> Could we solve some of the open questions by restricting the path to something like `./plugins/plugin_name/` instead of allowing arbitrary paths to be defined in the json? (i'm apprehensive to allow more general file access unless there's a compelling use case)

I took the arbitrary paths as it might be useful for IPC (think FIFO), those usually are placed in temporary folders. I decided to not be overly restrictive with this. I do agree that it is a little scary. If we restricted it to plugin directory only, we could do away with per-path permissions entirely and just have an *access filesystem* permission. That still would not relieve us of problems 2 and 3: overwriting code and `info.json`.

iProdigy:
to clarify, i mean a separate sub-directory that does not contain any code or the info.json file

Mm2PL:
That would solve basically all permission issues. Only problems I could imagine is malicious plugins filling up the user's drive and that should not be a major issue. However I am still not sure about restricting the `io` module more than absolutely necessary.

src/controllers/plugins/PluginPermission.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+    this->type = opt.value();
+}
+
+QString PluginPermission::toHtmlEscaped() const
+{
+    switch (this->type)
+    {
+        case PluginPermission::Type::FilesystemRead:
+            return "In its data directory.";
```
Nerixyz, line 36: This reads as `Required permissions: In its data directory`. Is this intended?

src/controllers/plugins/PluginPermission.cpp
```diff
@@ -0,0 +1,45 @@
...
+        case PluginPermission::Type::FilesystemWrite:
+            return "Write to or create files in its data directory";
+        default:
+            assert(false && "invalid PluginPermission type in toString()");
```
Nerixyz, line 40: This method is not called `toString()`.

src/controllers/plugins/PluginPermission.hpp
```diff
@@ -0,0 +1,27 @@
...
+    };
+    Type type;
+
+    std::vector<QString> errors;
+
+    bool isValid() const
+    {
+        return this->errors.empty();
+    }
+    QString toHtmlEscaped() const;
```
Nerixyz, line 21: Could this be `toHtml` instead? Some permissions might want to render information in HTML (like `"do something in <b>[html-escaped]</b>"`). There should probably be a newline above, too.

src/controllers/plugins/PluginPermission.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+    explicit PluginPermission(const QJsonObject &obj);
```
Nerixyz, line 23: This constructor should probably be at the top.

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
+#ifdef CHATTERINO_HAVE_PLUGINS
+#    include "controllers/plugins/api/IOWrapper.hpp"
+
+#    include "Application.hpp"
+#    include "controllers/plugins/LuaUtilities.hpp"
+#    include "controllers/plugins/PluginController.hpp"
+#    include "lauxlib.h"
+#    include "lua.h"
+namespace chatterino::lua::api {
```
Nerixyz, line 9: Nit: newline above

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
...
+    char major = 'r';  // 'r'|'w'|'a'
+    bool update{};     // '+'
+    bool binary{};     // 'b'
+    QString error;
+
+    LuaFileMode() = default;
+
+    LuaFileMode(const QString &smode)
+    {
+        auto major = smode.at(0);
```
Nerixyz, line 24: This is UB if `smode` is empty.

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
...
+    auto abs = file.absoluteFilePath();
+    qCDebug(chatterinoLua) << "[" << pl->id << ":" << pl->meta.name
+                           << "] Plugin is opening file at " << abs
+                           << " with mode " << mode.toString();
+    bool ok = pl->hasFSPermissionFor(
+        mode.update || mode.major == 'w' || mode.major == 'a', abs);
+    if (!ok)
+    {
+        return luaL_error(
+            L, "Plugin does not have permissions to access given file.");
```
Nerixyz, line 125: These shouldn't be runtime errors but `fail` and an error message (see [docs](https://www.lua.org/manual/5.4/manual.html#6.8)).

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
...
+    lua_remove(L, -2);  // remove LUA_REGISTRYINDEX[REAL_IO_NAME]
+    lua_insert(L, 1);   // move function to start of stack
+    lua::push(L, abs);
+    lua_insert(L, 2);  // move file name just after the function
+    lua_call(L, lua_gettop(L) - 1, LUA_MULTRET);
+    return lua_gettop(L);
+}
+
+namespace {
+    // This is the code for both io.input and io.output
```
Nerixyz, line 193: Nit: newline above

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
...
+    // (input).read
+    // args
+    lua_call(L, lua_gettop(L) - 1, 1);
+    return 1;
+}
+
+// NOLINTEND(*vararg)
+
+}  // namespace chatterino::lua::api
+#endif
```
Nerixyz, line 342: Nit: newline above

src/controllers/plugins/Plugin.cpp
```diff
@@ -201,5 +241,22 @@ void Plugin::removeTimeout(QTimer *timer)
     }
 }
 
+bool Plugin::hasFSPermissionFor(bool write, const QString &path)
+{
+    auto canon = QUrl(this->loadDirectory().absolutePath() + "/");
+    if (!canon.isParentOf(path))
```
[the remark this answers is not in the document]
Mm2PL: That is not intended and a major issue lol

src/controllers/plugins/api/IOWrapper.cpp
```diff
@@ -0,0 +1,342 @@
...
+            return luaL_error(L, mode.error.toStdString().c_str());
+        }
+    }
+    QString filename;
+    if (!lua::pop(L, &filename))
+    {
+        return luaL_error(L,
+                          "io.open filename (1st argument) must be a string");
+    }
+    QFileInfo file(pl->loadDirectory().filePath(filename));
```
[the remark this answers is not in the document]
Mm2PL: Corrected.

src/controllers/plugins/PluginPermission.cpp
Mm2PL: I think I wrote this at 3 in the morning. I'm happy to reword it.

src/controllers/plugins/PluginPermission.cpp
Mm2PL: I'll be renaming it to `toHtml`.

src/controllers/plugins/Plugin.cpp
```diff
@@ -201,5 +241,22 @@ void Plugin::removeTimeout(QTimer *timer)
...
+    {
+        return false;
+    }
+
+    using PType = PluginPermission::Type;
+    auto typ = write ? PType::FilesystemWrite : PType::FilesystemRead;
+
+    return std::ranges::any_of(this->meta.permissions,
```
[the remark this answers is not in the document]
Mm2PL: The docker builds don't return any useful errors, so I can't tell if its the CI's fault or mine.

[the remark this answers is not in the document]
Mm2PL: Fixed

docs/wip-plugins.md
```diff
@@ -35,12 +37,42 @@ Example file:
...
+
+Allows the plugin to read from its data directory.
+
+Example:
+
+```json
+{
+  "type": "FilesystemRead"
+}
+```
```
pajlada, line 62: maybe something like this for the examples to clarify
```json
{
  ...,
  "permissions": [
    {
      "type": "FilesystemRead"
    }
  ]
}
```

src/controllers/plugins/Plugin.cpp
```diff
@@ -111,6 +112,45 @@ PluginMeta::PluginMeta(const QJsonObject &obj)
...
+            for (const auto &err : parsed.errors)
+            {
+                this->errors.push_back(
+                    QString("permissions element #%1: %2").arg(i).arg(err));
+            }
+            if (parsed.isValid())
+            {
+                // ensure no invalid permissions slip through this
+                this->permissions.push_back(parsed);
+            }
```
pajlada, line 150: Consider iterating over errors only if it's invalid

src/controllers/plugins/api/IOWrapper.hpp
```diff
@@ -0,0 +1,97 @@
+#ifdef CHATTERINO_HAVE_PLUGINS
```
pajlada, line 1: Missing pragma once

pajlada (approved):

iProdigy (approved):
thanks for bearing with my security nits

Issue #4960, opened 2023-11-11
Double/triple click selection of long words does not select all characters


jupjohn:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When double clicking on a long word in a message, the selection does not span the entire word. This is also exhibited by triple clicking the message.

Example message:
```
@user-type=;color=#FF0000;emotes=;turbo=0;returning-chatter=0;client-nonce=2513080064f6a1335c3c4b460849e7ac;badges=glhf-pledge/1;user-id=130496419;id=e96f7752-d2d6-45c1-b23f-dd7a6173d535;rm-received-ts=1699732253387;flags=;tmi-sent-ts=1699732253150;first-msg=0;room-id=46673989;subscriber=0;mod=0;display-name=jendrina22;badge-info= :jendrina22!jendrina22@jendrina22.tmi.twitch.tv PRIVMSG #masayoshi widepepohappycliminghardo4housewithcomodohypeontopholdingapepegadragonpoganonactionfigure
```

### Screenshots

Double click selection:

![image](https://github.com/Chatterino/chatterino2/assets/4962764/463720c0-f33d-47de-a5bc-68d7db4f8f01)

Expected text selection: `widepepohappycliminghardo4housewithcomodohypeontopholdingapepegadragonpoganonactionfigure`
Actual text selection: `widepepohappycliminghardo4housewithcomodohypeontopholdingapepega`

The following videos demonstrate double clicking then triple clicking behaviour.

https://github-production-user-asset-6210df.s3.amazonaws.com/4962764/282263471-071856fe-1130-4830-afb8-d00ac0a31601.mp4

https://github-production-user-asset-6210df.s3.amazonaws.com/4962764/282263462-7317551e-2b3e-4137-b585-6c2454463409.mp4

### OS and Chatterino Version

Chatterino 2.4.6 (commit 95620e6e1) built with Qt 6.6.0 Running on Arch Linux, kernel: 6.6.1-zen1-1-zen

kornes:
double-click selection ends selection on the line's end rather than word end, i tested c2 2.4.4 and it behaves the same so likely it never worked as expected.

[...]

PR #5243, opened 2024-03-10
Fix double click to select full words

KleberPF:
Leaving this as a draft because I want some opinions if this is a good approach or not. Heavily inspired by [this](https://github.com/Chatterino/chatterino2/issues/4960#issuecomment-1981668329) comment from @Nerixyz, but adding an id to each element instead of the prev/next approach. Currently works on basic tests, haven't tested more advanced stuff (RTL text, CJK characters, etc)


Nerixyz (commented):
Using IDs for words is a good idea. However, I think the concept shouldn't "leak" beyond `MessageLayout`, i.e. `ChannelView` and `*MessageBuilder` shouldn't know about it.

One bug this implementation has is that it assumes that one `TextElement` has at most one word. This isn't the case. The `EmotePopup` for example, adds an element with multiple words:

https://github.com/Chatterino/chatterino2/blob/b9b1b8cf9cbb33618e445a100f0a9ef6e11a3a9b/src/widgets/dialogs/EmotePopup.cpp#L53-L55

`MessageLayoutContainer` should generate the IDs. When a `TextElement` adds its elements to the container, it generates a new ID for each word by calling something like `container.nextWordID()`. This way, message builders don't need to know about word-IDs. (if you put the next ID after `containsRTL`, it doesn't consume extra space)

Getting the word bounds should hide the concept of word-IDs too. It should have a single method like `layout.getWordBoundsAt(QPoint)` which finds the layout element and finds the bounds of all layout elements with that word-ID (or returns the element's bounds if it has an ID of -1).

KleberPF:
Thanks for the feedback. I think I implemented everything you pointed out, and it seems to be working well. There might be some off by one errors because the messages are kinda scuffed on KDE Plasma

image

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -456,6 +456,40 @@ size_t MessageLayoutContainer::getFirstMessageCharacterIndex() const
...
+    {
+        const auto &element = this->elements_[index];
+        if (element->getWordId() != wordId)
+        {
+            break;
+        }
+
+        wordEnd += element->hasTrailingSpace()
+                       ? element->getSelectionIndexCount() - 1
+                       : element->getSelectionIndexCount();
```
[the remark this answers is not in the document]
KleberPF: Now this is done after the loop checking the last element

src/widgets/helper/ChannelView.cpp
```diff
@@ -1817,7 +1817,7 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
     if (this->isDoubleClick_ && hoverLayoutElement)
     {
         auto [wordStart, wordEnd] =
-            getWordBounds(layout.get(), hoverLayoutElement, relativePos);
+            layout->getWordBounds(hoverLayoutElement->getWordId());
```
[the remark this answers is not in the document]
KleberPF: Now we choose which logic to use inside `layout->getWordBounds`

KleberPF:
I tested this a bit more and everything seems to be working. Not sure if there is any point in documenting the tests as they would be just screenshots of random selections but let me know if I should. I'll remove the draft as it seems the final version will be something close to this.

src/messages/layouts/MessageLayout.cpp
```diff
@@ -443,12 +443,31 @@ void MessageLayout::deleteCache()
...
+const MessageLayoutElement *MessageLayout::getElementAt(QPoint point) const
 {
     // go through all words and return the first one that contains the point.
     return this->container_.getElementAt(point);
 }
 
+std::pair<int, int> MessageLayout::getWordBounds(QPoint relativePos) const
+{
+    const auto *element = this->getElementAt(relativePos);
+    assert(element != nullptr);
```
pajlada, line 455: getElementAt can explicitly return a nullptr if no element is matched at the position
I see the assert might be ok since we call getElementAt right before this, but since we're just now making `getWordBounds`, couldn't it accept a `MessageLayoutElement` already so the `getElementAt` call isn't done twice?

src/messages/layouts/MessageLayout.cpp
```diff
@@ -443,12 +443,31 @@ void MessageLayout::deleteCache()
...
+
+    if (element->getWordId() != -1)
+    {
+        return this->container_.getWordBounds(element);
+    }
```
pajlada, line 460: Could you add a comment here for what this significes? It's not clear to me what it means when the element we're looking at has a word id of -1

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -456,6 +457,45 @@ size_t MessageLayoutContainer::getFirstMessageCharacterIndex() const
...
+        const auto &element = this->elements_[index];
+        if (element->getWordId() != hoveredElement->getWordId())
+        {
+            break;
+        }
+
+        wordEnd += element->getSelectionIndexCount();
+    }
+
+    const auto *lastElementInSelection = this->elements_[index - 1].get();
```
pajlada, line 490: Not sure how feasible it is, but could you add a unit test that runs getWordBounds on a MessageLayoutContainer with some edge cases? like 0 elements, 1 element, hoveredElement doesn't exist

Specifically the `index - 1` part to me is scary

pajlada (commented):
I pushed some stuff, mostly small things. Could you review the comments I added to make sure I understood the code correctly?

src/messages/layouts/MessageLayout.cpp
KleberPF: I did it that way first, but we need `relativePos` later when calling `this->getSelectionIndex(relativePos)`. Don´t know if there is a better way to do this here (unless you want to pass in both parameters)

src/messages/layouts/MessageLayout.cpp
KleberPF: -1 basically means that the element can't span across multiple lines (like an emote, for example). I agree the code is not very self explanatory here. Do you want a comment here or should I encapsulate that inside a function like `element->canBeMultiline()` (name is kinda bad but you get the idea)

src/messages/layouts/MessageLayoutContainer.cpp
KleberPF: Don't know if there is a way to create an empty `MessageLayoutContainer`, but I like having a failsafe here. I will add it and see what I can do about the tests

src/messages/layouts/MessageLayout.cpp
pajlada: I think a comment here is fine

tests/src/MessageLayout.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include "widgets/Notebook.hpp"
+#include "widgets/splits/Split.hpp"
+#include "widgets/splits/SplitInput.hpp"
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include <QDebug>
+#include <QString>
+
+#include <memory>
```
KleberPF, line 26: Commenting to remember to remove unused headers

tests/src/MessageLayout.cpp
```diff
@@ -0,0 +1,123 @@
...
+    MessageLayoutTest(const QString &text)
+    {
+        MessageBuilder builder;
+        builder.append(
+            std::make_unique<TextElement>(text, MessageElementFlag::Text));
+        this->layout = std::make_unique<MessageLayout>(builder.release());
+        this->layout->layout(WIDTH, 1, MessageElementFlag::Text, false);
+    }
+
+    MockApplication mockApplication;
```
KleberPF, line 102: I still don't quite understand how constructing `MockApplication` here makes it so any all to `getIApp` gets the MockApplication. Also this should probably only be done once instead of each time a test is created

tests/src/MessageLayout.cpp
```diff
@@ -0,0 +1,123 @@
...
+};
+
+}  // namespace
+
+TEST(TextElement, BasicCase)
+{
+    auto test = MessageLayoutTest("aaaaaaaa");
+
+    // Simulate we are clicking on the first word
+    auto point = QPoint(WIDTH / 20, test.layout->getHeight() / 2);
```
KleberPF, line 113: This is ugly

src/messages/layouts/MessageLayoutContainer.cpp
KleberPF: I battled my way through gtest and somehow managed to make a working test. I left some comments which I think you should take a look at. I'll add more tests and fix some of the stuff pointed out tomorrow

src/messages/MessageElement.cpp
```diff
@@ -454,7 +454,7 @@ TextElement::TextElement(const QString &text, MessageElementFlags flags,
 void TextElement::addToContainer(MessageLayoutContainer &container,
                                  MessageElementFlags flags)
 {
-    auto *app = getApp();
+    auto *app = getIApp();
```
KleberPF, line 457: Is this ok?

src/messages/layouts/MessageLayout.cpp
KleberPF: Done

src/messages/layouts/MessageLayout.cpp
KleberPF: Added

src/messages/MessageElement.cpp
pajlada: Yes :+1:

tests/src/MessageLayout.cpp
pajlada: The constructor of MockApplication sets a singleton to point to itself (check mocks/EmptyApplication)

tests/src/MessageLayout.cpp
pajlada: That's fine :+1: This test has already helped a lot

tests/src/MessageLayout.cpp
pajlada: I've gone ahead and done this in [`269e77d` (#5243)](https://github.com/Chatterino/chatterino2/pull/5243/commits/269e77d304edfbe444fc9fa73d01c2718330facd)

pajlada:
Testing RTL selections - it's very scuffed. Luckily, it's already scuffed before this PR, which means I don't think it's a blocker.

pajlada (approved):

PR #5245, opened 2024-03-10
Hide chatter list button for non-moderators

### LINKED ISSUE #1239: Mod/VIP status does not update when modded/VIPed
**Describe the bug**
When you join a channel, chatterino2 currently determines whether you are mod/VIP. The mod/VIP status however does not update when you are modded or unmodded. (Which is easily detectable via the received `PRIVMSG`, `USERSTATE` or even PubSub messages for the own user).

**To Reproduce**
- Join channel where you are neither mod or VIP
- Have the broadcaster mod/VIP you
- You can not spam without the global delay (if VIP or mod) and mod button does not appear (if modded)

**Expected behavior**
VIP/Mod status updates.

**Screenshots**

**Chatterino version**
3079765a0cb8d81721b1812901b55c38d3624c41

**Additional context**

pajlada:
this also makes sure moderator status is updated whenever possible (so when you type a message).

Fixes #1239


pajlada (commented):
image

Issue #5218, opened 2024-03-01
Add support for /announce[color] command variations


Mm2PL:
### Discussed in https://github.com/Chatterino/chatterino2/discussions/4367

<div type='discussions-op-text'>

<sup>Originally posted by **Pafon22** February 11, 2023</sup>
Unrecognized command when sending /announce with any color.
Insert /announceblue /announcegreen /announceorange /announcepurple in the options when possible. I don't know if Twitch already has the endpoint for this.</div>

However first colored announcement support needs to be implemented. I can see this functionality being merged into a single PR if neither part is too large.

pajlada:
When implementing the commands, they should replicate Twitch chat's behaviour
![image](https://github.com/Chatterino/chatterino2/assets/962989/7aec2c6a-9cfe-4b36-90ce-d5ef46c5b2af)

4rneee:
Supporting the `/announce[color]` commands would be very straight forward and I could create a PR for that.  
I also looked a bit into displaying colored announcements and don't know how (or if) this can be implemented because I didn't find any documentation. https://dev.twitch.tv/docs/irc/tags/#usernotice-tags doesn't even document the `announcement` notice in the `msg-id` tag which we currently use to detect announcements.

pajlada:
I'm ok with implementing sending, even if receiving colored announcements don't work

Mm2PL:
Here are some example announcement messages:

<details>
<summary>Examples</summary>

```
@badge-info=;badges=broadcaster/1,no_audio/1;color=#DAA520;display-name=Mm2PL;emotes=;flags=;id=98928e18-2c4c-4914-97ea-032a8b47d94c;login=mm2pl;mod=0;msg-id=announcement;msg-param-color=BLUE;room-id=117691339;subscriber=0;system-msg=;tmi-sent-ts=1710346144413;user-id=117691339;user-type=;vip=0 :tmi.twitch.tv USERNOTICE #mm2pl :blue
@badge-info=;badges=broadcaster/1,no_audio/1;color=#DAA520;display-name=Mm2PL;emotes=;flags=;id=c3e34a2e-b7be-4608-a8b2-18c6c455c302;login=mm2pl;mod=0;msg-id=announcement;msg-param-color=GREEN;room-id=117691339;subscriber=0;system-msg=;tmi-sent-ts=1710346157120;user-id=117691339;user-type=;vip=0 :tmi.twitch.tv USERNOTICE #mm2pl :green
[...]

PR #5250, opened 2024-03-13
feat: add /announce[color] commands

4rneee:
This adds support to send the `/announce[color]` commands.  
Resolves sending part for https://github.com/Chatterino/chatterino2/issues/5218.  
Note that Chatterino currently does not support receiving colored highlights, so the color is currently only visible in the browser chat.

Screenshot:
announce

### DIFF
--- a/src/controllers/commands/builtin/twitch/Announce.cpp
+++ b/src/controllers/commands/builtin/twitch/Announce.cpp
@@ -9,7 +9,8 @@
 #include "providers/twitch/TwitchAccount.hpp"
 #include "providers/twitch/TwitchChannel.hpp"
 
-namespace chatterino::commands {
+namespace {
+using namespace chatterino;
 
 QString sendAnnouncementColor(const CommandContext &ctx,
                               const HelixAnnouncementColor color)
@@ -91,6 +92,9 @@ QString sendAnnouncementColor(const CommandContext &ctx,
         });
     return "";
 }
+} // namespace
+
+namespace chatterino::commands {
 
 QString sendAnnouncement(const CommandContext &ctx)
 {
--- a/src/controllers/commands/builtin/twitch/Announce.cpp
+++ b/src/controllers/commands/builtin/twitch/Announce.cpp
@@ -26,11 +26,23 @@ QString sendAnnouncementColor(const CommandContext &ctx,
         return "";
     }
 
+    QString colorStr = "";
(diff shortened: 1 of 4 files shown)


Mm2PL (approved):
While I don't necessarily like having a ton of commands one for each color, this is how Twitch does it and commands are expected to work the mostly same on every client. 
Looks good to me.
I edited the PR description, so the issue wouldn't be closed when this is merged.

src/controllers/commands/builtin/twitch/Announce.cpp
```diff
@@ -25,25 +26,38 @@ QString sendAnnouncement(const CommandContext &ctx)
...
+
     if (ctx.words.size() < 2)
     {
         ctx.channel->addMessage(makeSystemMessage(
-            "Usage: /announce <message> - Call attention to your "
-            "message with a highlight."));
+            QString("Usage: /announce%1 <message> - Call attention to your "
+                    "message with a %1%2highlight.")
+                .arg(colorStr)
+                .arg(color == HelixAnnouncementColor::Primary ? "" : " ")));
```
pajlada, line 45: nit: do an if on the color. this makes the message formatting cleaner & moves the colorStr logic into the same color check, something along the lines of
```
if color == Primary:
  msg = "Usage: /announce <message> - Call attention to your message with a highlight"
else:
  colorstr = magic_enum(...)
  msg = QString("Usage: /announce%1 <message> - Call attention to your message with a %1 highlight", colorstr)

ctx.channel->addMessage(makeSystemMessage(msg));
```

src/controllers/commands/builtin/twitch/Announce.cpp
```diff
@@ -11,7 +11,8 @@
 
 namespace chatterino::commands {
 
-QString sendAnnouncement(const CommandContext &ctx)
+QString sendAnnouncementColor(const CommandContext &ctx,
```
pajlada, line 14: nit: move this function to an anon namespace, outsode if `namespace chatterino::commands`

so it'd be something along the lines of

```
namespace {
using namespace chatterino;
using namespace chatterino::commands; // maybe

QString sendAnnouncementColor...
```

pajlada (changes requested):
Some small nits, otherwise looks good to me

pajlada (approved):

PR #5252, opened 2024-03-16
Show line indicator instead of rectangle while dragging in tables

Nerixyz:
The user experience of dragging rows around in tables isn't the best, as you don't really have a great indicator of where a row will be dropped at. I found that you can proxy `PE_IndicatorItemViewItemDrop`, the drop indicator and draw your own stuff. To get the direction, a row is dragged in, the model indices are compared. If no target is found, a red border will appear around the viewport.

|| Before | After |
|---|---|---|
| Move | chatterino_2024-03-16_19-00-52 | chatterino_2024-03-16_18-56-39 |
| Invalid Move | chatterino_2024-03-16_19-02-01 | chatterino_2024-03-16_18-56-57 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - Minor: Add wrappers for Lua `io` library for experimental plugins feature. (#5231)
 - Minor: Add permissions to experimental plugins feature. (#5231)
 - Minor: Add support to send /announce[color] commands. (#5250)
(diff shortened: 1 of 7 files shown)


pajlada (commented):
Some small nits - haven't tested this yet but it looks good

pajlada (approved):

PR #5258, opened 2024-03-20
refactor: adapt magic_enum to Qt

Nerixyz:
Since `magic_enum` and `QString` don't really mix well, I adapted it to fit `QString` and `QStringView` in `qmagicenum`. Both `magic_enum` and `qmagicenum` can be used at the same time. In face, `qmagicenum` uses `magic_enum` under the hood. Most stuff is `constexpr` (except creation of `QString`s). Since this is Chatterino code, I used the Chatterino naming convention.

There are still a few `.toLower`s that should be removed in the future.

Only plugins need `magic_enum` (as Lua strings are 8bit).

Hopefully this works on Qt 5 and Ubuntu 20.04.

### DIFF
--- a/src/common/network/NetworkPrivate.hpp
+++ b/src/common/network/NetworkPrivate.hpp
@@ -60,7 +60,7 @@ public:
     void emitError(NetworkResult &&result);
     void emitFinally();
 
-    QStringView typeString() const;
+    QString typeString() const;
 
 private:
     QString hash_;
--- a/src/common/network/NetworkPrivate.hpp
+++ b/src/common/network/NetworkPrivate.hpp
@@ -60,7 +60,7 @@ public:
     void emitError(NetworkResult &&result);
     void emitFinally();
 
-    QLatin1String typeString() const;
+    QStringView typeString() const;
 
 private:
     QString hash_;
--- a/src/controllers/plugins/LuaAPI.cpp
+++ b/src/controllers/plugins/LuaAPI.cpp
@@ -119,9 +119,12 @@ int c2_register_callback(lua_State *L)
         return 0;
     }
 
-    auto callbackSavedName = QString("c2cb-%1").arg(
-        magic_enum::enum_name<EventType>(evtType).data());
-    lua_setfield(L, LUA_REGISTRYINDEX, callbackSavedName.toStdString().c_str());
+    auto typeName = magic_enum::enum_name(evtType);
+    std::string callbackSavedName;
+    callbackSavedName.reserve(5 + typeName.size());
+    callbackSavedName += "c2cb-";
+    callbackSavedName += typeName;
+    lua_setfield(L, LUA_REGISTRYINDEX, callbackSavedName.c_str());
 
     lua_pop(L, 2);
 
--- a/src/controllers/plugins/Plugin.hpp
+++ b/src/controllers/plugins/Plugin.hpp
@@ -107,14 +107,14 @@ public:
             return {};
         }
         // this uses magic enum to help automatic tooling find usages
+        auto typeName =
+            magic_enum::enum_name(lua::api::EventType::CompletionRequested);
+        std::string cbName;
+        cbName.reserve(5 + typeName.size());
+        cbName += "c2cb-";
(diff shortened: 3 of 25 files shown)


src/common/network/NetworkPrivate.hpp
```diff
@@ -60,7 +60,7 @@ class NetworkData
     void emitError(NetworkResult &&result);
     void emitFinally();
 
-    QLatin1String typeString() const;
+    QStringView typeString() const;
```
pajlada, line 63: Keep this as a string instead of a string view

src/controllers/plugins/LuaAPI.cpp
```diff
@@ -119,9 +119,12 @@ int c2_register_callback(lua_State *L)
...
 
-    auto callbackSavedName = QString("c2cb-%1").arg(
-        magic_enum::enum_name<EventType>(evtType).data());
-    lua_setfield(L, LUA_REGISTRYINDEX, callbackSavedName.toStdString().c_str());
+    auto typeName = magic_enum::enum_name(evtType);
+    std::string callbackSavedName;
+    callbackSavedName.reserve(5 + typeName.size());
+    callbackSavedName += "c2cb-";
+    callbackSavedName += typeName;
+    lua_setfield(L, LUA_REGISTRYINDEX, callbackSavedName.c_str());
```
pajlada, line 127: Would `std::format` work here? Not sure if all compilers support it

src/controllers/plugins/Plugin.hpp
```diff
@@ -107,14 +107,14 @@ class Plugin
...
+        cbName += typeName;
         auto typ =
-            lua_getfield(this->state_, LUA_REGISTRYINDEX,
-                         QString("c2cb-%1")
-                             .arg(magic_enum::enum_name<lua::api::EventType>(
-                                      lua::api::EventType::CompletionRequested)
-                                      .data())
-                             .toStdString()
-                             .c_str());
+            lua_getfield(this->state_, LUA_REGISTRYINDEX, cbName.c_str());
```
pajlada, line 117: Would `std::format` work here? Not sure if all compilers support it

src/widgets/settingspages/GeneralPageView.hpp
```diff
@@ -285,7 +286,8 @@ class GeneralPageView : public QWidget
 
         for (const auto &item : items)
         {
-            combo->addItem(QString::fromStdString(std::string(item)));
+            // Safety: item comes from qmagicenum (precondition)
+            combo->addItem(qmagicenum::staticString(item));
```
pajlada, line 290: Keep this simple & just add a regular QString instead

src/util/QMagicEnum.hpp
```diff
@@ -0,0 +1,309 @@
+#pragma once
+
+#include <magic_enum/magic_enum.hpp>
+#include <QString>
+#include <QStringView>
+
+namespace qmagicenum {
```
pajlada, line 7: Could you put this inside the chatterino namespace?

src/util/QMagicEnum.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+namespace detail {
```
pajlada, line 9: Could you put the detail namespace above the main namespace so the indentation doesn't get messed up?

pajlada (commented):
Just trying to keep the refactor as small as possible - overall looks great
Will take another more in-depth check in the weekend

src/common/network/NetworkPrivate.hpp
Nerixyz: In case of confusion, I'll note that `QLatin1String` is a string view too. It will be renamed to `QLatin1StringView` in Qt 7.

src/controllers/plugins/LuaAPI.cpp
Nerixyz: [cppreference](https://en.cppreference.com/w/cpp/compiler_support/20) says it requires GCC 13 (we're on 10).

pajlada (approved):

PR #5262, opened 2024-03-24
fix: Update Linux build instructions

pajlada:
- **fix: minimum qt version if 5.15**
- **fix: prioritize manual builds through terminal**
- **fix: use cmake --build . for building**
- **Update Arch Linux dependencies**
- **Update guide to use Qt6 (and fix Ubuntu instructions)**


pajlada:
image

PR #5263, opened 2024-03-24
fix: remove ":" from the message the user is replying to if it's a /me message

Heryin:
Screenshot_20240324_140000

This removes a `:` from the reply if it's a reply to a `/me` message to keep it consistent with `/me` messages not having a `:`.


pajlada (changes requested):
Add a new changelog "Bugfix" entry

pajlada (approved):

Issue #1077, opened 2019-05-18
OSX contextual program menus lacks details


coral:
![Screenshot 2019-05-18 at 15 43 10](https://user-images.githubusercontent.com/2163363/57975722-ace6e300-7983-11e9-826c-03bf648da2a1.png)

should contain a link to program settings and the other good stuff.

jammehcow:
Brainstorming here:

```
chatterino
  |-- About Chatterino
  |-- *separator*
  |-- Preferences
  |-- *default stuff added by macOS*

View
  |-- Zoom In
  |-- Zoom Out
  |-- Actual Size

Window:
  |-- Minimise
  |-- Zoom
  |-- *separator*
  |-- Bring all to front

Help
  |-- Chatterino Wiki
  |-- Chatterino GitHub
  |-- Chatterino Discord
```

I'm hesitant to add chatterino features like "New Split," "Close Tab" etc. for now unless someone has a good way to tie them to our keybinds in `Window.cpp` and be able to control context actions like "Cut," "Paste" where text fields don't exist.

pajlada:
I think this is completed since #5266

PR #5266, opened 2024-03-25
feat: add more items in macOS menu bar

mavjs:
This PR tries to address some of the items listed in the brainstormed list mentioned in #1077

- "About Chatterino" item which when clicked, opens the SettingsDialog with the "About" tab selected.
image

- "Help" menu bar option that has multiple items linking to Wiki, Github, Discord, when clicked they open the URLs with the default browser.
image

- "Minimise" item (under "Window" menu bar option) which when clicked, minimises the entire Window.
image


src/widgets/Window.cpp
```diff
@@ -711,6 +719,13 @@ void Window::addMenuBar()
     // Window menu.
     QMenu *windowMenu = mainMenu->addMenu(QString("Window"));
 
+    // Window->Minimise item
+    QAction *minimiseWindow = windowMenu->addAction(QString("Minimise"));
+    minimiseWindow->setShortcuts({QKeySequence("Meta+M")});
```
Mm2PL, line 724: Is this standard behavior to have a `Window->Minimise` action on macos? This is probably a window manager's job to handle minimising a window and have a shortcut for it.

src/widgets/Window.cpp
```diff
@@ -722,6 +737,27 @@ void Window::addMenuBar()
...
+    // Help->Chatterino Github
+    QAction *helpGithub = helpMenu->addAction(QString("Chatterino Github"));
+    connect(helpGithub, &QAction::triggered, this, []() {
+        QDesktopServices::openUrl(QUrl("https://github.com/Chatterino/chatterino2"));
+    });
+
+    // Help->Chatterino Discord
+    QAction *helpDiscord = helpMenu->addAction(QString("Chatterino Discord"));
+    connect(helpDiscord, &QAction::triggered, this, []() {
+        QDesktopServices::openUrl(QUrl("https://discord.gg/7Y5AYhAK4z"));
```
Mm2PL, line 759: This should use the same constant as the link that's currently in the AboutPage (`LINK_CHATTERINO_DISCORD`). Should this expire or need to be deleted, changing it shouldn't be unnecessarily hard.

Mm2PL (changes requested):

src/widgets/Window.cpp
mavjs: A lot of application on macOS do have minimise action. (See examples below)

I have not been using macOS very long, and also trying to wrap my head around this code, so I am going off from what I can see from the different applications I am using. I could very well be wrong here.

Here is one from Safari browser:
image

Here is one from VSCode:
image

src/widgets/Window.cpp
mavjs: Will do!

Wissididom:
Not sure, but this is what I find when googling for the question if a minimise option is standard for mac: https://developer.apple.com/design/human-interface-guidelines/the-menu-bar#Window-menu

As far as I understand it is recommended, but not necessary to do, but I have no experience with Mac at all.

src/widgets/Window.cpp
mavjs: @Mm2PL what would, in your opinion, be the best option to have that same constant (`LINK_CHATTERINO_DISCORD`) from `AboutPage.cpp` in `Window.cpp` as well?

I am still wrapping my head around the code base and haven't written C++ in a long time so I wanted to have some suggestions before I make any changes that may go beyond just the menu bar items.

Here are a couple of options:
* Create a header file in `src/common/ProjectInfo.hpp` which holds those similar constants that other things can include? (likely option and probably good for future.)
* Include `AboutPage.cpp` and use that in `Window.cpp` (probably the least preferred way?)
* Create the same constant in `Window.cpp` (duplication)

pajlada: You could move the `LINK_CHATTERINO_DISCORD` definition to `src/common/Common.hpp` & use that in both files

Felanbird:
should be `minimize` for both the fact that Apple documents it spelt that way, and we have no references in our code spelt with an s

src/widgets/Window.cpp
```diff
@@ -722,6 +738,27 @@ void Window::addMenuBar()
...
+    QMenu *helpMenu = mainMenu->addMenu(QString("Help"));
+
+    // Help->Chatterino Wiki item
+    QAction *helpWiki = helpMenu->addAction(QString("Chatterino Wiki"));
+    connect(helpWiki, &QAction::triggered, this, []() {
+        QDesktopServices::openUrl(QUrl(LINK_CHATTERINO_WIKI));
+    });
+
+    // Help->Chatterino Github
+    QAction *helpGithub = helpMenu->addAction(QString("Chatterino Github"));
```
pajlada, line 752: nit: capitalization
```suggestion
    QAction *helpGithub = helpMenu->addAction(QString("Chatterino GitHub"));
```

pajlada (approved):
Overall this looks great! Thank you! I'll test this briefly on my macOS Sonoma system then I'm happy to merge this in

mavjs:
> Overall this looks great! Thank you! I'll test this briefly on my macOS Sonoma system then I'm happy to merge this in

Thank you!

One thing that's low priority (I think) is about the use of `Meta` in the MenuBar, while in fact it should be `Control`. However, for this PR, I didn't want to change such keyboard shortcuts until we get some consensus about whether that change is desirable. :)

Referencing: https://doc.qt.io/qt-6/qkeysequence.html#details

> Note: On Apple platforms, references to "Ctrl", [Qt::CTRL](https://doc.qt.io/qt-6/qt.html#Modifier-enum), [Qt::Key_Control](https://doc.qt.io/qt-6/qt.html#Key-enum) and [Qt::ControlModifier](https://doc.qt.io/qt-6/qt.html#KeyboardModifier-enum) correspond to the Command keys on the Macintosh keyboard, and references to "Meta", [Qt::META](https://doc.qt.io/qt-6/qt.html#Modifier-enum), [Qt::Key_Meta](https://doc.qt.io/qt-6/qt.html#Key-enum) and [Qt::MetaModifier](https://doc.qt.io/qt-6/qt.html#KeyboardModifier-enum) correspond to the Control keys. In effect, developers can use the same shortcut descriptions across all platforms, and their applications will automatically work as expected on Apple platforms.

Mm2PL (approved):

pajlada:
I've tested this and everything works as expected :+1: 
I don't have edit rights on the PR, could you make this PR up-to-date with master @mavjs?

mavjs:
> I've tested this and everything works as expected 👍 I don't have edit rights on the PR, could you make this PR up-to-date with master @mavjs?

@pajlada oops, I toggled that off when I sent a PR. (Have re-enabled it.) Been a while since I have contributed to some upstream open source stuff, so played around with that option. 😛 

Seems @Felanbird (thanks!) has taken care of making it up-to-date with master. 🙏🏽

PR #5273, opened 2024-03-27
ci: run clang-tidy with Qt 6 and update action

Nerixyz:
Since https://github.com/Chatterino/chatterino2/blob/master/BUILDING_ON_LINUX.md now states that Chatterino should be built with Qt 6, it's time to use it in the clang-tidy CI.

- Removed unused/misleading steps
- Renamed jobs (should be a bit easier to understand than "build")
- Updated action to v0.18
	- Disabled `num_comments_as_exitcode`
- Qt is now installed in the workspace (`.qtinstall`)
- Added script to set up CMake in CI
	This will set up the environment variables like https://github.com/jurplel/install-qt-action, configure CMake, and build MOC and UIC targets (as a consequence it also builds dependencies).


.CI/SetupClangTidyCMake.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
pajlada, line 1: Rename to `setup-clang-tidy.sh`
Remove semicolons at the end of the lines

pajlada (changes requested):
small nit, overall looks good :+1:

pajlada (approved):

PR #5275, opened 2024-03-29
Add `reward.cost` `reward.id`, `reward.title` filter variables

Mm2PL:
This grows the Message struct by 16 bytes for the `shared_ptr`.

This adds the following filter variables:
- `reward.title`
- `reward.id`
- `reward.cost`

The use cases include having a separate split for high cost rewards that you don't want to miss, showing only some rewards in chat.

(felanbird): #4882 #4883


Felanbird:
title works
cost works but probably needs to be an int
id works

pajlada:
I have confirmed that all of these filters work as expected :+1: Thank you!

PR #5283, opened 2024-03-30
refactor: improve LuaLS generator

Nerixyz:
The previous generator had some bugs that would require keeping a lot more state or having more duplicate code when doing it line-by-line. Mainly, classes with methods weren't defined globally and some enums weren't read correctly.
I refactored it to fix the bugs. It should be a bit easier to extend the generator. ~~Furthermore, triple-slash comments are now supported.~~ Triple slash comments aren't supported anymore, they're doxygen-only now.
There were some issues in the definitions which I fixed as well.

### DIFF
--- a/docs/plugin-meta.lua
+++ b/docs/plugin-meta.lua
@@ -6,15 +6,6 @@
 
 c2 = {}
 
----@class IWeakResource
-IWeakResource = {}
-
---- Returns true if the channel this object points to is valid.
---- If the object expired, returns false
---- If given a non-Channel object, it errors.
----@return boolean
-function IWeakResource:is_valid() end
-
 ---@alias c2.LogLevel integer
 ---@type { Debug: c2.LogLevel, Info: c2.LogLevel, Warning: c2.LogLevel, Critical: c2.LogLevel }
 c2.LogLevel = {}
@@ -53,7 +44,7 @@ ChannelType = {}
 ---@type { Twitch: Platform }
 Platform = {}
 
----@class Channel: IWeakResource
+---@class Channel
 Channel = {}
 
 --- Returns true if the channel this object points to is valid.
--- a/docs/plugin-meta.lua
+++ b/docs/plugin-meta.lua
@@ -7,6 +7,7 @@
 c2 = {}
 
 ---@class IWeakResource
+IWeakResource = {}
 
 --- Returns true if the channel this object points to is valid.
 --- If the object expired, returns false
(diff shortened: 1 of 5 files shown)


docs/plugin-meta.lua
```diff
@@ -7,42 +7,54 @@
...
 
 ---@class CompletionList
 ---@field values string[] The completions
 ---@field hide_others boolean Whether other completions from Chatterino should be hidden/ignored.
--- Now including data from src/common/Channel.hpp.
+
+---@class CompletionEvent
+---@field query string The word being completed
+---@field full_text_content string Content of the text input
+---@field cursor_position integer Position of the cursor in the text input in unicode codepoints (not bytes)
```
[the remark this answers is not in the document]
Mm2PL: Oops. I made a mistake here. Also the problem with having utf-16 position here is that Lua doesn't know about UTF-16, it only knows byte offsets and UTF-8. Ideally it should be as this comment describes: in code points.

scripts/make_luals_meta.py
```diff
@@ -42,6 +45,7 @@
 c2 = {}
 
 ---@class IWeakResource
+IWeakResource = {}
```
Mm2PL, line 48: Should this really define a table? Chatterino itself does not expose that table. `IWeakResource` is an interface some objects can implement.

scripts/make_luals_meta.py
```diff
@@ -58,116 +62,272 @@
...
+        if comment.startswith("@lua@param"):
+            params.append(comment.split(" ", 2)[1])
+
+        out.write(f"---{comment.removeprefix('@lua')}\n")
+
+    if not comments[-1].startswith("@exposed "):
+        panic(path, line, f"Invalid function exposure - got '{comments[-1]}'")
+    name = comments[-1].split(" ", 1)[1]
+    printmsg(path, line, f"function {name}")
+    out.write(f"function {name}({", ".join(params)}) end\n\n")
```
Mm2PL, line 222: This requires Python 3.12, use different quotes or extract the join call into a variable. Also my syntax highlighter doesn't like it

```suggestion
    out.write(f"function {name}({', '.join(params)}) end\n\n")
```

Mm2PL (changes requested):

scripts/make_luals_meta.py
Nerixyz: Since you can't access `IWeakResource:is_valid` in Lua code, I removed it. `Channel` already has `Channel:is_valid` defined (with the same doc comment).

Mm2PL (approved):

PR #5286, opened 2024-03-31
Fix some documentations & comments

pajlada:
- **add comments for the new reward filters**
- **slightly improve documentation of r9k values**

### DIFF
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -462,7 +462,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
     layout.addTitle("Messages");
     layout.addCheckbox(
         "Separate with lines", s.separateMessages, false,
-        "Adds a line inbetween each message to help better tell them apart.");
+        "Adds a line between each message to help better tell them apart.");
     layout.addCheckbox("Alternate background color", s.alternateMessages, false,
                        "Slightly change the background behind every other "
                        "message to help better tell them apart.");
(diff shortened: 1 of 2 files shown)


pajlada (commented):
image

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -929,7 +935,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         },
         [](auto args) {
             return fuzzyToInt(args.value, 5);
-        });
+        },
+        true,
+        "A value of 5s means if there's a 5s break inbetween messages, we will "
```
Mm2PL, line 940: "inbetween" is wrong. IMO this should be:
```suggestion
        "A value of 5s means if there's a 5s break between messages, we will "
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -904,7 +904,10 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                           toggleLocalr9kShortcut + ".");
     layout.addCheckbox("Hide similar messages", s.similarityEnabled);
     //layout.addCheckbox("Gray out matches", s.colorSimilarDisabled);
-    layout.addCheckbox("By the same user", s.hideSimilarBySameUser);
+    layout.addCheckbox(
+        "By the same user", s.hideSimilarBySameUser, false,
+        "When checked, messages that are very similar to each other can still "
+        "be sent as long as they're sent by different users.");
```
Mm2PL, line 910: Messages can be shown, not sent

Mm2PL (changes requested):
:tf:

Mm2PL (approved):

Issue #4099, opened 2022-11-02
Create CI job that builds on Windows using VCPKG


pajlada:
vcpkg can be used on any platform iirc but it's mostly relevant for Windows developers

PR #5305, opened 2024-04-07
Clean up QtCreator's package manager setup

hemirt:
Added a QtCreatorPackageManager.cmake that gets auto imported when using Qt>6 with CMake and "Package manager auto setup" option (by default) turned on.

In this file, options to skip the automatic packagae manager setup is set up.

Options to skip exclusively conan, or vcpkg, are also set to on for self-documentation for the user.

Contributors using Qt Creator should pick just one (or none) to use and disable the other one, just now by default all are disabled.

A general status message with information about this gets sent right at the start of cmake command (as the QtCreatorPackageManager.cmake is one of the earliest things Qt Creator runs).

https://www.qt.io/blog/qt-creator-cmake-package-manager-auto-setup

[Extra Notes]
Vcpkg seems like it's not configured well to use with Qt Creator - since it downloads qt (and/or builds it).
Conan might not have this "issue" but I did not test it out, since even the build of openssl or boost was taking well over 15 minutes already.


QtCreatorPackageManager.cmake
```diff
@@ -0,0 +1,7 @@
+if (NOT QT_CREATOR_SKIP_PACKAGE_MANAGER_SETUP)
+    message(STATUS "Skipping Qt Creator automatic package manager (skipping Conan and VCPKG) - check QtCreatorPackageManager.cmake for more details")
+endif()
+
+set(QT_CREATOR_SKIP_PACKAGE_MANAGER_SETUP ON)
+set(QT_CREATOR_SKIP_CONAN_SETUP ON)
+set(QT_CREATOR_SKIP_VCPKG_SETUP ON)
```
Nerixyz, line 7: Maybe I read the blog-post incorrectly, but doesn't `QT_CREATOR_SKIP_PACKAGE_MANAGER_SETUP` imply that checks for conan & vcpkg are skipped?

There should also be a trailing newline here.

Nerixyz (commented):
> Contributors using Qt Creator should pick just one (or none) to use and disable the other one, just now by default all are disabled.

In my opinion, if you have installed Qt Creator, you should be using conan, because it doesn't bring in Qt but uses the version installed on the system. vcpkg will compile qt from source, which can take some time (and might conflict with some assumptions Qt Creator makes (?)). Furthermore, only conan is tested in CI (https://github.com/Chatterino/chatterino2/issues/4099).

> Conan might not have this "issue" but I did not test it out, since even the build of openssl or boost was taking well over 15 minutes already.

Conan doesn't build boost anymore. Only OpenSSL is built from source (this takes some time on Windows because they refuse to support any other build tool than nmake).

---

Since we support both package managers, I think it's a good idea to include a `QtCreatorPackageManager.cmake` (compared to adding it to `.gitignore`). However, only conan should be enabled. If a user doesn't want to have the auto-setup, they can turn it off in the Qt Creator settings.

---

This PR should also include documentation changes (at least) in `BUILDING_ON_WINDOWS.md`.

QtCreatorPackageManager.cmake
pajlada: It should be fine to comment out the `QT_CREATOR_SKIP_VCPKG_SETUP` option then, so people know it exists still. While adding comments in there, you could also add a link to https://www.qt.io/blog/qt-creator-cmake-package-manager-auto-setup in a comment so people know better where to configure/how to configure things

hemirt: Should I leave some skip setting off? As Nerixyz mentions conan supposedly uses Qt installed on system, whilst vcpkg does bring in a different version.

pajlada: The end result of the PR should, imo, be: Qt Creator now auto-configures Conan when loading the project, but skips vcpkg. Not 100% sure how that would look

hemirt: resolved as such

hemirt:
Conan now autoexecutes and downloads or builds the libraries (it builds zlib and openssl for me)

But when trying to run chatterino from qt creator, I get these errors

image

image

The folder `build\Desktop_Qt_6_7_1_MSVC2019_64bit-Debug\bin` where chatterino.exe is located is missing the `libcrypto-3-x64.dll` and `libssl-3-x64.dll` files.

If I copy the files from `build\Desktop_Qt_6_7_1_MSVC2019_64bit-Debug\conan-dependencies\build\conan\bin` (where these are the only two files), chatterino then starts.

I'm not sure if this is a problem with the conanfile, or the cmakelists file.

pajlada:
This PR works for me - I don't get the error you get with missing openssl DLLs, my kit's Run configuration looks like this:
image

If I disable "Add build library search path to PATH", I get your error.

If I launch the exe without qt creator, I also get your error.

I think this is all acceptable.

Could you check if you have the "Add build library search path to PATH" disabled in your run configuration?

hemirt:
> Could you check if you have the "Add build library search path to PATH" disabled in your run configuration?

I do have that checked. I can't launch the app inside Qt creator, nor outside it.
Did your configuration copy the libcrypto files into the bin folder next to the chatterino.exe?
`chatterino2\build\Desktop_Qt_6_7_1_MSVC2019_64bit-Debug\bin`
Mine does not.

Unchecking this setting did not change anything for me.

Maybe your build library search path (unsure what this variable is supposed to be) somehow has these openssl libraries?

pajlada:
My crypto DLLs are not added to the same directory as `chatterino.exe` is, no.
It seems my version of Qt creator (which was freshly installed at v13.0.2) handles adding conan to PATH before running `chatterino.exe`

image

pajlada (approved):

PR #5309, opened 2024-04-08
fix: hide tooltip on window leave event

### LINKED ISSUE #5308: Split header tooltip sometimes gets stuck
### Checklist

### Describe your issue

Right-clicking on a split header while the Chatterino window is inactive causes the split header tooltip stay visible until you hover over the split header again. Normally, the tooltip should disappear as soon as you right-click.

To reproduce:
1. Focus any other window than Chatterino
2. Right-click on the split header (this might take a few tries)
3. Split header tooltip stays stuck until the split header is hovered over again

### Screenshots

https://github.com/Chatterino/chatterino2/assets/58441/0dc78732-6c5f-4aea-8127-d25a0a283013

### OS and Chatterino Version

Chatterino Nightly 2.5.0-beta.1 (commit 25a69fd1) built on 2024-04-07 with Qt 6.5.0, MSVC 193833135 Running on Windows 11 Version 24H2, kernel: 10.0.26090

Nerixyz:
Disclaimer: This solution is really hacky.

I think https://github.com/Chatterino/chatterino2/issues/5308 is a bug in Qt. I'm not entirely sure what the root cause is (possibly some bad order of queued events). The bug happens when quickly moving the mouse and right-clicking on the split header. When the QMenu is shown, focus is transferred, eventually causing a leave-event to be sent to the focused window. In our case, it's a QWidgetWindow. Usually, the widget to receive the leave-event (and propagate it upwards) is [set to `qt_last_mouse_receiver`](https://github.com/qt/qtbase/blob/ce23c9083f4d06054eb59781c3309b1fcc807a1f/src/widgets/kernel/qwidgetwindow.cpp#L411-L412) - e.g. the QLabel showing the channel info. However, when things go wrong, [`qt_last_mouse_receiver` is set to the newly created QMenu](https://github.com/qt/qtbase/blob/ce23c9083f4d06054eb59781c3309b1fcc807a1f/src/widgets/kernel/qwidgetwindow.cpp#L546) before the leave-event is delivered. But wait, doesn't `sendMouseEvent` get the last mouse receiver, so it could deliver the leave-event? Yea, but it only does so [if the mouse is over the QMenu](https://github.com/qt/qtbase/blob/ce23c9083f4d06054eb59781c3309b1fcc807a1f/src/widgets/kernel/qapplication.cpp#L2320), which it isn't.

So, what do we do? If the receiver isn't set to `qt_last_mouse_receiver`, it's set to the window-widget, `m_widget`, which is our `BaseWindow`. To prevent the tooltip getting stuck, we use the window's leave-event to hide it. This isn't ideal, but it works.

If you test a lot, you might notice similar effects in other parts. For example, buttons sometimes get stuck in a hovered state when moving the mouse quickly and clicking. I suspect the same (or a really similar) cause. This is a pain to debug - it would be nice if there was an easier way to reproduce.

Fixes #5308.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed split tooltip getting stuck in some cases. (#5309)
+
 ## 2.5.0-beta.1
 
 - Major: Twitch follower emotes can now be correctly tabbed in other channels when you are subscribed to the channel the emote is from. (#4922)
--- a/src/widgets/BaseWindow.cpp
+++ b/src/widgets/BaseWindow.cpp
@@ -527,6 +527,7 @@ void BaseWindow::changeEvent(QEvent *)
 
 void BaseWindow::leaveEvent(QEvent *)
 {
+    this->leaving.invoke();
 }
 
 void BaseWindow::moveTo(QPoint point, widgets::BoundsChecking mode)
--- a/src/widgets/BaseWindow.hpp
+++ b/src/widgets/BaseWindow.hpp
@@ -85,6 +85,7 @@ public:
     void setTopMost(bool topMost);
 
     pajlada::Signals::NoArgSignal closing;
+    pajlada::Signals::NoArgSignal leaving;
 
     static bool supportsCustomWindowFrame();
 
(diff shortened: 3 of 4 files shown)


sando:
Confirmed this fixes the issue, thanks!

pajlada (approved):
Thank you!

Issue #3283, opened 2021-10-11
version numbers missing on macOS


core-code:
i've downloaded the version 2.3.4 but when selecting it in the Finder the version number is missing

all macOS apps are supposed to have version information

the `CFBundleShortVersionString` and `CFBundleVersion` entries in your `Info.plist` file are missing
<img width="219" alt="Screenshot 2021-10-11 at 16 30 37" src="https://user-images.githubusercontent.com/1261954/136807897-6ca9cdd2-5b69-4ee0-a2c6-78030e597f81.png">
<img width="256" alt="Screenshot 2021-10-11 at 16 28 21" src="https://user-images.githubusercontent.com/1261954/136807900-de98f909-3aa0-4ef8-b313-ca03a8f2a01d.png">

leon-richardt:
It appears the [keys are indeed set in our `plist`](https://github.com/Chatterino/chatterino2/blob/b0e79b603e388912ed58ca2cccb299d03b0116e5/cmake/MacOSXBundleInfo.plist.in) but the variables appear to be empty. Possibly an issue with the build system?

@core-code Did you download Chatterino from https://chatterino.com or somewhere else?

core-code:
i used this URL:

`https://chatterino.fra1.digitaloceanspaces.com/bin/2.3.4/Chatterino.dmg`

PR #5311, opened 2024-04-09
fix for version showing incorrectly in the macos get info window (fix for #3283)

nealxm:
fix for #3283

according to [apple](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html) the `CFBundleGetInfoString` plist key is not to be used. in old apple applications is used to be for showing the version which is why the current info string shows in the version field.

since the key has been replaced by `NSHumanReadableCopyright` i suggest removing the old key so that the version shows correctly in the get info window.

current:
Screenshot 2024-04-09 at 15 53 20

proposed:
Screenshot 2024-04-09 at 15 54 46


pajlada (approved):
Thanks! I confirmed this works as expected on my mac mini running macOS sonoma

nealxm:
Glad this fix worked! apologies I am not able to create a different pull request I should have pushed changes to different branches. Hope this is still okay

Felanbird:
> Glad this fix worked! apologies I am not able to create a different pull request I should have pushed changes to different branches. Hope this is still okay

It's not please go ahead and remove the contributor change from this PR, otherwise if there ends up being a problem with this PR we'll have to revert the commit, which contains your contributor addition.

As you mentioned for the future you should only make changes to separate branches, and keep your master branch out of it.

nealxm:
apologies again for messing that up. should be all good now. thank you both again!

Issue #5315, opened 2024-04-10
All tabs deleting themselves besides online tabs


Jawman2000:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Only tabs of streamers that are live stay open. Any other tabs i.e. offline chat rooms, /whispers, /mentions all remove themselves.

### Screenshots

https://github.com/Chatterino/chatterino2/assets/109163523/0c9c407b-c632-485a-ba37-f61c3531082d

### OS and Chatterino Version

Chatterino 7.4.6 (commit f2d677b2) built with Qt 6.5.0, MSVC 193532217 Running on Windows 11 Version 21H2, kernel: 10.0.22000

Wissididom:
Try pressing Ctrl+Shift+L :thinking:

Nerixyz:
> Try pressing Ctrl+Shift+L 🤔

This should probably be in the context menu of the notebook for better discoverability.

KleberPF:
Considering how often this gets asked, would a pop-up (with the option to hide next time) explaining why the new tab got hidden be a good idea? There is also the context menu option missing, as pointed out by Nerixyz.

PR #5318, opened 2024-04-11
Add context menu entry to toggle offline tabs

Nerixyz:
Adds an entry to toggle the visibility of offline tabs, as mentioned in https://github.com/Chatterino/chatterino2/issues/5315#issuecomment-2048367176.

I'll note that the context menus for the notebook, split-header and the menu of a split are shown when pressing the mouse, not when releasing it, which is unintuitive.


pajlada:
> I'll note that the context menus for the notebook, split-header and the menu of a split are shown when pressing the mouse, not when releasing it, which is unintuitive.

This is a platform-specific discrepancy we have. On Linux, the norm is to show the menu on mouse-down rather than mouse-up

pajlada (approved):
LGTM - there's some UX improvements that could be done with tab visibility in general but that doesn't have to be completed in this PR, or in this release. This will already help reduce support traffic of "oops all my tabs are gone"

PR #5293, opened 2024-04-02
feat: report duration for multi-month anon sub gifts

iProdigy:
Similar to #2026 but for subgift (and these irc tags are [actually](https://dev.twitch.tv/docs/irc/tags/#usernotice-tags) documented)

Before
Screenshot_20240401_163359

After
image

PR #5319, opened 2024-04-11
feat: include duration in more multi month gifts

iProdigy:
Fixes https://github.com/Chatterino/chatterino2/pull/5293#issuecomment-2048923176

New case (previously wouldn't include duration):
Screenshot_202
40411_115311

Old cases:
Screenshot_20240411_115234
image


pajlada (commented):
We're now back to "AnAnonymousGifter" as the username which feels meh

pajlada (approved):

PR #5321, opened 2024-04-12
fix: improve error messaging when pgrep fails streamer mode detection

pajlada:
We now correctly detect a timeout & show a special error message for it.
I've gone for the approach of possibly posting two streamer-mode messages, instead of sharing a boolean for them.

I've tested this locally on my system in 2 ways
1) Remove pgrep - the original messaging still works
2) Replace pgrep with a script that runs sleep 5 - the new timeout message shows


pajlada (commented):
image

Issue #5317, opened 2024-04-11
`player.twitch.tv` URLs require an unnecessary redirect


Felanbird:
Back in 2020 we were required to add `&parent=twitch.tv` to these URLs as they stopped working without it, sometime after this Twitch changed the scheme. While our version of the URL has continued to work, Twitch could stop offering us the redirect and we'll end up with broken URLs again. Also i'm sick of seeing the redirect

wrong (what we do): https://player.twitch.tv/?parent=twitch.tv&channel=pajlada 
correct: https://player.twitch.tv/?channel=pajlada&parent=twitch.tv

We talked about this years ago in chat, and IIRC pajlada wanted all of these to be handled in a singular way, so that should be done instead of just changing the existing 3 separate URLs

https://github.com/Chatterino/chatterino2/blob/dd62707d535cd556f89ee3080ac234cf2ba5c426/src/singletons/Toasts.cpp#L180-L182
https://github.com/Chatterino/chatterino2/blob/dd62707d535cd556f89ee3080ac234cf2ba5c426/src/widgets/splits/Split.cpp#L824-L827
https://github.com/Chatterino/chatterino2/blob/dd62707d535cd556f89ee3080ac234cf2ba5c426/src/providers/twitch/TwitchChannel.cpp#L87-L88

iProdigy:
Is there actually any redirect happening here?

```
curl -I -X GET https://player.twitch.tv/?parent=twitch.tv&channel=pajlada

HTTP/1.1 200 OK
```

seems like the query params are just sorted by client-side js for whatever reason

Felanbird:
well i'm sick of seeing it either way

Felanbird:
#5326

PR #5326, opened 2024-04-15
Change order of query parameters of Twitch Player URLs.

mavjs:
- This feature removes the unnecessary redirect of `player.twitch.tv` by moving the `channel` and `parent` key around.
- This also makes sure that the URL used in multiple place can be changed in one place.
- Attempts at fixing #5317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Minor: Added context menu action to toggle visibility of offline tabs. (#5318)
 - Minor: Report sub duration for more multi-month gift cases. (#5319)
 - Minor: Improved error reporting for the automatic streamer mode detection on Linux and macOS. (#5321)
+- Minor: Remove unnecessary redirect of player.twitch.tv URLs (#5327)
 - Bugfix: Fixed a crash that could occur on Wayland when using the image uploader. (#5314)
 - Bugfix: Fixed split tooltip getting stuck in some cases. (#5309)
 - Bugfix: Fixed the version string not showing up as expected in Finder on macOS. (#5311)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,13 @@
 - Minor: Added context menu action to toggle visibility of offline tabs. (#5318)
 - Minor: Report sub duration for more multi-month gift cases. (#5319)
 - Minor: Improved error reporting for the automatic streamer mode detection on Linux and macOS. (#5321)
-- Minor: Remove unnecessary redirect of player.twitch.tv URLs (#5327)
 - Bugfix: Fixed a crash that could occur on Wayland when using the image uploader. (#5314)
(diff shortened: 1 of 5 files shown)


src/common/Common.hpp
```diff
@@ -12,6 +12,9 @@
 #define LINK_CHATTERINO_DISCORD "https://discord.gg/7Y5AYhAK4z"
 #define LINK_CHATTERINO_SOURCE "https://github.com/Chatterino/chatterino2"
 
+const QString FSTRING_TWITCH_PLAYER_URL =
```
pajlada, line 15: Rename to `TWITCH_PLAYER_URL` and move into the chatterino namespace

src/widgets/splits/Split.cpp
```diff
@@ -836,8 +836,7 @@ void Split::openChannelInBrowserPlayer(ChannelPtr channel)
     if (auto *twitchChannel = dynamic_cast<TwitchChannel *>(channel.get()))
     {
         QDesktopServices::openUrl(
-            "https://player.twitch.tv/?parent=twitch.tv&channel=" +
-            twitchChannel->getName());
+            QUrl(FSTRING_TWITCH_PLAYER_URL.arg(twitchChannel->getName())));
```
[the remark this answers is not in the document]
pajlada: Not something this PR should fix, since the same "issue" was possible before this

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Minor: Added context menu action to toggle visibility of offline tabs. (#5318)
 - Minor: Report sub duration for more multi-month gift cases. (#5319)
 - Minor: Improved error reporting for the automatic streamer mode detection on Linux and macOS. (#5321)
+- Minor: Remove unnecessary redirect of player.twitch.tv URLs (#5327)
```
pajlada, line 8: I added a missing `.`, changed the PR number, made the category a `Dev` entry since it doesn't actually change anything, and updated the wording to better reflect what it does (which we can be more free with since it'd be a dev entry). You'd also need to move it to the bottom of the changelog entries since it's now a dev entry, which is why I haven't used the suggestion functionality.

`- Dev: Changed the order of the query parameters for Twitch player URLs. (#5326)`

pajlada (changes requested):

src/common/Common.hpp
Nerixyz: This should be declared `inline` and possibly use `QStringLiteral` to avoid allocating at runtime.

pajlada (approved):

Mm2PL (approved):

mavjs:
@pajlada - appreciate you stepping in to make the changes. <3

I did have a similar thought about moving the `TWITCH_URL` string under the `chatterino` namespace, however, I wanted to see first what others would recommend with the current state I submitted.

PR #5334, opened 2024-04-19
refactor: cleanup and document `Scrollbar`

Nerixyz:
Working on https://github.com/Chatterino/chatterino2/pull/5329, I realized that `Scrollbar` needed some documentation and a rename of `largeChange`. Then I saw some unused functions and clang-tidy warnings, so I decided to refactor the whole class.

The commits and documentation should work on their own - if you're missing information, please point that out and I'll add more docs.

Since I added some documentation, I decided to add a `doxygen` target (if available), so I don't have to re-add that every time I want to generate documentation. It's not meant to be pretty - we could use [breathe](https://github.com/breathe-doc/breathe) for that. But it's not really a priority since Chatterino isn't a library.


Pizzabelly:
Nice. I could suggest renaming `resetMaximum()` to `resetBounds()`. Maybe also a note about `getRelativeCurrentValue()` being the current position adjusted to the bounds.

pajlada:
I added some tests for `addHighlightsAtStart` that run on current master, but fails with this PR

src/widgets/Scrollbar.hpp
```diff
@@ -1,53 +1,130 @@
...
+/// still shift upwards (just at a different offset). To avoid this, there's a
+/// _relative current value_, which is `currentValue - minimum`. It's the
+/// actual index of the top message in the buffer. Since the minimum is shifted
+/// by 1 when messages come in, the view will remain idle (visually).
 class Scrollbar : public BaseWidget
 {
     Q_OBJECT
 
 public:
-    Scrollbar(size_t messagesLimit, ChannelView *parent = nullptr);
```
Nerixyz, line 21: Could use [`gsl::not_null`](https://github.com/microsoft/GSL/blob/main/docs/headers.md#user-content-H-pointers-not_null).

pajlada: the parent could technically be changed to a QObject since we don't use it as a ChannelView, nullptr here is currently used in tests

dnsge:
I have a couple more things that would be nice to refactor, see https://github.com/Nerixyz/chatterino2/pull/22 :)

pajlada (approved):

PR #5336, opened 2024-04-20
fix: Remove "Show chatter list" entry from split header menu for non-mods

### LINKED ISSUE #5335: Show chatter list button still available for non-moderators
### Checklist

### Describe your issue

https://github.com/Chatterino/chatterino2/pull/5245 hid the chatter list button for non-moderators because non-moderators don't have access to the chatter list anymore. However, the button `three dots -> more -> Show chatter list` still exists even for non-moderators.  
I am not sure if this is intentional or not but if not it should be removed.

### Screenshots

The chatter list button next to the three dots is no there but in the menu the button still exists.
Show chatter list button

### OS and Chatterino Version

Chatterino 2.5.0-beta.1 (commit 7c97e6bc modified) built with Qt 6.7.0 Running on Arch Linux, kernel: 6.8.7-arch1-1

pajlada:
Fixes #5335


pajlada:
image

PR #5340, opened 2024-04-21
fix: use openssl3 on qt6 windows builds

pajlada:


.github/workflows/build.yml
```diff
@@ -54,6 +54,8 @@ jobs:
       C2_PLUGINS: ${{ matrix.plugins }}
       C2_ENABLE_CRASHPAD: ${{ matrix.skip-crashpad == false }}
       C2_BUILD_WITH_QT6: ${{ startsWith(matrix.qt-version, '6.') }}
+      C2_USE_OPENSSL3: ${{ startsWith(matrix.qt-version, '6.') && 'True' || 'False' }}
+      C2_CONAN_CACHE_SUFFIX: ${{ startsWith(matrix.qt-version, '6.') && '-QT6' || '' }}
```
Nerixyz, line 58: This was added to the `build-ubuntu-docker` job but should be added to `build` job. Additionally, it should be added to `test-windows.yml`.

PR #5345, opened 2024-04-21
fix: use the full url when resolving

### LINKED ISSUE #5343: Unable to open links without protocol
### Checklist

### Describe your issue

I can't open links that don't have a protocol (http/https). I can click them, but nothing happens.

Links that work:
http://www.google.com
https://www.google.com

Links that do not work:
www.google.com
google.com

The link preview also does not work for these links.

image

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.0 DEBUG (commit 1a04bda5 modified) built with Qt 5.15.13 Running on Arch Linux, kernel: 6.8.5-arch1-1

Nerixyz:
The link passed to the link resolver is the one that ends up in the return value of `getLink()` (if it wasn't resolved in the meantime).

Calling `setLink` on a `LinkElement` doesn't actually do anything, because it provides its own `getLink`.

Fixes #5343.

<sub>voting for a patch release this week</sub>

### DIFF
--- a/tests/src/LinkParser.cpp
+++ b/tests/src/LinkParser.cpp
@@ -1,11 +1,8 @@
 #include "common/LinkParser.hpp"
 
 #include <gtest/gtest.h>
-#include <QList>
 #include <QString>
 #include <QStringList>
-
-#include <vector>
 
 using namespace chatterino;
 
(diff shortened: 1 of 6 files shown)


tests/src/LinkParser.cpp
```diff
@@ -173,3 +176,80 @@ TEST(LinkParser, doesntParseInvalidLinks)
...
+                << input.toStdString() << ": "
+                << p.result()->rest.toString().toStdString()
+                << " != " << expectedResult->rest.toString().toStdString();
+            ASSERT_EQ(p.result()->source, expectedResult->source)
+                << input.toStdString() << ": "
+                << p.result()->source.toStdString()
+                << " != " << expectedResult->source.toStdString();
+        }
+    }
+}
```
Nerixyz, line 255: Why are these tests separate? All of them have the property that `source == protocol + host + rest`. If the input should be printed in the assertion, this should be added in `Check::check`.

pajlada: When reading through the tests, this was not immediately clear to me. Having an explicit input & expected output from a test makes it clearer. I understand this does not hit any paths that weren't hit before.

If you want I can move these tests into a separate PR

pajlada (approved):

PR #5346, opened 2024-04-21
refactor: clean up toStdString usages in tests

pajlada:
- **tests: Add QString/QStringView << operators for std::ostream**
- **tests: clean up toStdString usages**

### DIFF
--- a/tests/src/TestHelpers.hpp
+++ b/tests/src/TestHelpers.hpp
@@ -1,6 +1,10 @@
 #pragma once
 
+#include <QString>
+#include <QStringView>
+
 #include <mutex>
+#include <ostream>
 
 template <typename T>
 class ReceivedMessage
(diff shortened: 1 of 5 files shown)


tests/src/TestHelpers.hpp
```diff
@@ -42,3 +46,21 @@ class ReceivedMessage
...
+
+inline std::ostream &operator<<(std::ostream &os, const QStringView &str)
+{
+    os << qUtf8Printable(str.toString());
+    return os;
+}
+
+inline std::ostream &operator<<(std::ostream &os, const QByteArray &bytes)
+{
+    os << qUtf8Printable(QString::fromUtf8(bytes));
```
Nerixyz, line 58: This is doing UTF8 → UTF16 → UTF8. There's [toStdString](https://doc.qt.io/qt-6/qbytearray.html#toStdString) and even this isn't _really_ needed.

pajlada: fixed in [`208b90a` (#5346)](https://github.com/Chatterino/chatterino2/pull/5346/commits/208b90a9b04a69162072ab0c366dbcac45bb8ef4)

Issue #1956, opened 2020-09-11
Clean Up Settings


leon-richardt:
# What should be added?
Devise a way to clean up and organize Chatterino's "Settings" section. I mean this issue to be place to collect ideas on how to go about this.

## Basic & Advanced Section
Divide settings into a "Basic" and "Advanced" section, shoving obscure and rarely used options into the "Advanced" section. Sections could be divided by having separate tabs, for example.

For example, Chromium uses this approach:
<details>
<summary>Screenshot</summary>

![Chromium settings](https://user-images.githubusercontent.com/35232120/92950063-6496e280-f45c-11ea-8336-92fceb4bf8b7.png)
</details>

## "Developer Options"-style Settings
(Originally proposed by @Supinic.)

In order to access these settings, users would have to check a box first. Somewhat similar to the first approach, however it hides more obscure settings behind an additional barrier.

For example, Android uses this approach:
<details>
<summary>Screenshot</summary>

![Android settings](https://user-images.githubusercontent.com/35232120/92950620-45e51b80-f45d-11ea-91b5-35957756223b.png)
</details>

## Hide obscure settings altogether
This is the most radical approach but I will mention it anyway as it might inspire some other solution.

Remove obscure options from the GUI altogether and expose them only via editing the configuration file directly. This is probably a hassle to migrate users to though. (And thus very unlikely to be implemented.)

# Why should it be added?
Chatterino's settings are getting increasingly cluttered as the application grows in functionality. A search bar exists to quickly navigate to a setting but still, discoverability of features is hurt by having to scroll through many different options.

At the same time, the average user likely only cares about a small subset of settings available to them. To demonstrate, I compiled a quick list of configuration options that I think are irrelevant to most users:

<details>
<summary>List of rarely used settings</summary>

- "Show which channels are live in tabs" 
- "Show input when it's empty"
- "Show message length while typing"
- "Timestamp format"
- The "R9K" section
- The "Cache" section
- The "AppData" section
- "Restart on crash"
- "Use libsecret/KWallet/Gnome keychain to secure passwords"
- "Automatically close user popup when it loses focus"
- "Enable experimental IRC support"
- "Show unhandled IRC messages"

(This is not meant to be a complete list but rather to demonstrate that there are a lot of rarely changed settings.)
</details>

I am looking forward to other suggestions!

fourtf:
I would like to see the "Basic & Advanced Section" added to the general settings.

zneix:
[...]

PR #5361, opened 2024-04-25
chore: remove unused timegates

Nerixyz:
https://github.com/Chatterino/chatterino2/issues/1956#issuecomment-2076293502 mentioned timegates, but we don't use these settings, so they clutter up the settings UI. Since Twitch wants to move away from IRC, there's no point in keeping them.


Felanbird (approved):
I know pajlada isn't the biggest fan of deleting settings, but all of these just go straight into the garbage can now, or are caught by our own handling (I have `/vips` on always use IRC but I get our normal error bounced back), so there's really no point to keep them.

pajlada (approved):

PR #5369, opened 2024-04-28
Add custom image functionality for inline mod buttons.

Mm2PL:
This also cleans up ModerationAction code


src/controllers/moderationactions/ModerationActionModel.cpp
```diff
@@ -1,28 +1,43 @@
...
 void ModerationActionModel::getRowFromItem(const ModerationAction &item,
                                            std::vector<QStandardItem *> &row)
 {
-    setStringItem(row[0], item.getAction());
+    setStringItem(row[Column::Command], item.getAction());
+    setFilePathItem(row[Column::Icon], item.iconPath());
+    if (!item.iconPath().isEmpty())
+    {
+        loadPixmapFromUrlLazy(
+            (*item.getImage())->url(), [row](const QPixmap &pixmap) {
```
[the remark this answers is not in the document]
Mm2PL: if iconPath is present, getImage will always be valid.

src/controllers/moderationactions/ModerationActionModel.cpp
```diff
@@ -1,28 +1,43 @@
...
+                row[Column::Icon]->setData(pixmap, Qt::DecorationRole);
+            });
```
Nerixyz, line 39: This should ensure the reference to `row` remains valid. Furthermore, `setData` should be called on the GUI thread.

> if iconPath is present, getImage will always be valid.

This should be noted in a comment.

src/controllers/moderationactions/ModerationActionModel.hpp
```diff
@@ -13,6 +13,11 @@ class ModerationActionModel : public SignalVectorModel<ModerationAction>
 public:
     explicit ModerationActionModel(QObject *parent);
 
+    enum Column {
```
Nerixyz, line 16: This should probably be an `enum class`.

src/controllers/moderationactions/ModerationAction.cpp
```diff
@@ -140,18 +122,18 @@ const std::optional<ImagePtr> &ModerationAction::getImage() const
...
+    }
+    else if (this->builtInImageToLoad_ == BuiltInImage::TrashCan)
+    {
+        this->image_ =
+            Image::fromResourcePixmap(getResources().buttons.trashCan);
+    }
+    else if (!this->iconPath_.isEmpty())
+    {
+        this->image_ = Image::fromUrl({this->iconPath_.toString()});
     }
```
Nerixyz, line 137: It was like this before, but the image should probably be set once (even though there's a cache).

src/util/LoadPixmapLazy.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+                // const cast since we are only reading from it
+                QBuffer buffer(const_cast<QByteArray *>(&data));
+                buffer.open(QIODevice::ReadOnly);
+                QImageReader reader(&buffer);
+
+                if (!reader.canRead() || reader.size().isEmpty())
+                {
+                    qCWarning(chatterinoSettings)
+                        << "Can't read mod action image at" << url.string << ":"
```
Nerixyz, line 32: - This mentions a mod action image, but the function is more general (it can read _any_ image).
- This will add a space between the URL and the colon (same as below).

src/util/LoadPixmapLazy.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include "common/Aliases.hpp"
+
+#include <QPixmap>
+
+namespace chatterino {
+void loadPixmapFromUrlLazy(const Url &url,
+                           std::function<void(QPixmap)> &&callback);
```
Nerixyz, line 8: - This function should be called `loadPixmapFromUrl`. There's nothing lazy about it.
- The documentation should mention that `callback` is invoked concurrently.
- The whitespace in this file should be adjusted.

Nerixyz (commented):
I can't comment there, but there's no title for the custom image:

src/controllers/moderationactions/ModerationActionModel.cpp
Mm2PL: I don't think we can even check if `row` is valid. I can't make it a smart pointer without a large refactor. I could look up the row by some kind of unique temporary ID, but it seems like a hack.

src/controllers/moderationactions/ModerationActionModel.hpp
Mm2PL: An `enum` non-`class` is implicitly convertible to int. This is really useful here as these are used for indexing a vector.

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -212,6 +214,34 @@ ModerationPage::ModerationPage()
...
         view->getTableView()->horizontalHeader()->setSectionResizeMode(
             0, QHeaderView::Stretch);
+        QObject::connect(
+            view->getTableView(), &QTableView::clicked,
+            [this, view](const QModelIndex &clicked) {
+                if (clicked.column() == ModerationActionModel::Column::Icon)
+                {
+                    auto fileUrl = QFileDialog::getOpenFileUrl(
+                        this, tr("Open Image"), QUrl(),
+                        tr("Image Files (*.png *.jpg *.jpeg)"));
```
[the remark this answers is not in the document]
Mm2PL: This was copypasted from age old highlight code. That one also has translations which don't work.

src/util/LoadPixmapLazy.cpp
Mm2PL: It does add a space, to be honest I don't care that much about that space.

src/controllers/moderationactions/ModerationAction.cpp
Mm2PL: I added an early-return

src/util/LoadPixmap.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+    NetworkRequest(url.string)
+        .concurrent()
+        .cache()
+        .onSuccess(
+            [callback = std::move(callback), url](const NetworkResult &result) {
+                auto data = result.getData();
+
+                // const cast since we are only reading from it
+                QBuffer buffer(const_cast<QByteArray *>(&data));
```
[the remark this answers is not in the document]
Nerixyz: I think I forgot to comment there - we can use [`QBuffer::setData`](https://doc.qt.io/qt-6/qbuffer.html#setData).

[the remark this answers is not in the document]
Mm2PL: Also the const cast isn't required (anymore?) it constructs just fine from a pointer to the local copy.

src/controllers/moderationactions/ModerationAction.cpp
```diff
@@ -124,6 +100,12 @@ ModerationAction::ModerationAction(const QString &action)
         this->line1_ = xD.mid(0, 2);
         this->line2_ = xD.mid(2, 2);
     }
+
+    if (!iconPath.isEmpty())
```
pajlada, line 104: Should probably use [QUrl::isValid](https://doc.qt.io/qt-6/qurl.html#isValid) here instead

src/controllers/moderationactions/ModerationAction.cpp
```diff
@@ -124,6 +100,12 @@ ModerationAction::ModerationAction(const QString &action)
...
+    {
+        this->builtInImageToLoad_ = BuiltInImage::None;
```
pajlada, line 106: I think since user-specified images are overrides, we should not set `builtInImageToLoad_` to None, and instead check `iconPath_` in `getImage` first.
With this change, I think it's clearer if the `builtInImageToLoad_` is instead the moderation action type (Delete, Ban, or Timeout) which would remove the None option

pajlada (approved):

PR #5378, opened 2024-05-04
fix: prefer reporting error over status for 200 OK

Nerixyz:
When the HTTP status code is 200 OK, but an error was reported by Qt, the error name is now reported (if available) instead of "200".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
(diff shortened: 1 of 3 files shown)


pajlada:
I broke out all test-changing things into https://github.com/Chatterino/chatterino2/pull/5379 & added the "wrapped test file" logic instead of forced include

pajlada (approved):

PR #5379, opened 2024-05-05
tests: better test initializing allowing for better printing

pajlada:
This PR just broke out the test changes from https://github.com/Chatterino/chatterino2/pull/5378 and changed the logic slightly, removing the force include stuff.


pajlada:
image

PR #5383, opened 2024-05-06
Introduce HTTP API for plugins

Mm2PL:
This adds a new API: `c2.HTTPRequest` that requires a new permission: `Network`.


docs/plugin-meta.lua
```diff
@@ -159,6 +160,68 @@ function Channel.by_twitch_id(id) end
 
 -- End src/controllers/plugins/api/ChannelRef.hpp
 
+-- Begin src/controllers/plugins/api/HTTP.hpp
+
+---@class HTTPResult
+---@field data string Data received from the server
```
pajlada, line 166: Data is an iffy name for this. `body` or `text` is a more accurate term.

Leaving these fields as methods instead means we could be more flexible when it comes to things like json parsing in the future (see https://docs.rs/reqwest/latest/reqwest/struct.Response.html#implementations)

src/controllers/plugins/Plugin.cpp
```diff
@@ -270,5 +273,19 @@ bool Plugin::hasFSPermissionFor(bool write, const QString &path)
     return false;
 }
 
+bool Plugin::hasHTTPPermissionFor(const QUrl &url)
+{
+    // XXX: Older compilers don't have support for std::ranges
+    // NOLINTNEXTLINE(readability-use-anyofallof)
```
pajlada, line 279: Can disable the `readability-use-anyofallof` lint in `.clang-tidy` instead

src/controllers/plugins/Plugin.cpp
```diff
@@ -270,5 +273,19 @@ bool Plugin::hasFSPermissionFor(bool write, const QString &path)
     return false;
 }
 
+bool Plugin::hasHTTPPermissionFor(const QUrl &url)
```
pajlada, line 276: Why do we accept a url here?

pajlada (commented):
Design question:
Maybe rename the permission from HTTP to Network? I think this more accurately describes what the plugin will actually be allowed to do, with less of a technical description.
"Allows the plugin to make network requests over your network/over the internet/to third party websites".

Bit of an early review, haven't looked into the implementation of anything

src/controllers/plugins/Plugin.cpp
Nerixyz: Which compiler doesn't support ranges? macOS supports it from Xcode 14.3 (we're on macos-13, so it's Xcode 15), Ubuntu is on GCC 10, which supports ranges too, and MSVC supports them.

src/controllers/plugins/api/HTTP.cpp
```diff
@@ -0,0 +1,433 @@
...
+    {"create", &HTTPRequest::create},
+    {nullptr, nullptr},
+};
+
+std::shared_ptr<HTTPRequest> HTTPRequest::getOrError(lua_State *L,
+                                                     StackIdx where)
+{
+    if (lua_gettop(L) < 1)
+    {
+        luaL_error(L, "Called c2.HTTPRequest method without a channel object");
```
Nerixyz, line 43:
```suggestion
        luaL_error(L, "Called c2.HTTPRequest method without a request object");
```

src/controllers/plugins/api/HTTP.cpp
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    QString value;
+    if (!lua::pop(L, &value))
+    {
+        luaL_error(L,
+                   "cannot get value (2nd argument of HTTPRequest:set_header)");
+        return 0;
+    }
+    QString name;
```
Nerixyz, line 257: Both `value` and `name` could be a `QByteArray` and a `NetworkRequest::header(const QByteArray&, const QByteArray&)` overload could be added (since that's the signature of `setRawHeader`).

Nerixyz (commented):
The following code segfaults after a short while:

```lua
local function send_req()
    local req = c2.HTTPRequest.create(c2.HTTPMethod.Get, "https://postman-echo.com/get")
    req:on_success(function(res)
    end)
    req:on_error(function(result)
    end)
    req:execute()
    c2.later(send_req, 100)
end

send_req()
```

src/controllers/plugins/Plugin.cpp
pajlada: idk, just reading the comment

src/controllers/plugins/api/HTTP.cpp
Mm2PL: of course i had to miss one

Mm2PL:
> The crash happens in the handlers of the request. Looking at the captured L, it's probably the thread from c2.later getting deleted while the callbacks hold on to it through lua_State.

I'll have to store an ID of the plugin or something and get the state back then.

src/controllers/plugins/Plugin.cpp
```diff
@@ -270,5 +273,19 @@ bool Plugin::hasFSPermissionFor(bool write, const QString &path)
...
+{
+    // XXX: Older compilers don't have support for std::ranges
```
iProdigy, line 278:
```suggestion
    if (url.scheme() != "http" && url.scheme() != "https")
    {
        return false;
    }
    // XXX: Older compilers don't have support for std::ranges
```

iProdigy (commented):
to avoid privilege escalation where network permission would allow reading local files:

src/controllers/plugins/Plugin.cpp
Mm2PL: This code was copy-pasted from the method above created in #5231.

> I don't think GCC 9 implements the ranges proposal (Ubuntu 20.04).
https://github.com/Chatterino/chatterino2/pull/5231#discussion_r1511419810

I'll give it a try and see if it builds.

src/controllers/plugins/Plugin.cpp
Mm2PL: For cases like restricting protocols or urls a plugin can use. This is used to only allow http and https now

docs/plugin-meta.lua
Mm2PL: To put methods on it and have data only in C++-land I'd have to rework `HTTPResult` to be a shared resource instead of just a table. I'm not sure if this is worth the complexity of that.

src/controllers/plugins/Plugin.cpp
Mm2PL: Fixed in [1cca807](https://github.com/Chatterino/chatterino2/pull/5383/commits/1cca8079af4ce9d2228ebd3292f9068b6efaac6e)

src/controllers/plugins/api/HTTP.cpp
Mm2PL: I added a new overload and changed this code to use it

Mm2PL:
> The following code segfaults after a short while:
> 
> ```lua
> local function send_req()
>     local req = c2.HTTPRequest.create(c2.HTTPMethod.Get, "https://postman-echo.com/get")
>     req:on_success(function(res)
>     end)
>     req:on_error(function(result)
>     end)
>     req:execute()
>     c2.later(send_req, 100)
> end
> 
> send_req()
> ```
> 
> The crash happens in the handlers of the request. Looking at the captured `L`, it's probably the thread from `c2.later` getting deleted while the callbacks hold on to it through `lua_State`.

Weird that this only happens with hosts that are remote that is have some latency. Seems like a pesky race condition.

pajlada:
Let me know when it's in a finished state and I'll take a look

docs/wip-plugins.md
```diff
@@ -85,6 +85,24 @@ Example:
...
+  ...,
+  "permissions": [
+    {
+      "type": "HTTP"
+    },
+    ...
+  ]
+}
+```
+
```
pajlada, line 105: I think the permission should be renamed to something along the lines of Network instead

docs/wip-plugins.md
```diff
@@ -370,6 +388,121 @@ Returns `true` if the channel can be moderated by the current user.
...
+This table describes HTTP methods available to Lua Plugins. The values behind
+the names may change, do not count on them. It has the following keys:
+
+- `Get`
+- `Post`
+- `Put`
+- `Delete`
+- `Patch`
+
+#### `HTTPResult`
```
pajlada, line 402: I prefer the name Response over Result

I know we internally call it Result, but most other libraries call them `Request` and `Response`

docs/wip-plugins.md
```diff
@@ -370,6 +388,121 @@ Returns `true` if the channel can be moderated by the current user.
...
+
+An `HTTPResult` is a table that contains the following keys:
+
+```lua
+{
+    data = "This is the data received from the server as a string",
```
pajlada, line 408: How do we see plugins use the returned data when it's only returned as a string?

src/controllers/plugins/api/HTTP.cpp
```diff
@@ -0,0 +1,462 @@
...
+                      "expected a string)");
+        lua_pushnil(L);
+        return 1;
+    }
+    auto *pl = getIApp()->getPlugins()->getPluginByStatePtr(L);
+    if (!pl->hasHTTPPermissionFor(parsedurl))
+    {
+        luaL_error(L, "Plugin does not have permission to send HTTP requests "
+                      "to this URL");
+        return 0;
```
pajlada, line 318: What makes this return 0 instead of 1 - is it because we don't push nil?
If so, why don't we push nil here?

src/controllers/plugins/api/HTTP.cpp
```diff
@@ -0,0 +1,462 @@
...
+    lua_settable(L, -3);  // metatable.__gc = SharedPtrUserData<...>::destroy
+
+    luaL_setfuncs(L, HTTP_REQUEST_METHODS, 0);
+}
+
+int HTTPRequest::on_success_wrap(lua_State *L)
+{
+    lua::StackGuard guard(L, -2);
+    auto ptr = HTTPRequest::getOrError(L, 1);
+    return ptr->on_success(L);
```
pajlada, line 93: `ptr` can be `nullptr` here, but we don't check for it

src/controllers/plugins/api/HTTP.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+public:
+    /**
+     * Sets the success callback
+     *
+     * @lua@param callback HTTPCallback Function to call when the HTTP request succeeds
+     * @exposed HTTPRequest:on_success
+     */
+    static int on_success_wrap(lua_State *L);
+    int on_success(lua_State *L);
```
pajlada, line 72: nit: why are these named `on_success` instead of `onSuccess`?
and what do their return values indicate?

docs/wip-plugins.md
Mm2PL: This isn't a unicode text string, it's handled more like a binary buffer, we convert a QByteArray into an std::string and give the plugin that. They can use it however they like.

src/controllers/plugins/api/HTTP.cpp
Mm2PL: `luaL_error` is technically `no return`. I'll update this to return imminently after calling `luaL_error`

docs/wip-plugins.md
pajlada: I was thinking more in terms of the usability of someone wanting to use a JSON API, which should be pretty common. How would that look like?

src/controllers/plugins/api/HTTP.hpp
Mm2PL: These are named `on_success` because they are API functions. This is the function you call to set the success callback. The return value is how many return values are on the stack.

src/controllers/plugins/api/HTTP.cpp
Mm2PL: `ptr` here can only be null if we sent Lua a `std::shared_ptr<HTTPRequest>` with a null value. Doing that is not sane. I don't think a check here is necessary.

docs/wip-plugins.md
Mm2PL: Making a json parser api shouldn't be incredibly hard, it's mostly just:
 1. take string from lua
 2. parse with favorite json parser
 3. push result as a table into lua

For serialization just go other way around. As long as we don't have recursive structures, step 3 will be easy.

src/controllers/plugins/api/HTTP.cpp
Mm2PL: I added one in getOrError anyway with a particularly scary error message.

pajlada (approved):

Issue #3411, opened 2021-12-19
don't ping that a channel is live when a channel is already live


killertofus:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running a recent nightly build or the *newest* stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

I get a ping that a channel is live when closing chatterino despite that channel being already live, it should only ping when the channel just got live not every time.

### OS and Chatterino Version

Chatterino 2.3.4 (commit 51ece94)

Mm2PL:
What?

Felanbird:
This behavior is intended and shouldn't be changed, it makes no sense to be given zero information when you startup Chatterino "just because" the streamer went live before you opened it.

PR #5388, opened 2024-05-10
feat: add option to suppress live notifications on startup

### LINKED ISSUE #3006: Feature Request - Suppress Windows desktop notifications on startup
Request:

It would be nice to have a checkbox (or by default) in the "Live Notifications" options to suppress desktop notifications for a while when starting chatterino.
I currently have about several selected channels for Live Notifications. And if most of them are already live when starting/restarting chatterino its a bit annoying seeing 10 notifications roll in one after another until they all are done.

Nerixyz:
This PR adds an option to suppress the initial live notification (🤓 but the title says "startup"). An initial live notification is a notification resulting from the first encountered live status update in a TwitchChannel and a fake channel (both are independent). A change could be the application starting up, but it's also a channel being joined for the first time or a channel going from TwitchChannel → fake or the other way around. I think having "startup" as the title of the option is clear enough.

Additionally, some inconsistencies have been fixed (e.g. streamer mode not suppressing toasts, toasts from fake channels not having a title, ...). TwitchChannel and fake channels now use the same implementation to send notifications.

Closes #3006.

I'm waiting for #4868, so I'll mark this as a draft.


killertofus:
also closes #3411

src/util/QCompareCaseInsensitive.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include <QtGlobal>
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+#    include <QUtf8StringView>
+#endif
+
+namespace chatterino {
+
+/// Case insensitive transparent comparator for Qt's string types
+struct QCompareCaseInsensitive {
```
Nerixyz, line 15: I'm going to use this for [`LinkParser`](https://github.com/Chatterino/chatterino2/blob/fdecb4a39f3e0340fe555797030454986b50be78/src/common/LinkParser.cpp#L46) in a future PR to remove the last allocation it does while parsing.

pajlada (approved):

PR #5404, opened 2024-05-18
fix: get rid of duplicate scale events

### LINKED ISSUE #1266: Tabs are too wide after changing zoom

Nerixyz:
Previously, two scale events were emitted to fix #1266. This PR fixes #1266 by introducing `NotebookTab::normalTabWidthForHeight` which computes the width for the height _after_ scaling (in `NotebookTab::updateSize`). This would previously require two invocations (first one to set the height and second one the width). Furthermore, it gets rid of refreshing the notebook too often (~> performing layout for every tab).

As a next step, this PR checks if the scale has changed at all in `BaseWidget::setScale`. This will probably introduce some bugs, but I think it's better to address these individually instead of spamming scale events (like I did for SplitHeader).

### DIFF
--- a/src/widgets/dialogs/SettingsDialog.cpp
+++ b/src/widgets/dialogs/SettingsDialog.cpp
@@ -397,12 +397,8 @@ void SettingsDialog::refresh()
     }
 }
 
-void SettingsDialog::scaleChangedEvent(float newDpi)
+void SettingsDialog::scaleChangedEvent(float /*newScale*/)
 {
-    assert(newDpi == 1.F &&
-           "Scaling is disabled for the settings dialog - its scale should "
-           "always be 1");
-
     for (SettingsDialogTab *tab : this->tabs_)
     {
         tab->setFixedHeight(30);
--- a/src/widgets/dialogs/SettingsDialog.cpp
+++ b/src/widgets/dialogs/SettingsDialog.cpp
@@ -397,8 +397,12 @@ void SettingsDialog::refresh()
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/widgets/dialogs/SettingsDialog.cpp
```diff
@@ -397,12 +397,8 @@ void SettingsDialog::refresh()
     }
 }
 
-void SettingsDialog::scaleChangedEvent(float newDpi)
+void SettingsDialog::scaleChangedEvent(float /*newScale*/)
 {
-    assert(newDpi == 1.F &&
-           "Scaling is disabled for the settings dialog - its scale should "
-           "always be 1");
```
pajlada, line 404: Why is this assert removed?

pajlada (commented):
Overall looks good, will test & look a bit closer tomorrow

pajlada (approved):

PR #5405, opened 2024-05-19
Fixed restricted users' usernames not being clickable

pajlada:
also made the linking be automatic from now on, ensuring we don't make the same mistake again with MentionElements


Mm2PL:
> also made the linking be automatic from now on, ensuring we don't make the same mistake again with MentionElements

Can you please also check how irc user cards work after this change? I am unable to do it for most of today

pajlada:
> > also made the linking be automatic from now on, ensuring we don't make the same mistake again with MentionElements
> 
> Can you please also check how irc user cards work after this change? I am unable to do it for most of today

The author is unchanged, both open usercards of the name

```
          V---- this opens usercard for "pajlada"
13:37 pajlada: xd
```

Mentions are unchanged, they don't open anything

```
                                  V------ this is not clickable
13:37 pajlada: yo @forsen what's up
```

I don't have to merge this in right now, so I can let you give it a try as well

pajlada:
image

PR #5406, opened 2024-05-19
fix: usercard jumping when loading data

Nerixyz:
As anticipated in #5404, there are places expecting `scaleChangedEvent` to be called more often. One such place is `Label` which is used in `UserInfoPopup`. Before, it would jump when follower and subscription data was loaded. This PR fixes that bug. Thank you to @brian6932 for reporting this.


brian6932 (approved):
Looks fixed to me.

pajlada (approved):

PR #5415, opened 2024-05-21
feat(emote-popup): save size of popup

### LINKED ISSUE #5414: Emote window position isnt saved after restarting Chatterino
### Checklist

### Describe your issue

Emote window position isnt saved afer restarting Chatterino, always spawns on left top corner. Please give option to make it always spawn left or right next to Chaterino window.

### Screenshots

_No response_

### OS and Chatterino Version

2.5.1 Windows 11

Nerixyz:
~~I don't know if this PR completely resolves #5414, but it does fix some bugs related to Qt's `frameMargins` that are added even though we're a frameless window (probably a Qt bug - I'll try to reproduce and report it). As mentioned in #5414, the emote-popup doesn't remember its position if the main window is closed while the popup is open. This PR fixes that bug as well.~~

As mentioned in https://github.com/Chatterino/chatterino2/issues/5414#issuecomment-2123260324, saving the width and height of the popup in addition to its position is reasonable. The position is now remembered correctly without Qt's invisible margins (`geometry` doesn't have that "bug"). The only tiny issue I can see is closing Chatterino with the popup open can cause the position to be off by a few pixels because of the delayed bounds on Windows, however, I don't really think this is a major issue.

Fixes #5414.


src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -203,13 +203,18 @@ EmoteMap filterEmoteMap(const QString &text,
...
     , search_(new QLineEdit())
     , notebook_(new Notebook(this))
 {
     // this->setStayInScreenRect(true);
-    this->moveTo(getIApp()->getWindows()->emotePopupPos(),
-                 widgets::BoundsChecking::DesiredPosition);
+    auto bounds = getIApp()->getWindows()->emotePopupBounds();
+    if (bounds.size().isEmpty())
+    {
+        bounds.setSize({300, 500});
```
pajlada, line 215: Previously, the default size of the popup was 300 * scale / 500 * scale

pajlada (commented):
When the emote popup is created (SplitInput.cpp), we call resize - that call needs to be removed now

pajlada (approved):

PR #5419, opened 2024-05-25
fix: use `irc-unknown` when irc server couldn't be loaded

pajlada:
If logging is enabled and you have joined an IRC server/channel before that's now been removed somehow (by irc server being removed or irc support being disabled), the channel logger will now log the channel contents to `irc-unknown` instead of erroring out


pajlada:
image

Mm2PL:
I think if a server doesn't exist we shouldn't even attempt to log these messages. The core issue is that this:
> by irc server being removed or irc support being disabled
can happen and we still receive and handle a message.

pajlada:
If we stop the message I noticed getting through from coming through, then we fix that crash. All it would take is another message coming through to cause another crash - regardless of what we do, this should still be the preferred behaviour over crashing

PR #5421, opened 2024-05-25
chore: refactor TwitchIrcServer

pajlada:
This PR aims to remove all direct uses of `getApp()->twitch`, ensuring we can add tests for anything adjacent to `TwitchIrcServer`

The class and its inheritance is ugly, so this refactor PR is also pretty ugly. To achieve the goal of this PR without making too many changes, I've added `getTwitchAbstract()` alongside `getTwitch()` which returns the `IAbstractIrcServer` of the `TwitchIrcServer`.


pajlada:
image

Small testing shows everything works as expected - I've done a full self-review, ensuring I haven't c-v-pasted anything incorrectly.

PR #5429, opened 2024-05-29
Improved error messaging for Update Channel API

### LINKED ISSUE #5332: No error handling on `/settitle`
### Checklist

### Describe your issue

When you set a title that goes over a character limit, `/settitle`'s response's `Title update failed! Are you missing the required scope?`. While not documented in https://dev.twitch.tv/docs/api/reference/#modify-channel-information, when a title's exceeds `140` characters, you receive the following error. An alternative to requesting here could be to set a client side limit, as there's no way to set titles longer than this, no matter the API.
```json
{
	"data": [],
	"error": "Bad Request",
	"message": "Title is too long.",
	"status": 400
}
```

### Screenshots

_No response_

### OS and Chatterino Version

Doesn't matter

JakeRYW:
Adding appropriate error handling for the `/settitle` and `/setgame` commands. Handles trying to set the title/game in unauthorized channels, missing scope, rate limits, and title being too long by providing  `Failed to set title - Title is too long.` error message. 

Also removed the `status` variable declaration from the `setTitle` function as it was not used.

Fixes #5332


src/controllers/commands/builtin/twitch/UpdateChannel.cpp
```diff
@@ -30,8 +30,15 @@ QString setTitle(const CommandContext &ctx)
...
 
-    auto status = ctx.twitchChannel->accessStreamStatus();
     auto title = ctx.words.mid(1).join(" ");
+
+    if(title.size() > 140)
+    {
+        ctx.channel->addMessage(
+            makeSystemMessage("Unable to set title: Title must be 140 characters or fewer."));
+        return "";
+    }
```
Nerixyz, line 40: This shouldn't be a client-side check. #5332 mentions that the response we get from Twitch's Helix API may include a `message` field which we can use to report back. Currently, we're not doing anything with the response in case of an error (see [Helix.cpp](https://github.com/Chatterino/chatterino2/blob/9ec1022405539b1a9d36870dfd645397e6ce4edf/src/providers/twitch/api/Helix.cpp#L648-L650)).

Nerixyz (changes requested):

src/controllers/commands/builtin/twitch/UpdateChannel.cpp
pajlada: Nerixyz is correct - the reason we try using server-side errors wherever possible is to ensure compatibility with API changes (e.g. if Twitch changes their max title length to 120 or 160)

JakeRYW: Removed the client side check, and added error handling for both the `settitle` and `setgame` since they are both utilizing the Update Channel API.

src/providers/twitch/api/Helix.cpp
```diff
@@ -646,7 +649,61 @@ void Helix::updateChannel(QString broadcasterId, QString gameId,
...
+                }
+                break;
+
+                case 400:
+                case 403: {
+                    failureCallback(Error::Forwarded, message);
+                }
+                break;
+
+                case 409:
```
Nerixyz, line 689: This seems wrong. Shouldn't 409 forward the message?

src/controllers/commands/builtin/twitch/UpdateChannel.cpp
```diff
@@ -40,10 +40,46 @@ QString setTitle(const CommandContext &ctx)
...
+                break;
+
+                case Error::Forwarded: {
+                    errorMessage += message;
+                }
+                break;
+
+                case Error::Unknown:
+                default: {
+                    errorMessage += "An unknown error has occured.";
```
Nerixyz, line 77: This should include the message - like "Failed to set title - An unknown error has occurred ({message})" (there's also a typo in occured → occurred, which is also present in the error handling of `manageAutoModMessages` and the comment for `secondsSinceDisconnect`).

src/controllers/commands/builtin/twitch/UpdateChannel.cpp
```diff
@@ -105,10 +141,46 @@ QString setGame(const CommandContext &ctx)
...
+                            errorMessage += message;
+                        }
+                        break;
+
+                        case Error::Unknown:
+                        default: {
+                            errorMessage += "An unknown error has occured.";
+                        }
+                        break;
+                    }
```
Nerixyz, line 181: This is almost identical to the code above and should be refactored into a function (in an anonymous namespace above `chatterino::commands`).

Nerixyz (changes requested):
This looks good, just a few things.

- Make sure you run `clang-format` on the files you edited ([currently, it's not formatted correctly](https://github.com/Chatterino/chatterino2/actions/runs/9342461969/job/25723081919?pr=5429))
- The [Helix mock declaration](https://github.com/Chatterino/chatterino2/blob/248cd46eb7cc3954b6cabed1769edf563ef24879/mocks/include/mocks/Helix.hpp#L126) needs to be updated - since the type now contains a comma, it has to be enclosed in parentheses: `(FailureCallback<HelixUpdateChannelError, QString> failureCallback)`.

src/providers/twitch/api/Helix.cpp
JakeRYW: According to the [Modify Channel Information](https://dev.twitch.tv/docs/api/reference/#modify-channel-information) docs, 409 for this endpoint will give a Too Many Requests response when trying to see the Branded Content flag too frequently, which wouldn't be hit by `/settitle` or `/settgame` of course but I just tried to handle all errors currently provided in the docs.

Nerixyz: Oh, right. I read Too Many Requests and thought it would be 429. Probably should say 409 Conflict in the docs.

JakeRYW: Since it looks like it's just a typo in the docs, I'll remove the 409, and leave it to just 429.

Nerixyz (approved):
Looks good, now you just need a changelog entry!

pajlada (approved):
Thank you for your contribution! As a first-time contributor, you can now add yourself to the contributors list that's shown inside the About page in Chatterino.

Issue #5439, opened 2024-06-05
New Warn Feature


19Topgun93:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Hello :)

Twitch has a new "Warn" Feature. "Unrecognized command: /warn".

I cannot figure out if I can resolve this issue alone. .warn also not Function.

### OS and Chatterino Version

Chatterino 2.5.1 (commit 248cd46eb) built with Qt 6.7.1 Running on Garuda Linux, kernel: 6.9.3-zen1-1-zen

pajlada:
This is a new Twitch feature, we need to implement the API(s), depending on their scope requirements & availability. There's nothing you can do (short of submitting a PR fixing it in our code)

iProdigy:
for now you can upvote https://twitch.uservoice.com/forums/310213-developers/suggestions/48491402-provide-an-api-to-trigger-the-new-warning-system-t

iProdigy:
endpoint to warn users is now in open beta https://dev.twitch.tv/docs/api/reference/#warn-chat-user

PR #5441, opened 2024-06-06
feat: notify mods when users are warned

iProdigy:
Currently the pubsub topic only contains the moderator user id but not name, so we say `A moderator` did the action (which is still clickable to open their usercard based on id)

image

Related to #5439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Minor: Added `flags.action` filter variable, allowing you to filter on `/me` messages. (#5397)
 - Minor: The size of the emote popup is now saved. (#5415)
 - Minor: Added the ability to duplicate tabs. (#5277)
+- Minor: Moderators can now see when users are warned. (#5441)
 - Bugfix: Fixed tab move animation occasionally failing to start after closing a tab. (#5426)
 - Bugfix: If a network request errors with 200 OK, Qt's error code is now reported instead of the HTTP status. (#5378)
 - Bugfix: Fixed restricted users usernames not being clickable. (#5405)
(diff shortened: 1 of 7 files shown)


pajlada (approved):
Thank you!
I made two commits
1) Added some TODO's regarding resolving the moderators user ID to a user in the future
2) I get the list of reasons instead of only the custom reason
  When a warning comes in over pubsub, the args list is a list where the first param is the user that was warned, second param is the free-text reason, and any param after that is one of the rules that the user broke
  User broke only one rule:
`QJsonObject({"args":["bbaper","","Pepega"],"created_at":"2024-06-06T09:37:10.457114704Z","created_by":"","created_by_user_id":"11148817","id":"WARNING__11148817__165495734__1717666630457","moderation_action":"warn","target_user_id":"165495734","target_user_login":"bbaper","type":"chat_login_moderation"})`  
  User broke 2 rules + a custom reason:
`QJsonObject({"args":["bbaper","custom reason xd","Pepega","Rule 2"],"created_at":"2024-06-06T09:37:05.977014158Z","created_by":"","created_by_user_id":"11148817","id":"WARNING__11148817__165495734__1717666625977","moderation_action":"warn","target_user_id":"165495734","target_user_login":"bbaper","type":"chat_login_moderation"})`
  User warned only for custom reason: `QJsonObject({"args":["bbaper","custom reason only"],"created_at":"2024-06-06T09:39:58.680292231Z","created_by":"","created_by_user_id":"11148817","id":"WARNING__11148817__165495734__1717666798680","moderation_action":"warn","target_user_id":"165495734","target_user_login":"bbaper","type":"chat_login_moderation"})`

iProdigy:
regarding the TODO, can subscribe to https://github.com/twitchdev/issues/issues/963 to track if it's fixed on twitch's end

PR #5459, opened 2024-06-16
chore: unsingletonize UserDataController

pajlada:
The `user-data.json` file will save immediately on change, and on exit
(on dtor) if necessary. So we don't need to manually call save


pajlada:
image

PR #5473, opened 2024-06-18
chore: cleanup, document, and test some RTL code

Nerixyz:
This PR is an attempt at cleaning up the RTL code and documentation a bit so it's easier to understand (that's subjective, I know).

- Renamed `containsRTL` to `lineContainsRTL_` and checked for `lineContainsRTL_ || isRTL()` in `breakLine`.
   This way, a few RTL words in a wrapped LTR message won't affect the lines below.
- Renamed `first` and `FirstWord` to `textDirection_` and `Direction` respectively.
  That's what they are about. `first` also "leaks" how it's detected (finding the first word that has a direction).
- Added `isRTL`, `isLTR`, and `isNeutral` to check the text direction.
- Moved from `int` to `size_t` and `qsizetype` (for signed values).
- Added some assertions in `addElement` to check for weird usages of elements.
- Moved from std containers (vector/stack) to [`QVarLengthArray`](https://doc.qt.io/qt-6/qvarlengtharray.html).
  Single lines usually have very few elements, so we can store them on the stack (if that's not enough, the container will allocate).
- Moved `wasPrevReversed_` from a member to a local variable (`isReversing`)
- Updated documentation for `addElement` and `reorderRTL` (and the other stuff I added)
- Added basic tests for reordering.
  The tests only test the reordering on a single line. The line must have a username element. ~~They're still a bit WIP, as e.g. images and mentions aren't tested.~~


src/messages/Image.cpp
```diff
@@ -60,7 +60,7 @@ Frames::Frames(QList<Frame> &&frames)
                                            return init + frame.duration;
                                        });
 
-    if (totalLength == 0)
+    if (totalLength <= 1)
```
Nerixyz, line 63: Not worth another PR imo.

pajlada (approved):

PR #5475, opened 2024-06-21
chore: post some ssl info on startup

pajlada:


src/main.cpp
```diff
@@ -90,6 +91,22 @@ int main(int argc, char **argv)
             attachToConsole();
         }
 
+        qCInfo(chatterinoApp)
+            << "Chatterino Qt SSL library build version:"
+            << QSslSocket::sslLibraryBuildVersionString()
+            << ", library version:" << QSslSocket::sslLibraryVersionString();
```
Nerixyz, line 97: This comma will float in space. This should probably use `.nospace()`.

Nerixyz (commented):
I don't think the info category is appropriate here. It's printing debug information, thus it should use `qCDebug`.

src/main.cpp
pajlada: fixed in [`7225b74` (#5475)](https://github.com/Chatterino/chatterino2/pull/5475/commits/7225b7475dc09ca84d89ced80a281d706a5db94e) by splitting the library build & library version strings up into separate prints

pajlada:
keeping it as info

PR #5482, opened 2024-06-23
chore: change pajlada's profile picture to v2

pajlada:


pajlada:
image
that profile picture looks good

Issue #4769, opened 2023-08-16
non-URI characters should be excluded from the link parser


brian6932:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Currently when you use a valid tld, the link parser doesn't check whether the domain characters are valid or not.

There are really 2 approaches you could go with this, you can either URI encode the characters within the white-space, imo that's unnatural and doesn't feel right for a chat program. The better thing to do is just exclude them like webchat does.

### Screenshots

c2:
![image](https://github.com/Chatterino/chatterino2/assets/18603393/90610092-8a2d-4a3b-b955-195bcb045452)
![image](https://github.com/Chatterino/chatterino2/assets/18603393/bcb4b571-0fa6-465d-b255-1fdaa027015a)
webchat:
![image](https://github.com/Chatterino/chatterino2/assets/18603393/0c29e2ac-ef67-4602-be10-4817a2e48857)

### OS and Chatterino Version

Chatterino Nightly 2.4.4 (commit 381d5c42) built on 2023-08-14 with Qt 5.15.2, MSVC 193532217 Running on Windows 10 Version 2009, kernel: 10.0.22621

cbclemmer:
Added a PR to help with this #4778

Felanbird:
I would prefer if the approach to fix this tackled the most user-facing problems, compared to patching tiny situations where a non-existent link is shown. Though this would likely require more of a refactor as it would result in two message elements.

Similar to the above mentioned `()` problem, links that trail with `.` are a more common user-facing problem. When this happens it's often due to a user writing a bot command, and ending the sentence with a `.`, which browser chat will ignore but Chatterino does not. In this case unless the link host handles this, the link will fail to load.

examples that automatically handle this problem:
[youtube](https://www.youtube.com/watch?v=lk-Br9AVVXU.)

examples where the link fails to load
[github](https://github.com/pajlada.)
[twitch](https://www.twitch.tv/pajlada.)
[runelite's plugin-hub](https://runelite.net/plugin-hub/show/dink.)
[dashduck's website](https://drscovy.com/raids.) | working version https://drscovy.com/raids

Nerixyz:
[...]

PR #5486, opened 2024-06-28
feat: strip prefixes and suffixes in links

Nerixyz:
This PR makes the link parser aware of prefixes and suffixes in links. Both are identical to the ones [accepted on GitHub](https://github.github.com/gfm/#autolinks-extension-) (prefixes: `<`, `*`, `_`, `~`, `(`; suffixes: `>`, `?`, `!`, `.`, `,`, `:`, `*`, `~`), however, the implementation doesn't count the number of parentheses. Additionally, `_` isn't a valid suffix, because I expect valid links to have `_` as a suffix (e.g. `twitch.tv/__xX_The_Gamer_Xx__`).

This adds a few more tests too.

Towards https://github.com/Chatterino/chatterino2/issues/4769 (we don't check for some weird characters in hostnames yet).


tests/src/LinkParser.cpp
```diff
@@ -56,6 +87,12 @@ TEST(LinkParser, parseDomainLinks)
         {"", "https.cat"},
         {"", "httpsd.cat"},
         {"", "http.cat", "/200"},
+        {"", "http.cat", "/200("},
+        {"", "chatterino.com><chatterino.com"},
+        {"https://", "chatterino.com><chatterino.com"},
```
pajlada, line 92: In markdown, this gets parsed as 2 links. We handle it as 1 link - that feels wrong

Nerixyz: > In markdown, this gets parsed as 2 links.

Only if the links contain a protocol (`<http://example.com><http://example.com>` → <http://example.com><http://example.com> | `<http://example.com><example.com>` → <http://example.com><example.com> | `<example.com><example.com>` → <example.com><example.com>).

I think we should reject all of them (that's much easier).

> If fixing this is out of the scope of this PR, that's fine, but I don't think they should be in the list of "valid links" if that's the case. I'd rather have them commented out in the list of invalid links

Done.

pajlada (approved):

PR #5487, opened 2024-06-30
fix: prevent user info popups re-logging messages

### LINKED ISSUE #5470: Opening the usercard logs all of the users messages again
### Checklist

### Describe your issue

At random times, a bunch of one posters comments are repeated in one big dump of comments in the logs.  All of the comments are duplicates from earlier in the same Twitch Stream.  They are all dropped into the log at the same time.  In the screenshot below, notice that every message is at the same time.  All of the messages listed were from earlier in the same stream.  They appear in the same order as before as well. I have a colleague who is experiencing the same thing, but not the same chatters.

### OS and Chatterino Version

Chatterino 2.5.1 (commit eafcb941) built with Qt 6.5.0, MSVC 193833135 Running on Windows 10 Version 22H2, kernel: 10.0.19045

jupjohn:
This PR aims to fix #5470 where opening a user's info card would cause all of the user's messages to be re-logged for that channel. This only happens with user popups because all other "fake" channels (search, reply threads etc.) explicitly set the `MessageFlags::DoNotLog` flag when adding messages from a channel snapshot.

I've also cleaned up the logic around choosing what messages to log as it would prevent logging when passed overriding flags that DID NOT contain a `DoNotLog` flag. The altered logic now handles that, as well as checks the original message on the off chance someone's tagged that too. I've retested search and reply threads which all retain the original behaviour.

Original behaviour (7bfb5ac0a):

https://github.com/Chatterino/chatterino2/assets/4962764/d243e691-234b-4d62-a44c-902cdc435282

Post-change behaviour (aca9e9b0, the hash in this clip was stale on my clipboard):

https://github.com/Chatterino/chatterino2/assets/4962764/5b123026-c10a-4674-9d3f-79590c7de2a6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Bugfix: Fixed message history occasionally not loading after a sleep. (#5457)
 - Bugfix: Fixed a crash when tab completing while having an invalid plugin loaded. (#5401)
 - Bugfix: Fixed windows on Windows not saving correctly when snapping them to the edges. (#5478)
+- Bugfix: Fixed user info card popups adding duplicate line to log files. (#5487)
 - Dev: Update Windows build from Qt 6.5.0 to Qt 6.7.1. (#5420)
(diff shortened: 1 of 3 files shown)


pajlada:
Closing in favour of https://github.com/Chatterino/chatterino2/pull/5499 - thank you!

PR #5491, opened 2024-07-04
Enable antialiasing for reply button

dnsge:
## Description

Enables antialiasing for the reply button. This has always bugged me. I saw a screenshot of Chatterino on Twitter and was instantly compelled to fix it. Before/after:

image


jupjohn (approved):
Would be nice to do a sweep of any other images that could do with the same hint - but that's for another day. Looks good to me!

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -392,6 +392,7 @@ void ImageWithCircleBackgroundLayoutElement::paint(
     if (pixmap && !this->image_->animated())
     {
         QRectF boxRect(this->getRect());
+        painter.setRenderHint(QPainter::Antialiasing);
```
Nerixyz, line 395: This should be changed to the value it had before, after calling `drawPixmap` (probably best to read it before calling `setRenderHint` rather than pushing the state).

dnsge: I went ahead and added `QPainter::save` and `QPainter::restore` calls around every message layout element paint in [a4daddf](https://github.com/Chatterino/chatterino2/pull/5491/commits/a4daddfd3de952ebfc00b28a9d5b979b6ff15ccc)

dnsge: https://doc.qt.io/qt-6/qpainter.html#save

Nerixyz (approved):
First off: this works.

The main thing I'm/was concerned with is that we do a bit too much, since we don't really need to `save` for every element.

Also noticed that the button is antialiased if you reply to a message (since the reply curve is antialiased).

dnsge:
Save/restore also affects all writes to pen, brush, etc so I added it outside since it seemed like we wanted to isolate the effects to the painter from inside each message element. But we could instead have it only wherever we use render hints, which is perhaps where it really matters.

pajlada (approved):

PR #5499, opened 2024-07-07
Add extra context to messages that are added to channels, allowing the logging controller to take more responsibility in what messages to log

### LINKED ISSUE #5470: Opening the usercard logs all of the users messages again

pajlada:
I have removed the `DoNotLog` flags from `UserInfoPopup` on purpose since all messages added from it are "reposts", i.e. messages that wouldn't be logged anyway

~~Currently it's the logger that checks whether or not to log reposts - this should just be moved to Channel's addMessage function instead~~

Fixes #5470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Bugfix: Fixed message history occasionally not loading after a sleep. (#5457)
 - Bugfix: Fixed a crash when tab completing while having an invalid plugin loaded. (#5401)
 - Bugfix: Fixed windows on Windows not saving correctly when snapping them to the edges. (#5478)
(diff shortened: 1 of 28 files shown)


pajlada:
image

PR #5500, opened 2024-07-07
refactor: add `Channel::addSystemMessage` function

pajlada:
this can be used instead of the very common `channel->addMessage(makeSystemMessage(...))` pattern


pajlada:
image

PR #5507, opened 2024-07-13
Added option to log streams by their ID, allowing for easier "per-stream" log analyzing

pajlada:
Fixes #5386


pajlada:
Will merge this in tomorrow if anyone wants to take a look, otherwise ipaja

pajlada (commented):
image

PR #5508, opened 2024-07-13
Leak a little less memory with dynamic allocations in Lua

Mm2PL:
This PR ensures that any potentially large objects will get deallocated on error instead of getting forgotten and leaked.

This would have leaked memory:
```lua
function cmd_leak()
    local LARGE = string.rep("12345", 100)
    for i = 1, 1000000 do
        pcall(c2.register_command, LARGE) -- this errors and the error is thrown away
    end
end
c2.register_command("/leak", cmd_leak)
```

Now it doesn't leak this memory.


Nerixyz:
[You're invoking undefined behavior in many cases by using `luaL_error` which uses `longjmp` under the hood](https://eel.is/c++draft/csetjmp.syn#2). On Windows, your approach _can_ (and currently does; at least in the debug configuration) result in a double-free, as [`longjmp` uses the same stack-unwinding as exceptions](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/longjmp?view=msvc-170).

Speaking of exceptions... why not use them? At that point, it's probably best to use some bindings library (most (if not all) support incremental adoption). [sol2 has a built-in trampoline](https://sol2.readthedocs.io/en/latest/functions.html?highlight=trampoline#exception-safety-handling). [LuaBridge3](https://kunitoki.github.io/LuaBridge3/Manual.html) doesn't seem to have anything like it. There are probably others with support for that. I tried to convert the channel-api to use sol2 in https://github.com/Nerixyz/chatterino2/tree/experiment/sol3 (https://github.com/Nerixyz/chatterino2/commit/45b7c0a91a1e927f9ba111ed77f994e046ffc1de) which has, despite adding adaptions for Qt types, more deletions than additions.

Mm2PL:
Actually lua has an exception backend which could be used if we compiled it as c++ instead of c

Mm2PL:
I took a look at the sol this. And not having to do the SharedUserData<Crap> is appealing but I'm not sure that it's appealing enough to rip everything out and convert it to sol.

Mm2PL:
image
Closing in favor of moving everything to sol.

PR #5510, opened 2024-07-14
chore: make `FlagsEnum` constexpr

Nerixyz:
... and `noexcept` and fix `hasAll` and use concepts and add tests and remove `std::initializer_list`.

All tests test an unscoped (`enum`) and scoped (`enum class`) enum, hence the template. Since this is used in many files, I've added `<concepts>` to the PCH.

Clangd was complaining about `Test.hpp`, so I added an export to the gtest headers.

No other code was changed.

### DIFF
--- a/src/common/FlagsEnum.hpp
+++ b/src/common/FlagsEnum.hpp
@@ -82,14 +82,14 @@ public:
 
     constexpr FlagsEnum operator|(T flag) const noexcept
     {
-        return FlagsEnum(static_cast<T>(static_cast<Int>(this->value_) |
-                                        static_cast<Int>(flag)));
+        return static_cast<T>(static_cast<Int>(this->value_) |
+                              static_cast<Int>(flag));
     }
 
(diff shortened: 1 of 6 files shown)


src/common/FlagsEnum.hpp
```diff
@@ -65,43 +80,59 @@ class FlagsEnum
...
+                                        static_cast<Int>(rhs.value_)));
+    }
+
+    constexpr bool has(T flag) const noexcept
+    {
+        return static_cast<Int>(this->value_) & static_cast<Int>(flag);
+    }
 
-        return xd;
+    constexpr bool hasAny(FlagsEnum flags) const noexcept
```
pajlada, line 100: does `FlagsEnum` without a template parameter mean "FlagsEnum with the same template parameter"?

Nerixyz: Yes. Unfortunately, I don't know how that's called.

pajlada (approved):

PR #5514, opened 2024-07-20
chore: remove Singleton & replace getIApp with getApp

pajlada:
This removes the `Singleton` base class of the remaining controllers/singletons, and manually calls their initialize/save if necessary instead, with only the necessary parameters.

With this change, I've gone made the `getApp()` function call return the `IApplication*` interface (like `getIApp()` did), so we can have one single "get god class" function.
It's been possible for a little bit already, but now everything that must be accessible is accessible through the `IApplication` interface


pajlada:
image

PR #5515, opened 2024-07-20
fix: count parentheses in links

Nerixyz:
Links like `a.com/(foo)` would be parsed as `a.com/(foo`. Although this was documented, it could result in undesirable behavior.
Parentheses are now counted inside the rest (i.e. path/query/hash).


tests/src/LinkParser.cpp
```diff
@@ -89,14 +89,21 @@ TEST(LinkParser, parseDomainLinks)
         {"", "https.cat"},
         {"", "httpsd.cat"},
         {"", "http.cat", "/200"},
-        {"", "http.cat", "/200("},
-        {"", "a.com", "?("},
-        {"", "a.com", "#("},
+        {"", "http.cat", "/200()"},
+        {"", "a.com", "?()"},
+        {"", "a.com", "#()"},
```
Nerixyz, line 94: We can't have test cases that don't have balanced parentheses, as all possible suffixes are checked and having a suffix that balances the parentheses would result in an error (e.g. `a.com#(` with `).` would result in `a.com#()` being the link and `.` being the suffix).

pajlada (approved):

PR #5516, opened 2024-07-20
ci: collect code coverage

Nerixyz:
Uses gcov and codecov to collect code coverage. The test build is quite a lot slower (about 20 min). The `CODECOV_TOKEN` secret must be set to upload.


codecov.yml
```diff
@@ -0,0 +1 @@
+comment: false
```
pajlada, line 1: rename to `.codecov.yml`

pajlada (approved):

Issue #3262, opened 2021-10-01
Chatterino reverts back to default settings after system restart


Shigure02:
**Describe your issue**
It's been a long time since this issue has been going on and sometimes when I restart my computer it reverts back to its default settings 

    Examples:
    - It reverts back to its default settings like clears the settings.
    - Tabs still stay the same and format. Splits remain, Close button re-enabled, 
    

 **Possible fix**
      -Copying the Settings folder from the %AppData% of Chatterino `` AppData > Roaming > Chatterino2 `` 

**OS and Chatterino Version**

        Win 10 Home, V 20H2 (OS Build 19042.867)
          Chatterino 2.3.4

LosFarmosCTL:
So only the current settings.json gets reset and everything else stays the same?

What about the commands and the .bkp backup settings files?

Mm2PL:
I have a little suspicion that winblows kills Chatterino while saving settings, so it leaves and invalid file behind and therefore can load the settings after restarting

Akemi02:
> So only the current settings.json gets reset and everything else stays the same?
> 
> What about the commands and the .bkp backup settings files?

I'm assuming the commands are reset as well as it wasn't saved for me and I'm not sure about the backup files as I haven't compared the differences of it yet, I apologize I'm quite new to Chatterino

Felanbird:
It's my strong belief that this issue only happens due to the users C: drive being a HDD, and it's a very old Windows issue.
I often ran into it with World of Warcraft addons, as many of them stored all of their data in `.json` files.

We discussed this in the Runelite discord back in July, as a `config.json` file is used for Runelite, and if it corrupts due to this, the client would silently fail to start. Another dev reported their belief that this was an issue that only happened on HDD's, and I confirmed I hadn't ran into it for programs installed on an SSD (chatterino, runelite, WoW addons after getting an SSD)

afaik the changes made here: [runelite/runelite@compare](https://github.com/runelite/launcher/compare/15d5b73ba7e92e0f745d74456b4ac8c8d35913e3...runelite:952d224b126dbd58691d76841634227666c18e63) fixed the problem for Runelite, but they are a java app and using fsync so this info probably doesn't help.

Nerixyz:
[...]

Issue #3260, opened 2021-09-29
Emote picker scrolling is inconsistent with multiline emote sets


jammehcow:
**Describe your issue**
When scrolling through the emote picker in any tab where emote sections span multiple lines, scrolling via scrollbar dragging inconsistently scrolls depending on the scroll position.

**Screenshots**
Was painful to demonstrate but I tried to pull the scrollbar down at a constant speed.
https://user-images.githubusercontent.com/4962764/135192618-0816154d-962e-4219-96a8-cb45494164ae.mp4

**OS and Chatterino Version**
Chatterino 2.3.4 (commit 9b9fd7d4)
Arch using Linux Zen 5.14.6 + BSPWM

jammehcow:
I've seen this occur since Is tarted using chatterino so it's by no means a new development - couldn't find an existing issue for it.

My best guess is that all emotes in each sub section (in the video) are treated as one line when they actually span multiple. It's most obvious when the top of the page goes from a label (e.g. Mizkif) to a large emote section.

sando:
Can repro on Windows, Chatterino Nightly 2.3.4 (commit 9b9fd7d4). 

Wasn't quite sure what was going on in the video at first, but trying it myself made the problem pretty obvious. 😄

jammehcow:
> Wasn't quite sure what was going on in the video at first, but trying it myself made the problem pretty obvious. smile

I've added some average steps to reproduce which hopefully explain it better. I'm sick at the moment and can't quite explain it so happy to edit.

Mm2PL:
This should also be reproducible in normal channels since it's the same ChannelView widget. Can reproduce on 177ab0eaa7f65a49c002ba9b02a9ff3f65659d9d.

jammehcow:
> This should also be reproducible in normal channels since it's the same ChannelView widget. Can reproduce on [177ab0e](https://github.com/Chatterino/chatterino2/commit/177ab0eaa7f65a49c002ba9b02a9ff3f65659d9d). It's also possible to repro this bug with middle click scrolling

That's what I thought a while ago but it's not noticeable since most messages are <2 lines. Just tested with some emote spam and can confirm it also occurs in standard channels.

jupjohn:
Coming back to this as it's more noticeable ever since 7TV emote support was merged. The whole scrollbar implementation would need to change and I don't think I have the brains to figure that one out. 

[...]

PR #5524, opened 2024-07-26
Handle panning touch gestures

camporter:
This adds support for the panning gesture on `ChannelView`. This lets users who have touchscreens perform a two-finger pan of the chat (effectively a scroll).

Mouse move events aren't handled while panning is occurring so that erroneous text selection when going into and out panning is avoided.

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -54,6 +54,7 @@
 #include <QDebug>
 #include <QDesktopServices>
 #include <QEasingCurve>
+#include <QGestureEvent>
 #include <QGraphicsBlurEffect>
 #include <QJsonDocument>
 #include <QMessageBox>
@@ -368,6 +369,8 @@ ChannelView::ChannelView(InternalCtor /*tag*/, QWidget *parent, Split *split,
     QObject::connect(&this->scrollTimer_, &QTimer::timeout, this, [this] {
         this->scrollUpdateRequested();
     });
+
+    this->grabGesture(Qt::PanGesture);
 
     // TODO: Figure out if we need this, and if so, why
     // StrongFocus means we can focus this event through clicking it
@@ -1786,8 +1789,48 @@ void ChannelView::leaveEvent(QEvent * /*event*/)
     this->unpause(PauseReason::Mouse);
 }
 
+bool ChannelView::event(QEvent *event)
+{
+     if (event->type() == QEvent::Gesture)
+     {
+         return gestureEvent(dynamic_cast<QGestureEvent*>(event));
+     }
+
+    return BaseWidget::event(event);
+}
+
+bool ChannelView::gestureEvent(const QGestureEvent *event)
+{
+     if (QGesture *pan = event->gesture(Qt::PanGesture))
+     {
+         auto const *gesture = dynamic_cast<QPanGesture *>(pan);
+
+         switch (gesture->state())
+         {
+             case Qt::GestureStarted:
+             case Qt::GestureUpdated:
+                 this->isPanning_ = true;
+                 this->clearSelection();
+                 break;
+             default:
+                 this->isPanning_ = false;
+                 break;
+         }
+
+         this->scrollBar_->offset(-gesture->delta().y() * 0.5);
+     }
+
+     return false;
+}
+
 void ChannelView::mouseMoveEvent(QMouseEvent *event)
 {
+    if (this->isPanning_)
+    {
+        // Don't do any text selection, hovering, etc while panning
+        return;
+    }
+
     /// Pause on hover
     if (float pauseTime = getSettings()->pauseOnHoverDuration;
         pauseTime > 0.001F)
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1819,6 +1819,8 @@ bool ChannelView::gestureEvent(const QGestureEvent *event)
                     {
                         this->disableScrolling();
                     }
+
+                    return true;
                 }
                 break;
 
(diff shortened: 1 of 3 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -1786,8 +1789,48 @@ void ChannelView::leaveEvent(QEvent * /*event*/)
     this->unpause(PauseReason::Mouse);
 }
 
+bool ChannelView::event(QEvent *event)
+{
+     if (event->type() == QEvent::Gesture)
+     {
+         return gestureEvent(dynamic_cast<QGestureEvent*>(event));
```
Nerixyz, line 1796:
```suggestion
         return this->gestureEvent(dynamic_cast<QGestureEvent*>(event));
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -1786,8 +1789,48 @@ void ChannelView::leaveEvent(QEvent * /*event*/)
...
+     }
+
+     return false;
+}
+
 void ChannelView::mouseMoveEvent(QMouseEvent *event)
 {
+    if (this->isPanning_)
+    {
+        // Don't do any text selection, hovering, etc while panning
```
Nerixyz, line 1830: How does this interact with `mousePressEvent`? Does `mousePressEvent` get called before? If so, it's worth checking that its state gets "reset", as it could set e.g. `isLeftMouseDown_`.

Nerixyz (commented):
Thank you for looking into this! Since I don't have a touchscreen, it's hard to review. Some thoughts:

- How does text selection work on a touchscreen now? At least on mobile, it's common to have users hold for a short amount of time for the UI to transition into "text-selection" mode. I don't know how that's handled in web browsers on desktop, but that should be a source of inspiration.

For the future:

- It could be nice to be able to "flick" though tabs by swiping left and right (maybe restricted to swipes originating from the edges). This would also be nice to have for touchpads on laptops, since browsers allow that.
- Scrolling on a graphics tablet should also be possible, but that requires a bit more effort (especially since our scrollbar isn't a traditional one - https://github.com/Chatterino/chatterino2/issues/3260).

src/widgets/helper/ChannelView.cpp
```diff
@@ -1786,8 +1789,48 @@ void ChannelView::leaveEvent(QEvent * /*event*/)
...
+     }
+
+    return BaseWidget::event(event);
+}
+
+bool ChannelView::gestureEvent(const QGestureEvent *event)
+{
+     if (QGesture *pan = event->gesture(Qt::PanGesture))
+     {
+         auto const *gesture = dynamic_cast<QPanGesture *>(pan);
```
pajlada, line 1806:
```suggestion
         const auto *gesture = dynamic_cast<QPanGesture *>(pan);
```

additionally, you should probably sanity-check that `gesture` is not a `nullptr` since we've used `dynamic_cast`

src/widgets/helper/ChannelView.cpp
```diff
@@ -1786,8 +1789,48 @@ void ChannelView::leaveEvent(QEvent * /*event*/)
...
+
+         switch (gesture->state())
+         {
+             case Qt::GestureStarted:
+             case Qt::GestureUpdated:
+                 this->isPanning_ = true;
+                 this->clearSelection();
+                 break;
+             default:
+                 this->isPanning_ = false;
```
pajlada, line 1816: I assume this is here to handle `Qt::GestureFinished` or `Qt::GestureCancelled` - I'd rather we handle those two states explicitly

pajlada (commented):
Overall looks pretty good, will test this out on my touchscreen and give some more feedback.
Thank you for looking into this!

pajlada:
Just tested this out on my 4k touchscreen, it works fine but I feel like the speed at which it scrolls is wrong. This is most likely mostly related to #3262 that Nerixyz already mentioned above.

src/widgets/helper/ChannelView.cpp
camporter: Yes these should be clearer now.

camporter:
> How does text selection work on a touchscreen now? At least on mobile, it's common to have users hold for a short amount of time for the UI to transition into "text-selection" mode. I don't know how that's handled in web browsers on desktop, but that should be a source of inspiration.

Thanks for reviewing, text selection is handled with one finger movement on a touch screen, since Qt by default does a mouse click and move with it.

> It could be nice to be able to "flick" though tabs by swiping left and right (maybe restricted to swipes originating from the edges). This would also be nice to have for touchpads on laptops, since browsers allow that.

I tried working with some of the swipe gesture support, but it conflicts with the normal mouse event handling. Swiping would be nice, but it seems like it would look and feel better with a dedicated animation, etc.

I also experimented with pinch to zoom the UI, but it feels a bit ugly. Not sure if it's specific to Qt or if there's something I'm doing wrong. Left it off this review.

> Scrolling on a graphics tablet should also be possible, but that requires a bit more effort (especially since our scrollbar isn't a traditional one - [Scrolling should behave more like browsers #3260](https://github.com/Chatterino/chatterino2/issues/3260)).

Yeah, panning on the emote menus with this has the same issues as scrolling normally :frowning_face: 

> Just tested this out on my 4k touchscreen, it works fine but I feel like the speed at which it scrolls is wrong. This is most likely mostly related to https://github.com/Chatterino/chatterino2/issues/3262 that Nerixyz already mentioned above.

I adjusted the pan scaling down a bit more, so it should feel a bit more natural now hopefully. The scroll speed setting is ignored for panning (unless I missed something). Would guess you wouldn't want the scroll speed to affect panning, since panning is expected to match finger movements as close as possible.

src/widgets/helper/ChannelView.cpp
camporter: I tried changing the mouse press and release event handling while panning, but it caused some selection issues when doing mouse movement after panning.

pajlada (approved):
Pushed two small changes:
 - ran formatter & made the switch look a bit more like the ones we normally write
 - added `return true` to the handled gesture to indicate to qt that we've handled the event, and it does not need to be propagated further. I have not tested this but will assume it works

PR #5525, opened 2024-07-27
fix: sort elements after RTL reordering

Nerixyz:
Since selection depends on elements being in order, this PR sorts the elements after reordering. This can be expensive, but allows selection of the text. As mentioned in the changelog entry, selection is still wrong, but now it's consistently wrong, which is an improvement.

It might not be too hard to get a more correct selection for RTL messages now (i.e. one that is "aligned" to the right instead of the left), but it's not too important for me.


pajlada (approved):
Nice! This feels like a good start, even if it's wrong, at least it's wrong in a nicer way

PR #5527, opened 2024-07-30
fix: replace defines with constexpr/const and use more absolute paths for includes

Nerixyz:
- Replaces constant `#define`s with `constexpr` where possible. In case a `QString` is desired, they're replaced with a static `QString`.
	- For concatenations, QStringBuilder is required. This is included with QString in Qt 6 but requires an include in Qt 5
- Uses paths relative to `src/` for all includes. Check with `rg -P '^#include "(?!Application|BrowserExtension|RunGui|ForwardDecl)[^"/]+"' -g 'src/**/*.{cpp,hpp}'`.

### DIFF
--- a/src/common/Credentials.cpp
+++ b/src/common/Credentials.cpp
@@ -11,6 +11,7 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QSaveFile>
+#include <QStringBuilder>
 
 #include <variant>
 
--- a/src/common/Credentials.cpp
+++ b/src/common/Credentials.cpp
@@ -25,6 +25,7 @@
 #    else
 #        include "keychain.h"
 #    endif
+#endif
 
 namespace {
 
@@ -38,19 +39,19 @@ QString formatName(const QString &provider, const QString &name)
 
 bool useKeyring()
 {
-#    ifdef NO_QTKEYCHAIN
+#ifdef NO_QTKEYCHAIN
     return false;
-#    endif
+#endif
     if (Modes::instance().isPortable)
     {
         return false;
     }
 
-#    ifdef Q_OS_LINUX
+#ifdef Q_OS_LINUX
     return getSettings()->useKeyring;
-#    else
+#else
     return true;
-#    endif
+#endif
 }
 
 // Insecure storage:
@@ -116,7 +117,7 @@ std::queue<Job> &jobQueue()
 
 void runNextJob()
 {
-#    ifndef NO_QTKEYCHAIN
+#ifndef NO_QTKEYCHAIN
     auto &&queue = jobQueue();
 
     if (!queue.empty())
@@ -153,7 +154,7 @@ void runNextJob()
 
         queue.pop();
     }
-#    endif
+#endif
 }
 
 void queueJob(Job &&job)
@@ -188,7 +189,7 @@ void Credentials::get(const QString &provider, const QString &name_,
 
     if (useKeyring())
     {
-#    ifndef NO_QTKEYCHAIN
+#ifndef NO_QTKEYCHAIN
         // if NO_QTKEYCHAIN is set, then this code is never used either way
         auto *job = new QKeychain::ReadPasswordJob("chatterino");
         job->setAutoDelete(true);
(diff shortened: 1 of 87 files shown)


src/common/Credentials.cpp
```diff
@@ -11,31 +11,24 @@
...
 #include <variant>
 
 #ifndef NO_QTKEYCHAIN
-#    ifdef CMAKE_BUILD
-#        if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-#            include "qt6keychain/keychain.h"
-#        else
-#            include "qt5keychain/keychain.h"
-#        endif
-#    else
```
pajlada, line 24: this breaks my build

src/common/Version.cpp
```diff
@@ -10,8 +10,31 @@ namespace chatterino {
...
+ *  - 2.4.0-beta.3
+ *  - 2.4.0-beta.2
+ *  - 2.4.0-beta
+ *
+ * Alpha:
+ *  - 2.4.0-alpha.3
+ *  - 2.4.0-alpha.2
+ *  - 2.4.0-alpha
+ **/
+constexpr QStringView CHATTERINO_VERSION = u"2.5.1";
```
pajlada, line 34: Why not keep this defined in the header and continue to use `CHATTERINO_VERSION` everywhere?

src/widgets/helper/CommonTexts.hpp
```diff
@@ -1,8 +1,13 @@
...
 
-#define OPEN_IN_BROWSER "Open stream in browser"
-#define OPEN_PLAYER_IN_BROWSER "Open player in browser"
-#define OPEN_MOD_VIEW_IN_BROWSER "Open mod view in browser"
-#define OPEN_IN_STREAMLINK "Open in streamlink"
-#define DONT_OPEN "Don't open"
-#define OPEN_WHISPERS_IN_BROWSER "Open whispers in browser"
+#include <QString>
+
+inline const QString OPEN_IN_BROWSER = QStringLiteral("Open stream in browser");
```
pajlada, line 5: do we avoid using the `u"..."_s` syntax here to keep the header file simpler?

src/widgets/helper/NotebookButton.cpp
```diff
@@ -14,8 +14,6 @@
 #include <QPainterPath>
 #include <QRadialGradient>
 
-#define nuuls nullptr
```
pajlada, line 17: ?

src/widgets/settingspages/IgnoresPage.cpp
```diff
@@ -19,12 +20,14 @@
 #include <QTableView>
 #include <QVBoxLayout>
 
+namespace chatterino {
+
+using namespace literals;
+
 // clang-format off
-#define INFO "/block <user> in chat blocks a user.\n/unblock <user> in chat unblocks a user.\nYou can also click on a user to open the usercard."
+inline const QString INFO = u"/block <user> in chat blocks a user.\n/unblock <user> in chat unblocks a user.\nYou can also click on a user to open the usercard."_s;
```
pajlada, line 28: Can just inline this in `addUsersTab` instead, like you've done in HighlightingPage

pajlada (commented):
small incomplete review

src/common/Version.cpp
Nerixyz: Done.

src/widgets/helper/CommonTexts.hpp
Nerixyz: We can't use it in header files, as we'd need to add a `using` directive - effectively making the operator `""_s`  available everywhere.

src/widgets/helper/NotebookButton.cpp
Nerixyz: This was never used.

pajlada: sad

Nerixyz: I mean I can add it back

pajlada: No it's ok, just end of an era

Mm2PL: sadge

pajlada (approved):

PR #5353, opened 2024-04-22
fix(notebook): align actions and hotkeys

Nerixyz:
Aligns <kbd>Ctrl</kbd>+<kbd>U</kbd> and "Toggle visibility of tabs" to do the same. The hotkey will now trigger the same `QAction`. I've done the same for <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd> and removed the visibility filter reset when the hotkey is activated with `on` or `off`.

Fixes #5341.

PR #5530, opened 2024-08-04
fix: tristate toggle logic for tab visibilty

### LINKED ISSUE #5341: Toggle visibility of tabs hotkey behaves differently than right click when used with Show live tabs only
### Checklist

### Describe your issue

When using Show live tabs only, using right click to toggle visibility of tabs will toggle back to only showing live channels, but when using the Ctrl+U hotkey it toggles back to showing all tabs regardless of live or not.

### Screenshots

right click example:

https://github.com/Chatterino/chatterino2/assets/62087307/5e0b3ca2-f73a-4019-8cf9-1a9e91194088

hotkey example:

https://github.com/Chatterino/chatterino2/assets/62087307/6e1212f9-c3c6-458a-a7c7-a255ecdaff11

### OS and Chatterino Version

Chatterino Nightly 2.5.0 (commit 3aead093) built on 2024-04-21 with Qt 6.5.0, MSVC 193833135 Running on Windows 11 Version 23H2, kernel: 10.0.22631

pajlada:
This removes the "toggle live" option for tab visibilty options, including the hotkey. so this tries to remove the hotkey if it hasn't been touched by the user

This changes the tab visibilty options in the menu from "toggle" & "toggle live" to "show all", "show live only", and "hide all" in a submenu

This removes all tab visibility options for the emote popup

For previous context, see https://github.com/Chatterino/chatterino2/pull/5353

The existing toggle option will reset the `live only` filter, and switch between "hide all" and "show all"

Fixes #5341


Nerixyz (commented):
It would probably be good to know if the removed hotkey was used.

This PR should also mention https://github.com/Chatterino/chatterino2/issues/5341. One thing that gets lost is the visibility of the toggle hotkey inside the context menu.

Now the state graph looks like this:

```mermaid
graph TD;
    All-->|liveOnly|LiveOnly
    LiveOnly-->|on|All
    LiveOnly-->|toggle|None
    LiveOnly-->|liveOnly|LiveOnly
    LiveOnly-->|off|None
    All<-->|toggle|None
    All-->|off|None
    All-->|on|All
    None-->|off|None
    None-->|liveOnly|LiveOnly
```

src/controllers/hotkeys/HotkeyController.cpp
```diff
@@ -524,6 +523,14 @@ void HotkeyController::resetToDefaults()
     this->loadHotkeys();
 }
 
+void HotkeyController::removeDeprecatedHotkeys()
+{
+    // The "toggleLiveOnly" argument was removed 2024-08-04
+    this->tryRemoveDefault(HotkeyCategory::Window, QKeySequence("Ctrl+Shift+L"),
+                           "setTabVisibility", {"toggleLiveOnly"},
+                           "toggle live tabs only");
+}
```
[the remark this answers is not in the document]
pajlada: Fixed in https://github.com/Chatterino/chatterino2/pull/5530/commits/ffbc2cf19f833575601c77102291d045895b3ce0 by renaming the function to `clearRemovedDefaults` and adding the same note from `tryRemoveDefault`

src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -326,15 +326,16 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
...
-              .possibleArguments{{"Toggle", {}},
-                                 {"Set to on", {"on"}},
-                                 {"Set to off", {"off"}},
-                                 {"Live only on", {"liveOnly"}},
-                                 {"Live only toggle", {"toggleLiveOnly"}}},
+              .possibleArguments{
+                  {"Toggle", {}},
+                  {"Set to on", {"on"}},
+                  {"Set to off", {"off"}},
+                  {"Live only on", {"liveOnly"}},
```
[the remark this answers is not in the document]
pajlada: Should be fixed in [`8163a7b` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/8163a7bbea3916fb8a060a9d077cf33da542665a) - I've renamed the options to better match the menu options

pajlada:
> It would probably be good to know if the removed hotkey was used.

Anecdotally, I've only heard of people using the removed hotkey accidentally and wondering why some of their tabs are missing

src/widgets/Window.cpp
```diff
@@ -642,31 +642,38 @@ void Window::addShortcuts()
...
              {
-                 this->notebook_->setShowTabs(true);
-                 getSettings()->tabVisibility.setValue(
-                     NotebookTabVisibility::LiveOnly);
+                 this->notebook_->onlyShowLiveTabsAction->trigger();
              }
              else if (arg == "toggleLiveOnly")
              {
-                 this->notebook_->toggleOfflineTabs();
+                 // NOOP: Removed 2024-08-04
```
Mm2PL, line 661: This should probably link the PR

src/controllers/hotkeys/HotkeyController.cpp
```diff
@@ -524,6 +523,14 @@ void HotkeyController::resetToDefaults()
     this->loadHotkeys();
 }
 
+void HotkeyController::clearRemovedDefaults()
+{
+    // The "toggleLiveOnly" argument was removed 2024-08-04
```
Mm2PL, line 528: I think the user should be somehow notified of this change on update. This also does not account for people who rebound it. If you don't want a message box when starting the program for everybody. IMO it would be nice to notify people who changed the default setting, maybe with a message box or a new element in the hotkey settings:

chatterino-manual-config-migration

This cannot be a thing merged after this PR because then we completely break users' configs from one nightly to another. I can see it being a PR merged before this one and this one being made dependent on that.

src/widgets/Window.cpp
```diff
@@ -642,31 +642,38 @@ void Window::addShortcuts()
...
+                 auto *selectedPage = this->notebook_->getSelectedPage();
+                 if (selectedPage != nullptr)
+                 {
+                     auto *selectedSplit = selectedPage->getSelectedSplit();
+                     if (selectedSplit != nullptr)
+                     {
+                         auto selectedChannel = selectedSplit->getChannel();
+                         if (selectedChannel != nullptr)
+                         {
+                             selectedChannel->addSystemMessage(
```
Mm2PL, line 671: IMO this should return a hotkey error string, not add a system message to the channel. Like the code below

src/widgets/Window.cpp
```diff
@@ -642,31 +642,38 @@ void Window::addShortcuts()
...
+                                 "setTabVisibility's toggleLiveOnly has been "
+                                 "removed");
+                         }
+                     }
+                 }
```
Mm2PL, line 676: Update the code below to no longer include toggleLiveOnly.

Mm2PL (dismissed):

src/widgets/Window.cpp
pajlada: :+1: 
Resolved in [`b3f27d8` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/b3f27d89d28d8a5014472aa2a412e74011533b29)

src/widgets/Window.cpp
pajlada: :+1:
Fixed in [`d7ba3a0` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/d7ba3a0a5366532579e2d0003d467dc6f87fce2a)

src/widgets/Window.cpp
pajlada: :+1: 
Fixed in [`8b5c532` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/8b5c53234f599cc8bb5ca0a579a6dda33b5aea1e)

src/controllers/hotkeys/HotkeyController.cpp
pajlada: Effort has been made to make this more apparent for users in [`ea642bb` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/ea642bb19b77540211bc09d504807f36859e577e) and [`23339d1` (#5530)](https://github.com/Chatterino/chatterino2/pull/5530/commits/23339d1d128f41c260fc5fe605abb772ce3d7923).

It doesn't highlight the rows themselves since it's a lot of code that's very messy to update, I'd rather that be part of another PR.

It currently looks like this:
2024-08-10-152840_915x600_scrot

It's not pretty, but I think it gets the job done.
I chose yellow/the position because that's how it's done for duplicate Commands.

pajlada (commented):
image

KogasaPls:
> > It would probably be good to know if the removed hotkey was used.
> 
> Anecdotally, I've only heard of people using the removed hotkey accidentally and wondering why some of their tabs are missing

Also anecdotally, but I used it (hence why I originally added it) and wondered why it was removed :)

I do think the configuration state graph was a bit confusing, it should never have been part of a "tab visibility" option with a toggle already on it. It should rather be two toggles, "hide tabs" and "show live tabs only". I only used the latter functionality, which now requires two separate hotkeys (one to "show live tabs only" and another to "show all tabs"). I used it to hide a large number of offline tabs while I'm browsing live channels, and reveal them when I'm browsing offline chats.

Anyway, just providing some context, not complaining. Take care.

PR #5532, opened 2024-08-04
refactor: move seventv/bttv event apis to application

pajlada:
- **Gracefully quit streamer mode**
- **Set Application::instance to nullptr in dtor, allow us to better catch usages of getApp() after Application is destroyed**
- **move SeventvEventAPI from TwitchIrcServer to Application**
- **move BttvLiveUpdates to Application**


pajlada (commented):
image

PR #5535, opened 2024-08-08
refactor: move "delete old update files" logic to Updates

pajlada:
This PR moves the "delete old update files" logic from RunGui to Updates, and uses `QtConcurrent::run` instead of an `std::thread`

I have purposefully not added a platform-check to this yet


pajlada:
image

PR #5537, opened 2024-08-10
fix: cleanly exit on shutdown

pajlada:
- **Use Qt's lifetime system for the tab completion model**
- **Exit cleanly, but leak 7tv eventapi**
- **add renameThread function**
- **rename basicpubsubmanager thread**
- **implement stopImpl interface for basicpubsubclient**
- **implement pubsub stop thing for 7tv's pubsub client**
- **Add changelog entry**

This is the final step of our exit-refactor, and uses work from @Mm2PL 's https://github.com/Chatterino/chatterino2/pull/4993

### DIFF
--- a/src/singletons/StreamerMode.cpp
+++ b/src/singletons/StreamerMode.cpp
@@ -238,7 +238,7 @@ void StreamerModePrivate::start()
 StreamerModePrivate::~StreamerModePrivate()
 {
     this->thread_.quit();
-    this->thread_.wait(50);
+    this->thread_.wait(500);
 }
 
 bool StreamerModePrivate::isEnabled() const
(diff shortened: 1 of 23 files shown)


pajlada (commented):
image

I tested on macOS & Linux and it works as expected

Windows was tested by @Nerixyz and @brian6932

brian6932 (approved):
Tested on Windows, and works as expected

src/singletons/StreamerMode.cpp
```diff
@@ -194,9 +202,15 @@ bool StreamerMode::isEnabled() const
...
 
+void StreamerMode::start()
+{
+    this->private_->start();
+}
+
 StreamerModePrivate::StreamerModePrivate(StreamerMode *parent)
     : parent_(parent)
 {
+    this->thread_.setObjectName("StreamerMode");
```
Mm2PL, line 213: This could use `renameThread`. Would be nice to only use a single mechanism to name threads

Mm2PL (changes requested):

Mm2PL (approved):

PR #5538, opened 2024-08-10
refactor: rename miniaudio thread

pajlada:
- **use ref instead of doubleref for renameThread**
- **rename miniaudio thread name**
- **add changelog entry**


pajlada (commented):
image

PR #5539, opened 2024-08-10
chore: rename threads on Windows too

Nerixyz:
Additionally, this adds a few names for threads not yet renamed. Names on Windows don't have tight restrictions on the length. On macOS, this would need to be done from inside the thread, as it doesn't have the first argument to `pthread_setname_np`.

Code_2024-08-10_15-22-10

### DIFF
--- a/src/util/RenameThread.cpp
+++ b/src/util/RenameThread.cpp
@@ -2,14 +2,31 @@
 
 #include "common/QLogging.hpp"
 
+#include <QOperatingSystemVersion>
+
 #ifdef Q_OS_WIN
+
 #    include <Windows.h>
-#endif
 
 namespace chatterino::windows::detail {
 
 void renameThread(HANDLE hThread, const QString &threadName)
 {
+#    if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)  // Qt 6 requires Windows 10 1809
+    // Windows 10, version 1607
+    constexpr QOperatingSystemVersion minVersion{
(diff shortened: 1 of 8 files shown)


src/util/RenameThread.hpp
```diff
@@ -7,13 +7,25 @@
...
 
+#ifdef Q_OS_WIN
+using HANDLE = void *;
+#endif
+
 namespace chatterino {
 
+#ifdef Q_OS_WIN
+namespace windows::detail {
+    void renameThread(HANDLE hThread, const QString &name);
```
pajlada, line 18: change to void* here instead

src/util/RenameThread.cpp
```diff
@@ -0,0 +1,22 @@
...
+void renameThread(HANDLE hThread, const QString &threadName)
+{
+    auto hr = SetThreadDescription(hThread, threadName.toStdWString().c_str());
+    if (!SUCCEEDED(hr))
+    {
+        qCWarning(chatterinoCommon).nospace()
+            << "Failed to set thread description, hresult=0x"
+            << QString::number(hr, 16);
+    }
+}
```
pajlada, line 20: needs to be if-deffed - doesn't build on non-windows platforms

and you wanted to check if you need to do a win7 or something check too

pajlada (commented):
small nits, otherwise :+1:

pajlada (approved):

PR #5547, opened 2024-08-17
Remove experimental IRC support

pajlada:
https://github.com/Chatterino/chatterino2/discussions/5498


Mm2PL:
Maybe I wasn't clear enough that I didn't want this to happen until we could migrate IRC to plugins. Maybe building all this back from the ground won't be that terrible Clueless

PR #5548, opened 2024-08-17
refactor: make a single MessageBuilder

pajlada:
This removes TwitchMessageBuilder and SharedMessageBuilder, and moves all relevant functionality into MessageBuilder

MessageBuilder is chunky, but it's at least in one file now instead of spread out with little reason.

There's a few more things that could probably be done here, like making sure we make better use of static functions for message makers so consumers don't have to set message flags themselves.

A common pattern I still see is:
```c++
MessageBuilder builder(foo, bar);
builder.flags = asd;
auto msg = builder.release();
```

where we could just have a static function instead
```c++
auto msg = MessageBuilder::buildFooBar(foo, bar);
```

with an alternative parameter adding extra flags for ones that are a bit more dynamic
```c++
auto msg = MessageBuilder::buildFooBar(foo, bar, {ExtraFlag});
```


pajlada:
image

PR #5551, opened 2024-08-18
refactor: some more Application refactors

pajlada:
specifically this tries to move some responsibility of signal/conenction stuff from Application to WindowManager


pajlada:
image

PR #5558, opened 2024-08-24
fix: Fixed account switch not being saved if no other settings were changed

pajlada:
Previously, if a user changed their user through the account switch dialog but didn't change any settings, we never saved this until the end of the application. If the application shut down uncleanly, that means the user would have to change account again on next startup.

With this PR, we now request a save of settings when the user changes their account.


pajlada (commented):
image

PR #5560, opened 2024-08-25
fix: remove deprecated `Application::getTwitchAbstract`

pajlada:
- **fix: remove deprecated `getTwitchAbstract`**


pajlada:
image

PR #5562, opened 2024-08-25
Build for Ubuntu 24.04

Wissididom:
I wanted to check which dependencies can be removed, but I wasn't able to reliably tell which dependencies are necessary and which aren't, so I thought I'd just PR it without that.

I also removed a few duplicate dependencies in that dependency lists.

It should still be the same state as when it worked before and still work, but I'll test again, when CI is finished building (the .deb file).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,15 @@ jobs:
             skip-crashpad: false
             build-appimage: true
             build-deb: true
+          - os: ubuntu-24.04
+            container: ghcr.io/chatterino/chatterino2-build-ubuntu-24.04:latest
+            qt-version: 6.6.1
+            force-lto: false
+            plugins: true
+            skip-artifact: false
+            skip-crashpad: false
+            build-appimage: false
+            build-deb: true
     env:
       C2_ENABLE_LTO: ${{ matrix.force-lto }}
       C2_PLUGINS: ${{ matrix.plugins }}
--- a/.CI/CreateUbuntuDeb.sh
+++ b/.CI/CreateUbuntuDeb.sh
@@ -12,7 +12,7 @@ install_prefix="appdir/usr"
 # The directory we finally pack into our .deb package
 packaging_dir="package"
 
-# Get the Ubuntu Release (e.g. 20.04 or 22.04)
+# Get the Ubuntu Release (e.g. 20.04, 22.04 or 24.04)
 ubuntu_release="$(lsb_release -rs)"
 
 # The final path where we'll save the .deb package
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -49,6 +49,15 @@ jobs:
...
             build-appimage: true
             build-deb: true
+          - os: ubuntu-24.04
+            container: ghcr.io/chatterino/chatterino2-build-ubuntu-24.04:latest
+            qt-version: 6.6.1
+            force-lto: false
+            plugins: true
+            skip-artifact: false
+            skip-crashpad: false
+            build-appimage: false
```
Nerixyz, line 59: I don't know how AppImages work exactly, but couldn't we build the AppImage on 24.04 now?

.github/workflows/build.yml
```diff
@@ -49,6 +49,15 @@ jobs:
             skip-crashpad: false
             build-appimage: true
             build-deb: true
+          - os: ubuntu-24.04
+            container: ghcr.io/chatterino/chatterino2-build-ubuntu-24.04:latest
+            qt-version: 6.6.1
```
Nerixyz, line 54: (not for this PR and probably for Chatterino/docker): We could bump this to 6.7.2 (as well as the Windows version).

Nerixyz (commented):
I can't comment there, but the artifact is not yet downloaded in the `create-release` job.

.github/workflows/build.yml
Wissididom: I don't know either, but I disabled it because I had conflicts with the filename in CI and iirc pajlada said something on stream about disabling it.

.github/workflows/build.yml
Wissididom: Actually I also thought that, but wasn't sure if there was a specific reason for it. I think it would need to be adjusted in both Chatterino/docker and Chatterino/chatterino2 (See https://github.com/Chatterino/docker/blob/master/Dockerfile_chatterino2-build-ubuntu-24.04#L97)

.github/workflows/build.yml
pajlada: The reason I don't want to "just" bump the AppImage is that generally, the AppImage is built on older versions since that ensures it runs on older distros due to some glibc magic. We build it in 22.04 instead of 20.04 because of 22.04 using qt6 before and 20.04 used qt5, and we didn't want to distribute qt5 there

.github/workflows/build.yml
pajlada: Can definitely be done, just needs testing on the various distros

PR #5570, opened 2024-08-28
refactor: deserialize theme automatically

Nerixyz:
No more macros. Uses [Boost.PFR](https://www.boost.io/doc/libs/1_86_0/doc/html/boost_pfr.html) to get the names of fields. Hopefully this works with all compilers.


Nerixyz:
Well... seems like we need to up the requirements on compilers. All compilers support this, but their requirement is too low. MSVC fails because I didn't update the boost version.

PR #5573, opened 2024-08-30
fix(streamlink): set working-directory to streamlink dir

pajlada:


pajlada:
solution is a bit more complex.
we could set the working directory to a temporary dir, but it feels jank

PR #5578, opened 2024-08-31
fix: some tooltips not being readable

pajlada:
our "budget fusion palette" had the same color for the tooltip
background as the tooltip text


pajlada:
image

PR #5585, opened 2024-09-08
feat: add "tags" or "keyword" support for settings

pajlada:
This will create more widgets per setting, but the goal is to more simply allow us to add various free text to settings that need them.

I don't plan on migrating all settings, but I migrated all the enum dropdowns so I removed that implementation.

It's not really a builder pattern but it tries to be similar to it


pajlada:
image

PR #5588, opened 2024-09-09
chore: ensure `static`s are only present once in the final app

Nerixyz:
`static` shouldn't be used for constants (nor functions) in headers, as this causes them to be compiled _separately_ into every TU they're included in. For small constants, it's not that bad, but for more complex types, this requires initialization for each of these instances (e.g. a `QString` needs to be constructed by calling into the respective library - LTO can't eliminate this). These constants should be defined as `inline`, meaning they will essentially be "merged" into a single constant when linking (`inline` doesn't really mean that the compiler will inline them).

Even though `constexpr` implies `inline` for functions, it doesn't imply `inline` for constants (these will default to `static`).


src/providers/twitch/TwitchChannel.hpp
```diff
@@ -32,17 +32,11 @@ namespace chatterino {
 // instead.
 // See https://github.com/Chatterino/chatterino2/issues/3384 and
 // https://mm2pl.github.io/emoji_rfc.pdf for more details
-const QString ZERO_WIDTH_JOINER = QString(QChar(0x200D));
+inline const QString ZERO_WIDTH_JOINER = QStringLiteral("\u200D");
```
pajlada, line 35: Leave the changes from this file out until we have unit tests added for this RFC

pajlada (approved):
Looks good, just the one note about the emoji RFC-thing

PR #5594, opened 2024-09-14
refactor: move zero width replacement to a function

Nerixyz:
Refactors and tests the Unicode zero-width-joiner replacement. The replacement is now in a helper function (so we can test it) and the individual components use Unicode string literals. (From #5588).


pajlada (commented):
I would prefer if this was split into two PRs (or at the very least commits with tests being allowed to fully run)
1. Moving the logic to a function, and adding tests for it
2. Change way we encode the characters

pajlada (approved):

Mm2PL (approved):

PR #5595, opened 2024-09-15
ci: set test cmake build job count to logical processor count

jupjohn:
This PR aims to increase (depending on hardware) the amount of parallel build jobs used when building chatterino inside of the ubuntu test suite. Currently it's running a single build job, which partially explains the long run times.


Nerixyz:
Good catch! We could probably use (s)ccache for the tests and maybe even the builds. From https://github.com/Chatterino/chatterino2/actions/caches, it looks like we have some remaining space (the caches could be limited to 100 MiB). Getting results early is really valuable imo.

jupjohn:
Sure, I'll implement that if it's simple enough

pajlada (approved):

PR #5598, opened 2024-09-16
test: add snapshot tests for MessageBuilder

Nerixyz:
I want to refactor the `MessageBuilder` for IRC messages, but it's an integral part of Chatterino, so I'm adding some tests first. These test that the output message is the same. Adding tests requires the following steps (also documented in the `MessageBuilder` test (at the top).

1. Add an IRC message in `tests/snapshots/MessageBuilder/IRC` as a new JSON file (`{ "input": "<IRC-msg>" }`) and save it
2. Set `UPDATE_SNAPSHOTS` in `tests/src/MessageBuilder.cpp` to `true`
3. In the same file, add the filename from step 1 without the extension to `IRC_SNAPSHOTS`
4. Run the new test (or run all message-builder tests with `ctest -R TestMessageBuilderP --output-on-failure`). Don't worry about the integrity test for now.
5. Set `UPDATE_SNAPSHOTS` back to `false`
6. Ensure the output is stable by running the message-builder tests again `ctest -R TestMessageBuilderP --output-on-failure` (everything should pass now).

On Windows, you might see that all snapshot files will show as modified. However, after staging them (`git add`), they shouldn't show up as modified anymore (as just the line-endings changed).

Coverage from this PR: https://app.codecov.io/github/Chatterino/chatterino2/pull/5598


src/providers/twitch/TwitchChannel.cpp
```diff
@@ -462,6 +462,12 @@ void TwitchChannel::addChannelPointReward(const ChannelPointReward &reward)
     }
 }
 
+void TwitchChannel::addKnownChannelPointReward(const ChannelPointReward &reward)
+{
+    auto channelPointRewards = this->channelPointRewards_.access();
```
pajlada, line 467: Assert isTest

tests/src/MessageBuilder.cpp
```diff
@@ -627,3 +988,197 @@ TEST_F(TestMessageBuilder, IgnoresReplace)
...
+        addPhrase(u"&\\[ (\\w+) \\]&"_s, true, false, u"{ \\1 }"_s, false);
+        addPhrase(u"BLOCK"_s, false, true, u"?"_s, true);
+        addPhrase(u"block!{2,}"_s, true, true, u"?"_s, true);
+        // empty
+        addPhrase(u""_s, false, false, u"empty"_s, true);
+        // invalid regex
+        addPhrase(u"("_s, true, false, u"invalid"_s, true);
+        // too many iterations
+        addPhrase(u"(?<=infinite-loop)$"_s, true, false, u"infinite-loop"_s,
+                  true);
```
pajlada, line 1031: I'd prefer if these are instead read as part of settings, if that's possible

pajlada (changes requested):
Partial review, haven't taken a full look
Overall looks good and will definitely bring confidence in changing the MessageBuilder logic going forward

pajlada (approved):

PR #5600, opened 2024-09-20
feat: include more data when copying messages as JSON

Nerixyz:
From #5598. This PR extends `Message` to have a `toJson` method which serializes all members of a message (except for `parseTime` in tests). It's useful to have this while debugging message parsing. If binary-size is an issue, this could be behind a `#define`.


src/messages/MessageElement.hpp
```diff
@@ -183,10 +186,14 @@ class MessageElement
     virtual void addToContainer(MessageLayoutContainer &container,
                                 MessageElementFlags flags) = 0;
 
+    virtual QJsonObject toJson() const = 0;
```
pajlada, line 189: We could make toJson non-pure and have derived classes call the base toJson instead of jsonBase

src/widgets/helper/ChannelView.cpp
```diff
@@ -256,40 +256,14 @@ void addHiddenContextMenuItems(QMenu *menu,
...
-        }
-
-        jsonDocument.setObject(jsonObject);
-
-        auto jsonString =
-            jsonDocument.toJson(QJsonDocument::JsonFormat::Indented);
-
-        menu->addAction("Copy message &JSON", [jsonString] {
-            crossPlatformCopy(jsonString);
+        menu->addAction("Copy message &JSON", [message] {
```
pajlada, line 263: The menu is deleted when it's closed right? If it is, this is fine

Nerixyz: Yes, delete-on-close is set.

pajlada (approved):

PR #5610, opened 2024-09-29
feat: run bench in ci

pajlada:
we only run each iteration once (essentially a benchmark dry run). this is just here to ensure we don't break benchmarks with some change we make


pajlada:
image

PR #5617, opened 2024-10-02
fix: use correct selection index when double-clicking into the void

Nerixyz:
From pajlada:

- Double click emote/emoji to select it
- Double click outside a message (in the void) and drag
- *Expected*: The selection starts at the nearest element - *Previous*: The selection starts at that selected emote/emoji

Technically, this fix does some duplicate work, since we did the work of looking at all elements, but shouldn't be that bad.


pajlada (approved):
Thanks for the quick fix!

PR #5626, opened 2024-10-05
feat: launch arg to start logged in as a supplied user

lbrooney:
Add a new optional launch arg , `--login <username>`, that allows a user to specify which account they want to start Chatterino logged in as. If an account with the supplied username does not exist, then it starts anonymous. This arg is essentially equivalent to modifying `current` in `settings.json` before launch.

I tried to implement similar to how channel args are applied in `WindowManager` by applying the username in `TwitchAccountManager`, but I don't know if it would be preferred to elevate it to `AccountController` instead. I did not implement the platform specificity that the `channel` and `activate` options have. 

I don't know if this is worth noting, but a side effect is that if the user never switches accounts, the `current` saved in `setting.json` will be set to whatever username they provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Minor: Removed experimental IRC support. (#5547)
 - Minor: Moderators can now see which mods start and cancel raids. (#5563)
 - Minor: The emote popup now reloads when Twitch emotes are reloaded. (#5580)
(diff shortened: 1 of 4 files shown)


pajlada (approved):
Great first PR - thank you!

As a first-time contributor, you are now eligible to be listed in the About page in Chatterino.

PR #5649, opened 2024-10-14
highlight tabs only on unviewed messages

hemirt:
This draft tackles the problem of unnecessary higlighting of images viewed in currently selected notebook tab, as discussed here: https://github.com/Chatterino/chatterino2/discussions/5412

It only tackles the part 1) do not highlight other tabs on messages shown in current tab, does not yet tackle the part 2) on selection unhiglight possible other tabs.

This is just a proof of concept, just to see what might be necessary to do these changes and to tackle the ideas.


src/widgets/helper/ChannelView.hpp
```diff
@@ -179,6 +179,9 @@ class ChannelView final : public BaseWidget
 
     LimitedQueueSnapshot<MessageLayoutPtr> &getMessagesSnapshot();
 
+    // Returns true if message should be included
+    bool shouldIncludeMessage(const MessagePtr &m) const;
+
```
hemirt, line 184: When currently selected tab has a channel that is filtered by filters, and a message is received in the underlying channel, but the current tab does not show it (because it is filtered out by the filters), this needed to be used to make other tabs highlight on this message.

(It's not really needed, so it doesnt have to be public, since you can copy the code from inside with the use of public getFilterSet and similar)

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,65 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+        newHighlightStyle == HighlightState::NewMessage)
+    {
+        return;
+    }
+
+    if (this->highlightState_ == newHighlightStyle ||
+        this->highlightState_ == HighlightState::Highlighted)
+    {
+        return;
+    }
```
hemirt, line 405: As was before, on no change, just dont do anything.

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,65 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+                isSubset(uniqueFilterIdsSource, uniqueFilterIdsSplit))
+            {
+                return;
+            }
+        }
+    }
+
+    this->highlightState_ = newHighlightStyle;
+    this->update();
+}
```
hemirt, line 443: As from before.

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,65 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            };
+
+            if (channelViewSource.underlyingChannel() == split->getChannel() &&
+                split->getChannelView().shouldIncludeMessage(message) &&
+                isSubset(uniqueFilterIdsSource, uniqueFilterIdsSplit))
```
hemirt, line 434: We check if the channels are same &&
if message got shown in currently viewed notebook tab splits &&
if source filters (i.e. other tabs split) is subset of current filters (i.e. we only dont-higlight when our current view is "lesser", i.e. more specific))

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,65 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+        return;
+    }
+
+    auto splitContainer =
+        dynamic_cast<SplitContainer *>(this->notebook_->getSelectedPage());
+    if (splitContainer != nullptr)
+    {
+        const auto &splits = splitContainer->getSplits();
+        for (const auto &split : splits)
+        {
```
hemirt, line 413: setHighlightState is called from each "source" split, i.e. if message is received on channel X, then each split with channel X calls this function, and we always only check currently being shown tab and compare it to the source split

therefore "splits" and "split" here refers to currently being shown/selected tab
and source refers to the split with a channel which received a message in another tab

hemirt:
So far I haven't noticed any performance issues either, although the decisions surely could be condensed into a precomputed NotebookTab<->NotebookTab, (except maybe the filter stuff, as I had to figure out to _do_ highlight other tabs when a current tab shows filtering splits and a message is received but filtered out. i.e. not shown)

src/widgets/helper/ChannelView.hpp
```diff
@@ -179,6 +179,9 @@ class ChannelView final : public BaseWidget
 
     LimitedQueueSnapshot<MessageLayoutPtr> &getMessagesSnapshot();
 
+    // Returns true if message should be included
```
Nerixyz, line 182: This should mention that `m` must "logically" belong to this channel, or rather that the origin of `m` (the channel it originates from or smth) isn't checked.

src/widgets/helper/NotebookTab.hpp
```diff
@@ -60,6 +61,8 @@ class NotebookTab : public Button
     bool isLive() const;
 
     void setHighlightState(HighlightState style);
```
Nerixyz, line 63: This function is only used once (and in tests). It should probably be renamed to avoid confusion.

Nerixyz (commented):
Tests would be great. Unfortunately, we can't mock `ChannelView`s. Maybe we can find some way to test this.

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,60 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
     this->update();
 }
 
+void NotebookTab::setHighlightState(HighlightState newHighlightStyle,
+                                    ChannelView &channelViewSource,
```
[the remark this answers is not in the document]
hemirt: I think at first I used .channel() which is not marked as const, I'll take a look, now it's underlyingChannel() only I believe, which is marked const.

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,60 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+                QSet(filterIdsSource.cbegin(), filterIdsSource.cend());
+            auto &&filterIdsSplit = split->getChannelView().getFilterIds();
+            auto uniqueFilterIdsSplit =
+                QSet(filterIdsSplit.cbegin(), filterIdsSplit.cend());
+
+            auto isSubset = []<typename T>(QSet<T> sub, QSet<T> super) {
+                return std::ranges::none_of(sub, [&super](const auto &subItem) {
+                    return !super.contains(subItem);
+                });
+            };
```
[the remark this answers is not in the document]
hemirt: 1) I see, I used auto&& universal reference so that it always works, without any possible needless copies. In current impl it should be the same as you gave.
2) True, just wrote out my intent, to better spell it out.
3) sure

src/widgets/helper/ChannelView.hpp
hemirt: I just moved the function from private to public part, realistically I do not like the name that was used for it, I was searching for a function like "isMessageFiltered" or "filterMessage", both of these names better describe what the function does in my opinion, though shouldIncludeMessage was the name that was already there.

src/widgets/helper/NotebookTab.hpp
hemirt: yeah its used when a tab is duplicated

src/widgets/helper/NotebookTab.cpp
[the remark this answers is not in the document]
hemirt: fixed

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,60 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+    {
+        return;
+    }
+
+    auto *splitContainer =
+        dynamic_cast<SplitContainer *>(this->notebook_->getSelectedPage());
+    if (splitContainer != nullptr)
+    {
+        const auto &splits = splitContainer->getSplits();
+        for (const auto &split : splits)
```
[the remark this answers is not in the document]
hemirt: done

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,56 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+
+            auto isSubset = []<typename T>(const QList<T> &sub, const QList<T> &super) {
+                return std::ranges::all_of(sub, [&super](const auto &subItem) {
+                    return super.contains(subItem);
+                });
+            };
+
+            if (channelViewSource.underlyingChannel() == split->getChannel() &&
+                split->getChannelView().shouldIncludeMessage(message) &&
+                isSubset(uniqueFilterIdsSource, uniqueFilterIdsSplit))
```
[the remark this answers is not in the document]
hemirt: fixed

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,6 +383,56 @@
...
```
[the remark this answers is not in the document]
hemirt: fixed

src/widgets/helper/NotebookTab.hpp
hemirt: // needs to be checked, with the duplicate tab

hemirt:
for now selecting a tab correctly unselects other tabs based on their status of viewed channels as a whole

hemirt:
I believe it should all be now working, unhiglighting on selection based on filters is based on name&filters identity, rather than any filters subsetting, makes more sense for each filter set to be considered its own to-be-highlighted, as it makes more sense for the user to know that each of their filters highlighted (even if they be subsets).

I.e. splits higlight if they should based on inclusion and unselection is based on identity.

hemirt:
Everything right now should be implemented and I'd appreciate it if someone could test it out. I've been using it for quite some time and it worked quite well for me, but I don't use any filters, so I only rudimentary tested those.

Nerixyz (commented):
From what I can tell, this works. Some test-cases would be great (not unit-tests, but written steps). For example:

Setup (could be the same for all tests):
- Tab 1: [TwitchDev]
- Tab 2: [TwitchDev]
- Tab 3: empty

Steps:
- Select Tab 3
- (some other account): Write `@{current-user}` in `TwitchDev` (i.e. ping the current user)
- expectation: Tab 1 and 2 are highlighted red
- Select Tab 1
- expectation: Tab 2 is no longer highlighted

If these get too long, you can use a `<details>` block (write `/details` and tab-complete):

```md
<details><summary>Steps</summary>

text...
</details>
```

src/widgets/helper/NotebookTab.hpp
```diff
@@ -107,6 +122,46 @@ class NotebookTab : public Button
...
+            bool operator()(const ChannelViewProxy &l,
+                            const ChannelViewProxy &r) const;
+        };
+
+        std::unordered_set<ChannelViewProxy, ChannelViewProxyHash,
+                           ChannelViewProxyEqual>
+            newMessageSource;
+        std::unordered_set<ChannelViewProxy, ChannelViewProxyHash,
+                           ChannelViewProxyEqual>
+            highlightedSource;
```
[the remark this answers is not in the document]
hemirt: Yes, I believe I see a possible issue. It might also crash on rehash, if those items no longer exist.

I believe I only require the
1) underlying channel
a) for comparison of channel == channel
b) for getName() -> identity (ID) of the source
2) channel view
a) for shouldIncludeMessage - uses the filter set (and the excludeUserMessagesFromFilter setting) - therefore it could work as a free function (provided we have the filter set)
b) getFilterIds -> identity (ID) of the source

Yes this should be replaced with some ID, I believe for the `ChannelView` rather than a `Channel`, granted it does not matter as long as the Name+FilterSet/FilterIds can be uniquely identified.

src/widgets/helper/NotebookTab.hpp
```diff
@@ -59,11 +60,25 @@ class NotebookTab : public Button
      **/
     bool isLive() const;
 
+    /**
+     * @brief Sets the highlight state of this tab clearing highlight sources
+     *
+     * Obeys the HighlightsEnabled setting and highlight states hierarchy
```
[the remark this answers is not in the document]
hemirt: that mentioned (`HighlightState::Highlighted`) is appearing with higher priority than new message (`HighlightState::NewMessage`)

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,9 +551,96 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+    // message is highlighting unvisible tab
+
+    auto underlyingChannel = channelViewSource.underlyingChannel();
+    auto newFilters = channelViewSource.getFilterIds();
+    auto channelViewProxy =
+        HighlightSources::ChannelViewProxy{&channelViewSource};
+
+    // the unvisible tab should unhighlight other tabs iff
+    // the other tab's filters are more generic therefore
+    // the other tab's filter set is subset of the unvisible tab
```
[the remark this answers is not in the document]
hemirt: yes

src/widgets/helper/NotebookTab.hpp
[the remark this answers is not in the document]
hemirt: Highlight source is just the split that is asking for a tab to be activated 
https://github.com/Chatterino/chatterino2/blob/e5e5a79645b7e3163468b0ae73ad75fbeb88ab00/src/widgets/splits/SplitContainer.cpp#L216-L224

src/widgets/Notebook.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include "widgets/BaseWidget.hpp"
 
+#include <ForwardDecl.hpp>
```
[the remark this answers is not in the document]
hemirt: yes, a left over from other various ways i tried out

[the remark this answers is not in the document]
hemirt: removed

src/widgets/helper/NotebookTab.cpp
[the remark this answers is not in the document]
hemirt: removed leftover

hemirt:
> From what I can tell, this works. Some test-cases would be great (not unit-tests, but written steps). For example:
> 
> Setup (could be the same for all tests):
> 
> * Tab 1: [TwitchDev]
> * Tab 2: [TwitchDev]
> * Tab 3: empty
> 
> Steps:
> 
> * Select Tab 3
> * (some other account): Write `@{current-user}` in `TwitchDev` (i.e. ping the current user)
> * expectation: Tab 1 and 2 are highlighted red
> * Select Tab 1
> * expectation: Tab 2 is no longer highlighted
> 
> If these get too long, you can use a `<details>` block (write `/details` and tab-complete):
> 
> ```
> <details><summary>Steps</summary>
> 
> text...
> </details>
> ```

I'm unsure where I would even write these - (as a comment here?), or whats the verbiage on these actions.

src/widgets/helper/NotebookTab.hpp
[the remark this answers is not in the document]
hemirt: > Yes this should be replaced with some ID, I believe for the `ChannelView` rather than a `Channel`, granted it does not matter as long as the Name+FilterSet/FilterIds can be uniquely identified.

Yes I believe that for example a ChannelView could be generating these IDs, the sets would then just hold the IDs, as we actually do not do anything special with the ChannelView. I'm gonna try it out in the meantime to figure out where and how would be the best way.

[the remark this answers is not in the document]
hemirt: Replaced it with an ID based on those two things which now shouldnt suffer from any dangling pointers, (or the rehash problem due to them), as the ChannelView is only accessed when it is asked by the ChannelView (or the split) and IDs are saved at this access time, and the ChannelView is not stored anymore - https://github.com/Chatterino/chatterino2/pull/5649/commits/0e4897969b8e132f2f675c84f418c9412ac1d313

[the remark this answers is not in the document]
hemirt: As the ID was being asked for with every highlight request (i.e. all messages that ultimately appear in a currently unviewed tab's splits), I moved the ID into the `ChannelView`, it is updated when filters are updated or the underlying channel is updated. https://github.com/Chatterino/chatterino2/pull/5649/commits/1bdb11839129fe3ceff801746b653dc77f4d4622.

src/widgets/helper/NotebookTab.hpp
```diff
@@ -107,6 +122,40 @@ class NotebookTab : public Button
 
     int normalTabWidthForHeight(int height) const;
 
+    bool shouldMessageHighlight(const ChannelView &channelViewSource,
+                                const MessagePtr &message) const;
+
+    struct HighlightSources {
+        using ChannelViewId = std::size_t;
+
+        static ChannelViewId GetChannelViewId(
```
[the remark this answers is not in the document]
hemirt: outdated

hemirt:
Wrote out a bunch of test scenarios, hopefully I didn't confuse myself somewhere there.

<details><summary>Details</summary>
<p>

Common setup:
- Account name: User
- a message containing User will Highlight (red)
- a message containing Hello will just NewMessage (white)
- Tab 1: [Earth, Moon]
- Tab 2: [Moon]
- Tab 3: [Earth]
- Tab 4: [Mars, Phobos, Deimos]

Scenario 1:
- Selected Tab 1 [Earth, Moon]
- (other account) types (a) User / (b) Hello in Earth
- Expectation: no tab is highlighted

Scenario 2:
- Selected Tab 1 [Earth, Moon]
- (other account) types (a) User / (b) Hello in Mars
- Expectation: Tab 2 highlights in (a) Highlighted (red) / (b) NewMessage (white) state 

Scenario 3:
- Selected Tab 4 [Mars, Phobos, Deimos]
- (other account) types (a) User / (b) Hello in Earth
- Expectation: Tab 1 [Earth, Moon] and Tab 3 [Earth] highlighted in (a) Highlighted (red) / (b) NewMessage (white) state
- Select Tab 3 [Earth]
- Expectation: Tab 1 [Earth, Moon] in unhighlight (gray) state
- Expectation: Tab 3 [Earth] in active (blue) state

Scenario 4:
- Selected Tab 4 [Mars, Phobos, Deimos]
- (other account) types (a) User / (b) Hello in Earth
- (other account) types (a) User / (b) Hello in Moon
- Expectation: Tab 1, 2, 3 are (a) Highlighted (red) / (b) NewMessage (white) state
- Subscenario i)
-- Select Tab 1 [Earth, Moon]
-- Expectations: Tab 1 active, Tab 2,3 unhighlighted (gray)
- Subscenario ii)
-- Select Tab 2 [Moon]
-- Expectations: Tab 2 active, Tab 1, 3 still in (a) red / (b) white state
-- Select Tab 3 [Earth]
-- Expectations: Tab 3 active, Tab 1, 2 unhighlighted (gray)

Scenario 5:
- Selected Tab 4 [Mars, Phobos, Deimos]
- (other account) types User in Earth
- (other account) types Hello in Moon
- Expectations: Tab 1 [Earth, Moon] Highlighted (red), Tab 2 [Moon] NewMessage (white), Tab 3 [Earth] Highlighted (red)
- Subscenario i)
-- Select Tab 1 [Earth, Moon]
-- Expectations: Tab 1 active, Tab 2, 3 unhighlighted (gray)
- Subscenario ii)
-- Select Tab 2 [Moon]
-- Expectations: Tab 2 active, Tab 1, 3 still Highlighted (red)
- Subscenario iii)
-- Select Tab 3 [Earth]
-- Expectations: Tab 3 active, Tab 1, 2 NewMessage (white)

</p>
</details>

src/widgets/helper/NotebookTab.hpp
```diff
@@ -107,6 +122,25 @@ class NotebookTab : public Button
 
     int normalTabWidthForHeight(int height) const;
 
+    bool shouldMessageHighlight(const ChannelView &channelViewSource,
+                                const MessagePtr &message) const;
+
+    struct HighlightSources {
+        std::unordered_set<ChannelView::ChannelViewID> newMessageSource;
+        std::unordered_set<ChannelView::ChannelViewID> highlightedSource;
```
pajlada, line 130: Document these members - I'm not sold on the naming of `highlight source`, so my thought is that with these documented to better describe your intent I might understand whether the name is ok or not

pajlada (changes requested):
one "bug": Two splits with the same filter applied in two separate tabs will both highlight with the current code

Setup:
Tab 1: [Moon with filter `(message.content contains "forsen")`]
Tab 2: [Moon with the same filter `(message.content contains "forsen")`]

Scenario:
Other user types "forsen" in Moon - both tabs highlight


not a final review, unit tests would be super handy but it's not something I think you should try to tackle in this PR

src/widgets/helper/NotebookTab.hpp
hemirt: the source in here refers to the split in which the message appears, I'll add this as a comment there

hemirt:
> one "bug": Two splits with the same filter applied in two separate tabs will both highlight with the current code
> 
> Setup: Tab 1: [Moon with filter `(message.content contains "forsen")`] Tab 2: [Moon with the same filter `(message.content contains "forsen")`]
> 
> Scenario: Other user types "forsen" in Moon - both tabs highlight
> 
> not a final review, unit tests would be super handy but it's not something I think you should try to tackle in this PR

Not sure on which Tab youre actually selected, but if youre selected on a different Tab, e.g. Tab 3, then both Tab 1 and Tab 2 should highlight, then selecting either tab will unhiglight the other tab.

Fixed the issue when you're on Tab 1 and Tab 2 gets highlighted, when it shouldn't (as you're viewing the same thing in Tab 1).


I.e. each split (view) acts based on its own Identity (ID), i.e. having different filters will act as if its a different channel.

src/widgets/helper/ChannelView.hpp
hemirt: so far I just renamed the m to message, the other naming way would require a logic flip everywhere where its used.
( isMessageFiltered == not shouldIncludeMessage)

src/widgets/helper/ChannelView.hpp
```diff
@@ -212,9 +215,13 @@ class ChannelView final : public BaseWidget
 
     Scrollbar *scrollbar();
 
+    using ChannelViewID = std::size_t;
+    ChannelViewID getID() const;
```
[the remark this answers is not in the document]
hemirt: >The ID is made of the underlying channel's name combined with the filter set IDs

This makes the views act equivalently, despite not being the same object. (A view in Tab 1 and view in Tab 2 are different objects, yet they can show the same things.)

pajlada (commented):
Nitpicky, but I think it will help me understand it more

I'd prefer if instead of storing "sources" of highlights, we store "unread highlights"

Instead of storing `newMessage` and `highlighted` unread highlights in separate sets, store them in a single map instead (e.g. `map<ChannelViewID, HighlightState>` where the HighlightState is the "highest" highlight state). If for some reason we must store both the `new message` and `highlighted` highlight states of an unread highlight, we could instead use a `FlagsEnum`, but I'd like to understand why that would be necessary first

Then, instead of having separate `removeNewMessageSource` etc it would be called `markHighlightsFromChannelsAsRead(ChannelViewID)` or something without having to specify the highlight type

let me know if there's anything in that feedback that doesn't make sense

hemirt:
> Instead of storing `newMessage` and `highlighted` unread highlights in separate sets, store them in a single map instead (e.g. `map<ChannelViewID, HighlightState>` where the HighlightState is the "highest" highlight state). If for some reason we must store both the `new message` and `highlighted` highlight states of an unread highlight, we could instead use a `FlagsEnum`, but I'd like to understand why that would be necessary first

It is needed in this scenario

Tab 1: [Earth, Moon]
Tab 2: [Earth]
Tab 3: [Moon]

you're on Tab 4 [Mars]

if you in channel #Earth get a new message (white)
and in channel #Moon get pinged (red)

then Tab 1 shows as red (i.e. the highest state)

If you select the Tab 3: [Moon] and view this ping message, then Tab 1 goes to the new message state (as you have yet to see the new message in channel #Earth

Just state flags on a Tab would not work (as I see it) since you need to know _who_ is the source of the highlight state (and as said above, also which state), to properly unmark them.

hemirt:
> Yes, I believe we only need the highest state of any given source, as it gets all removed on selection. So it might be possible, gonna take a look and try to implement it that way.

incorportaed the changes in [05ab3b2](https://github.com/Chatterino/chatterino2/pull/5649/commits/05ab3b27e783c33aceaed78f283a6b05aeed54e5)

src/widgets/helper/NotebookTab.cpp
```diff
@@ -381,9 +517,82 @@ void NotebookTab::setHighlightState(HighlightState newHighlightStyle)
...
+    {
+        return;
+    }
+
+    this->highlightState_ = newHighlightStyle;
+    this->update();
+}
+
+bool NotebookTab::shouldMessageHighlight(const ChannelView &channelViewSource,
+                                         const MessagePtr &message) const
```
pajlada, line 576: This function is essentially: `isMessageInAnyCurrentlyRenderedChannelView` - could it & its logic be moved to ChannelView instead (maybe as an anonymous function) so `tabHighlightRequested` doesn't need the `message` param, and a tab highlight always means a highlight will take place (except if the tab is selected)?

hemirt: I've looked through it and I believe in here:

https://github.com/Chatterino/chatterino2/blob/05ab3b27e783c33aceaed78f283a6b05aeed54e5/src/widgets/helper/NotebookTab.cpp#L585-L588

the channelViewSource.shouldIncludeMessage is superfluous, the way the message flows indicates it already was checked.

But the problem is that we need to figure out if _other_ visible splits can see this message.

Unfortunately I don't see a way to remove the message param, since the handling happens both asynchronously and most notably - independently.

hemirt: The only need for this message would be if we want the underlying channel to not highlight if a message appears and is also shown in visible filtered split.


What should then be done?
1) message appearing in currently visible filtered channel should not highlight other splits that are of same underlying channel but _unfiltered_ -> this needs the message there
2) they should highlight -> this does not need to check any message

hemirt: Currently with the switch to identity based checking and treating each filtered channel as its own different channel, the message does not need to be there. So I removed it now.

There is a possibility for message checking if we want to make messages that appear in currently visible filtered split to not highlight some other splits where the message also appears. But it only makes sense for a message not to appear in unfiltered split, as other filtered splits are set by the user to show the message for some reason.

pajlada: Awesome :+1: This looks better, even if NotebookTab has some logic I don't want it to have, this is acceptable

src/widgets/helper/ChannelView.hpp
hemirt: is no longer needed to be public, so reverted

pajlada (approved):

Felanbird:
Am I misunderstanding what this PR was supposed to do? from what I can see it still sets the duplicates of a tab as unread, even when you have one of them focused

PR #5652, opened 2024-10-17
refactor(message-builder): move static helper methods to functions

Nerixyz:
These functions shouldn't live in the message builder. They're related to IRC and Ignores.
IRC related functions are in `TwitchIrc` and in the `chatterino::twitchirc` namespace. All functions were copied. Only parameters equal to `originalMessage` were renamed to `content` to avoid confusion with the original message text, which might be bigger.

In the future, we should get rid of the `int`s in this code.

clang-tidy will complain, but I didn't change anything.


CHANGELOG.md
```diff
@@ -107,6 +107,7 @@
 - Dev: Twitch messages are now sent using Twitch's Helix API instead of IRC by default. (#5607)
 - Dev: `GIFTimer` is no longer initialized in tests. (#5608)
 - Dev: Emojis now use flags instead of a set of strings for capabilities. (#5616)
+- Dev: Removed static `MessageBuilder` helpers to standalone functions. (#5652)
```
pajlada, line 110:
```suggestion
- Dev: Refactored static `MessageBuilder` helpers to standalone functions. (#5652)
```

pajlada (changes requested):
> parameters equal to originalMessage were renamed to content to avoid confusion with the original message text, which might be bigger.

I don't understand the reasoning behind the change - I don't really care too much about it, but I don't see how `originalMessage` would be incorrect

Nerixyz:
> but I don't see how `originalMessage` would be incorrect

It's not incorrect, but it's misleading. For the reply `@user foo bar`, the content could be `foo bar`. Yes, `foo bar` is `originalMessage_` of `MessageBuilder`, but it's not the "original" message - that would be `@user foo bar`.

pajlada:
> > but I don't see how `originalMessage` would be incorrect
> 
> It's not incorrect, but it's misleading. For the reply `@user foo bar`, the content could be `foo bar`. Yes, `foo bar` is `originalMessage_` of `MessageBuilder`, but it's not the "original" message - that would be `@user foo bar`.

Sure, I guess that depends on how you interpret "message" and "content". Libcommuni interprets "content" as `@user foo bar`

PR #5655, opened 2024-10-18
test: use httpbox fork release

pajlada:
still need to change windows test runner, but i want to test the macos one first


pajlada (commented):
image

PR #5661, opened 2024-10-19
feat: add shared chat badge

iProdigy:
image


src/singletons/WindowManager.cpp
```diff
@@ -195,6 +195,7 @@ void WindowManager::updateWordTypeMask()
     flags.set(settings->animateEmotes ? MEF::BitsAnimated : MEF::BitsStatic);
 
     // badges
+    flags.set(MEF::BadgeSharedChannel);
```
iProdigy, line 198: should there be a new setting to toggle visibility of this badge? (or shall we lean on an existing setting?)

iProdigy:
technically we could reduce the number of helix requests made by using https://dev.twitch.tv/docs/api/reference/#get-shared-chat-session

e.g., if shared chat has 6 channels, the current approach could make up to 5 helix requests while the alternate approach would just do 1 - but not a big deal in the context of the full app lifetime

src/singletons/WindowManager.cpp
pajlada: I'm fine to not have a setting for this to begin with

src/messages/MessageBuilder.cpp
```diff
@@ -2746,6 +2757,28 @@ void MessageBuilder::appendTwitchBadges(const QVariantMap &tags,
         return;
     }
 
+    if (this->message().flags.has(MessageFlag::SharedMessage))
+    {
+        QString sourceId = tags["source-room-id"].toString();
+        std::optional<QString> sourceName;
+        if (twitchChannel->roomId() == sourceId)
+        {
+            sourceName = twitchChannel->getName();
```
pajlada, line 2766: This assumes an incoming message to channel A is a shared message where the `source-room-id` is also channel A - can this actually happen?

src/messages/MessageBuilder.cpp
```diff
@@ -2746,6 +2757,28 @@ void MessageBuilder::appendTwitchBadges(const QVariantMap &tags,
...
+            sourceName =
+                getApp()->getTwitch()->getOrPopulateChannelCache(sourceId);
+        }
+
+        if (sourceName.has_value())
+        {
+            this->emplace<BadgeElement>(makeSharedChatBadge(sourceName.value()),
+                                        MessageElementFlag::BadgeSharedChannel)
+                ->setLink({Link::UserInfo, sourceName.value()});
+        }
```
pajlada, line 2779: Since `sourceName` can also be empty, we should check for both has_value and `sourceName.value().isEmpty()`

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -1128,6 +1129,38 @@ std::shared_ptr<Channel> TwitchIrcServer::getChannelOrEmptyByID(
...
+{
+    {
+        const auto cache = this->channelNamesById_.access();
+        if (cache->exists(channelId))
+        {
+            return cache->get(channelId);
+        }
+
+        // prevent multiple helix requests for single user
+        cache->put(channelId, "");
```
pajlada, line 1143: Document this behaviour in the header file for this method

pajlada (changes requested):
small nits, not tested yet - overall looks good. badge looks good

TODO for myself: test right-click actions on shared channel badges

src/messages/MessageBuilder.cpp
iProdigy: due to the latest refactor, yeah `twitchChannel` in that method can be the `sourceChannel` (field no longer exists) rather than the TwitchChannel that received the message: https://github.com/Chatterino/chatterino2/blob/e35fabfabe2f430e5614340a0fc6e5009db415b3/src/messages/MessageBuilder.cpp#L1946

src/messages/MessageBuilder.cpp
iProdigy: `makeSharedChatBadge` has that check & can intentionally be passed an empty string

but yeah that can be problematic for the UserInfo link

iProdigy: Done in https://github.com/Chatterino/chatterino2/pull/5661/commits/021f1a0bfc7c77c18aaa7f65a62f7263b7db6893

8thony:
I got a crash with all settings set to default and then joined a channel with Shared Chat enabled.
Testet on a9b446f

Another option could be that instead of a badge, we put the channel profile picture in front of the message.

pajlada:
> Another option could be that instead of a badge, we put the channel profile picture in front of the message.

I'd rather that be part of a follow-up PR as a "polish pass", which would let us discuss it first too

iProdigy:
@8thony could you share the channel name (& the other channels that were in the shared chat session & whether you were also connected to any of those channels)

also the stacktrace from the crash dump would be very helpful

8thony:
The Shared Chat Session had 2 People and I opened only one channel from those. 
c2 had no settings, no account, no cache files, nothing setup.
You can rename your %appdata%\chatterino2 folder to something else to test this.

src/providers/twitch/TwitchIrcServer.hpp
```diff
@@ -95,6 +100,14 @@ class TwitchIrcServer final : public ITwitchIrcServer, public QObject
     std::shared_ptr<Channel> getChannelOrEmptyByID(
         const QString &channelID) override;
 
+    /**
+     * Obtains the channel display name associated with the passed ID,
+     * so that Shared Chat messages can provide source channel context.
+     * Can yield an empty string if a helix request is already in-flight.
+     */
+    std::optional<QString> getOrPopulateChannelCache(
+        const QString &channelId) override;
```
Nerixyz, line 109: This duplicates `TwitchUsers::resolveID` (kinda). It doesn't use an LRU cache.

resources/twitch/sharedChat.png
Nerixyz, line 1: The edges look blurred. Is that intended? We should also think about providing 1x, 2x, and 3/4x versions of badges.

src/messages/MessageBuilder.cpp
```diff
@@ -380,6 +380,17 @@ EmotePtr makeAutoModBadge()
...
 
+EmotePtr makeSharedChatBadge(const QString &sourceName)
+{
+    return std::make_shared<Emote>(
+        Emote{"SharedChat_" + sourceName,
+              ImageSet{Image::fromResourcePixmap(
+                  getResources().twitch.sharedChat, 0.25)},
+              Tooltip{"Shared Message" +
+                      (sourceName.isEmpty() ? "" : " from " + sourceName)},
+              Url{"https://link.twitch.tv/SharedChatViewer"}});
```
Nerixyz, line 391:
```suggestion
              Url{"https://link.twitch.tv/SharedChatViewer"},});
```
+ reformat

src/messages/MessageBuilder.cpp
```diff
@@ -380,6 +380,17 @@ EmotePtr makeAutoModBadge()
         Url{"https://dashboard.twitch.tv/settings/moderation/automod"}});
 }
 
+EmotePtr makeSharedChatBadge(const QString &sourceName)
+{
+    return std::make_shared<Emote>(
+        Emote{"SharedChat_" + sourceName,
```
Nerixyz, line 386: We don't need a name for this.

resources/twitch/sharedChat.png
iProdigy: I wanted the corners to look rounded (kinda similar to our automod badge) so I feathered that part of the outline, but it's probably not the prettiest approach since i'm not a graphic designer

iProdigy:
@8thony could you confirm the crash is now fixed?

pajlada (approved):

PR #5668, opened 2024-10-21
fix: subscribed threads not being marked as subscribed

pajlada:
Got another branch going which updates unit tests to properly catch this

cc @Nerixyz if you want to take a peek - I imagine this was roughly the intent you had

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,7 +110,7 @@
 - Dev: Emojis now use flags instead of a set of strings for capabilities. (#5616)
 - Dev: Move plugins to Sol2. (#5622)
 - Dev: Refactored static `MessageBuilder` helpers to standalone functions. (#5652)
-- Dev: Decoupled reply parsing from `MessageBuilder`. (#5660)
+- Dev: Decoupled reply parsing from `MessageBuilder`. (#5660, #5668)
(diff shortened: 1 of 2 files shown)


Nerixyz:
> cc @Nerixyz if you want to take a peek - I imagine this was roughly the intent you had


Not exactly. If the thread isn't highlighted, that's an issue with the IrcMessageHandler, which sets the flag (https://github.com/Chatterino/chatterino2/blob/master/src/providers/twitch/IrcMessageHandler.cpp#L729).

pajlada:
> > cc @Nerixyz if you want to take a peek - I imagine this was roughly the intent you had
> 
> Not exactly. If the thread isn't highlighted, that's an issue with the IrcMessageHandler, which sets the flag ([`master`/src/providers/twitch/IrcMessageHandler.cpp#L729](https://github.com/Chatterino/chatterino2/blob/master/src/providers/twitch/IrcMessageHandler.cpp?rgh-link-date=2024-10-21T11%3A05%3A03Z#L729)).

Setting `SubscribedThread` here is too late for the highlight controller to act on, will keep current implementation for now

Nerixyz:
> Setting `SubscribedThread` here is too late for the highlight controller to act on, will keep current implementation for now

Then we can drop the `highlight` in the reply context.

PR #5678, opened 2024-10-26
refactor: deduplicate IRC parsing

Nerixyz:
This deduplicates the code for parsing historical vs. live messages. It introduces a `MessageSink` which hold some traits (some behavior as to how messages should be added). It is modelled after the `Channel`, which is why it has the `addMessage` signature. Ideally, I'd want to make a new function that _just_ takes a message, but I couldn't come up with a good name (because ultimately, it will add a message to the sink). An alternative could be to make `MessageSink` a wrapper around `Channel`.

This should make it easier to parse specific messages (e.g. subs) without introducing differences between recent and live messages.

### DIFF
--- a/src/common/Channel.hpp
+++ b/src/common/Channel.hpp
@@ -1,8 +1,10 @@
 #pragma once
 
+#include "common/enums/MessageContext.hpp"
 #include "controllers/completion/TabCompletionModel.hpp"
 #include "messages/LimitedQueue.hpp"
 #include "messages/MessageFlag.hpp"
+#include "messages/MessageSink.hpp"
 
 #include <magic_enum/magic_enum.hpp>
 #include <pajlada/signals/signal.hpp>
@@ -26,15 +28,7 @@ enum class TimeoutStackStyle : int {
     Default = DontStackBeyondUserMessage,
 };
 
-/// Context of the message being added to a channel
-enum class MessageContext {
-    /// This message is the original
-    Original,
-    /// This message is a repost of a message that has already been added in a channel
-    Repost,
-};
-
-class Channel : public std::enable_shared_from_this<Channel>
+class Channel : public std::enable_shared_from_this<Channel>, public MessageSink
 {
 public:
     // This is for Lua. See scripts/make_luals_meta.py
@@ -55,7 +49,7 @@ public:
     };
 
     explicit Channel(const QString &name, Type type);
-    virtual ~Channel();
+    ~Channel() override;
 
     // SIGNALS
     pajlada::Signals::Signal<const QString &, const QString &, bool &>
@@ -85,8 +79,9 @@ public:
     // overridingFlags can be filled in with flags that should be used instead
     // of the message's flags. This is useful in case a flag is specific to a
     // type of split
-    void addMessage(MessagePtr message, MessageContext context,
-                    std::optional<MessageFlags> overridingFlags = std::nullopt);
+    void addMessage(
+        MessagePtr message, MessageContext context,
+        std::optional<MessageFlags> overridingFlags = std::nullopt) final;
     void addMessagesAtStart(const std::vector<MessagePtr> &messages_);
 
     void addSystemMessage(const QString &contents);
@@ -94,8 +89,8 @@ public:
     /// Inserts the given messages in order by Message::serverReceivedTime.
     void fillInMissingMessages(const std::vector<MessagePtr> &messages);
 
-    void addOrReplaceTimeout(MessagePtr message);
-    void disableAllMessages();
+    void addOrReplaceTimeout(MessagePtr message, QTime now) final;
+    void disableAllMessages() final;
     void replaceMessage(MessagePtr message, MessagePtr replacement);
     void replaceMessage(size_t index, MessagePtr replacement);
     void deleteMessage(QString messageID);
@@ -104,8 +99,13 @@ public:
     void clearMessages();
 
     MessagePtr findMessage(QString messageID);
+    MessagePtr findMessageByID(QStringView messageID) final;
 
     bool hasMessages() const;
+
+    void applySimilarityFilters(const MessagePtr &message) const final;
+
+    MessageSinkTraits sinkTraits() const final;
 
     // CHANNEL INFO
     virtual bool canSendMessage() const;
--- a/src/messages/MessageSimilarity.hpp
+++ b/src/messages/MessageSimilarity.hpp
@@ -1,0 +1,73 @@
+#pragma once
+
+#include "Application.hpp"
+#include "controllers/accounts/AccountController.hpp"
+#include "messages/Message.hpp"
+#include "providers/twitch/TwitchAccount.hpp"
+#include "singletons/Settings.hpp"
+
+#include <ranges>
+
+namespace chatterino::similarity::detail {
+
+float relativeSimilarity(QStringView str1, QStringView str2);
+
+float inMessages(const MessagePtr &msg,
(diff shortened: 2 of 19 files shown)


src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -778,32 +343,41 @@ void IrcMessageHandler::handlePrivMessage(Communi::IrcPrivateMessage *message,
...
-            {
-                auto parsedBadges = parseBadges(badgesTag.toString());
-                twitchChannel->setMod(parsedBadges.contains("moderator"));
-                twitchChannel->setVIP(parsedBadges.contains("vip"));
-                twitchChannel->setStaff(parsedBadges.contains("staff"));
-            }
+            auto parsedBadges = parseBadges(badgesTag.toString());
+            channel->setMod(parsedBadges.contains("moderator"));
+            channel->setVIP(parsedBadges.contains("vip"));
+            channel->setStaff(parsedBadges.contains("staff"));
```
Nerixyz, line 367: Not sure if I'm a fan of this running for historic messages, but I don't see how it could be that bad right now.

src/util/VectorMessageSink.hpp
```diff
@@ -0,0 +1,36 @@
...
+    void disableAllMessages() override;
+
+    void applySimilarityFilters(const MessagePtr &message) const override;
+
+    MessagePtr findMessageByID(QStringView id) override;
+
+    MessageSinkTraits sinkTraits() const override;
+
+    const std::vector<MessagePtr> &messages() const;
+    std::vector<MessagePtr> takeMessages() &&;
```
pajlada, line 28: `&&` here means the user must call this on an rvalue reference (e.g. by calling `std::move(vecotMessageSink)`) first, and we use it here to ensure a VectorMessageSink is never left with a moved-from `messages_` field - is that correct?

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -1405,76 +954,12 @@ void IrcMessageHandler::handlePartMessage(Communi::IrcMessage *message)
...
-                message->flags.set(MessageFlag::Disabled, true);
-            }
-        }
-    }
-}
-
 void IrcMessageHandler::addMessage(Communi::IrcMessage *message,
-                                   const ChannelPtr &chan,
-                                   const QString &originalContent,
-                                   ITwitchIrcServer &server, bool isSub,
```
pajlada, line 1471: Undo the changes to `ITwitchIrcServer` and `ChannelPtr` here please, unless they are necessary for the deduplication

src/messages/MessageSimilarity.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include "providers/twitch/TwitchAccount.hpp"
+#include "singletons/Settings.hpp"
+
+#include <ranges>
+
+namespace chatterino::similarity::detail {
+
+float relativeSimilarity(QStringView str1, QStringView str2);
+
+float inMessages(const MessagePtr &msg,
```
pajlada, line 15: move to source file

src/messages/MessageSimilarity.hpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    return similarityPercent;
+}
+
+}  // namespace chatterino::similarity::detail
+
+namespace chatterino {
+
+void setSimilarityFlags(const MessagePtr &message,
```
pajlada, line 46: move to source file

src/messages/MessageSink.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace chatterino {
+
+struct Message;
+using MessagePtr = std::shared_ptr<const Message>;
+
+enum class MessageSinkTrait : uint8_t {
+    None = 0,
+    AddMentionsToGlobalChannel = 1 << 0,
+    RequiresKnownChannelPointReward = 1 << 1,
```
pajlada, line 19: Document these

src/messages/MessageSink.hpp
```diff
@@ -0,0 +1,58 @@
...
+};
+using MessageSinkTraits = FlagsEnum<MessageSinkTrait>;
```
pajlada, line 21: add FlagsEnum include

src/messages/MessageSink.hpp
```diff
@@ -0,0 +1,58 @@
...
+    virtual ~MessageSink() = default;
+
+    /// Add a message to this sink
+    ///
+    /// @param message The message to add (non-null)
+    /// @param ctx The context in which this message is being added.
+    /// @param overridingFlags
+    virtual void addMessage(
+        MessagePtr message, MessageContext ctx,
+        std::optional<MessageFlags> overridingFlags = std::nullopt) = 0;
```
pajlada, line 36: add optional include

src/messages/MessageSink.hpp
```diff
@@ -0,0 +1,58 @@
...
+    /// Searches for similar messages and flags this message as similar
+    /// (based on the current settings).
+    virtual void applySimilarityFilters(const MessagePtr &message) const = 0;
+
+    /// @brief Searches for a message by an ID
+    ///
+    /// If there is no message found, an empty shared-pointer is returned.
+    virtual MessagePtr findMessageByID(QStringView id) = 0;
+
+    ///
```
pajlada, line 54: `///`

src/common/Channel.hpp
```diff
@@ -104,9 +99,14 @@ class Channel : public std::enable_shared_from_this<Channel>
     void clearMessages();
 
     MessagePtr findMessage(QString messageID);
```
pajlada, line 101: can add a deprecated annotation on this one if we want to phase it out later

pajlada (commented):
a little mini review, from what I've seen so far it looks good

I'm not super fond of the word "sink" but I don't know of a better word that's not super wordy

src/messages/MessageSimilarity.hpp
Nerixyz: That's not possible as it takes any range (e.g. vector, limitedqueue, map, etc)

pajlada: We only use 2 types right now (limitedqueue & vector), if we explicitly instantiate those in the source file, the two detail functions can now both be anonymous in the source file instead.
Less flexibility, but given what we support now I think it's ok. If we start supporting a bunch of other types of random range types, we can reconsider it

Nerixyz: Sure

src/util/VectorMessageSink.hpp
Nerixyz: Yes, that's the intention.

src/providers/twitch/IrcMessageHandler.cpp
Nerixyz: Re-added the server, but there's no point in using a `ChannelPtr` there. We're dealing with `TwitchChannel`s and the signatures should show that.

pajlada: Since this is not a tiny refactor, I wanted as few things changed as possible unless they were directly related to the IRC/RecentMessages parsing. Function signature changes & renames like `chan` to `channel` makes this more difficult.
If changing `ChannelPtr` to `TwitchChannel` makes sense as part of this refactor (e.g. because RecentMessages paths already deal directly with a `TwitchChannel`), at least keep the variable name the same.

pajlada (approved):

PR #5685, opened 2024-11-02
fix(7TV): ignore `entitlement.reset`

Nerixyz:
We currently log something like "Unknown subscription type: INVALID". This just adds the event type to the known ones and silently ignores the event. `entitlement.reset` is pushed every time a user without 7TV cosmetics updates their presence. As most users don't have cosmetics, it's really spammy. I don't see a big reason to handle it. In my opinion, we should handle it once there's a bug that would be fixed by handling the event (keep in mind: we didn't need this in the past).


pajlada (approved):

RainbowScientist5 (commented):
0x00000000009726632680fb29d3f7a9734e3010e2

PR #5690, opened 2024-11-03
feat: add setting to hide blocked term automod messages

pajlada:
Blocked terms can be added at https://dashboard.twitch.tv/settings/moderation/blocked-terms (replace with your username). Twitch fairly recently made messages blocked by blocked terms show up as automod messages. This PR aims to make this customizable, just like it is in Twitch Web Chat.

This is controlled with the new setting "Show blocked term automod messages"
image

I left it as an enum with two options because I see a near future where this can have the option "Only in streamer mode" or something like that

Instead of not adding the message to the channel view, it does add it but it's not laid out, thus making it hidden. This is how some other moderation action messages work, and I didn't want this PR to be the PR to solidify this behaviour, or figure out which is correct, so I left a comment for future adventurers.


iProdigy:
could also improve this ui:
Screenshot_20241104_021111

https://gist.github.com/iProdigy/6b94599326bc991f7ceb4b33fb20888c

PR #5693, opened 2024-11-04
chore: clarify Lua submodule version

brian6932:
Updating submodules causes
```
lib\sol2\include\sol/compatibility/compat-5.3.h(404): fatal error C1189: #error: "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
```

- sol requires `501`-`504`
https://github.com/ThePhD/sol2/blob/2b0d2fe8ba0074e16b499940c4f3126b9c7d3471/include/sol/compatibility/compat-5.3.h#L401-L406
- lua `master` uses `505`
https://github.com/lua/lua/blob/e3ce88c9e850b7e79751083014699c5eae1bff31/lua.h#L20-L24


Nerixyz:
> lua `master` uses `505`

That shouldn't matter. We're checked out on the v5.4.7 tag. If we'd use the latest master commit, Sol would break in CI.

brian6932:
>We're checked out on the v5.4.7 tag

Yes, but it's not exactly clear that updating submodules will break the build. Imo the branch should just be specified in `.gitmodules`, especially since lua releases are set to their own branches anyway.

pajlada:
This is less of a fix, and more of a hint to future developers clarifying what branch we're tracking.
It also makes it easier for us to be able to track changes to that specific branch, since Lua maintains old versions as branches

pajlada (approved):

PR #5698, opened 2024-11-05
chore: standardize references to FrankerFaceZ & BetterTTV in settings

Felanbird:
started because half the FFZ settings were `ffz` and the other half were `frankerfacez` so if you're used to searching ffz sometimes you don't get the setting you want, but if we update ffz we should update bttv also
now i know what you're thinking, doesn't this make it worse for everyone, and yeah it does lol got u, but standardization

anyway we're respecting branding now and it pushes pajlada towards implementing search tags to restore `bttv` and `ffz` search


pajlada (approved):

leon-richardt:
> implementing search tags to restore bttv and ffz search

or just fuzzy search 😌

PR #5699, opened 2024-11-05
feat: improve automod blocked term messaging

pajlada:
Top two AutoMod messages when streamer mode is enabled & the default-enabled "Hide blocked terms" under Streamer Mode is enabled, bottom two is when that is disabled.

image


pajlada (commented):
image

PR #5705, opened 2024-11-08
Deduplicate emotes by name in smart tab emote completion strategy

ilya-zlobintsev:
This PR makes it so if there are multiple emotes with the same name (for example from different sources) and you are using the "smarter emote completion" option, when tabbing through them it no longer cycles through all the instances of the same name.

Originally I changed this in `SmartEmoteStrategy`, but this ended up removing duplicates from the : dropdown menu as well, which is not something we want I believe. Now it's deduplicated specifically in `TabCompletionModel` regardless of the emote strategy, but I think this should be fine as it's a cheap operation.


pajlada (approved):

Mm2PL (approved):
looks good to me

Issue #5694, opened 2024-11-04
Qt `6.8.0` breaks bold usernames


brian6932:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Screenshots

left `6.7.2`, right `6.8.0`
![](https://github.com/user-attachments/assets/f6658bb2-92ce-4ec2-906a-da5732187048)

### OS and Chatterino Version

68304272a26c12b7d72fa3d95a732471341e3d15 built on 2024-11-04 with Qt 6.8.0, MSVC 194134123 Running on Windows 10 Version 21H2, kernel: 10.0.19044

Nerixyz:
I couldn't see a difference in the screenshot you provided, so here are they next to each other.

6.8.0:
![6.8.0](https://github.com/user-attachments/assets/2b589e10-8c3c-4ba1-94ee-729180a1d523)
6.7.1:
![6.7.1](https://github.com/user-attachments/assets/961e949a-4372-4e21-8466-722bb214f746)

The username is still bold (just not as bold as before). This doesn't look like the default font. Does this happen with the default font too? I'm guessing your font provides weights that GDI couldn't display, but DirectWrite can. We currently don't use OpenType font weights, because of Qt 5 support. But in the future, we could revisit the font-weights.

brian6932:
>This doesn't look like the default font. Does this happen with the default font too?

Yea it's not the default font. https://github.com/belluzj/fantasque-sans This is the font used, specifically the `NoLoopK` variant, `ttf`s (TrueType) only are installed, not OpenType. I can't really tell if the default font exhibits it or not, it doesn't seem much bolder in the first place.
`6.7.1`,`6.7.2`,`6.8.0`
![](https://github.com/user-attachments/assets/4678f69d-e5a0-40d4-87c9-6dddbb278c0d)

Nerixyz:
[...]

PR #5710, opened 2024-11-12
chore: keep GDI fontengine on Windows for now

Nerixyz:
From https://github.com/Chatterino/chatterino2/issues/5694: The DirectWrite font engine in Qt 6.8.0 doesn't match the behavior of the old GDI engine. As explained in https://github.com/Chatterino/chatterino2/issues/5694#issuecomment-2455665969, I filed [QTBUG-130861](https://bugreports.qt.io/browse/QTBUG-130861), but that won't be fixed until at least 6.8.2 (most likely - 6.8.1 is to be released on 21st of November). The issue doesn't affect the default configuration on Windows that uses _Segoe UI_ as it has a style for the 600 weight. Only fonts that don't have a style for the 600 weight are affected.

Currently, we don't gain much by using DirectWrite exclusively (note the DirectWrite is already used for some fonts that require it - it's also used in the linked issue), so I disabled that fontengine by default. Users can still opt-out by disabling `CHATTERINO_USE_GDI_FONTENGINE`.

I also removed the `dpiawareness=2` as this was never read, because the key was wrong (`WindowArguments` had the `s` in `Windows` missing).


cmake/resources/generate_resources.cmake
```diff
@@ -88,6 +88,15 @@ if (WIN32)
     list(APPEND RES_AUTOGEN_FILES "${CMAKE_BINARY_DIR}/autogen/windows.rc")
 endif ()
 
+set(WINDOWS_ARGUMENTS "@Invalid()") # empy QVariant() in QSettings
```
pajlada, line 91:
```suggestion
set(WINDOWS_ARGUMENTS "@Invalid()") # empty QVariant() in QSettings
```

cmake/resources/generate_resources.cmake
```diff
@@ -88,6 +88,15 @@ if (WIN32)
     list(APPEND RES_AUTOGEN_FILES "${CMAKE_BINARY_DIR}/autogen/windows.rc")
 endif ()
 
+set(WINDOWS_ARGUMENTS "@Invalid()") # empty QVariant() in QSettings
+if (CHATTERINO_USE_GDI_FONTENGINE AND Qt${MAJOR_QT_VERSION}_VERSION GREATER_EQUAL "6.8.0")
```
pajlada, line 92:
```suggestion
if (CHATTERINO_USE_GDI_FONTENGINE AND Qt${MAJOR_QT_VERSION}_VERSION VERSION_GREATER_EQUAL "6.8.0")
```

@Nerixyz  can you confirm this works for you, and then apply this change? see https://cmake.org/cmake/help/v3.7/command/if.html#command:if - iirc 6.8.1 is GREATER_EQUAL 6.8.0, and 6.10.0 is not GREATER_EQUAL 6.8.0

Nerixyz: Oh right i missed the version 🤦

pajlada (approved):

PR #5757, opened 2024-12-07
fix: initial scrollbar behaviour in inactive tab

pajlada:
Channel views could sometimes have their scrollbar be at the top instead
of bottom when they were first activated if it had messages in it

Scrollbars now assume their default behaviour is: "I'm scrolled down to the bottom" - this seems reasonable & fixes the issue


pajlada:
image

PR #5760, opened 2024-12-07
feat: Add a basic implementation of shared chat profile pictures

treuks:
This PR adds small(ish) profile pictures next to the badges of the users, to make it easy to tell from which chat the message came from at a quick glance:

image

The biggest problems with this implementation, and the ones which need to be fixed ASAP are:
* It stores the URL of the profile picture on the TwitchUser class for each user, which may or may not increase the ram usage.
* It uses a hard-coded replacement of the 300x300 size which is, while seemingly consistent across channels, just looks very bad.
* Quite a bit of code duplication.
* Some inconsistent variable naming.

Besides that, from the minimal amount of testing I've done, it seems to be pretty usable.


Felanbird:
functionally seems fine, the badge being larger than other badges makes me crave death

src/messages/MessageBuilder.cpp
```diff
@@ -385,8 +385,26 @@ EmotePtr makeAutoModBadge()
...
+
+        return std::make_shared<Emote>(Emote{
+            .name = EmoteName{},
+            .images = ImageSet{Image::fromUrl(
+                Url{modifiedUrl},
+                0.257142857143)},  // get as close to 18x18 as possible
+            .tooltip =
+                Tooltip{"Shared Message" +
+                        (sourceName.isEmpty() ? "" : " from " + sourceName)},
+            .homePage = Url{"https://link.twitch.tv/SharedChatViewer"},
```
Nerixyz, line 404: We could link to the channel here (but we need to pass the login name in that case).

src/messages/MessageBuilder.cpp
```diff
@@ -385,8 +385,26 @@ EmotePtr makeAutoModBadge()
         Url{"https://dashboard.twitch.tv/settings/moderation/automod"}});
 }
 
-EmotePtr makeSharedChatBadge(const QString &sourceName)
+EmotePtr makeSharedChatBadge(const QString &sourceName, const QString &url)
 {
+    if (!url.isEmpty())
+    {
+        QString modifiedUrl = url;
+
```
Nerixyz, line 393:
```suggestion
```

pajlada (approved):
Would love some eyes on this - from my perspective this works.

I don't know if it's worth trying to only store profile pictures for users that need it.

src/messages/MessageBuilder.cpp
```diff
@@ -385,8 +385,26 @@ EmotePtr makeAutoModBadge()
...
+        modifiedUrl.replace("300x300", "70x70");
```
Nerixyz, line 394: Twitch provides 64x64 as well as 28x28. We should use these to avoid scaling issues.

src/messages/MessageBuilder.cpp
```diff
@@ -385,8 +385,26 @@ EmotePtr makeAutoModBadge()
...
+
+        return std::make_shared<Emote>(Emote{
+            .name = EmoteName{},
+            .images = ImageSet{Image::fromUrl(
+                Url{modifiedUrl},
+                0.257142857143)},  // get as close to 18x18 as possible
```
Nerixyz, line 400:
```suggestion
                18.F/28.F)},  // get as close to 18x18 as possible
```

(or whatever the desired size is)

src/messages/MessageBuilder.cpp
treuks: do they? i tried putting 64x64 on my pfp and it gave me a 403 error

example:
<https://static-cdn.jtvnw.net/jtv_user_pictures/a577667e-2c71-4420-983d-9609aac088f2-profile_image-64x64.png>

src/messages/MessageBuilder.cpp
pajlada: Looks like we don't support the homePage thing on badges atm - right-clicking a badge doesn't give me any link options

src/messages/MessageBuilder.cpp
pajlada: Fixed in [`d99a439` (#5760)](https://github.com/Chatterino/chatterino2/pull/5760/commits/d99a439945097ae2b8aaaa0ddc48547066c2648b)

src/messages/MessageBuilder.cpp
pajlada: Fixed in [`dbde6f3` (#5760)](https://github.com/Chatterino/chatterino2/pull/5760/commits/dbde6f3d3ef732778ff3738a7afcbeb57fee0ddb)

src/messages/MessageBuilder.cpp
pajlada: In the mini bit of digging I've done, it looks like 28x28 has always worked.

Switched to 28x28 in [`d3f6587` (#5760)](https://github.com/Chatterino/chatterino2/pull/5760/commits/d3f65871ec33358573584d63d46859c8c743305f)

src/messages/MessageBuilder.cpp
pajlada: I updated the link in [`88b65e6` (#5760)](https://github.com/Chatterino/chatterino2/pull/5760/commits/88b65e61568d88ca7d110213df332de7cf9afbcd) anyway - won't try to fix the right-click action for badge elements in this PR though

src/messages/MessageBuilder.cpp
```diff
@@ -2959,22 +2987,32 @@ void MessageBuilder::appendTwitchBadges(const QVariantMap &tags,
...
         }
         else
         {
-            sourceName =
-                getApp()->getTwitchUsers()->resolveID({sourceId})->displayName;
+            auto twitchUser = getApp()->getTwitchUsers()->resolveID({sourceId});
+            sourceName = twitchUser->displayName;
+            sourceProfilePicture = twitchUser->profilePictureUrl;
+            sourceLogin = twitchUser->name;
         }
```
Nerixyz, line 3011: The last two branches are the same except for the name and the first one isn't necessary, because `QString`'s default constructor creates an empty string. Maybe something like this:

```cpp
if (!sourceId.isEmpty())
{
    auto twitchUser = getApp()->getTwitchUsers()->resolveID({sourceId});
    sourceProfilePicture = twitchUser->profilePictureUrl;
    sourceLogin = twitchUser->name;
    if (twitchChannel->roomId() == sourceId)
    {
        sourceName = twitchChannel->getName();
    }
    else
    {
        sourceName = twitchUser->getName();
    }
}
```

PR #5762, opened 2024-12-09
ci: build macOS on GitHub Actions

Nerixyz:
Currently, the released macOS builds are not from GitHub Actions. This PR adds back builds on macOS. This uses parts of the code from the local builder as well as adaptions I made for https://github.com/SevenTV/chatterino7.


pajlada:
I'm currently not interested in maintaining macOS builds in GH Actions

PR #5783, opened 2025-01-01
fix(bttv): use `channel` if no user was sent

### LINKED ISSUE #5781: BTTV owner emotes doesn't show the actual username
### Checklist

### Describe your issue

I added this emote today to my channel https://betterttv.com/emotes/676bfc34f9a9712d25b71cdf and it shows I'm the owner from the emote when in reality it's [EsperDG](https://betterttv.com/users/5fa6eddb4dfba16440294cfa)

Screenshot of the problem:
image
(note that if you F5 or restart chatterino, it will show the right owner, it's only when you add it in the moment)

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.5.2-beta.1 (commit becc1004) built on 2024-12-31 with Qt 6.7.1, MSVC 194234435, Crashpad Running on Windows 10 Version 22H2, kernel: 10.0.19045

Nerixyz:
Weird "bug"... For some messages in BTTV's WebSocket API, the `user` object of the emote is missing even though it should be there (i.e. the author isn't the current channel, and it's not some private emote). I reproduced it the first time, but after removing the emote and adding it back, the JSON included `user`. After that, adding any emote would always include `user`. Might be a BTTV bug.

If there isn't a `user` (the author is empty), we now fall back to the `channel`. This is the login name, not the display name, but that's the best we have.

Fixes #5781.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Bugfix: Fixed scrollbar highlights being visible in overlay windows. (#5769)
 - Bugfix: Make macos fonts look the same as v2.5.1. (#5775)
 - Bugfix: Fixed 7TV usernames messing with Qt's HTML (#5780)
+- Bugfix: Fixed BTTV emotes occasionally showing the wrong author. (#5783)
 - Dev: Hard-code Boost 1.86.0 in macos CI builders. (#5774)
 
 ## 2.5.2-beta.1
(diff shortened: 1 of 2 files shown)


pajlada (approved):

jupjohn:
Have you reported this upstream (BTTV)? Would be nice to get a proper fix

PR #5802, opened 2025-01-07
fix: don't force 96 dpi

### LINKED ISSUE #5411: Tab widths do not fit properly to text at 125% scale on Windows
### Checklist

### Describe your issue

Specifications:
- Font: Sogoe UI
- Font size: 10pt
- Zoom: 1.2x
- Screen resolution: 1920 x 1080
- Scale: 125%

Note: I did not make any changes to the configuration before or after updating Chatterino. I also noticed that changing the scale from 125% (default) to 100% fixes it.

Possible causes: #5404 #4868

### Screenshots

image

### OS and Chatterino Version

Chatterino Nightly 2.5.1 (commit 8ba57041) built on 2024-05-19 with Qt 6.5.0, MSVC 193933523 Running on Windows 11 Version 22H2, kernel: 10.0.22621

Nerixyz:
Setting [`AA_Use96Dpi`](https://doc.qt.io/qt-6/qt.html#ApplicationAttribute-enum) would fix fonts to a specific pixel/point size ([usage in Qt](https://github.com/search?q=repo%3Aqt%2Fqtbase+AA_Use96Dpi+NOT+path%3Atest&type=code)). We want the fonts to scale and use the scale provided by the OS. I only tested this on Windows.

TODO: list stuff to test

Fixes #5411.

### DIFF
--- a/src/RunGui.cpp
+++ b/src/RunGui.cpp
@@ -77,8 +77,6 @@ namespace {
     void initQt()
     {
         // set up the QApplication flags
-        QApplication::setAttribute(Qt::AA_Use96Dpi, true);
-
 #ifdef Q_OS_WIN32
         // Avoid promoting child widgets to child windows
         // This causes bugs with frameless windows as not all child events


pajlada:
Good catch!

I will test this on Linux (Wayland w/ sway) on 100% & 150% scale
I can test this on macOS on 100% scale & 200% scale

karl-police:
I tested this on Windows with

```bat
set QT_SCALE_FACTOR_ROUNDING_POLICY=Round
chatterino.exe
```

And the Font Size in the chat was the same as commit ``Chatterino Nightly 2.4.6 (commit 4a5d4ef3)`` which is an older build, where things are "okay".

However, things that didn't scale with this, were badges and a lot of other things. But the badges were smaller compared to on the Twitch Website.

My LogPixels is at 120.

Nerixyz:
> But the badges were smaller compared to on the Twitch Website.

You only showed a screenshot from Chatterino.

> This build from this merge check artifact

The badges in the second screenshot are 18x18 (original size) and in the first, they're 22x22, which is roughly $\frac{120}{96} \cdot 18 = 22.5$ We might need to look at the scaling there. Similar issues are likely with emote images.

> Turning Round off brings issues like these
>  [...]
> which alter as you drag the window...

This is a different issue - it's not the intention of this PR to fix this.

karl-police:
Badge screenshot from website
image

PR #5805, opened 2025-01-08
consolidate firefox incognitobrowser logic (and deprecate opera/launcher)

teknsl:
Mozilla’s firefox-nightly apt package now points to firefox-bin in its desktop file, catch all firefox variants by mozilla with one if statement

not sure what the changelog should be

### DIFF
--- a/src/util/IncognitoBrowser.cpp
+++ b/src/util/IncognitoBrowser.cpp
@@ -16,8 +16,8 @@ using namespace chatterino;
 QString getPrivateArg(const QString &exePath)
 {
     struct Entry {
-        QString exe{};
-        QString arg{};
+        QString exe;
+        QString arg;
     };
 
     static std::vector<Entry> lut{
@@ -48,7 +48,9 @@ QString getPrivateArg(const QString &exePath)
     for (const auto &entry : lut)
     {
         if (exe == entry.exe)
+        {
             return entry.arg;
+        }
     }
 
     // catch all mozilla distributed variants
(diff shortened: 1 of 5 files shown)


src/util/IncognitoBrowser.cpp
```diff
@@ -30,30 +34,31 @@ QString getPrivateSwitch(const QString &browserExecutable)
...
         {"brave", "-incognito"},
-        {"firefox-devedition", "-private-window"},
-        {"firefox-developer-edition", "-private-window"},
-        {"firefox-beta", "-private-window"},
-        {"firefox-nightly", "-private-window"},
     };
 
-    // compare case-insensitively
-    auto lowercasedBrowserExecutable = browserExecutable.toLower();
+    QString exe = QFileInfo(exePath).baseName().toLower();
```
pajlada, line 39: Undo the variable name change, and you can continue to use `auto`

pajlada (changes requested):
Just some nits, will test better later.

Changelog could be:

 - Minor: Treat all browsers starting with the filename `firefox` as Firefox browsers

src/util/IncognitoBrowser.cpp
```diff
@@ -30,30 +34,31 @@ QString getPrivateSwitch(const QString &browserExecutable)
         {"firefox-esr", "-private-window"},
```
Wissididom, line 34: Is there a reason why there were 4 firefox entries removed, but this `firefox-esr` is still there? Could it be that removing it was missed?

Mm2PL:
> Mozilla’s firefox-nightly apt package now points to firefox-bin in its desktop file, catch all firefox variants by mozilla with one if statement

How long will it take for all potential distributions to use this? When 
will this change land in stable packages?

pajlada (approved):

PR #5809, opened 2025-01-11
Update release docs

pajlada:
[skip ci]


8thony:
Two things would be useful to add, I think
- Something about winget publish action
- how to handle a bugfix release (like you did with 2.4.6 release)

PR #5824, opened 2025-01-14
fix: use `serverReceivedTime` over `parseTime` for stacking

Nerixyz:
[The tests are dying](https://github.com/Chatterino/chatterino2/actions/runs/12758619842/job/35560988846?pr=5823). [Twitch hates us](https://discuss.dev.twitch.com/t/legacy-pubsub-deprecation-and-shutdown-timeline/58043). The animals are leaving. The aliens aren’t contacting us. We might be alone. It just might be you and me. But that’s okay. Because do you really need anyone else?

Anyways, this replaces the use of `parseTime` with `serverReceivedTime`, because `parseTime` isn't set for regular messages (it's set to the current time, meaning our tests will be correct just once throughout the day).

I marked this as a draft, because I don't have time and this needs tests for the stacking of regular timeout messages. If you want to add some, feel free to push to this branch.

### DIFF
--- a/tests/snapshots/IrcMessageHandler/timeout-stack-never.json
+++ b/tests/snapshots/IrcMessageHandler/timeout-stack-never.json
@@ -1,0 +1,217 @@
+{
+    "input": "@target-user-id=141981764;ban-duration=600;historical=1;room-id=111448817;rm-received-ts=1736968967915;tmi-sent-ts=1736968967813 :tmi.twitch.tv CLEARCHAT #pajlada twitchdev",
+    "output": [
+        {
+            "badgeInfos": {
+            },
+            "badges": [
+            ],
+            "channelName": "",
+            "count": 3,
+            "displayName": "",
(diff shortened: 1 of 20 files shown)


tests/snapshots/IrcMessageHandler/timeout-stack-never.json
```diff
@@ -0,0 +1,217 @@
...
+                        "times)"
+                    ]
+                }
+            ],
+            "flags": "System|Timeout|DoNotTriggerNotification",
+            "id": "",
+            "localizedName": "",
+            "loginName": "",
+            "messageText": "twitchdev has been timed out for 10m. (3 times) ",
+            "searchText": "twitchdev has been timed out for 10m. (3 times) ",
```
Nerixyz, line 83: This seems wrong, but that's the current behavior.

pajlada (approved):

PR #5830, opened 2025-01-17
fix: ensure overrideScale_ is initialized before scale_

pajlada:
since we now read overrideScale_ from scale(), this order needs to be changed


pajlada:
No longer needed

Issue #5645, opened 2024-10-13
Border issues in nightly versus release (W11)


hemirt:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

If Windows 11 users could confirm how it appears for them, that would be nice, since this could just be some weird W11 setting combination.

windows 11 has these stupid borders around everything, they also change colors when window is selected or not (also whether its pinned to the side or not)
- some apps do NOT change these colors when active vs inactive (this would be preferred for me, would be nice to be a setting)

in stable it is that way, stays the same as active border whether the window is active or not
in nightly this is no longer that way, and it switches between inactive and active border states

the main issue is now that there is a small margin around chatterino nightly which looks bad, also the input box is smaller (for some reason the font in the input box too)

both unselected:
<img width="837" alt="{21051AE4-9F09-489D-8154-5F2A33170B81}" src="https://github.com/user-attachments/assets/8b934554-5166-4ba7-845b-25f9738dc66c">

stable selected:
<img width="836" alt="{C8DD56DB-2334-4F64-AC07-F25C156A2B57}" src="https://github.com/user-attachments/assets/6e77c20f-0ed5-49a8-b358-d2ecfd9fbc62">

nightly selected:
<img width="830" alt="{92229976-93CA-4D4E-9112-6EAE8F0F4442}" src="https://github.com/user-attachments/assets/38cb0a51-45fb-4047-aba1-f89b649b3c55">

as you can see, there is something off about the borders in nightly version, on windows 11

what can be problematic about this issue
- windows has like several settings affecting borders, border accents, colors, states, pin states, etc, so the way people see this can be different for everyone, so this may just be affecting a certain combination of settings that I have, so if anyone can also confirm if they see the same issues, that would be nice
- - that said, the chat input box should be the same (i.e. not filling up to the borders) for each user, if anyone could confirm <img width="300" alt="image" src="https://github.com/user-attachments/assets/d7070466-b9c1-4b10-8b2b-19dda0cf0b0f"> here I zoomed in (and selected the nightly version, so that both versions have the same active border color, stable is in the top right back, whilst nightly is in front bottom left)

[...]

PR #5836, opened 2025-01-19
fix: make borders on windows 11 thinner

Nerixyz:
It seems like Windows 11 automatically adds a border for us. If we manually add one, we make it thicker, which doesn't look nice (notice the thick border on the left).
Now, the border doesn't change its color when selected/unselected, but that seems to be the common way on Windows 11 (I tested Firefox).

✨ Comparison ✨ 

| Description | Before | After |
| ---|---|---|
|Selected|  |  |
| Unselected |  |  |

(I hope this fixes) https://github.com/Chatterino/chatterino2/issues/5645.


Nerixyz:
> but the chat box border still looks bad compared to the 2.5.1 release.

I think that's a separate issue. This PR only fixes the border of the application window.

For the split input, I don't know what the best solution is. Yes, I can remove the margins, but then multiple horizontal splits will look bad next to each other (notice: the margin is inside the _split_, not inside the _window_).

The question is basically: How should this look?



And... I don't know.

> Resizing the window just causes it to shift around. 🤔

Is that a separate bug with this PR? Resizing the window works as expected for me.

pajlada (approved):

PR #5839, opened 2025-01-19
chore(tests): print more errors when snapshot tests fail

pajlada:

### DIFF
--- a/tests/src/IrcMessageHandler.cpp
+++ b/tests/src/IrcMessageHandler.cpp
@@ -603,7 +603,11 @@ TEST_P(TestIrcMessageHandlerP, Run)
 
     delete ircMessage;
 
-    ASSERT_TRUE(snapshot->run(got, UPDATE_SNAPSHOTS));
+    ASSERT_TRUE(snapshot->run(got, UPDATE_SNAPSHOTS))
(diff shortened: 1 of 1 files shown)


pajlada (commented):
image

tests/src/IrcMessageHandler.cpp
```diff
@@ -603,7 +603,11 @@ TEST_P(TestIrcMessageHandlerP, Run)
 
     delete ircMessage;
 
-    ASSERT_TRUE(snapshot->run(got, UPDATE_SNAPSHOTS));
+    ASSERT_TRUE(snapshot->run(got, UPDATE_SNAPSHOTS))
+        << "Snapshot " << snapshot->name() << " failed. Expected JSON to be\n"
+        << QJsonDocument::fromVariant(snapshot->output().toVariant()).toJson()
```
Nerixyz, line 608: `fromVariant` can only create objects or arrays.
This should be something like

```cpp
QString debugRepr(const auto& it) {
    QString str;
    QDebug dbg(&str);
    dbg << it;
    return str;
}
```

Qt 6.9 will allow `QJsonValue.toJson()`.

PR #5850, opened 2025-01-23
feat: prefer visible/selected channels when joining

### LINKED ISSUE #3129: JOIN rate limit improvements
With the release of v2.3.4, we now join channels following a proper rate limit as suggested by Twitch.

Our solution was the bare minimum to get things going again, but there are a few issues and UX improvements left for us to make.

 - [x] Add a system message when we start joining a chat, update that message to "Joined" when we've confirmed that we've joined through the ROOMSTATE message
 - [ ] ~~Add a system where JOINs can be retried in case they have failed (i.e. if after sending the JOIN we do not receive a ROOMSTATE after a certain amount of time). This needs to be able to handle banned channels (they send a NOTICE with a specific msg-id), and nonexistant channels (they send nothing, so maybe a max retry? or helix call?)~~
 - [ ] Right now channels are joined in alphabetical order - we should have a prioritized queue where channels in your currently selected split are priotized

I'm making this mega-issue instead of splitting it out into separate issues as this work will be very tightly knit.

If you come up with anything else that's weird with how the current rate limiting is done, post a comment and I'll update the checklist

Nerixyz:
When we join channels because Chatterino (re-)connected, the join requests for channels that are currently selected are now put first. When starting Chatterino, this makes it seem like it's faster than before (even though it's as fast as before).

One downside of this PR is that `TwitchIrcServer` now knows about `WindowManager`, `Window`, `SplitNotebook`, `SplitContainer`, and `Split`.
An alternative would be to track the selected channels *somewhere* (i.e. add/remove them from methods in `Split` or `SplitNotebook`). I think that approach is more complex with little to no benefit.

- Fixes #3129 (it's the last item)

---

When searching for `joinBucket_` in `TwitchIrcServer`, you might find [another use](https://github.com/Chatterino/chatterino2/blob/da2384d695fa12e17b177b8372448435559070f5/src/providers/twitch/TwitchIrcServer.cpp#L1560). We don't call this when starting Chatterino, because the windows are initialized before the IRC connections (the connection isn't connected yet).

### DIFF
--- a/src/providers/twitch/TwitchIrcServer.cpp
+++ b/src/providers/twitch/TwitchIrcServer.cpp
@@ -24,8 +24,12 @@
 #include "providers/twitch/TwitchChannel.hpp"
 #include "singletons/Settings.hpp"
 #include "singletons/StreamerMode.hpp"
+#include "singletons/WindowManager.hpp"
 #include "util/PostToThread.hpp"
 #include "util/RatelimitBucket.hpp"
+#include "widgets/splits/Split.hpp"
+#include "widgets/splits/SplitContainer.hpp"
+#include "widgets/Window.hpp"
 
 #include <IrcCommand>
 #include <IrcMessage>
@@ -37,6 +41,7 @@
 #include <cassert>
 #include <functional>
 #include <mutex>
+#include <set>
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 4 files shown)


src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -935,31 +940,55 @@ void TwitchIrcServer::onReadConnected(IrcConnection *connection)
...
+
+        for (auto *split : page->getSplits())
         {
-            this->joinBucket_->send(channel->getName());
+            selected.emplace(split->getChannel().get());
         }
     }
 
+    // put the selected channels first
+    std::ranges::partition(activeChannels, [&](const auto &chan) {
```
pajlada, line 974: Could you swap to [std::ranges::stable_partition](https://en.cppreference.com/w/cpp/algorithm/ranges/stable_partition) to ensure we retain the relative order of the two segments? imo it's nice if you can somewhat know what order of the join of your inactive channels

src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -935,31 +940,55 @@ void TwitchIrcServer::onReadConnected(IrcConnection *connection)
...
+        {
+            continue;
+        }
+
+        for (auto *split : page->getSplits())
         {
-            this->joinBucket_->send(channel->getName());
+            selected.emplace(split->getChannel().get());
         }
     }
```
pajlada, line 971: I would prefer if `WindowManager` instead got a public function that returns a list of channel names that are visible (e.g. `getVisibleChannelNames`), and `TwitchIrcServer` uses those channel names to prioritize which channels to load

pajlada (changes requested):
Small nits - the `activeChannel` & lock change is a nice bonus refactor here

pajlada (approved):

PR #5854, opened 2025-01-24
fix: more force relayout on timeout/clear/delete

pajlada:
not optimized, but it means the message is hidden as soon as the message becomes disabled, instead of sometimes slightly after it becomes disabled

an alternative to the "force relayout always" is to return if any messages were disabled from the timeout/clearchat helpers

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -468,15 +468,16 @@ void IrcMessageHandler::handleClearChatMessage(Communi::IrcMessage *message)
     {
         chan->disableAllMessages();
         chan->addOrReplaceClearChat(std::move(clearChat.message), time);
-        return;
+    }
+    else
+    {
+        chan->addOrReplaceTimeout(std::move(clearChat.message), time);
     }
 
-    chan->addOrReplaceTimeout(std::move(clearChat.message), time);
-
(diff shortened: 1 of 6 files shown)


src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -468,15 +468,16 @@ void IrcMessageHandler::handleClearChatMessage(Communi::IrcMessage *message)
...
+    else
+    {
+        chan->addOrReplaceTimeout(std::move(clearChat.message), time);
     }
 
-    chan->addOrReplaceTimeout(std::move(clearChat.message), time);
-
-    // refresh all
-    getApp()->getWindows()->repaintVisibleChatWidgets(chan.get());
     if (getSettings()->hideModerated)
```
Nerixyz, line 477: There should be a check if these are historical messages. In that case we shouldn't force a layout as those messages aren't added to any view yet.

Nerixyz (changes requested):

src/providers/twitch/IrcMessageHandler.cpp
pajlada: As we investigated, this handle function isn't called for historical messages so this is not an issue. Thanks!

pajlada:
image

PR #5856, opened 2025-01-26
feat: `setHighlightSounds` & `openSubscriptionPage` split hotkeys

brian6932:
I noticed these hotkeys were missing from the split header context menu entries, so added them. I didn't add any default binds, wasn't sure if they'd be wanted (e.g. <kbd>CTRL</kbd>+<kbd>M</kbd> for a `setHighlightSounds` toggle).


src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -195,6 +195,18 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
...
+              .displayName = "Set highlight sounds",
+              .argumentDescription = "[on or off. default: toggle]",
+              .minCountArguments = 0,
+              .maxCountArguments = 1,
+              .possibleArguments = HOTKEY_ARG_ON_OFF_TOGGLE,
+              .argumentsPrompt = "New value:",
+              .argumentsPromptHover =
+                  "Should highlight sounds be enabled, disabled or toggled",
+          }},
+         {"subscribe", ActionDefinition{"Open subscription page"}},
```
Nerixyz, line 209: bikeshed: "subscribe" feels wrong (I know that's from the split header). Just from the name, I wouldn't know what it does (of course, if you know what it does, thinking about what you think it would do if you didn't know is hard). When I read the description, I wasn't sure what "subscribe" did. Maybe `openSubscriptionPage`?

src/singletons/Settings.hpp
```diff
@@ -677,11 +677,10 @@ class Settings
     bool isMutedChannel(const QString &channelName);
     bool toggleMutedChannel(const QString &channelName);
     std::optional<QString> matchNickname(const QString &username);
-
-private:
     void mute(const QString &channelName);
     void unmute(const QString &channelName);
```
Nerixyz, line 681: If you make these public, please ensure that the channel name only occurs once in the `mutedChannels` (e.g. when repeatedly calling `mute("foo")`).

src/widgets/splits/Split.cpp
```diff
@@ -807,7 +808,60 @@ void Split::addShortcuts()
...
+                     mode = 0;
+                 }
+                 else if (arg == "on")
+                 {
+                     mode = 1;
+                 }
+                 else
+                 {
+                     mode = 2;
+                 }
```
Nerixyz, line 837: `mode` is already initialized to 2.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -537,51 +537,69 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
...
+            h->getDisplaySequence(HotkeyCategory::Split, "subscribe"));
 
-        auto *action = new QAction(this);
-        action->setText("Notify when live");
-        action->setCheckable(true);
-
-        auto notifySeq = h->getDisplaySequence(
-            HotkeyCategory::Split, "setChannelNotification", {{"toggle"}});
-        if (notifySeq.isEmpty())
         {
```
Nerixyz, line 544: Please undo this. This makes the diff look more complicated.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -537,51 +537,69 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
...
 
-        moreMenu->addAction(action);
+            auto notifySeq = h->getDisplaySequence(
+                HotkeyCategory::Split, "setHighlightSounds", {{"toggle"}});
+            if (notifySeq.isEmpty())
+            {
+                notifySeq = h->getDisplaySequence(HotkeyCategory::Split,
+                                                  "setHighlightSounds",
+                                                  {std::vector<QString>()});
+                // this makes a full std::optional<> with an empty vector inside
```
Nerixyz, line 587: This was here before, but I don't think it should be here. If you want to keep it, move it one line up (so it's clear what this is referring to).

Nerixyz (commented):
The few unrelated changes could be done in another PR (possibly combined with passing hotkey arguments as `const&`).

src/widgets/splits/Split.cpp
brian6932: Sure the implementation was copied here.

src/widgets/splits/SplitHeader.cpp
brian6932: The code looked kinda confusing to me, as the blocks weren't doing logic isolation, had some extra code, and were repeating the condition. I'll put it in a separate cleanup PR, if you prefer that.

src/widgets/splits/SplitHeader.cpp
brian6932: Just to make sure, you're talking about the comment here, and not the whole `empty` branch?

Nerixyz: Yes, it's about the comment.

src/widgets/splits/Split.cpp
```diff
@@ -807,7 +808,59 @@ void Split::addShortcuts()
...
+        {"openSubscriptionPage", [this](const auto &) -> QString {
+             if (!this->getChannel()->isTwitchChannel())
+             {
+                 return "Cannot subscribe to a non-Twitch "
+                        "channel.";
+             }
+
+             this->openSubPage();
+             return "";
+         }}};
```
Mm2PL, line 863:
```suggestion
         }},
    };
```

pajlada (approved):

PR #5858, opened 2025-01-26
Add compact tabs to settings and implement them

### LINKED ISSUE #5807: More compact tab bar?
### Checklist

### Describe your issue

In the most recent update, the tab bar was made less compact, leading to my tab layout for open channels overflowing into anther line of tabs. It would be nice to have an option for a more compact tab bar with less empty space around channel names – either reverting to the previous interface design, or perhaps customizable (like a "normal" and a "compact" layout.

Screenshot 2025-01-09 at 10 20 18 AM

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.2 (commit f53d92c7) built with Qt 6.7.0 Running on macOS 15.0, kernel: 24.0.0

apa420:
Fixes #5807
I did find a bug regarding the tab width not adjusting properly when using left and right tab mode when switching this setting or the "Show tab close button". But the setting will take effect after either changing the layout location back and forth or restarting chatterino.

The numbers used here are what I found looks the best while keeping other settings such as the tab close button still looking good. I have not tested this on scaling, would probably be good if someone who has that tests this pr.
    
<pre>     default              compact         show tab close              show tab close compact</pre>

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -168,6 +168,7 @@ public:
     FloatSetting boldScale = {"/appearance/boldScale", 63};
     BoolSetting showTabCloseButton = {"/appearance/showTabCloseButton", true};
     BoolSetting showTabLive = {"/appearance/showTabLiveButton", true};
+    BoolSetting compactTabs = {"/appearance/compactTabs", false};
     BoolSetting hidePreferencesButton = {"/appearance/hidePreferencesButton",
                                          false};
     BoolSetting hideUserButton = {"/appearance/hideUserButton", false};
--- a/src/widgets/helper/NotebookTab.cpp
+++ b/src/widgets/helper/NotebookTab.cpp
@@ -20,7 +20,6 @@
 #include <boost/container_hash/hash.hpp>
 #include <QAbstractAnimation>
 #include <QApplication>
-#include <QDebug>
 #include <QDialogButtonBox>
 #include <QLabel>
 #include <QLinearGradient>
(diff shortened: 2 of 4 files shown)


8thony (approved):
Feature works as expected. I like it.
The bug you mentioned reproducible on Nightly and 2.5.2 should be fixed in a separate PR I think.

src/singletons/Settings.hpp
```diff
@@ -168,6 +168,7 @@ class Settings
     FloatSetting boldScale = {"/appearance/boldScale", 63};
     BoolSetting showTabCloseButton = {"/appearance/showTabCloseButton", true};
     BoolSetting showTabLive = {"/appearance/showTabLiveButton", true};
+    BoolSetting compactTabs = {"/appearance/compactTabs", false};
```
pajlada, line 171: Can you make the this an `EnumSetting<TabStyle>` instead?

Define `TabStyle` as
```c++
enum class TabStyle : std::uint8_t {
  Normal,
  Compact,
};
```

In the general page, use `SettingWidget` instead

```c++
SettingWidget::dropdown("Tab style", s.tabStyle)
  ->addTo(layout);
```


With this change, you might want to implement a few anonymous functions inside `NotebookTab.cpp` to get the relevant value based on the tab style.
Something like
```c++
float getCompactDivider(TabStyle tabStyle) {
  switch (tabStyle) {
    case TabStyle::Normal:
      return 1.5;
    case TabStyle::Compact:
      return 1.0;
  }
}
```

and same for the `compactReducer`

pajlada (changes requested):

src/singletons/Settings.hpp
Nerixyz: An option to have the tabs as wide as normal but with the tab close button hidden would be nice too. In the future, an option to stretch the tabs to the width of the window would be great.

pajlada: Yeah, I feel like there's more styles than just "Compact" that we might want to implement. touch-friendly tabs!

pajlada (approved):

jupjohn:
This feature is a nice touch for people like me who don't like too many rows of tabs. Thanks @apa420 👌

PR #5864, opened 2025-01-27
docs: update linux & test docs

pajlada:


apa420:
Mentioning the cmake flag when building for tests would be a nice addition imo

apa420 (approved):
lgmt :+1:

PR #5865, opened 2025-01-27
fix: make suspicious treatment updates searchable

iProdigy:
`searchText` was not populated for pubsub notif of (un)monitor/(un)restrict


pajlada (approved):
We could technically leverage `emplaceSystemTextAndUpdate` here but given the branching & usernames here I'm not sure if that really lowers complexity

PR #5870, opened 2025-01-29
feat: Added moderator actions into usercard

### LINKED ISSUE #3069: On Mod screen, delete message button on user popup
Enhancement:

When looking at a user's history to decide if I take moderation actions, I'd like the ability to delete a specific message and not just the top bar of larger actions. 

e.g. I want to delete just the comment circled in blue:

image

apa420:
Fixes #3069
Maybe look into making the usercard update on sword click.

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1403,11 +1403,17 @@ MessageElementFlags ChannelView::getFlags() const
             split = dynamic_cast<Split *>(searchPopup->parentWidget());
         }
 
-        auto *userInfoPopup = dynamic_cast<UserInfoPopup *>(
-            this->parentWidget()->parentWidget()->parentWidget());
-        if (userInfoPopup != nullptr)
+        if (this->parentWidget() != nullptr)
         {
-            split = userInfoPopup->getSplit();
+            if (this->parentWidget()->parentWidget() != nullptr)
+            {
+                auto *userInfoPopup = dynamic_cast<UserInfoPopup *>(
+                    this->parentWidget()->parentWidget()->parentWidget());
+                if (userInfoPopup != nullptr)
+                {
+                    split = userInfoPopup->getSplit();
+                }
+            }
         }
     }
 


src/widgets/helper/ChannelView.cpp
```diff
@@ -1402,6 +1402,13 @@ MessageElementFlags ChannelView::getFlags() const
         {
             split = dynamic_cast<Split *>(searchPopup->parentWidget());
         }
+
+        auto *userInfoPopup = dynamic_cast<UserInfoPopup *>(
+            this->parentWidget()->parentWidget()->parentWidget());
```
Mm2PL, line 1407: I'm here to let you know I strongly dislike this solution. I know that we use this all over the place so it can stay, however I will not be silent.

Also what if ChannelView's parent does not have a parent? This segfaults right? I think we need to check the entire painful chain for objects we expect.

pajlada: yeah I'm not super fond of this either, I'd rather we find some other solution to this that doesn't involve digging through 3 sets of parents, or relying on the original channels "mod view" state.
maybe a mod view button in the userchannelpopup itself to control whether or not it should be visible?

Mm2PL: We could have a global mod state for usercards and if it's opened in the context of a moderated channel it could show some moderation buttons, ideally excluding ban and timeout ones, so they wouldn't duplicate the ones already available in the usercard.

Nerixyz: > Also what if ChannelView's parent does not have a parent? This segfaults right?

This segfaults when opening the overlay window right now.

apa420: I'll make it check if each parentWidget() is a nullptr. That should should solve it right?

apa420: This should be solved now.

pajlada:
is there a scenario where we can't use `this->split_` for figuring out the flags?

PR #5874, opened 2025-01-30
feat: Search improvement

### LINKED ISSUE #5872: Improvements to hotkey settings
### Checklist

### Describe your issue

Add a few improvements to hotkeys
- [x] Search hotkeys - #5874 
- [x] Search hotkeys with a hotkey to see if it's being used - #5874
- [x] Find the hotkey or hotkey-bind name through global search - #5874
- [ ] Have a secondary keybinding for the hotkey (maybe allow for unlimited binds for the same key, but most programs only allow for 2-3 keybinds for the same command)
- [ ] The adding/editing window should be a popup instead of a normal window (annoying on tiling window managers)

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.2 DEBUG (commit 6f3cf284 modified) built with Qt 6.8.1 Running on Arch Linux, kernel: 6.12.10-arch1-1

apa420:
This pr adds the functionality of searching for fields on the tables we have in the settings pages. You can for example search for the name you gave a hotkey and it'll show up. Or you can search for "whisper" and find that you have a highlight for it in the highlights tab!

Partially fixes #5872

### DIFF
--- a/src/widgets/helper/EditableModelView.cpp
+++ b/src/widgets/helper/EditableModelView.cpp
@@ -150,26 +150,37 @@ void EditableModelView::addRegexHelpLink()
     this->addCustomButton(regexHelpLabel);
 }
 
-void EditableModelView::filterSearchResults(QString *query,
-                                            std::vector<int> *columnSelect)
+bool EditableModelView::filterSearchResults(const QString &query,
+                                            std::vector<int> &columnSelect)
 {
+    bool searchFoundSomething = false;
     auto rowAmount = this->model_->rowCount();
+
+    // make sure to show the page even if the table is empty,
+    // but only if we aren't search something
+    if (rowAmount == 0 && query.isEmpty())
+    {
+        return true;
+    }
+
     for (int i = 0; i < rowAmount; i++)
     {
-        tableView_->showRow(i);
+        tableView_->hideRow(i);
     }
-    for (int j : *columnSelect)
+    for (int j : columnSelect)
     {
         for (int i = 0; i < rowAmount; i++)
         {
             QModelIndex idx = model_->index(i, j);
             QVariant a = model_->data(idx);
-            if (!a.toString().contains(*query, Qt::CaseInsensitive))
+            if (a.toString().contains(query, Qt::CaseInsensitive))
             {
-                tableView_->hideRow(i);
+                tableView_->showRow(i);
+                searchFoundSomething = true;
             }
         }
     }
+    return searchFoundSomething;
 }
 
 void EditableModelView::filterSearchResultsHotkey(
--- a/src/widgets/helper/EditableModelView.cpp
+++ b/src/widgets/helper/EditableModelView.cpp
@@ -150,6 +150,51 @@ void EditableModelView::addRegexHelpLink()
     this->addCustomButton(regexHelpLabel);
 }
 
+void EditableModelView::filterSearchResults(QString *query,
+                                            std::vector<int> *columnSelect)
+{
+    auto rowAmount = this->model_->rowCount();
+    for (int i = 0; i < rowAmount; i++)
+    {
+        tableView_->showRow(i);
+    }
+    for (int j : *columnSelect)
+    {
+        for (int i = 0; i < rowAmount; i++)
+        {
+            QModelIndex idx = model_->index(i, j);
+            QVariant a = model_->data(idx);
+            if (!a.toString().contains(*query, Qt::CaseInsensitive))
+            {
+                tableView_->hideRow(i);
+            }
+        }
+    }
+}
+
+void EditableModelView::filterSearchResultsHotkey(
+    const QKeySequence *keySequenceQuery)
+{
+    auto rowAmount = this->model_->rowCount();
+    for (int i = 0; i < rowAmount; i++)
+    {
+        tableView_->hideRow(i);
+    }
+    for (int i = 0; i < rowAmount; i++)
+    {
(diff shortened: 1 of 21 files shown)


src/widgets/settingspages/CommandPage.cpp
```diff
@@ -87,16 +89,16 @@ CommandPage::CommandPage()
...
     auto layout = layoutCreator.setLayoutType<QVBoxLayout>();
 
-    auto *view = layout
-                     .emplace<EditableModelView>(
-                         getApp()->getCommands()->createModel(nullptr))
-                     .getElement();
+    view_ = layout
+                .emplace<EditableModelView>(
+                    getApp()->getCommands()->createModel(nullptr))
+                .getElement();
```
Nerixyz, line 95: It's easier to read the diff if this is in the form of
```cpp
auto *view = layout.emplace(...);
this->view_ = view;
```
Then you don't have to change the next lines. (same for the other pages - it makes the whole PR smaller)

src/widgets/settingspages/CommandPage.cpp
```diff
@@ -33,17 +33,17 @@ QString c1settingsPath()
     return combinePath(qgetenv("appdata"), "Chatterino\\Custom\\Commands.txt");
 }
 
-void checkCommandDuplicates(EditableModelView *view, QLabel *duplicateWarning)
+void checkCommandDuplicates(EditableModelView *view_, QLabel *duplicateWarning)
```
Nerixyz, line 36: This doesn't need to be renamed.

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -64,45 +65,49 @@ HighlightingPage::HighlightingPage()
                     "Message highlights are prioritized over badge highlights "
                     "and user highlights.");
 
-                auto *view =
+                viewMessages_ =
```
Nerixyz, line 68: This could be kept as `view` too (same reason as above).

src/widgets/settingspages/HighlightingPage.cpp
```diff
@@ -53,7 +53,8 @@ HighlightingPage::HighlightingPage()
         // getSettings()->enableHighlights));
 
         // TABS
-        auto tabs = layout.emplace<QTabWidget>();
+        this->tabWidget_ = new QTabWidget();
+        auto tabs = layout.append(tabWidget_);
```
Nerixyz, line 57:
```suggestion
        auto tabs = layout.emplace<QTabWidget>();
        this->tabWidget_ = tabs;
```

src/widgets/helper/EditableModelView.cpp
```diff
@@ -150,6 +150,62 @@ void EditableModelView::addRegexHelpLink()
...
+                tableView_->showRow(i);
+                searchFoundSomething = true;
+            }
+        }
+    }
+    return searchFoundSomething;
+}
+
+void EditableModelView::filterSearchResultsHotkey(
+    const QKeySequence *keySequenceQuery)
```
Nerixyz, line 187: Could be a const-ref - the pointer implies that it could be `nullptr`.

src/widgets/helper/EditableModelView.cpp
```diff
@@ -150,6 +150,62 @@ void EditableModelView::addRegexHelpLink()
     this->addCustomButton(regexHelpLabel);
 }
 
+bool EditableModelView::filterSearchResults(const QString &query,
+                                            std::vector<int> &columnSelect)
```
Nerixyz, line 154: This should be a `std::span<const int> columnSelect` (since we use C++ 20).
That would allow callers to pass a `std::array`:

```cpp
bool AccountsPage::filterElements(const QString &query)
{
    std::array fields{0, 1};

    return view_->filterSearchResults(query, fields);
}
```

<sub>🤓 Technically, the query could be a `QStringView` too, but we use `const QString&` everywhere, so it's fine.</sub>

src/widgets/settingspages/CommandPage.cpp
apa420: I dislike the idea of having the same variable have two names in the code. Especially since one of them explicitly through naming convention states that it's a class variable and probably will have impact somewhere else in the code.

src/widgets/settingspages/HighlightingPage.cpp
apa420: This does not work. Using emplace creates a `LayoutCreator<QTabWidget>` which was exactly the reason for using append in this instance.
You can do `this->tabWidget_ = &*tabs;` but for readability I prefer the previous version.

src/widgets/settingspages/CommandPage.cpp
Nerixyz: > Especially since one of them explicitly through naming convention states that it's a class variable

We have a [convention of referring to class members through `this`](https://wiki.chatterino.com/Contributing%20for%20Developers/#this).

apa420: Good catch, I seem to have been overly aggressive with regex replace

pajlada (commented):
Not a full review

I like the Hotkeys "Search keybind" functionality
I like that I can CTRL+F to find a hotkey name
I don't like that I can CTRL+F to find the hotkey keybinding
I don't like it searching through accounts/nicknames/highlights/ignoresfilters/logs
I'm torn on searching through commands with this

apa420:
Closed in favor of #5884

PR #5875, opened 2025-01-30
Alt hotkeys [Feedback wanted]

apa420:
#5872 
I would appreciate feedback on this (not on the code rn). This would basically add the possiblity of a alternate keybind to an already existing action. This would make creating more keybinds for the same action easier and also allow us to set multiple keybinds for the same default action more intuitavely (see shift+enter and ctrl+shift+enter).


Felanbird:
nitpick nothing to do with the code: user-facing references to alternative/alternate should not be shortened, since `alt` is a valid key press

apa420:
Is there a request to have infinite keybinds per hotkey action or is just 2 enough?

Issue #1000, opened 2019-04-10
Rewrite the toasts implementation for crossplatform


apa420:
The current implementation of toasts in chatterino is currently not that good since it doesn't support very basic features and isn't crossplatform which chatterino strives to be.

pajlada:
Toasts currently work on Linux and Windows, but we lack macOS support.
I think the correct approach going forward is to implement macOS support through another library or whatever we need, not rewrite it to use a cross platform library.

PR #5881, opened 2025-02-01
feat: add Linux support for Live Notifications toasts w/ libnotify

### LINKED ISSUE #5782: Notifications don't work on Ubuntu  22.04 lts
### Checklist

### Describe your issue

Hi and hello, happy new year to you. I downloaded and installed from your site (https://chatterino.com/linux), but I don't get a desktop notification when a channel goes live. Is there anything I can do? 

### Screenshots

_No response_

### OS and Chatterino Version

2.5.1

camporter:
Add libnotify notification support for channels going live. Would be nice to have notifications for other things as well, but this is just adding an equivalent for wintoast.

Disabled QT keywords to not clash with glib :disappointed: there's probably a better way to do this.

Fixes #5782

There's a lot of testing that needs to be done and build concerns, so leaving as a draft.


Nerixyz:
> Disabled QT keywords to not clash with glib 😞 there's probably a better way to do this.

Could you do this in another PR? This makes the diff way larger than it actually is. You can remove `emit`s (imo) - they're cosmetic and don't do anything.

pajlada:
Hi! Thanks for your initial PR, I just wanted to let you know that the use of `libnotify` seems like a good idea from my perspective.

8thony:
- Towards #1000

killertofus:
the aur version of [chatterino](https://aur.archlinux.org/packages/chatterino2-git) will also need to include libnotify as a dep then

camporter:
Thanks, happy to provide a PKGBUILD patch but it should just be `libnotify` in the deps.

Missed nixos in the docs so updated.

camporter:
I have some Highlight -> Toast behavior done 
image

but that should probably wait for the live notification and build issues to be resolved first.

pajlada:
@camporter I would prefer if desktop notifications for highlights is not added in this PR - It's too much of a separate feature

If you want to add some way to test desktop notifications easier, feel free to add an additional branch to the `/debug-test` command

camporter:
Thanks, added a debug option for notifications and tried a variety of random channels. 
Screenshot From 2025-02-05 19-38-38

Remaining things:
* The linux builds I believe are breaking because the container images will need an update.
* For freebsd, considering leaving libnotify off..

Wissididom:
For the container update you could make a PR in https://github.com/Chatterino/docker where you update `Dockerfile_chatterino2-build-ubuntu-20.04`, `Dockerfile_chatterino2-build-ubuntu-22.04` and `Dockerfile_chatterino2-build-ubuntu-24.04`. I think from what I've read in the error message that it might be enough to just install libnotify or smth in those container images. After pajlada than merges the PR in Chatterino/docker and you wait until the build on master is complete, the new docker images should be available for pulling/in the workflow.

camporter:
Ah thanks, confused the separate repo with the `.docker` dir.  Opened a PR over there :slightly_frowning_face:

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Minor: When (re-)connecting, visible channels are now joined first. (#5850)
 - Minor: Added the ability to filter on messages by the author's user ID (example: `author.user_id == "22484632"`). (#5862)
 - Minor: Improved error messaging of the `/clip` command. (#5879)
+- Minor: Add linux desktop live notification support (#5881)
```
pajlada, line 12:
```suggestion
- Minor: Added Linux support for Live Notifications toasts. (#5881)
```

src/CMakeLists.txt
```diff
@@ -1062,6 +1062,15 @@ if (UNIX)
...
         target_link_libraries(${LIBRARY_PROJECT} PUBLIC ${CMAKE_DL_LIBS})
     endif ()
+
+    if (NOT APPLE AND BUILD_WITH_LIBNOTIFY)
+        find_package(PkgConfig REQUIRED)
+        pkg_check_modules(LIBNOTIFY REQUIRED libnotify)
+        include_directories(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_INCLUDE_DIRS})
+        target_link_libraries(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_LIBRARIES})
+        target_link_directories(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_LIBRARY_DIRS})
+        target_compile_definitions(${LIBRARY_PROJECT} PUBLIC LIBNOTIFY)
```
pajlada, line 1072: Rename the `LIBNOTIFY` definition to `CHATTERINO_WITH_LIBNOTIFY`

pajlada (changes requested):
Small nits, otherwise seems good

src/CMakeLists.txt
```diff
@@ -1062,6 +1062,15 @@ if (UNIX)
         message(STATUS "Linking with CMake DL libs: '${CMAKE_DL_LIBS}'")
         target_link_libraries(${LIBRARY_PROJECT} PUBLIC ${CMAKE_DL_LIBS})
     endif ()
+
+    if (NOT APPLE AND BUILD_WITH_LIBNOTIFY)
+        find_package(PkgConfig REQUIRED)
+        pkg_check_modules(LIBNOTIFY REQUIRED libnotify)
+        include_directories(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_INCLUDE_DIRS})
+        target_link_libraries(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_LIBRARIES})
+        target_link_directories(${LIBRARY_PROJECT} PUBLIC ${LIBNOTIFY_LIBRARY_DIRS})
```
Nerixyz, line 1071: I'm pretty sure we can link to `libnotify` as `PRIVATE` as we only use it internally (not exposed in any header or through other functions). We should define `CHATTERINO_WITH_LIBNOTIFY` as `PUBLIC` however, as it changes the layout of `Toasts`.

src/singletons/Toasts.cpp
```diff
@@ -282,6 +295,38 @@ void Toasts::sendWindowsNotification(const QString &channelName,
...
+{
+    if (this->initialized_)
+    {
+        return;
+    }
+    auto result = notify_init("chatterino2");
+
+    if (result == 0)
+    {
+        qCDebug(chatterinoNotification) << "Failed to initialize libnotify";
```
Nerixyz, line 310:
```suggestion
        qCWarning(chatterinoNotification) << "Failed to initialize libnotify";
```

src/singletons/Toasts.hpp
```diff
@@ -35,6 +35,11 @@ class Toasts final
     void sendWindowsNotification(const QString &channelName,
                                  const QString &channelTitle, Platform p);
 
+    bool initialized_ = false;
+#elif defined(LIBNOTIFY)
+    void ensureInitialized();
+    void sendLibnotify(const QString &channelName, const QString &channelTitle);
+
     bool initialized_ = false;
```
Nerixyz, line 43: Could `ensureInitialized` and `initialized_`  be defined in a single block? They're the same on Windows and with libnotify. Maybe the `send*` functions could be united as well. Something like `sendPlatformNotification`? It would include the platform for now (https://github.com/Chatterino/chatterino2/pull/5914).

Nerixyz (commented):
I wonder if we could make our own small notification library (I'd be fine with a submodule here). Implementing this for macOS doesn't seem to be that complicated either ([KNotifications impl](https://github.com/KDE/knotifications/blob/85e71b58fba5b978cb8493926fff640de2480020/src/notifybymacosnotificationcenter.mm) - ObjectiveC might be the hardest part here). Maybe after EventSub is in a decent state.

src/CMakeLists.txt
pajlada: If this works that's fine :+1:

src/singletons/Toasts.hpp
camporter: I left these separate just to avoid getting into too much rework here, but combining them or moving them into a library seems good. The macos notification stuff required signing and other conditions that I didn't wade into. I don't boot windows up very often to test there either :laughing: 

For linux/freebsd, you could also avoid libnotify and use dbus directly, but a bit more involved to do.

pajlada:
Looks like you can apply this patch to fix the FreeBSD build, and it also just looks simpler

```patch
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 24ddbe7b..05a83ebd 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1065,10 +1065,8 @@ if (UNIX)
 
     if (NOT APPLE AND BUILD_WITH_LIBNOTIFY)
         find_package(PkgConfig REQUIRED)
-        pkg_check_modules(LIBNOTIFY REQUIRED libnotify)
-        include_directories(${LIBRARY_PROJECT} ${LIBNOTIFY_INCLUDE_DIRS})
-        target_link_libraries(${LIBRARY_PROJECT} PRIVATE ${LIBNOTIFY_LIBRARIES})
-        target_link_directories(${LIBRARY_PROJECT} PRIVATE ${LIBNOTIFY_LIBRARY_DIRS})
+        pkg_check_modules(LIBNOTIFY REQUIRED IMPORTED_TARGET libnotify)
+        target_link_libraries(${LIBRARY_PROJECT} PRIVATE PkgConfig::LIBNOTIFY)
         target_compile_definitions(${LIBRARY_PROJECT} PUBLIC CHATTERINO_WITH_LIBNOTIFY)
     endif ()
 endif ()
```

pajlada (approved):
Notifications works for me, valgrind doesn't complain about anything (new) - Thank you!

PR #5882, opened 2025-02-02
Disable QT keywords to avoid clashes with other libraries.

camporter:
To be able to add libnotify support in #5881, QT keywords conflict with glib and other libraries so try disabling them.

Not sure what the feeling is on including/excluding `emit` and not using keywords. It avoids some ugliness having to undef things and redef them :slightly_smiling_face:

### DIFF
--- a/src/common/SignalVectorModel.hpp
+++ b/src/common/SignalVectorModel.hpp
@@ -182,7 +182,7 @@ public:
 
             QVector<int> roles = QVector<int>();
             roles.append(role);
-            dataChanged(index, index, roles);
+            this->dataChanged(index, index, roles);
         }
 
         return true;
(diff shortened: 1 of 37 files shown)


src/common/SignalVectorModel.hpp
```diff
@@ -182,7 +182,7 @@ class SignalVectorModel : public QAbstractTableModel,
 
             QVector<int> roles = QVector<int>();
             roles.append(role);
-            emit dataChanged(index, index, roles);
+            dataChanged(index, index, roles);
```
Nerixyz, line 185: It would be great if these could be prefixed with `this->` to indicate that they're class-methods.

Nerixyz (commented):
Thank you, now we can use `emit`, `slots`, and `signals` in member names! 🎉

pajlada (approved):

PR #5888, opened 2025-02-02
feat: Search ignorespage [5/9]

apa420:
Filters columns 0 and only for messages


pajlada:
I don't want ignores to be searchable through CTRL+F. If we want to implement a separate search as part of the Ignores list itself that's fine, but I'd like there to be justification for adding it first.

PR #5889, opened 2025-02-02
feat: Search highlightingpage [6/9]

apa420:
Filters columns 0 in all tables


pajlada:
I'm looking to make some heavier changes to the highlights page soon :tm: so I don't want this as a temporary feature.

PR #5892, opened 2025-02-02
feat: Search accountspage [9/9]

apa420:
Filters columns 0 and 1


pajlada:
Not interested in having the accounts page be searchable

PR #5896, opened 2025-02-02
fix: force quit miniaudio thread after 1s

pajlada:
tested by not destroying the work guard


pajlada:
image

apa420:
I noticed issues with this yesterday after getting a live notification sound and closing chatterino after. Would take up to 30s for the miniaudio to close and release the terminal.

PR #5900, opened 2025-02-04
added libnotify as a dep

killertofus:
when #5881 gets merged into the main branch, you will need to have libnotify as a dep for building


pajlada:
The linked pr already updates these I think, you're probably better off PRing to their PR or adding a comment to it with your feedback is they've missed something

PR #5903, opened 2025-02-05
refactor(eventsub): enforce static error categories

Nerixyz:
Pop quiz: is there a requirement for the `error_category` passed to the `error_code` constructor ([docs](https://www.boost.org/doc/libs/1_87_0/libs/system/doc/html/system.html#ref_constructors_2))?

---

Doesn't look like it. [`boost::system::error_category`](https://www.boost.org/doc/libs/1_87_0/libs/system/doc/html/system.html#ref_boostsystemerror_category_hpp) doesn't say anything either (maybe I missed something). Well, turns out, `error_category` instances are treated like singletons. So we shouldn't construct them at runtime. One additional constraint we'd like to enforce is that they're supposed to be constant initialized (without the implicit lock inserted by the compiler for `static`s). We can achieve both of these by constructing a single global instance of the category and by using `std/boost::source_location`.

Initially, I had some funky workarounds for not being able to pass strings through the code (first commit). However, I realized that the location in the file where an error was emitted from contains the same (if not more) information. This is also simpler, because it's encouraged by `boost::system::error_code`.


pajlada:
This type of error seems fine to me :+1: 

```
Missing required key (/home/pajlada/git/chatterino/chatterino2/lib/twitch-eventsub-ws/src/generated/payloads/channel-moderate-v2.cpp:228:53 in function 'boost::json::result_for<Vip, boost::json::value>::type chatterino::eventsub::lib::payload::channel_moderate::v2::tag_invoke(boost::json::try_value_to_tag<Vip>, const boost::json::value &)')
```

pajlada:
Since we use generated code, we don't have to use the macro

pajlada (approved):

PR #5904, opened 2025-02-06
refactor(eventsub): Allow for skipping generation of json implementations

pajlada:
This can be done using the `SKIP_JSON_GENERATION` flag (off by default)


pajlada:
image

PR #5932, opened 2025-02-08
feat(eventsub): Add support for `channel.moderate` Warn

pajlada:
Unrelated to Warn, this change makes us handle unset & null vectors as an empty vector, and null strings as an empty string


pajlada:
image

PR #5933, opened 2025-02-08
chore: disable Qt5 builds in CI

pajlada:
The intent of this is not to remove every single ifdef qt version thing we have, but instead to stop building for Qt5 in CI.

New code we write should not have to pass Qt5 CI to be accepted in the codebase.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,6 @@ concurrency:
 env:
   C2_ENABLE_LTO: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/bugfix-release/') || startsWith(github.ref, 'refs/heads/release/') }}
   CHATTERINO_REQUIRE_CLEAN_GIT: On
-  C2_BUILD_WITH_QT6: Off
   CONAN_VERSION: 2.11.0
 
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -143,7 +142,7 @@ jobs:
         include:
           # macOS
           - os: macos-13
```
Nerixyz, line 144: Could we bump this to `macos-latest` too? iirc, the reason we were using `macos-13` was because Qt5 didn't work on ARM macs.

Wissididom: I think the reason for macos-13 was some C++20 code not working in https://github.com/Chatterino/chatterino2/pull/4970 (See third bullet point of description)
I don't remember which one, tho.

Nerixyz: That was because [`macos-latest` resolved to `macos-12`](https://web.archive.org/web/20231123025253/https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners) at the time. So we should be fine with using latest.

pajlada: Will do this in a follow-up PR

PR #5943, opened 2025-02-09
feat(eventsub): properly unsubscribe once no more handles are interested

pajlada:
Does some smaller refactors too, removing some xds and some xDs and even some XDs


pajlada (commented):
image

PR #5952, opened 2025-02-12
feat(eventsub): add basic parse tests and benchmark

### LINKED ISSUE #5918: Add unit tests (we already have `googletest` in the repository)

Nerixyz:
Adds `twitch-eventsub-ws-test` and `twitch-eventsub-ws-benchmark`. I added some messages we used in https://github.com/twitch-rs/twitch_api (as we know those can be received compared to the examples in the docs). The tests already found some issues in our bindings (imo, `std::optional<bool>` should default to `false`, but that's a task for the generator).

In the tests, we could use gmock to test that the handlers are actually called. `handleMessage` provides a good interface to test the parsing pipeline.

Closes #5918


pajlada (approved):
Looks great, thank you!

PR #5956, opened 2025-02-15
feat: delete messages from context menu

Nerixyz:
> user: how do you delete messages on chatterino?

Chatterino has many moderator focused features, but you can't delete a message from the context menu without reading through the settings or knowing that `/delete` exists. I don't see a reason why there shouldn't be a button if the user has moderator permissions.


pajlada (approved):

pajlada:
> I find myself accidentally clicking this all the time when I mean to reply. Would it be possible to get more control over it's display?

Sure, I'll add a setting to disable the option

pajlada:
@VirtuallyJesse with #6100 which will be included in the 2.5.3 release you can disable the feature with this setting
image

PR #5959, opened 2025-02-16
fix: opening a usercard from `/live` no longer makes failing ivr API call

kyleScarmack:
When the channelName is "/live", the API call fails because "/live" is not a valid channelName. To fix this, replace "/live" with the userName, which is the actual channelName.

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -945,9 +945,9 @@ void UserInfoPopup::updateUserData()
         this->ui_.block->setEnabled(true);
         this->ui_.ignoreHighlights->setChecked(isIgnoringHighlights);
 
-        auto type = this->channel_->getType();
+        auto type = this->underlyingChannel_->getType();
 
-        if (type != Channel::Type::TwitchLive)
+        if (type == Channel::Type::Twitch)
         {
             // get followage and subage
             getIvr()->getSubage(
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -945,44 +945,49 @@ void UserInfoPopup::updateUserData()
         this->ui_.block->setEnabled(true);
         this->ui_.ignoreHighlights->setChecked(isIgnoringHighlights);
 
-        // get followage and subage
-        getIvr()->getSubage(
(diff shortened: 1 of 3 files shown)


pajlada (commented):
Will this end up showing the streamer's subage & followage for their own channel? (e.g. click `forsen` and it will show `forsen`'s subage to `forsen`?)

To me, that information doesn't feel super useful - do you think it would make sense to not make the request at all in UserInfoPopup.cpp if the request comes from the `/live` tab?

kyleScarmack:
Yes, this will end up showing the streamer's subage to their own channel, whicn indirectly shows how long they have been affiliated or partnered. If you do not think it is super useful, I can not make the request at all in UserInfoPopup.cpp.

kyleScarmack:
Added a check to ensure type != Channel::Type::TwitchLive before retrieving followage and subage, allowing the function to proceed only when the condition is met.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -945,44 +945,49 @@ void UserInfoPopup::updateUserData()
...
 
-        // get followage and subage
-        getIvr()->getSubage(
-            this->userName_, this->underlyingChannel_->getName(),
-            [this, hack](const IvrSubage &subageInfo) {
-                if (!hack.lock())
-                {
-                    return;
-                }
+        auto type = this->channel_->getType();
```
Nerixyz, line 948:
```suggestion
        auto type = this->underlyingChannel_->getType();
```

`channel_` and `underlyingChannel_` can be different. I think you could check for `type == Channel::Type::Twitch`. Any other channel will have a bad name like `/live`, `/watching`, `/automod`, etc. (note: this doesn't prevent loading in usercards from these channels completely - they will set the channels appropriately where possible). Just in the live channel, there's no reasonable channel to set.

kyleScarmack:
Fixed as suggested.

pajlada:
Looks good to me @kyleScarmack - do you have a formatter set up in your IDE to fix the linting issues, or would you prefer I fix it for you?

kyleScarmack:
If you could fix it for me that would be great. I will learn for next time though.

pajlada (approved):

pajlada:

If you want this, you can open a new PR where you modify the `resources/contributors.txt` file and add yourself to the list. Read the comments at the top of the file for instructions.

Issue #5380, opened 2024-05-05
Split input doesn't scale down when Chatterino scales down


pajlada:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When zooming in & out of Chatterino, the height of the split input will change with it. Sometimes when doing this, the height will not change. Might be related to having multiple tabs open.
I don't have clear reproduction steps

### OS and Chatterino Version

Chatterino 2.5.1 DEBUG (commit 6f489b791 modified) built with Qt 6.7.0 Running on Arch Linux, kernel: 6.8.9-arch1-1

Mm2PL:
Might happen when changing DPI (`xrandr --output $YOUROUTPUT --dpi 500`). I think I've seen this happen when messing with DPI settings

4rneee:
For me the placeholder text in the split input (i.e. "Send message as xyz...") does not scale when zooming in/out until switching to a different tab.  
This might be a different issue and I can create one if you would like me to.

After taking a look at the code I am not sure if this might be a issue with QT as you set the font for the `textEdit` widget:
https://github.com/Chatterino/chatterino2/blob/bdd903bb42ee277d016a1acbcb6f40a336d76d57/src/widgets/splits/SplitInput.cpp#L236-L237
which works for the plain text but not the placeholder.

pajlada:
> For me the placeholder text in the split input (i.e. "Send message as xyz...") does not scale when zooming in/out until switching to a different tab. This might be a different issue and I can create one if you would like me to.
> 
> After taking a look at the code I am not sure if this might be a issue with QT as you set the font for the `textEdit` widget:
> 
> [chatterino2/src/widgets/splits/SplitInput.cpp](https://github.com/Chatterino/chatterino2/blob/bdd903bb42ee277d016a1acbcb6f40a336d76d57/src/widgets/splits/SplitInput.cpp#L236-L237)
> 
> Lines 236 to 237 in [bdd903b](/Chatterino/chatterino2/commit/bdd903bb42ee277d016a1acbcb6f40a336d76d57)
>  this->ui_.textEdit->setFont( 
[...]

PR #5960, opened 2025-02-16
fix: update the input style sheet when scaling

Nerixyz:
From https://github.com/Chatterino/chatterino2/issues/5380#issuecomment-2661376330 (not sure if it fixes the issue).

We set a style sheet in the input. Apparently, this prevents fonts from being applied to that element's children. When setting the font, it [always takes the parent font](https://github.com/qt/qtbase/blob/cf2f10268461d7e0dbd8331fbd1c83bf66c4b8c6/src/widgets/styles/qstylesheetstyle.cpp#L6489-L6490) for some reason. Previously, the font size adjustment only worked because we [always call `themeChangedEvent` when showing a widget](https://github.com/Chatterino/chatterino2/blob/bdd903bb42ee277d016a1acbcb6f40a336d76d57/src/widgets/BaseWidget.cpp#L162) and we [set the style there](https://github.com/Chatterino/chatterino2/blob/bdd903bb42ee277d016a1acbcb6f40a336d76d57/src/widgets/splits/SplitInput.cpp#L258).

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -235,6 +235,7 @@ void SplitInput::scaleChangedEvent(float scale)
     }
     this->ui_.textEdit->setFont(
         app->getFonts()->getFont(FontStyle::ChatMedium, scale));
+    this->ui_.textEdit->setStyleSheet(this->theme->splits.input.styleSheet);
     this->ui_.textEditLength->setFont(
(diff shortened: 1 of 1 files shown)


pajlada (approved):
With this change, the text resizes together with the scale :+1: 
It does not fully fix the issue but it does fix the comment you linked to

PR #5971, opened 2025-02-22
feat(notifications): show user avatar when sending live notification on Linux

### LINKED ISSUE #5938: Notifications (Toasts) should show the user Avatar
The libnotify implementation doesn't add the avatar image to the toast.

4rneee:
Add avatar images to live notifications in the libnotify implementation.
Fixes #5938 

There are two things I am unsure about:
1. do I need to check for `hasAvatarForChannel(channelName)` or can I assume this will be true once `sendLibnotify` is called?
2. Is `g_object_unref(img)` the correct way of 'freeing' the buffer? I copied it from [here](https://gitlab.gnome.org/GNOME/libnotify/-/blob/master/tests/test-image.c?ref_type=heads#L97-L98).


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Minor: Improved error messaging of the `/clip` command. (#5879)
 - Minor: Added Linux support for Live Notifications toasts. (#5881)
 - Minor: Messages can now be deleted from the context menu in a channel. (#5956)
+- Minor: Show user avatar in Live Nofications on Linux. (#5971)
```
Nerixyz, line 14: This should be added to the original changelog entry (as we didn't have a release with notifications on Linux yet).

Nerixyz (commented):
> 1. do I need to check for `hasAvatarForChannel(channelName)` or can I assume this will be true once `sendLibnotify` is called?

You shouldn't need it. `gdk_pixbuf_new_from_file` should error if the file doesn't exist when it tries to open it.

> 2. Is `g_object_unref(img)` the correct way of 'freeing' the buffer? I copied it from [here](https://gitlab.gnome.org/GNOME/libnotify/-/blob/master/tests/test-image.c?ref_type=heads#L97-L98).

That should be correct. `notify_notification_set_image_from_pixbuf` [stores a reference and increments the refcount](https://gitlab.gnome.org/GNOME/libnotify/-/blob/570982f616838abba6bdd8ca2bdb2f32f3b1d1de/libnotify/notification.c#L1413-1414), so one unref on our side is needed.

pajlada (approved):
This works for me :+1: thank you!

PR #5974, opened 2025-02-22
chore(eventsub): add operator== and indicator for shared chat

Nerixyz:
To not confuse moderators, we should ignore moderation events when using shared chat (I believe that's what we're implicitly doing in PubSub as well).


iProdigy:
i don't think we do this in pubsub, but maybe we could add this suffix to the moderation messages: ` for the duration of the shared chat session`

e.g., sharedMod banned badGuy for the duration of the shared chat session

pajlada:
After claryfing with @iProdigy  - moderation actions that happen in the other room during a shared chat session _do_ take effect in your chat, so I think it would make sense to show them still, but maybe with an extra piece of text attached to it like they mentioned above

Nerixyz:
> moderation actions that happen in the other room during a shared chat session _do_ take effect in your chat,

There's little to no documentation on this well-thought-out feature. Do emoteonly and friends apply as well? Also, what's the purpose of having `shared_chat_*` then?

iProdigy:
yeah i agree it's not adequately explained in the dev docs

the user facing docs do explain it with a complicated table https://link.twitch.tv/SharedChatMod

also ive written my own explainer at https://twitch4j.github.io/chat/shared-chat/

back to your question: roomstate does not apply to other channels, neither do automod settings or blocked/permitted terms

pajlada (approved):

iProdigy:
sorry if i wasn't clear earlier: we only have to annotate (un)ban/(un)timeout - warn/(un)mod/(un)timeout/(un)raid/etc are not mirrored (and `shared_chat_delete` is a legit deletion, so no suffix needed for that either)

Nerixyz:
> warn/(un)mod/(un)timeout/(un)raid/etc are not mirrored

... so this is a mistake in the docs? The [`warn` example here](https://dev.twitch.tv/docs/eventsub/eventsub-subscription-types/#channelmoderate-v2) shows different users in broadcaster vs source.

iProdigy:
yeah believe that's a docs error

Nerixyz:
Thank you for clarifying. tl;dr: anything, but the `shared_*` events aren't mirrored. 

`operator==` is still a useful feature, same as `isSharedChat` (maybe we need that for the shared-chat events).

PR #5975, opened 2025-02-22
fix: special shortcut for Windows live notifications

### LINKED ISSUE #5954: WinToast needs a special shortcut
### Checklist

### Describe your issue

Windows uses a Shortcut with AUMI (App User Model ID) to associate a notification with an application.
If this shortcut is missing, notifications may fail to display or behave incorrectly [See here](https://redirect.github.com/mohabouje/WinToast/issues/60)
Since 6e0852f we disabled creation of this shortcut.

For me, if this is missing, the following unexpected behavior occurs:
- notification has no icon and a different name 
- unable to customize the Toast behavior within Windows settings `System -> Notifications & actions`

```
Steps to Reproduce

- Enable live notification for any current streaming channel
- Close Chatterino
- Remove all Chatterino shortcuts at:
  - %AppData%\Roaming\Microsoft\Windows\Start Menu\Programs
  - %ProgramData%\Microsoft\Windows\Start Menu\Programs
- Clear Notification Center
- Run Chatterino
- Receive "broken" notification
```

<details><summary>Maybe fixable with this patch</summary>

<p>
This should create a shortcut, unless you explicitly disable it (what #3817 was about).

I couldn't figure out a different way to get this working.

```
diff --git a/src/singletons/Settings.hpp b/src/singletons/Settings.hpp
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -526,6 +526,7 @@ public:
         "/notifications/suppressInitialLive", false};
 
     BoolSetting notificationToast = {"/notifications/enableToast", false};
+    BoolSetting notificationToastDisableShortcut = {"/notifications/disableWinToastShortcut", false};
     IntSetting openFromToast = {"/notifications/openFromToast",
                                 static_cast<int>(ToastReaction::OpenInBrowser)};
 
diff --git a/src/singletons/Toasts.cpp b/src/singletons/Toasts.cpp
--- a/src/singletons/Toasts.cpp
+++ b/src/singletons/Toasts.cpp
@@ -230,7 +230,9 @@ void Toasts::ensureInitialized()
     instance->setAppUserModelId(
         WinToast::configureAUMI(L"", L"Chatterino 2", L"",
                                 Version::instance().version().toStdWString()));
-    instance->setShortcutPolicy(WinToast::SHORTCUT_POLICY_IGNORE);
+    if (getSettings()->notificationToastDisableShortcut) {
+        instance->setShortcutPolicy(WinToast::SHORTCUT_POLICY_IGNORE);
+    }
     WinToast::WinToastError error{};
     instance->initialize(&error);
 
diff --git a/src/widgets/settingspages/NotificationPage.cpp b/src/widgets/settingspages/NotificationPage.cpp
--- a/src/widgets/settingspages/NotificationPage.cpp
+++ b/src/widgets/settingspages/NotificationPage.cpp
@@ -49,6 +49,8 @@ NotificationPage::NotificationPage()
 #ifdef Q_OS_WIN
                 settings.append(this->createCheckBox(
                     "Show notification", getSettings()->notificationToast));
+                settings.append(this->createCheckBox(
+                    "Disable creation of WinToast Shortcut (requires restart)", getSettings()->notificationToastDisableShortcut));
                 auto openIn = settings.emplace<QHBoxLayout>().withoutMargin();
                 {
                     openIn

```

</p>
</details> 

<details><summary>Check if your shortcut has AUMI configured</summary>
Use a Tool called [LECmd](https://github.com/EricZimmerman/LECmd) with this parameter on a shortcut file:

<code>LECmd.exe -f Chatterino2.lnk</code>

If the shortcut has AUMI configured, you find this in the output `App User Model ID                   ==> .Chatterino 2`
</details> 

<details><summary>Manually create a shortcut with AUMI</summary>
Use a Tool called [make-shortcut-with-appusermodelid](https://github.com/Robertof/make-shortcut-with-appusermodelid)
<code>makelnk.exe ".Chatterino 2" "C:\PathToChatterino\chatterino.exe" Chatterino2</code>
</details> 

@Nerixyz if you have time, could you try to reproduce this?

### Screenshots

| Shortcut without AUMI  | Shortcut with AUMI  |
| ------------- | ------------- |
| Image  | Image  |
| Image  | Image  |

*Reply Box removed since d0cf6c4
**Shortcut was renamed to `Chatterino` by myself

### OS and Chatterino Version

Chatterino Nightly 2.5.2 (commit 449aefc) Qt 6.7.1 Running on Windows 10

8thony:
Fixes #5954

The shortcut is created by default when the first notification fires, unless one of the `portable` or `externally-packaged` (new mode) modes are enabled.

`externally-packaged` is a new mode with this released, and can be useful when packaging Chatterino (e.g. to SCOOP)


.CI/chatterino-installer.iss
```diff
@@ -77,7 +77,7 @@ Source: "{#WORKING_DIR}vc_redist.x64.exe"; DestDir: "{tmp}"; Tasks: vcredist;
 ; NOTE: Don't use "Flags: ignoreversion" on any shared system files
 
 [Icons]
-Name: "{autoprograms}\{#MyAppName}"; Filename: "{app}\{#MyAppExeName}"
+Name: "{autoprograms}\{#MyAppName}"; Filename: "{app}\{#MyAppExeName}"; AppUserModelID: ".Chatterino 2"
```
pajlada, line 80: Is `AppUserModelID` the string that appears on top of the notification?

8thony: No, it is used to associate the shortcut with the app.
I probably add it to the Desktop Shortcut as well.

8thony:
@Nerixyz there might be a bug with pinning c2 to the taskbar.

src/widgets/settingspages/NotificationPage.cpp
```diff
@@ -51,6 +51,14 @@ NotificationPage::NotificationPage()
...
 #endif
 #ifdef Q_OS_WIN
+                settings.append(this->createCheckBox(
+                    "Create start menu shortcut (requires "
+                    "restart)",
+                    getSettings()->createShortcutForToasts,
+                    "When enabled, a shortcut will be created inside your "
+                    "start menu folder if needed by live notifications."
+                    "\n(On portable mode, this is disabled by "
+                    "default)"));
```
pajlada, line 61: You can use `SettingWidget` here, which means you wouldn't have to update the `createCheckBox` implementation - see https://github.com/Chatterino/chatterino2/blob/7fee3f7e375a099b5207c215389343f8f1d95bf2/src/widgets/settingspages/GeneralPage.cpp#L1231-L1235 as a reference

src/singletons/Settings.hpp
```diff
@@ -526,6 +527,9 @@ class Settings
         "/notifications/suppressInitialLive", false};
 
     BoolSetting notificationToast = {"/notifications/enableToast", false};
+    BoolSetting createShortcutForToasts = {
+        "/notifications/createShortcutForToasts",
+        Modes::instance().isPortable ? false : true};
```
pajlada, line 532: add a , at the end of the ternary, that way it formats a bit nicer, causing nicer diffs in the future if we have to edit it

pajlada (commented):
We can just change all AUMI/AppNames to "Chatterino", feel free to do that in this PR - I'm assuming that should not break anything. Rather that be changed now than before we start creating shortcuts

Nerixyz:
> Should we change the name of the App [here](https://github.com/Chatterino/chatterino2/blob/7fee3f7e375a099b5207c215389343f8f1d95bf2/src/singletons/Toasts.cpp#L231)?

This worked for me. I'd advocate for this.

---

Honestly, I now think this should be disabled by default. From what we know: installations with the installer (probably the majority) aren't affected. But we add a shortcut there by default. In Chatterino7, for example, the created shortcut is named `Chatterino7.lnk`. With this PR, an additional shortcut (`Chatterino.lnk`) would be created. But that's not necessary (could also change the app name there, I suppose). It's really only for portable users (or users of `cmake --install`).

Nerixyz:
> Does it show the correct name and icon for you on stable 2.5.2 installed with its installer?

Hmm, no, it doesn't. The installer changes are good, though, I'm fine with them. I guess we could enable it by default... But we should definitely change the name that we give to WinToast then. Because the installer creates `Chatterino.lnk`, not `Chatterino2.lnk`.

8thony:
Changes made:
- Renamed shortcut to `Chatterino`
- Renamed all AppUserModelID
- Added AppUserModelID to main app [more about this here](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid)
   - This makes sure the taskbar shortcut uses the same AppUserModelID.
	   - Downside: the old taskbar shortcut "break" (c2 need to be re-pinned one time)

src/widgets/settingspages/NotificationPage.cpp
pajlada: nvm all good, you can't do that here yet

pajlada (approved):

PR #5980, opened 2025-02-23
chore(eventsub): simplify building for system messages

Nerixyz:
This should make it easier to build simple system messages in the future. A new message builder is born, `EventSubMessageBuilder`. But we merged all builders some time ago? Yes, we could do this in the `MessageBuilder` as well. However, I do feel like this is a bit specific to EventSub. Especially, having `builder.appendUser(esString, esString, text)` is neat and avoids boilerplate. Note: we will need to have even more handlers if we support `channel.chat.notification`.

For more complex messages, a second function name like `onModerateMessage(chan, time, event, action)` can be used.


pajlada (approved):
I just want to make sure when we start adding more message builders that we don't end up with the IRC/PubSub message builder - a class with so many static functions and helpers that only work for certain scenarios. What you've done here is good :+1: Good use of concepts & leaving the main building as separate functions still, with the Message Builder only having some useful helper funtions

PR #5981, opened 2025-02-23
feat(eventsub): implement clearchat

Nerixyz:
Basically the same as in PubSub. While testing, I noticed that we call `getTwitch` in `onChannelModerate`. That function isn't called on the GUI thread, though. I moved the assertions to the functions of `ITwitchIrcServer` (where needed).

### DIFF
--- a/src/providers/twitch/eventsub/Connection.cpp
+++ b/src/providers/twitch/eventsub/Connection.cpp
@@ -6,6 +6,7 @@
 #include "messages/Message.hpp"
 #include "messages/MessageBuilder.hpp"
 #include "providers/twitch/eventsub/MessageBuilder.hpp"
+#include "providers/twitch/eventsub/MessageHandlers.hpp"
 #include "providers/twitch/PubSubActions.hpp"
 #include "providers/twitch/TwitchChannel.hpp"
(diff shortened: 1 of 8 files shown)


src/providers/twitch/eventsub/Connection.cpp
```diff
@@ -23,15 +25,41 @@ namespace {
...
 
+namespace channel_moderate = lib::payload::channel_moderate::v2;
+
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 const auto &LOG = chatterinoTwitchEventSub;
 
+void handleModerateMessage(
+    TwitchChannel *chan, [[maybe_unused]] const QDateTime &time,
+    const channel_moderate::Event &event,
+    [[maybe_unused]] const channel_moderate::Clear &action)
```
Nerixyz, line 36: Could be done in something like `providers/twitch/eventsub/MessageHandlers.{c,h}pp`?

pajlada: That file structure seems fine to me :+1:

pajlada (approved):

Issue #5908, opened 2025-02-06
EventSub tracking issue


Nerixyz:
The EventSub implementation is very basic right now, but it needs to be ready in the near future (https://discuss.dev.twitch.com/t/legacy-pubsub-deprecation-and-shutdown-timeline/58043). A few tasks that need to be done at some point (feel free to update) and I can't keep them all in my head or in the `TODO.md` file.

**twitch-eventsub-ws**
- [x] #5911
- [x] #5912
- [x] #5913
- [x] #5909
- [x] #5917
- [x] #5918
- [x] #5919

**Chatterino Integration**
- [x] #5926
- [ ] #5929
- [x] #5928
- [ ] #5927
- [x] Reconnect after network outage #6059 #6058
- [x] Use Qt's logging instead of `std::cerr` #6161

**PubSub Parity**
- [x] `moderation.chatCleared` (handled by `channel.moderate` `clear` #5981)
- [x] `moderation.messageDeleted` (handled by `channel.moderate` `(shared_chat_)delete` #5993)
- [x] `moderation.modeChanged` (handled by `channel.moderate` #5992)
- [x] `moderation.moderationStateChanged` (handled by `channel.moderate` `(un)mod` #6001)
- [x] `moderation.raidStarted` (handled by `channel.moderate` `raid` #6002)
- [x] `moderation.raidCanceled` (handled by `channel.moderate` `unraid` #6002)
- [x] `moderation.userBanned` (handled by `channel.moderate` `(shared_chat_)ban` #5985 and `(shared_chat_)timeout` #5995)
- [x] `moderation.userUnbanned` (handled by `channel.moderate` `unban` and `shared_chat_unban` #5990)
- [x] `moderation.userWarned` (handled by `channel.moderate` `warn` #5932)
- [x] `moderation.suspiciousMessageReceived` (handled by `channel.suspicious_user.message` #6007)
- [x] `moderation.suspiciousTreatmentUpdated` (handled by `channel.suspicious_user.update` #6012)
- [x] `moderation.autoModMessageCaught` (handled by `automod.message.hold` #6005 and `automod.message.update` #6010)
- ~~`moderation.autoModMessageBlocked` (handled by `automod.message.update`)~~ was superseded by autoModMessageCaught but never removed
- [x] `moderation.automodUserMessage` (handled by `channel.moderate` `{add,remove}_{permitted,blocked}_term` #6000)
[...]

PR #5992, opened 2025-02-26
feat(eventsub): implement mode changes

Nerixyz:
Towards #5908.

Implements
- followers (we didn't handle this in PubSub)
- followersoff (we didn't handle this in PubSub)
- emoteonly
- emoteonlyoff
- slow
- slowoff
- subscribers
- subscribersoff
- uniquechat
- uniquechatoff

### DIFF
--- a/src/providers/twitch/eventsub/MessageBuilder.cpp
+++ b/src/providers/twitch/eventsub/MessageBuilder.cpp
@@ -203,8 +203,12 @@ void makeModerateMessage(
     const lib::payload::channel_moderate::v2::Event &event,
     const lib::payload::channel_moderate::v2::Followers &action)
 {
-    makeModeMessage(builder, event, u"followers-only"_s, true,
-                    u"(%1 minutes)"_s.arg(action.followDurationMinutes));
+    QString duration;
+    if (action.followDurationMinutes > 0)
+    {
+        duration = u"(%1 minutes)"_s.arg(action.followDurationMinutes);
+    }
+    makeModeMessage(builder, event, u"followers-only"_s, true, duration);
 }
 void makeModerateMessage(
     EventSubMessageBuilder &builder,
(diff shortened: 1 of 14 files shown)


src/providers/twitch/eventsub/MessageBuilder.cpp
```diff
@@ -1,8 +1,40 @@
...
+    builder.emplaceSystemTextAndUpdate(u"turned"_s, text);
+    QString op = on ? u"on"_s : u"off"_s;
+    builder.emplaceSystemTextAndUpdate(op, text);
+    builder.emplaceSystemTextAndUpdate(mode, text);
+    builder.emplaceSystemTextAndUpdate(u"mode."_s, text);
+
+    if (!duration.isEmpty())
+    {
+        builder.emplaceSystemTextAndUpdate(duration, text);
+    }
```
pajlada, line 30: Followers-only mode can have a "0 minute" duration, we should omit the duration text if that's the case.

image

The `followers` object looks like this:

```json
    "followers": {
      "follow_duration_minutes": 0
    },
```

src/providers/twitch/eventsub/MessageBuilder.cpp
```diff
@@ -166,4 +198,84 @@ void makeModerateMessage(
...
+{
+    makeModeMessage(builder, event, u"r9k"_s, true);
+}
+void makeModerateMessage(
+    EventSubMessageBuilder &builder,
+    const lib::payload::channel_moderate::v2::Event &event,
+    const lib::payload::channel_moderate::v2::UniquechatOff & /*action*/)
+{
+    makeModeMessage(builder, event, u"r9k"_s, false);
+}
```
pajlada, line 279: Let's call it unique-chat - it's what the feature's new name is
```suggestion
void makeModerateMessage(
    EventSubMessageBuilder &builder,
    const lib::payload::channel_moderate::v2::Event &event,
    const lib::payload::channel_moderate::v2::Uniquechat & /*action*/)
{
    makeModeMessage(builder, event, u"unique-chat"_s, true);
}
void makeModerateMessage(
    EventSubMessageBuilder &builder,
    const lib::payload::channel_moderate::v2::Event &event,
    const lib::payload::channel_moderate::v2::UniquechatOff & /*action*/)
{
    makeModeMessage(builder, event, u"unique-chat"_s, false);
}
```

pajlada (approved):

PR #5993, opened 2025-02-26
feat(eventsub): implement (shared chat) delete

pajlada:
Towards #5908

Implements:
 - Delete  
   
image
image

 - Shared chat delete (we didn't handle this in PubSub)
 
image
image

Will add changelog entry after the chat modes PR is merged in


Nerixyz (commented):
In PubSub we also [tried to merge messages from IRC](https://github.com/Chatterino/chatterino2/blob/5ea420b7ae41f976f05db3b4df6428705af0e89f/src/providers/twitch/TwitchIrcServer.cpp#L377-L391). I think we should focus on building the messages first and merging them afterward (especially if we want to restructure `Message` soon-ish).

PR #5998, opened 2025-02-26
Chatterino warning fixes

TheComputerGuy96:
This fixes the warnings inside Chatterino code with GCC 14 on Linux (which progresses #4825 forward)

Note that sol2 may still have a stubborn `-Wmaybe-uninitialized` warning even after this change


pajlada:

PR #6002, opened 2025-02-27
feat(eventsub): Add support for `/(un)raid`

pajlada:
### `/raid`

Looks the same as from PubSub (first line is PubSub, second line is EventSub)
raid

### `/unraid`

Since we have the info, I've added the channel we've cancelled the raid to (first line is PubSub, second line is EventSub)
unraid


pajlada (commented):
image

PR #6005, opened 2025-02-28
feat(eventsub): implement automod message hold

Nerixyz:
Towards https://github.com/Chatterino/chatterino2/issues/5908. I mostly copied the code from the PubSub implementation, but split it in two. The message is "just" added. The fragments (=emotes) are ignored like in PubSub.

(top message is PubSub, because EventSub is slow)


pajlada (approved):
PubSub does deduping of matched terms which I think would be nice, but I'll add it as a "nice to have issue" instead

image

PR #6011, opened 2025-03-01
fix: don't save commands on exit

### LINKED ISSUE #6009: commands.json will reset unpredictably
### Checklist

### Describe your issue

Every so often when I start Chatterino, it seems to not properly read or handle the command settings file. Chatterino doesn't recognize my custom commands and when I check the settings folder under %appdata% it created an empty `commands.json` file instead of reusing my existing settings. I need to manually restore the settings from one of the created backups.

I have multiple ideas what may causes this: Either Chatterino isn't always properly making sure that the file was properly written to the file system (stream not properly flushed/closed?) or Chatterino isn't handling running multiple instances well enough. Sometimes I accidently open a second Chatterino instance because my muscle memory opens a new window. I cannot guarantee that this is causing it.

I can not reproduce it deterministically, but it happens like once every two weeks and it's starting to become annoying enough that I wrote this report 🖖 

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.5.1 (commit e3268767) built on 2024-12-08 with Qt 6.7.1, MSVC 194234435,

pajlada:
also update Settings to v0.3.0, giving us the ability to more sparingly save the commands.

We can now call `CommandController::save()` as much as we want, but it'll only save if something has actually saved, thanks to the combined use of `OnlySaveOnChange` and `SettingOption::CompareBeforeSet`

Fixes #6009

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Bugfix: Fixed channel point redemptions with messages not showing up if PubSub is disconnected. (#5948)
 - Bugfix: Fixed the input font not immediately updating when zooming in/out. (#5960)
 - Bugfix: Fixed color input thinking blue is also red. (#5982)
+- Bugfix: Fixed an issue where commands would sometimes reset if Chatterino was improperly shut down. (#6011)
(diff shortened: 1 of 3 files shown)


pajlada (commented):
image

PR #6017, opened 2025-03-02
fix(eventsub): connection race

pajlada:
Sometimes when subscriptions came in too fast onto each other (e.g. when you had multiple messages in recentmessages that marked you as mod, then not mod, then mod, then not mod etc), we could start multiple subscription attempts for the same topic. This fired an assertion in the retry logic.

We now store the state the subscription is currently in, and properly manage that state (copium) to ensure we know from what states various calls are allowed to be made.
I've added a million new asserts, if any of them fire we need to double check that things are ok :)


src/providers/twitch/eventsub/Controller.cpp
```diff
@@ -154,133 +222,157 @@ SubscriptionHandle Controller::subscribe(const SubscriptionRequest &request)
...
+        auto *listener = dynamic_cast<Connection *>(connection->getListener());
+
+        qCDebug(LOG) << "Make helix request for" << request;
+        getHelix()->createEventSubSubscription(
+            request, listener->getSessionID(),
+            [this, request, connection,
+             weakConnection{std::weak_ptr<lib::Session>(connection)}](
+                const auto &res) {
+                qCDebug(LOG) << "Subscription success" << request;
+                // TODO: Move marathon stick in here
```
Nerixyz, line 266: Leftover?

src/providers/twitch/eventsub/Controller.cpp
```diff
@@ -154,133 +222,157 @@ SubscriptionHandle Controller::subscribe(const SubscriptionRequest &request)
 
 void Controller::subscribe(const SubscriptionRequest &request, bool isRetry)
 {
-    qCInfo(LOG) << "Subscribe request for" << request.subscriptionType;
-    boost::asio::post(this->ioContext, [this, request, isRetry] {
-        // 1. Flush dead connections (maybe this should not be done here)
-        // TODO: implement
+    // boost::asio::post(this->ioContext, [this, request, isRetry] {
```
Nerixyz, line 225: Leftover?

src/providers/twitch/eventsub/Controller.cpp
```diff
@@ -154,133 +222,157 @@ SubscriptionHandle Controller::subscribe(const SubscriptionRequest &request)
...
+        this->getViableConnection(openButNotReadyConnections);
+
+    if (viableConnection.has_value())
+    {
+        const auto &connection = *viableConnection;
+        auto *listener = dynamic_cast<Connection *>(connection->getListener());
+
+        qCDebug(LOG) << "Make helix request for" << request;
+        getHelix()->createEventSubSubscription(
+            request, listener->getSessionID(),
```
Nerixyz, line 261: Should `assert(listener)` here (→ that all our listeners are `Connection`s)

src/providers/twitch/eventsub/Controller.cpp
```diff
@@ -154,133 +222,157 @@ SubscriptionHandle Controller::subscribe(const SubscriptionRequest &request)
...
+
+        if (listener == nullptr)
         {
-            // No connection was available to handle this subscription request, create a new connection
-            this->createConnection();
-            this->retrySubscription(request, boost::posix_time::millisec(500),
-                                    10);
+            // something really goofy is going on
+            qCWarning(LOG) << "listener was not the correct type";
+            continue;
```
Nerixyz, line 360: Here we continue - I think we should assert here as well.

src/providers/twitch/eventsub/Controller.cpp
pajlada: Fixes in 2dfc02354088bb8e5c3080060fabf770ddf03f7d

src/providers/twitch/eventsub/Controller.cpp
pajlada: Fixed in 786cc8b09ef0afcd6ba51bba442381f68f1541e0

src/providers/twitch/eventsub/Controller.cpp
pajlada: Fixed in 1b4205bd33c78d1bd39ea69abb7eedbdf5ee72e5

src/providers/twitch/eventsub/Controller.cpp
pajlada: Fixed in dc3330b0b1267ef7e1ca416f87906fa26d9f3503

pajlada:
image

PR #6019, opened 2025-03-02
reduce pubsub/liveupdates pubsub clients "wait for clean exit" time from 1s to 100ms

pajlada:
this means "unclean" exits take less time


pajlada:
image

PR #6030, opened 2025-03-05
chore: Cleanup of hotkeys & settings in #5856

brian6932:


src/widgets/splits/Split.cpp
```diff
@@ -669,24 +669,18 @@ void Split::addShortcuts()
...
-                 this->setModerationMode(true);
-             }
-             else
+             switch (mode)
              {
-                 this->setModerationMode(!this->getModerationMode());
+                 case 0:
+                 case 1:
+                     this->setModerationMode((bool)mode);
+                     break;
```
pajlada, line 679: I'd rather case 0 and case 1 are separate calls to `this->setModerationMode`

The rest of this file looks good

src/widgets/splits/SplitHeader.cpp
```diff
@@ -541,59 +541,64 @@ std::unique_ptr<QMenu> SplitHeader::createMainMenu()
...
+                        getApp()->getNotifications()->isChannelNotified(
+                            this->split_->getChannel()->getName(),
+                            Platform::Twitch));
+                });
+            QObject::connect(action, &QAction::triggered, this, [this]() {
+                getApp()->getNotifications()->updateChannelNotification(
+                    this->split_->getChannel()->getName(), Platform::Twitch);
+            });
 
-    if (twitchChannel)
```
pajlada, line 571: This change just means the `Mute highlight sounds` action shows up when i right click e.g. the `/mentions` tab right?

pajlada (changes requested):

src/widgets/splits/SplitHeader.cpp
brian6932: No, this logically changes nothing, it just removes a repeated condition and replaces it with 2 empty block scopes within the same condition. The diff looks weird here, apart of the reason why it was moved to the separate PR.

src/widgets/splits/Split.cpp
```diff
@@ -498,7 +498,7 @@ void Split::addShortcuts()
          }},
         {"focus",
          [this](std::vector<QString> arguments) -> QString {
```
Nerixyz, line 500: These arguments could be `const std::vector<QString> &`. Maybe we can move them to `std::span<const QString>` some day...

pajlada (approved):

PR #6031, opened 2025-03-05
feat: support `/watching` on all platforms

Nerixyz:
Adds support for the browser extension on all platforms. Attaching to a browser is still only supported on Windows.

Tested on:

- Google Chrome, Linux
- Firefox, Linux
- Google Chrome, Windows
- Firefox, Windows

The browser extension should probably disable the "replace Twitch chat" option on non-windows platforms.

While debugging on Linux, I found a bug with the native messaging thread where it would be destructed too soon.


src/singletons/NativeMessaging.cpp
```diff
@@ -169,18 +216,18 @@ NativeMessagingServer::~NativeMessagingServer()
...
-    this->thread.quit();
+    this->thread->requestInterruption();
+    this->thread->quit();
     // Most likely, the receiver thread will still wait for a message
-    if (!this->thread.wait(250))
+    if (!this->thread->wait(250))
     {
-        this->thread.terminate();
+        this->thread->terminate();
     }
```
pajlada, line 225: Seems like we're supposed to wait after calling terminate too https://doc.qt.io/qt-6/qthread.html#terminate

Maybe something like this instead - reducing the max timeout & waiting after the call to terminate too - 
can't do a suggestion because it spans a deleted line or something
```c++
    if (!this->thread->wait(100))
    {
        this->thread->terminate();

        if (!this->thread->wait(100))
        {
            qCWarning(chatterinoNativeMessage)
                << "Failed to terminate thread cleanly";
        }
    }
```

src/singletons/NativeMessaging.cpp
```diff
@@ -158,9 +205,9 @@ namespace nm::client {
 
 // SERVER
 NativeMessagingServer::NativeMessagingServer()
-    : thread(*this)
+    : thread(new ReceiverThread(*this))
 {
-    this->thread.setObjectName("NativeMessagingReceiver");
+    this->thread->setObjectName("NativeMessagingReceiver");
```
pajlada, line 210: Rename the thread so it fits within posix thread limitations
```suggestion
    this->thread->setObjectName("C2NMReceiver");
```

pajlada (changes requested):
Looks good, just some small nits. Will test this in macOS

pajlada (approved):
I've confirmed this works on macOS using Google Chrome and Firefox

There's this little bouncing icon in my task bar whenever i switch to and from a Twitch tab now, which we suspect is the little chatterino native host thing. I don't mind if that's fixed in a follow-up PR since I don't like dealing with macOS :-)

crazysmc:
The Windows-only `#ifdef` on the settings page was overlooked and thus no extension IDs can be added on other platforms nor is the *recommended* extension linked https://github.com/Chatterino/chatterino2/blob/f6a820cff0ab3fc6c56bfb43dbe4fee968e39282/src/widgets/settingspages/GeneralPage.cpp#L892-L941

Nerixyz:
> The Windows-only `#ifdef` on the settings page was overlooked and thus no extension IDs can be added on other platforms nor is the _recommended_ extension linked

Ah, right, I forgot about that. Feel free to make a PR. Not a huge issue as you can manually edit the JSON file.

PR #6032, opened 2025-03-05
fix: Qt 6.4.3 build

pajlada:
I started this PR with just attempting to fix the compilation, but without some of the further fixes the snapshot tests failed due to outputting garbage strings instead of the real strings.

Something with Qt 6.4.3's QStringBuilder is broken, so I'm using these changes as a temporary fix until we don't support Qt 6.4.3 anymore.

### DIFF
--- a/lib/twitch-eventsub-ws/tests/src/string.cpp
+++ b/lib/twitch-eventsub-ws/tests/src/string.cpp
@@ -1,11 +1,13 @@
 #include <gtest/gtest.h>
 #include <QString>
+#include <QStringBuilder>
 #include <twitch-eventsub-ws/string.hpp>
 
 #include <type_traits>
 
 namespace {
 
+constexpr const char *MINI = "mini";
 constexpr const char *LONGEST_SSO = "mylongstringyesitslongwo";
 constexpr const char *TOO_LONG = "mylongstringyesitslongwow";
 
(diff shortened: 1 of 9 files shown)


lib/twitch-eventsub-ws/tests/src/string.cpp
```diff
@@ -567,3 +569,18 @@ TEST(String, Equals)
...
+{
+    String s(MINI);
+
+    ASSERT_EQ("foo" % s.qt() % "bar", "foominibar");
+    ASSERT_EQ(u"foo" % s.qt() % "bar", "foominibar");
+    ASSERT_EQ("foo" % s.qt() % u"bar", "foominibar");
+    ASSERT_EQ(u"foo" % s.qt() % u"bar", "foominibar");
+
+    // the code below breaks on qt 6.4.3, but i think it's unrelated to our String stuff
+    // auto text2 = u"a" % s.qt() % u"b";
```
Nerixyz, line 583: A trivial fix would be 
```suggestion
    QString text2 = u"a" % s.qt() % u"b";
```

Nerixyz (commented):
Too many unrelated changes (for the title "chore(ci): build with Qt 6.4.3 in CI").

pajlada:
image

PR #6035, opened 2025-03-06
feat(eventsub): reconnect

Nerixyz:
Towards https://github.com/Chatterino/chatterino2/issues/5908.

When we disconnect, or we're asked to reconnect, we should reconnect. A `Session` can now be orphaned, because we want to transplant the listener. That's mostly it. If we don't have a reconnect URL and are disconnected, we reset all the subscriptions a connection was subscribed to and try to subscribe again.

You can try this with the Twitch CLI. I opened https://redirect.github.com/twitchdev/twitch-cli/pull/347 to have at least one event that we use in there. Otherwise, you can also edit some event in `TwitchChannel` to use `channel.ban`.

```shell
twitch event websocket start-server --ssl --port 3012 # start local server
chatterino --use-local-eventsub # start chatterino
twitch event websocket close --reason=4000 --session=<id> # disconnect without preserving subs
twitch event websocket reconnect  # reconnect while preserving subs
twitch event trigger automod-message-hold -T websocket # trigger an event (should only trigger once)
```

I converted the local eventsub thing into an argument, because it needs to use the local server for subscribing to EventSub. To avoid a qbittorrent situation, this is limited to debug builds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - Bugfix: Fixed color input thinking blue is also red. (#5982)
 - Bugfix: Fixed an issue where commands would sometimes reset if Chatterino was improperly shut down. (#6011)
 - Dev: Subscriptions to PubSub channel points redemption topics now use no auth token, making it continue to work during PubSub shutdown. (#5947)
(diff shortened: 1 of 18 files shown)


Nerixyz:
Should we increment the retry counter if there aren't any open connections? Feels a bit wrong, but when the user is disconnected, they will also be disconnected from IRC, meaning that when they reconnect, they should get resubscribed (need to test that).

pajlada (approved):

pajlada:
Tested this and it works, nice!

pajlada:
> Should we increment the retry counter if there aren't any open connections? Feels a bit wrong, but when the user is disconnected, they will also be disconnected from IRC, meaning that when they reconnect, they should get resubscribed (need to test that).

You're thinking if the user doesn't have a network connection at all, to ensure the reconnects happen indefinitely? That definitely makes sense to try & ensure it works

PR #6036, opened 2025-03-07
refactor(eventsub): use backoff over retry count

Nerixyz:
From https://github.com/Chatterino/chatterino2/pull/6035: if there's no connection, we should retry indefinitely.

Thinking a bit more about this, I came to the conclusion that we don't need the retry count at all but a backoff for each subscription. This makes the code a bit simpler.

Three semi-related changes:
- If Helix returns _403 Forbidden_, we can mark the request as failed (no need to retry). We might need to retry if we get a _429 Too Many Requests_. Unfortunately, that status is also used if a subscription already exists ([docs](https://dev.twitch.tv/docs/api/reference/#create-eventsub-subscription)).
- If someone unsubscribed and immediately subscribed again, we could've discarded the subscription request if it came in while we were `Unsubscribing`. To detect this, we check when marking a request as unsubscribed if there are any more references. If so, we subscribe again.
- Used a `boost::asio::system_timer`, because the `boost::asio::deadline_timer` is deprecated (and doesn't support `std::chrono`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - Bugfix: Fixed color input thinking blue is also red. (#5982)
 - Bugfix: Fixed an issue where commands would sometimes reset if Chatterino was improperly shut down. (#6011)
 - Dev: Subscriptions to PubSub channel points redemption topics now use no auth token, making it continue to work during PubSub shutdown. (#5947)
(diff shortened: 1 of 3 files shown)


pajlada:
Looks good - testing this out a bit

pajlada (approved):
This has worked well in my brief testing

PR #6037, opened 2025-03-07
feat: allow custom chat font weight

Nerixyz:
Since I always used GDI as the font engine, I never noticed that it parses some weights (styles) of fonts as different families (e.g. there's "Segoe UI", "Segoe UI Bold", "Segoe UI Light"). DirectWrite doesn't do that, and properly combines them into one font family ("Segoe UI" with "Normal", "Bold", and "Light" styles). My selected chat font has a different weight than 400 (normal) and I'd like to keep that.

This PR adds a setting to customize the weight of the chat font. Some font engines support variable fonts, and we should allow for customization of this. The setting takes an OpenType font weight that is directly passed to Qt.

Since this touches `Fonts.cpp`, I refactored that a bit to have "getter" functions that each resolve a font property (family, size, weight, italic).

Future work: The font dialog should automatically set the selected font weight and size.

### DIFF
--- a/src/singletons/Fonts.cpp
+++ b/src/singletons/Fonts.cpp
@@ -12,7 +12,7 @@ namespace {
 
 using namespace chatterino;
 
-int getBoldness()
+int getUsernameBoldness()
 {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
     // From qfont.cpp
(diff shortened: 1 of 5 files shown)


src/singletons/Fonts.cpp
```diff
@@ -105,83 +221,20 @@ Fonts::FontData &Fonts::getOrCreateFontData(FontStyle type, float scale)
...
-        UiFontData &data = defaultSize[type];
-        QFont font(data.name, int(data.size * scale), data.weight, data.italic);
-        return FontData(font);
-    }
+    return QFont{
+        fontFamily(type),
+        static_cast<int>(fontSize(type) * scale),
+        fontWeight(type),
+        isItalic(type),
+    };
```
pajlada, line 237: nice

pajlada (approved):

PR #6041, opened 2025-03-08
fix(eventsub): Respect "Hide moderation actions"

pajlada:
This change is bigger than it technically needed to be.

I've added a "ModerationAction" message flag which is what's actually being used to hide messages instead of incorrectly flagging messages as `Timeout` or `Untimeout`.
This will have some consequences for message search & filters, but I believe they are all more correct than they were before.
If the user case for being able to filter on previously-timeout-filtered-messages comes up we can add more search predicates for this.


pajlada (commented):
image

PR #6042, opened 2025-03-08
feat: Hide messages from suspicious users in streamer mode

pajlada:
draft waiting for previous PR to be merged in

### DIFF
--- a/src/messages/layouts/MessageLayout.cpp
+++ b/src/messages/layouts/MessageLayout.cpp
@@ -166,6 +166,17 @@ void MessageLayout::actuallyLayout(const MessageLayoutContext &ctx)
             continue;
         }
 
+        if (this->message_->flags.hasAny(MessageFlag::RestrictedMessage,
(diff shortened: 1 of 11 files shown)


pajlada (commented):
image

src/messages/layouts/MessageLayout.cpp
```diff
@@ -166,6 +166,17 @@ void MessageLayout::actuallyLayout(const MessageLayoutContext &ctx)
             continue;
         }
 
+        if (this->message_->flags.hasAny(MessageFlag::RestrictedMessage,
+                                         MessageFlag::MonitoredMessage))
```
iProdigy, line 170: a note for when we eventually use eventsub for monitored messages: it'd be nice if streamer mode only hid the indication that the message was monitored rather than hiding the message outright. With the latter, monitored users may realize they are monitored as the streamer shows their c2 on stream

PR #6043, opened 2025-03-08
fix: set application policy on macOS for extension

Nerixyz:
An attempt to avoid showing Chatterino in the dock/menu bar when the browser extension proxy is launched. No clue if it works, but that reverts what Qt does [when it loads the coca plugin](https://github.com/qt/qtbase/blob/a72077a88903fe532f6a749677eb4da4ea99f79f/src/plugins/platforms/cocoa/qcocoahelpers.mm#L179).


pajlada (approved):
Confirmed this works - the application pops up for a millisecond in the taskbar, but it's not noticeable unless you're looking. And only the first launch, not whenever you switch tabs

PR #6049, opened 2025-03-09
fix: only hide messages from restricted users, not all suspicious users

pajlada:
it's probably best to leave this setting solo'd out as restricted users
so monitored users can get their own behaviour in their own setting later


pajlada:
image

PR #6053, opened 2025-03-09
feat: gracefully shut down on SIGINT/SIGTERM on Linux/mac

pajlada:


src/util/UnixSignalHandler.cpp
```diff
@@ -0,0 +1,86 @@
+#include <QtGlobal>
+#if QT_VERSION >= QT_VERSION_CHECK(6, 5, 0)
+#    include <QtSystemDetection>
```
Nerixyz, line 3: This is included in `<QtGlobal>` ([docs](https://doc.qt.io/qt-6/qtglobal.html))

src/util/UnixSignalHandler.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+public:
+    UnixSignalHandler(int signal);
+
+    static void fired(int signal);
+
+Q_SIGNALS:
+    void signalFired();
+
+private Q_SLOTS:
```
Nerixyz, line 28: You don't really _need_ `Q_SLOTS` here, since you pass a function pointer to `handleSignal` instead of a name.

src/util/UnixSignalHandler.cpp
pajlada: Fixed in [`c91f894` (#6053)](https://github.com/Chatterino/chatterino2/pull/6053/commits/c91f894f6b12b32ce1645925f6df16c1d921c1b2) - Thanks!

pajlada:
image

PR #6057, opened 2025-03-09
fix: crash when searching uninitialized channel

pajlada:
To reproduce the crash:

1. Open a new split (e.g. through CTRL+T)
2. **Don't** click ok or cancel in the channel selector dialog
3. Click the new split that was created
4. press CTRL+F


pajlada:
image

PR #6058, opened 2025-03-09
fix(eventsub): check keepalive

Nerixyz:
Towards https://github.com/Chatterino/chatterino2/issues/5908.

When suspending the PC or disconnecting for some time, we need to reconnect. We didn't detect this before, because we never did anything with the socket (e.g. writing). Now, we're checking every `keepalive_timeout_seconds * 2` that we received any message.


pajlada (approved):
Things still work as normal, tested local eventsub websocket server doing reconnects

I have not tried things timing out

PR #6067, opened 2025-03-12
fix: hide more pubsub mod actions under streamer mode

iProdigy:
`/warn`, `/(un)monitor`, `/(un)restrict`, `/(add|remove)_(blocked|permitted)_term` were not hidden as moderation actions under streamer mode


pajlada:
Since we handle this properly in EventSub, and next release will not use PubSub for these actions, I kind of left most pubsub changes out on purpose. Do you have a use for PubSub while being on nightly still? EventSub is on by default now

iProdigy:
Fair point, there's no particular use case for pubsub mod actions if eventsub is working (will test this weekend) - a streamer pinged me about this bug but I'll just tell them to use nightly & reauth for latest scopes

PR #6073, opened 2025-03-13
chore: check if WEBP is supported

Nerixyz:
All third party emote providers require WEBP. Some users forget to install qtimageformats. This results in missing images. This PR makes sure we are able to load WEBP images and shows a warning if we aren't.

In the future, we could move PNGs to lossless WEBPs. They're almost always smaller.


pajlada:
In what scenarios can this happen? I imagine only if the user builds Chatterino from source

Nerixyz:
> In what scenarios can this happen? I imagine only if the user builds Chatterino from source

Yes, this or users on Linux where packaging has gone wrong.

pajlada (approved):

PR #6074, opened 2025-03-13
refactor: some mini feature changes in BaseWindow

pajlada:
this renames some features BaseWindow already had and moves some functionality from AccountSwitcherPopup to BaseWindow

### DIFF
--- a/src/widgets/AccountSwitchPopup.cpp
+++ b/src/widgets/AccountSwitchPopup.cpp
@@ -14,14 +14,15 @@ namespace chatterino {
 using namespace literals;
 
 AccountSwitchPopup::AccountSwitchPopup(QWidget *parent)
-    : BaseWindow({BaseWindow::TopMost, BaseWindow::Frameless,
-                  BaseWindow::DisableLayoutSave},
-                 parent)
+    : BaseWindow(
+          {
+              BaseWindow::TopMost,
+              BaseWindow::Frameless,
+              BaseWindow::DisableLayoutSave,
+              BaseWindow::LinuxPopup,
+          },
+          parent)
 {
-#ifdef Q_OS_LINUX
-    this->setWindowFlag(Qt::Popup);
-#endif
-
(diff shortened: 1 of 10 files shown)


src/widgets/AccountSwitchPopup.cpp
```diff
@@ -14,13 +14,16 @@ namespace chatterino {
...
 AccountSwitchPopup::AccountSwitchPopup(QWidget *parent)
-    : BaseWindow({BaseWindow::TopMost, BaseWindow::Frameless,
-                  BaseWindow::DisableLayoutSave},
-                 parent)
+    : BaseWindow(
+          {
+              BaseWindow::TopMost,
+              BaseWindow::Frameless,
+              BaseWindow::DisableLayoutSave,
+              BaseWindow::LinuxPopup,
```
pajlada, line 22: This conditionally sets the `Qt::Popup` flag on the window _if_ on Linux

src/widgets/AccountSwitchPopup.cpp
```diff
@@ -92,11 +95,6 @@ void AccountSwitchPopup::refresh()
     this->ui_.accountSwitchWidget->refresh();
 }
 
-void AccountSwitchPopup::focusOutEvent(QFocusEvent *)
-{
-    this->hide();
```
pajlada, line 97: This functionality was moved to `BaseWindow` with `focusOutAction`

src/widgets/BaseWindow.cpp
```diff
@@ -571,29 +584,6 @@ void BaseWindow::wheelEvent(QWheelEvent *event)
     }
 }
 
-void BaseWindow::onFocusLost()
```
pajlada, line 574: This is now handled directly in the `event` method using the renamed `windowDeactivateAction` member

pajlada:
I have briefly tested this on Linux, macOS & Windows with any popups/windows that _may_ be affected by me doing something wrong, but everything looks good to me

PR #6075, opened 2025-03-13
chore: Enable plugins in builds by default

pajlada:
- **chore: Enable plugins by default**
- **chore: add changelog entry**


Mm2PL:
Do we intend to still support builds without plugins?

pajlada:
> Do we intend to still support builds without plugins?

I don't personally think that's necessary

pajlada:
@Mm2PL The intent right now is:
Change this default
Wait a bit
Stop using the option


This would let us get rid of the big blocks of ifdefs (and reduces the types of "build matrices" we support)

PR #6078, opened 2025-03-15
feat(eventsub): handle multi enum constant keys

pajlada:
enum constants can have `/// extra_enum_constant_names` applied to allow for different strings to match to the same types

this is useful if the documentation is wrong and we're unsure for which is actually correct :-)


pajlada:
image

PR #6079, opened 2025-03-15
chore(eventsub): some more tests & print if message is ignored

pajlada:
- **add some more tests**
- **output to info if a low trust status message is ignored**
- **chore: add changelog entry**


pajlada:
image

PR #6080, opened 2025-03-15
fix: emote button is now rendered smoothlier

pajlada:
- **feat: implement svg rendering for buttons**
- **fix: make emoji an svg so it's rendered smoothlier**


pajlada (commented):
image

PR #6081, opened 2025-03-15
fix: correctly scale SelectChannelDialog

pajlada:
We no longer use a Notebook for this, this means the layout we have will autosize appropriately.
To make it autosize on slightly larger font sizes I've done manual word wrapping on the labels instead of letting them word wrap themselves.

If the user resizes the dialog after opening it, things might look a little bit more goofy than before, but the default size it opens as should be better.

Something with how I add labels now (not using the layout creator) I messed up the focus chain.
I've fixed this by checking 2 steps ahead in the nextInFocuschain and previousInFocusChain.

the way we want this to work, we cannot use Qt's own focus chain logic since the QTextEdit that belongs with the "Channel" radiobox will always be ignored since it's not visible

### DIFF
--- a/src/widgets/dialogs/SelectChannelDialog.cpp
+++ b/src/widgets/dialogs/SelectChannelDialog.cpp
@@ -46,7 +46,6 @@ SelectChannelDialog::SelectChannelDialog(QWidget *parent)
 
     ui.channelLabel = new QLabel("Join a Twitch channel by its channel name");
     ui.channelLabel->setVisible(false);
-    ui.channelLabel->setFocusPolicy(Qt::FocusPolicy::NoFocus);
     layout->addWidget(ui.channelLabel);
 
     ui.channelName = new QLineEdit();
@@ -76,7 +75,6 @@ SelectChannelDialog::SelectChannelDialog(QWidget *parent)
     ui.whispersLabel = new QLabel(
         "Shows the whispers that you receive\nwhile Chatterino is running");
     ui.whispersLabel->setVisible(false);
-    ui.whispersLabel->setFocusPolicy(Qt::FocusPolicy::NoFocus);
     layout->addWidget(ui.whispersLabel);
 
     QObject::connect(ui.whispers, &AutoCheckedRadioButton::toggled, this,
@@ -94,7 +92,6 @@ SelectChannelDialog::SelectChannelDialog(QWidget *parent)
     ui.mentionsLabel = new QLabel(
         "Shows all the messages that\nhighlight you from any channel");
     ui.mentionsLabel->setVisible(false);
-    ui.mentionsLabel->setFocusPolicy(Qt::FocusPolicy::NoFocus);
     layout->addWidget(ui.mentionsLabel);
 
     QObject::connect(ui.mentions, &AutoCheckedRadioButton::toggled, this,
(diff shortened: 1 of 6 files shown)


Nerixyz:
We might want tabs there again if plugins can create their own channels.

pajlada (commented):
This works as expected on Arch Linux with Sway, will test artifacts on Windows/macOS if no one does it before tomorrow

pajlada:
> We might want tabs there again if plugins can create their own channels.

I think I'd prefer if that uses some built in qt tabs or just an additional radio button rather than its own tab - automatic window size hint scaling with notebooks is not good

src/widgets/dialogs/SelectChannelDialog.cpp
```diff
@@ -302,90 +278,109 @@ bool SelectChannelDialog::hasSeletedChannel() const
...
+
+            if ((keyEvent->key() == Qt::Key_Backtab ||
+                 keyEvent->key() == Qt::Key_Down) &&
+                keyEvent->modifiers() == Qt::NoModifier)
+            {
+                return true;
+            }
         }
+        break;
     }
```
Nerixyz, line 378: Missing a default case.

src/widgets/dialogs/SelectChannelDialog.cpp
```diff
@@ -36,159 +31,138 @@ SelectChannelDialog::SelectChannelDialog(QWidget *parent)
...
+    ui.channelLabel = new QLabel("Join a Twitch channel by its channel name");
+    ui.channelLabel->setVisible(false);
+    layout->addWidget(ui.channelLabel);
+
+    ui.channelName = new QLineEdit();
+    ui.channelName->setVisible(false);
+    layout->addWidget(ui.channelName);
+
+    QObject::connect(ui.channel, &AutoCheckedRadioButton::toggled, this,
+                     [this](bool enabled) mutable {
```
Nerixyz, line 56:
```suggestion
                     [this](bool enabled) {
```

Same for the other cases.

Nerixyz (commented):
This wrapping looks a bit arbitrary:

image

Imo, this dialog shouldn't scale at all like the other dialogs that don't display a chat. This would make it possible for the labels to wrap properly. It looks wrong.

src/widgets/dialogs/SelectChannelDialog.cpp
pajlada: not part of this PR but resolved in [`8d02c53` (#6081)](https://github.com/Chatterino/chatterino2/pull/6081/commits/8d02c535036ded4a43deaa3db2fafa3a01b2302e)

src/widgets/dialogs/SelectChannelDialog.cpp
pajlada: fixed in [`34ccfad` (#6081)](https://github.com/Chatterino/chatterino2/pull/6081/commits/34ccfad5f9ff5a3ffa578e8bf6e843c96ebefe36)

pajlada:
word-wrapping returned for some of the labels in [`408747f` (#6081)](https://github.com/Chatterino/chatterino2/pull/6081/commits/408747f80601495bd26cfa32abc08b218ba98cbd)

pajlada:
Tested & works as expected for me on macOS

PR #6092, opened 2025-03-17
fix(eventsub): stack bans

Nerixyz:
In PubSub and IRC, we stack bans. We should do the same with EventSub (In https://github.com/Chatterino/chatterino2/pull/5985 I didn't want to do this, but now I realized that I was wrong). For the message itself, all it takes is to set the `PubSub` flag.
We now set the shared message info correctly on these messages.

Towards #5908.

### DIFF
--- a/src/providers/twitch/eventsub/MessageHandlers.cpp
+++ b/src/providers/twitch/eventsub/MessageHandlers.cpp
@@ -92,8 +92,53 @@ void handleModerateMessage(
         builder.emplaceSystemTextAndUpdate(action.reason.qt(), text);
     }
 
-    builder->messageText = text;
-    builder->searchText = text;
+    builder.setMessageAndSearchText(text);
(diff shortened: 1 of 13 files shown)


src/providers/twitch/eventsub/MessageHandlers.cpp
```diff
@@ -92,8 +92,53 @@ void handleModerateMessage(
...
-    builder->messageText = text;
-    builder->searchText = text;
+    builder.setMessageAndSearchText(text);
+    builder->timeoutUser = action.userLogin.qt();
+
+    auto msg = builder.release();
+    runInGuiThread([chan, msg, time] {
+        chan->addOrReplaceTimeout(msg, time);
+    });
+}
```
Nerixyz, line 102: This diff looks goofy, because I used `builder.setMessageAndSearchText(text);` here (that's the only change in that method).

pajlada (approved):

PR #6096, opened 2025-03-18
fix: focus input after updating visibility

Nerixyz:
When opening the dialog for a regular Twitch channel, the input wasn't focused. With this change, the input is first shown and focused afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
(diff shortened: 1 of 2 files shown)


pajlada:
Will take a peek, thanks

I remember messing about in here because Ctrl+R didn't correctly focus the text

pajlada (approved):
Tested this PR!

| OS | Before | After |
| ------------- | ------------- | ------- |
| Windows 11 | fail | works |
| Linux w/ Sway | works | works |
| macOS | works | works |

PR #6100, opened 2025-03-20
Move "Delete message" to a sub-menu

pajlada:


Mm2PL:
why do we need this?

Wissididom:
> why do we need this?

Someone wrote "I find myself accidentally clicking this all the time when I mean to reply. Would it be possible to get more control over it's display?" in the linked closed PR.

pajlada:
> why do we need this?

I've been accidentally clicking it a few times too. I didn't know the "right click then click reply to message" was so much muscle memory for me, so this is a feature I want to be able to turn off when I'm not doing a lot of deleting

Mm2PL:
I have never encountered this as I mostly use "Right Click then 'r'" to 
reply. I don't think making a setting is a good way to resolve the 
problem of it being too easy to hit accidentally.

I propose trying to make a moderation submenu there, maybe we could put 
the hidden copy msg id there and unhide it. This way it's harder to by 
accident with mouse or keyboard.

W dniu 20.03.2025 o 22:30, pajlada pisze:
> pajlada left a comment (Chatterino/chatterino2#6100)
> 
>> why do we need this?
> 
> I've been accidentally clicking it a few times too. I didn't know the "right click then click reply to message" was so much muscle memory for me, so this is a feature I want to be able to turn off when I'm not doing a lot of deleting
>

iProdigy:
I also like the moderation submenu idea - eventually it can have monitor/restrict buttons when twitch gets around to implementing those endpoints

pajlada:
Moving it to a submenu as `Mo&derate`, with `&Delete message` seems to work fine for the keyboard usecase (right-click, press "D", press "D" to delete a message), and is ok for mouse usecase, even though it looks silly with only "Delete message" there right now.

image

This will still require some muscle memory changes, but it won't easily have you accidentally delete a message.

@VirtuallyJesse would this be an ok compromise for you?

src/widgets/helper/ChannelView.cpp
```diff
@@ -2578,7 +2578,11 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
     if (!layout->getMessage()->id.isEmpty() && twitchChannel &&
         twitchChannel->hasModRights())
     {
-        menu->addAction(
+        menu->addSeparator();
+        auto *moderateAction = menu->addAction("Mo&derate");
```
Nerixyz, line 2582: Hmm, we call it "Moderation" in the settings.

pajlada: My reasoning for writing "Moderate" was to fit with the rest of the right-click options which are written in action form

I'm open to:
a) Leave as is
b) Change to "Moderate..."
c) Change to "Moderation"
d) Change to "Moderation..."

pajlada:
image

PR #6104, opened 2025-03-21
fix: disable precompiled headers for macOS

pajlada:


Nerixyz (changes requested):
This shouldn't be the solution. Since it's only one file, we could compile the ObjC code in another CMake library (without PCH) and link to that.

pajlada:
I've clarified in the changelog entry & in the cmake option that this is a temporary fix

PR #6111, opened 2025-03-23
fix cmd+backspace behavior in main chat dialog input

### LINKED ISSUE #4755: Macos: Command + Backspace doesn't clear text in main chat dialog window
### Checklist

### Describe your issue

Command + Backspace doesn't clear typed text in the main chat dialog window, however it works as intended in the find dialog.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.4.4 (commit 29a14627) built with Qt 6.5.0 Running on macOS Ventura (13.5), kernel: 22.6.0

jacob-thompson:
fixes #4755 

it seems that [QTextEdit](https://doc.qt.io/qt-6/qtextedit.html#editing-key-bindings) does not handle this behavior on its own like [QLineEdit](https://doc.qt.io/qt-6/qlineedit.html#default-key-bindings) does. handling the behavior explicitly fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Bugfix: Don't set default binding for "Toggle local R9K" on macOS. Was <kbd>CTRL</kbd> + <kdb>H</kdb> before, which clashes with a system binding. (#5764)
 - Bugfix: Don't add moderation buttons to your own Usercard. (#6107)
 - Dev: Temporarily disable precompiled header support for macOS. (#6104)
-- Bugfix: Fix <kbd>CMD</kbd> + <kbd>BACKSPACE</kbd> not clearing main chat dialog input. (#6111)
+- Bugfix: Handle <kbd>CMD</kbd> + <kbd>BACKSPACE</kbd> behavior explicitly in main chat dialog input for macOS. (#6111)
 
 ## 2.5.3-beta.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Handle <kbd>CMD</kbd> + <kbd>BACKSPACE</kbd> behavior explicitly in main chat dialog input for macOS. (#6111)
+
 ## 2.5.3
 
 - Minor: Shared chat messages now use the source channel's profile picture to denote it's a shared chat message. (#5760)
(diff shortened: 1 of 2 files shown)


Mm2PL:
Are you really sure this is correct? [Qt's documentation says that Cmd+Backspace should remove the word before the cursor](https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts), not everything before the cursor.

We have selection behavior changes for the emote popup that ensure that `QKeySequence::DeleteStartOfWord` (or whatever key combo invokes it) [deletes the selection before deleting the word before the cursor](https://github.com/Chatterino/chatterino2/blob/master/src/widgets/dialogs/EmotePopup.cpp#L515). I would like the behavior of that to behave the same way in all inputs on a given operating system.

jacob-thompson:
> Are you really sure this is correct? [Qt's documentation says that Cmd+Backspace should remove the word before the cursor](https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts), not everything before the cursor.

the behavior you describe is the result of pressing Alt+Backspace on macOS, as it says in that documentation next to `DeleteStartOfWord`. this behavior wouldn't be defined in that documentation as a "standard shortcut" since this is just a macOS thing

Nerixyz:
> We have selection behavior changes for the emote popup that ensure that `QKeySequence::DeleteStartOfWord` (or whatever key combo invokes it) [deletes the selection before deleting the word before the cursor](https://github.com/Chatterino/chatterino2/blob/master/src/widgets/dialogs/EmotePopup.cpp?rgh-link-date=2025-03-23T19%3A49%3A01Z#L515). I would like the behavior of that to behave the same way in all inputs on a given operating system.

That behavior is done for line edits when pressing the platform shortcut for `DeleteStartOfWord` [on 6.5+](https://github.com/qt/qtbase/commit/1d953b2fbc3deccad6eb2a5ebe39b24908dbcf3e). I wonder why there's a special case for [CTRL+backspace specifically](https://github.com/qt/qtbase/blob/5be58ac9271555eed3a2884d827f30c164d2277d/src/widgets/widgets/qwidgetlinecontrol.cpp#L1855-L1861). Seems like it was there from the start and I forgot it... But it's not in `QTextEdit` for some reason...

pajlada (approved):
Works as expected on my macbook, thanks for noticing _and_ fixing!

PR #6116, opened 2025-03-25
Fix: Don't create native messaging manifest file if browser directory doesn't exist

### LINKED ISSUE #6115: Don't create native messaging manifest file if browser directory doesn't exist
### Checklist

### Describe your issue

Change that introduced this behavior: #6031

After updating to the 2.5.3 AppImage on Linux, I noticed some files created by chatterino at `~/.mozilla/native-messaging-hosts` and `~/.config/google-chrome/NativeMessagingHosts/`. Since I use my distribution's Chromium package and don't use either Firefox or Chrome, I deleted them but then realized these files are recreated each time chatterino is launched. For the sake of reducing clutter and unnecessary files in the user's home directory (especially in the case of `~/.mozilla/` since Firefox still doesn't support [XDG Base Spec](https://bugzilla.mozilla.org/show_bug.cgi?id=259356)), I think chatterino should only create these files if the browser specific directory exists since we can assume the user has installed and used the browser. This issue probably applies to other platforms as well.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.2 Running on Arch Linux, kernel: 6.13.8-arch1-1

4rneee:
Fixes #6115

I did not find a clean way of checking this without creating a new `QDir`(or similar) object since `dir.cdUp()` changes the dir if it succeeds and `dir.exists("..")` returns false if `dir` does not exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Don't create native messaging manifest file if browser directory doesn't exist. (#6116)
+
 ## 2.5.3
 
 - Minor: Shared chat messages now use the source channel's profile picture to denote it's a shared chat message. (#5760)
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -77,6 +77,15 @@ void writeManifestTo(QString directory, const QString &filename,
     }
 
     QDir dir(directory);
+
+    // check if the parent directory (i.e. the browsers directory) exists
+    // if it does not exist we abort because it means that the browser is not installed
+    QDir dirCpy = dir;
+    if (!dirCpy.cdUp())
+    {
+        return;
+    }
+
     if (!dir.mkpath(u"."_s))
     {
         qCWarning(chatterinoNativeMessage) << "Failed to create" << directory;
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -27,6 +27,7 @@
 
 namespace {
 
+using namespace chatterino::nm::detail;
 using namespace chatterino;
 using namespace chatterino::literals;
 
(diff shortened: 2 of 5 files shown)


Nerixyz:
> I did not find a clean way of checking this without creating a new `QDir`(or similar) object since `dir.cdUp()` changes the dir if it succeeds and `dir.exists("..")` returns false if `dir` does not exist.

Since we're on Qt 6 now, we could also use `std::filesystem::path` ([example](https://godbolt.org/z/oGboYaTG1)), but that's verbose too (easier in Qt 6.7+ because of the added string view operator).

pajlada:
If we're not looking to create all subdirectories, we can use [QDir::mkdir](https://doc.qt.io/qt-6/qdir.html#mkdir-1) or [std::filesystem::create_directory](https://en.cppreference.com/w/cpp/filesystem/create_directory) instead. That way we let that check for the existance of the parent directory instead

4rneee:
> we can use [QDir::mkdir](https://doc.qt.io/qt-6/qdir.html#mkdir-1)

If I read the documentation and code for `QDir::mkdir` correctly it can only be used to create subdirectories. This means that we would first require a QDir object of the parent path. For this we could extent the Config struct to contain the path to the browser's directory or use `QDir::cdUp`.

> or [std::filesystem::create_directory](https://en.cppreference.com/w/cpp/filesystem/create_directory)

This seems easier and we could use the following code instead of the call to `mkpath`:
```C++
if (!dir.exists())
{
    std::error_code ec;
    if (!std::filesystem::create_directory(directory.toStdString(), ec) ||
        ec)
    {
        qCWarning(chatterinoNativeMessage)
            << "Failed to create" << directory;
        return;
    }
}
```
And maybe change/remove the warning message as it would be intended that the creation fails in certain cases.

pajlada:
I'm fine to modify the Config struct to allow for `QDir::mkdir`

4rneee:
> I'm fine to modify the Config struct to allow for `QDir::mkdir`

I implemented this now in the latest commit

pajlada:
> ## Pull Request Overview
> 
> This pull request addresses a bug by preventing the creation of a native messaging manifest file when the browser directory does not exist.
> 
>     * Adds a changelog entry documenting the fix.
> 
>     * Updates the documentation to clarify that the file is not created without a valid directory.
> 
> 
> Files not reviewed (4)
> 
>     * **src/singletons/NativeMessaging.cpp**: Language not supported
> 
>     * **src/singletons/NativeMessaging.hpp**: Language not supported
> 
>     * **tests/CMakeLists.txt**: Language not supported
> 
>     * **tests/src/NativeMessaging.cpp**: Language not supported

wow, thanks!

pajlada (approved):

PR #6117, opened 2025-03-27
chore(conan): don't create `CMakeUserPresets.json`

Nerixyz:
I was building with VisualStudio recently and used CMakePresets, but it conflicted with conan, because conan would create `CMakeUserPresets.json` which confused VS. The file that conan generates isn't useful.

I should revise the documentation for builds on Windows and add at least a small section about VS and CMakePresets, because with [cmake-conan](https://github.com/conan-io/cmake-conan), they're nice to use and will result in a vcpkg-like experience in VS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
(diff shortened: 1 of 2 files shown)


pajlada (approved):
works for me and makes no difference for my windows build process

Issue #6118, opened 2025-03-28
Remove reply menu item from context menu of deleted messages


hemirt:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Deleted messages cannot be replied to, yet the UI allows it.

![Image](https://github.com/user-attachments/assets/2bf7453b-edcc-4ee3-b47d-a7e1a8cdd7a3)

(Received Helix error: The request's input was not valid. Message cannot be replied to.)

Thus the option should be removed.

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 11 Version 24H2, kernel: 10.0.26100

hemirt:
Twitch is not allowing replies to messages in threads whose root has been deleted.

PR #6119, opened 2025-03-28
fix context menu extra option to reply to invalid messages

hemirt:
fixes https://github.com/Chatterino/chatterino2/issues/6118

adds InvalidReplyTarget flag to messages that are 1) single-deleted, 2) deleted via user timeout (CLEARCHAT user)

such messages are invalid targets to reply and the reply wont go through

interesting thing to note, that each message in a replythread is invalid if the root is deleted (Twitch behaviour)


hemirt:
thing to consider, whether to add a greyed out menu item instead
i.e.

"Cannot reply to a deleted message"
"Cannot reply to a message in a thread with deleted root"

(or obviously some other shorter variants)

However it is possible to reply to original message via a deleted message in the thread, thus seeing both the greyed out "Cannot reply to a deleted message" and "Reply to original message" might be confusing for some.

hemirt:
Q. Why add in new flag instead of using the "Disabled" flag
A. Disabled flag is used in various places, usually for the "Disabled" styling of the message.

Q. Why is not a simple check for flag possible?
A. Because Twitch disables replies for messages in reply threads if the reply thread root message gets deleted, and because the reply thread root message does not reference its own "replyThread" (as per Message.hpp), flag-setting would require full downwards traversal on message-deletion, instead of single-hop upwards-traversal only on message click.

Wissididom:
> thing to consider, whether to add a greyed out menu item instead

Personally I think it should be greyed out if easily possible, because I can imagine people with muscle memory to then click something else they didn't mean to click. I haven't looked how it currently looks like yet, tho.

hemirt:
For tests,
https://github.com/hemirt/chatterino2/blob/24fbab99138dc3475232b5459d97fdf936f9889e/tests/snapshots/IrcMessageHandler/rm-deleted.json#L248

I think now also requires `InvalidReplyTarget`

src/messages/MessageFlag.hpp
```diff
@@ -65,6 +65,14 @@ enum class MessageFlag : std::int64_t {
     ///  - forsen added "blockedterm" as a blocked term
     ///  - Your message is being checked by mods and has not been sent
     ModerationAction = (1LL << 41),
+    /// The message is not a valid reply target
+    /// Examples:
+    ///  - message was deleted via single channel chat message deletion (IRC: CLEARMSG, EVENTSUB: channel.chat.message_delete)
+    ///  - message was deleted via chat clear user messages (IRC: CLEARCHAT(user), EVENTSUB: channel.chat.clear_user_messages)
+    /// However as a reply thread root does not know its own thread [The root of the thread does not have replyThread set.(Message.hpp)]
+    ///  - messages that are invalid reply targets due to the root being an invalid reply target
+    ///    are NOT marked as invalid reply targets themselves, thus it is required to check their possible reply thread roots.
```
Nerixyz, line 74: Maybe: "Note: If this message is inside a reply thread, the root must not have the flag either."

src/messages/MessageFlag.hpp
```diff
@@ -65,6 +65,14 @@ enum class MessageFlag : std::int64_t {
     ///  - forsen added "blockedterm" as a blocked term
     ///  - Your message is being checked by mods and has not been sent
     ModerationAction = (1LL << 41),
+    /// The message is not a valid reply target
```
Nerixyz, line 68:
```suggestion
    /// The message can't be replied to
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -2557,20 +2557,36 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
...
         if (messagePtr->replyThread != nullptr)
         {
-            menu->addAction("Reply to &original thread", [this, &messagePtr] {
-                this->setInputReply(messagePtr->replyThread->root());
-            });
+            const auto &rootPtr = messagePtr->replyThread->root();
+            if (!rootPtr->flags.has(MessageFlag::InvalidReplyTarget))
+            {
+                // if the root of the thread is an invalid reply target (i.e. has been deleted)
+                // the whole thread is now invalid reply target (Twitch behaviour)
```
Nerixyz, line 2566: This should be above the `if` - it looks like this is for the `messagePtr->flags.has(MessageFlag::InvalidReplyTarget)` check.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Bugfix: Fixed inputs in dialogs not having a border around and padding in them. (#6098)
 - Bugfix: Don't set default binding for "Toggle local R9K" on macOS. Was <kbd>CTRL</kbd> + <kdb>H</kdb> before, which clashes with a system binding. (#5764)
 - Bugfix: Don't add moderation buttons to your own Usercard. (#6107)
+- Bugfix: Don't add reply buttons to messages that are invalid reply targets. (#6119)
```
Nerixyz, line 13: Should be under `Unversioned`.

Nerixyz (commented):
I agree that we need a new flag for this. `Disabled` is too different.

A snapshot for messages affected by timeouts (and `/clear` later) could be nice. Something like

```json
{
	"input": "timeout",
    "prevMessages": [msg1, msg2],
    "nAdditional": 2
}
```

Feel free to copy IRC messages from other snapshots.

CHANGELOG.md
hemirt: moved

src/widgets/helper/ChannelView.cpp
hemirt: moved

src/messages/MessageFlag.hpp
hemirt: changed

src/messages/MessageFlag.hpp
hemirt: changed

hemirt:
> I agree that we need a new flag for this. `Disabled` is too different.

yeah, that's why I added one


> A snapshot for messages affected by timeouts (and `/clear` later) could be nice. Something like
> 
> ```json
> {
> 	"input": "timeout",
>     "prevMessages": [msg1, msg2],
>     "nAdditional": 2
> }
> ```
> 
> Feel free to copy IRC messages from other snapshots.

Not quite sure what do you mean here and how it pertains to current PR.
However the /clear command works a bit differently to the /clear(user) that happens on timeout/ban, and the delete-message. Full-clear chat messages _are_ able to be replied to.

Too old messages can't be replied to, however I was unable to find any mention of the threshold, so far I have tested it to be between 24:47 and 26 hours. However I do not know if it's affected by anything other than the time between those messages.

hemirt:
Note:
reply buttons

src/messages/MessageFlag.hpp
```diff
@@ -65,6 +65,12 @@ enum class MessageFlag : std::int64_t {
     ///  - forsen added "blockedterm" as a blocked term
     ///  - Your message is being checked by mods and has not been sent
     ModerationAction = (1LL << 41),
+    /// The message can't be replied to
+    /// Examples:
+    ///  - message was deleted via single channel chat message deletion (IRC: CLEARMSG, EVENTSUB: channel.chat.message_delete)
+    ///  - message was deleted via chat clear user messages (IRC: CLEARCHAT(user), EVENTSUB: channel.chat.clear_user_messages)
+    /// Note: If this message is inside a reply thread, the root must not have the flag either.
+    InvalidReplyTarget = (1LL << 42),
```
fourtf, line 73: I would invert this, calling it something like `UserCanReply`. I think the flags should each be off by default and be enabled if a special condition is met. The default here is that you can't reply to a message (if we're thinking of messages as an item in the chat and not a real message from a chatter).

hemirt: I'm not sure I understand it, but by default each message is reply-able, unless it has been deleted or its reply-thread root was deleted (deleted = deleted as single message, or as timeouted user), or 24 hours has passed since it was sent.

To have this inverted would mean that pretty much every message would almost always have the flag turned on, instead of some special few messages.

hemirt: And before the invalidReplyTarget flag gets resolved, on right click when building the menu, we only consider messages that are replyable

https://github.com/Chatterino/chatterino2/blob/07ad07d5258c5fe26ca7ca5022e4a4b69ae952ab/src/messages/layouts/MessageLayout.cpp#L517-L533

hemirt:
> * [x]  make historic messages check the time so that they are or are not replyable
>   -> proposed solution > check on right click
>   -> (because there's no need to add more processing while loading historic messages, as there can be multitudes of channels)
> * [x]  invalidate replyability of old messages [not those loaded as historic] in a live (or reconnect) scenario ?
>   -> proposed solution > check on right click

as mentioned, fixed when message is invalidated by time, via the proposed solution of check on click

hemirt:
last issue are the reply buttons, unsure what to do about them, I was not able to make them disappear, unless using

getApp()->getWindows()->forceLayoutChannelViews();
and adding here https://github.com/Chatterino/chatterino2/blob/d3bab9132f66898783e96abd104e736c10f9a99d/src/messages/layouts/MessageLayout.cpp#L201 the check for the message flag for the element of reply button

this forced layout would happen only on these clearMsgs messages, so that might be fine, wasnt able to find out how to force just one channel (layoutChannelViews(channel) did not seem to work)

however idk how the timed invalidation should work in that regards

might just be better instead to not open the reply message input and send a system message about invalid message upon clicking the reply button

it might also be good to disable input in the "popped out" thread window, if it is not possible to reply

pajlada (changes requested):
Move the logic of `MessageLayout::isReplyable` to `Message`, then add your `InvalidReplyTarget` and timestamp checking logic into there instead

hemirt:
> Move the logic of `MessageLayout::isReplyable` to `Message`, then add your `InvalidReplyTarget` and timestamp checking logic into there instead

sorry it took some time, was dealing with IRL stuff

moved the logic into message

also made it so that clicking on the reply message-element icon on a nonreplyable message does nothing

however as I said, I have no idea how to remove the icon (best would be change to view thread probably at the message deletion instead), _and_ it should also be made not-possible to try to type into the "reply-thread popup" when it is non-replyable (another thing to look at)

however thus so far, the context menu and replyability checking is resolved, if needed try catch me in my chat/notify, I know my comments are sometimes hard to understand

Mm2PL:
> however as I said, I have no idea how to remove the icon (best would be change to view thread probably at the message deletion instead) [...]

A thing you could try is removing its ReplyButton MessageElementFlag. However this may require a relayout to actually hide these buttons. I am not fully sure

hemirt:
> > however as I said, I have no idea how to remove the icon (best would be change to view thread probably at the message deletion instead) [...]
> 
> A thing you could try is removing its ReplyButton MessageElementFlag. However this may require a relayout to actually hide these buttons. I am not fully sure

sounds "expensive", however sounds doable, since it would only be done on 1) user-timedout 2) message-deleted, but probably c/sh-ould be ignored for messages that become older than 1 day (and thus non-replyable due to their age)

src/messages/Message.hpp
```diff
@@ -62,6 +62,8 @@ struct Message {
     // The root of the thread does not have replyThread set.
     std::shared_ptr<MessageThread> replyThread;
     MessagePtr replyParent;
+    enum class ReplyStatus { NotReplyable, Replyable, ReplyableWithThread };
```
pajlada: Comments on this enum would be nice

If I understand it correctly:
NotReplyable means: This message cannot be replied to
Replyable: This message can be replied to
ReplyableWithThread: This message can be replied to, and is also part of a thread that you can reply to

src/widgets/helper/ChannelView.cpp
```diff
@@ -2594,25 +2594,42 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
...
+                    "Reply to &original thread", [this, &messagePtr] {
+                        this->setInputReply(messagePtr->replyThread->root());
+                    });
+            }
         }
     }
 
+    if (const auto &messagePtr = layout->getMessagePtr();
+        messagePtr->replyThread != nullptr)
+    {
```
pajlada: You've moved this logic out of the `if (this->canReplyToMessages())` scope, this means you can right-click view thread inside a thread popup, which wasn't possible before.
Is there some reasoning behind this change I don't see?

hemirt: originally to be able to view thread of nonreplyable threads, however you're right, it is supposed to be inside the channelView->canReplyToMessages check

src/messages/Message.hpp
hemirt: yes, replyable with thread means that the message can be replied to and is a part of a thread (which implies that the thread must be replyable, otherwise the message could not have been replyable)

hemirt:
I fixed the issues, also made the reply status more explicit (the 5 possible states), thus it now shows what it should on each option, and the option is grayed out if not available where it makes sense


I did however maybe found a bug, if you open a thread (view thread), (without autoclose on unfocus option), and close main chatterino, I had some race condition issue destructing the popup, (assert(STOPPED.load() == false); fired up in Application.cpp:667). Should be unrelated to anything I did though. The series of events seemed to be ~ReplyThreadPopup, ~DraggablePopup, ~BaseWindow, ~BaseWidget, ~TwitchChannel (first line) getApp(), might be worth looking at if its possible to replicate, just started chatterino in debug mode, and opened any thread via view thread, and then clicked the X on the main window

src/messages/Message.hpp
hemirt: made it more explicit

src/messages/Message.hpp
```diff
@@ -62,6 +62,14 @@ struct Message {
     // The root of the thread does not have replyThread set.
     std::shared_ptr<MessageThread> replyThread;
     MessagePtr replyParent;
+    enum class ReplyStatus {
```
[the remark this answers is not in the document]
hemirt: done

pajlada (approved):

pajlada:
the assert you mentioned should be fixed in https://github.com/Chatterino/chatterino2/pull/6271

PR #6120, opened 2025-03-29
feat: add REPL for plugins

Nerixyz:
This PR adds a small REPL for plugins. Sometimes, I want to try something out when testing some plugin stuff, or I simply want to view the log messages. This REPL provides an easy interface for that without needing a debugger or starting Chatterino from the command line.

With the REPL, you can
- Run Lua code in the global scope (implicitly adds `return ` unless input is prefixed with `@ ` - if the prefixed version fails to parse, the original code is executed)
- Read log messages
- Reload the plugin
- View tables (prefix input with `!` to expand all items)

Additionally, the font can be customized (the default monospace font is usually quite bad) and the window can be pinned. It works on light and dark mode - I used the console colors from Firefox's DevTools for errors/warnings.

The goal was to be simple and helpful. The window is attached to an ID and listens for plugins that are loaded with this ID (to survive reloads). When stringifying I used a custom function (compared to `luaL_tolstring`) to (1) expand tables and (2) avoid errors from implicitly calling Lua code through `__tostring`. Users can still call `print` to have this functionality.


pajlada (commented):
Are the icons from [Fluent icons](https://github.com/microsoft/fluentui-system-icons)?

src/widgets/settingspages/PluginsPage.cpp
```diff
@@ -231,6 +232,13 @@ void PluginsPage::rebuildContent()
         {
             reloadButton->setEnabled(false);
         }
+
+        auto *replButton = new QPushButton("Open REPL", this->dataFrame_);
+        QObject::connect(replButton, &QPushButton::clicked, [id]() {
+            auto *repl = new PluginRepl(id);
+            repl->show();
+        });
+        pluginEntry->addRow(replButton);
```
pajlada, line 241: My first instinct is to only have this button available when Chatterino is compiled in debug mode.

I think I'd prefer that to start, and we can think of how we can safely allow this in non-debug modes later, maybe something like
Extra run flag? `chatterino.exe --enable-lua-repl`
Environment variable? `CHATTERINO_ENABLE_LUA_REPL=1`
Hidden setting in `settings.json` that you can only enable by manually editing settings.json? `{"I_KNOW_WHAT_IM_DOING": true}`

src/controllers/plugins/Plugin.hpp
```diff
@@ -154,6 +171,9 @@ class Plugin
     std::vector<QTimer *> activeTimeouts;
     int lastTimerId = 0;
 
+    OnDestroyed onDestroyed_;
```
pajlada, line 174: I think I'd prefer if the signals were publicly available, and consumers access them & call `connect` themselves

pajlada (changes requested):
Overall looks good, some nits with permissions/naming/use of signals for the first pass.
Haven't played around with the UI myself yet

src/widgets/settingspages/PluginsPage.cpp
Nerixyz: I would like to have them be available in release builds without too much effort. The goal was to allow developers, also people who just started, to easily debug their plugins. I could see the following: unconditionally enable in debug mode, enable if the author name matches the current user, always enable if a hidden setting is enabled.

Nerixyz:
> Are the icons from [Fluent icons](https://github.com/microsoft/fluentui-system-icons?rgh-link-date=2025-04-06T11%3A56%3A40Z)?

Yes.

src/widgets/settingspages/PluginsPage.cpp
pajlada: I'd like the REPL to be completely disabled unless a user has started chatterino with the `--enable-lua-repl` flag (I'm ok with different naming)

Mm2PL: imo if we must hide it, we should do the hidden setting. It's easy enough to explain to other programmers and persistent so a user won't need to enable it every time the use the program which is a pain in the ass

Mm2PL (changes requested):
I had forgotten to click Review

src/widgets/settingspages/PluginsPage.cpp
pajlada: What I don't like with it is how easy it is to just brick your settings, but I am ok with it

src/widgets/PluginRepl.cpp
```diff
@@ -0,0 +1,706 @@
...
+        auto v = QStringView(code).sliced(1).trimmed();
+        auto nChars = v.constData() - code.constData();
+        code.remove(0, nChars);
+    }
+
+    if (code.startsWith(u"@ "))
+    {
+        code.remove(0, 2);
+    }
+    else if (!code.startsWith(u"return "))
```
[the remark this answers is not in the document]
Nerixyz: Removed `@`. `!` is still in there to force expand types without any limit.

src/singletons/Settings.hpp
```diff
@@ -645,6 +645,12 @@ class Settings
 
     BoolSetting showSendButton = {"/ui/showSendButton", false};
 
+    struct {
+        QStringSetting fontFamily = {"/ui/pluginRepl/fontFamily", {}};
```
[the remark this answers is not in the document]
Nerixyz: An empty font resolves to the default monospace font for the platform. I added a comment about that now.

Nerixyz:
We should use clang-format 20 already....

Mm2PL (changes requested):
This can execute stuff in the context of an unloaded plugin, but only when opened with the plugin already disabled. I would like this either not be able to execute things in the context of disabled plugins or always be able to. 
Assuming we go for the no unloaded eval route, I would also like the input placeholder to change when the plugin is unloaded and have a warning while trying to eval code. Ideally not outright disabling the input because then it's harder to copy stuff out of history.
This may be a weird ask but I'd actually like there to be a way to reload the target plugin, ideally an Action. This may be difficult to do with hiding either the category or an action in Popup Windows. For default hotkeys probably bind it to `F5`.

src/widgets/PluginRepl.cpp
```diff
@@ -0,0 +1,718 @@
...
+        default:
+            s.append(stringifyValue(it));
+    }
+}
+
+}  // namespace
+
+namespace chatterino {
+
+PluginRepl::PluginRepl(QString id, QWidget *parent)
```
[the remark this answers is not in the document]
Nerixyz: I added a check if a plugin is enabled.

PR #6121, opened 2025-03-30
Fixes #3545 Allow disabling fading recent messages

### LINKED ISSUE #3545: Faded text is difficult to read

fourtf:

### DIFF
--- a/src/widgets/settingspages/GeneralPage.cpp
+++ b/src/widgets/settingspages/GeneralPage.cpp
@@ -505,6 +505,13 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                      "message to help better tell them apart.")
         ->addTo(layout);
 
+    SettingWidget::checkbox("Reduce opacity of past messages",
+                            s.fadeRecentMessages)
+        ->setTooltip(
+            "Reduce opacity of messages that were posted before Chatterino "
+            "was started or while re-connection.")
+        ->addTo(layout);
+
     SettingWidget::checkbox("Hide deleted messages", s.hideModerated)
         ->setTooltip(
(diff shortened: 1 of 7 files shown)


src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -505,6 +505,13 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                      "message to help better tell them apart.")
         ->addTo(layout);
 
+    SettingWidget::checkbox("Reduce opacity of past messages",
```
Mm2PL, line 508: i think we call them 'historical' in the UI?

Felanbird: we call them `history` even tho `historical` is the correct word

fourtf: So true I changed it to "message history" which is the name some other settings also use

pajlada (approved):

PR #6123, opened 2025-03-31
make paused chat indicator more apparent

### LINKED ISSUE #2688: Paused chat indicator should be less discrete
The current indicator isn't so easy to see for someone like me, sometimes I forget that chat is paused due to my mouse being hovered over it. Given the chat moves from the bottom, it doesn't make sense to have the indicator up the top.

We should either move the pause indicator down the bottom, or maybe change it to something less discrete (the latter is preferred). Maybe another setting would be good. Something like webchat would be good:

webchat lul

jacob-thompson:
fixes #2688

the paused chat indicator in its current form is a little bit difficult to see. I am suggesting:

- adding an outline so that the indicator stands out from other widget elements such as chat timestamps
- moving the indicator to the bottom of the widget since users' eyes are more likely to be focused on the bottom half of the widget than the top half of the widget
- adding "Chat Paused" text next to the symbol to make the indicator bigger and easier to notice

# screenshots

<details>
<summary>before</summary>

</details>

<details>
<summary>after (dark theme)</summary>

</details>

<details>
<summary>after (black theme)</summary>

</details>

<details>
<summary>after (light theme)</summary>

</details>

<details>
<summary>after (white theme)</summary>

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Unversioned
 
+- Minor: Make paused chat indicator more visible. (#6123)
 - Bugfix: Don't create native messaging manifest file if browser directory doesn't exist. (#6116)
 - Dev: Conan will no longer generate a `CMakeUserPresets.json` file. (#6117)
-- Minor: Make paused chat indicator more visible. (#6123)
 
 ## 2.5.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Make paused chat indicator more visible. (#6123)
+- Minor: Make paused chat indicator more visible, and fix its zoom behavior. (#6123)
 - Bugfix: Don't create native messaging manifest file if browser directory doesn't exist. (#6116)
 - Dev: Conan will no longer generate a `CMakeUserPresets.json` file. (#6117)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bugfix: Don't create native messaging manifest file if browser directory doesn't exist. (#6116)
 - Dev: Conan will no longer generate a `CMakeUserPresets.json` file. (#6117)
+- Minor: Make paused chat indicator more visible. (#6123)
 
 ## 2.5.3
 
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -1549,42 +1549,27 @@ void ChannelView::paintEvent(QPaintEvent *event)
     if (this->paused())
     {
         auto a = this->scale() * 20;
-        auto color = QColor(127, 127, 127, 255);
+        auto color = QColor(180, 180, 180, 255);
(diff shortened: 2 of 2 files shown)


Nerixyz:
I don't think the indicator should be at the bottom. Considering we have "Pause chat on hover", users likely want to read the last messages. Imo, an outline around the pause icon should be fine. Or maybe just the string "paused".

iProdigy:
Could you experiment with a red outline around the pause icon (and no text)?

jacob-thompson:
> Imo, an outline around the pause icon should be fine. Or maybe just the string "paused".

I think if anything the icon should stay:

<details>
<summary>screenshot</summary>

</details>

I do feel, though, that the outlined icon alone still may be difficult to notice for those with poor vision. how would you feel about also making it a touch bigger, @Nerixyz ?:

<details>
<summary>screenshot</summary>

</details>

jacob-thompson:
> Could you experiment with a red outline around the pause icon (and no text)?

certainly pops out more since red is a hot color, but I just think it looks ugly:

<details>
<summary>top-left, red outline</summary>

</details>

I feel it'd look especially ugly if the outline blended with chat elements such as red chatters' names

pajlada:
To confirm previous feedback, position shouldn't change with this PR but feel free to experiment more what makes it pop out and isn't ugly

Nerixyz:
> I think if anything the icon should stay:

Oh, sorry for the confusion. I meant that the icon should stay, but the text should only say paused.

> I do feel, though, that the outlined icon alone still may be difficult to notice for those with poor vision.

What about adding a background similar to the tooltip, and possibly making the pause icon lighter? If we do that, we can ensure sufficient contrast on both light and dark themes:



(should probably be smaller and the lighter color shouldn't be `#ffffff`)

jacob-thompson:
> What about adding a background similar to the tooltip, and possibly making the pause icon lighter? If we do that, we can ensure sufficient contrast on both light and dark themes

I feel this looks nice:

<details>
<summary>dark theme screenshot</summary>

</details>


<details>
<summary>light theme screenshot</summary>

</details>

in this case the icon & text are `#B4B4B4`

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Bugfix: Don't create native messaging manifest file if browser directory doesn't exist. (#6116)
 - Dev: Conan will no longer generate a `CMakeUserPresets.json` file. (#6117)
+- Minor: Make paused chat indicator more visible. (#6123)
```
pajlada, line 7: Move to the top of the changelog entries. They're generally sorted like this:
Major
Minor
Bugfix
Dev

pajlada (changes requested):
Not part of your changes, but the Pause icon does not handle different zoom levels well.
Would you like to fix that in this PR?

pause-0 5x-zoom
pause-2x-zoom
pause-1x-zoom

jacob-thompson:
> Not part of your changes, but the Pause icon does not handle different zoom levels well. Would you like to fix that in this PR?

fixed

pajlada (approved):
Looks good, thank you!

PR #6124, opened 2025-03-31
Add cached emotes fallback for failed provider fetches

occluder:
This PR introduces a cached fallback mechanism for emotes when provider API requests fail, lessening the pain of service hiccups

- Emote loading now starts with cached emotes (if available) to display them immediately
- An async call is then made to fetch latest data
- If the call succeeds, the response is used and the cache gets updated
- If the call fails, the cached version remains loaded, and a system message alerts users about the fallback

Things I'd like to note:
1. This is only for third-party emote providers, Twitch global/channel emotes aren't considerd
2. Only the cached emotes are loaded from the stored caches, meaning this currently will not load ffz's custom vip/mod badges or 7tv's emoteSet string even though they are saved

Closes #5676


occluder:
I'll fix my fork and re-open

PR #6125, opened 2025-03-31
Add cached emotes fallback for failed provider fetches

occluder:
This PR introduces a cached fallback mechanism for emotes when provider API requests fail, lessening the pain of service hiccups

- Emote loading now starts with cached emotes (if available) to display them immediately
- An async call is then made to fetch latest data
- If the call succeeds, the response is used and the cache gets updated
- If the call fails, the cached version remains loaded, and a system message alerts users about the fallback

Things I'd like to note:
1. This is only for third-party emote providers, Twitch global/channel emotes aren't considerd
2. Only the cached emotes are loaded from the stored caches, meaning this currently will not load ffz's custom vip/mod badges or 7tv's emoteSet string even though they are saved

Closes #5676


occluder:
> Haven't looked at the changes (which would answer this,) but would be interested to know if you'd start seeing emotes in messages pulled from recent messages on startup 🤔

Isn't this already the case? I'm on latest nightly and I see what you described

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -218,18 +222,35 @@ void BttvEmotes::loadEmotes()
...
+    readProviderEmotesCache("global", "bttv", [this](auto jsonDoc) {
+        auto emotes = this->global_.get();
+        auto pair = parseGlobalEmotes(jsonDoc.array(), *emotes);
+        if (pair.first)
+        {
+            this->setEmotes(std::make_shared<EmoteMap>(std::move(pair.second)));
+        }
+    });
+
+    std::ignore = QtConcurrent::run([this]() {
```
Nerixyz, line 234: This goes for all of these uses: We don't need to initiate this in a different thread. The network request will already run in a different thread. You should've seen assertions fail if you built in debug mode.

src/providers/bttv/BttvEmotes.cpp
```diff
@@ -218,18 +222,35 @@ void BttvEmotes::loadEmotes()
...
-            auto emotes = this->global_.get();
-            auto pair = parseGlobalEmotes(result.parseJsonArray(), *emotes);
-            if (pair.first)
-            {
-                this->setEmotes(
-                    std::make_shared<EmoteMap>(std::move(pair.second)));
-            }
-        })
-        .execute();
+    readProviderEmotesCache("global", "bttv", [this](auto jsonDoc) {
```
Nerixyz, line 225: This goes for all of these uses: This could be done concurrently. We could have one shared state (e.g. `std::shared_ptr<bool>`) to indicate that the main request is done. Parsing of the emotes could be done there as well. Only the `setEmotes` (or similar) should happen on the GUI thread (`runInGuiThread`):

```cpp
// something like this
std::shared_ptr<bool> state = std::make_shared<bool>(false);
// this reads in another thread
readProviderEmotesCache("..", "..", [this, state](const auto&jsonDoc) {
  auto map = parseEmotesOrWhatever(jsonDoc);
  // we need this for global emotes as well, because the state isn't a mutex
  runInGuiThread([this, map{std::move(map)}]() mutable {
    if (!*state) setEmotes(std::move(map));
  });
});
fetchEmotes([this, state](const auto&res) {
  // ^ if the network request isn't .concurrent(), we're already in the GUI thread.
  // I'm not sure if we can make these requests concurrent.
  *state = true;
  setEmotes(parseEmotesOrWhatever(res));
});
```

src/util/Helpers.cpp
```diff
@@ -316,6 +324,50 @@ QLocale getSystemLocale()
     return QLocale::system();
 }
 
+void writeProviderEmotesCache(const QString &id, const QString &provider,
+                              const QByteArray &bytes)
+{
+    std::ignore = QtConcurrent::run([bytes, id, provider]() {
```
Nerixyz, line 330: Please use [`QThreadPool::globalInstance()->start(callable)`](https://doc.qt.io/qt-6/qthreadpool.html#start-1), because we don't need the result of this.

src/util/Helpers.cpp
```diff
@@ -316,6 +324,50 @@ QLocale getSystemLocale()
...
+        QFile responseCache(getApp()->getPaths().cacheDirectory() + "/" + id +
+                            "." + provider);
+
+        if (responseCache.open(QIODevice::WriteOnly))
+        {
+            qCDebug(chatterinoCache)
+                << "Saved json response " << id << "." << provider;
+            responseCache.write(bytes);
```
Nerixyz, line 338: I think we should compress the data. Using the free function [`qCompress`](https://doc.qt.io/qt-6/qbytearray.html#qCompress) here (and `qUncompress` on the other side) caused the cached 7TV response of my channel to be just 5% of the original size.

Nerixyz (commented):
> but would be interested to know if you'd start seeing emotes in messages pulled from recent messages on startup

Recent messages and the emote providers are racing. If we load cached emotes, we should be more likely to have loaded emotes when the recent messages response comes back. I couldn't find an issue, but there was some talk about saving the user-id in the window layout such that a channel will know its user-id on startup. This way we could load emotes even earlier, but should delay loading the message history.

src/providers/bttv/BttvEmotes.cpp
occluder: I don't have a problem doing it this way, but doesn't this feel slightly over-engineered? I'm not as familiar with the codebase as you are, but this looks like creating a race condition and solving it for tiny perf gain

src/util/Helpers.cpp
```diff
@@ -316,6 +322,50 @@ QLocale getSystemLocale()
...
+        QFile responseCache(getApp()->getPaths().cacheDirectory() + "/" + id +
+                            "." + provider);
+
+        if (responseCache.open(QIODevice::WriteOnly))
+        {
+            qCDebug(chatterinoCache)
+                << "Saved json response " << id << "." << provider;
+            responseCache.write(qCompress(bytes));
+        }
+    });
```
Mm2PL, line 338: is writing to a file really so slow that this needs to be done on a separate thread?

Wissididom: In my opinion, yes. For example I have an external 10 TB HDD, which if it hasn't been used a while needs time to be responsive again. If the drive, the file is located on, is being used a lot anyways I don't see it as a big problem, tho.

src/providers/bttv/BttvEmotes.cpp
Nerixyz: I don't mind doing this synchronously too much. My only worry is that this could slow us down if we save the user-ID in the window layout and immediately load emotes on startup. I haven't tried or measured, so I don't know if this is even noticeable.

pajlada (approved):

src/util/Helpers.cpp
pajlada: nymn's 7TV channel emotes, even compressed, is 589KB

some of it has to do with how large the network payload is since we're not only storing raw emote id/url stuff, but still kinda large

pajlada:

PR #6126, opened 2025-04-01
feat: setting to change emote/badge thumbnail size

4rneee:
This adds a setting to change the size of emote/badge thumbnails when hovering over them.

The default behavior is unchanged.


src/widgets/helper/ChannelView.cpp
```diff
@@ -2027,12 +2027,15 @@ void ChannelView::mouseMoveEvent(QMouseEvent *event)
...
-                    element->getTooltip(),
-                });
+                auto scale = getSettings()->emoteTooltipScale.getValue();
+                TooltipEntry entry = {
+                    .image =
+                        showThumbnail
+                            ? emoteElement->getEmote()->images.getImage(scale)
+                            : nullptr,
+                    .text = element->getTooltip()};
+                entry.scaleImg(scale);
```
pajlada, line 2037: I'd prefer if the `scaleImg` function is removed.

You could add a static function to `TooltipEntry` that creates a scaled version, e.g. `auto entry = TooltipEntry::scaled(image, text, scale);` instead, or set `customWidth`/`customHeight` yourself right here

pajlada (changes requested):

pajlada:
> I don't know if it matters much but is there a way to hide the setting for the emote thumbnail size if ThumbnailPreviewMode is set to "Don't show"? A simple if statement did not do the trick.
There's no simple way to do this, no.
In some of the non-general-pages we gray the setting out (i.e. disable it, so it also can't be interacted with) if it's not relevant but it's not something we support generically.
I'd say don't worry about it

4rneee:
Any idea why this test fails all of a sudden?
The `OnceFlag.waitFor` test seems to fail because it waited for 38ms longer than expected. I guess it's nondeterminism. Is there a way for me to trigger a rerun?

pajlada:
You could always push an empty commit, but that'll retry the entire CI. I've manually restarted the failing job now

pajlada (approved):

PR #6132, opened 2025-04-06
fix: receive bttv live updates again

iProdigy:
bttv socket server changed at some point to _only_ support TLS 1.3 but we previously maxed at TLS 1.2 so the bttv ws stopped connecting due to a tls handshake error

I've tested that 7tv ws (supports tls 1.0-1.3) still works with this change (out of paranoia I disabled tls 1.0 & 1.1, which are both deprecated)

also updates outdated `broadcast_me` payload: https://betterttv.com/developers/websocket#user-broadcast

### DIFF
--- a/src/providers/liveupdates/BasicPubSubManager.hpp
+++ b/src/providers/liveupdates/BasicPubSubManager.hpp
@@ -337,7 +337,7 @@ private:
     WebsocketContextPtr onTLSInit(const websocketpp::connection_hdl & /*hdl*/)
     {
         WebsocketContextPtr ctx(
-            new boost::asio::ssl::context(boost::asio::ssl::context::tlsv13));
+            new boost::asio::ssl::context(boost::asio::ssl::context::tls_client));
 
         try
         {
--- a/src/providers/liveupdates/BasicPubSubManager.hpp
+++ b/src/providers/liveupdates/BasicPubSubManager.hpp
@@ -342,8 +342,6 @@ private:
         try
         {
             ctx->set_options(boost::asio::ssl::context::default_workarounds |
-                             boost::asio::ssl::context::no_sslv2 |
-                             boost::asio::ssl::context::no_sslv3 |
                              boost::asio::ssl::context::no_tlsv1 |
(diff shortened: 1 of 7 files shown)


src/providers/liveupdates/BasicPubSubManager.hpp
```diff
@@ -337,12 +337,15 @@ class BasicPubSubManager
     WebsocketContextPtr onTLSInit(const websocketpp::connection_hdl & /*hdl*/)
     {
         WebsocketContextPtr ctx(
-            new boost::asio::ssl::context(boost::asio::ssl::context::tlsv12));
+            new boost::asio::ssl::context(boost::asio::ssl::context::tlsv13));
```
Nerixyz, line 340: Should probably be `tls_client` to support both v1.2 and v1.3.

src/providers/liveupdates/BasicPubSubManager.hpp
```diff
@@ -337,12 +337,15 @@ class BasicPubSubManager
...
 
         try
         {
             ctx->set_options(boost::asio::ssl::context::default_workarounds |
                              boost::asio::ssl::context::no_sslv2 |
+                             boost::asio::ssl::context::no_sslv3 |
```
Nerixyz, line 346: No need to disable SSLv2/3 because `tls_client` [implies TLS 1+](https://github.com/boostorg/asio/blob/62864c570afcb7234442168ecacd53af701d7d9e/include/boost/asio/ssl/impl/context.ipp#L334).

src/providers/liveupdates/BasicPubSubManager.hpp
pajlada: If possible this would be preferred yeah. We know the issues of TLS v1.3-only support :disappointed:

iProdigy: confirmed `tls_client` also works for bttv & 7tv, thanks!

pajlada (approved):
Good catch, thank you!

PR #6137, opened 2025-04-07
Feat: Themes allow transparency on chat windows background

apa420:
This allows you to create themes with different opacity settings. This only applies for the main window, popups and not any sub windows such as settings, accounts, emotes etc.

Example theme:
[Transparent.json](https://github.com/user-attachments/files/19638126/Transparent.json)
image

### DIFF
--- a/src/widgets/Window.cpp
+++ b/src/widgets/Window.cpp
@@ -71,6 +71,7 @@ Window::Window(WindowType type, QWidget *parent)
             this->onAccountSelected();
         }));
     this->onAccountSelected();
+    this->setAttribute(Qt::WA_TranslucentBackground);
 
     if (type == WindowType::Main)
     {
--- a/src/widgets/splits/SplitContainer.cpp
+++ b/src/widgets/splits/SplitContainer.cpp
@@ -643,14 +643,7 @@ void SplitContainer::paintEvent(QPaintEvent * /*event*/)
     }
     else
     {
-        if (getApp()->getThemes()->isLightTheme())
-        {
-            painter.fillRect(rect(), QColor("#999"));
-        }
-        else
-        {
-            painter.fillRect(rect(), QColor("#555"));
-        }
+        painter.fillRect(rect(), this->theme->splits.background);
     }
 
     for (DropRect &dropRect : this->dropRects_)


src/widgets/Window.cpp
```diff
@@ -71,6 +71,7 @@ Window::Window(WindowType type, QWidget *parent)
             this->onAccountSelected();
         }));
     this->onAccountSelected();
+    this->setAttribute(Qt::WA_TranslucentBackground);
```
Nerixyz, line 74: Could we only set this if we know the color we will paint with is actually transparent? I suppose that would be `getTheme()->windows->background`. Then it should probably be set in `BaseWindow` instead, since it paints with this color.

src/widgets/splits/SplitContainer.cpp
```diff
@@ -643,14 +643,7 @@ void SplitContainer::paintEvent(QPaintEvent * /*event*/)
...
     {
-        if (getApp()->getThemes()->isLightTheme())
-        {
-            painter.fillRect(rect(), QColor("#999"));
-        }
-        else
-        {
-            painter.fillRect(rect(), QColor("#555"));
-        }
+        painter.fillRect(rect(), this->theme->splits.background);
```
Nerixyz, line 646: If I remember correctly, this is the color that's shown between splits. At least in the dark theme, it's the same as `tabs.dividerLine` (not in light themes, though).

apa420:
I've noticed an issue with the border between splits being gone. Seems like some form of fix for that is required.

src/widgets/Window.cpp
apa420: I don't think the transparency should appear on all windows. Only the main one is preferable (`BaseWindow` doesn't know what a `Window::Main` is), I'll make the window check for the theme being transparent before setting the flag.

apa420:
> I've noticed an issue with the border between splits being gone. Seems like some form of fix for that is required.

Fixed

Nerixyz:
This doesn't work on Windows. You'd need to handle NCPAINT correctly on both Windows 10 and Windows 11. Note that NCPAINT also paints the window shadow, so skipping NCPAINT won't work (also on Windows 10 the non-client area is bigger, and you get epic white or Windows 7 edges).

apa420:
I'm not in a position to test this on windows so I'm simply gonna make this pr for linux only.

Nerixyz:
Okay, I got a hacky solution that works on Windows 11:

```cpp
DWM_BLURBEHIND bb = {0};
bb.dwFlags = DWM_BB_ENABLE;
bb.fEnable = TRUE;
bb.hRgnBlur = nullptr;
::DwmEnableBlurBehindWindow(hwnd, &bb);
```

This doesn't actually do a blur behind the window, but only makes it transparent. I don't know why exactly, but I'm pretty sure it's one of the legacy DWM functions. Neither do I know if this is really CPU intensive (shirley Microsoft would have a check for a 0px blur, right?).

pajlada:
Couldn't figure out an easy way to make it work on Windows using the blur magic, so I'm ok if this is a Linux-only feature for now

pajlada (approved):

PR #6138, opened 2025-04-08
Fix: 6133 drag and drop split loss

### LINKED ISSUE #6133: Split drag-and-drop failure will cause split to be lost
### Checklist

## Issue
If one drag the split and the dropping fails, the split is lost. The expected behavior is the split goes back to its old place

## Reproduction
1. Make sure you have at least 2 tabs in Chatterino 
2. On tab 1, make sure you have a split
3. Drag the split by dragging the header in any of following areas
  - drop it on tab 2 on the tab bar (your cursor changes to 🚫 )
  - drop it on empty area of the tab bar (your cursor changes to 🚫 )
  - Not in recording, more obscure cases: dropping outside Chatterino: may or may not work correctly, regardless whether mouse cursor is 🚫 or dragging hand
     - dragging into empty area of KDE desktop (cursor: dragging hand) -> lost
     - dragging into Dolphin (KDE's file manager) (cursor: 🚫) -> lost
     - dragging into Konsole (KDE's terminal) (cursor: 🚫) -> work correctly 
     - dragging into Google Chrome, in various area, regardless of cursor being 🚫 or dragging hand -> work correctly 
4. Observe that the split is lost, forever.

## Notes:
- Dropping on + button will create new tab and move split there as expected
- Strangely, dropping on settings or account button also behave same as dropping on + button

### Screenshots

Note: because I'm capturing chatterino window only, in the middle part of clip,  where I add forsen split to reproduce again, the popup wasn't visible (that's not relevant to the bug tho)

https://github.com/user-attachments/assets/69fd049e-15f6-4cab-9383-51f519894a06

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3481c9336) built with Qt 6.9.0 Running on EndeavourOS, kernel: 6.14.0-rc7-3-cachyos-rc

- DE: KDE Plasma, Wayland

rod41732:
This fixes #6133

Instead of checking drag->exec() which contains limited information and behave inconsistently on various platforms (e.g. broken on KDE, but works fine on sway).

I now added the state for explicitly checking whether split are dropped. 

Tested and seems working as intended now
- no more split lost when dragging into tab bar, desktop, dolphin
- existing cases that worked still working correctly.

[Screencast_20250408_175040.webm](https://github.com/user-attachments/assets/5153db07-2903-4a34-9e31-6f941c7db179)

### DIFF
--- a/src/widgets/helper/NotebookButton.cpp
+++ b/src/widgets/helper/NotebookButton.cpp
@@ -199,6 +199,8 @@ void NotebookButton::dropEvent(QDropEvent *event)
         return;
     }
 
+    draggedSplit->markAsDropped();
+
     event->acceptProposedAction();
 
     auto *page = new SplitContainer(notebook);
--- a/src/widgets/splits/Split.cpp
+++ b/src/widgets/splits/Split.cpp
@@ -1685,11 +1685,6 @@ void Split::drag()
     this->dragState_ = SplitDragState::Dragging;
 
     drag->exec(Qt::MoveAction);
-    qCWarning(chatterinoWidget)
-        << "DOGE: drag state after"
-        << static_cast<std::underlying_type_t<SplitDragState>>(
-               this->dragState_);
-    ;
 
     // drag->exec is a blocking action
     if (this->dragState_ != SplitDragState::Dropped)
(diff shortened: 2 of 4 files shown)


pajlada:
Closing in favour of #6147 - thank you for looking into it, it helped me know where to look!

PR #6076, opened 2025-03-14
feat(plugins): add `c2.WebSocket`

Nerixyz:
Adds a small WebSocket client. Connections must use TLS.

<details><summary>Example</summary>

Go in a Twitch channel and run `/mirror-chat`. All chat messages will now appear twice.

```lua
---@type c2.WebSocket|nil
local ws = nil
c2.register_command("/mirror-chat", function(ctx)
    if ws then
        ws:close()
    end
    local proto = "wss"
    if #ctx.words >= 2 then
        proto = ctx.words[2]
    end
    ws = c2.WebSocket.new(proto .. "://irc-ws.chat.twitch.tv")
    local chan = ctx.channel
    ws.on_text = function(text)
        -- Don't care about multiple messages in one
        text = text:sub(1, -3)
        if #text > 4 and text:sub(1, 4) == 'PING' then
            ws:send_text("PONG" .. text:sub(5))
            return
        end

        local sender_end = text:find(' ', 1, true)
        if sender_end == nil then
            return
        end
        local sender = text:sub(2, sender_end - 1)

        local command_end = text:find(' ', sender_end + 1, true)
        if command_end == nil then
            return
        end
        local command = text:sub(sender_end + 1, command_end - 1)
        if command ~= "PRIVMSG" then
            return
        end

        local name_end = sender:find('!', 1, true)
        if name_end == nil then
            return
        end
        local sender_name = sender:sub(1, name_end - 1)

        local msg_start = text:find(':', command_end + 1, true)
        if msg_start == nil then
            return
        end
        chan:add_system_message(sender_name .. ": " .. text:sub(msg_start + 1))
    end
    ws:send_text("NICK justinfan123456")
    ws:send_text("JOIN #" .. ctx.channel:get_name())
    chan:add_system_message("Mirroring messages")
end)
```
</details> 

Requests: https://github.com/Chatterino/chatterino2/discussions/4931, https://github.com/Chatterino/chatterino2/discussions/4999#discussioncomment-7747236

PR #6141, opened 2025-04-08
fix(websockets): add query to URL

Nerixyz:
I missed this in #6076. If the URL has a query, we now append it.

Depends on https://github.com/Chatterino/twitch-pubsub-server-test/pull/126.


pajlada (approved):
I'll just need to fix the changelog entry when merging this in, but otherwise looks good

PR #6145, opened 2025-04-12
fix: don't clear message when closing reply

### LINKED ISSUE #6144: The close button in message replies shouldn't delete history state
### Checklist

### Describe your issue

When you <kbd>SHIFT</kbd>+<kbd>RMB</kbd> a user's name in a message to reply, type out your reply, and then realize that you've replied to the wrong message, sometimes you may click the ⮾ button, instead of/before <kbd>SHIFT</kbd>+<kbd>RMB</kbd>ing again to change the target. Currently, once the ⮾ button's pressed, the message you're typing's cleared, and all history for the split input box's gone, so you can't <kbd>CTRL</kbd>+<kbd>Z</kbd> to get the text back.

### Implementation details/somewhat unrelated
I'm not even sure if clearing the split input box's the right call here, just closing the panel, and deleting the `@reply` text should be enough, but this isn't a big deal. I'd also like an <kbd>ESC</kbd> bind on the split input box to close this reply panel by default.

### Screenshots

_No response_

### OS and Chatterino Version

a70acbfda

pajlada:
Fixes #6144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Bugfix: Fixed scrolling now working on inputs in the settings. (#6128)
 - Bugfix: Make reply-cancel button less coarse-grained. (#6106)
 - Bugfix: Fixed missing BetterTTV live updates of emotes. (#6132)
+- Bugfix: Fixed message contents being cleared when using the close button for replies. (#6145)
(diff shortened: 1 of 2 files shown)


pajlada:
image

PR #6147, opened 2025-04-12
fix: split drag & drop on KDE Plasma

pajlada:
### Tests

1. Drag a split & drop it next to another split in the same tab. The split should be in the tab you had selected in the position you dropped it at.
2. Drag a split onto your desktop or onto Discord or something. The Split should go back to its original tab
3. Drag a split onto another tab. The Split should be moved to the tab you dropped it onto, and not exist in the old tab.
4. Drag a split onto the plus button next to your tabs. The split should be moved to a new tab
5. Drag a split onto another Chatterino instance. The split should return back to its original tab
6. Drag a split onto a Chatterino popup window. The split should move to that popup as expected.

Supersedes #6138

I have tested this on Sway & Plasma. Leppunen & @rod41732 have tested this in Plasma

Would be great to have a windows user test this to ensure I haven't broken drag & drop somehow.


pajlada:
I've now tested this on Windows.
Windows had similar issues to KDE where you could "delete" splits by dropping the split in certain locations, namely in the explorer window sidebar at the bottom, or onto the recycle bin (lol). - this PR "fixes" this too

pajlada:
image

PR #6148, opened 2025-04-12
refactor: split

pajlada:
just some mini refactors in split, not too disruptive

- **refactor(Split): remove some unused includes**
- **refactor(Split): remove unused `select_randomly` function**
- **refactor(Split): remove unused includes**
- **refactor(Split): use auto when possible**
- **refactor(Split): remove unneccessary static_cast**
- **refactor(Split): const ref function params**
- **refactor(Split): simplify signal connects**
- **refactor(Split): const ref argument parsing in hotkeys**
- **refactor(Split): don't return "const" from getFilters**


pajlada:
image

PR #6155, opened 2025-04-16
feat: update emojis to Unicode 16.0

Nerixyz:
(AI generated text on why emojis are essential). https://github.com/iamcal/emoji-data is pretty dead (though I hope by writing this, I can magically revive the project). Although there aren't frequent updates to Unicode standards, there hasn't been an update in the repo for over a year now, while Unicode 16.0 has been out for half a year. I opened https://github.com/iamcal/emoji-data/pull/252 to update to 16.0, and I think it's time that we update as well. It's only Facebook that hasn't updated yet, but I don't think many people use that emoji set. I got the `emoji.json` from that PR.

Requires `https://pajbot.com/static/emoji-v2` to be updated.

---

In the future, we could probably trim that JSON down by removing fields we don't use.


pajlada (approved):

pajlada:
https://pajbot.com/static/emoji-v2 has been updated for Emoji 16

Issue #6157, opened 2025-04-18
Settings are partially reset randomly since 2.5.3


th-bu:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Since updating to 2.5.3 Chatterino will seemingly at random remove whole sections from the settings.json.

In particular (when it's doing this) it's removing the following sections/values:

* behaviour
  * mentionUsersWithComma
  * autocompletion
* hotkeys
  * addedDefaults.toggle live tabs only
* misc
  * twitch
  * showPronouns
* accounts *(whole section)*
* appearance *(whole section)*
* streamerMode *(whole section)*
* ui *(whole section)*
* whispers *(whole section)*
* filtering *(whole section)*
* ignore *(whole section)*
* highlighting *(whole section)*
* notifications *(whole section)*

It is also adding the following:

* hotkeys
  * delete

The important thing is that it produces a valid json file, so no missing braces at the end.

Before updating to 2.5.3 Chatterino was running flawlessly for over a year in that regard.

I also tried reverting back to 2.5.2 and did not have any problems during the 10 days I used it. So it must be some change in 2.5.3.

Before you ask, yes, I let Chatterino be shut down by Windows. But as I said it was running without any problems with the same workflow for over a year now.

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 11 Version 23H2, kernel: 10.0.22631

th-bu:
I just installed the nightly that contains the change and will keep using it for the next 2 weeks or so (maybe even until the 2.5.4 final comes out). I will report here if I still get the same behavior as before.

pajlada:
Thank you!

th-bu:
7 days in and it currently working without any problems. Not a single reset during that time period.

th-bu:
I'd say we can close this issue. I've had no problems since upgrading to the nightly version.

PR #6159, opened 2025-04-18
fix: handle `aboutToQuit()`

Nerixyz:
In an attempt to solve https://github.com/Chatterino/chatterino2/issues/6157, I added a handler for [`aboutToQuit`](https://doc.qt.io/qt-6/qcoreapplication.html#aboutToQuit) - specifically because this is called when the application is shut down by Windows ([source](https://github.com/qt/qtbase/blob/53ff8897c5c8bc6175cf94ed24e2d2c5fa17365b/src/plugins/platforms/windows/qwindowscontext.cpp#L1312-L1315)).

If we do this, we shouldn't save afterward - `aboutToQuit` will be called (immediately) before `Application::run` returns. I added `disableSave` to try to avoid saving after we exit.


pajlada (approved):

pajlada:
Seems like the `aboutToQuit` signal is the correct place to do this type of thing, good catch

https://doc.qt.io/qt-6/qapplication.html#exec

PR #6160, opened 2025-04-19
add Import/Export for Nicknames

T0biii:
Hey we wanted to exchange nicknames among friends, so we came up with the idea of ​​adding import and export buttons to the Nicknames view.
This is my first time with C++ code, so feedback is very welcome.

Tested on Ubuntu:
image
image
image
image

### DIFF
--- a/src/widgets/settingspages/NicknamesPage.cpp
+++ b/src/widgets/settingspages/NicknamesPage.cpp
@@ -7,9 +7,13 @@
 #include "util/LayoutCreator.hpp"
 #include "widgets/helper/EditableModelView.hpp"
 
+#include <QFileDialog>
 #include <QHeaderView>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QMessageBox>
 #include <QTableView>
-
 namespace chatterino {
 
 NicknamesPage::NicknamesPage()
@@ -44,6 +48,18 @@ NicknamesPage::NicknamesPage()
             Nickname{"Username", "Nickname", false, false});
     });
 
+    // Add Import and Export buttons to the EditableModelView
+    auto *importButton = new QPushButton("Import");
+    auto *exportButton = new QPushButton("Export");
+
+    QObject::connect(importButton, &QPushButton::clicked, this,
+                     &NicknamesPage::importNicknames);
+    QObject::connect(exportButton, &QPushButton::clicked, this,
+                     &NicknamesPage::exportNicknames);
+
+    view->addCustomButton(importButton);
+    view->addCustomButton(exportButton);
+
     QTimer::singleShot(1, [view] {
         view->getTableView()->resizeColumnsToContents();
         view->getTableView()->setColumnWidth(0, 200);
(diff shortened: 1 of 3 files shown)


resources/contributors.txt
```diff
@@ -85,6 +85,7 @@ MiguVT | https://github.com/MiguVT | :/avatars/miguvt.png
 jacob-thompson | https://github.com/jacob-thompson |
 Jujufrost | https://github.com/Jujufrost |
 occluder | https://github.com/occluder | :/avatars/occluder.png
+T0biii | https://github.com/T0biii |
```
pajlada, line 88: Leave this change for a follow-up PR

pajlada (changes requested):
I don't mind this feature, but if the goal is to share nicknames between friends I'd prefer if this used something like your system clipboard instead.

So the export button would export the nickname json to your clipboard, and the import button would read that json blob from your clipboard

The label of the buttons should probably be updated to reflect that change as well. "Import from clipboard" and "Export to clipboard" if it fits.
If it doesn't fit where they are now, you can move it to below the nicknames instead.

T0biii:
I think it's easier to share a file depending on how many nicknames are present, e.g. Discord has a character limit of 2000. 4000 for Nitro Users. 
But if you think clipboard would be better i can try to update the pr to clipboard

pajlada:
I did not consider character limits, let me give it a think

pajlada:
Sorry for the delay!

I think files are fine, however I think I'd like to be a bit forward-thinking with the import & export format.
Currently you're exporting the nicknames as-is. I'd prefer if we provide some metadata alongside the export that can be used in the import.

I think something like this would be good:
```json5
{
    "type": "nickname_export",
    "version": "1",
    "comment": "Exported from Version::instance().buildString() at CURRENT TIME",
    "data": [
        // the nicknames in the same format you were already using
    ]
}
```


I'm not convinced of the place for the "exportNicknames" and "importNicknames" functions, so I might want to move those around a bit but that's easy to do as a last step.

src/widgets/settingspages/NicknamesPage.cpp
```diff
@@ -57,4 +75,119 @@ bool NicknamesPage::filterElements(const QString &query)
     return this->view_->filterSearchResults(query, fields);
 }
 
+void NicknamesPage::importNicknames()
+{
+    QString filePath = QFileDialog::getOpenFileName(
+        this, tr("Import Nicknames"), "", tr("JSON Files (*.json)"));
```
Nerixyz, line 81: We don't use Qt's translation feature.

T0biii: got it from here:
https://github.com/Chatterino/chatterino2/blob/f3559280786dbabc8cebe1951a525580d11cb2a5/src/widgets/settingspages/NotificationPage.cpp#L92-L94

Nerixyz: Hmm, we should remove it there as well at some point. If we ever use Qt's translations, we can add it back.

T0biii:
Unfortunately, I don't have time to work on it right now; perhaps I'll try again in the future.

PR #6161, opened 2025-04-19
fix: use qlogging in eventsub lib (through a hoop)

pajlada:
This PR adds a logging proxy to the EventSub lib

It gives the session a warn and debug function it can use to log things, and we can decide how things are logged.
I made a mini qLogging proxy for it, which just forwards the data to categorized Qt logging.
On <Qt6.5 no logging will take place since they don't support qDebug << std::string_view

It's not the most optimized way of logging things but I think it's ok.


pajlada (commented):
image

PR #6165, opened 2025-04-20
fix: emote popup logging messages to "Other" directory

### LINKED ISSUE #6056: emote popup is logging to Others dir
bug

pajlada:
some mini unrelated changes

this fixes the issue in two ways:
1) None channel types no longer log messages
2) Channels without an explicit platform set are not logged

Fixes #6056


pajlada (commented):
image

PR #6175, opened 2025-04-24
chore: re-export Qt's string literals

Nerixyz:
We require Qt 6.4+, which has all the string literals we need. [QString](https://doc.qt.io/qt-6/qstring.html#operator-22-22_s), [QByteArray](https://doc.qt.io/qt-6/qbytearray.html#operator-22-22_ba), [QLatin1StringView](https://doc.qt.io/qt-6/qlatin1stringview.html#operator-22-22_L1-1). The only one missing is for `QStringView` which is coming in 6.10, but we haven't needed it.


pajlada (approved):

pajlada:
good catch!

PR #6181, opened 2025-05-02
fix: don't include blocked users in join/part messages

### LINKED ISSUE #2505: Ignore does not hide join/parts in channel
**What should be added?**
Not sure if this is implemented in any way, but i ignored maybe 10-20 bots that regularly join/part most twitch channels but can still see their joins and leaves. Assuming ignore is just to not see their messages.

**Why should it be added?**
Hides clutter that is not needed to see in chat.

pajlada:
Users who are blocked no longer show up in the "Users joined:" or "Users parted:" messages

Fixes #2505

The biggest chunk of code of this fix comes from adding user logins to the list of blocks. There are some container types that allow for use of multiple keys (basically keeping the same type of logic we use with multiple sets but internally instead), but that seemed a bit out of scope of this change.

### DIFF
--- a/src/common/ChannelChatters.cpp
+++ b/src/common/ChannelChatters.cpp
@@ -1,10 +1,8 @@
 #include "common/ChannelChatters.hpp"
 
-#include "Application.hpp"
 #include "common/Channel.hpp"
-#include "controllers/accounts/AccountController.hpp"
+#include "controllers/ignores/IgnoreController.hpp"
 #include "debug/AssertInGuiThread.hpp"
-#include "messages/Message.hpp"
 #include "messages/MessageBuilder.hpp"
 #include "providers/twitch/TwitchAccount.hpp"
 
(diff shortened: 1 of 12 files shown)


src/providers/twitch/TwitchAccount.cpp
```diff
@@ -138,51 +140,60 @@ void TwitchAccount::loadBlocks()
...
 {
     assertInGuiThread();
     assert(getApp()->isTest());
 
     TwitchUser blockedUser;
     blockedUser.id = userID;
+    blockedUser.name = userLogin;
     this->ignores_.insert(blockedUser);
     this->ignoresUserIds_.insert(blockedUser.id);
+    this->ignoresUserIds_.insert(blockedUser.name);
```
Nerixyz, line 196:
```suggestion
    this->ignoresUserLogins_.insert(blockedUser.name);
```

Nerixyz (changes requested):

src/providers/twitch/TwitchAccount.cpp
pajlada: Fixed in 59b163b91b3c2f81252a7319767bb024e1b1f518

src/common/ChannelChatters.cpp
```diff
@@ -27,8 +31,18 @@ void ChannelChatters::addRecentChatter(const QString &user)
...
-    joinedUsers->append(user);
+    assertInGuiThread();
+
+    if (getApp()
+            ->getAccounts()
+            ->twitch.getCurrent()
+            ->blockedUserLogins()
+            .contains(user))
+    {
+        return;
```
iProdigy, line 42: imo moderator status should override & show join/part, even if user was blocked

iProdigy: Done in https://github.com/Chatterino/chatterino2/pull/6181/commits/f805385655844209bc2516a1f2e6d125774b60db thanks!

pajlada (commented):
image

PR #6188, opened 2025-05-04
feat: run-and-kill script

pajlada:
this is a shell script you can run to automatically run & kill Chatterino at a given path at random intervals

example usage:
`./scripts/run-and-kill.sh ./build-mold/bin/chatterino`


pajlada:
image

PR #6190, opened 2025-05-05
fix(plugins): TSTL `HTTPRequest.create` generation

2547techno:
Currently, the use of `c2.HTTPRequest.create(...)` in typescript results in code that produces the following error after being compiled with TSTL:
```
Failed to evaluate command from plugin HTTP Error: stack index 1, expected number, received table: (bad argument into 'std::shared_ptr<chatterino::lua::api::HTTPRequest>(sol::this_state, enum chatterino::NetworkRequestType, QString)')
```

This is because the generated lua code for this function call is:
```lua
c2.HTTPRequest:create(...)
```
This PR updates the types file so that the generated lua code is:
```lua
c2.HTTPRequest.create(...)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Dev: Updated GoogleTest to v1.17.0. (#6180)
 - Dev: Mini refactor of `TwitchAccount`. (#6182)
 - Dev: Simplified string literals to be a re-export of Qt functions. (#6175)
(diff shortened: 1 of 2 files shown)


pajlada (approved):

Nerixyz:
https://typescripttolua.github.io/docs/advanced/writing-declarations#classes mentions that you'd create an interface of the same name with a function that takes `this` as `void`.

pajlada:
Nerixyz's link seems reasonable - if you notice any issues please follow up with a new PR that changes the static `create` function to match the given example in the link

PR #6213, opened 2025-05-17
fix: don't construct images after we're quitting

pajlada:
i also snuck in a mini Application.cpp refactor, no one will know


pajlada:
image

PR #6215, opened 2025-05-17
refactor(notebook): merge top/bottom and left/right layout

Nerixyz:
Towards https://github.com/Chatterino/chatterino2/discussions/6153.
Should make it easier to implement refactorings to notebook layout (and potentially fix the issues linked in https://github.com/Chatterino/chatterino2/discussions/6153#discussioncomment-12846958).

Top/bottom and left/right layouts are almost the same, except for the direction. Unfortunately, they're not that similar, but we can still put them together and get the same result. The functions are named horizontal for left-to-right/right-to-left layout and vertical for top-to-bottom layout.

The easiest way to test this is to make screenshots and use the "difference" blend-mode in your favorite photo editor (tbf, all I did was diff the branches and insert an `if` where they differed). One small difference is that the custom buttons will show 1dip higher in the bottom layout. I don't think this is a major issue.

I sneaked in some `static_cast`s in place of C-casts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 - Dev: Refactored away some `getApp` usages in `WindowManager`. (#6194)
 - Dev: Simplified string literals to be a re-export of Qt functions. (#6175)
 - Dev: Fixed incorrect lua generation of `c2.HTTPRequest.create` for typescript plugins. (#6190)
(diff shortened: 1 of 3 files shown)


Nerixyz:
I could fix those, but I feel like it's better to do in another PR.

pajlada (approved):

PR #6229, opened 2025-05-23
fix: crash on older versions of Qt

### LINKED ISSUE #6228: Crash when built with Qt 6.6.0
### Checklist

### Describe your issue

```
PS C:\Users\\GitHub\chatterino2> $env:QT_WIN_DEBUG_CONSOLE = 'attach'
PS C:\Users\\GitHub\chatterino2> build/bin/chatterino.exe
PS C:\Users\\GitHub\chatterino2> QObject::moveToThread: Cannot move objects with a parent
QWidget::setLayout: Attempting to set QLayout "" on chatterino::SplitInput "", which already has a layout
QWidget::setLayout: Attempting to set QLayout "" on chatterino::SplitInput "", which already has a layout
QWaitCondition: Destroyed while threads are still waiting
QWaitCondition: Destroyed while threads are still waiting
QWaitCondition: Destroyed while threads are still waiting
```
I've also tried versions 6.6.3, 6.6.2, and 6.6.1. All versions result in an instant crash of the app.

### Screenshots

_No response_

### OS and Chatterino Version

Windows 11, version 24H2 (26100.4188), Chatterino (built from source, branch main)

pajlada:
QStringBuilder is a bit more finicky on older versions of Qt. This "builds" the builders fully into a QString before we attempt to use them.

This crash was not part of the latest stable release, as it was introduced in #6125 

Fixes #6228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unversioned
 
 - Minor: Added user notes to the user info dialog (when clicking a username). (#6122)
-- Minor: Added cached emotes fallback when fetching from a provider fails. (#6125)
+- Minor: Added cached emotes fallback when fetching from a provider fails. (#6125, #6229)
 - Minor: Add an option for the reduced opacity of message history. (#6121)
 - Minor: Make paused chat indicator more visible, and fix its zoom behavior. (#6123)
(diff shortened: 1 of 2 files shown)


pajlada:
image

PR #6233, opened 2025-05-24
Fix eventsub message delete disable

### LINKED ISSUE #6232: Moderator Delete Notifications can't be turned off
### Checklist

### Describe your issue

Just upgraded from a 2.5.1 build to the most recent 2.5.3 nightly build (deed1061). 

"Show deletions of single messages" doesn't impact notifications of moderator deletes, which didn't used to be appear. Is there an option I might be missing to hide them?

### Screenshots

Image

### OS and Chatterino Version

Chatterino Nightly 2.5.3 (commit [deed1061](https://github.com/Chatterino/chatterino2/commit/deed1061)) built on 2025-05-24 with Qt 6.7.1, MSVC 194334808, Crashpad

Mm2PL:
Fixes #6232

### DIFF
--- a/src/providers/twitch/eventsub/Connection.cpp
+++ b/src/providers/twitch/eventsub/Connection.cpp
@@ -188,6 +188,10 @@ void Connection::onChannelModerate(
 
             if constexpr (CanMakeModMessage<Action>)
             {
+                if (getSettings()->hideDeletionActions)
+                {
+                    return;
+                }
                 EventSubMessageBuilder builder(channel, now);
                 builder->loginName = payload.event.moderatorUserLogin.qt();
                 makeModerateMessage(builder, payload.event, action);


src/providers/twitch/eventsub/Connection.cpp
```diff
@@ -188,6 +188,10 @@ void Connection::onChannelModerate(
 
             if constexpr (CanMakeModMessage<Action>)
             {
+                if (getSettings()->hideDeletionActions)
+                {
+                    return;
+                }
```
Nerixyz, line 194: This will hide all moderation actions. At the very least, this should check for a deletion action. However, I think we concluded that we still add the messages but don't show them (like with "hide moderator actions"). Given that it's implemented as "don't add the message" in IRC, I'd be fine with a check & early out here but with a FIXME above.

pajlada: FIXME added in [`e65aa5b` (#6233)](https://github.com/Chatterino/chatterino2/pull/6233/commits/e65aa5ba0ddd43f20c238c91a5355e9236319241)

pajlada (approved):

PR #6238, opened 2025-05-25
feat: implement display name specializations for enums

pajlada:
Currently not implemented anywhere, but I've tested implementing it for a setting and it works as expected.

Example implementation (put in same file/namespace as the enum itself)
```c++
constexpr chatterino::qmagicenum::customize_t qmagicenumDisplayName(
    TabStyle value) noexcept
{
    switch (value)
    {
        case TabStyle::Normal:
            return "Normal (default)";

        default:
            return {};
    }
}
```


pajlada:
image

PR #6250, opened 2025-06-04
feat: add support for automatic streamer mode check from flatpak

ilya-zlobintsev:
Previously, the automatic streamer mode check did not work in flatpak builds, as `pgrep` was executed inside of the sandbox, which only has chatterino's processes and not anything else running on the system.

Now the command will be executed outside of the sandbox through `flatpak-spawn`, the same way as it's done for spawning streamlink. I've tested locally that this works and detects obs running correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Minor: The JSON selector in the upload response can now query arrays using their indices like `foo.0`. (#6193)
 - Minor: Made nicknames searchable in the Settings dialog search bar. (#5886)
 - Minor: Added hotkey Action for opening account selector. (#6192)
-- Minor: Automatic streamer mode detection now works from Flatpak. (#6249)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Minor: The JSON selector in the upload response can now query arrays using their indices like `foo.0`. (#6193)
 - Minor: Made nicknames searchable in the Settings dialog search bar. (#5886)
 - Minor: Added hotkey Action for opening account selector. (#6192)
+- Minor: Automatic streamer mode detection now works from Flatpak. (#6250)
```
Nerixyz:
```suggestion
- Bugfix: Automatic streamer mode detection now works from Flatpak. (#6250)
```

pajlada (approved):

Issue #5944, opened 2025-02-09
Incorrect word wrapping when breaking a word


Nerixyz:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

On some fonts, we don't do word wrapping correctly when a single word needs to be split. Consider the following example:

- Set font to `Arial`
- (use any font engine)
- Generate a message with a lot of numbers, for example:
    ```
    /timeout 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
    ```
- Resize the window to break the word

| Incorrect | (more) Correct |
|---|---|
| ![Image](https://github.com/user-attachments/assets/24be30b8-57d6-4c7d-92c5-b8a7ee65b6fc) | ![Image](https://github.com/user-attachments/assets/b68a7ad3-492e-4da6-bafc-3ea35995780c) |

We're using `horizontalAdvance` on every codepoint in the text. This doesn't always return the actual advancement in the word, because a character can have `left/rightBearing`.

My "(more) correct" version is from https://github.com/Nerixyz/chatterino2/commit/1ead4bbd8399b2877d6ee2b353c0197fe20e6b4d where I used the private API to do the advancing. There might be a way of doing this through `left/rightBearing`.

### OS and Chatterino Version

Chatterino 2.5.2 DEBUG (commit 1ead4bbd8 modified) built with Qt 6.9.0, MSVC 194234436 Running on Windows 11 Version 24H2, kernel: 10.0.26100

Mm2PL:
Can replicate with `FiraCode Nerd Font @ 11pt.`

PR #6265, opened 2025-06-13
feat: add optional improved text wrapping

Nerixyz:
It's a heartbreaking story, but improving the text wrapping while remaining fairly efficient requires private Qt APIs (https://github.com/Chatterino/chatterino2/issues/5944). Specifically, I ran into this because Twitch now allows Arabic display names. Those can get longer (e.g. [`\uFDFD`](https://symbl.cc/en/FDFD)). `horizontalAdvance`ing on single characters there will yield incorrect widths. Thus, when the channel view is narrow enough for the username to require wrapping, the message text would appear in the wrong position ("Noto Sans" font).

In the linked issue, I already experimented with using `QStackTextEngine` to do the wrapping. Since it's a private API, I put it behind a configuration option. On Ubuntu we already build Qt, so we control the exact Qt version and can use private APIs.


pajlada (changes requested):
I'm fine with allowing users enabling this, but I don't want this enabled by default anywhere (even where we're comfortable with it).

If you change it to just the compile flag, I'd be OK merging this.

CMakeLists.txt
```diff
@@ -120,6 +121,11 @@ endif()
 
 include(${CMAKE_CURRENT_LIST_DIR}/cmake/GIT.cmake)
 
+if (CHATTERINO_ALLOW_PRIVATE_QT_API)
+    # For qtextengine_p.h
+    set(_additional_qt_components GuiPrivate)
+endif()
```
pajlada: What change does this make?

Nerixyz: I didn't remove this in an earlier commit.

pajlada (approved):

PR #6267, opened 2025-06-14
refactor: Make Account & Settings buttons SVGs

pajlada:
instead of rendering the paths directly in Chatterino, they're now SVGs

This has the following consequences:
1. The color does _not_ respect your custom theme. For most users, this
   will not be a problem/concern.
2. You can no longer drag a split onto those buttons to create a new tab


pajlada:
image

PR #6268, opened 2025-06-14
feat(SvgButton): Allow custom color

pajlada:
Not a clean solution, but it works & given that we cache the output it feels ok

### DIFF
--- a/src/widgets/buttons/SvgButton.cpp
+++ b/src/widgets/buttons/SvgButton.cpp
@@ -77,8 +77,14 @@ void SvgButton::paintContent(QPainter &painter)
     if (this->color_.has_value())
     {
         painter.save();
+
+        // Set the composition mode so that the upcoming color fill only applies the color
+        // on top of the pre-existing SVG contents
+        //
+        // More info on how the composition modes work can be found here: https://doc.qt.io/qt-6/qpainter.html#CompositionMode-enum
         painter.setCompositionMode(QPainter::CompositionMode_SourceIn);
         painter.fillRect(bounds, *this->color_);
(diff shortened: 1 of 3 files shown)


src/widgets/buttons/SvgButton.cpp
```diff
@@ -65,7 +71,16 @@ void SvgButton::paintContent(QPainter &painter)
...
-    this->svg_->render(&painter, {topLeft, contentSize});
+    auto bounds = QRectF{topLeft, contentSize};
+    this->svg_->render(&painter, bounds);
+
+    if (this->color_.has_value())
+    {
+        painter.save();
+        painter.setCompositionMode(QPainter::CompositionMode_SourceIn);
+        painter.fillRect(bounds, *this->color_);
+        painter.restore();
```
Nerixyz: There should be a comment here explaining what this does, because it's not entirely obvious.

Nerixyz (commented):
Seems like this is the best hacky way.

We should ask Qt to provide something like a default fill/stroke color. It would be really helpful for icons in general, I think.

src/widgets/buttons/SvgButton.cpp
pajlada: added a comment about it in [`59135a1` (#6268)](https://github.com/Chatterino/chatterino2/pull/6268/commits/59135a15c21379f91f84bd4b2198df06c23eee75) and linked to https://doc.qt.io/qt-6/qpainter.html#CompositionMode-enum which explains it in more detail

pajlada:
image

PR #6273, opened 2025-06-15
fix: don't autocomplete commands mid-sentence

pajlada:
https://github.com/Chatterino/chatterino2/discussions/6288


pajlada:
image

PR #6281, opened 2025-06-17
feat: optionally parallelize `clang-format-all.sh`

pajlada:
this can be enabled by setting the FAST environment variable to 1


pajlada:
image

PR #6282, opened 2025-06-17
chore: slight restructuring of `check-clang-tidy.sh`

pajlada:
default now single thread and not using parallel unless FAST=1 is
specified


pajlada:
image

Issue #6283, opened 2025-06-18
Import image uploader settings from clipboard


Nerixyz:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

As a user of the image uploader, I want to be able to easily import settings without having to type them out. 

[ShareX](https://getsharex.com/docs/custom-uploader#sxcu-file) has something like this, where users can import a [`.sxcu`](https://getsharex.com/docs/custom-uploader#sxcu-file) file which contains the definition for an uploader. My idea was to define a small JSON schema that Chatterino can read from the clipboard. By pressing a button in the settings, Chatterino would import the settings. Example JSON:

```json
{
  "version": "1.0.0",
  "request_url": "https://foo.bar",
  "form_field": "file",
  "extra_headers": "Authorization: Foo ; X-Header: 1",
  "image_link": "{result.link}",
  "deletion_link": "{result.delete}",
}
```

This will have more value once image uploaders show this to users.
The expected flow would be:
- User sees new cool image uploader
- Image uploader shows a button for the user to copy the Chatterino JSON
- User copies the JSON
- User clicks "Import from Clipboard" in Chatterino
- Image uploader works as expected

... as opposed to the user manually typing or manually copying each field, which is more error-prone, especially for non-technical users.

I think this is a good first issue as it's fairly self-contained. The settings are defined in the [ExternalToolsPage](https://github.com/Chatterino/chatterino2/blob/da96528b5511226c24562c9a3d77cab56acaad4f/src/widgets/settingspages/ExternalToolsPage.cpp#L104-L138). [QJsonObject](https://doc.qt.io/qt-6/qjsonobject.html) and friends can be used to import JSON.

### OS and Chatterino Version

\-

PR #6284, opened 2025-06-19
Enhancement: add image uploader settings import/export functionality

knobby-tires:
Adds import/export functionality for image uploader settings as requested in #6283.

### DIFF
--- a/src/widgets/settingspages/ExternalToolsPage.cpp
+++ b/src/widgets/settingspages/ExternalToolsPage.cpp
@@ -5,9 +5,16 @@
 #include "util/StreamLink.hpp"
 #include "widgets/settingspages/SettingWidget.hpp"
 
+#include <QApplication>
+#include <QClipboard>
 #include <QFormLayout>
 #include <QGroupBox>
+#include <QHBoxLayout>
+#include <QJsonDocument>
+#include <QJsonObject>
 #include <QLabel>
+#include <QMessageBox>
+#include <QPushButton>
 
 namespace chatterino {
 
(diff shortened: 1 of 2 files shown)


pajlada:
Hi!
At a quick glance through the code, things look good. I'll give it a proper review & test this weekend.
Thanks for your contribution!

knobby-tires:
Awesome! I am leaving town this weekend and will have very limited access to a computer. I am happy to make necessary adjustments when/if I can. Just want to make sure you know I am not abandoning the PR lol.

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -137,7 +144,191 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
             ->addTo(layout, form);
     }
 
+    {
```
pajlada: I'd prefer if the image uploader export/import buttons are part of the "Image Uploader" section

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -137,7 +144,191 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
...
+        });
+        buttonLayout->addWidget(importButton);
+
+        buttonLayout->addStretch();
+        layout.addLayout(buttonLayout);
+    }
+
     layout.addStretch();
 }
+void ExternalToolsPage::exportSettings()
```
pajlada: Could you move this to `ImageUploader.hpp/cpp` as part of the `detail` namespace instead?
That way you can write a few unit tests for this to ensure we handle certain setting combinations ok.
What I'd like to test is:
 - What does it look like when the user does not have any settings set at all?
 - What does it look like if the user has never set the deletion link?

I imagine the export function to look something like thisinside `chatterino::imageuploader::detail`:
`QJsonObject exportSettings(Settings &s);`

Feel feel to move the `importSettings` function along, and not write any unit tests for it. It's a bit tricky given that it expects user input.

If you're having a hard time writing the unit tests yourself, I'm happy to help out there.

pajlada (changes requested):
As a first pass - will do another more in-depth review once it's more testable

When I was reviewing this I was noticing the text fields don't update when the setting updates from external sources - I'll be fixing that in a separate PR.

knobby-tires:
Hi @pajlada, I believe I addressed most of what you asked :

- I moved import/export functions to `ImageUploader.hpp/cpp` in the `chatterino::imageuploader::detail` namespace
- I changed `exportSettings` to take `Settings &s` parameter as suggested: `QJsonObject exportSettings(Settings &s);`
- I moved all related functions (`importSettings`, `validateImportJson`, `parseAndApplyHeaders`) to the new file
- Updated `ExternalToolsPage` to use the new functions
- Added the new files to `CMakeLists.txt`
- Confirmed functionality still works correctly (export/import both working)


Next I'll work on writing unit tests for the `exportSettings` function next 

Thanks for the detailed feedback, if I have misunderstood anything up until now please let me know. I am really trying to learn and gain real experience here.

src/util/ImageUploader.cpp
```diff
@@ -0,0 +1,118 @@
...
+        }
+        if (!headersObj.isEmpty()) {
+            settingsObj["Headers"] = headersObj;
+        }
+    }
+
+    return settingsObj;
+}
+
+bool importSettings(const QJsonObject &settingsObj, Settings &s)
```
pajlada, line 64: Question:
This function would currently return true if the user _only_ has "URL" set.
Is the constraint that only one of the fields need to be set, or that all fields need to be set?

src/util/ImageUploader.hpp
```diff
@@ -0,0 +1,20 @@
...
+class QJsonObject;
+
+namespace chatterino {
+
+class Settings;
+
+namespace imageuploader {
+namespace detail {
+
+QJsonObject exportSettings(Settings &s);
```
pajlada: Make this take a `const Settings &s` instead, to clarify that we're not looking to make any modifications to the settings in this function.

src/util/ImageUploader.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+class QJsonObject;
```
pajlada, line 5: You're missing a forward declaration to `QString` here

src/util/ImageUploader.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
pajlada, line 1: It would be nice with a tiny comment for each of the 4 functions you've made. It does not have to be a long comment, but just a tiny insight into what it does (e.g. "Exports settings from Chatterino to the ShareX export JSON format" or something)

src/util/ImageUploader.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+class Settings;
+
+namespace imageuploader {
+namespace detail {
+
+QJsonObject exportSettings(Settings &s);
+bool importSettings(const QJsonObject &settingsObj, Settings &s);
+bool validateImportJson(const QString &clipboardText, QJsonObject &settingsObj);
+void parseAndApplyHeaders(const QJsonObject &headersObj, Settings &s);
```
pajlada: This function is only ever used by the internal pieces - you can make it a source-only function.
Additionally, it would be nice if it only did the parsing, letting the `importSettings` function do the importing.
I imagine the signature would change to something like
```c++
QStringList parseHeaders(const QJsonObject &headersObj);
```

src/widgets/settingspages/ExternalToolsPage.hpp
```diff
@@ -14,8 +14,10 @@ class ExternalToolsPage : public SettingsPage
 
 private:
     void initLayout(GeneralPageView &layout);
+    void exportSettings();
+    void importSettings();
```
pajlada: I'd prefer if these functions were source-only, and accepted a `QWidget *parent` parameter instead of using `this` for the dialog parents. You can still pass `this` to them. They should also be renamed to clarify that they are specifically importing or exporting image uploader settings, not any generic settings.

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -135,9 +143,73 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
...
+    QJsonDocument doc(settingsObj);
+    QApplication::clipboard()->setText(doc.toJson(QJsonDocument::Indented));
+
+    QMessageBox::information(this, "Settings Exported",
+        "Image uploader settings have been copied to clipboard as JSON.");
+}
+
+void ExternalToolsPage::importSettings()
+{
+    QString clipboardText = QApplication::clipboard()->text().trimmed();
```
pajlada: Use our `getClipboardText` function (so `getClipboardText().trimmed()`)

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -135,9 +143,73 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
...
     layout.addStretch();
 }
 
+void ExternalToolsPage::exportSettings()
+{
+    auto &s = *getSettings();
+
+    QJsonObject settingsObj = imageuploader::detail::exportSettings(s);
+    QJsonDocument doc(settingsObj);
+    QApplication::clipboard()->setText(doc.toJson(QJsonDocument::Indented));
```
pajlada: Use our `crossPlatformCopy` function (so `crossPlatformCopy(doc.toJson(QJsonDocument::Indented));`

src/util/ImageUploader.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    bool hasValidSettings = false;
```
pajlada: Are there any "meta" fields that _always_ need to be part of the export that we can check?
e.g. if the json blob always needs `Name` and `Version`, we could check those here and error out early if something is wrong

pajlada (changes requested):
File structure looks good from my perspective.
Some nits with how functions are structured, some documentation, and some questions regarding validation.

src/util/ImageUploader.cpp
knobby-tires: Currently it returns true if any field is set but it should check for all required fields. I'll fix this.

src/util/ImageUploader.cpp
knobby-tires: The export includes Version, Name, RequestMethod, and Body fields. I'm not sure which are standard across different image uploader formats. What fields should I validate?

pajlada: Only the meta fields from ShareX exports, which I think are `Version` and `Name`

knobby-tires:
Hey, I addressed your feedback from yesterday. Let me know what you think

pajlada (approved):
This looks good to me now - I took the liberty of doing a few minor modifications.
Could you take a look at the changes I made and see if they make sense to you @knobby-tires ?

knobby-tires:
@pajlada, Looks good to me! Thanks for the help.

pajlada:
Thank you @knobby-tires! As a first-time contributor, you are now eligible to be listed in the About page in Chatterino.

PR #6285, opened 2025-06-19
build-win add workflow

Veatra:


Veatra:
a

Veatra (commented):
ok

Nerixyz:
As I see, you're trying to compile Chatterino with Qt 5: 2.5.2 was the last release to support Qt 5. Since then, a few things have been added that require Qt 6.

PR #6286, opened 2025-06-21
fix: update lineedit if setting changed externally

pajlada:
This does not fix every single "text box" setting in the settings dialog, only the ones that use `SettingWidget::lineEdit`

I'm OK with that, since that's where I want our development to move anyway.


pajlada (commented):
image

PR #6293, opened 2025-06-22
fix: `SettingWidget::dropdown` not working well on enums with value gaps

pajlada:
slightly messy pr, review commit by commit

In the process of using `SettingWidget::dropdown` in some more places I found this (e.g. `EmojiData::Capability`)


pajlada:
image

PR #6297, opened 2025-06-24
Added a system to generate Lua definitions at build time

Mm2PL:
Did you know I do not like CMake all that much?

### DIFF
--- a/scripts/make_luals_meta.py
+++ b/scripts/make_luals_meta.py
@@ -32,8 +32,9 @@ Non-command lines of comments are written with a space after '---'
 from io import TextIOWrapper
 from pathlib import Path
 from typing import Optional
-import re
 import argparse
+import logging
+import re
 
 BOILERPLATE = """
 ---@meta Chatterino2
@@ -49,7 +50,6 @@ repo_root = Path(__file__).parent.parent
 lua_api_file = repo_root / "src" / "controllers" / "plugins" / "LuaAPI.hpp"
 lua_meta = repo_root / "docs" / "plugin-meta.lua"
 
-print("Writing to", lua_meta.relative_to(repo_root))
 
 
 def strip_line(line: str):
@@ -186,11 +186,11 @@ def finish_class(out, name):
 
 
 def printmsg(path: Path, line: int, message: str):
-    print(f"{path.relative_to(repo_root)}:{line} {message}")
+    logging.debug(f"{path.relative_to(repo_root)}:{line} {message}")
 
 
 def panic(path: Path, line: int, message: str):
-    printmsg(path, line, message)
+    logging.critical(f"{path.relative_to(repo_root)}:{line} {message}")
     exit(1)
 
 
@@ -216,7 +216,7 @@ def write_func(path: Path, line: int, comments: list[str], out: TextIOWrapper):
 
 
 def read_file(path: Path, out: TextIOWrapper):
-    print("Reading", path.relative_to(repo_root))
+    logging.debug(f"Reading {path.relative_to(repo_root)}")
     with path.open("r") as f:
         lines = f.read().splitlines()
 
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -794,6 +794,35 @@ list(APPEND SOURCE_FILES ${RES_AUTOGEN_FILES})
...
+        set(outputs "${CMAKE_SOURCE_DIR}/docs/plugin-meta.lua")
+    endif()
+
+    # If additional build performance is needed: consider having the script
+    # generate a proper listing of files containing lua meta directives or at
+    # least use only headers here. This command takes only ~25ms for me on
+    # Linux.
+    add_custom_command(
+        COMMENT "Generating Lua definitions"
+        DEPENDS "${CMAKE_SOURCE_DIR}/scripts/make_luals_meta.py" ${SOURCE_FILES}
```
Nerixyz: You can make the script emit a depfile like GCC or Clang. It already knows which files it touches.

Something like
```
../docs/plugin-meta.lua: path/to/header.hpp path/to/other/header.hpp
```

`DEPFILE` is supported with make, ninja, and VS. Paths might need to be relative (I'm not 100% sure). It could be an argument to the script like `make_luals_meta.py --depfile luals_meta.d`.

src/CMakeLists.txt
```diff
@@ -794,6 +794,35 @@ list(APPEND SOURCE_FILES ${RES_AUTOGEN_FILES})
...
 
+if (SKIP_LUA_DEFINITIONS_GENERATION)
+    message(STATUS "Skipping lua definition generation")
+else()
+    message(STATUS "Doing lua definition generation")
+    find_package(Python3 COMPONENTS Interpreter)
+
+    # HACK(Mm2PL): Trick make into running our script
+    if (FORCE_LUA_DEFINITIONS_GENERATION)
+        set(outputs "${CMAKE_SOURCE_DIR}/docs/plugin-meta.lua" _fake_output)
```
Nerixyz, line 805: Why is this needed?

Mm2PL: @pajlada asked me to have an option to ALWAYS generate lua definitions and fail the build if it fails.

Nerixyz: I suppose this is for CI(?). If depfiles are used, then the generation will always the first time to generate the depfile (iirc). This could also use a post build event in that case - or use [`add_custom_target`](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target) as that will always be considered out-of-date.

pajlada: By default, we should try to build lua definitions at build time. But if this fails for whatever reason, we should not fail the build. Right now, if lua generation fails with no flag specifies, the entire chatterino build fails. I'd prefer if that doesn't happen.
If the user specifies `SKIP_LUA_DEFINITIONS_GENERATION`, I want the lua definition building to not be attempted at all. This is correctly implemented right now.
If the user specifies `FORCE_LUA_DEFINITIONS_GENERATION`, I want the lua definition build to use the same logic of whether or we need to build something, but it should error out if the lua definition building fails (e.g. due to missing python, or some error in the python script).

That's the same logic we have for the json generation in the eventsub library. See the `GenerateJson.cmake` file for how it's handled there in case I've explained things weird

pajlada (changes requested):
By default, we should _try_ to build LUA definitions at build time. If this fails, the build should not fail.
If the user does not have Python available, the build should not fail.

If the user enables `SKIP_LUA_DEFINITIONS_GENERATION`, we should not even try to find Python to start the LUA definition building.

If the user enables the `FORCE_LUA_DEFINITIONS_GENERATION`, we should fail if Python is unavailable, or the script fails. I don't think we need to mess with changed files. If nothing has changed, and FORCE_LUA_DEFINITIONS_GENERATION is on, then nothing would build, and that's fine. I want this to be enabled in CI somewhere just to ensure the script, on a clean build, will always run once and not error.

~~This is the same logic we apply to the eventsub library python generation.~~

Nerixyz:
> or the script fails.

We should always fail if the script fails. The script doesn't use any dependencies. If the script fails (→ the user has Python installed), something is really wrong.

> This is the same logic we apply to the eventsub library python generation.

We don't do it there either (only checking if Python is available and that a venv can be created).

> `FORCE_LUA_DEFINITIONS_GENERATION`

There should be some option to do this for all build-time generated files (i.e. Lua and EventSub). I'd imagine we'll have more in the future.

pajlada:
> > or the script fails.
> 
> We should always fail if the script fails. The script doesn't use any dependencies. If the script fails (→ the user has Python installed), something is really wrong.
> 
> > This is the same logic we apply to the eventsub library python generation.
> 
> We don't do it there either (only checking if Python is available and that a venv can be created).

I updated the above post to not compare it to the eventsub python generation

> 
> > `FORCE_LUA_DEFINITIONS_GENERATION`
> 
> There should be some option to do this for all build-time generated files (i.e. Lua and EventSub). I'd imagine we'll have more in the future.

If we get a lot more we can consider it yeah

PR #6299, opened 2025-06-25
feat(luals-meta): add support for depfile and custom log levels

Nerixyz:
Mentioned this in https://github.com/Chatterino/chatterino2/pull/6297/files#r2166689940 so I added this. This isn't dependent on that PR.

This does two things:

- Adds the ability to (effectively) disable the logging. This is required for #6297 (I don't want to have these logs when building the project)
- Adds support for generating depfiles. They use absolute paths. Tested this with Ninja, and it worked (just needs [CMP0116](https://cmake.org/cmake/help/latest/policy/CMP0116.html) to silence a CMake warning).

### DIFF
--- a/scripts/make_luals_meta.py
+++ b/scripts/make_luals_meta.py
@@ -30,9 +30,12 @@ Non-command lines of comments are written with a space after '---'
 """
 
 from io import TextIOWrapper
+import os
 from pathlib import Path
 import re
 from typing import Optional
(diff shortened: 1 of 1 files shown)


Mm2PL:
I have already done basically this on my branch today except I did not 
commit and push it yet. I will merge your logging changes manually into 
my branch.

Nerixyz:
Oops, sorry

PR #6300, opened 2025-06-28
refactor: use `SettingWidget::dropdown` for emoji style

pajlada:
less repeating of supported emoji styles

This means we are now storing emoji styles in lowercase
(so "twitter" instead of "Twitter"). I've made the map lookup case
insensitive to compensate

### DIFF
--- a/src/widgets/settingspages/SettingWidget.cpp
+++ b/src/widgets/settingspages/SettingWidget.cpp
@@ -227,6 +227,8 @@ template SettingWidget *SettingWidget::dropdown<TabStyle>(
     const QString &label, EnumStringSetting<TabStyle> &setting);
 template SettingWidget *SettingWidget::dropdown<ShowModerationState>(
     const QString &label, EnumStringSetting<ShowModerationState> &setting);
+template SettingWidget *SettingWidget::dropdown<EmojiStyle>(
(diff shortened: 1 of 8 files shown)


src/widgets/settingspages/SettingWidget.cpp
```diff
@@ -227,6 +227,8 @@ template SettingWidget *SettingWidget::dropdown<TabStyle>(
     const QString &label, EnumStringSetting<TabStyle> &setting);
 template SettingWidget *SettingWidget::dropdown<ShowModerationState>(
     const QString &label, EnumStringSetting<ShowModerationState> &setting);
+template SettingWidget *SettingWidget::dropdown<EmojiStyle>(
+    const QString &label, EnumStringSetting<EmojiStyle> &setting);
```
Nerixyz, line 231: We should avoid this...

pajlada: I switched to this in [`5dd92ad` (#6293)](https://github.com/Chatterino/chatterino2/pull/6293/commits/5dd92ad55a8e0abab885eb48b068217bc683eea1) to prevent the template function from being defined inside the header file. imo this ugliness outweighs having the file in the header

pajlada:
image

PR #6303, opened 2025-06-28
feat: add SettingWidget::dropdown support for int style enums

pajlada:
I have migrated the `LastMessageLineStyle` and `ThumbnailPreviewMode` enums / settings over to this SettingWidget.


pajlada:
image

PR #6304, opened 2025-06-28
refactor: remove unused ignoredPhraseReplace setting

pajlada:
never exposed to user, no use case explained in original PR


pajlada:
image

PR #6305, opened 2025-06-29
refactor: remove GeneralPageView.addCheckbox

pajlada:
this PR cleans up some includes, and converts all remaining addCheckbox usages to use SettingWidget.

Additionally, I've added `[[nodiscard]]` to all SettingWidget creators, meaning your IDE will scream at you if you don't use `addTo` at the end of the builder pattern


pajlada:
image

PR #6308, opened 2025-06-29
refactor(liveupdates): use `WebSocketPool` over websocketpp

Nerixyz:
The big goal is to get rid of websocketpp in the future and use Boost.Beast exclusively.

This PR moves the live-updates (BTTV and 7TV) to use the WebSocketPool initially added for plugins. This mixes up the design of the live-updates manager and client a bit.

- Messages are now handled in the client. In my opinion, that's where they should've been handled in the beginning as well (but websocketpp doesn't think so). Connection close/open events are handled on the manager. The latter events are handled on the GUI thread (I could also add a lock there, not sure). Messages are still delivered on the websocket thread.
- The manager is now a `QObject` to be able to guard `QTimer`s (for reconnect backoff and heartbeats on 7TV).
- Removed virtual inheritance over CRTP - the rationale here is that the manager and client are instantiated once per "type" (BTTV/7TV), so there's no need for virtual here.

This is a draft because I added `onOpen` which I want to add separately.


tests/src/BasicPubSub.cpp
```diff
@@ -6,6 +7,7 @@
 #include <QJsonDocument>
 #include <QJsonObject>
 #include <QString>
+#include <QtCore/qtestsupport_core.h>
```
Nerixyz, line 10: Not sure about this one. We could easily roll our own functions...

pajlada:
At a quick glance this looks nice. I will take a more proper look when `onOpen` is moved to a separate PR & this PR is no longer a draft.
Thank you!

tests/src/BasicPubSub.cpp
pajlada: Up to you - I'm fine either way.
Worst case tests stop building if they change something and we have to fix that

CHANGELOG.md
```diff
@@ -32,12 +33,6 @@
 - Bugfix: Fixed flickering tooltips on Wayland when the mouse cursor is over them. (#6451)
 - Bugfix: Fixed `c2.Channel` comparing `false` to the same channel. (#6456)
 - Bugfix: Fixed an issue where the moderation icon was missing from the Moderation tab. (#6457)
-- Bugfix: Update the message length font when updating your chat font. (#6462)
```
pajlada, line 53: did an oopsie here

src/providers/seventv/SeventvEventAPI.hpp
```diff
@@ -65,34 +62,11 @@ class SeventvEventAPI
...
 
-protected:
-    std::shared_ptr<BasicPubSubClient<seventv::eventapi::Subscription>>
-        createClient(liveupdates::WebsocketClient &client,
-                     websocketpp::connection_hdl hdl) override;
-    void onMessage(
-        websocketpp::connection_hdl hdl,
-        BasicPubSubManager<seventv::eventapi::Subscription>::WebsocketMessagePtr
-            msg) override;
+    void stop();
```
pajlada, line 69: Looks like this function is unused, and after a discussion with you, it seems like the pimpl is calling the relevant stop functions.

Removing this will break tests, so needs a little bit of finagling.

pajlada (commented):
shortName_ is currently unused - you can forward it to the pool for use in thread name if you'd like, no requirement for this PR

src/providers/bttv/BttvLiveUpdates.hpp
```diff
@@ -44,15 +47,11 @@ class BttvLiveUpdates : public BasicPubSubManager<BttvLiveUpdateSubscription>
      */
     void partChannel(const QString &id);
 
-protected:
-    void onMessage(
-        websocketpp::connection_hdl hdl,
-        BasicPubSubManager<BttvLiveUpdateSubscription>::WebsocketMessagePtr msg)
-        override;
+    void stop();
```
pajlada, line 54: This stop function, just like the 7TV one, also looks unused. looks to also break tests so remove with caution

PR #6315, opened 2025-07-03
feat(WebSocket): add open event

Nerixyz:
Adds an open event for when the WebSocket handshake completed successfully.

Depends on #6314 and #6316.

### DIFF
--- a/tests/src/Plugins.cpp
+++ b/tests/src/Plugins.cpp
@@ -653,6 +653,7 @@ TEST_F(PluginTest, testTcpWebSocket)
         EXPECT_TRUE(open);
         messages.emplace_back(isText, std::move(data));
     });
+    // On GCC in release mode, using set() would cause the done function to be called instead.
     lua->set_function("open", [&] {
         EXPECT_FALSE(open);
         open = true;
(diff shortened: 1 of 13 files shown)


tests/src/Plugins.cpp
```diff
@@ -645,12 +645,19 @@ TEST_F(PluginTest, testTcpWebSocket)
...
+    bool open = false;
     lua->set("done", [&] {
         waiter.requestDone();
     });
     lua->set("add", [&](bool isText, QByteArray data) {
+        EXPECT_TRUE(open);
         messages.emplace_back(isText, std::move(data));
     });
+    // On GCC in release mode, using set() would cause the done function to be called instead.
+    lua->set_function("open", [&] {
```
Nerixyz, line 657: Don't know if it's us using sol incorrectly or GCC doing some weird optimization. It doesn't happen in debug mode, and doesn't happen on Clang and MSVC.

pajlada: Related to https://github.com/ThePhD/sol2/issues/608#issuecomment-372876206 https://sol2.readthedocs.io/en/latest/functions.html#binding-callable-objects

Also fixable by making the "open" function here accept a param, but using set_function is perfectly reasonable

pajlada (approved):

PR #6318, opened 2025-07-05
fix: elide user notes

pajlada:
This PR adds support for basic text elision in the Label class.

I have not tested many of the options combined with text elision, only the default setup + text elision on or off.

Unrelated changes:
 - Label: Renamed `preferedSize_` to `sizeHint_`.
 - Label: Added a `minimumSizeHint_` param - I didn't end up using this but I'd like to play with it more.
 - Label: Added a `getFontMetrics` private method that returns the font metrics for the current font style & scale
 - Label: Added a `getWidthWithoutOffset` private method that returns the width we have to deal with when it comes to text rendering. I just didn't want to type width - offset - offset again.

### DIFF
--- a/src/widgets/Label.cpp
+++ b/src/widgets/Label.cpp
@@ -33,6 +33,11 @@ void Label::setText(const QString &text)
     if (this->text_ != text)
     {
         this->text_ = text;
+        if (this->shouldElide_)
+        {
+            this->updateElidedText(this->getFontMetrics(),
+                                   this->getWidthWithoutOffset());
+        }
         this->updateSize();
         this->update();
     }
@@ -76,33 +81,40 @@ void Label::setWordWrap(bool wrap)
     this->update();
 }
 
+void Label::setShouldElide(bool shouldElide)
+{
+    this->shouldElide_ = shouldElide;
+    this->updateSize();
+    this->update();
+}
+
 void Label::setFontStyle(FontStyle style)
 {
     this->fontStyle_ = style;
     this->updateSize();
 }
 
-void Label::scaleChangedEvent(float scale)
+void Label::scaleChangedEvent(float /*scale*/)
 {
     this->updateSize();
 }
 
 QSize Label::sizeHint() const
 {
-    return this->preferedSize_;
+    return this->sizeHint_;
 }
 
 QSize Label::minimumSizeHint() const
 {
-    return this->preferedSize_;
+    return this->minimumSizeHint_;
 }
 
-void Label::paintEvent(QPaintEvent *)
+void Label::paintEvent(QPaintEvent * /*event*/)
 {
     QPainter painter(this);
 
-    auto metrics = getApp()->getFonts()->getFontMetrics(this->getFontStyle(),
-                                                        this->scale());
+    auto metrics = this->getFontMetrics();
+
     painter.setFont(
         getApp()->getFonts()->getFont(this->getFontStyle(), this->scale()));
 
@@ -111,7 +123,16 @@ void Label::paintEvent(QPaintEvent *)
     // draw text
     QRect textRect(offset, 0, this->width() - offset - offset, this->height());
 
-    int width = static_cast<int>(metrics.horizontalAdvance(this->text_));
+    auto text = [this] {
+        if (this->shouldElide_)
+        {
+            return this->elidedText_;
+        }
+
+        return this->text_;
+    }();
+
+    int width = static_cast<int>(metrics.horizontalAdvance(text));
     Qt::Alignment alignment = !this->centered_ || width > textRect.width()
                                   ? Qt::AlignLeft | Qt::AlignVCenter
                                   : Qt::AlignCenter;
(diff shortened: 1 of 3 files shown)


src/widgets/Label.hpp
```diff
@@ -30,23 +32,45 @@ class Label : public BaseWidget
...
 private:
     void updateSize();
-    int getOffset();
+    int getOffset() const;
+
+    /// Returns the current font style's font metric based on the current scale.
+    QFontMetricsF getFontMetrics() const;
+
+    /// Returns the width of this widget, with the offset calculated out if the offset is enabled.
+    qreal getWidthWithoutOffset() const;
```
Nerixyz:
```suggestion
    /// Returns the width of the content without padding.
    int innerWidth() const;
```

Could be used when painting as well.

src/widgets/Label.hpp
```diff
@@ -30,23 +32,45 @@ class Label : public BaseWidget
     bool getWordWrap() const;
     void setWordWrap(bool wrap);
 
+    /// Sets whether the text should elide if there's not enough room to
+    /// render the current text.
+    ///
+    /// When using text eliding, you should most likely set the horizontal size policy to Minimum
```
Nerixyz: This isn't needed, because the size hint provides a width of -1 → 0. By default, the size-hint it `Preferred`, so the only difference is that the widget can shrink... but it can't be smaller than 0dp.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -490,6 +490,8 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, Split *split)
 
     auto notesPreview = layout.emplace<Label>().assign(&ui_.notesPreview);
     notesPreview->setVisible(false);
+    notesPreview->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
```
Nerixyz: See above. The vertical size policy makes sense.

src/widgets/Label.cpp
```diff
@@ -127,30 +148,85 @@ void Label::paintEvent(QPaintEvent *)
...
+}
+
+qreal Label::getWidthWithoutOffset() const
+{
+    if (this->hasOffset_)
+    {
+        return this->width() - this->getOffset() - this->getOffset();
+    }
+
+    return this->width();
```
Nerixyz:
```suggestion
    return this->width() - (2 * this->getOffset());
```

src/widgets/Label.cpp
```diff
@@ -127,30 +148,85 @@ void Label::paintEvent(QPaintEvent *)
...
         option.setWrapMode(QTextOption::NoWrap);
     }
-    painter.drawText(textRect, this->text_, option);
+    painter.drawText(textRect, text, option);
 
 #if 0
-    painter.setPen(QColor(255, 0, 0));
-    painter.drawRect(0, 0, this->width() - 1, this->height() - 1);
+        painter.setPen(QColor(255, 0, 0));
+        painter.drawRect(0, 0, this->width() - 1, this->height() - 1);
```
Nerixyz: Should be reverted again.

pajlada: resolved in [`0184ab6` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/0184ab6b7470c0c2c6bd3c33bc7c379719e2be93)

src/widgets/Label.cpp
pajlada: resolved in [`d21587c` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/d21587c605d24a7b2d6e3738d567acbfb03fa69d)

src/widgets/Label.hpp
pajlada: resolved
renamed "offset" to "padding" in [`309a0c7` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/309a0c7a065ba0e5f22ba905fb7dcfb5292964c8)
renamed `getWidthWithoutOffset` in [`4f61acb` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/4f61acbbf64483f0a4dce0ea997ab9524875187c)
used when painting in [`8ee77fd` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/8ee77fdd374aee88ffef64a51f7b1a44c42bce57) and [`a363741` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/a36374172abbd3d2a5931937760d4dfb5c39e45a)

src/widgets/dialogs/UserInfoPopup.cpp
pajlada: resolved in [`8a184f5` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/8a184f5008f2a338ba06e16c38683c5ce5b04c26)

src/widgets/Label.hpp
pajlada: resolved in [`8a184f5` (#6318)](https://github.com/Chatterino/chatterino2/pull/6318/commits/8a184f5008f2a338ba06e16c38683c5ce5b04c26)

pajlada:
image

PR #6320, opened 2025-07-05
fix: set the app user model id when starting

Nerixyz:
Currently, Windows groups my Chatterino7 and debug builds of Chatterino together, even though they have a different AUMID.

Chatterino already knows which AUMID it should have. So it should set this when starting (this PR).


pajlada:
@brian6932 Could you chime in if this is any issue for you?

Nerixyz:
Also, @8thony might have some comments.

8thony:
@Nerixyz in Chatterino7 you forgot to change the AUMID [here](https://github.com/SevenTV/chatterino7/blob/b1190123270e21f4f6c4c5cca2b039b0fc7ed933/src/main.cpp#L48).
It should be consistent with what you have used [elsewhere](https://github.com/SevenTV/chatterino7/commit/ae8bbab7c83b71727f6f4554c93dbf85c3c2a335).

Nerixyz:
Ah, thank you. I didn't notice we did it there as well.

Nerixyz:
... we should probably have some constant for this string ...

PR #6321, opened 2025-07-06
chore: factor out AUMID creation

Nerixyz:
From https://github.com/Chatterino/chatterino2/pull/6320#issuecomment-3041252899 - it's always bad to have the same string twice if you can avoid it.

This factors out the AUMID to be in `Version` instead of being created in `Toasts` as well as being spelled out in `main`.

### DIFF
--- a/src/common/Version.cpp
+++ b/src/common/Version.cpp
@@ -36,6 +36,11 @@ Version::Version()
 
     this->generateBuildString();
     this->generateRunningString();
+
+#ifdef Q_OS_WIN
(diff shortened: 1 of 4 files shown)


pajlada (approved):

pajlada:
the `WinToast::configureAUMI` function confused me - I was thinking you missed that but apparently it just builds the AUMI string

Nerixyz:
> the `WinToast::configureAUMI` function confused me - I was thinking you missed that but apparently it just builds the AUMI string

Right, it should probably be `formatAUMI`... some day 😀

Issue #6313, opened 2025-07-03
highlighted system messages don't respect unstable-set-user-color


Felanbird:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [ ] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

yes I know this is an UNSTABLE command that's been basically forgotten, but someone asked in the discord as they wanted to change the cyan of someone, which is normalized to the exact same cyan that I force changed brian the day he changed, so at least 2 of us hate it.

### Screenshots

![Image](https://github.com/user-attachments/assets/2a45a72b-7c2f-44bf-bc3e-a7f76645cedc)

### OS and Chatterino Version

Chatterino Nightly 2.5.2 (commit fb375e14) built on 2025-03-13 with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 10 Version 22H2, kernel: 10.0.19045

PR #6322, opened 2025-07-06
fix: correctly override user color in sub gifts

pajlada:
This PR adds a helper function to figure out what color a user should have.

It reads from these sources, in order
1. UserDataController
2. A custom provided color
3. From ChannelChatters

I've also added some tests, including some IRCMessageHandler snapshot tests to support testing this a bit

Works towards fixing https://github.com/Chatterino/chatterino2/issues/6313 but doesn't actually fix it


pajlada:
image

PR #6326, opened 2025-07-10
feat: support window transparency on Windows

Nerixyz:
Extends https://github.com/Chatterino/chatterino2/pull/6137 to work on Windows. I only tested this on Windows 11.

### DIFF
--- a/src/widgets/Window.cpp
+++ b/src/widgets/Window.cpp
@@ -50,6 +50,10 @@
 #include <QStandardItemModel>
 #include <QVBoxLayout>
 
+#ifdef Q_OS_WIN
+#    include <dwmapi.h>
+#endif
+
 namespace chatterino {
 
 Window::Window(WindowType type, QWidget *parent)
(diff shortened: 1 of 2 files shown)


8thony:
On Windows 10, the window buttons get glitched, but there is no transparency on the window.
I used the normal `Dark` theme and the `Transparent.json` from the linked PR.  
Tested on commit c513010
Note: You might have to restart for the buttons to get glitched.
<table>
<tr>
 <td>Dark
 <td>Transparent
<tr>
 <td>dark
 <td>transparent
</table>

Nerixyz:
> On Windows 10

Thank you for testing on Windows 10!

> the window buttons get glitched

One set of the window buttons is from Windows, because `WS_SYSMENU`, `WS_MAXIMIZEBOX`, and `WS_MINIMIZEBOX` are set. I don't remember if it's okay to unset these. 

> but there is no transparency on the window.

That's unfortunate. I can't really test on Windows 10 right now, so this will stay a draft for now.

> Note: You might have to restart for the buttons to get glitched.

That's a problem on Linux as well (the implementation only changes the window transparency when creating the window).

PR #6332, opened 2025-07-13
fix: Fix a crash that would occur on exit if a ping sound played recently

pajlada:
Tested on:
 - Arch Linux (btw) with Pipewire
 - Windows 11 in a VM

Both no longer output "Audio thread did not stop within 1 second" on exit, which could crash things as we detached the thread.

### DIFF
--- a/src/controllers/sound/MiniaudioBackend.cpp
+++ b/src/controllers/sound/MiniaudioBackend.cpp
@@ -1,22 +1,17 @@
 #include "controllers/sound/MiniaudioBackend.hpp"
 
-#include "Application.hpp"
 #include "common/QLogging.hpp"
 #include "debug/Benchmark.hpp"
-#include "singletons/Paths.hpp"
-#include "singletons/Settings.hpp"
-#include "singletons/WindowManager.hpp"
+#include "util/QMagicEnum.hpp"
 #include "util/RenameThread.hpp"
-#include "widgets/Window.hpp"
 
-#include <boost/asio/executor_work_guard.hpp>
+#include <boost/asio.hpp>
 
 #define MINIAUDIO_IMPLEMENTATION
 #include <miniaudio.h>
 #include <QFile>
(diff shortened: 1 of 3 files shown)


src/controllers/sound/MiniaudioBackend.cpp
```diff
@@ -219,37 +220,43 @@ MiniaudioBackend::~MiniaudioBackend()
...
+        {
+            this->audioThread->join();
+            return;
+        }
 
-    qCWarning(chatterinoSound) << "Audio thread did not stop within 1 second";
-    this->audioThread->detach();
+        qCWarning(chatterinoSound)
+            << "Audio thread did not stop within 1 second";
+        this->audioThread->detach();
```
Nerixyz, line 238: Not from this PR, but this has the same use-after free issue as the other `detach` calls. If the audio thread resumes after the backend is destructed, it will read already free'd data (e.g. the io-context). The only safe thing we can do at this point is `std::terminate` (or we need to leak `this`).

pajlada: yeah that's the crash that occurred that led me to this fix. we detached, but the timer for stopping the sound engine was still running, which would sometimes segfault.

Operating systems like macOS have problematic behaviour when the app doesn't exit in a timely manner, where this maybe-can-crash behaviour is preferable to the OS itself force-killing Chatterino. We could ifdef it to macOS only, but this PR doesn't really change the detach behaviour of this or other controllers.

pajlada: Will do a follow-up PR and look into our usages of detach

pajlada:
image

src/controllers/sound/MiniaudioBackend.cpp
pajlada: https://github.com/Chatterino/chatterino2/pull/6333

PR #6333, opened 2025-07-13
fix: don't detach threads

pajlada:
Previously, we would detach threads that did not exit in a reasonable
time, meaning they'd continue running and potentially accessing data
that had been freed, causing us to crash.
With this change, we _dont_ detach the thread, and let the thread
destructor do as it pleases (which will be terminating).
Neither solution is clean - old solution would sometimes work, but this
should give us a better indicator of where we're doing things wrong.

CC @Nerixyz


pajlada:
image

PR #6334, opened 2025-07-13
fix: Re-set SVG renderer aspect ratio after load

pajlada:
The `QSvgRenderer::setAspectRatioMode` function only sets the aspect
ratio setting for the currently loaded SVG document.
https://github.com/qt/qtsvg/blob/29c3b512d9ca25941b18f7ee52ce8b80f003f3b4/src/svg/qsvgrenderer.cpp#L311-L320

When the theme changes, or we otherwise call load, this does not retain
our aspect ratio settings.
https://github.com/qt/qtsvg/blob/29c3b512d9ca25941b18f7ee52ce8b80f003f3b4/src/svg/qsvgrenderer.cpp#L412-L432

This means we need to ensure we re-set the SVG renderer's aspect ratio
after calling load.


pajlada:
image

PR #6335, opened 2025-07-13
fix: convert the Split Header "add split" button to an SVG

pajlada:
Dependant on https://github.com/Chatterino/chatterino2/pull/6334


Nerixyz:
Although I suggested it, I'm not sure if this is the way to go anymore - at least for these simple icons. If we drew these ourselves, they would probably look better. We could ensure that they line up with the pixel grid.

If you still want to go with this, I suggest matching the style we had before. That would mean the button doesn't have padding and a background:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="17" height="28" viewBox="0 0 17 28">
   <rect x="0" y="0" width="17" height="28" fill="#191919" opacity="0.7" />
   <rect x="3.5" y="13" width="10" height="1" fill="#c1c1c1" opacity="0.7" />
   <rect x="8" y="8.5" width="1" height="10" fill="#c1c1c1" opacity="0.7" />
</svg>
```

This will look sharp on 1x scale at a DPR of 1.

PR #6338, opened 2025-07-15
Adding option to customize log file timestamps.

droidicus:
This PR add an option to use the same timestamp format for log files as is shown in the Chatterino UI:
image

This option is False by default to preserve the previous behavior of using "HH:mm:ss" as the formatting string.


src/singletons/Settings.hpp
```diff
@@ -152,6 +152,8 @@ class Settings
         "/appearance/enableAnimationsWhenFocused", false};
     QStringSetting timestampFormat = {"/appearance/messages/timestampFormat",
                                       "h:mm"};
+    BoolSetting customLogTimestamp = {
+        "/appearance/messages/customLogTimestamp", false};
```
Nerixyz: Bikeshed: This isn't really a custom timestamp. It's more like that the timestamp format is inherited. Wouldn't it be better to turn this into a string, so it's fully customizable and independent of the chat timestamp format?

droidicus: I wanted to follow the existing dropdown example for simplicity and uniformity, but would be fine with any changes.
I think my preference would be to either change the wording to more closely match "inherited" or adding a separate dropdown in the moderation/logs settings dialog. If a simple text box is prefered I can also make that change, but it is less user friendly in my view.

droidicus: After looking at this a bit more, there are complications from the chat timestamp display option as well, I will refactor.

droidicus: I refactored this to two dropdowns to allow the message timestamp format and log timestamp format to differ, let me know what you think.

src/singletons/Settings.hpp
```diff
@@ -152,6 +152,8 @@ class Settings
         "/appearance/enableAnimationsWhenFocused", false};
     QStringSetting timestampFormat = {"/appearance/messages/timestampFormat",
                                       "h:mm"};
+    QStringSetting logTimestampFormat = {
+        "/appearance/messages/logTimestampFormat", "hh:mm:ss"};
```
pajlada: Please reformat the setting to take up 4 full lines

```suggestion
    QStringSetting logTimestampFormat = {
        "/appearance/messages/logTimestampFormat",
        "hh:mm:ss",
    };
```

Additionally, the default value here should be the same as we used in logs before (i.e. `HH:mm:ss`)

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -521,7 +521,7 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         ->addTo(layout);
 
     layout.addDropdown<QString>(
-        "Timestamp format",
+        "Message Timestamp format",
```
pajlada: no title casing
```suggestion
        "Message timestamp format",
```

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -538,6 +538,21 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                                    : args.value;
         },
         true, "a = am/pm, zzz = milliseconds");
+
+    layout.addDropdown<QString>(
```
pajlada: Please move this dropdown to the Moderation sidebar, under the Logs tab, and un-titlecase the label

pajlada (changes requested):
Just some small nits wrt formatting & its default value, overall this looks good to me

fourtf:
I'm on my phone, so I can't check the settings myself. But it looks to me like the log timestamp doesn't really fit into the messages section of the settings. It's not related to the messages that get displayed in the chat. Is there maybe a log section that this could fit in?

pajlada:
> I'm on my phone, so I can't check the settings myself. But it looks to me like the log timestamp doesn't really fit into the messages section of the settings. It's not related to the messages that get displayed in the chat. Is there maybe a log section that this could fit in?

Yup! Under the moderation sidebar, under the Logs tab https://github.com/Chatterino/chatterino2/pull/6338#discussion_r2209690663

src/singletons/Settings.hpp
droidicus: The deleted line 156 shows the value that was used in the logs as `hh:mm:ss`, not `HH:mm:ss`.

pajlada: you added that line, it's new with this PR
https://github.com/Chatterino/chatterino2/pull/6338/files#diff-fa312dd94554fc49b9a2ecea5f290b0e78bb6bec70526d8ff712eaa79077051aR153-R160

the original value was defined in LoggingChannel.cpp (look at the deleted lines) https://github.com/Chatterino/chatterino2/pull/6338/files#diff-8e34776cc4565c7189d1d88abdfeb80155c932f6285e7af9d81f33f8f4003c8fL185-L187

droidicus: Ah whoops, I was reading from the suggested change accidently!

droidicus: I remember now...
From the `Qtime.toString()` docs `hh` and `HH` are equivalent (both use 24 hour formatting) when there is no am/pm display, such as this case. I changed the default to match the message timestamp option list which uses `hh:mm:ss` and `hh:mm:ss a`

If changing to use `HH` is still preferred, I suggest that the timestamp format lists for both message timestamp formats and the log timestamp formats be changed as well. Let me know your preference, and I can make this change if desired.

image
https://doc.qt.io/qt-6/qtime.html#toString

src/widgets/settingspages/GeneralPage.cpp
droidicus: Moved to the moderation sidebar, logs tab.
Updated PR screenshots to reflect.
The `ComboBox` methods differs between the `General` and `Moderation` sidebar selection, let me know if this implementation looks good.

src/singletons/Settings.hpp
pajlada: The important part is that it's not a breaking change, so if it's not breaking this is fine by me

src/widgets/settingspages/GeneralPage.cpp
pajlada: This looks fine to me - it can be updated to match at a later point, or upon request.

pajlada (approved):

pajlada:
Thank you @droidicus! As a first-time contributor, you are now eligible to be listed in the About page in Chatterino.

PR #6346, opened 2025-07-19
Add option to use server timestamp from the message instead of the local clock time for logging.

droidicus:
This PR adds the option to use `message->serverReceivedTime` instead of the local clock time for logging timestamps.

Feedback request: I do not know how stable the server timestamps are over time, but this change does show non-monotonic timestamps in logs. I have left the default behavior using the local clock to prevent these non-monotonic timestamps in the default configuration. This is significantly more likely to occur with miliseconds enabled in the logs, as the vast majority of the negative timedeltas that I have observed so far in a moderately fast chat channel are < 100 ms:

image
image

A corner-case example: If non-monotonic timestamps are received that span midnight, the file-partition logic will cause the logger to open and close files multiple times before stabilizing (e.g. if the message receive timestamps come in the order `23:59:57`, `00:00:01`, `23:59:59`, `00:00:02`, `00:00:05`, etc. then the active log file will change three times instead of once).

I have not seen this in my own testing, but if it occurs it is more likely to happen on extremely high velocity chat channels.

image


Nerixyz:
We should have documentation for these fields in `Message`...

`serverReceivedTime` isn't always set. When it's not set (e.g. [here](https://github.com/Chatterino/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/tests/snapshots/IrcMessageHandler/sub-first-time.json#L84) for some reason), we should fall back to the current time. 

For the interface, there should be a different name for the setting. As a user, I wouldn't know what a server receive time is (as a developer that's also a bit confusing, hence the comment at the start). A tooltip should list the tradeoffs.

droidicus:
> For the interface, there should be a different name for the setting. As a user, I wouldn't know what a server receive time is (as a developer that's also a bit confusing, hence the comment at the start). A tooltip should list the tradeoffs.

Thanks for the feedback, I will update.
Do you have recommendation for the naming?

droidicus:
> `serverReceivedTime` isn't always set. When it's not set (e.g. [here](https://github.com/Chatterino/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/tests/snapshots/IrcMessageHandler/sub-first-time.json#L84) for some reason), we should fall back to the current time.

@Nerixyz Looking into the source of this timestamp, I believe there is already a fallback to current time if there is not a server-side timestamp in the parsing code. In addition the tests use mocked timestamps, likely why the example json files have blank strings: https://github.com/droidicus/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/src/util/IrcHelpers.cpp#L9

That being said, I have added checks for null timestamps and Epoch timestamps (e.g. if it is instantiated from a zero-length string) to fallback to the current time as well.

Let me know if this is what you were looking for.

pajlada:
Have you looked into in which scenarios it's instantiated from a zero-length string?
this feels very ugly: `message->serverReceivedTime != QDateTime::fromMSecsSinceEpoch(QString("").toLongLong())`

src/singletons/helper/LoggingChannel.cpp
```diff
@@ -166,9 +166,20 @@ void LoggingChannel::openStreamLogFile(const QString &streamID)
 void LoggingChannel::addMessage(const MessagePtr &message,
                                 const QString &streamID)
 {
-    QDateTime now = QDateTime::currentDateTime();
+    QDateTime messageTimestamp;
+    if (getSettings()->logTimestampFromMessage &&
+        !message->serverReceivedTime.isNull() &&
+        message->serverReceivedTime !=
+            QDateTime::fromMSecsSinceEpoch(QString("").toLongLong()))
```
Nerixyz: Do we ever construct a date like `QDateTime::fromMSecsSinceEpoch(0)`? `isNull` should be enough for default constructed date times.

Nerixyz:
> In addition the tests use mocked timestamps, likely why the example json files have blank strings: [droidicus/chatterino2@`8085d6f`/src/util/IrcHelpers.cpp#L9](https://github.com/droidicus/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/src/util/IrcHelpers.cpp#L9)

That should still result in a string like `1970-01-01T00:00:00Z`, not an empty string. The empty string is only returned if the date is invalid (i.e. default constructed): https://github.com/qt/qtbase/blob/c9ed0591ab56006030499139491a4271666ec642/src/corelib/time/qdatetime.cpp#L4652-L4654.

> Do you have recommendation for the naming?

Something like "Use Twitch's timestamps"? Although that will be wrong if we add support for other platforms.

src/singletons/helper/LoggingChannel.cpp
droidicus: @Nerixyz @pajlada 
If it is possible for `tmi-sent-ts` to exist as a tag in a message, but the value is a zero length string/not a number, that could result in this case. Note I haven't seen this happen in my limited testing: https://github.com/droidicus/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/src/util/IrcHelpers.cpp#L25-L30

Epoch timestamps are also constructed for tests: https://github.com/droidicus/chatterino2/blob/8085d6f633da2347afa8aa0a2ba3f76a3926d600/src/util/IrcHelpers.cpp#L44-L49

I would be happy to remove this check if you don't think it is needed with the `isNull` check, or change to `QDateTime::fromMSecsSinceEpoch(0)` to be less ugly.

pajlada:
"Use Twitch's timestamps" seems good to me

droidicus:
> "Use Twitch's timestamps" seems good to me

Done. I also added a tooltip which may need the verbage adjusted as well.

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -165,6 +165,16 @@ ModerationPage::ModerationPage()
         logTimestampFormat->setToolTip("a = am/pm, zzz = milliseconds");
         logsTimestampFormatLayout.append(logTimestampFormat);
 
+        QCheckBox *logTimestampFromMessage = this->createCheckBox(
+            "Use Twitch's timestamps", getSettings()->logTimestampFromMessage);
+        logTimestampFromMessage->setToolTip(
+            "Using Twitch's timestamps for logging can result in out-of-order "
+            "timestamps in the log files.\nE.g. Consecutive timestamps may "
+            "appear like this: 11:32:39, 11:32:38, 11:32:40, 11:32:40, "
+            "11:32:42, etc.");
```
Nerixyz: I was more thinking of an explanation on _why_ someone would want to enable this. There could be a sentence like "This might result in out-of-order timestamps in log files" at the end.

src/singletons/helper/LoggingChannel.cpp
Nerixyz: Yeah, I don't think we ever construct them like this except for tests. And if we do, that's a bug. After all, this is the timestamp we would show to the user.

src/widgets/settingspages/ModerationPage.cpp
droidicus: Done, let me know if the wording still needs work.

Nerixyz: Looks good 👍

src/singletons/helper/LoggingChannel.cpp
droidicus: Removed Epoch test, left `isNull()` test.

Nerixyz (commented):
Looks good from my side. I don't have any right to approve though.

src/singletons/Settings.hpp
```diff
@@ -571,6 +571,10 @@ class Settings
         "/logging/logTimestampFormat",
         "hh:mm:ss",
     };
+    BoolSetting logTimestampFromMessage = {
+        "/logging/logTimestampFromMessage",
+        false,
+    };
```
pajlada: naming nit, will need changes to places that use it

```suggestion
    BoolSetting logTryUseTwitchTimestamps = {
        "/logging/tryUseTwitchTimestamps",
        false,
    };
```

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -165,6 +165,16 @@ ModerationPage::ModerationPage()
         logTimestampFormat->setToolTip("a = am/pm, zzz = milliseconds");
         logsTimestampFormatLayout.append(logTimestampFormat);
 
+        QCheckBox *logTimestampFromMessage = this->createCheckBox(
+            "Use Twitch's timestamps", getSettings()->logTimestampFromMessage);
+        logTimestampFromMessage->setToolTip(
+            "Twitch's timestamp is when the message was recieved by the chat "
+            "server.\nOtherwise the timestamp is when the message was recieved "
+            "by the chat client.\nTwitch's timestamp is more accurate, however "
+            "it can result in out-of-order timestamps in the log files.");
```
pajlada: Some bikeshedding for the tooltip that I think reads better
```
Try to use Twitch's timestamp (the time when the message was received by Twitch's chat server), rather than your computer's local timestamp.
Note that using this setting can result in out-of-order timestamps in the log files, and that if Twitch's timestamp was unavailable for a message, it will fall back to your computer's local timestamp.
```

The main issues I have with the current tooltip is:
 - Mentioning accuracy doesn't convey anything relevant in my opinion. The accurate thing to do would be to include _more_ information (i.e. both server received time _and_ local time), which is not something we support.
 - We should mention that if the server-side-time is unavailable, that this will fall back to your local time.

src/singletons/Settings.hpp
droidicus: Done.

src/widgets/settingspages/ModerationPage.cpp
droidicus: Done.

pajlada (approved):

PR #6348, opened 2025-07-20
fix: don't use same websocket connection for multiple accounts

### LINKED ISSUE #5926: The websocket connections need to die when user is changed

pajlada:
It's not permitted to use the same EventSub WebSocket connection for subscriptions regarding multiple users.
In this PR, I add a new signal that fires _before_ Helix updates, so we can properly unsubscribe from any subscriptions made by the previous user.
To ensure we don't reuse WebSocket connections, each WebSocket connection now has a User ID attached to it.
This attachment happens when the first subscription is confirmed, and the User ID gets detached from it when the last subscription is removed from the connection.
Since these are async operations, changing users will often result in creating a fresh connection that we try to use instead, but it ensures we don't reuse connections.

Additionally, I have made the following changes:
1) Clear subscriptions that we have unsubscribed from
2) Added a `/debug-eventsub` command to print all subscriptions -> connection maps, and connection -> subscription maps

Fixes #5926


pajlada:
image

Issue #1574, opened 2020-02-28
Icons are inconsistent


fourtf:
**Describe the bug**
The icons in the main UI are inconsistent. The screenshot below are showing 4 different ways of drawing icons:
- Custom drawn, well scaling icons (Settings and User)
- Tab '+' button, which is made of 1px lines - fixed in https://github.com/Chatterino/chatterino2/pull/6371
- Split kebab menu is a PNG - fixed in https://github.com/Chatterino/chatterino2/pull/6372
- Split '+' button is a PNG - fixed in https://github.com/Chatterino/chatterino2/pull/6349
- Moderation sword which is also a png but has a border - fixed in https://github.com/Chatterino/chatterino2/pull/6398
- Emote icon is an svg, which scales well but is drawn pixelated - fixed in #6080 

**Screenshots**
![image](https://user-images.githubusercontent.com/6320036/75578404-12641600-5a64-11ea-8ac0-b40402b5ce46.png)

![image](https://user-images.githubusercontent.com/6320036/75578534-4808ff00-5a64-11ea-87df-c362cc1173b5.png)

**Additional information**
The solution should draw high quality at all scales.

pajlada:
Fixed in https://github.com/Chatterino/chatterino2/pull/6371 https://github.com/Chatterino/chatterino2/pull/6372 https://github.com/Chatterino/chatterino2/pull/6349 https://github.com/Chatterino/chatterino2/pull/6398 #6080 
 (and maybe some more PRs I haven't listed)

PR #6349, opened 2025-07-20
fix: split header "add split" button inconsistency

pajlada:
This button was previously a PNG with a background.

With this change, I have added a new type of Button which intends to draw primitive shapes using configured colors, line thicknessess, and padding.

For this change, I have only implemented the plus button (which is pretty similar to the NotebookButton plus painting).

Works towards solving the meta issue of https://github.com/Chatterino/chatterino2/issues/1574

Previously, I attempted to use an SVG for this but it did not end up good enough.

### DIFF
--- a/src/widgets/buttons/DrawnButton.cpp
+++ b/src/widgets/buttons/DrawnButton.cpp
@@ -1,0 +1,108 @@
+#include "widgets/buttons/DrawnButton.hpp"
+
+#include "singletons/Theme.hpp"
+
+#include <QPainter>
+
+namespace chatterino {
+
+DrawnButton::DrawnButton(Type type_, Options options, BaseWidget *parent)
+    : Button(parent)
+    , foreground(this->theme->messages.textColors.system)
+    , foregroundHover(this->theme->messages.textColors.regular)
+    , basePadding(options.padding)
+    , baseThickness(options.thickness)
+    , type(type_)
+{
+    this->setContentCacheEnabled(true);
+}
+
+void DrawnButton::setBackground(QColor color)
+{
+    this->background = color;
+    this->invalidateContent();
+}
+
+void DrawnButton::setBackgroundHover(QColor color)
+{
+    this->backgroundHover = color;
+    this->invalidateContent();
+}
+
+void DrawnButton::setForeground(QColor color)
+{
+    this->foreground = color;
+    this->invalidateContent();
+}
+
+void DrawnButton::setForegroundHover(QColor color)
+{
+    this->foregroundHover = color;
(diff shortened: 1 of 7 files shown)


src/widgets/buttons/DrawnButton.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+DrawnButton::DrawnButton(Type type_, Options options, BaseWidget *parent)
+    : Button(parent)
+    , foreground(this->theme->messages.textColors.system)
+    , foregroundHover(this->theme->messages.textColors.regular)
+    , basePadding(options.padding)
+    , baseThickness(options.thickness)
+    , type(type_)
+{
+    this->setContentCacheEnabled(true);
```
Nerixyz, line 13: Not sure if we even need this here, but I guess it doesn't hurt.

src/widgets/buttons/DrawnButton.cpp
```diff
@@ -0,0 +1,108 @@
...
+            painter.setPen(pen);
+
+            auto innerSize = this->rect().size();
+            innerSize.setHeight(innerSize.width());
+            QRect inner;
+            inner.setSize(innerSize);
+            inner.moveCenter(this->rect().center());
+
+            int padding = static_cast<int>(
+                static_cast<float>(this->basePadding) * this->scale());
```
Nerixyz: The thickness is rounded, but the padding isn't. Is that intended? Also, for future types, it might be helpful to move these to a separate method.

src/widgets/buttons/DrawnButton.cpp
pajlada: Yeah I didn't run any benchmarks for whether caching was worth it or not, my thinking was that it probably won't hurt enough to have the content cache flag enabled to matter.

src/widgets/buttons/DrawnButton.cpp
pajlada: Padding and line thickness were moved to private methods in [`bb22455` (#6349)](https://github.com/Chatterino/chatterino2/pull/6349/commits/bb224555976d22d7be4e27fc330440da1638bc18)

I made padding round in [`8266638` (#6349)](https://github.com/Chatterino/chatterino2/pull/6349/commits/8266638363143dd318d92d3d017c8a234682f828)

pajlada:
I made one more change that wasn't technically necessary, changing the width of the "add split" button from ~17.5px to 16px. The 16px matches the scrollbar's full width, which I think looks nicer.

image

pajlada:
image

PR #5754, opened 2024-12-05
feat(plugins): add basic message construction

Nerixyz:
This adds the API I described in https://github.com/Chatterino/chatterino2/discussions/4999#discussioncomment-11450186. There are minor differences (e.g. flags belong to a specific element). Some helper functions aren't there yet. This is only the bare-bones construction.

The `/hello-world` example works with this implementation.

PR #6353, opened 2025-07-22
feat(plugins): allow message introspection

Nerixyz:
Continuation of #5754. This allows users to introspect messages (i.e. look at the elements they're made up of).

The fields of messages are modifiable. Plugins can also add/remove elements. To allow Lua access to elements, they are wrapped in a custom iterator that keeps a reference to the original message. The fields of elements are mostly static right now. But it would be possible to allow modification.
A future feature could allow plugins to clone messages, which would then be modifiable again.


tests/src/Plugins.cpp
```diff
@@ -970,6 +971,167 @@ TEST_F(PluginTest, ChannelAddMessage)
...
+{
+    configure();
+    sol::protected_function pfn = lua->script(R"lua(
+        return function(msg)
+            return {
+                function()
+                    assert(msg.flags == c2.MessageFlag.Debug)
+                    msg.flags = c2.MessageFlag.System
+                    assert(msg.flags == c2.MessageFlag.System)
+                end,
```
pajlada: Move this test function to a separate lua->Script or something for "functions that should always work even if the message is frozen"

tests/src/Plugins.cpp
```diff
@@ -970,6 +971,167 @@ TEST_F(PluginTest, ChannelAddMessage)
...
+        {
+            ASSERT_TRUE(res.valid());
+        }
+        else
+        {
+            ASSERT_FALSE(res.valid());
+            sol::error err = res;
+            ASSERT_EQ(std::string_view(err.what()), "Message is frozen");
+        }
+    }
```
pajlada, line 1202: If these are not dependant on each other,  put them in separate local scopes to communicate that better

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +231,504 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+
 void createUserType(sol::table &c2)
 {
-    c2.new_usertype<Message>("Message",
-                             sol::factories([](const sol::table &tbl) {
-                                 return messageFromTable(tbl);
-                             }));
+    c2.new_usertype<ElementRef>(
+        "MessageElement", sol::no_constructor, "type",
+        sol::property([](const ElementRef &el) {
```
Mm2PL: do something like this to force a clang-format to insert a new line
```suggestion
        "MessageElement", sol::no_constructor, //
        "type", sol::property([](const ElementRef &el) {
```

src/messages/MessageElement.hpp
```diff
@@ -194,6 +194,9 @@ class MessageElement
 
     virtual QJsonObject toJson() const;
 
+    // for plugins
```
pajlada: better comment

src/messages/MessageColor.cpp
```diff
@@ -50,4 +50,43 @@ QString MessageColor::toString() const
...
+        case Type::System:
+            return QStringLiteral("system");
+        case Type::Link:
+            return QStringLiteral("link");
+        default:
+            return {};
+    }
+}
+
+MessageColor MessageColor::fromLua(const QString &spec, Type fallback)
```
pajlada, line 70: looks unused

src/messages/MessageElement.hpp
```diff
@@ -277,16 +297,26 @@ class SingleLineTextElement : public MessageElement
...
+        return this->words_;
+    }
 
 private:
     MessageColor color_;
     FontStyle style_;
 
-    struct Word {
-        QString text;
-        int width = -1;
```
pajlada, line 277: this was unused

src/controllers/plugins/api/Message.hpp
```diff
@@ -7,33 +7,29 @@
...
----@field user_id? string The ID of the user who sent the message
----@field channel_name? string The name of the channel this message appeared in
----@field username_color? string The color of the username
----@field server_received_time? number The time the server received the message (in milliseconds since epoch)
----@field highlight_color? string|nil The color of the highlight (if any)
----@field elements? MessageElementInit[] The elements of the message
+---@class c2.MessageElementBase
+---@field flags c2.MessageElementFlag The element's flags
+---@field tooltip string The tooltip (if any)
+---@field trialing_space boolean Whether to add a trailing space after the element
```
pajlada: trailing_space

pajlada (changes requested):

Nerixyz:
I refactored the access for message elements. If a property for a message element doesn't exist, we now return `nil` instead of throwing. This is more consistent with non-existing properties (e.g. `element.words` on a badge will behave like `element.foobar`).

tests/src/Plugins.cpp
```diff
@@ -1132,6 +1175,188 @@ TEST_F(PluginTest, MessageConstness)
...
+    ASSERT_TRUE(nonExistent[3].is<std::nullptr_t>());
+
+    // test that accessing anything outside the elements vector causes an error
+    auto res = pfn(msg, 0, "flags");
+    ASSERT_FALSE(res.valid());
+    res = pfn(msg, 42, "flags");
+    ASSERT_FALSE(res.valid());
+}
+
+// Test that we can access properties of message elements
```
pajlada: nit: update comment to match what the test does

pajlada (changes requested):
just mini nit, otherwise looks good

pajlada (approved):

pajlada:
@Mm2PL feel free to merge this in when you've taken a look / tested it / you're happy with it

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+        if (!this->msg || this->index >= this->msg->elements.size())
+        {
+            return nullptr;
+        }
+        return this->msg->elements[this->index].get();
+    }
+
+    MessageElement &ref() const
+    {
+        auto *el = element();
```
Mm2PL:
```suggestion
        auto *el = this->element();
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+        if (!el)
+        {
+            throw std::runtime_error("Element does not exist or expired");
+        }
+        return *el;
+    }
+
+    const MessageElement &cref() const
+    {
+        const auto *el = constElement();
```
Mm2PL:
```suggestion
        const auto *el = this->constElement();
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+        return *el;
+    }
+
+    /// Cast this element to `T`. Otherwise nullopt is returned.
+    /// Use `.map()` to access the content.
+    template <typename T>
+    sol::optional<T &> as() const
+    {
+        // using ref() to error if the reference is invalid
+        auto *el = dynamic_cast<T *>(&ref());
```
Mm2PL:
```suggestion
        auto *el = dynamic_cast<T *>(&this->ref());
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+        return *el;
+    }
+
+    /// Cast this element to `const T`. Otherwise nullopt is returned.
+    /// Use `.map()` to access the content.
+    template <typename T>
+    sol::optional<const T &> asConst() const
+    {
+        // using cref() to error if the reference is invalid
+        const auto *el = dynamic_cast<const T *>(&cref());
```
Mm2PL:
```suggestion
        const auto *el = dynamic_cast<const T *>(&this->cref());
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+    ///
+    /// This is called recursively.
+    /// If the callback returns something, we return an `optional<T>` otherwise
+    /// we return `void`.
+    template <typename TReturn, typename T, typename... Rest>
+    auto visitOne(auto &&cb, auto &&...rest) const
+        -> std::conditional_t<std::is_void_v<TReturn>, void,
+                              sol::optional<TReturn>>
+    {
+        auto *el = dynamic_cast<T *>(maybeConstElement<std::is_const_v<T>>());
```
Mm2PL:
```suggestion
        auto *el = dynamic_cast<T *>(this->maybeConstElement<std::is_const_v<T>>());
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -232,16 +208,557 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+                return el.cref().hasTrailingSpace();
+            },
+            [](const ElementRef &el, bool trailingSpace) {
+                el.ref().setTrailingSpace(trailingSpace);
+            }),
+        "padding", sol::property([](const ElementRef &el) {
+            return el.asConst<CircularImageElement>().map(
+                [](const CircularImageElement &el) {
+                    return el.padding();
+                });
```
Mm2PL: You could use references to methods for many of these.
```suggestion
            return el.asConst<CircularImageElement>().map(
                &CircularImageElement::padding);
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -282,16 +263,576 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+        {
+            return sol::nullopt;
+        }
+        return *el;
+    }
+
+    template <typename T>
+    bool is() const
+    {
+        return dynamic_cast<const T *>(&cref()) != nullptr;
```
Mm2PL:
```suggestion
        return dynamic_cast<const T *>(&this->cref()) != nullptr;
```

src/controllers/plugins/api/Message.cpp
```diff
@@ -282,16 +263,576 @@ std::shared_ptr<Message> messageFromTable(const sol::table &tbl)
...
+                        std::make_shared<QColor>(QColor::fromString(sv));
+                }
+            }),
+        // must be read only (but it might be helpful for generic Lua functions)
+        "frozen", sol::property([](Message *msg) {
+            return msg->frozen;
+        }),
+        "elements",
+        [](const std::shared_ptr<Message> &msg) {
+            if (!msg)
```
Mm2PL: Why is this the only method to worry about having a valid pointer? I guess std::shared_ptr<Message>(nullptr) is technically a value, but I think we should worry more about returning such a bad pointer into Lua.

Nerixyz: Yea, we can't get a nullptr here - removed the check.

PR #6361, opened 2025-07-27
refactor(SettingWidget): use custom setting widgets

pajlada:
- **Adding option to allow using server timestamp from the message instead of the local clock time for logging.**
- **Add changelog entry**
- **update formatting**
- **adding checks for null timestamps, and epoch timestamps from zero-length string initialization**
- **Changing verbage on checkbox and add tooltop explaining non-monotonic timestamps**
- **Update formatting.**
- **Changing tooltip to explain the difference between the timestamps.**
- **Removing epoch test before using message->serverReceivedTime.**
- **Code review changes to variable names and tooltip.**
- **Partially implement SettingWidgets for bare QLayout's**
- **make use of SettingWidget for the new checkbox**
- **refactor: move our custo msettings widgets to their own file**
- **fix: make use of custom setting widget types**


pajlada:
image

PR #6362, opened 2025-07-29
Add emote completion widget filter for only zero-width emotes using :~

droidicus:
Searching with `:` is unaffected:
image

Searching with `:~` will only show zero-width emotes:
image


src/controllers/completion/strategies/ClassicEmoteStrategy.cpp
```diff
@@ -24,7 +33,17 @@ void ClassicEmoteStrategy::apply(const std::vector<EmoteItem> &items,
...
-    // Second pass: if there is an exact match, put that emote first
+    // Second pass: filter only zero-width if needed
+    if (zeroWidthOnly)
+    {
+        auto [first, last] =
+            std::ranges::remove_if(output, [](const EmoteItem &emoteItem) {
+                return !emoteItem.emote->zeroWidth;
+            });
+        output.erase(first, last);
+    }
```
Nerixyz: We could check for zeroWidth in the first pass.

pajlada (changes requested):
I would prefer if this is not a setting.

Can you (or someone else) think of a scenario where a user might not want this functionality enabled?

src/controllers/completion/strategies/ClassicEmoteStrategy.cpp
droidicus: @Nerixyz do you mean combining the logic to check both in the first pass, or swapping the first and second passes?

Nerixyz: The first one - checking for both in the first pass. Ideally by early `continue` if some item doesn't match to avoid nesting.

droidicus:
@pajlada I would be happy to remove it as a setting if desired. The corner case that I thought about was for emotes that start with a `~` might be filtered. I did a bit of looking around to see if there were any emotes or emoji that start with a `~` and I could not find any, if they do exist they are not common.

src/controllers/completion/strategies/ClassicEmoteStrategy.cpp
droidicus: Done!

pajlada:
after receiving feedback, I'd prefer if you do merge pushes :pray: it makes it easier to review

droidicus:
@pajlada sorry about that, I can do that going forward. Changes made since feedback: https://github.com/Chatterino/chatterino2/pull/6362/files/576be22f97e62a78136e8d7d33613af370cf459f..b8ed95dcc51cdfc1b7c09dc25928db9ef7624f91

src/controllers/completion/strategies/ClassicEmoteStrategy.cpp
```diff
@@ -3,21 +3,34 @@
...
 
+#include <algorithm>
+
 namespace chatterino::completion {
 
 void ClassicEmoteStrategy::apply(const std::vector<EmoteItem> &items,
                                  std::vector<EmoteItem> &output,
                                  const QString &query) const
 {
     QString normalizedQuery = query;
```
Nerixyz, line 23: Aside: I wonder if we could use string views here and in the other strategy

Nerixyz (commented):
I think it's okay if we clear non-zero-width emotes in the smart strategy first. Realistically, that should reduce the size of the completions drastically, so the other processing will benefit.

pajlada (changes requested):
Behaviour with this PR:
If the user has colon-completion enabled and the user types `:~`, all zero-width emotes will show up in their completion popup, no full width emotes.
This is due to the changes you made to `ClassicEmoteStrategy::apply` https://github.com/Chatterino/chatterino2/pull/6362/files#diff-4971f08e369368d8e58f81fbfde37e4cfe91aea14c53c6b5879d5202695a0dbbR20
This is good and expected

If the user has colon-completion disabled and the user types `:~Rai<TAB>`, it won't tab complete anything. If the user types `~Rai<TAB>` it won't tab complete anything.
You've made changes to the `ClassicTabEmoteStrategy` but it never comes into effect because we only use `normalizedQuery` if the user is attempting to tab-complete an emoji (e.g. `:jo<TAB>` for `:joy:`). For this to work, you'd need to modify the logic here https://github.com/Chatterino/chatterino2/pull/6362/files#diff-4971f08e369368d8e58f81fbfde37e4cfe91aea14c53c6b5879d5202695a0dbbR85-R99 - however it feels out of scope of what you've intended to do, so I would suggest undoing changes to the `*TabEmoteStrategy` classes.

droidicus:
Thanks, I wasn't aware of this behavior when colon completion was disabled. I will do another pass.

droidicus:
Changes to the `*TabEmoteStrategy` classes have been undone. Thanks!

pajlada (approved):

Issue #1989, opened 2020-09-26
Backup settings


fourtf:
**What should be added?**

**Why should it be added?**

fourtf:
make sure that accounts are not included when exporting settings

PR #6363, opened 2025-07-30
feat: add backup restore UI

### LINKED ISSUE #2751: Load backup settings from ui
Dropdown menu to load backup settings in ui.
Would share implementation with #1989

Nerixyz:
This adds a backup restore UI to Chatterino. It can be accessed from the about page: 
<details><summary>Image</summary>

</details>

This shows a dialog where users can select the individual backups to restore. Each backup shows the timestamp.
<details><summary>Image</summary>

</details> 

After the user selects the files, Chatterino creates a restore plan and shows it to the user.
<details><summary>Image</summary>

</details>

Chatterino force-restarts and applies the action (idk how flatpak works). We need to force exit to avoid running any destructors or anything that could try to save anything.
<details><summary>Image</summary>

</details>

This doesn't have any detection for corrupted settings yet. But it's built such that the dialog could run before initializing the settings.
Closes #2751.


src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            auto res = QMessageBox::question(
+                this, "Are you sure?", formatPlan(actions),
+                QMessageBox::No | QMessageBox::Yes);
+            if (res == QMessageBox::Yes)
+            {
+                QProcess twin;
+                // 😱 what about flatpak?????
+                twin.setProgram(QApplication::applicationFilePath());
```
Mm2PL: This should probably use the same thing as src/RunGui.cpp:138 for macos

src/util/Helpers.hpp
Nerixyz, line 1: This file is a mess of different unrelated things and we should organize the utils.

Mm2PL:
I really wish this wouldn't need the force restart business and we could shut down gracefully just not saving anything. 
 - We already have `Settings::disableSave()` which seems that it could be used to bypass that. 
 - `WindowManager`'s saving could be skipped by closing all windows via `closeAll` which sets a don't save flag. 
 - Commands are only saved when settings dialog is confirmed.
 - The only thing that does not already have a flag for this situation is `UserDataController`.

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,259 @@
...
+    }
+    else
+    {
+        this->combo = new QComboBox;
+        this->combo->addItem("None (keep current)");
+        auto dtf = QLocale::system().dateTimeFormat(QLocale::ShortFormat);
+        for (const auto &backup : availableBackups)
+        {
+            this->combo->addItem(
+                u"%1 (%2)"_s.arg(stdPathToQString(backup.path.filename()),
```
Mm2PL: Could you look into adding the file size into this text? It's a good way to tell if the file was truncated or otherwise emptied.

pajlada:
It would not be far fetched for us to see if the follow scenario happens:
 - Current `settings.json` doesn't exist, or fails to load
 - There is a valid settings backup file

In that scenario, we could prompt the user whether they'd want to load their settings from the backup _before_ we start Application, meaning we'd not need to do any restarts.
Unless I'm missing some scenario, I'd prefer we go for that sort of approach instead.

Mm2PL:
> pajlada left a comment (Chatterino/chatterino2#6363)
> [...]
> Unless I'm missing some scenario, I'd prefer we go for that sort of approach instead.

I would like to be able to restore backups of settings if I've for 
example accidentally deleted my commands or restored all the hotkeys to 
defaults.

pajlada:
> I would like to be able to restore backups of settings if I've for example accidentally deleted my commands or restored all the hotkeys to defaults.

ok maybe I'm misunderstanding what it is we're aiming to do then, that sounds more like https://github.com/Chatterino/chatterino2/issues/1989 to me

Nerixyz:
> that sounds more like #1989 to me

#1989 calls for exporting all settings such that they can be shared or backed up _elsewhere_.
#2751 wants to have _some UI_ to restore the backups Chatterino makes on its own.

Nerixyz:
This PR does too much (see review comments). Let's discuss what we actually want in the linked issue. I currently see three things:

- Notification and restore dialog when Chatterino fails to load some settings but detects some backups
- Restore dialog in the settings to restore any backup (mostly what this PR does)
- Settings import/export

Issue #3691, opened 2022-04-24
Chatter list and changing channels leads to unexpected behavior


Mm2PL:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

If you open a chatter list for a channel and then close the split, the chatter list closes itself as expected. However if you change the split's channel the chatter list stays. If you open a usercard from this bugged chatter list, it will be opened in the new channel instead. Maybe this kind of behavior could be useful to someone cross-checking chatter lists, I'd rather have this be remade into a full-on feature.

### Screenshots

https://user-images.githubusercontent.com/25011746/164982833-a40756e8-83fd-41f5-a817-e34912bfff62.mp4

### OS and Chatterino Version

Chatterino 2.3.5 (commit c994a589) running on Arch Linux, kernel: 5.17.3-arch1-1

PR #6365, opened 2025-08-02
refactor: make chatter list its own widget

pajlada:
This makes the Chatter List widget/window/popup into an actual
BaseWindow instead of a QDockWidget. the QDockWidget only had
functionality on Linux, and it barely functioned since we're not using
it with any dockable layouts/windows


8thony:
- Should solve #3691

pajlada:
image

PR #6366, opened 2025-08-02
fix: tabular number message length

pajlada:
technically 2 fixes in 1, uh oh.

This makes the Message Length label in the text input use the same font / font features as the timestamps do (so tabular numbers), meaning the width of the message count will only change with the number of digits.

While I was here, I made the Emote Button be right-aligned, so it doesn't move at all when the Message Length label changes in width.


pajlada:
image

PR #6367, opened 2025-08-02
feat: dynamically track mutability of messages

Nerixyz:
This adds a flag to `Message` (`frozen`) that indicates that it shouldn't be modified anymore. This is set once it gets added to a channel.

**Why?**
Plugins (#6353). Consider the following:
- Create a message `msg`
- Add `msg` to some channel
- (some time later) add an element to `msg` or modify it in some other way
We don't allow this because we assume messages are immutable (except flags) once they're in a channel. Technically, this is also possible in C++. We can't enforce this statically - mutable messages are `std::shared_ptr<Message>` and immutable ones are `std::shared_ptr<const Message>`. There can be const and non-const pointers to the same message at the same time.
In Lua, we need to check this invariant.

### DIFF
--- a/src/messages/Message.hpp
+++ b/src/messages/Message.hpp
@@ -86,6 +86,13 @@ struct Message {
     };
     ReplyStatus isReplyable() const;
     uint32_t count = 1;
+
+    /// Can this message be modified?
+    ///
+    /// This is only used for plugins right now. This value is only ever set to
+    /// true.
+    mutable bool frozen = false;
+
     std::vector<std::unique_ptr<MessageElement>> elements;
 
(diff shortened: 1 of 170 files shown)


src/messages/Message.hpp
```diff
@@ -86,13 +86,25 @@ struct Message {
     };
     ReplyStatus isReplyable() const;
     uint32_t count = 1;
+
+    /// Can this message be modified?
```
Mm2PL, line 90: I'd like to see the comment say something about the rendering code expecting the messages to never change (except `flags` and `frozen`). And that this should be checked before modifying messages.

Mm2PL (approved):

PR #6371, opened 2025-08-03
chore: use DrawnButton instead of NotebookButton

pajlada:
The button no longer gets colored blue when a split is being dragged.
I'm OK with that being fixed in a follow-up PR, since it's kinda difficult to manage. It currently just uses a global static bool to check if a split is being dragged - no events.


pajlada (commented):
Image

PR #6375, opened 2025-08-06
fix(plugins): lookup by main lua state in http request

Nerixyz:
When making an HTTP request from a timeout handler, the associated thread with the timeout handler could be destroyed right before the `finally` handler ran. This would invalidate the Lua state that we captured. To work around this, only keep track of the main state, which we can get from the plugin.


Mm2PL:
https://github.com/Remahy/Chatterino-Plugins/blob/main/crash-poc/init.lua

Mm2PL:
This fixes the issue with Remahy's plugin

pajlada (approved):

PR #6376, opened 2025-08-07
Stop sending `JOIN`/`PART` commands for channels starting with `/`

Mm2PL:
Plugins tend to use channels beginning with `/` as their own empty channels. We shouldn't send invalid JOIN commands for channels that are local. This is a simple hack, the fix for the root of the issue would be to give plugins a way to actually own channels, but that is going to take a lot of time to finalize.

We already do not send `JOIN`s for `/watching` and friends (with the exception of `/watching` for some reason)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - Bugfix: Emotes that failed to load their images now show as text. (#6355)
 - Bugfix: Fixed a crash that occurs when searching for emotes in channel-less contexts. (#6357)
 - Bugfix: Fixed command triggers showing as '/...' when the value is longer than the column width. (#6369)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
 - Bugfix: Emotes that failed to load their images now show as text. (#6355)
 - Bugfix: Fixed a crash that occurs when searching for emotes in channel-less contexts. (#6357)
 - Bugfix: Fixed command triggers showing as '/...' when the value is longer than the column width. (#6369)
+- Bugfix: Fixed sending IRC `JOIN`/`PART` commands for channels starting with `/`. (#6376)
```
pajlada: make dev entry and mention this is more a temporary fix for plugins

PR #6380, opened 2025-08-09
feat: pulse split input on self message received

pajlada:
This feature, disabled by default, pulses the split input in a green glow whenever one of your messages is received in the channel.


pajlada:
image

PR #6383, opened 2025-08-10
feat: show live indicator in usercard

### LINKED ISSUE #2699: Twitch stream live indicator in Usercard
# **What should be added?** #

**A LIVE or OFFLINE indicator should be shown (respectively) in a user's usercard to allow Chatterino users to easily identify whether or not the person is LIVE or OFFLINE.**
_**If you need more clarification on this, let me know in the comments.**_

# **Why should it be added?** #

**I believe this should be added so that users of Chatterino can know whether someone is live or not by just clicking on their username or whatnot to bring up the person's usercard and see some type of LIVE or OFFLINE indicator there.**

Nerixyz:
This PR adds a live indicator in the usercard:

- Default:
  
- When hovering the icon:
  

This isn't done as a text, because I think that would bloat the UI. A tooltip shows the current viewers. Maybe it could also show the uptime?

Fixes #2699.


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -355,6 +356,10 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, Split *split)
 
                 this->ui_.nameLabel = addCopyableLabel(box, "Copy name");
                 this->ui_.nameLabel->setFontStyle(FontStyle::UiMediumBold);
+                this->ui_.nameLabel->setPadding(QMargins(8, 0, 0, 0));
+                this->ui_.liveIndicator = new LiveIndicator(8);
```
pajlada: I'd prefer if the LiveIndicator is shown if the stream is live, and hidden if the stream is offline. Not used as padding.
It would mean modifying the `nameLabel` padding on the fly if you want it to look like it does now, but that seems trivial with your modifications to `Label::setPadding`

pajlada (changes requested):

pajlada (approved):

PR #6384, opened 2025-08-10
feat: add tooltips with follow/sub durations to usercard

Nerixyz:
So many bots have some followage and subage command. We know the dates in the usercard, but only show the absolute date. With this PR, the tooltip of these labels will contain the formatted duration - like "3 years, 2 months and 1 hour ago".

Computing these components is a bit more complicated. What we want is equivalent to the following JS code:
```js
Temporal.Now.instant()
  .since(Temporal.Instant.from("2018-07-04T06:34:24Z"))
  .round({
    smallestUnit: "hours",
    largestUnit: "years",
    relativeTo: Temporal.PlainDate.from({ year: 2025, month: 1, day: 1 }),
  });
```

Qt doesn't have any functionality for this, so we're using Boost.DateTime (header only).


src/util/FormatTime.cpp
```diff
@@ -71,4 +161,49 @@ QString formatTime(std::chrono::seconds totalSeconds)
...
+        if (n == 0)
+        {
+            return;
+        }
+
+        if (!out.isEmpty())
+        {
+            if (remaining == 1)
+            {
+                out += u" and ";
```
pajlada, line 211:
```suggestion
                out += u", and ";
```

src/util/FormatTime.hpp
```diff
@@ -11,4 +12,10 @@ QString formatTime(int totalSeconds);
 QString formatTime(const QString &totalSecondsString);
 QString formatTime(std::chrono::seconds totalSeconds);
 
+/// Formats a duration that's expected to be long (sevreal months or years) like
+/// "4 years, 5 days and 8 hours".
```
pajlada, line 16:
```suggestion
/// "4 years, 5 days, and 8 hours".
```

src/util/FormatTime.cpp
```diff
@@ -71,4 +161,49 @@ QString formatTime(std::chrono::seconds totalSeconds)
             std::numeric_limits<int>::max()))));
 }
 
+QString formatLongFriendlyDuration(const QDateTime &from, const QDateTime &to)
```
pajlada, line 176: As suggested by @Mm2PL in chat, it would be cool if it would use the "4 oldest" units instead

e.g.
1 year 1 month 1 day 1 hour ago
1 month 1 day 1 hour 1 minute ago
1 hour 1 minute ago

pajlada (changes requested):

pajlada (changes requested):
I think something is wrong when one of the "middle" fields is empty, e.g. if it would output:
1 year, 1 month, and 1 hour ago

see (ONLY TODAY!!)
image

8thony:
Is there a reason the tooltip is set to light mode?
The tooltips in Settings use dark mode.

Nerixyz:
> I think something is wrong when one of the "middle" fields is empty, e.g. if it would output: 1 year, 1 month, and 1 hour ago
> 
> see (ONLY TODAY!!)

Not quite. It was when the time of the "to" datetime is lower than that of the "from" datetime.

Nerixyz:
> Is there a reason the tooltip is set to light mode?
> The tooltips in Settings use dark mode.

Not intentional. Probably something with the palette we're using there.

PR #6385, opened 2025-08-10
fix: self drawn add button alignment and color

Nerixyz:
After #6349:
- If the width of the add button was even, the center would always be off.
   This is still the case, but we now ensure that we don't create the button with an even width.
- The button was bigger than before
- The background color was too dark (before, the background wasn't fully opaque)

This PR attempts to fix the issues.
It really only fixes them on non-fractional DPRs. On fractional DPR, the center of the button still looks off.


pajlada (changes requested):
>  If the width of the add button was even, the center would always be off.
    This is still the case, but we now ensure that we don't create the button with an even width.

Please change the scrollbar "expanded width" logic to match this change

> The button was bigger than before

Which button was bigger than before? The split add button is smaller in nightly than in stable

> The background color was too dark (before, the background wasn't fully opaque)

This change was intentional. Feel free to modify the highlight color if you want it to match the split header selected or unselected background on hover

Nerixyz:
> Please change the scrollbar "expanded width" logic to match this change

I found a better way.

> Which button was bigger than before? The split add button is smaller in nightly than in stable

Sorry, I didn't describe this correctly. The plus was larger after #6349 - with the PNG, there were some scaling artifacts where one pixel was slightly occupied.

Nerixyz:
> This change was intentional. Feel free to modify the highlight color if you want it to match the split header selected or unselected background on hover

Changed this to the (unselected) split background color. That makes it look more like before (except for the lighter border when unselected).

pajlada (approved):

PR #6386, opened 2025-08-11
Add support for plugin links

Mm2PL:
Gives plugins the ability to add links to message elements

image


Mm2PL:
@Nerixyz I'd like your review on this PR.

src/controllers/plugins/api/Message.cpp
```diff
@@ -163,7 +162,60 @@ std::unique_ptr<MessageElement> elementFromTable(const sol::table &tbl)
...
+        }
+        QString tooltip = tbl.get_or("tooltip", QString{});
+        if (!tooltip.isEmpty())
+        {
+            tooltip += "<br>";
+        }
+
+        switch (link->type)
+        {
+            case Link::Url:
```
Nerixyz, line 182: I'm not sure about this.

On the one hand I like this, because it shows the user what clicking would do, but on the other hand, this feels a bit too far - a plugin won't be able to really control the tooltip.

Also, with message modification, a plugin could overwrite this (mabye it could also inject some HTML to make the message disappear? Not sure).

An alternative could be to add some special marker to a link (that it came from a plugin) and have the channel-view show something special. Maybe even a special thing where holding shift would show the value of the link? Could be nice for regular links too.

src/controllers/plugins/api/Message.cpp
```diff
@@ -163,7 +162,60 @@ std::unique_ptr<MessageElement> elementFromTable(const sol::table &tbl)
...
+                tooltip += QString("<b>URL:</b> %1").arg(link->value);
+                break;
+            case Link::UserAction:
+                tooltip += QString("<b>Command:</b> %1").arg(link->value);
+                break;
+            case Link::CopyToClipboard:
+                tooltip +=
+                    QString("<b>Copy to clipboard:</b> %1").arg(link->value);
```
Nerixyz: If they will be in there, the values should probably be trimmed (or even left out?).

- For Url: Trimming is probably the best here. Though, one could still hide malicious stuff in the end.
- For UserAction: Trim. We should also make sure that we only ever send one message, so `hello\n/unban PluginAuthor` won't send as two messages.
- For CopyToClipboard: Maybe hiding (i.e. only show that this copies _something_)? There's not really much harm you can do other than to overwrite the clipboard. Not sure about all use-cases, but I could imagine something like "click to copy token" or something. You wouldn't want that to show.

src/controllers/plugins/api/Message.cpp
```diff
@@ -163,7 +162,60 @@ std::unique_ptr<MessageElement> elementFromTable(const sol::table &tbl)
...
+            case Link::UserInfo:
+            case Link::UserWhisper:
+            case Link::ReplyToMessage:
+                break;
+
+            // these types are not exposed to plugins
+            case Link::None:
+            case Link::AutoModAllow:
+            case Link::AutoModDeny:
+            case Link::InsertText:
```
Nerixyz, line 204: InsertText might be cool too.

src/controllers/plugins/SolTypes.cpp
```diff
@@ -139,6 +141,49 @@ int sol_lua_push(sol::types<chatterino::lua::ThisPluginState>, lua_State *L,
 
 }  // namespace chatterino::lua
 
+namespace chatterino {
+
+// Link
+bool sol_lua_check(sol::types<chatterino::Link>, lua_State *L, int index,
+                   std::function<sol::check_handler_type> handler,
```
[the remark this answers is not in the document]
Nerixyz: We could probably make this a bit lighter by using a function reference over a `std::function`, but that's for another PR.

src/controllers/plugins/api/Message.cpp
Mm2PL: InsertText seemed like it needed a rework to use. I probably should have said that before.

src/controllers/plugins/api/Message.hpp
```diff
@@ -86,6 +90,20 @@ c2.Message = {}
...
+// Note: code dependant on this needs these values to be meaningful `Link::Type`s
+/** @exposeenum c2.LinkType */
+enum class ExposedLinkType : std::uint8_t {
+    Url = Link::Type::Url,
+    UserInfo = Link::Type::UserInfo,
+    UserAction = Link::Type::UserAction,  // run a command/send message
+    JumpToChannel = Link::Type::JumpToChannel,
+    CopyToClipboard = Link::Type::CopyToClipboard,
+    JumpToMessage = Link::Type::JumpToMessage,
+};
```
pajlada, line 105: Could we expose Link Type instead of having this wrapper?
I understand we want to limit which link types we expose to plugins, but could we make the lua doc generator have some `/** hidethisenumvalue **/` feature instead?

src/messages/MessageElement.hpp
```diff
@@ -178,7 +178,7 @@ class MessageElement
     MessageElement(MessageElement &&) = delete;
     MessageElement &operator=(MessageElement &&) = delete;
 
-    virtual MessageElement *setLink(const Link &link);
+    virtual MessageElement *setLink(const Link &link, bool *ok = nullptr);
```
pajlada: I'd prefer if the `setLink` calls are reverted, and you instead to the checking based on the element type inside `src/controllers/plugins/api/Message.cpp`

src/controllers/plugins/api/Message.cpp
pajlada: I'd prefer if we wait with plugin-provided tooltips for Links until we have a safe interface for it (i.e. a way to set a tooltip that only allows for plaintext stuff)

src/controllers/plugins/api/Message.hpp
Mm2PL: The problem isn't documentation. I wanted unwanted types to not be exposed or named in the Lua VM. That felt like it would require messing with magicenum in complex ways.

pajlada: ~~How are types exposed to the Lua VM if not through the `/** @exposeenum c2.LinkType */` comment and `make_luals_meta.py`?~~

ok nvm I understand, `createEnumTable` would not know to skip certain values unless we do some other magicry with `magic_enum` there

Mm2PL: This is exposed via the makeEnumTable<...>() function call in PluginController::initSol() which takes an entire enum and makes a table for it to put it into `_G.c2`. https://github.com/Chatterino/chatterino2/pull/6386/files#diff-a99a29045e38fcb918fbb7a4f6100d9d4d6567a76bac4f5fa140e9be7a6f36f2R237

Mm2PL:
>  I'd prefer if we wait with plugin-provided tooltips for Links until we have a safe interface for it (i.e. a way to set a tooltip that only allows for plaintext stuff)

The tooltip data is now ignored when using links.

###### I can't mark your comment as resolved

src/controllers/plugins/api/Message.cpp
Mm2PL: Urls aren't truncated normally, so I don't see why plugins should have that specifically. UserAction doesn't send two messages in such configurations. It's validated like something you'd send manually. I've changed CopyToClipboard to only show that it copies *something*.

src/controllers/plugins/api/Message.cpp
Mm2PL: Custom tooltips can no longer be mixed with links

Nerixyz:
From testing:

- Links on timestamp elements don't work because they create a text element without passing the link/tooltip
- Links on reply curves don't work because it doesn't have a hitbox - its height is 0

<details><summary>Test</summary>

```lua
c2.register_command("/links", function(ctx)
    local chan = ctx.channel
    chan:add_message(c2.Message.new({
        id = "testing42",
        elements = { { type = "text", text = "testing" } },
    }))
    local lb = { type = "linebreak", flags = c2.MessageElementFlag.Text };
    chan:add_message(c2.Message.new({
        elements = {
            {
                type = "text",
                text = "Text Element",
                link = { type = c2.LinkType.Url, value = "https://twitch.com/forsen" },
            },
            {
                type = "single-line-text",
                text = "single-line-text",
                link = { type = c2.LinkType.UserInfo, value = "id:141981764" },
            },
            {
                type = "mention",
                display_name = "display_name[Mention]",
                login_name = "twitchdev",
                user_color = "link",
                fallback_color = "link",
            },
            {
                type = "timestamp",
                link = { type = c2.LinkType.JumpToChannel, value = "twitchdev" }
            },
            lb,
            {
                type = "reply-curve",
                link = { type = c2.LinkType.CopyToClipboard, value = "hello world" }
            },
            {
                type = "text",
                text = "some text",
            }
        }
    }))
    chan:add_message(c2.Message.new({
        flags = c2.MessageFlag.Centered,
        elements = {
            {
                type = "text",
                text = "Available Links",
                style = c2.FontStyle.ChatVeryLarge,
            },
            lb,
            {
                type = "text",
                text = "JumpToChannel",
                link = { type = c2.LinkType.JumpToChannel, value = "twitchdev" },
            },
            lb,
            {
                type = "text",
                text = "JumpToMessage",
                link = { type = c2.LinkType.JumpToMessage, value = "testing42" },
            },
            lb,
            {
                type = "text",
                text = "CopyToClipboard",
                link = { type = c2.LinkType.CopyToClipboard, value = "Hello, World!" },
            },
            lb,
            {
                type = "text",
                text = "Url",
                link = { type = c2.LinkType.Url, value = "https://example.com" },
            },
            lb,
            {
                type = "text",
                text = "UserAction",
                link = { type = c2.LinkType.UserAction, value = "My user action" },
            },
            lb,
            {
                type = "text",
                text = "UserInfo",
                link = { type = c2.LinkType.UserInfo, value = "twitchdev" },
            }
        }
    }))
end)
```

</details>

src/messages/MessageElement.cpp
```diff
@@ -1122,8 +1122,18 @@ TextElement *TimestampElement::formatTime(const QTime &time)
 
     QString format = locale.toString(time, getSettings()->timestampFormat);
 
-    return new TextElement(format, MessageElementFlag::Timestamp,
-                           MessageColor::System, FontStyle::TimestampMedium);
+    auto *text =
+        new TextElement(format, MessageElementFlag::Timestamp,
+                        MessageColor::System, FontStyle::TimestampMedium);
+    text->setLink(this->getLink());
```
Nerixyz, line 1129: Can't test this right now, but does the tooltip work as well?

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -609,11 +609,28 @@ qreal TextIconLayoutElement::getXFromIndex(size_t index)
     }
 }
 
+namespace {
+QSizeF calculateCurveSize(float neededMargin, qreal width, float radius)
```
Nerixyz: Add a space here and at the closing braces.

src/messages/MessageElement.cpp
Mm2PL: huh it doesn't i forgor
e261694

src/messages/MessageElement.cpp
```diff
@@ -1122,8 +1122,19 @@ TextElement *TimestampElement::formatTime(const QTime &time)
...
+    text->setTooltip(this->getTooltip());
```
Nerixyz, line 1130: There should be a `setTooltip` in `TimestampElement::addToContainer` too, I think (or make setTooltip virtual). Or maybe we could say that `TimestampElement` doesn't support tooltips.

Nerixyz (commented):
Works fine for me now.

pajlada:
this pr changes the way the curve element renders. I'm ok with either re-breaking curve element links in this PR, or ensuring they look the same.

new:
unsorted_20250823_12h15m46s

old:
unsorted_20250823_12h15m45s

Mm2PL:
@pajlada I have re-broken reply-curves, they now should render the same as before.

Mm2PL:
@pajlada could you review this when you have time?

pajlada:
> @pajlada could you review this when you have time?

I'll start taking a look at these leftover PRs this weekend

CHANGELOG.md
```diff
@@ -20,7 +11,7 @@
 - Minor: Added WebSocket API for plugins. (#6076, #6186, #6314, #6315)
 - Minor: Allow for themes to set transparent values for window background on Linux. (#6137)
 - Minor: Popup overlay now only draws an outline when being interacted with. (#6140)
-- Minor: Added basic message API to plugins. (#5754)
+- Minor: Added basic message API to plugins. (#5754, #6386)
```
pajlada: fix the changelog entry (move it to the right place)

pajlada (approved):
Just fix the changelog entry and you can merge this in

PR #6387, opened 2025-08-11
fix: image uploader setting import 

x3rt:
- If multiple headers are defined in the sxcu they would be stored delimited by a `\n` newline character, instead of the expected `;` semicolon causing uploads to use invalid headers
- The `URL` and `DeletionURL` settings would store the raw string meaning ShareX's `{response}` or `{json:files[0].url}` would be stored as such and therefore `getLinkFromResponse` would not parse the response correctly.

This PR attempts to fix the above. 
As this is my first time working with Chatterino's code I expect there to be things to nit-pick about my changes. I did attempt my best to follow the Contributing guidelines & match the format of other functions.

Some things to note;
1. I'm unsure whether you wish to support converting back to proper ShareX formatting during export.


Nerixyz:
> 1. I'm unsure whether you wish to support converting back to proper ShareX formatting during export.

Yes, we do want to be ShareX compatible - [we even advertise this](https://github.com/Chatterino/chatterino2/blob/f3b02f370bb3661c357d8cd6723bd051b511c3cb/src/widgets/settingspages/ExternalToolsPage.cpp#L205-L206).

In #6394 I added some test for this, because there were none. Once this gets merged, you can add more tests here and update the snapshots (change `UPDATE_SNAPSHOTS` to true, rerun tests, and reset `UPDATE_SNAPSHOTS` to false).

pajlada:
If you could fix the tests to update the "new reality" that'd be great.
The changes look fine from my perspective

x3rt:
Will do so when available

src/util/ImageUploader.cpp
```diff
@@ -4,11 +4,68 @@
...
 namespace {
 
+/**
+  * @brief Converts ShareX-style tokens throughout the URL to the expected format.
+  *
+  * Examples:
+  *  - https://example.com/{json:foo.bar[1]}/{json:foo.bar[2]}
+  *     -> https://example.com/{foo.bar.1}/{foo.bar.2}
+  *  - {json:{response}|foo.bar[1]}
+  *     -> {foo.bar.1}
```
Nerixyz, line 20: These should be added as tests too.

pajlada: yep i know, i've set up coverage so i can make sure i cover the cases
image

pajlada:
Somewhat unrelated: in [`9210736` (#6387)](https://github.com/Chatterino/chatterino2/pull/6387/commits/92107360f59a743b5c0ca47460f68f4f5c8e02f7) I removed the requirement for the `Name` field in the `sxcu` file. I now instead check that the `RequestURL` field exists, which exists in both ShareX's format & our format.

In [`5405f0f` (#6387)](https://github.com/Chatterino/chatterino2/pull/6387/commits/5405f0fd1ff785e8cb9406fb05d414d5973356e0) I added a few snapshot tests for testing the jsonpath feature - as far as I can tell from my brief testing this now works as expected. My sharex uploader settings I've tested has been very small though, so happy to have someone else test this as well.

pajlada:
Thank you @x3rt! As a first-time contributor, you are now eligible to be listed in the About page in Chatterino.

PR #6393, opened 2025-08-14
fix: take handler for `sol_lua_check` as function reference

Nerixyz:
I mentioned this in https://github.com/Chatterino/chatterino2/pull/6386#discussion_r2266900601 - we shouldn't take the handler as a `std::function` which could potentially allocate. The handler is called when the check fails, so it's not stored anywhere. Sol passes the handler as a template argument. We could do the same, but this would mean that we'd need to define the functions in the header. Instead, we can use a function reference (like [`std::function_ref`](https://en.cppreference.com/w/cpp/utility/functional/function_ref.html)). Since we aren't on C++ 26, we need to add this ourselves. I used [`llvm::function_ref`](https://github.com/llvm/llvm-project/blob/9315d701eb6f693fa1ca5d1e89ef437d42466000/llvm/include/llvm/ADT/STLFunctionalExtras.h) as the base for this, since it's concise (the actual implementation in libc++ and libstdc++ is larger because they handle more cases like noexcept/const/volatile, but the idea of keeping two pointers is the same). We're using C++ 20, so we can use concepts instead of SFINAE.


pajlada (approved):
Want this to wait until the Link PR is merged in?

Nerixyz:
> Want this to wait until the Link PR is merged in?

I'd be fine with both. Maybe @Mm2PL has a preferred order.

Mm2PL:
I'm fine with this being merged before link PR is merged, I'll deal with the conflicts. That one still needs a little more time in the oven.

PR #6397, opened 2025-08-17
fix: svgify chatters button

pajlada:
based on fluenticons `Person` - from https://github.com/Nerixyz/tmp-c2-icons

new looks:

unsorted_20250817_13h17m06s

unsorted_20250817_13h17m13s


Nerixyz (commented):
The chatters button looks a bit too noticeable in the dark mode (from the screenshots) because it's completely white. It should probably be some gray color.

pajlada:
image

Updated to use the same color as the kebab icon

pajlada:
image

PR #6412, opened 2025-08-21
fix: make moderation button pretty

### LINKED ISSUE #6401: Design touchup: Moderation icon
If someone feels inclined to do some design work, I'd love it if someone could touch up the moderation swords we're doing now.
They need to look somewhat similar to our current design, so:
 - Light & dark mode disabled (preferably a hollow sword)
 - Light & dark mode enabled (currently a filled in sword, green, but open to ideas)

See https://github.com/Chatterino/chatterino2/pull/6398 for latest change to the icons, which svgified them but also didn't really make them pretty.

pajlada:
Icon by @Nerixyz

Fixes #6401


jupjohn:
Visually, I'm not a fan of this icon (and the one it replaced) due to the border being nearly indistinguishable from the background when enabled on the dark theme. It gives the illusion that the sword has shrunk when enabled.

My alternative suggestions would be:
1. Replace the enabled icon's border with the same solid fill colour;
2. ~~Change the enabled icon's border to the same colour as the disabled one so it remains consistent (need to test, unsure if it would end up the same.)~~ this looks horrible.
3. Perhaps a darker green outline would work (like `#278400`?)

pajlada:
> Visually, I'm not a fan of this icon (and the one it replaced) due to the border being nearly indistinguishable from the background when enabled on the dark theme. It gives the illusion that the sword has shrunk when enabled.

That's fair - I will merge this in as-is because it's still an upgrade over the one in our main branch.

> My alternative suggestions would be:
> 1. Replace the enabled icon's border with the same solid fill colour;
> 2. ~~Change the enabled icon's border to the same colour as the disabled one so it remains consistent (need to test, unsure if it would end up the same.)~~ this looks horrible.
> 3. Perhaps a darker green outline would work (like `#278400`?)

I'm happy to have you experiment more - feel free to submit a PR if you find something that you feel like is an improvement!

Thanks for your feedback as always

PR #6417, opened 2025-08-26
fix: Twitch dropping `MAGIC_MESSAGE_SUFFIX`

brian6932:
The invisible Unicode block that `\u{e0000}`'s a part of, can't be sent to TMI any longer, it just gets dropped. `\u{34f}` shouldn't ever be dropped by TMI, it's future-proof, because emojis rely on this character to join. Chatterino currently only appends these characters when the duplicate message doesn't contain any spaces, so make sure to note when testing.

### DIFF
--- a/src/providers/twitch/TwitchChannel.cpp
+++ b/src/providers/twitch/TwitchChannel.cpp
@@ -80,9 +80,9 @@ using detail::isUnknownCommand;
 
 namespace {
 #if QT_VERSION < QT_VERSION_CHECK(6, 1, 0)
-const QString MAGIC_MESSAGE_SUFFIX = QString((const char *)u8" \U000E0000");
+const QString MAGIC_MESSAGE_SUFFIX = QString((const char *)u8" \u034f");
 #else
-const QString MAGIC_MESSAGE_SUFFIX = QString::fromUtf8(u8" \U000E0000");
+const QString MAGIC_MESSAGE_SUFFIX = QString::fromUtf8(u8" \u034f");
 #endif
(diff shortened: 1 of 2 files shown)


jupjohn:
Works for me :+1: 

image

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -80,9 +80,9 @@ using detail::isUnknownCommand;
 
 namespace {
 #if QT_VERSION < QT_VERSION_CHECK(6, 1, 0)
-const QString MAGIC_MESSAGE_SUFFIX = QString((const char *)u8" \U000E0000");
+const QString MAGIC_MESSAGE_SUFFIX = QString((const char *)u8" \u034f");
 #else
```
Nerixyz: This is no longer needed, as we don't support Qt <6.4.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -80,9 +80,9 @@ using detail::isUnknownCommand;
...
-const QString MAGIC_MESSAGE_SUFFIX = QString::fromUtf8(u8" \U000E0000");
+const QString MAGIC_MESSAGE_SUFFIX = QString::fromUtf8(u8" \u034f");
```
Nerixyz: Can we use `QStringLiteral` here?

pajlada (approved):
CC @Nerixyz @TETYYS

TETYYS:
I'm TETYYS and I approve this message

Nerixyz:
LGTM

PR #6420, opened 2025-08-26
feat(plugins): add JSON parsing/serialization

Nerixyz:
Most HTTP and WebSocket APIs use JSON. Config files often use JSON. We use three JSON APIs in Chatterino (if not more). Plugins should not need to ship their own JSON library.

This PR adds a `'json'` module.

**`json.parse`**

Uses Boost.JSON to parse a string into Lua types (tables/scalars). It uses the SAX API to avoid creating intermediate values. It optionally supports trailing spaces and comments.

**`json.stringify`**

This uses Rapidjson's `Writer`. Why Rapidjson and not Boost? Boost doesn't have a writer you can drive yourself, and their direct serialization is almost completely undocumented. Furthermore, direct serialization (i.e. letting Boost drive the serialization) requires much more boilerplate code. If I understand them correctly, then incremental serialization is a big goal for them and integrating that would be hard.

`nil` in Lua tables means that an element isn't there. We provide a special `json.null` to indicate that an element should be serialized as `null`. This aligns with https://github.com/mpx/lua-cjson which provides an equivalent value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Minor: Added setting for character limit of deleted messages. (#6491)
 - Minor: Added link support to plugin message API. (#6386, #6527)
 - Minor: Added a description for the logging option under moderation tab. (#6514)
-- Minor: Added a JSON API for plugins (`require('json')`). (#6420)
+- Minor: Added a JSON API for plugins (`require('chatterino.json')`). (#6420)
 - Minor: Consolidate font picking into one dialog. (#6531)
 - Minor: Added a menu action to sort tabs alphabetically. (#6551)
 - Minor: Fixed "edit hotkey" dialog opening like a normal window. (#6540)


Mm2PL:
Is it intended that `{"test": null}` parses as `{}`, not `{test=json.null}`?

src/controllers/plugins/api/JSON.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace chatterino::lua::api {
+
+sol::object loadJson(sol::state_view lua)
+{
+    return lua.create_table_with(    //
+        "parse", jsonParse,          //
+        "stringify", jsonStringify,  //
+        // pushed as lightuserdata
+        "null", static_cast<void *>(nullptr)  //
```
Mm2PL, line 86: Can you give this object a metatable and with `__tostring` which returns `json.null`?

Nerixyz:
> Is it intended that `{"test": null}` parses as `{}`, not `{test=json.null}`?

Yes, it's ["intended"](https://github.com/Chatterino/chatterino2/pull/6420/files#diff-8684c27f4960f911437bb162b4b77e5ba45cd3cc22fc496234fc0bb6f0b26f28R88). I'm not sure if that's the correct approach. Basically, if null should be parsed as `nil` or `json.null`. I'm happy to change that.

Mm2PL:
I would want it to be parsed as `json.null` so 
`json.parse(json.stringify(X))` is as close to `X` as possible, ideally 
not changing its meaning.

W dniu 15.10.2025 o 21:46, nerix pisze:
> Nerixyz left a comment (Chatterino/chatterino2#6420)
> 
>> Is it intended that `{"test": null}` parses as `{}`, not `{test=json.null}`?
> 
> Yes, it's ["intended"](https://github.com/Chatterino/chatterino2/pull/6420/files#diff-8684c27f4960f911437bb162b4b77e5ba45cd3cc22fc496234fc0bb6f0b26f28R88). I'm not sure if that's the correct approach. Basically, if null should be parsed as `nil` or `json.null`. I'm happy to change that.
>

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+        return true;
+    }
+
+    bool on_object_end(size_t /* nElements */, boost::system::error_code &ec)
+    {
+        if (this->elements.empty())
+        {
+            ec = makeCode(ParseError::NoElements);
+            assert(false && "We shouldn't hit this");
```
Mm2PL: If you want to keep it, make assert text more descriptive. Why do we have an error code for this scenario but we also assert it never happens?

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,379 @@
...
+        return true;
+    }
+
+    bool on_array_end(size_t /* nElements */, boost::system::error_code &ec)
+    {
+        if (this->elements.empty())
+        {
+            ec = makeCode(ParseError::NoElements);
+            assert(false && "We shouldn't hit this");
+            return false;
```
Mm2PL, line 162: Same as in `on_object_end`.

If you want to keep it, make assert text more descriptive. Why do we have an error code for this scenario but we also assert it never happens?

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,379 @@
...
+                  boost::system::error_code &ec)
+    {
+        lua_pushinteger(this->state, static_cast<lua_Integer>(i));
+        return this->appendValue(ec);
+    }
+
+    bool on_uint64(uint64_t i, boost::json::string_view /* source */,
+                   boost::system::error_code &ec)
+    {
+        lua_pushinteger(this->state, static_cast<lua_Integer>(i));
```
Mm2PL: `lua_Integer` is signed which means that parsing `18446744073709551615` ($2^{64}-1$) yields `-1`. I think this should be converted to a floating point like JS does to garble the data less. Consider `json.stringify(json.parse(18446744073709551615)` case:
 - currently this returns `-1`
 - JS returns `18446744073709552000` because that's the closest number it can represent. 
 - The closest float is `18446744073709551616`. My best guess is my tools are using a different rounding mode than JS.

src/controllers/plugins/api/JSONStringify.cpp
```diff
@@ -0,0 +1,252 @@
...
+            writer.Bool(lua_toboolean(L, -1) != 0);
+            break;
+        case LUA_TNUMBER: {
+            if (lua_isinteger(L, -1))
+            {
+                writer.Int64(lua_tointeger(L, -1));
+            }
+            else
+            {
+                writer.Double(lua_tonumber(L, -1));
```
Mm2PL: This yields bad JSON if given infinity. `json.stringify(math.huge)` returns an empty string.
My browser returns `null` for an equivalent call. Python is insane and returns `Infinity`, I really don't want to go this route. We could achieve `inf` and `-inf` by writing a double with a sufficiently large exponent such that the closest available one is `inf` or `-inf`, ex. `json.parse('1e309')`. Sadly only some implementations handle this gracefully:
- Firefox, Python, Node.js, our `json.parse`: parses as inf.
- jq: parses as `1.7976931348623157e+308` which is the largest possible double.
- go: craps out and says `1e309` is not a valid float64 (even if given free reign on what the value's type should be).


This also yields bad JSON if given NaN. My browser is insistent on `null` for this case too. Python accepts and returns `NaN`. There is no nice solution here. I think we should follow browsers and the standard here:
> Numeric values that cannot be represented in the grammar below (such
> as Infinity and NaN) are not permitted.
> [RFC 8259, section 6. Numbers](https://datatracker.ietf.org/doc/html/rfc8259#section-6)


Or we should decide we don't want to serialize these values and make it error.

src/controllers/plugins/api/JSONStringify.cpp
```diff
@@ -0,0 +1,252 @@
...
+        break;
+        case LUA_TLIGHTUSERDATA: {
+            auto *ptr = lua_touserdata(L, -1);
+            if (ptr == nullptr)
+            {
+                writer.Null();
+            }
+            else
+            {
+                luaL_error(L, "Non-null lightuserdata not supported");
```
Mm2PL: This error text does not make sense without knowing implementation details of this module.

src/controllers/plugins/api/JSONStringify.cpp
Nerixyz: Rapidjson can be statically instructed to behave like python, but I opted to error out.

src/controllers/plugins/api/JSON.cpp
Nerixyz: All lightuserdata use one global metatable, but it should be fine if we set this here, since we don't use them elsewhere (user-facing).

Mm2PL: This could be a small full userdata if you wanted.

src/controllers/plugins/api/JSON.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#ifdef CHATTERINO_HAVE_PLUGINS
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api {
+
+sol::object loadJson(sol::state_view lua);
```
pajlada, line 12: A little comment

src/controllers/plugins/api/JSONParse.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#ifdef CHATTERINO_HAVE_PLUGINS
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api {
+
+int jsonParse(lua_State *L);
```
pajlada, line 13: A little comment

src/controllers/plugins/api/JSONStringify.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#ifdef CHATTERINO_HAVE_PLUGINS
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api {
+
+int jsonStringify(lua_State *L);
```
pajlada, line 13: A little comment

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,388 @@
...
+    {
+        boost::system::error_code ec;
+        auto nRead = this->parser.write_some(false, sv.data(), sv.length(), ec);
+        if (ec)
+        {
+            luaL_error(this->parser.handler().state,
+                       "Failed to parse JSON: %s (at %s) pos: %d",
+                       ec.message().c_str(), ec.location().to_string().c_str(),
+                       static_cast<int>(nRead));
+            std::terminate();
```
pajlada: Add a comment about luaL_error's behaviour and that it behaves as if it's noreturn

src/controllers/plugins/api/JSONStringify.cpp
```diff
@@ -0,0 +1,277 @@
...
+/// table an array if it only has positive non-zero integer keys. One notable
+/// exception is an empty array. An empty table is considered an empty object.
+/// If the user wants to specify an empty array, they specify a table with
+/// `json_null` at item `0` (i.e. `{ [0] = c2.json_null }`). `json_null` is a
+/// `nullptr` lightuserdata.
+///
+/// Arrays might have holes in them (e.g. `{1, [10]=2}` or `{1, nil, 2}`), so we
+/// keep track of the maximum index we saw.
+///
+/// @returns The size of the array if  it is one - otherwise it's an object.
```
pajlada:
```suggestion
/// @returns The size of the array if it is one - otherwise it's an object.
```

src/controllers/plugins/api/JSONStringify.cpp
```diff
@@ -0,0 +1,277 @@
...
+            max = std::max(max, static_cast<size_t>(key));
+        }
+
+        lua_pop(L, 1);  // value
+    }
+
+    if (max > ENCODE_MAX_TABLE_LENGTH)
+    {
+        luaL_error(L, "Table is too big");
+        std::terminate();
```
pajlada: comment about terminate (or remove, whichever you prefer)

src/controllers/plugins/api/JSONStringify.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    return max;
+}
+
+/// luaL_error but with [[noreturn]]
+[[noreturn]]
+void fail(lua_State *L, const char *msg, auto &&...args)
+{
+    luaL_error(L, msg, std::forward<decltype(args)>(args)...);
+    std::terminate();
```
pajlada: comment about terminate (or remove, whichever you prefer)

tests/lua/json/stringify.lua
```diff
@@ -0,0 +1,137 @@
+local JSON = require('json')
```
pajlada: cc @Mm2PL 

what should we "recommend" / is the standard for json imports in lua.
`JSON`? `json`?

Just thinking that we should be consistent in our tests, examples, and documentation from the start to give the highest chance of plugins being consistent in their style.

docs/chatterino.d.ts
```diff
@@ -375,4 +375,13 @@ declare namespace c2 {
         Original,
         Repost,
     }
+
+    function json_parse(
```
pajlada: @Mm2PL I'd really prefer if we focus on pure Lua support rather than trying to support transpiling.
How strongly do you want to support TS transpiling?

tests/lua/json/stringify.lua
Mm2PL: I think we should keep the name of the module and its local variable the same. 
This is [something the LuaRocks project insists on](https://github.com/luarocks/lua-style-guide#modules)

So this would be the convention:
```lua
local name = require('long.module.name')
local json = require('json')
```

docs/chatterino.d.ts
Mm2PL: With proper static type checking, it isn't really that necessary. Describing the API in two different ways in two different spots is not what I want to do. The ideal solution would be to generate these somehow from the same source as `docs/lua-meta/globals.lua`. That likely isn't happening due to special cases like `c2.register_callback` which takes different callback function signatures depending on the event given

docs/wip-plugins.md
```diff
@@ -769,3 +769,80 @@ require("data.file") -- tried to load Plugins/name/data/file.lua and errors beca
...
+local str = json.stringify({ foo = 1 }, { pretty = true })
+-- {
+--     "foo": 1
+-- }
+```
+
+`options` can be an optional table with the following optional keys:
+
+- `pretty` (boolean): Use newlines and indentation when stringifying
+- `indent_char` (string, default: ` `): Character to use when indenting object/array items
```
Mm2PL: Consider changing this to say 'default: space' not:
```markdown
` `
```

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    bool on_document_end(boost::system::error_code & /* ec */)
+    {
+        return true;
+    }
+
+    bool on_object_begin(boost::system::error_code &ec)
+    {
+        elements.append(Element{
```
Mm2PL:
```suggestion
        this->elements.append(Element{
```

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,416 @@
...
+            .index = 0,
+            .isObject = 1,
+        });
+        if (elements.size() > MAX_NESTING)
```
Mm2PL:
```suggestion
        if (this->elements.size() > MAX_NESTING)
```

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,416 @@
...
+        }
+        assert(this->elements.back().isObject);
+        this->elements.pop_back();
+
+        return this->appendValue(ec);
+    }
+
+    bool on_array_begin(boost::system::error_code &ec)
+    {
+        elements.append(Element{
```
Mm2PL:
```suggestion
        this->elements.append(Element{
```

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -0,0 +1,416 @@
...
+            .index = 0,
+            .isObject = 0,
+        });
+        if (elements.size() > MAX_NESTING)
```
Mm2PL:
```suggestion
        if (this->elements.size() > MAX_NESTING)
```

Nerixyz:
Should this use `@chatterino/json` or should that be done in a separate PR?

pajlada:
follow-up PR imo, but I don't have a strong opinion so if @Mm2PL disagrees we can follow him

Mm2PL (commented):
Missed a `this` in a couple places.

pajlada:
were we not able to use `@chatterino/json`?

Wissididom:
> were we not able to use `@chatterino/json`?

image

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Minor: Added setting for character limit of deleted messages. (#6491)
 - Minor: Added link support to plugin message API. (#6386, #6527)
 - Minor: Added a description for the logging option under moderation tab. (#6514)
+- Minor: Added a JSON API for plugins (`require('json')`). (#6420)
```
Mm2PL:
```suggestion
- Minor: Added a JSON API for plugins (`require('chatterino.json')`). (#6420)
```

Nerixyz:
> were we not able to use `@chatterino/json`?

Yes. LuaLS doesn't understand `@meta @chatterino/json`. It doesn't understand `@meta chatterino/json` either, so `chatterino.json` is used. This can conflict with local files (if you have a `chatterino/json.lua` in your local plugin). The workaround here would be to use that path as the import name.

PR #6436, opened 2025-09-03
refactor: Emotes

Nerixyz:
Quite a big one. I want to put some parts in separate PRs.

**The issue**: We currently have three third party emote providers. Each of them is a separate entry in `Application` and each of them needs to be separately wired up throughout different parts of the app (emote popup, message building, split header refreshing, twitch channel refreshing, input completion, etc.). In the future, plugins should also be able to have their own emote providers that integrate with the app (not just message building).

**The solution**: Abstract emote providers into [`EmoteProvider`](https://github.com/Nerixyz/chatterino2/blob/refactor/emote-providers/src/controllers/emotes/EmoteProvider.hpp)s. Most parts of the app don't care if emotes are from BTTV, FFZ, or 7TV. All they care about is their order and "origin" (channel/global). Of course, there are exceptions to this. Namely, live updates which are still specific to one provider and drive-by updates of `TwitchChannel` (FFZ badges and 7TV user data). The key thing is that they're scoped to `TwitchChannel` - they don't leak into other parts such as input completion.

We now have one [`EmoteController`](https://github.com/Nerixyz/chatterino2/blob/refactor/emote-providers/src/controllers/emotes/EmoteController.hpp) in the `Application` (the old `singletons/Emotes` + third party emotes). The emote controller holds all providers. Currently, providers are set at startup, but they could be added at runtime in the future. Each channel with third party emotes contains an [`EmoteHolder`](https://github.com/Nerixyz/chatterino2/blob/refactor/emote-providers/src/controllers/emotes/EmoteHolder.hpp). This contains the emote maps for each provider for that channel. It's responsible for (re-)loading them.

Most of the changes to other parts here depend on this. A few more important ones:

- Emote message element flags are only used for links. We now put the provider ID in the emote to resolve a URL.
- Badges defined `homePage` but it was never used in context menus
- Zero-width checking for BTTV should be done in the emote provider.


Nerixyz:
I'll do this in chunks.

PR #6441, opened 2025-09-04
fix(plugin): handle non-string errors gracefully

### LINKED ISSUE #6201: Errors from plugins can panic the whole program
### Checklist

### Describe your issue

Calling `error()` with non-string arguments in Lua can crash Chatterino because we only expect errors to be strings. Throwing numbers and nils is lovely is it not?

### Screenshots

```
PANIC: unprotected error in call to Lua API (stack index 1, expected string, received nil)
...
#3  0x000055555673f694 in luaD_throw (L=0x55555c363128, errcode=2) at lib/lua/src/ldo.c:132
...
#19 0x0000555555ca26a2 in tryCall<std::optional<QString>, sol::basic_table_core<false, sol::basic_reference<false> >&, sol::basic_protected_function<sol::basic_reference<false>, false, sol::basic_reference<false> > > (function=...)
    at src/controllers/plugins/SolTypes.hpp:81
```

### OS and Chatterino Version

Chatterino 2.5.3 DEBUG (commit c702f8fba modified) built with Qt 6.9.0 Running on Arch Linux, kernel: 6.13.8-arch1-1

Nerixyz:
I haven't encountered this in the current version of Chatterino, but there might be hidden cases where a failed function call won't have a string at the top of the stack. Currently, if that's the case, we'd crash, because we try to convert the error result to a `sol::error`. That type expects a string at the top of the stack (which would contain the error message). As explained in the comment, that's not always the case, and we shouldn't crash because of that.

Here's some example code that demonstrates the third and most unimportant case from the comment (but it's the easiest to reproduce):
```lua
ws = c2.WebSocket.new("wss://echo.websocket.org")
ws.on_text = function() error(42) end -- oopsie
ws:send_text("lol")
```

Fixes #6201.

### DIFF
--- a/src/controllers/plugins/SolTypes.cpp
+++ b/src/controllers/plugins/SolTypes.cpp
@@ -12,6 +12,8 @@
 
 namespace chatterino::lua {
 
+using namespace Qt::Literals;
+
 Plugin *ThisPluginState::plugin()
 {
     if (this->plugptr_ != nullptr)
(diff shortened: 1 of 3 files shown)


src/controllers/plugins/SolTypes.cpp
```diff
@@ -27,6 +29,51 @@ Plugin *ThisPluginState::plugin()
...
+
+    auto optString = sol::stack::check_get<QString>(result.lua_state(), -1);
+    if (optString)
+    {
+        return *std::move(optString);
+    }
+
+    // If we get here, the stack didn't contain a string at the top. This is
+    // valid in Lua, but unconventional. Error handlers typically expect a
+    // string at the top of the stack.s
```
pajlada:
```suggestion
    // string at the top of the stack.
```

PR #6446, opened 2025-09-05
feat: spellcheck input

### LINKED ISSUE #16: spellcheck for input

Nerixyz:
Yesterday evening, I wanted to experiment how Hunspell would work in Chatterino. I discovered that it's not that much effort to get spell checking to work.

This is a minimal implementation of spell checking. Dictionaries must be saved as `index.aff` and `index.dic` in `Dictionaries/` (next to `Settings/`). You can find dictionaries [here](https://github.com/wooorm/dictionaries) (this collects some from various sources).

**Features**
- Toggle checking for all inputs in the settings
- Toggle checking for one input via context menu
- Apply suggestions via context menu (even when spellchecking is disabled)
- Works with undo/redo
- Emotes are ignored

**Missing Features** (might be cool to have this in the future)
- Currently, it's bring-your-own-dictionary. It would be cool if Chatterino could install dictionaries.
- You can't switch the language per split right now.
- Custom dictionaries would be great (and "Add to dictionary").
- The spellchecker should be able to reload.

One thing I dislike is that `QSyntaxHighlighter` is synchronous, meaning it reruns every time a character is typed. Technically, you could hack around that by ignoring its requests to highlight a block and manually call `rehighlight` (at some timer tick, for example).

Closes #16.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable ALPHA plugin support in Chatterino" ON)
 option(CHATTERINO_USE_GDI_FONTENGINE "Use the legacy GDI fontengine instead of the new DirectWrite one on Windows (Qt 6.8.0 and later)" ON)
 option(CHATTERINO_ALLOW_PRIVATE_QT_API "Allow uses of Qt's private API - when enabling this, Chatterino must use the EXACT Qt version it was compiled against" OFF)
+option(CHATTERINO_SPELLCHECK "Enable spellchecking in Chatterino (requires Hunspell)" OFF)
 
 option(CHATTERINO_UPDATER "Enable update checks" ON)
 mark_as_advanced(CHATTERINO_UPDATER)
@@ -146,6 +147,10 @@ find_package(Boost REQUIRED OPTIONAL_COMPONENTS headers)
 find_package(OpenSSL REQUIRED)
 
 find_package(Threads REQUIRED)
+
+if(CHATTERINO_SPELLCHECK)
+    find_package(hunspell REQUIRED)
+endif()
 
 find_library(LIBRT rt)
 
--- a/src/controllers/spellcheck/SpellChecker.cpp
+++ b/src/controllers/spellcheck/SpellChecker.cpp
@@ -1,0 +1,209 @@
+#include "controllers/spellcheck/SpellChecker.hpp"
+
+#include "Application.hpp"
+#include "common/QLogging.hpp"
+#include "controllers/accounts/AccountController.hpp"
+#include "messages/Emote.hpp"
+#include "providers/bttv/BttvEmotes.hpp"
+#include "providers/seventv/SeventvEmotes.hpp"
+#include "providers/twitch/TwitchAccount.hpp"
+#include "providers/twitch/TwitchChannel.hpp"
+#include "singletons/Paths.hpp"
+#include "util/FilesystemHelpers.hpp"
+
+#include <QTextCharFormat>
+
+#ifdef CHATTERINO_WITH_SPELLCHECK
+#    include <hunspell/hunspell.hxx>
+#endif
+
+namespace {
+
+using namespace chatterino;
+
+bool shouldIgnore(TwitchChannel *twitch, const QString &word)
+{
+    EmoteName name(word);
+    if (twitch)
+    {
+        if (twitch->bttvEmote(name) || twitch->ffzEmote(name) ||
+            twitch->seventvEmote(name))
+        {
+            return true;
+        }
+        auto locals = twitch->localTwitchEmotes();
+        if (locals->contains(name))
+        {
+            return true;
+        }
+    }
+    if (getApp()->getBttvEmotes()->emote(name) ||
+        getApp()->getFfzEmotes()->emote(name) ||
+        getApp()->getSeventvEmotes()->globalEmote(name))
+    {
+        return true;
+    }
+
+    return getApp()
+        ->getAccounts()
+        ->twitch.getCurrent()
+        ->twitchEmote(name)
+        .has_value();
+}
+
+}  // namespace
+
+namespace chatterino {
+
+#ifdef CHATTERINO_WITH_SPELLCHECK
+class SpellCheckerPrivate
+{
+public:
+    static std::unique_ptr<SpellCheckerPrivate> tryLoad();
+
+    Hunspell hunspell;
+
+private:
+    SpellCheckerPrivate(const char *affpath, const char *dpath);
+};
+
+std::unique_ptr<SpellCheckerPrivate> SpellCheckerPrivate::tryLoad()
+{
+    auto stdPath = qStringToStdPath(getApp()->getPaths().dictionariesDirectory);
+    auto aff = stdPath / "index.aff";
+    auto dic = stdPath / "index.dic";
+    if (!std::filesystem::exists(aff) || !std::filesystem::exists(dic))
+    {
+        qCInfo(chatterinoSpellcheck)
+            << "Failed to find index.aff or index.dic in 'Dictionaries'";
+        return nullptr;
+    }
+    std::error_code ec;
+    auto affCanonical = std::filesystem::weakly_canonical(aff, ec);
+    if (ec)
+    {
+        qCInfo(chatterinoSpellcheck)
+            << "Failed to canonicalize" << stdPathToQString(aff)
+            << "error:" << ec.message();
+        return nullptr;
+    }
+    auto dicCanonical = std::filesystem::weakly_canonical(dic, ec);
+    if (ec)
+    {
+        qCInfo(chatterinoSpellcheck)
+            << "Failed to canonicalize" << stdPathToQString(dic)
+            << "error:" << ec.message();
+        return nullptr;
+    }
+
+    return std::unique_ptr<SpellCheckerPrivate>{new SpellCheckerPrivate(
+        affCanonical.string().c_str(), dicCanonical.string().c_str())};
+}
+
+SpellCheckerPrivate::SpellCheckerPrivate(const char *affpath, const char *dpath)
+    : hunspell(affpath, dpath)
+{
+}
+
+SpellChecker::SpellChecker()
+    : private_(SpellCheckerPrivate::tryLoad())
+{
+}
+#else
+class SpellCheckerPrivate
+{
+};
+SpellChecker::SpellChecker() = default;
+#endif
+
+SpellChecker::~SpellChecker() = default;
+
+bool SpellChecker::isLoaded() const
+{
+    return this->private_ != nullptr;
+}
+
+void SpellChecker::reload()
+{
+#ifdef CHATTERINO_WITH_SPELLCHECK
+    this->private_ = SpellCheckerPrivate::tryLoad();
+#endif
+}
+
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+bool SpellChecker::check(QStringView word)
+{
+#ifdef CHATTERINO_WITH_SPELLCHECK
+    if (!this->private_)
+    {
+        return true;
+    }
+
+    return this->private_->hunspell.spell(word.toUtf8().toStdString());
+#else
+    (void)word;
+    return true;
+#endif
+}
+
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::vector<std::string> SpellChecker::suggestions(const QString &word)
+{
+#ifdef CHATTERINO_WITH_SPELLCHECK
+    if (!this->private_)
+    {
+        return {};
+    }
+
+    auto stdWord = word.toStdString();
+    if (this->private_->hunspell.spell(stdWord))
+    {
+        return {};
+    }
+
+    return this->private_->hunspell.suggest(stdWord);
+#else
+    (void)word;
+    return {};
+#endif
+}
+
+SpellCheckHighlighter::SpellCheckHighlighter(QObject *parent)
+    : QSyntaxHighlighter(parent)
+    , wordRegex(R"(\p{L}+)",
+                QRegularExpression::PatternOption::UseUnicodePropertiesOption)
+{
+    this->spellFmt.setUnderlineStyle(QTextCharFormat::SpellCheckUnderline);
+    this->spellFmt.setUnderlineColor(Qt::red);
+}
+
+void SpellCheckHighlighter::setTwitchChannel(TwitchChannel *channel)
+{
+    this->channel = channel;
+    this->rehighlight();
+}
+
+void SpellCheckHighlighter::highlightBlock(const QString &text)
+{
+    auto *spellChecker = getApp()->getSpellChecker();
+    if (!spellChecker->isLoaded())
+    {
+        return;
+    }
+
+    QStringView textView = text;
+    auto it = this->wordRegex.globalMatchView(textView);
+    while (it.hasNext())
+    {
+        auto match = it.next();
+        auto text = match.capturedView();
+        if (!shouldIgnore(this->channel, text.toString()) &&
+            !spellChecker->check(text))
+        {
+            this->setFormat(static_cast<int>(text.data() - textView.data()),
+                            static_cast<int>(text.size()), this->spellFmt);
+        }
+    }
+}
+
+}  // namespace chatterino
--- a/src/controllers/spellcheck/SpellChecker.hpp
+++ b/src/controllers/spellcheck/SpellChecker.hpp
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <QRegularExpression>
+#include <QSyntaxHighlighter>
+
+#include <memory>
+#include <string>
+#include <vector>
+
+class QTextDocument;
+
+namespace chatterino {
+
+class TwitchChannel;
+
+class SpellCheckerPrivate;
+class SpellChecker
+{
+public:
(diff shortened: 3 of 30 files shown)


Mm2PL:
5 wrz 2025 18:59:05 nerix ***@***.***>:

> One thing I dislike is that *QSyntaxHighlighter* is synchronous, meaning it reruns every time a character is typed. Technically, you could hack around that by ignoring its requests to highlight a block and manually call *rehighlight* (at some timer tick, for example).
That doesn't seem all that bad. In my experience other programs already behave this way. Ideally I would want the timer to be rather low. I believe most programs currently have this rehighlight timer set WAY too high.

Felanbird:
probably needs to ignore the leading `@mention` in replies
image

Mm2PL:
Ideally we should ignore names of people in the chat not just the leading `@username` for replies

pajlada:
I'm fine if this PR doesn't achieve everything we want for a spellcheck input in one go
**Ignoring mentions** seems good for a followup PR
**Running on a timer instead of each key input** seems good for a followup PR

Feature-wise, I feel like we should aim to be only one of "bring your own dictionary" or "supply dictionaries" to start.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added spell checking to the chat input. (#6446)
```
fourtf: It could be noted that this is experimental. Later this text can be changed.

```suggestion
- Dev: Started work on spell checking. (#6446)
```

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#ifdef CHATTERINO_WITH_SPELLCHECK
+#    include <hunspell/hunspell.hxx>
+#endif
+
+namespace {
+
+using namespace chatterino;
+
+bool shouldIgnore(TwitchChannel *twitch, const QString &word)
```
fourtf: I feel like this could be moved into the `Channel` class maybe and then overwritten in the `TwitchChannel` class? Then other channel types could implement their own logic as well.

src/widgets/splits/Split.cpp
```diff
@@ -894,6 +894,16 @@ bool Split::getModerationMode() const
     return this->moderationMode_;
 }
 
+std::optional<bool> Split::checkSpellingOverride() const
```
fourtf, line 897: I don't really get why this would be a setting for a split. Most UI settings are unavailable at the split level.

src/widgets/splits/SplitInput.cpp
```diff
@@ -72,14 +73,27 @@ SplitInput::SplitInput(QWidget *parent, Split *_chatWidget,
         new QCompleter(this->split_->getChannel()->completionModel);
     this->ui_.textEdit->setCompleter(completer);
 
+    // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer)
+    this->spellcheckHighlighter = new SpellCheckHighlighter(this);
+    this->spellcheckHighlighter->setTwitchChannel(
+        dynamic_cast<TwitchChannel *>(this->split_->getChannel().get()));
```
fourtf: Handling raw pointers is a bit crash-y. Couldn't we give it a shared pointer instead?

src/widgets/splits/Split.cpp
Nerixyz: I added this because I want to selectively enable spellchecking for a few splits only.

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -0,0 +1,219 @@
...
+void SpellCheckHighlighter::setChannel(const std::shared_ptr<Channel> &channel)
+{
+    auto twitch = std::dynamic_pointer_cast<TwitchChannel>(channel);
+    this->channel = twitch;
+    this->rehighlight();
+}
+
+void SpellCheckHighlighter::highlightBlock(const QString &text)
+{
+    auto *spellChecker = getApp()->getSpellChecker();
```
pajlada: It would be nice if SpellCheckHighlighter took the SpellChecker class as a constructor argument instead so it doesn't have to "call out" to `getApp`. I know we have this pattern everywhere, but it's nice to avoid where possible, and this feels like an easy one.

src/controllers/spellcheck/SpellChecker.hpp
```diff
@@ -0,0 +1,50 @@
...
+    void reload();
+
+    bool check(const QString &word);
+    std::vector<std::string> suggestions(const QString &word);
+
+private:
+    std::unique_ptr<SpellCheckerPrivate> private_;
+};
+
+class SpellCheckHighlighter : public QSyntaxHighlighter
```
pajlada: This class should probably live in its own file, but you don't have to do this in this PR if you don't want to.

src/singletons/Settings.hpp
```diff
@@ -349,6 +349,8 @@ class Settings
         false,
     };
 
+    BoolSetting enableSpellChecking = {"/behaviour/spellChecking", false};
```
pajlada: nit: store this under `/behaviour/spellChecked/enabled` instead

src/widgets/settingspages/AboutPage.cpp
```diff
@@ -144,6 +144,10 @@ AboutPage::AboutPage()
             addLicense(form.getElement(), "Unicode",
                        "https://www.unicode.org/copyright.html",
                        ":/licenses/unicode.txt");
+#ifdef CHATTERINO_WITH_SPELLCHECK
+            addLicense(form.getElement(), "Hunsppell",
```
pajlada:
```suggestion
            addLicense(form.getElement(), "Hunspell",
```

src/widgets/splits/SplitInput.cpp
```diff
@@ -72,14 +73,26 @@ SplitInput::SplitInput(QWidget *parent, Split *_chatWidget,
...
+    // NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer)
+    this->spellcheckHighlighter = new SpellCheckHighlighter(this);
+    this->spellcheckHighlighter->setChannel(this->split_->getChannel());
+
     this->signalHolder_.managedConnect(this->split_->channelChanged, [this] {
         auto channel = this->split_->getChannel();
         auto *completer = new QCompleter(channel->completionModel);
         this->ui_.textEdit->setCompleter(completer);
+        this->spellcheckHighlighter->setChannel(this->split_->getChannel());
+        this->spellcheckHighlighter->setChannel(this->split_->getChannel());
```
pajlada: You are calling `setChannel` twice here

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -0,0 +1,127 @@
...
+std::unique_ptr<SpellCheckerPrivate> SpellCheckerPrivate::tryLoad()
+{
+    auto stdPath = qStringToStdPath(getApp()->getPaths().dictionariesDirectory);
+    auto aff = stdPath / "index.aff";
+    auto dic = stdPath / "index.dic";
+    if (!std::filesystem::exists(aff) || !std::filesystem::exists(dic))
+    {
+        qCInfo(chatterinoSpellcheck)
+            << "Failed to find index.aff or index.dic in 'Dictionaries'";
+        return nullptr;
```
Mm2PL, line 35: I don't like this as a permanent error handler. This can stay while we work on the feature but digging through the logs to find that your dictionaries are not installed correctly is bad.

src/widgets/splits/InputHighlighter.hpp
```diff
@@ -0,0 +1,43 @@
...
+class QTextDocument;
+
+namespace chatterino {
+
+class Channel;
+class TwitchChannel;
+class SpellChecker;
+
+/// This highlights the text in the split input.
+/// Currently, it only does syntax highlighting.
```
Mm2PL: This comment doesn't really make sense to me. Where is the syntax highlighting

pajlada:
For followup PRs, https://github.com/wooorm/dictionaries seems to contain a bunch of dictionaries with their licenses that we could potentially embed.

Mm2PL:
> For followup PRs, https://github.com/wooorm/dictionaries

Is that what arch Linux uses for hunspell* packages?

Nerixyz:
> For followup PRs, [wooorm/dictionaries](https://github.com/wooorm/dictionaries?rgh-link-date=2025-12-30T13%3A49%3A01Z) seems to contain a bunch of dictionaries with their licenses that we could potentially embed.

I also had this in the PR description, but I should mention that on Linux and FreeBSD (maybe also on macOS?), we'd want to search in the system directories.

src/widgets/splits/InputHighlighter.cpp
```diff
@@ -0,0 +1,103 @@
...
+        {
+            return true;
+        }
+        auto locals = twitch->localTwitchEmotes();
+        if (locals->contains(name))
+        {
+            return true;
+        }
+
+        if (twitch->accessChatters()->contains(word))
```
Mm2PL, line 36: This function shouldn't be called `isEmote` if it checks for chatters

src/widgets/splits/InputHighlighter.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+}  // namespace
+
+namespace chatterino {
+
+InputHighlighter::InputHighlighter(SpellChecker &spellChecker, QObject *parent)
+    : QSyntaxHighlighter(parent)
+    , spellChecker(spellChecker)
+    , wordRegex(R"(\p{L}+)",
+                QRegularExpression::PatternOption::UseUnicodePropertiesOption)
```
Mm2PL: This doesn't account for the existence of contractions, so `doesn't` will highlight as a misspelling. This also means that user names with numbers or underscores in them will be multiple words.

pajlada:
@Mm2PL Do you want a chance to test/review? Otherwise I'm happy to merge this one in as is so we can continue iterating on the feature

Mm2PL:
I have a couple comments pending. I'll publish them today or tomorrow

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+private:
+    SpellCheckerPrivate(const char *affpath, const char *dpath);
+};
+
+std::unique_ptr<SpellCheckerPrivate> SpellCheckerPrivate::tryLoad()
+{
+    auto stdPath = qStringToStdPath(getApp()->getPaths().dictionariesDirectory);
+    auto aff = stdPath / "index.aff";
+    auto dic = stdPath / "index.dic";
```
Mm2PL, line 30: It's worth noting that hunspell dictionaries are usually named like `en_GB.{aff,dic}`.

CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ option(CHATTERINO_LTO "Enable LTO for all targets" OFF)
 option(CHATTERINO_PLUGINS "Enable ALPHA plugin support in Chatterino" ON)
 option(CHATTERINO_USE_GDI_FONTENGINE "Use the legacy GDI fontengine instead of the new DirectWrite one on Windows (Qt 6.8.0 and later)" ON)
 option(CHATTERINO_ALLOW_PRIVATE_QT_API "Allow uses of Qt's private API - when enabling this, Chatterino must use the EXACT Qt version it was compiled against" OFF)
+option(CHATTERINO_SPELLCHECK "Enable spellchecking in Chatterino (requires Hunspell)" OFF)
```
Mm2PL, line 34: For the future: I want this default to ON once this is usable for end users.

CHANGELOG.md
```diff
@@ -71,6 +71,8 @@
 - Dev: Unwrapped `LimitedQueueSnapshot` to `std::vector`. (#6606)
 - Dev: Simplified uses of `getMessageSnapshot`. (#6607)
 - Dev: Disabled `llvm-prefer-static-over-anonymous-namespace` in clang-tidy. (#6610)
+- Dev: Added options to close multiple visible tabs. (#6515)
```
Mm2PL: unrelated change?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -481,6 +481,20 @@ void GeneralPage::initLayout(GeneralPageView &layout)
             "yours is successfully sent in the matching channel.")
         ->addTo(layout);
 
+#ifdef CHATTERINO_WITH_SPELLCHECK
+    SettingWidget::checkbox("Check spelling (experimental)",
```
Mm2PL: This setting feels like it should be needed for Chatterino to do any spell checking but that is not correct. The idea of enabling it for a few splits only should probably be better articulated in the tooltip

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -481,6 +481,20 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                            s.enableSpellChecking)
+        ->setTooltip(
+            u"Check the spelling of words in the input box. Chatterino does "
+            u"not "
+            "include dictionaries - they have to be downloaded or created "
+            "manually. Chatterino expects Hunspell dictionaries in '" %
+            getApp()->getPaths().dictionariesDirectory %
+            u"'. The file index.aff has to contain the affixes and index.dic "
+            u"must contain the dictionary (subject to change).")
```
Mm2PL: Please make most of this a label with a clickable directory link. You could point to the website you suggested to download dictionaries from.

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -0,0 +1,127 @@
...
+#endif
+
+SpellChecker::~SpellChecker() = default;
+
+bool SpellChecker::isLoaded() const
+{
+    return this->private_ != nullptr;
+}
+
+void SpellChecker::reload()
```
Mm2PL: This is unused

Mm2PL (changes requested):
It's mostly nitpicks. The one comment I really want resolved in this PR is the word regex one, all of the rest are pretty much nice-to-haves

src/controllers/spellcheck/SpellChecker.cpp
pajlada: ack, I imagine the UI for this feature will make this apparent. This is good for now

src/widgets/splits/InputHighlighter.hpp
pajlada: I interpret syntax highlighting as the red underline on misspelled words

src/controllers/spellcheck/SpellChecker.cpp
pajlada: ack, this is also something I imagine will change as the feature evolves (e.g. by loading all aff/dic pairs in the dictionaries directory), but this is preferred for now

CMakeLists.txt
pajlada: ack, we should treat this just like we did plugins. disable by default for now, then start enabling once the feature is well implemented for the platform

src/widgets/settingspages/GeneralPage.cpp
pajlada: Spellchecking should have its own little category later, similar to Streamlink. I'm fine with making that change later

src/widgets/settingspages/GeneralPage.cpp
pajlada: Not important for this PR imo, we'll be changing how the spellchecker is configured in the settings dialog

src/widgets/settingspages/GeneralPage.cpp
Nerixyz: I'll wait with this until we figure out how multiple dictionaries are loaded and how that's presented to the user.

src/widgets/settingspages/GeneralPage.cpp
Nerixyz: Added "by default". But yea, that mechanism might change as we iterate on the feature.

src/controllers/spellcheck/SpellChecker.cpp
Nerixyz: I only want to support one dictionary with this PR. Having this named `en_GB` would be wrong if the actual dictionary wasn't English.

src/widgets/splits/InputHighlighter.cpp
Nerixyz: It's now `\p{L}(?:\P{Z}+\p{L}+)*`, which is better but still not enough. We have to filter URLs as well (could use the link parser or some simple heuristic for this).

Nerixyz:
For future work: I don't have any specific area I want to focus on first. So if anyone wants to fix/add something, feel free to do so and open a PR.

mjeveritt:
Thanks for your work on this! :heart_eyes: looking forward to the [nightly?] build ... :crossed_fingers: 
_(hint: please tag here!)_

Agreed that on-trigger being every character is a pain, and ALSO that the timers being WAAY too high in other apps as the 'pendulum' invariably swings the other (opposite) way. Hopefully we can [all] find a decent balance (configurable, ever?!) so that we can all enjoy this vital tool. :abc: :wrench: 

Hope it doesn't reach a point that we need/have AI integration .. but at least this app is mostly immune from that *** for now! :laughing:

pajlada:
@mjeveritt 
Yeah the nightly build from https://github.com/Chatterino/chatterino2/releases/tag/nightly-build will have spellchecking enabled once it's done building (~20-30 minutes from now)

If after you've used it you have some feedback, feel free to open an issue or discussion about it (whichever you deem most fitting for the feedback)

PR #6447, opened 2025-09-06
fix: relayout tabs when update button visibility changes

pajlada:
CC @Wissididom if you could try this out - this is what i added to ensure the client actually responded yes to if there are updates

You can use `/debug-test update-check` to manually check for updates

```patch
diff --git a/src/common/Version.hpp b/src/common/Version.hpp
index faac0110d..6e77e7cd6 100644
--- a/src/common/Version.hpp
+++ b/src/common/Version.hpp
@@ -29,7 +29,7 @@ namespace chatterino {
  *  - 2.4.0-alpha.2
  *  - 2.4.0-alpha
  **/
-inline const QString CHATTERINO_VERSION = QStringLiteral("2.5.4-beta.1");
+inline const QString CHATTERINO_VERSION = QStringLiteral("2.5.4-beta.0");
 
 class Version
 {
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bugfix: Fixed crashes that could occur when Lua functions errored with values other than strings. (#6441)
 - Bugfix: Fixed zero-width global BTTV emotes not showing in the `:~` completions. (#6440)
+- Bugfix: Fixed an issue where the update button would be unclickable on macOS and Linux. (#6447)
 
 ## 2.5.4-beta.1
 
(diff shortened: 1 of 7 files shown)


Wissididom:
Works for me. Basically I didn't have to do anything to get it to show, because I'm using the master (beta) version that apparently isn't the latest version, but I also didn't have beta versions enabled.
Before:

https://github.com/user-attachments/assets/21f97404-bd5f-41cf-930b-c3ada9c7b6c8

After:

https://github.com/user-attachments/assets/6588f969-b2bd-41c0-8a20-c901a1da2c55

pajlada:
image

PR #6451, opened 2025-09-06
fix: mouse over tooltips

2547techno:
On wayland, the tooltip sometimes gets positioned in a way where the mouse is over both the tooltip and the emote/link. This causes some buggy behaviour where the tooltip is only displayed momentarily.

Adding the `Qt::WA_TransparentForMouseEvents` attribute to the tooltip widget seems to fix this.

Before: 
https://github.com/user-attachments/assets/9eee03e5-abe9-4ec5-8d88-42183d638513

After:

https://github.com/user-attachments/assets/1508a937-010a-442d-88b7-08b098f2fd37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Bugfix: Fixed crashes that could occur when Lua functions errored with values other than strings. (#6441)
 - Bugfix: Fixed zero-width global BTTV emotes not showing in the `:~` completions. (#6440)
 - Bugfix: Fixed an issue where the update button would be unclickable on macOS and Linux. (#6447)
+- Bugfix: Fixed flickering tooltips on Wayland when the mouse cursor is over them. (#6451)
 
 ## 2.5.4-beta.1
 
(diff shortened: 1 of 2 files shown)


Nerixyz (approved):
I wonder why we never set this in the first place.

Works on Windows (and it was broken there before, too). I couldn't replicate your exact issue there - I had to move the window to the bottom of the screen, but not placed at any corner. In the bottom right corner, Chatterino should try to place the tooltip above the cursor. Not sure why this isn't happening on Wayland (maybe because they gatekeep the global positions?).

pajlada (approved):
This works great on my setup (sway 1.11 on Arch Linux btw)

pajlada:
Also works great on KDE Plasma on Fedora, although it wasn't really a problem to begin with there.

2547techno:
> Also works great on KDE Plasma on Fedora, although it wasn't really a problem to begin with there.

hmm, i'm not sure why your tooltips were positioned correctly on plasma on fedora, the attached videos are also plasma but on nixos (+1 confirmed on the same setup). Not sure if knowing that will help figure out why tooltip positions screw up but I thought I'd mention it anyways.

pajlada:
I was running a somewhat old version of plasma to be fair, fedora 42. Either way this change is great. Thank you @2547techno

2547techno:
credit goes to `JOEioX` on twitch, he proposed the solution only minutes after i complained about it :)

PR #6452, opened 2025-09-06
fix(plugins): Log potential errors in HTTP request callbacks

### LINKED ISSUE #6224: Lua HTTPRequest fails to report callbacks' errors
### Checklist

### Describe your issue

When HTTPRequest's on_error() callback errors there is no information that it did. This makes debugging a bit nightmarish.

### Screenshots

```lua
c2.register_command('/http', function()
    local r = c2.HTTPRequest.create(c2.HTTPMethod.Get, "http://localhost:8080")
    r:set_timeout(100)
    r:on_success(function(result)
        print('success callback')
        error('failed success')
    end)
    r:on_error(function(result)
        print('error callback')
        error('failed to fail')
    end)
    r:finally(function()
        print('finally callback')
        error('finally failed')
    end)
    r:execute()
end)
```

Notice that running this does not produce any feedback about errors in callbacks.

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 11 Version 24H2, kernel: 10.0.26100

Nerixyz:
When an error occurred in a handler for an HTTP request, it wasn't printed to the console or the REPL (linked issue).
With this PR, the error is printed.

To test this, you can try the code from https://github.com/Chatterino/chatterino2/issues/6224#issuecomment-2901860640 - it should only print "finally" in the chat. It should (with this PR) print an error like `chatterino.lua: [test:Test plugin] HTTPRequest::on_error - (path): attempt to concatenate a nil value` in both the Chatterino log as well the REPL. Before this, there was no output.

Fixes #6224.

### DIFF
--- a/src/controllers/plugins/api/HTTPRequest.cpp
+++ b/src/controllers/plugins/api/HTTPRequest.cpp
@@ -115,10 +115,9 @@ void HTTPRequest::execute(sol::this_state L)
     auto hack = this->weak_from_this();
     auto *pl = getApp()->getPlugins()->getPluginByStatePtr(L);
     pl->httpRequests.push_back(this->shared_from_this());
-    auto *mainState = pl->state().lua_state();
 
     std::move(this->req_)
-        .onSuccess([hack](const NetworkResult &res) {
+        .onSuccess([pl, hack](const NetworkResult &res) {
             auto self = hack.lock();
             if (!self)
             {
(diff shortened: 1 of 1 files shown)


src/controllers/plugins/api/HTTPRequest.cpp
```diff
@@ -141,17 +142,17 @@ void HTTPRequest::execute(sol::this_state L)
             {
                 return;
             }
-            (*self->cbError)(HTTPResponse(res));
+            loggedVoidCall(*self->cbError, u"HTTPRequest::on_error", pl,
+                           HTTPResponse(res));
             self->cbError = std::nullopt;
         })
-        .finally([mainState, hack]() {
+        .finally([pl, hack]() {
```
Nerixyz, line 149: Note: we can just pass the plugin pointer here, because the plugin owns `hack` (see line 117).

I think, in the future, we could have some more generic plugin handle that we could store here (the plugin shouldn't need to know that HTTP requests exist). We can't put the plugin in a `std::shared_ptr` itself, but we could make the plugin own some shared pointer. That's for the next release, though.

pajlada:
CC @Mm2PL

Mm2PL (approved):
LGTM

PR #6453, opened 2025-09-07
fix: notebook custom button resizing/layout

pajlada:
The big change here is that we use `isHidden` instead of `!isVisible` for the custom button visibility test.
The former returns true if it _will_ be shown when the Notebook is shown, where the latter returns true only if the Notebook is currently shown.

Using `isHidden` means we can always perform the layout safely, regardless of the state of the Notebook.

As a bonus, I've removed the layout that happens whenever a custom button is added, and instead now perform a single un-animated layout when _all_ custom buttons have been added.

With these changes, I can no longer reproduce any of the custom buttons being hidden behind tabs. I have tested all tab layouts.


pajlada:
image

PR #6456, opened 2025-09-08
fix(plugins): Implement `__eq` for `Channel`

Nerixyz:
`c2.Channel.by_name('user') == c2.Channel.by_name('user')` was always false, even though the metatable had an `__eq`. We can manually add `operator==` which sol will pick up. This PR adds the operator.


Mm2PL (approved):

src/util/WeakPtrHelpers.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
pajlada, line 1: Good file

pajlada (approved):

PR #6462, opened 2025-09-14
fix: update input fonts on fontChanged signal

Obertura777:
I saw [the TODO on message length counter font not updating](https://github.com/Chatterino/chatterino2/blob/c7dea40b6baa9f401cc5559723b103edfb75741e/src/widgets/splits/SplitInput.cpp#L278C1-L278C77) on font change and thought this would be a simple fix. Now the `fontChanged` signal is connected to update the `textEditLength` widget's font. I've also removed the TODO comments.

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -217,6 +217,8 @@ void SplitInput::initLayout()
         app->getFonts()->fontChanged, [=, this]() {
             this->ui_.textEdit->setFont(
                 app->getFonts()->getFont(FontStyle::ChatMedium, this->scale()));
+            this->ui_.textEditLength->setFont(
+                app->getFonts()->getFont(FontStyle::TimestampMedium, this->scale()));
             this->ui_.replyLabel->setFont(app->getFonts()->getFont(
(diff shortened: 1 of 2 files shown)


pajlada (approved):

src/widgets/splits/SplitInput.cpp
```diff
@@ -217,6 +217,8 @@ void SplitInput::initLayout()
         app->getFonts()->fontChanged, [=, this]() {
             this->ui_.textEdit->setFont(
                 app->getFonts()->getFont(FontStyle::ChatMedium, this->scale()));
+            this->ui_.textEditLength->setFont(app->getFonts()->getFont(
+                FontStyle::TimestampMedium, this->scale()));
             this->ui_.replyLabel->setFont(app->getFonts()->getFont(
                 FontStyle::ChatMediumBold, this->scale()));
```
Nerixyz, line 223: This is basically the same code as in `scaleChangedEvent`. We should probably have some `updateFonts` method to deduplicate. Doesn't have to be in this PR.

Obertura777: sure. I might make another PR after this one gets merged.

PR #6465, opened 2025-09-15
fix: log deletion messages to file log

### LINKED ISSUE #6464: Logging not recording message deleted
### Checklist

### Describe your issue

In the logs for chat, it records an empty line when the "A message from <user> was deleted: <message>" shows in chatterino. If you are a moderator it shows the log from the moderator action, but the one that can be seen by everyone does not log properly.

### Screenshots

Image

Image

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 10 Version 22H2, kernel: 10.0.19045

jupjohn:
This PR aims to fix an issue where "A message from X was deleted" messages aren't being logged to chatterino's file logs. This was caused by the text being used by the logger (`Message.messageText`) not being set, so an empty string was being logged. I've also set the `searchText` so they show up in searches.

Fixes #6464.

Post-fix screenshot of log:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Bugfix: Fixed flickering tooltips on Wayland when the mouse cursor is over them. (#6451)
 - Bugfix: Fixed `c2.Channel` comparing `false` to the same channel. (#6456)
 - Bugfix: Fixed an issue where the moderation icon was missing from the Moderation tab. (#6457)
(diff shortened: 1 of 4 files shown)


Nerixyz (commented):
This also fixes the messages not being searchable 👍 

Looks like, we don't have a snapshot test for deletion messages from IRC yet. Could you add one (or maybe two: with and without truncation)?

Nerixyz (approved):
👍 Thank you!

PR #6469, opened 2025-09-16
fix(macos): check selection in cmd+backspace

Nerixyz:
When invoking the cmd+backspace handler, we didn't check if a selection existed and always changed the existing one. With this PR, the selection is only made if none exists.

Fixes #6468.


pajlada:
@jacob-thompson Could you verify that this makes sense?

jacob-thompson:
yes, this is how it should behave

pajlada (approved):
Tested this and this now mostly aligns with the macOS Notes app

`foo <SELECTION>bar</SELECTION> baz` and pressing CMD+Backspace results in:
 - chatterino: `foo  baz` (note the double space)
 - notes app: `foo baz` (note the single space)

I don't believe this is something we should aim to fix, as it feels like a notes-specific thing.

PR #6471, opened 2025-09-18
feat: update emoji data to Unicode 17.0

Nerixyz:
Again, like with 16.0, this uses my [feature branch](https://redirect.github.com/iamcal/emoji-data/pull/255), because I don't think that will get merged any time soon. This also runs into https://github.com/twitchdev/issues/issues/1104, so it's not worth merging before the next release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -432,7 +432,7 @@ jobs:
         shell: bash
 
       - name: Create release
-        uses: ncipollo/release-action@v1.20.0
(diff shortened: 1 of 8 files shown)


pajlada:
Marking as a draft since we are pending emoji-data to update, which is waiting for Apple's release of unicode 17.0 emojis

Nerixyz:
This still uses the data from my branch, because it's not merged yet, but Apple images are now available.

PR #6473, opened 2025-09-20
Refactor font update logic in SplitInput

Obertura777:
Following [this comment](https://github.com/Chatterino/chatterino2/pull/6462#discussion_r2347576216) in #6462, I've created a `updateFonts()` function for code deduplication.

### DIFF
--- a/src/widgets/splits/SplitInput.cpp
+++ b/src/widgets/splits/SplitInput.cpp
@@ -213,15 +213,10 @@ void SplitInput::initLayout()
     QObject::connect(this->ui_.textEdit, &QTextEdit::textChanged, this,
                      &SplitInput::onTextChanged);
 
-    this->managedConnections_.managedConnect(
-        app->getFonts()->fontChanged, [=, this]() {
-            this->ui_.textEdit->setFont(
-                app->getFonts()->getFont(FontStyle::ChatMedium, this->scale()));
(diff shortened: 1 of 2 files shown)


src/widgets/splits/SplitInput.hpp
```diff
@@ -188,6 +188,9 @@ class SplitInput : public BaseWidget
 
     QPropertyAnimation backgroundColorAnimation;
 
+private:
```
Nerixyz: Hmm, this `private` should be a bit higher (after the last `override`). Looks like this was never added before and everything is protected now. Would be fine if you leave it out too.

src/widgets/splits/SplitInput.cpp
```diff
@@ -277,15 +272,7 @@ void SplitInput::scaleChangedEvent(float scale)
             this->ui_.vbox->setSpacing(this->marginForTheme());
         }
     }
-    this->ui_.textEdit->setFont(
-        app->getFonts()->getFont(FontStyle::ChatMedium, scale));
-
-    // NOTE: We're using TimestampMedium here to get a font that uses the tnum font feature,
-    // meaning numbers get equal width & don't bounce around while the user is typing.
```
Nerixyz, line 284: Keep this comment in the new function.

pajlada (approved):

pajlada:

Issue #5027, opened 2023-12-15
Showing only live tabs also hides utility tabs


PretzelVector:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When tab visibility is set to only "Only live tabs", tab types other than channel are also hidden.

While I understand these tabs are not *technically* live, they will never be live and their utility does not change based on live-ness.

It would be nice to have `/live` and `/mentions` in addition to live tabs.

### OS and Chatterino Version

Chatterino 2.4.6 (commit GIT-REPOSITORY-NOT-FOUND) built with Qt 5.15.11 Running on NixOS 24.05 (Uakari), kernel: 6.6.6

Felanbird:
Implementing this within the current solution seems to be assumed as not-easy https://github.com/Chatterino/chatterino2/pull/4358#issuecomment-1591808977

x9136:
The only reason why it is inconvenient to leave only Live tabs.

x9136:
> to be assumed as not-easy https://github.com/Chatterino/chatterino2/pull/4358#issuecomment-1591808977

I don't quite understand why it turned out that the pinned tabs feature needs to be done through `isLive`.

But since this is how everything works now, can you tell me if it is possible to add the "Mark as Live" item to the tab menu in order to manually "pin" some tabs so that they are always visible?

PR #6480, opened 2025-09-22
Add tab visibility options: only unread tabs, live + unread tabs

brilliantdrink:
This is my draft for adding options to show only unread tabs and live + unread tabs.

It changes how tab visibility is stored, from incremental enum values (0, 1, 2) to powers of two to use as a bit field. This allows combination of multiple "x only" options. Im thinking that I might be good to continue that logic to the input fields and convert it from a select menu to checkboxes.

closes #6475


brilliantdrink:
Currently, there are no tests for tab visibility afaics. Creating those could be within the scope of this pr aswell.

src/widgets/helper/NotebookTab.hpp
```diff
@@ -71,7 +71,7 @@ class NotebookTab : public Button
      *
      * Obeys the HighlightsEnabled setting and the highlight state hierarchy and tracks the highlight state update sources
      */
-    void updateHighlightState(HighlightState style,
+    bool updateHighlightState(HighlightState style,
```
pajlada, line 80: Update documentation with what the return value is meant to reflect

src/widgets/splits/SplitContainer.cpp
```diff
@@ -222,8 +222,13 @@ void SplitContainer::addSplit(Split *split)
                          [this, split](HighlightState state) {
                              if (this->tab_ != nullptr)
                              {
-                                 this->tab_->updateHighlightState(
-                                     state, split->getChannelView());
+                                 if (this->tab_->updateHighlightState(state, split->getChannelView()))
+                                 {
+                                     if (auto *notebook = dynamic_cast<Notebook *>(this->parentWidget()))
```
pajlada, line 234: Add a comment here - I imagine it's something along the lines of
"The highlight state of this tab changed, force refresh the notebook to ensure the tabs visibilities are recalculated"

src/widgets/Notebook.cpp
```diff
@@ -1325,17 +1325,19 @@ SplitNotebook::SplitNotebook(Window *parent)
...
-                case NotebookTabVisibility::LiveOnly:
-                    this->setTabVisibilityFilter([](const NotebookTab *tab) {
-                        return tab->isLive();
-                    });
-                    break;
-                case NotebookTabVisibility::AllTabs:
-                default:
-                    this->setTabVisibilityFilter(nullptr);
-                    break;
+            if ((visibility & (NotebookTabVisibility::LiveOnly | NotebookTabVisibility::UnreadOnly)) != 0) {
```
pajlada: You could make use of our `FlagsEnum` class to make the comparisons a little bit more readable.

Additionally, if possible, I'd prefer if the actual visibility-callbacks for these states stay as simple as possible. Meaning if someone has set the filter to "live only", we should be able to use the `{ return tab->isLive(); } ` callback
If someone has set the filter to "unread only" we should be able to use the `{ return tab->highlightState() != HighlightState::None; }` callback

a lot of yapping, but move as much of the logic out from the callback as possible, since it's run many times compared to the setting changing which is called less often

pajlada (changes requested):
UX-wise, I think right-clicking the tab menu would work well with a menu that looks like this:
Show all tabs (toggles all 3 checkboxes)
Hide all tabs (untoggles all 3 checkboxes)

---------------------

[ ] Show live tabs
[ ] Show unread tabs
[ ] Show other tabs

Not sure - happy to bikeshed with the UX


For comparison, this is what the menu looks like now:
( ) Show all tabs
( ) Only show live tabs
( ) Hide all tabs

8thony:
> UX-wise, I think right-clicking the tab menu would work well with a menu that looks like this:
> Show all tabs (toggles all 3 checkboxes)
> Hide all tabs (untoggles all 3 checkboxes)

I just want to throw in that there was another request in #5027 with tab visibility "live + utility tabs".
Not sure how that's going to fit in the UX as well.

brilliantdrink:
Took a while, but here are some updates. I've updated the visibility submenu in the tab menu to look like this

Bildschirmfoto 2026-01-12 um 19 41 26

Selecting which tabs are shown feels to me like a filtering option, which is why i would opt not to add a "other" option to show all tabs, but rather solve it like this: those two options in the middle are not mutually exclusive, you can add and remove them like you would filters. Selecting to show all or to hide all does deselect all other options, though. If you deselect the hide option, it reverts to whatever was chosen before, e.g. if you select both unread and live, then select hide, and then click on "Hide all tabs" again it reverts to have unread and live selected. Deselecting "Show all tabs" does not work atm.

I am seeing a bug still where changing the option inside settings does change the actual behaviour but does not update the list in the tab menu (which seems to be pre-existing but id say is within the scope of this pr)

Also I think I have to take a look at `ActionNames.hpp` again

src/widgets/Notebook.cpp
brilliantdrink: Thanks! I've used the `FlagsEnum` in `Notebook.cpp:1240`  and `Notebook.cpp:1314` for my new commit, I am not sure if I used it as intended.

I did attempt to change the type in the `EnumSetting` but it seems its not intended for that, or at least not yet made compatible (it is saying `no type named 'type' in 'std::underlying_type<chatterino::FlagsEnum<chatterino::NotebookTabVisibilityFlag>>'`).


As for the visibility callback, I've changed it so that the flag check is outside the call back and only the tab-related checks are inside. I haven't found a way to make this non-redundant with the combination of flags yet, though.

brilliantdrink:
> > UX-wise, I think right-clicking the tab menu would work well with a menu that looks like this:
> > Show all tabs (toggles all 3 checkboxes)
> > Hide all tabs (untoggles all 3 checkboxes)
> 
> I just want to throw in that there was another request in #5027 with tab visibility "live + utility tabs". Not sure how that's going to fit in the UX as well.

Thanks for bringing that up! Ive had a look at it and it seems harder to implement than it seems.

Afais, Chatterino doesn't really have a concept of "utility tabs", because every tab is just made up of multiple splits. You could check for each tab of any of the splits have a "utility" split, technically, but I would vouch for adding the option to pin single tabs so that they show regardless of the tab visibility setting as x9136 hinted at

brilliantdrink:
Okay, it turns out that not having a "other" option leads to ambiguous states, where "unchecking" `Hide all tabs` or `Show all tabs` may not do what the user expect, so after some more thinking it turns out that pajladas original idea is basically the most straightforward UX-wise imo. This is what it looks like in the tab menu:


The only problem I see now is in the settings page: tab visibility is now is conceptually a multi-select, but qt6 doesn't seem to have a multi select dropdown at all. Listing all permutations only works for two values, like in "Username style", but gets confusing.
I see two options here: either tab visibility becomes more of a section with three checkboxes for the three categories of tabs, or we implement or use a pre-existing combobox with multi select

src/singletons/Settings.hpp
```diff
@@ -199,9 +199,9 @@ class Settings
 
     EnumSetting<NotebookTabLocation> tabDirection = {"/appearance/tabDirection",
                                                      NotebookTabLocation::Top};
-    EnumSetting<NotebookTabVisibility> tabVisibility = {
+    EnumSetting<NotebookTabVisibilityFlag> tabVisibility = {
         "/appearance/tabVisibility",
-        NotebookTabVisibility::AllTabs,
+        static_cast<NotebookTabVisibilityFlag>(-1),
     };
```
pajlada, line 205: This will require a settings migration, so this PR is best left until I'm done with the #6695 which includes a mechanism for migrating settings

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -212,26 +212,66 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                }
+            }
+            label.append(" tabs");
+            return QString::fromStdString(label);
         },
         [](auto args) {
-            if (args.value == "Only live tabs")
+            if (args.value == "Live tabs") {
+                return static_cast<std::underlying_type_t<NotebookTabVisibilityFlag>>(NotebookTabVisibilityFlag::Live);
+            }
```
pajlada, line 251: Try using `std::to_underlying` here instead https://en.cppreference.com/cpp/utility/to_underlying

pajlada (changes requested):

PR #6490, opened 2025-09-25
feat: add markdown support for user notes

### LINKED ISSUE #6319: Markdown support for user notes
### Checklist

### Describe your issue

Add support for Markdown formatting in user notes, for better organization and all.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.3 (commit 2568d8a6e) built with Qt 6.9.1 Running on EndeavourOS, kernel: 6.15.4-arch2-1

jacob-thompson:
fixes #6319

extends the `Label` widget to render markdown for user notes, and adds a toggle for a live markdown preview to the `EditUserNotesDialog`

image

https://github.com/user-attachments/assets/817a31b5-591f-4750-a455-b65f03ac5ce2

### DIFF
--- a/src/widgets/Label.cpp
+++ b/src/widgets/Label.cpp
@@ -242,11 +242,9 @@ void Label::updateSize()
     if (this->markdownEnabled_ && this->markdownDocument_ &&
         !this->text_.isEmpty())
     {
-        // Size based on Markdown document
         this->markdownDocument_->setDefaultFont(
             getApp()->getFonts()->getFont(this->getFontStyle(), this->scale()));
 
-        // Ensure markdown content is set
         this->markdownDocument_->setMarkdown(this->text_);
 
         // Use word wrap width if enabled, otherwise use a reasonable default
@@ -265,7 +263,6 @@ void Label::updateSize()
     }
     else
     {
-        // Original sizing logic
         auto height = metrics.height() + yPadding;
         if (this->shouldElide_)
         {
--- a/src/widgets/Label.cpp
+++ b/src/widgets/Label.cpp
@@ -3,6 +3,7 @@
 #include "Application.hpp"
 
 #include <QPainter>
+#include <QTextDocument>
 
 namespace chatterino {
 
@@ -38,6 +39,10 @@ void Label::setText(const QString &text)
         {
             this->updateElidedText(this->getFontMetrics(),
                                    this->textRect().width());
+        }
+        if (this->markdownEnabled_ && this->markdownDocument_)
+        {
+            this->markdownDocument_->setMarkdown(text);
         }
         this->updateSize();
         this->update();
@@ -84,6 +89,33 @@ void Label::setShouldElide(bool shouldElide)
     this->update();
 }
 
+bool Label::getMarkdownEnabled() const
+{
+    return this->markdownEnabled_;
+}
+
+void Label::setMarkdownEnabled(bool enabled)
+{
+    this->markdownEnabled_ = enabled;
+    if (enabled)
+    {
+        if (!this->markdownDocument_)
+        {
+            this->markdownDocument_ = std::make_unique<QTextDocument>();
+        }
+        if (!this->text_.isEmpty())
+        {
+            this->markdownDocument_->setMarkdown(this->text_);
+        }
+    }
+    else
+    {
+        this->markdownDocument_.reset();
+    }
+    this->updateSize();
+    this->update();
+}
+
 void Label::setFontStyle(FontStyle style)
 {
     this->fontStyle_ = style;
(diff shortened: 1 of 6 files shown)


fourtf:
Hello Jacob, thank you for your submission. It looks technically sound and I like the usage of QTextDocument to handle markdown.

What I fear is that this feature will only be available to very technical users because it's necessary to know markdown to use it. I also fear that the markdown preview is confusing for users as it will likely show the same text as the left side if no markup is added.

To open this feature up for everyone I think it's necessary to hide the fact that we are using markdown under the hood. A graphical way of formatting (a toolbar with buttons for bold, italics, lists, etc.) would be intuitive and inviting. After all, this is a sub-dialog so why not keep the user interface simple there. The document can be loaded and saved as markdown automatically, so it would "just work" for the user.

What do you think about this? It's also possible to go ahead with your suggested GUI for now and reconsider it later.

Mm2PL:
I don't think we need to hide that we're using markdown. I think we 
should take some inspiration from Discord and don't do full WYSIWYG but 
instead do buttons for inserting md syntax and syntax highlighting like:
  - <b>**this text is bold**</b>
  - <em>_this is in italics_</em>
  - ...


W dniu 26.09.2025 o 13:23, fourtf pisze:
> *fourtf* left a comment (Chatterino/chatterino2#6490) <https:// 
> github.com/Chatterino/chatterino2/pull/6490#issuecomment-3338215817>
> 
> Hello Jacob, thank you for your submission. It looks technically sound 
> and I like the usage of QTextDocument to handle markdown.
> 
> What I fear is that this feature will only be available to very 
> technical users because it's necessary to know markdown to use it. I 
> also fear that the markdown preview is confusing for users as it will 
> likely show the same text as the left side if no markup is added.
> 
> To open this feature up for everyone I think it's necessary to hide the 
> fact that we are using markdown under the hood. A graphical way of 
> formatting (a toolbar with buttons for bold, italics, lists, etc.) would 
> be intuitive and inviting. After all, this is a sub-dialog so why not 
> keep the user interface simple there. The document can be loaded and 
> saved as markdown automatically, so it would "just work" for the user.
> 
> What do you think about this? It's also possible to go ahead with your 
> suggested GUI for now and reconsider it later.

fourtf:
Yeah maybe it's fine. The user doesn't have to use the markdown preview if he doesn't understand or want it.

jacob-thompson:
image

resources/buttons/bold-darkMode.svg
Nerixyz, line 1: We usually use (outlined?) fluent icons for new ones.

src/widgets/dialogs/EditUserNotesDialog.cpp
```diff
@@ -17,12 +21,180 @@ EditUserNotesDialog::EditUserNotesDialog(QWidget *parent)
...
+            this->textEdit_->setFocus();
+        }
+        else
+        {
+            cursor.insertText("### ");
+            this->textEdit_->setTextCursor(cursor);
+            this->textEdit_->setFocus();
+        }
+    });
+    headingButton->setToolTip("Insert a heading at the start of the line");
```
Nerixyz: Does this make sure we're at the start of the line?

src/widgets/Label.cpp
```diff
@@ -117,32 +153,60 @@ void Label::paintEvent(QPaintEvent * /*event*/)
...
 
-    qreal width = metrics.horizontalAdvance(text);
-    Qt::Alignment alignment = !this->centered_ || width > textRect.width()
-                                  ? Qt::AlignLeft | Qt::AlignVCenter
-                                  : Qt::AlignCenter;
+        QString colorName = this->palette().windowText().color().name();
+        this->markdownDocument_->setDefaultStyleSheet(
+            QString("body { color: %1; } p { margin: 0; } h1, h2, h3, h4, h5, "
+                    "h6 { margin: 0; }")
+                .arg(colorName));
```
Nerixyz: Do we need this, or can we avoid using style sheets? Either way, this shouldn't be done in the `paintEvent`.

src/widgets/Label.hpp
```diff
@@ -68,10 +76,13 @@ class Label : public BaseWidget
     bool centered_ = false;
     bool wordWrap_ = false;
     bool shouldElide_ = false;
+    bool markdownEnabled_ = false;
```
Nerixyz: Many methods in the implementation now branch on this. Maybe this label should be a separate class?

src/widgets/Label.hpp
```diff
@@ -5,7 +5,10 @@
 
 #include <pajlada/signals/signalholder.hpp>
 
+#include <memory>
+
 class QFontMetricsF;
+class QTextDocument;
```
pajlada: You can undo the include & forward-declaration of QTextDocument here.

src/widgets/MarkdownLabel.hpp
```diff
@@ -0,0 +1,37 @@
...
+///
+/// MarkdownLabel inherits from Label and adds markdown rendering capabilities.
+/// It automatically handles markdown document creation and rendering.
+class MarkdownLabel : public Label
+{
+public:
+    explicit MarkdownLabel(QString text = QString(),
+                           FontStyle style = FontStyle::UiMedium);
+    explicit MarkdownLabel(BaseWidget *parent, QString text = QString(),
+                           FontStyle style = FontStyle::UiMedium);
```
pajlada: Seems like we only use the default-constructor. I'd prefer if you only create one constructor that requires the parent & default text, meaning the places where you create this with the LayoutCreator emplace would need to change a bit.

so something like
```c++
MarkdownLabel(BaseWidget *parent, QString text, FontStyle style = ....)
```

src/widgets/MarkdownLabel.hpp
```diff
@@ -0,0 +1,37 @@
...
+protected:
+    void paintEvent(QPaintEvent *event) override;
+    void mousePressEvent(QMouseEvent *event) override;
+    void mouseMoveEvent(QMouseEvent *event) override;
+
+private:
+    void updateSize() override;
+    void initializeMarkdownDocument();
+
+    mutable std::unique_ptr<QTextDocument> markdownDocument_;
```
pajlada: This shouldn't need to be mutable anymore

resources/buttons/bold-darkMode.svg
pajlada: 👍 Preferably use icons from https://fluenticons.co/outlined instead where possible, and where not possible let us know where you got these icons from so we can correctly attribute them / follow their license.

pajlada (changes requested):
Layout nits:
I think the "markdown preview" checkbox should be on the right side of the top bar

The style buttons should probably be grouped a little closer - right now they stretch to take up as much space as they can

Not a full review, but overall seems cool

Thank you!

src/widgets/Label.cpp
```diff
@@ -214,7 +215,7 @@ bool Label::updateElidedText(const QFontMetricsF &fontMetrics, qreal width)
 
 QRectF Label::textRect() const
 {
-    return this->rect().toRectF().marginsRemoved(this->currentPadding_);
+    return QRectF(this->rect()).marginsRemoved(this->currentPadding_);
```
pajlada, line 218: What did this change achieve?

src/widgets/dialogs/EditUserNotesDialog.cpp
```diff
@@ -17,12 +21,199 @@ EditUserNotesDialog::EditUserNotesDialog(QWidget *parent)
...
+                               .dark = ":/buttons/bold-darkMode.svg",
+                               .light = ":/buttons/bold-lightMode.svg",
+                           })
+                           .getElement();
+    QObject::connect(boldButton, &Button::leftClicked, this, [&] {
+        auto cursor = this->textEdit_->textCursor();
+        if (cursor.hasSelection())
+        {
+            auto selectedText = cursor.selectedText();
+            QChar newLine(0x2029);
```
pajlada: Make use of https://doc.qt.io/qt-6/qchar.html#SpecialCharacter-enum instead of hardcoding the hex code

pajlada (changes requested):
Overall looks & works well enough. Thanks for your work on this feature.
I pushed some minor fixes, let me know if any of those changes look off to you.

Extra non-blocking feedback, let me know if you want to work on improving its UX or if you want to leave that to future PRs:
 - The heading and list button should probably insert the text at the start of the selected line, not at the cursor position.
 - The heading button works a bit weird with selected text. Maybe improved after the previous nit is fixed?
 - The bold, italizice, and link button inserts a newline at the end _if_ you have a selection. This is ignored in markdown, but still looks weird in the raw text.
 - Clicking the heading/bold/italizice button again should, imo, try to remove the boldness/italicness/heading of the word/line if it's currently active. Very stretch goaly though.

src/widgets/Label.cpp
jacob-thompson: this line shouldn't have been touched, but I used an agent to throw together the first draft of this PR, and they tend to do pointless crap like this. I should have caught this when I reverted the `Label` changes in favor of `MarkdownLabel`, apologies.

it clearly changes nothing functionally, but I _would_ argue that the latter is more readable. I feel it is easier to make sense of one function call on a constructor than three strung-together function calls. I'll revert it if you disagree

pajlada: It doesn't matter much, so feel free to leave it. If the functions are identical, the only downside to leaving it is it makes the blame for the file a little worse.
Thanks for the clarification

jacob-thompson:
> * The heading and list button should probably insert the text at the start of the selected line, not at the cursor position.

in the case of the heading button, I thought the same early on, but I opted to copy the behavior of GitHub's markdown editor

in the case of the list button, you are right and I've just fixed this

> * The bold, italizice, and link button inserts a newline at the end _if_ you have a selection. This is ignored in markdown, 
but still looks weird in the raw text.

this was also a mistake and I've just fixed this

jacob-thompson:
> * Clicking the heading/bold/italizice button again should, imo, try to remove the boldness/italicness/heading of the word/line if it's currently active.

I've implemented a relatively simple solution with regex

pajlada (approved):

PR #6496, opened 2025-09-26
fix: usercard resizing improperly without recent messages

teknsl:
before:
before

after:
after

very niche but easy fix so uh


pajlada (approved):
This is good for now - something to keep in mind is that user notes might need to take up more space if we support newlines in them.
Right now we don't, so right now it's not a problem.

PR #6505, opened 2025-09-28
Add obertura777 to contributor list

Obertura777:
Add self to contributor list


Obertura777:
I put my github ID as the name; does it have to be my twitch ID?

iProdigy:
github id is correct

PR #6509, opened 2025-09-29
Enable browser integration settings on all platforms

crazysmc:
Referencing PR #6031

Thanks for the quick response.

The browser integration settings will be available on all platforms and the description will be different, depending on the availability of the chat replace feature.


pajlada (approved):
Thanks! I updated the Windows text a little bit, and disabled the `Attach to any browser (may cause issues)` setting on non-Windows systems since that is actually a Windows-only feature.

PR #6511, opened 2025-09-30
refactor: merge emote element flags

Nerixyz:
All emote elements are "equal" - there's no special handling for one specific emote provider. Thus, the flags can be merged. This is mainly required to abstract emote providers and to make it easy for new ones to be added (#6436).

With this PR, `{Twitch,BTTV,FFZ,Seventv}Emote{Image,Text}` are merged to `Emote{Image,Text}`.
This frees up six "slots" in the enum. I'll make a followup PR to clean up the enum values.


pajlada (approved):
This PR would have been a bit easier to review if you split up the `MessageElementFlag` changes and actual code/function signature changes into separate commits

pajlada:
>  This frees up six "slots" in the enum. I'll make a followup PR to clean up the enum values.

I don't see a reason to do this, but happy to listen

PR #6514, opened 2025-10-01
fix: add explanation for the logging feature under moderation

### LINKED ISSUE #6258: Improve explanation about the Logging feature under Moderation
### Checklist

### Describe your issue

The log feature is under Moderation:

The wiki is brief:
https://wiki.chatterino.com/Moderation/#logging

Since the Logging is under Moderation, it looks like that's meant to log only moderation related things, like people banned, things like that. But actually it records the whole chat in plain text UTF-8.

Some brief explanation could be added to Wiki and the UI, just to let people know `Record chat in plain text with time and date and emote codes` and any other thing it does.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.0 Running on macOS 15.5, kernel: 24.5.0

Nikk0o:
Adds an explanation about the logging feature in the moderation tab.

Fixes #6258


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Dev: Correct handling of eventsubs without any account. (#6503)
 - Dev: Removed dependency to Qt5 Compatibility module by updating libcommuni. (#6500)
 - Dev: Merged emote element flags from different providers into two. (#6511)
+- Minor: Added a description for the logging option under moderation tab. (#6514)
```
Nerixyz: Move this up to the other `Minor` entries.

Nerixyz:
Are you also planning to update the [wiki entry](https://github.com/Chatterino/wiki/blob/b42415b110f7c3991c691eaecb65a973411e5d24/docs/Moderation.md?plain=1#L15)? That currently mentions "moderators" even though you don't need to be a moderator to save logs. That entry could also be a bit longer - possibly showing some example logs(?).

The Chatterino settings page could then link to the wiki entry for further information.

Nikk0o:
That sounds good. I was also thinking about making it a tooltip that shows up when you hover over the Enable option to take up less space. What do you think?

And yes, I will update the wiki, but I want to wait until this PR is merged to update the picture of the logging tab.

Nerixyz:
> I was also thinking about making it a tooltip that shows up when you hover over the Enable option to take up less space. What do you think?

The issue with tooltips is that you usually don't know it's there, unless there's some hint like a '?' icon or similar. Maybe others have an opinion on this.

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -81,6 +81,11 @@ ModerationPage::ModerationPage()
             "Enable logging", getSettings()->enableLogging);
         logs.append(enableLogging);
 
+		auto description = logs.emplace<QLabel>(
+				"Logging saves the chat history in plain text "
+				"with time and date and emote codes.");
+		description->setStyleSheet("color: #bbb");
```
jupjohn: Should this be theme-dependant?

src/widgets/settingspages/ModerationPage.cpp
```diff
@@ -81,6 +81,11 @@ ModerationPage::ModerationPage()
             "Enable logging", getSettings()->enableLogging);
         logs.append(enableLogging);
 
+		auto description = logs.emplace<QLabel>(
+				"Logging saves the chat history in plain text "
+				"with time and date and emote codes.");
```
jupjohn:
```suggestion
				"Log the time and content of chat/system messages to disk.");
```
I don't think "emote codes" makes sense as chatterino logs them as text.

jupjohn:
I feel leaving this as a label is fine. Moving it into a tooltip (even with a `?` icon) makes it harder to know what this is at a glance.

pajlada:
> > I was also thinking about making it a tooltip that shows up when you hover over the Enable option to take up less space. What do you think?
> 
> The issue with tooltips is that you usually don't know it's there, unless there's some hint like a '?' icon or similar. Maybe others have an opinion on this.

I'd like to replicate Factorio's system where if we have a tooltip, we'd append a '?' icon next to it. CC @Mm2PL as I know we've discussed this before

jupjohn:
> I'd like to replicate Factorio's system where if we have a tooltip, we'd append a '?' icon next to it.

If this starts to get done application wide (so there's a user expectation of how it works,) then that would be nice!

EDIT: I still feel this shouldn't be a tooltip since it describes the whole page/feature instead of the single toggle control.

pajlada:
> EDIT: I still feel this shouldn't be a tooltip since it describes the whole page/feature instead of the single toggle control.

Agreed - a label works fine here

Nikk0o:
I added the description next to the log path. Unfortunately, it still creates a new line because the text is too long, but I think it looks prettier than before.

I also want to make the tooltip show up only over the "Logs are saved at XYZ", but I couldn't find a way to do that and the text wrapping at the same time.

Nerixyz:
> I'd like to replicate Factorio's system where if we have a tooltip, we'd append a '?' icon next to it.

For those unaware like me, here's an example: https://www.youtube.com/watch?v=idklVQ6dhUQ&t=150s
They have a `(i)` next to the setting label. If you hover over it, a tooltip immediately appears (similar to how we show tooltips in the chat).

> Unfortunately, it still creates a new line because the text is too long, but I think it looks prettier than before.

That's totally fine. My concern was that if we have two labels, (1) we start two lines with "Log*" and (2) the space between them is too large. With one label that wraps, the space between the lines isn't that large.

Nikk0o:
Okay, I pushed the changes. Now it's only one label with the path and the explanation.

pajlada:

PR #6515, opened 2025-10-02
feat: add options to close multiple visible tabs in context menu

otaviobia:
This is a draft PR for the "close all tabs" and "close other tabs" features, as commonly seen in web browsers.

I recently had to close all my Chatterino tabs and my first instinct was to look for it in the context menu, there wasn't an option so I closed them manually, having to tediously click yes on the confirmation prompts every time. I now know I could have spammed "ctrl + shift + W" to do it quicker, but I still think the dedicated buttons are more user friendly.

A few of my concerns:
- Is there demand for it? Most users don't close all tabs that frequently.
- How to avoid missclicks? I added confirmation prompts to avoid accidental mass closing.
- Is my proposed solution conflicting with "split tabs"? I think the "ctrl + G" hotkey for reopening closed splits might display some strange behaviors when done after closing all or other tabs. I need to investigate it further.

Feedback is appreciated, thanks!


otaviobia:
https://github.com/user-attachments/assets/e42befa8-47e0-4c1a-80b4-563cc8eebc05

Nerixyz:
> * Is there demand for it? Most users don't close all tabs that frequently.

Since this feature is pretty small, it isn't that much of an issue. Browsers have these options when closing tabs, so in my opinion, it's worth having this in Chatterino too.

> * How to avoid missclicks? I added confirmation prompts to avoid accidental mass closing.

Speaking of browsers: in Firefox, these options are in a submenu. To close multiple tabs using the mouse, you'd right-click the tab and select "Close Multiple Tabs > Close Other Tabs". However, in Chrome, they're in the main context menu.

> * I think the "ctrl + G" hotkey for reopening closed splits might display some strange behaviors when done after closing all or other tabs.

As long as it reopens _some_ split closed by the operation, that should be fine. Which split is reopened should be considered implementation-defined until someone has a strong opinion about it and opens an issue.

otaviobia:
I added the options to close all tabs to the left or to the right. I also took the Firefox approach and grouped everything into a submenu (which I think removed the need for confirmation prompts). Lastly, I made it so this submenu is disabled if there's only one tab and the same thing for "close left" or "close right" depending if the tab is in a corner and which one.

Regarding tab visibility, right now it closes everything, even if it's hidden, I think it makes more sense to close only what's visible. My next steps will be related to this, but I wanted to ask first if that's also the behavior you would expect.

src/widgets/helper/NotebookTab.cpp
```diff
@@ -909,6 +1049,39 @@ void NotebookTab::mousePressEvent(QMouseEvent *event)
...
+                    {
+                        if (auto *tab = container->getTab();
+                            tab && tab->isVisible())
+                        {
+                            visibleTabs.push_back(tab);
+                        }
+                    }
+                }
+
+                const int visibleTabCount = int(visibleTabs.size());
```
pajlada: No need to cast to an int here, then you can avoid casting it back to size_t later

src/widgets/helper/NotebookTab.cpp
```diff
@@ -909,6 +1049,39 @@ void NotebookTab::mousePressEvent(QMouseEvent *event)
...
+
+                int visibleTabIndex = -1;
```
pajlada: Use an std::optional<size_t> or w/e here instead of an int, and I think it's meant to be called `selectedTabIndex`

pajlada:
Alternatively, you could use Notebook's `getSelectedIndex` & solve all the visibility checks in the first loop

otaviobia:
It seems Notebook already had the functions I needed! I also changed `visibleTabIndex` to `selectedTabIndex` and I think everything should be fine now

src/widgets/helper/NotebookTab.cpp
```diff
@@ -124,6 +124,146 @@ NotebookTab::NotebookTab(Notebook *notebook)
...
+    this->closeMultipleTabsMenu_ = new QMenu("Close Multiple Tabs", this);
+    this->menu_.addMenu(closeMultipleTabsMenu_);
+
+    closeMultipleTabsMenu_->addAction("Close All Visible Tabs", [this]() {
+        auto reply = QMessageBox::question(
+            this, "Close All Visible Tabs",
+            "Are you sure you want to close all visible tabs?",
+            QMessageBox::Yes | QMessageBox::Cancel);
+
+        if (reply == QMessageBox::Yes)
```
Nerixyz: Invert this to `if (reply != QMessageBox::Yes)` and early-return. This way, the code inside isn't as indented. The same goes for the other actions below.

PR #6518, opened 2025-10-03
ci(freebsd): skip imageformats install

pajlada:
it's not necessary for just building

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,7 @@ freebsd_instance:
 
 task:
   install_script:
-    - pkg install -y boost-libs git libnotify qt6-base qt6-svg qt6-imageformats qtkeychain-qt6 cmake pkgconf
+    - pkg install -y boost-libs git libnotify qt6-base qt6-svg qtkeychain-qt6 cmake pkgconf
   script: |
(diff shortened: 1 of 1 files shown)


Nerixyz:
I think we should just update to 14.3 since 14.2 is EOL.

pajlada:
> I think we should just update to 14.3 since 14.2 is EOL.

We should, but it's unrelated to this issue.
image

Nerixyz:
> We should, but it's unrelated to this issue.

Then either we're installing the package wrong or something else with FreeBSD is wrong - the package exists in the package search: https://ports.freebsd.org/cgi/ports.cgi?query=qt6-imageformats&stype=all&sektion=all.

pajlada:
> Then either we're installing the package wrong or something else with FreeBSD is wrong - the package exists in the package search: [ports.freebsd.org/cgi/ports.cgi?query=qt6-imageformats&stype=all&sektion=all](https://ports.freebsd.org/cgi/ports.cgi?query=qt6-imageformats&stype=all&sektion=all).

I'm aware, hence the PR. I want FreeBSD to be able to build, even if not exactly how I'd like it to build.

PR #6519, opened 2025-10-04
fix: compilation error with Clang 21

Nerixyz:
The range of `MyOpen` is `[1, 15]` but we instantiated `16` - this is an error in Clang 21. The point of the test is to see what happens if the name of an unnamed variant is requested. This PR fixes that by using a value inside the allowed range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Minor: Fixed usercard resizing improperly without recent messages. (#6496)
 - Minor: Added setting for character limit of deleted messages. (#6491)
 - Minor: Added link support to plugin message API. (#6386)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Minor: Fixed usercard resizing improperly without recent messages. (#6496)
 - Minor: Added setting for character limit of deleted messages. (#6491)
 - Minor: Added link support to plugin message API. (#6386)
+- Bugfix: Fixed compilation error with Clang 21. (#6519)
```
pajlada: Make this a dev entry & clarify that it's an error in the test

PR #6521, opened 2025-10-05
dev: Use system or package-manager-defined RapidJSON

pajlada:
This removes our RapidJSON submodule and expects the user to either
install their RapidJSON package from their operating system / package
manager, or have a CMake package manager like conan install it for them.


pajlada:
This has essentially been solved in the main branch.
If the user has rapidjson installed on their system, it will be used. (assuming it can be found with `find_package`).
If the user wants to customize where Chatterino finds rapidjson, they can set the `FETCHCONTENT_SOURCE_DIR_RAPIDJSON` cmake flag

PR #6530, opened 2025-10-08
refactor: move LuaLS meta to a subdirectory

Nerixyz:
In preparation for modules that can be `require()`'d separately. For these modules, there needs to be at least one file per module. The [`---@meta <name>`](https://luals.github.io/wiki/annotations/#meta) creates a logical module that can be `require()`'d. The current definitions are global - they can't be imported through `require('Chatterino2')`, so they're named `_`.

New modules such as the JSON utilities from #6420 will go in `docs/lua-meta`.


Mm2PL:
It'd be kinda cool to auto-generate meta files other than globals

PR #6531, opened 2025-10-08
refactor: consolidate font picking into one dialog

### LINKED ISSUE #6038: Refactor font picker/settings
Currently, choosing the font consists of a few dropdowns & the font dialog.

Instead, the font, font size, and font weight should all be controlled by the font dialog

The settings dialog's font options should then instead only consist of a single button "Choose font" or something that opens the font dialog, with an accompanying always-updating label stating your font family, font size, and font weight

teknsl:
Intends to close #6038

* Should the current font label have its width restricted?
* Should the row be placed higher? I have placed it lower due to finding it better looking with long font names
* ~Is using QFont::to/fromString ok for merging the different settings? This turns 3 event fires into one when the font changes and simplifies a little imo~
* ~Should QFontDialog be completely replaced? Effects and writing system properties are ignored even though they are stored in the setting, as Fonts::createFontData makes new QFonts based on limited properties of the stored font~
* ~Should DEFAULT_FONT be rewritten to a "safer" const QFont instead of raw format?~
* ~Should a different icon be used or one at all? Current icon was taken from Google Material Icons~
* Should more than family and point size be shown in the label?

<details>
<summary>Image</summary>
image
</details>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1094,6 +1094,10 @@ target_compile_definitions(${LIBRARY_PROJECT} PUBLIC
     IRC_NAMESPACE=Communi
     $<$<BOOL:${WIN32}>:_WIN32_WINNT=0x0A00> # Windows 10
     $<$<BOOL:${BUILD_TESTS}>:CHATTERINO_WITH_TESTS>
+    # Disable min/max macros from Windows.h
+    $<$<BOOL:${WIN32}>:NOMINMAX>
+    # Exclude some headers when including Windows.h
+    $<$<BOOL:${WIN32}>:WIN32_LEAN_AND_MEAN>
     )
 
 set(_c2_qt_defs
@@ -1259,9 +1263,6 @@ if (MSVC)
             -Wno-missing-designated-field-initializers
         )
     endif()
-
-    # Disable min/max macros from Windows.h
-    target_compile_definitions(${LIBRARY_PROJECT} PUBLIC NOMINMAX)
 else ()
     target_compile_options(${LIBRARY_PROJECT} PUBLIC
         -Wall
--- a/BUILDING_ON_LINUX.md
+++ b/BUILDING_ON_LINUX.md
@@ -14,10 +14,10 @@ Use <https://github.com/Chatterino/docker/pkgs/container/chatterino2-build-ubunt
 
 The built binary should be exportable from the final image & able to run on your system assuming you perform a static build. See our [build.yml GitHub workflow file](.github/workflows/build.yml) for the CMake line used for Ubuntu builds.
 
-### Debian 12 (bookworm) or later
+### Debian 13 (trixie) or later
 
 ```sh
-sudo apt install qt6-base-dev qt6-svg-dev qt6-image-formats-plugins libboost1.81-dev libnotify-dev libssl-dev cmake g++ git
+sudo apt install qt6-base-dev qt6-svg-dev qt6-image-formats-plugins libboost-dev libnotify-dev libssl-dev libsecret-1-dev pkg-config cmake g++ git
 ```
 
 ### Arch Linux
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Dev: Fixed compilation warnings on clang-cl. (#6528)
 - Dev: Fixed compilation error in tests with Clang 21. (#6519)
 - Dev: The LuaLS meta files moved from `docs/plugin-meta.lua` to `docs/lua-meta/globals.lua`. (#6530)
+- Dev: Compile time definitions for `Windows.h` are now conditional based on `WIN32` instead of `MSVC`. (#6534)
 
 ## 2.5.4
 


pajlada:
>     * Should a different icon be used or one at all? Current icon was taken from Google Material Icons

Use an icon from https://fluenticons.co/outlined if possible, or just use text "Change font" or something like that

src/singletons/Settings.hpp
```diff
@@ -189,17 +187,9 @@ class Settings
...
-    IntSetting chatFontSize{
-        "/appearance/currentFontSize",
-        DEFAULT_FONT_SIZE,
-    };
-    IntSetting chatFontWeight = {
-        "/appearance/currentFontWeight",
-        QFont::Normal,
+    QStringSetting chatFont{
+        "/appearance/chatFont",
+        DEFAULT_FONT,
```
pajlada: I'd prefer if you can keep the same 3 settings.
Otherwise we'll have to do our world first setting migration (which I have some uncommited code for already).

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -48,6 +48,63 @@ const QStringList ZOOM_LEVELS = {
     "1.6x", "1.8x", "2x",   "2.33x", "2.66x", "3x",      "3.5x", "4x",
 };
 
+class FontSettingWidget : public QWidget
```
Nerixyz: If you keep this, could you move this to its own file, as it's not tiny? Maybe somewhere in `widgets/helper`.

We also have the convention to access class/struct members with `this->` to make code review easier.

teknsl: should i move the implementations into .cpp or just inline?

Nerixyz: > should i move the implementations into .cpp or just inline?

In its own .cpp file, if any.
To be clear, the class is at this threshold where it's neither tiny nor medium/large, so I'd also be fine if it's kept here. Given that there aren't any other widgets defined in this file, I feel like it's better to move it to its own .cpp file.

teknsl: i decided to seperate the implementation - if QFontDialog should be replaced its replacement fits well inside the new unit

teknsl:
Screenshot_20251010_152152

~theres some issues with the font not updating properly in the chat after the settings have been set, but updates when resized(not entirely sure this does it 100% correctly) or application is restarted~

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+    {
+        auto const *cast = dynamic_cast<const IntItem *>(&other);
+        assert(cast);
+        return this->value < cast->value;
+    }
+
+    void setValue(int v)
+    {
+        this->value = v;
+        this->QListWidgetItem::setText(QString::number(v));
```
Nerixyz:
```suggestion
        QListWidgetItem::setText(QString::number(v));
```

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+class FontSizeWidget : public QWidget
+{
+    Q_OBJECT
+
+    IntItem *customItem = new IntItem;
+    QListWidget *list = new QListWidget;
+    QSpinBox *edit = new QSpinBox;
+
+    void setListTo(int size)
```
Nerixyz: What does this do? From the name, I'd expect it to set `list`.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+class IntItem : public QListWidgetItem
+{
+    int value;
+
+public:
+    void setText(QString const &) = delete;
+
+    IntItem(int v = 0)
+        : QListWidgetItem(QString::number(v))
```
Nerixyz: Set `type` to `QListWidgetItem::UserType + something` (see [docs](https://doc.qt.io/qt-6/qlistwidgetitem.html#subclassing)).

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    return nullptr;
+}
+
+class FontSizeWidget : public QWidget
+{
+    Q_OBJECT
+
+    IntItem *customItem = new IntItem;
```
Nerixyz: Document this.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+    QListWidget *list = new QListWidget;
+    QSpinBox *edit = new QSpinBox;
```
Nerixyz: Initialize these in the constructor.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+        if (item->getValue() == value)
+        {
+            return item;
+        }
+    }
+
+    return nullptr;
+}
+
+class FontSizeWidget : public QWidget
```
Nerixyz: I think it would be better if this and the other widgets were declared first, and the methods were implemented below. `IntItem` is tiny enough.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+            return;
+        }
+
+        this->customItem->setValue(size);
+        this->customItem->setHidden(false);
+        this->list->setCurrentItem(this->customItem);
+    }
+
+public:
+    FontSizeWidget(QFont const &initialFont, QWidget *parent = nullptr)
```
Nerixyz:
```suggestion
    FontSizeWidget(const QFont &initialFont, QWidget *parent = nullptr)
```

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+        header->addWidget(new QLabel("Size"));
+        header->addWidget(this->edit);
+        header->setContentsMargins(0, 0, 0, 0);
+
+        this->edit->setValue(initialFont.pointSize());
+        this->setListTo(initialFont.pointSize());
+
+        QObject::connect(this->edit, &QSpinBox::valueChanged, this,
+                         [this](int value) {
+                             this->list->blockSignals(true);
```
Nerixyz: Use [`QSignalBlocker`](https://doc.qt.io/qt-6/qsignalblocker.html)

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+        QObject::connect(search, &QLineEdit::textChanged, this->proxy,
+                         &QSortFilterProxyModel::setFilterFixedString);
+
+        QObject::connect(
+            this->list->selectionModel(), &QItemSelectionModel::currentChanged,
+            this, [this](QModelIndex const &modelIndex, QModelIndex const &) {
+                if (modelIndex.isValid())
+                {
+                    Q_EMIT selectedChanged();
```
Nerixyz:
```suggestion
                    this->selectedChanged();
```

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+                }
+            });
+
+        this->setSelected(initialFont.family());
+    }
+
+    void setSelected(QString const &family)
```
Nerixyz:
```suggestion
    void setSelected(const QString &family)
```

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+        , fontFamiliesW(new FontFamiliesWidget(initialFont))
+        , fontSizeW(new FontSizeWidget(initialFont))
+        , fontWeightW(new FontWeightWidget(initialFont))
+    {
+        auto *layout = new QVBoxLayout;
+        auto *choiceLayout = new QHBoxLayout;
+        auto *choiceSideLayout = new QVBoxLayout;
+        auto *buttons = new QDialogButtonBox;
+        auto *confirmBtn = new QPushButton("Ok");
+        auto *cancelBtn = new QPushButton("Cancel");
```
Nerixyz: You can use
```cpp
buttons->addButton("Ok", QDialogButtonBox::AcceptRole);
buttons->addButton("Cancel", QDialogButtonBox::RejectRole);
```
this creates the buttons for you.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+#include <QTextEdit>
+#include <QVBoxLayout>
+
+namespace chatterino {
+
+namespace {
+
+class IntItem : public QListWidgetItem
+{
+    int value;
```
Nerixyz: We usually put the private members at the bottom of the class (see the other classes in Chatterino).

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+            return;
+        }
+
+        QFont font = dialog.getSelected();
+        getSettings()->chatFontFamily = font.family();
+        getSettings()->chatFontSize = font.pointSize();
+        getSettings()->chatFontWeight = font.weight();
+
+        this->updateCurrentLabel();
+        getApp()->getWindows()->forceLayoutChannelViews();
```
Nerixyz: Not an issue with this PR (happens on master too), but we also need this when the user clicks "Cancel" in the settings dialog and settings are reverted.

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+    {
+        return this->value;
+    }
+};
+
+IntItem *findIntItemInList(QListWidget *list, int value)
+{
+    int n = list->count();
+    int i = 0;
+    for (; i < n; ++i)
```
Nerixyz:
```suggestion
    for (int i = 0; i < n; ++i)
```

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,457 @@
...
+        this->list->blockSignals(true);  // redundant signals
+
+        this->list->clear();
+        this->list->addItem(defaultItem);
+
+        weights.insert(defaultWeight);
+
+        /* the goal is to only display valid weights and this gets close, but
+         * is not perfect for all fonts.
+         */
```
Nerixyz: Prefer `//` comments.

src/widgets/helper/FontSettingWidget.cpp
teknsl: i’ve forward declared them above their implementations, i think this makes it clear and more navigable than putting them all at the top - im ok with moving them all to the top if that is whats wanted

src/widgets/helper/FontSettingWidget.cpp
teknsl: i assume the missing this-> was the problem here and Q_EMIT got caught in it?

src/widgets/helper/FontSettingWidget.cpp
teknsl: should it be fixed with this PR?

Nerixyz: No, in another one.

teknsl:
> In the future, we could also try to display a chat message there. We'd need to change something in `MessagePaintContext` or its members to allow custom fonts to be used. We already have a `MessageView`.

we could also expose the scaling in `src/singletons/Fonts.cpp` and apply it to the font, useful if we keep the current display or switch to `MessageView`.

teknsl:
@Nerixyz I have patches ready that expose font size scaling (1) and to use them in the dialog (2). Wondering how i should go about it - make PR with patch 1 and hold onto patch 2 until that one is merged or aborted?

Nerixyz:
> Wondering how i should go about it - make PR with patch 1 and hold onto patch 2 until that one is merged or aborted?

Yep, sounds good.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -263,6 +236,8 @@ void GeneralPage::initLayout(GeneralPageView &layout)
 
     SettingWidget::dropdown("Tab style", s.tabStyle)->addTo(layout);
 
+    layout.addWidget(new FontSettingWidget, {"font"});
```
Nerixyz: Could include "size" and "weight" in the keywords.

src/CMakeLists.txt
```diff
@@ -728,6 +728,8 @@ set(SOURCE_FILES
         widgets/helper/TableStyles.hpp
         widgets/helper/TrimRegExpValidator.cpp
         widgets/helper/TrimRegExpValidator.hpp
+        widgets/helper/FontSettingWidget.cpp
+        widgets/helper/FontSettingWidget.hpp
```
Nerixyz: Sort this in alphabetical order (just the widgets/helper block).

src/widgets/helper/FontSettingWidget.hpp
```diff
@@ -0,0 +1,15 @@
+#include <QLabel>
```
Nerixyz, line 6: Add `#pragma once` at the top (see other headers).

Nerixyz:
I can't approve & someone else should look over this too. But it looks fine now.

teknsl:
Would like some thoughts on which of these looks best.

<details><summary><code>QPushButton</code> with distinct icon (current)</summary>
Screenshot_20251023_170937
</details>

<details><summary><code>QToolButton</code> with distinct icon</summary>
Screenshot_20251023_171046
</details>

<details><summary><code>QToolButton</code> with arrow</summary>
Screenshot_20251023_171431
</details>

pajlada:
how did https://github.com/Chatterino/chatterino2/pull/6531#issuecomment-3382008380 look

teknsl:
The current icon is from fluent icons, https://github.com/Chatterino/chatterino2/pull/6531/commits/d1295abd7324b322a6ba9a0a647cf5e20ce93edc

teknsl:
Here is the one with text button

<details><summary>Image</summary>
image
</details>

src/widgets/helper/FontSettingWidget.cpp
```diff
@@ -0,0 +1,500 @@
...
+                     });
+}
+
+void FontWeightWidget::setFamily(const QString &family)
+{
+    QSignalBlocker listSignalBlocker(this->list);
+    QVector<int> weights;
+    unsigned closest = -1;
+
+    this->list->clear();
```
teknsl: Re-allocations here make the function very simple but if that is not wanted I have a patch which at the cost of lines (+44 -12) reuses the items.

teknsl:
To note: there are some fonts where weights don’t get applied to chat but only to the sample, maybe QPainter’s fault?

This is part of why I have added a "Not all weights are supported" label, I think users would complain if its not acknowledged.

Could try to make the sample render the same way the chat does so its consistent, but would add some complexity.

<video src="https://github.com/user-attachments/assets/a3978a56-3b3d-4a7d-b57e-29f48fd10db7">

Nerixyz:
> maybe QPainter’s fault?

That's most likely our fault of not setting the font correctly. Every paint operation in Qt goes through QPainter, though text is a bit more complicated.

Maybe this is an issue with variable fonts (i.e. fonts with a variable `wght` axis)? The Inter font is a good example for this.

teknsl:
There is something weird going on with Inter too, but it doesn’t seem like Qt considers it variable; `QFontInfo{QString{"Inter"}}.variableAxes().count()` produces 0. No idea what is going on here.

Atkinson Hyperlegible Mono has a variable weight axis, its weird that `QPlainTextEdit` renders it properly with standard `QFont{family, size, weight}` constructor but the painting in chat doesn’t - which uses the same(?).  For now I think we can just provide one weight in the case of variable weight fonts, as it is the only that works.

<hr>

Google Sans Code has no variable axes but unlike Inter it’s weights just get ignored and is rendered only in the sample like Atkinson Hyperlegible Mono (but Atkinson had variable weight axis) This is very inconsistent and the same issues are in the master branch.

With this I don’t think its worth it to only show one weight from variable weight fonts when the issue also exists for non-variable. 

<hr>

It seems to be QPainter’s fault with it being unable to draw some weights. I think the best course of action for now will be to replace the current sample display with something custom that uses a lone QPainter so it is consistent with what appears in chat.

<details>

```cpp
#include <QApplication>
#include <QPainter>
#include <QWidget>

#define FONT "Google Sans Code"
#define WEIGHT_1 800
#define WEIGHT_2 300
#define SIZE 23

class Widget : public QWidget
{
public:
    void paintEvent(QPaintEvent *) override
    {
        QPainter p1(this);
        p1.setFont({FONT, SIZE, WEIGHT_1});
        p1.drawText(20, 40, "Painter 1");

        QPainter p2(this);
        p2.setFont({FONT, SIZE, WEIGHT_2});
        p2.drawText(20, 80, "Painter 2");
    }
};

int main(int argc, char **argv)
{
    QApplication app(argc, argv);

    Widget widget;
    widget.resize(500, 500);
    widget.show();

    QApplication::exec();
}
```
Screenshot_20251027_002746
</details>

Nerixyz:
> `QFontInfo{QString{"Inter"}}.variableAxes().count()` produces 0. No idea what is going on here.

This returns 0 for me too. However, "Inter" itself isn't variable for me - I'd need "Inter Variable". This has the two expected axes: `QList(Optical size(opsz)[14...32; default=14], Weight(wght)[100...900; default=400])`.

> It seems to be QPainter’s fault with it being unable to draw some weights. I think the best course of action for now will be to replace the current sample display with something custom that uses a lone QPainter so it is consistent with what appears in chat.

That's strange. It doesn't happen for me on Windows with Qt 6.10.

If this is the same problem on master, that's fine, imo.

Nerixyz:
The dialog works well. The "Not all weights are supported" label is confusing to me, because for me, all weights are supported (on both GDI and DirectWrite). 

In the future, we could add an "apply" button. While the preview in the dialog works well, I found myself often clicking "ok" and reopening the dialog to try a font out. Could be in a followup.

teknsl:
I though it was a problem on all platforms but if it isn’t then I will remove it, didn’t really like the look of it anyway.

teknsl:
Apply button was a good idea and one I’ll probably use alot myself so I went and implemented it for this PR.

Nerixyz:
> I though it was a problem on all platforms but if it isn’t then I will remove it, didn’t really like the look of it anyway.

Now that the text is rendered like it is in messages (through `drawText`), you should always see the expected result. If choosing a different weight doesn't make a difference in the preview, that's fine.

pajlada (changes requested):
Use `show` on the `FontSettingDialog` so it doesn't block all interaction

e.g.
```c++
        static auto *dialog = new FontSettingDialog(this);
        dialog->show();
```


Add QWidget include in `FontSettingWidget.hpp`

Make new folder under `src/widgets` or `src/widgets/dialogs` and put each of your class in their own file

PR #6536, opened 2025-10-11
fix: do forceLayoutChannelViews on font change

teknsl:
* https://github.com/Chatterino/chatterino2/pull/6531#discussion_r2420896314

since its one line, does it need a changelog?

### DIFF
--- a/.github/workflows/test-arch-linux.yml
+++ b/.github/workflows/test-arch-linux.yml
@@ -28,9 +28,8 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
-          sudo pacman -Sy
-          sudo pacman -S --noconfirm 'qt6-base' 'qt6-tools' 'boost-libs' 'openssl' 'qt6-imageformats' 'qtkeychain-qt6' 'qt6-svg' 'libnotify'
-          sudo pacman -S --noconfirm 'git' 'boost' 'cmake' 'benchmark'
+          sudo pacman -Syu --noconfirm
+          sudo pacman -S --noconfirm qt6-base qt6-tools boost-libs qt6-imageformats qtkeychain-qt6 qt6-svg libnotify git boost cmake benchmark
 
       - uses: actions/checkout@v5
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Dev: Fixed warnings on Qt 6.10. (#6422)
 - Dev: Removed unused method in `Emojis`. (#6517)
 - Dev: Refactored `Emotes` into `EmoteController`. (#6516)
+- Dev: Fix Arch Linux partial upgrade error in CI. (#6553)
 - Dev: Added Qt keyword and warning flags project wide. (#6520)
 - Dev: Fixed compilation error in tests with Clang 21. (#6519)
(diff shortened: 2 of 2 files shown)


Nerixyz:
> since its one line, does it need a changelog?

Yea, something like "Bugfix: Fixed fonts not immediately reverting when cancelling the settings dialog"

teknsl:
I’ve spotted some logic in `WindowManager` that looks responsible for this situation, would additionally remove the need to manually call `forceLayoutChannelViews()` after the user has changed the font. Was there a reason this was in `layoutChannelViewsListener` instead of `forceLayoutChannelViewsListener`?

```diff
diff --git a/src/singletons/WindowManager.cpp b/src/singletons/WindowManager.cpp
index 5718dda2..bcb9cf73 100644
--- a/src/singletons/WindowManager.cpp
+++ b/src/singletons/WindowManager.cpp
@@ -163,9 +163,9 @@ WindowManager::WindowManager(const Args &appArgs_, const Paths &paths,
         settings.streamerModeHideModActions);
     this->forceLayoutChannelViewsListener.add(
         settings.streamerModeHideRestrictedUsers);
+    this->forceLayoutChannelViewsListener.add(fonts.fontChanged);
 
     this->layoutChannelViewsListener.add(settings.timestampFormat);
-    this->layoutChannelViewsListener.add(fonts.fontChanged);
 
     this->invalidateChannelViewBuffersListener.add(settings.alternateMessages);
     this->invalidateChannelViewBuffersListener.add(settings.separateMessages);
```

Nerixyz:
> Was there a reason this was in `layoutChannelViewsListener` instead of `forceLayoutChannelViewsListener`?

I don't think so. This seems like the better way. We should only test that this isn't called too often, as these forced layouts are expensive. But I suspect it's only when the settings are changed.

teknsl:
> We should only test that this isn't called too often

As far as I can tell `fonts.fontChanged` only gets fired from `Fonts.cpp:192` when any of the font related settings change. The new font dialog would fire this 3 times as family, size and weight are updated at the same time - is it possible or worth it to avoid this?

Nerixyz:
> The new font dialog would fire this 3 times as family, size and weight are updated at the same time - is it possible or worth it to avoid this?

I don't think you can avoid this completely. There's a `SignalArgs` in the settings library where you can pass some additional context like "Hey, I'll change the other settings too", but this will only work when setting the settings, not when resetting them (i.e. clicking "Cancel"). I'd guess that most of the time only one or two of the signal will fire, because one of them would be constant (e.g. keeping the font size or weight). iirc, then no signal is emitted if a setting didn't change.

teknsl:
> There's a SignalArgs in the settings library where you can pass some additional context like "Hey, I'll change the other settings too", but this will only work when setting the settings

If this is something we should do for https://github.com/Chatterino/chatterino2/pull/6531, I need help - I can’t grasp how to do this after trying to look through.

Nerixyz:
> If this is something we should do for #6531, I need help - I can’t grasp how to do this after trying to look through.

I'd say we shouldn't do this in that PR, because in most cases, users will only change the font family.

There are also other ways we could choose. We could also do an approach similar to what we do in image/frame loading, where we queue an event to Qt to batch updates. Either way, it's better to do this separately.

pajlada (approved):

PR #6537, opened 2025-10-11
dev: expose font size scaling function

teknsl:
* Facilitates previewing the chat scaled font in #6531

### DIFF
--- a/src/singletons/Fonts.cpp
+++ b/src/singletons/Fonts.cpp
@@ -56,40 +56,6 @@ int getUsernameBoldness()
 #endif
 }
 
-float fontSize(FontStyle style)
-{
-    auto chatSize = [] {
-        return static_cast<float>(getSettings()->chatFontSize);
-    };
-    switch (style)
-    {
-        case FontStyle::ChatSmall:
-            return 0.6F * chatSize();
-        case FontStyle::ChatMediumSmall:
-            return 0.8F * chatSize();
-        case FontStyle::ChatMedium:
-        case FontStyle::ChatMediumBold:
-        case FontStyle::ChatMediumItalic:
-        case FontStyle::TimestampMedium:
-            return chatSize();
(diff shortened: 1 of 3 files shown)


teknsl:
maybe removing the `scale` argument and let the caller scale the returned font size would be better, essentially just exposing the original function instead of this

src/singletons/Fonts.cpp
```diff
@@ -207,6 +173,48 @@ QFontMetricsF Fonts::getFontMetrics(FontStyle type, float scale)
     return this->getOrCreateFontData(type, scale).metrics;
 }
 
+float Fonts::scaleFontSize(FontStyle style, float size, float scale)
```
Nerixyz: It's not immediately clear what `size` and `scale` are.

`size` is the base **chat** font size and `scale` is the container scale. I'm not sure if we even need `scale` here, since that's just a factor on the result.

Imo, the function should be named `scaledFontSize` or `scaledSize`, since it doesn't modify anything.

teknsl:
is `auto v = getSettings()->chatFontSize` expensive enough to warrant only retrieving when needed like done in master?

Nerixyz:
> is `auto v = getSettings()->chatFontSize` expensive enough to warrant only retrieving when needed like done in master?

The fonts should be cached, so it should be fine to always get the value. Though, note that your example would copy the setting instead of getting the value. You could also do something like `const auto &s = getSettings()->chatFontSize` and use `s.getValue()`.

teknsl:
This PR could be skipped if assuming that the chat font size (`ChatMedium`) doesnt scale with anything except `getSettings()->uiScale` is fine.

pajlada:
not doing it this way, if we want to scale it we'll scale it differently anyway. this would only be relevant if we want to preview the small or large chat font size which I think is overkill.

PR #6547, opened 2025-10-18
fix: overflow when computing scrollbar rect

Nerixyz:
If `maximum <= minimum`, then we would divide by `0.0000001`. This could result in values for the y-position and height of the scrollbar rectangle that don't fit into a 32-bit signed integer. Consider the following:

- We're in a search popup without a search - all messages are displayed, we're at the bottom.
- `currentValue_ = 800`, `pageSize_ = 20`, `trackHeight_ = 500`
- Now, we're searching something, so the channel gets cleared.
- During this, `maximum_` and `minimum_` get set to 0 (while the rest stays as is)
- Then, calculating the y-position: $\frac{800}{\max(0.0000001, 0)} \cdot 500 + 1 = \frac{800}{0.0000001} \cdot 500 + 1 = 8000000000 \cdot 500 + 1 = 4000000000001 > 2147483648$

The difference between `maximum` and `minimum` is the number of messages currently in the channel. With this PR, we assume there's always at least one message when calculating the thumb rect. If there are no messages in a channel, we'll have an incorrect rect, but this won't be shown, because the scrollbar is always hidden in that case.

I updated the documentation of `Scrollbar` to better indicate the values and meaning of `minimum` and other members.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Minor: Added link support to plugin message API. (#6386)
 - Minor: Added a description for the logging option under moderation tab. (#6514)
 - Bugfix: Expose the "Extra extension IDs" setting on non-Windows systems too. (#6509)
+- Bugfix: Fixed scrollbar rect cmputation potentially resulting in overflows. (#6547)
 - Dev: Update release documentation. (#6498)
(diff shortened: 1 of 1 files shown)


Wissididom:
Changelog: `cmputation`

PR #6551, opened 2025-10-22
feat: menu action to sort tabs alphabetically

### LINKED ISSUE #2177: Sort using alphabetical order
**What should be added?**
Add the option to sort channel tabs using alphabetical order.

**Why should it be added?**
If you quickly want to find a chat, it gets confusing to find the one specific channel. I know you can reorder it, but once a new channel is added you have to reorder again. This would be very helpful if you plan to add a feature to bulk import channels, along with my other suggestion of removing duplicates.

teknsl:
* ~Potentially~ closes #2177
* Should the menu action text just be "Sort Tabs"?

When having many tabs it can get annoying to organize tabs and you need a reset. If this PR is wanted should an option to always sort alphabetically be included or is it enough to have the option to do it manually? (would be enough for my use)


src/widgets/Notebook.cpp
```diff
@@ -104,6 +105,20 @@ Notebook::Notebook(QWidget *parent)
             << "Notebook must be created within a BaseWindow";
     }
 
+    this->sortTabsAlphaAction_ = new QAction("Sort Tabs Alphabetically", this);
+
+    QObject::connect(this->sortTabsAlphaAction_, &QAction::triggered, [this] {
+        if (!this->isNotebookLayoutLocked())
```
Nerixyz: It could be confusing that nothing happens if you trigger this when the layout is locked.

We should make sure this can't be triggered from the emote popup. Though, I think that's already the case.

src/widgets/Notebook.cpp
```diff
@@ -104,6 +105,20 @@ Notebook::Notebook(QWidget *parent)
...
+        {
+            std::ranges::sort(this->items_, [](Item &a, Item &b) {
```
Nerixyz:
```suggestion
            std::ranges::sort(this->items_, [](const Item &a, const Item &b) {
```

Nerixyz (commented):
My hope is that eventually, plugins should be able to do something like this. But that will probably take a long time.

src/widgets/Notebook.cpp
teknsl: Should we hide/hint that you cant sort when the layout is locked or allow it to be used while locked? - Went with option 2 for now.

~Forgot about the emote popup, and it does show up there alongside lock tab layout. I will see if i can make it only show up in the channel tabs notebook~ Made the menu button exclusive to `SplitNotebook`.

Nerixyz: That's great! In my view, the locked notebook is for preventing accidental drags and here, the user explicitly clicked on the button.

src/widgets/Notebook.cpp
```diff
@@ -1244,6 +1245,16 @@ bool Notebook::shouldShowTab(const NotebookTab *tab) const
     return true;
 }
 
+void Notebook::sortTabsAlphabetically()
+{
+    std::ranges::sort(this->items_, [](const Item &a, const Item &b) {
+        return a.tab->getTitle() < b.tab->getTitle();
```
Nerixyz: Just noticed, this should be case-insensitive.

Nerixyz (commented):
Currently, once this is merged, #2177 will be closed. Not sure if that's intended.

From the issue, there's no indication of how the sorting should work ("always sort" or "click on a button and sort"). I think the current solution is good (so the issue should be closed afterward). You get the benefit of having the ability to have some tabs not in alphabetical order (through manually moving them after sorting). And I don't think tabs are added/removed that often.

src/widgets/Notebook.cpp
```diff
@@ -1244,6 +1246,19 @@ bool Notebook::shouldShowTab(const NotebookTab *tab) const
     return true;
 }
 
+void Notebook::sortTabsAlphabetically()
+{
+    QCollator cmp;
+    cmp.setCaseSensitivity(Qt::CaseInsensitive);
```
pajlada: In what scenario is the locale relevant for sorting tabs?

The only scenario I can think of is thousand separators/decimal separators

teknsl: Locale aware compare was a side effect of wanting case insensitive, I think it will handle non-ascii characters better but I don’t actually know - It just gave an easy way to do case insensitive compares.

If you don’t want to do that we could hardcode it to use the C locale or roll our own compare function (unless there is another way to do case insensitive compare.)

pajlada: I think you can just use QString's compare function with the CaseInsensitive flag https://doc.qt.io/qt-6/qstring.html#compare-1

src/widgets/Notebook.hpp
```diff
@@ -264,6 +266,8 @@ class SplitNotebook : public Notebook
     void showEvent(QShowEvent *event) override;
 
 private:
+    QAction *sortTabsAlphaAction_;
```
pajlada: No need to shorten `Alphabetically` to `Alpha`

pajlada (approved):
Just a small nit, otherwise looks good & works as expected and imo solves the linked issue

PR #6553, opened 2025-10-25
dev: fix Arch Linux deps

pajlada:
Sometimes, if we're building at the wrong time, we might be building
Qt with the wrong version of a dependent library (e.g. pcre2 in this case)

Make sure we always upgrade the system before attempting to build
Chatterino


pajlada:
image

PR #6554, opened 2025-10-25
feat(plugins): minimal account API

Nerixyz:
This adds an API to get the current Twitch account from Lua:
```lua
print(c2.current_account():user_id())
```

This resolves https://github.com/Chatterino/chatterino2/discussions/4999#discussioncomment-14779372. I didn't add a `ctx.executor` to the command context, as we only have one user logged in at any time. Furthermore, the current user might be useful for other things such as message building or while handling HTTP responses.

### DIFF
--- a/tests/src/Plugins.cpp
+++ b/tests/src/Plugins.cpp
@@ -1150,9 +1150,9 @@ TEST_F(PluginTest, testAccounts)
 
     auto res = lua->script(R"lua(
         local current = c2.current_account()
-        assert(current:user_login() == "testaccount_420")
-        assert(current:user_id() == "117166826")
-        assert(current:color() == "#ff000000") -- unset
+        assert(current:login() == "testaccount_420")
+        assert(current:id() == "117166826")
+        assert(current:color() == nil) -- unset
         assert(not current:is_anon())
     )lua");
     ASSERT_TRUE(res.valid()) << res.get<sol::error>().what();


docs/lua-meta/globals.lua
```diff
@@ -54,6 +54,35 @@ c2.ChannelType = {
...
+
+
+c2.accounts = {}
+c2.accounts.twitch = {}
+
+---@class c2.accounts.twitch.TwitchAccount
+c2.accounts.twitch.TwitchAccount = {}
+
+---@return string user_name The (login) name of the account
+function c2.accounts.twitch.TwitchAccount:user_name() end
```
pajlada: Just noting, not sure if I want any changes:
Should we ensure plugins are consistent with Twitch's terminology?
Login meaning the lowercased version of the username
Display Name meaning the potentially upercased version, potentially with CJK characters.

so, should this be called `user_login`?

src/controllers/plugins/api/Accounts.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api {
+
+/* @lua-fragment
+---@class c2.TwitchAccount
+c2.TwitchAccount = {}
+
+---@return string user_name The (login) name of the account
```
pajlada: Update to match new function name

pajlada (changes requested):
Just this nit, then this is good from my perspective

PR #6557, opened 2025-10-30
docs: replace debian bookworm deps with trixie

teknsl:
Old dependency list does not work on trixie, this new list does and is not backwards compatible.


Nerixyz:
Do we not need to install qt6-imageformats?

teknsl:
worked in a fresh container without

teknsl:
Nevermind, It builds but will complain at runtime instead.

pajlada (approved):
Thank you - we actually don't support Debian bookworm anymore if you want to remove that segment. Let me know if you prefer to leave it intact for a follow-up PR.

With the addition of Eventsub support we raised our minimum Boost version from 1.81 to 1.83

pajlada (approved):

PR #6562, opened 2025-11-04
deps: bump rapidjson for CMake 4.0 compatibility

Nerixyz:
CMake 4.0 dropped support for anything older than 3.5. So projects need to target at least 3.5. This was done in https://redirect.github.com/Tencent/rapidjson/pull/2250, so update to that commit. We're using their CMake since #6560 - before we only added the include directory.


pajlada:
This downgrades RapidJSON from  `24b5e7a8b27f42fa16b96fc70aade9106cf7102f` (Feb 5 2025) to `ebd87cb468fb4cb060b37e579718c4a4125416c1` (Dec 2 2024), so this looks like a downgrade rather than a bump. Is this intended?

Nerixyz:
I didn't see that the fetch content pr updated rapidjson.

PR #6570, opened 2025-11-08
feat: introduce emote providers

Nerixyz:
This is part 1/4 of refactoring emotes. It introduces the framework for emote providers but doesn't use it for any provider yet (except in tests). See https://github.com/Nerixyz/chatterino2/tree/refactor/third-party-emotes for all providers.

It introduces `EmoteProvider`s. These are the services (BTTV/FFZ/7TV) that provide emotes. They are managed by the `EmoteController`. In a channel, the emotes are managed by `ChannelEmotes`. 

The core idea hasn't changed since https://github.com/Chatterino/chatterino2/pull/6436 - this contains a more detailed explanation.


src/controllers/emotes/EmoteChannel.cpp
```diff
@@ -0,0 +1,21 @@
+#include "controllers/emotes/EmoteChannel.hpp"
+
+namespace chatterino {
+
+EmoteChannel::EmoteChannel(const QString &name, Type type)
```
pajlada: I'm not super fond of the additional channel layer. It seems to me like we could store the EmoteHolder in the Channel class as a unique_ptr or optional or something instead of this additional layer.

src/controllers/emotes/EmoteHolder.hpp
```diff
@@ -0,0 +1,70 @@
...
+using ChannelPtr = std::shared_ptr<Channel>;
+
+/// An entry for a single provider
+struct EmoteHolderItem {
+    std::weak_ptr<EmoteProvider> provider;
+    QString id;
+    EmoteMapPtr emotes;
+};
+
+class EmoteHolder
```
pajlada: comment

src/controllers/emotes/EmoteHolder.hpp
```diff
@@ -0,0 +1,70 @@
...
+class EmoteMap;
+using EmoteMapPtr = std::shared_ptr<const EmoteMap>;
+class EmoteController;
+class EmoteChannel;
+
+class Channel;
+using ChannelPtr = std::shared_ptr<Channel>;
+
+/// An entry for a single provider
+struct EmoteHolderItem {
```
pajlada: I don't like the term "item" or "entry", at least not when it's this generic.
How about something like "provider handle"?

src/controllers/emotes/EmoteProvider.hpp
```diff
@@ -0,0 +1,78 @@
...
+namespace chatterino {
+
+class EmoteChannel;
+
+struct EmoteLoadResult {
+    EmoteMapPtr emotes;
+    bool isPreemptiveCached = false;
+};
+
+class EmoteProvider
```
pajlada, line 24: What do you foresee the derived classes of EmoteProvider to be?

src/messages/MessageBuilder.cpp
```diff
@@ -420,13 +421,21 @@ EmotePtr makeSharedChatBadge(const QString &sourceName,
...
 {
     // Emote order:
+    //  - Channel emotes
     //  - FrankerFaceZ Channel
     //  - BetterTTV Channel
     //  - 7TV Channel
+    //  - Global emotes
     //  - FrankerFaceZ Global
     //  - BetterTTV Global
     //  - 7TV Global
```
pajlada, line 471: I'd rather user-specific emotes don't take precedence over something a streamer controls

src/common/Atomic.hpp
```diff
@@ -71,7 +71,7 @@ class Atomic<std::shared_ptr<T>>
     {
     }
 
-    Atomic(std::shared_ptr<T> &&val)
+    Atomic(std::shared_ptr<T> val)
```
pajlada: What made this change necessary?

src/common/QLogging.cpp
```diff
@@ -17,6 +17,7 @@ Q_LOGGING_CATEGORY(chatterinoCompletion, "chatterino.completion", logThreshold);
 Q_LOGGING_CATEGORY(chatterinoCrashhandler, "chatterino.crashhandler",
                    logThreshold);
 Q_LOGGING_CATEGORY(chatterinoEmoji, "chatterino.emoji", logThreshold);
+Q_LOGGING_CATEGORY(chatterinoEmotes, "chatterino.emotes", logThreshold);
```
pajlada, line 24: Random thought unrelated to your change here: It would be nice if emote providers would have their own logging category, and if that didn't require us to define something here

src/controllers/emotes/EmoteHolder.cpp
```diff
@@ -0,0 +1,192 @@
...
+#include "controllers/emotes/EmoteController.hpp"
+#include "controllers/emotes/EmoteProvider.hpp"
+#include "util/Functional.hpp"
+
+#include <QStringBuilder>
+
+namespace {
+
+template <typename T>
+bool ownerEquals(const std::weak_ptr<T> &lhs, const std::weak_ptr<T> &rhs)
```
pajlada: We have a `weakOwnerEquals` inside `src/util/WeakPtrHelpers.hpp`

src/util/Helpers.cpp
```diff
@@ -437,7 +439,17 @@ bool readProviderEmotesCache(const QString &id, const QString &provider,
 
         qCDebug(chatterinoCache)
             << "Loaded emote cache: " << id << "." << provider;
-        callback(doc);
+
```
pajlada: is this change important? I'd rather review this PR without this change if possible

pajlada (changes requested):
I'm OK with this approach & direction

src/messages/MessageBuilder.cpp
Nerixyz: These aren't only user-specific emotes. These are _all_ emotes (eventually). I migrated the providers in this order, starting from FFZ.

src/controllers/emotes/EmoteProvider.hpp
Nerixyz: For BTTV, FFZ, and 7TV: https://github.com/Nerixyz/chatterino2/blob/b8f9a7581b79cca504509d08c91b55f310f5ad64/src/controllers/emotes/BuiltinEmoteProvider.hpp.

Others would implement the pure virtual methods.

src/controllers/emotes/EmoteHolder.hpp
Nerixyz: It's not a handle, because it's specific to a channel, and it's only stored in the holder (users of the class won't store this). I'll try to find something else.

src/messages/MessageBuilder.cpp
pajlada: We could resolve these at the end instead of at the top then, since this will not change the emote order

src/controllers/emotes/EmoteProvider.hpp
pajlada: So a single BuiltinEmoteProvider class that we initialize 3 times for our 3 builtin emote providers?

Nerixyz: Yes.

src/messages/MessageBuilder.cpp
Nerixyz: We could, but then the priorities would be wrong while providers are being converted.

pajlada: That would depend on which order we would convert the providers in.
With the current code, if we convert one at a time, if we convert BTTV or 7TV first then the order is wrong
With my suggestion, if we convert one at a time, if we convert FFZ or BTTV first then the order is wrong.
So no matter what we do, we have to be either careful with the conversion order if we convert one at a time, or convert all 3 builtin providers at the same time.

Essentially, what I want to make sure of with this is that future plugin providers cannot override our builtin emote order. So when the API is finalized and plugins can provide emote providers, it should still be FrankerFaceZ Channel -> BetterTTV Channel -> 7TV Channel -> Plugin-provided channel emotes

Nerixyz: How plugins play into this is mostly orthogonal, because they will use the priorities in `EmoteProvider`. There, we can make sure they set at least priority X.

My plan is:

- This PR
- Migration of FFZ, BTTV, 7TV
- Some plugin interface to _get_ these emotes from existing providers (e.g. search for emote by name)
- A plugin interface to create new emote providers

pajlada: That sounds reasonable, but I don't understand this part:

> We could, but then the priorities would be wrong while providers are being converted.

Nerixyz: Sorry, I never mentioned that I did these migrations already in the order FFZ, BTTV, 7TV. So in that scenario, placing the new emote provider source at the bottom would temporarily yield "BTTV > 7TV > FFZ" and "7TV > FFZ > BTTV".

Of course, it's possible to migrate the other way around.

src/controllers/emotes/EmoteHolder.hpp
Nerixyz: Renamed to `ProviderData` since it's data for a provider in a channel.

src/controllers/commands/builtin/twitch/SendWhisper.cpp
```diff
@@ -134,8 +135,16 @@ bool appendWhisperMessageWordsLocally(const QStringList &words)
...
             }
         }  // bttv/ffz emote
+        {  // third party emotes
+            emote = emoteController->resolveGlobal(EmoteName{words[i]});
+            if (emote)
+            {
+                b.emplace<EmoteElement>(*emote, MessageElementFlag::Emote);
+                continue;
+            }
+        }  // third party emotes
```
pajlada: This order is inconsistent with the one in MessageBuilder

src/controllers/emotes/EmoteController.hpp
```diff
@@ -16,12 +25,24 @@ class EmoteController
 
     virtual void initialize();
 
+    EmoteProviderPtr findProviderByName(QStringView name) const;
+    EmoteProviderPtr findProviderByID(QStringView id) const;
```
pajlada: What is the use case for these? I don't see them used in this PR

src/controllers/emotes/ChannelEmotes.hpp
```diff
@@ -0,0 +1,72 @@
...
+/// channel. It stores the emotes for each provider and takes care of
+/// initializing and reloading them. Providers are sorted by their priority.
+class EmoteHolder
+{
+public:
+    /// An entry for a single provider
+    struct ProviderData {
+        std::weak_ptr<EmoteProvider> provider;
+        QString id;
+        EmoteMapPtr emotes;
```
pajlada: Are these channel emotes?

src/controllers/emotes/ChannelEmotes.hpp
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    EmoteHolder(Channel *channel);
+
+    void initialize(const EmoteController &controller);
+
+    /// Get an emote by its name.
+    ///
+    /// If there's no emote, an empty shared_ptr is returned.
+    EmotePtr resolve(const EmoteName &query) const;
```
pajlada, line 55: Does this resolve channel emotes only?

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -341,6 +341,11 @@ class TwitchChannel final : public Channel, public ChannelChatters
     const QString &getDisplayName() const override;
     void updateDisplayName(const QString &displayName);
 
+    /**
+     * Returns the localized name of the user
+     **/
+    const QString &getLocalizedName() const override;
+
```
pajlada: What was this change needed for?

src/controllers/emotes/EmoteHolder.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto channelSP = channel->shared_from_this();
+    if (!channelSP)
+    {
+        assert(false && "Channel without shared_from_this");
+        return;
+    }
+
+    provider->loadChannelEmotes(
+        channelSP,
+        weakGuarded(
```
pajlada: I don't think this increases readability

src/controllers/emotes/EmoteHolder.cpp
```diff
@@ -0,0 +1,188 @@
...
+                    }
+                }
+
+                if (!message.isEmpty())
+                {
+                    chan->addSystemMessage(message);
+                }
+            },
+            channelSP),
+        EmoteProvider::LoadChannelArgs{.manualRefresh = manualRefresh});
```
pajlada:
```suggestion
        EmoteProvider::LoadChannelArgs{.manualRefresh = manualRefresh,});
```

src/messages/MessageBuilder.cpp
```diff
@@ -420,13 +422,24 @@ EmotePtr makeSharedChatBadge(const QString &sourceName,
...
 
+    if (auto *holder = twitchChannel->emotes())
+    {
+        auto channel = holder->resolve(name);
+        if (channel)
+        {
+            return channel;
+        }
+    }
+
```
pajlada: This should be grouped with the other channel emote parsing code

src/controllers/emotes/EmoteProvider.hpp
```diff
@@ -0,0 +1,78 @@
...
+        return this->id_;
+    }
+
+    uint32_t priority() const
+    {
+        return this->priority_;
+    }
+
+    virtual bool hasChannelEmotes() const = 0;
+    virtual bool hasGlobalEmotes() const = 0;
```
pajlada, line 65: This is pure virtual yet we have an implementation for it in the source file

src/messages/Emote.hpp
```diff
@@ -59,7 +59,9 @@ class EmoteMap : public std::unordered_map<EmoteName, EmotePtr>
                                        const QString &emoteID) const;
 };
 
-inline const std::shared_ptr<const EmoteMap> EMPTY_EMOTE_MAP = std::make_shared<
+using EmoteMapPtr = std::shared_ptr<const EmoteMap>;
```
pajlada: Undo this for now please

src/common/Channel.hpp
```diff
@@ -122,6 +124,9 @@ class Channel : public std::enable_shared_from_this<Channel>, public MessageSink
 
     MessageSinkTraits sinkTraits() const final;
 
+    EmoteHolder *emotes();
+    const EmoteHolder *emotes() const;
```
pajlada: I'd prefer if emote logic was constrained to `TwitchChannel` since that's where all the current emote logic lives

pajlada (changes requested):
incomplete review / questions

src/messages/Emote.hpp
Nerixyz: I feel like this makes it less readable...

src/controllers/emotes/ChannelEmotes.hpp
Nerixyz: I renamed the outer class to `ChannelEmotes`. That should make things clearer.

src/controllers/emotes/ChannelEmotes.hpp
Nerixyz: See above.

src/providers/twitch/TwitchChannel.hpp
Nerixyz: `getLocalizedName` is incorrectly `private`, but wasn't needed here.

PR #6571, opened 2025-11-08
feat: seperate watchstreak highlights from sub highlights

devPoland:
This simply seperates watchstreaks alerts, so they dont look like sub alerts, and so the user can also set the color to whatever they want


src/controllers/highlights/HighlightModel.hpp
```diff
@@ -30,11 +30,12 @@ class HighlightModel : public SignalVectorModel<HighlightPhrase>
         SelfHighlightRow = 0,
         WhisperRow = 1,
         SubRow = 2,
-        RedeemedRow = 3,
-        FirstMessageRow = 4,
-        ElevatedMessageRow = 5,
-        ThreadMessageRow = 6,
-        AutomodRow = 7,
+        WatchStreakRow = 3,
```
pajlada: Could you move the `WatchStreakRow` to the end, so below `AutomodRow` instead?

src/controllers/highlights/HighlightModel.cpp
```diff
@@ -125,6 +125,22 @@ void HighlightModel::afterInit()
 
     this->insertCustomRow(subRow, HighlightRowIndexes::SubRow);
 
+    std::vector<QStandardItem *> watchStreakRow = this->createRow();
```
pajlada: Move this block of code to underneath where the automod row is created

src/singletons/Settings.hpp
```diff
@@ -526,6 +526,13 @@ class Settings
                                            ""};
     QStringSetting subHighlightColor = {"/highlighting/subHighlightColor", ""};
 
+    BoolSetting enableWatchStreakHighlight = {
+        "/highlighting/watchStreakHighlight/highlighted", true
+    };
+    QStringSetting watchStreakHighlightColor = {
+        "/highlighting/watchStreakHighlightColor", ""
```
pajlada: Please change the setting path to `/highlighting/watchStreak/color`

src/singletons/Settings.hpp
```diff
@@ -526,6 +526,13 @@ class Settings
                                            ""};
     QStringSetting subHighlightColor = {"/highlighting/subHighlightColor", ""};
 
+    BoolSetting enableWatchStreakHighlight = {
+        "/highlighting/watchStreakHighlight/highlighted", true
```
pajlada: Please change the setting path to `/highlighting/watchStreak/enabled`

pajlada (changes requested):
Just some nits, otherwise this looks fine.

Looks like you might need to run the formatter too, let me know if you need any help with that.

devPoland:
Do these changes look good?

src/controllers/highlights/HighlightModel.hpp
```diff
@@ -30,12 +30,12 @@ class HighlightModel : public SignalVectorModel<HighlightPhrase>
         SelfHighlightRow = 0,
         WhisperRow = 1,
         SubRow = 2,
-        WatchStreakRow = 3,
         RedeemedRow = 4,
         FirstMessageRow = 5,
         ElevatedMessageRow = 6,
         ThreadMessageRow = 7,
         AutomodRow = 8,
+        WatchStreakRow = 3,
```
pajlada: Number should be updated to reflect that it comes after the automod row

devPoland:
I changed the indexes back to how they were, I changed those in my first commit, so now I've reverted them

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -783,7 +783,15 @@ void IrcMessageHandler::parseUserNoticeMessageInto(Communi::IrcMessage *message,
                 parseTagString(messageText), tags,
                 calculateMessageTime(message).time(), channel);
 
-            msg->flags.set(MessageFlag::Subscription);
+            if (msgType == "viewermilestone")
```
Nerixyz, line 786: This is always false.

Nerixyz (commented):
Could you add a test in https://github.com/Chatterino/chatterino2/tree/8a58fcf0dfa26b44b2026758741a737b7ffd96c8/tests/snapshots/IrcMessageHandler?

Should be pretty similar to the `sub-message.json` and `sub-no-message.json`.

src/providers/twitch/IrcMessageHandler.cpp
pajlada: Do you want to fix this & add a unit test in a follow-up PR @devPoland? If not we can do it later :)

devPoland: I see what you mean, this check should not have been put inside the 
msgType == "subgift"
check, this is why this returned false, when I built chatterino2 with my changes it worked, so I assumed it was all fine. Thank you for pointing this out, I can commit this right now without the unit test if thats okay?

PR #6577, opened 2025-11-13
fix: middle-clicking username in local channels

Julia-Roman:
Middle-clicking usernames in a local channel would open an invalid viewercard link (e.g. `https://www.twitch.tv/popout//mentions/viewercard/pajlada`).

This PR fixes this by either opening a viewercard in the source channel (for `/mentions`), or just opening the Twitch channel (in `/whispers`, `/live`)

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -2360,7 +2360,9 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
                     MessageElementFlag::Username))
             {
                 const auto userName = hoverLayoutElement->getLink().value;
-                const auto type = this->channel_->getType();
+                const auto type = this->hasSourceChannel()
+                                      ? this->sourceChannel_->getType()
+                                      : this->channel_->getType();
                 switch (type)
                 {
                     case Channel::Type::TwitchWhispers:


pajlada (changes requested):
The fix to `/mentions` and `/live` are fine, the fix to `/whispers` is incomplete.
Inline whispers used to open the correct viewercard/channelwheretheinlinewhisperwasreceived/...` but now opens `viewercard//whispers/...`

Feel free to force push if you prefer since the changes are small.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2359,8 +2359,24 @@ void ChannelView::mouseReleaseEvent(QMouseEvent *event)
...
+                const auto userName = hoverLayoutElement->getLink().value;
+                const auto type = this->channel_->getType();
+                switch (type)
+                {
+                    case Channel::Type::TwitchWhispers:
+                    case Channel::Type::TwitchLive:
+                        QDesktopServices::openUrl(
+                            QUrl(u"https://www.twitch.tv/" % userName));
+                        break;
+                    case Channel::Type::TwitchMentions:
```
Nerixyz, line 2373: Unrelated: we should probably rename this to just `Mentions`.

PR #6579, opened 2025-11-14
fix: add Unicode license

pajlada:

### DIFF
--- a/resources/licenses/unicode.txt
+++ b/resources/licenses/unicode.txt
@@ -1,0 +1,39 @@
+UNICODE LICENSE V3
+
+COPYRIGHT AND PERMISSION NOTICE
+
+Copyright © 1991-2025 Unicode, Inc.
+
+NOTICE TO USER: Carefully read the following legal agreement. BY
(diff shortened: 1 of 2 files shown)


Nerixyz:
Where do we use Unicode directly? We don't directly depend on icu and the emoji data license is already included.

If we were to include all transitive licenses, we'd also need to include something like zlib which is one of Qt's dependencies.

pajlada:
> Where do we use Unicode directly? We don't directly depend on icu and the emoji data license is already included.

If I understand things correctly, we need to include licenses for anything we include statically in our program. This includes Unicode since emoji-data's json blob is based off of Unicode data.
I think we would be fine if the emoji-data license already contained the Unicode license blurb, but since it doesn't I added it as a standalone inclusion.

> If we were to include all transitive licenses, we'd also need to include something like zlib which is one of Qt's dependencies.

This is not what I'm aiming to do. If someone with more open source legal requirements can provide more insight on what we should and should not do, I'll follow those directions.

Nerixyz:
> If I understand things correctly, we need to include licenses for anything we include statically in our program.

Thank you, this would be a great description.

PR #6594, opened 2025-11-21
feat(plugins): add display name change event to `Channel`

Nerixyz:
The main purpose of this PR is to introduce signals to plugins. We should be able to pass (wrapped) Lua callbacks to our C++ signals and account for two things: signals outliving plugins and plugins outliving signals.

For the first one (signals outliving plugins), I added a `PluginWeakRef` and `PluginRef` (wrappers around `std::shared_ptr<Plugin>`). Plugins are still in a `unique_ptr` and the state is still allocated by Lua.
Signals are given a `SignalCallback` which holds a Lua function and a `PluginWeakRef`. Before invoking the callback, we check that the plugin is still alive. Similarly, when we destroy a callback, we only destroy the Lua function if the plugin is still alive.

The second one (plugins outliving signals) is handled by the signal libraries that hand out connection handles. The handles are still valid when the signal is destroyed.

Some things we need to watch out for in the future:
- Signals might be invoked recursively. For example, a "message added" signal handler could add another message which would invoke the same signal again.
- Deletion signals might expose objects in some partial state. For example, `Channel::destroyed` is currently called in the destructor of `Channel`, so accessing any `TwitchChannel` fields is UB.
- Invoking signals might destroy objects. If plugins can close splits, they might do that when a message is added to a channel. This would also destroy the channel while it's in the middle of handling a signal. We could work around this by destroying the split in the next event loop tick.


src/controllers/plugins/PluginRef.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if (this->shared.use_count() <= 1)
+    {
+        this->shared.reset();
+        return;
+    }
+
+    auto *app = getApp();
+    if (app && !app->isTest())
+    {
+        QMessageBox::critical(
```
pajlada: as discussed, please soft fail for users & assert for devs

pajlada (approved):
CC @Mm2PL

docs/lua-meta/globals.lua
```diff
@@ -181,6 +187,35 @@ function c2.Channel.by_twitch_id(id) end
...
 -- End src/controllers/plugins/api/ChannelRef.hpp
 
+-- Begin src/controllers/plugins/api/ConnectionHandle.hpp
+
+
+
+---@class c2.ConnectionHandle
+---A handle to a connection.
+---This handle does not automatically disconnect a signal by itself.
+---Destroying/closing it will not have any effect.
```
Mm2PL: Does this refer to the situation of the ConnectionHandle outliving the signal?

src/controllers/plugins/SignalCallback.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#ifdef CHATTERINO_HAVE_PLUGINS
+
+#    include "controllers/plugins/PluginRef.hpp"
+#    include "controllers/plugins/SolTypes.hpp"
+#    include "debug/AssertInGuiThread.hpp"
+
+namespace chatterino::lua {
+
+struct SignalCallback {
```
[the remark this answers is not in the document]
Nerixyz: Ugh, the copy-ctor that the compiler generates is already enough. If I add that, another clang-tidy rule will complain about that one not being `= default`.

docs/chatterino.d.ts
```diff
@@ -46,12 +46,22 @@ declare namespace c2 {
...
+        is_connected(): boolean;
+    }
+
     class Channel implements IWeakResource {
         is_valid(): boolean;
         get_name(): string;
         get_type(): ChannelType;
         get_display_name(): string;
 
+        on_display_name_changed(): ConnectionHandle;
```
Mm2PL: This needs to be updated to take a callback function

docs/lua-meta/globals.lua
```diff
@@ -244,6 +244,12 @@ function c2.Channel:is_vip() end
 ---@return string
 function c2.Channel:__tostring() end
 
+--- Callback when the channel display name changes.
+---
+---@param cb fun()
+---@return c2.ConnectionHandle hdl
+function c2.Channel:on_display_name_changed(cb) end
```
Mm2PL, line 251: Prose documentation for this please

docs/lua-meta/globals.lua
```diff
@@ -264,6 +270,35 @@ function c2.Channel.by_twitch_id(id) end
 
 -- End src/controllers/plugins/api/ChannelRef.hpp
 
+-- Begin src/controllers/plugins/api/ConnectionHandle.hpp
+
+
+
+---@class c2.ConnectionHandle
```
Mm2PL, line 277: Would be nice if this had prose docs too

src/controllers/plugins/PluginRef.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    return this->shared.get();
+}
+
+void PluginRef::destroy()
+{
+    auto useCount = this->shared.use_count();
+    if (useCount > 1)
+    {
+        qCWarning(chatterinoLua)
```
Mm2PL, line 40: Off topic: Weird that we use `Warning` category for something that we assert never happens all over the code base. It'd be cool to change that

Nerixyz: I had a hard-fail here before (https://github.com/Chatterino/chatterino2/pull/6594#discussion_r2552923997), but we usually use soft-fails. Having a warning can be helpful if Chatterino crashes as a result of this, because on some systems, the logs are captured in a crashdump.

pajlada: Yeah, having logs in these scenarios is nice.
@Mm2PL Are you thinking we should use something other than warn in situations like this?

Mm2PL: First of all using a "Warning" does not feel right, this should probably be a Critical assuming those are always logged. Ideally these situations should produce a message visible to the user with a description of what could have led to Chatterino to crash. Does crashpad have utilities to tell it any of this?

Nerixyz: > Does crashpad have utilities to tell it any of this?

Not that I know of. That's mainly because chrashpad is a separate process. You can only annotate the whole program execution with some initial data.

2547techno:
excited for signal support :)

PR #6598, opened 2025-11-22
Add categories to emoji viewer

treejadey:
Adds category separation to the emoji viewer

Before:
image

After:
image


src/providers/emoji/Emojis.cpp
```diff
@@ -219,6 +221,8 @@ void Emojis::loadEmojis()
                 parseEmoji(variationEmojiData, variation,
                            emojiData->shortCodes[0] + "_" + toneName);
 
+                variationEmojiData->category = emojiData->category;
```
pajlada, line 226: I would love a comment here stating that since this is a variation of the "base" emoji, we inherit the base emoji category

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -164,11 +165,45 @@ void addTwitchEmoteSets(const std::shared_ptr<const EmoteMap> &local,
 
 void loadEmojis(ChannelView &view, const std::vector<EmojiPtr> &emojiMap)
 {
+    std::map<QString, std::vector<EmojiPtr>> emoteCategoryMap;
```
pajlada: This categorization is currently done every time the emote popup is opened.
We could either store this in the Emojis class, or do this once on first emote popup by storing this statically with something like this:

```
diff --git a/src/widgets/dialogs/EmotePopup.cpp b/src/widgets/dialogs/EmotePopup.cpp
index 933dda3fb..56cbc870a 100644
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -165,22 +165,26 @@ void addTwitchEmoteSets(const std::shared_ptr<const EmoteMap> &local,
 
 void loadEmojis(ChannelView &view, const std::vector<EmojiPtr> &emojiMap)
 {
-    std::map<QString, std::vector<EmojiPtr>> emoteCategoryMap;
+    static auto emoteCategoryMap = [] {
+        std::map<QString, std::vector<EmojiPtr>> emoteCategoryMap;
 
-    for (const auto &emoji : emojiMap)
-    {
-        auto cat = emoteCategoryMap.find(emoji->category);
-        if (cat != emoteCategoryMap.end())
+        for (const auto &emoji : emojiMap)
         {
-            auto &vec = cat->second;
-            vec.push_back(emoji);
+            auto cat = emoteCategoryMap.find(emoji->category);
+            if (cat != emoteCategoryMap.end())
+            {
+                auto &vec = cat->second;
+                vec.push_back(emoji);
+            }
+            else
+            {
+                emoteCategoryMap.emplace(emoji->category,
+                                         std::vector<EmojiPtr>{emoji});
+            }
         }
-        else
-        {
-            emoteCategoryMap.emplace(emoji->category,
-                                     std::vector<EmojiPtr>{emoji});
-        }
-    }
+
+        return emoteCategoryMap;
+    }();
 
     ChannelPtr emojiChannel(new Channel("", Channel::Type::None));
     // set the channel first to make sure the scrollbar is at the top

```


I think the static version is ok for now, but we should be aware of the fact that it does increase our memory usage permanently rather than the current solution which is temporary. This might be premature optimization?

pajlada (commented):
Overall seems good - functionally works well.

src/providers/emoji/Emojis.cpp
```diff
@@ -219,6 +221,10 @@ void Emojis::loadEmojis()
                 parseEmoji(variationEmojiData, variation,
                            emojiData->shortCodes[0] + "_" + toneName);
 
+                // NOTE: After parsing an emoji variation, it's category gets set to ""
+                // We have to manually inherit the category from the original emojiData.
```
Nerixyz: Makes it sound like we're taking/moving the category.

```suggestion
                // NOTE: Emoji variations don't have a "catgory".
                // We have to manually inherit the category from the original emojiData.
```

PR #6600, opened 2025-11-24
feat: provide readable nightly versions

Nerixyz:
Currently, you have to look at the commit hash Chatterino was built with to find out which nightly version you're on. To compare two versions, you'd need to check each hash.

With this PR, the nightly versions get an increasing number starting from the last version tag. This is appended to the displayed version in the about page. It's also shown in the nightly release description. The format is "{tag-name}-{nightly-commits}+{fork-commits}". The last two parts are hidden if they're 0. Forks will need to fetch upstream to compute this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 - Dev: Mock headers are now added as a header set if supported by CMake. (#6561)
 - Dev: Set settings directory to temporary one used in tests. (#6584)
 - Dev: Check Lua unwinding and version in tests. (#6586)
+- Dev: Nightly versions are now numbered. (#6600)
 
 ## 2.5.4
 


pajlada:
I don't really understand the purpose of this.
The commit hash provides an immediate link to what commit hash your version is built with.
What does the number provide, other than something that's maybe easier for a user to understand?

Nerixyz:
> What does the number provide, other than something that's maybe easier for a user to understand?

It is mainly for users. However, I also see value besides this. If we're thinking long-term, plugins could make use of this. If we have some plugin distribution, then users will inevitably access this with different versions - including nightly. Having the possibility to query the host for "I want feature X which was added in A.B.C-D" could be helpful in that case.

pajlada:
The plugin thing is a good point - maybe we could version our APIs / modules and allow plugins to specify minimum requirements, or for run-time checks
```json
"requires": [
  "http=^1.0.0",
  "websocket=1.0.3"
]
```

```lua
if chatterino.provides("http=^1.0.0") then
  chatterino.new_request("bip")
elif chatterino.provides("http=^2.0.0") then
  chatterino.make_http_request_with_new_api("bap")
end
```

This could have the side-benefit of being able to build Chatterino with limited plugin support

PR #6604, opened 2025-11-27
Fix crash in notifyTwitchChannelOffline when snapshot is empty

vxtk:
Add guard clause to prevent accessing an empty LimitedQueueSnapshot in NotificationController::notifyTwitchChannelOffline.

Fixes: Assertion failure in stl_vector.h operator[] when snapshot is empty


pajlada:
As a first-time contributor, you are now eligible to be listed in the About page in Chatterino.

PR #6583, opened 2025-11-16
feat: add poll and prediction commands for broadcasters

iProdigy:
Work towards #2739 and #2740

This PR includes `/poll` and `/prediction` commands. Follow-up PRs will cover `/cancelpoll`, `/endpoll`, `/cancelprediction`, `/lockprediction`, `/completeprediction`

These commands are broadcaster-only due to Twitch API restrictions and these initial PRs focus on command-based interaction (GUI can be attempted later)

Sample usage:
* `/prediction --title "what lang will i use today" --duration 5m --choice "golang" --choice "kotlin" --choice "rust"`
* `/poll --title "what game to play today" --duration 60 --choice "expedition 33" --choice "eu5" --choice "disco elysium" --points 69`

Relevant uservoices:
* https://twitch.uservoice.com/forums/310213-developers/suggestions/43347684-allow-channel-moderators-to-control-polls-predicti
* https://twitch.uservoice.com/forums/310213-developers/suggestions/43371027-allow-getting-realtime-updates-on-polls-prediction

Issue #2739, opened 2021-05-04
Twitch Polls in Chatterino


gempir:
**Describe your issue**
Twitch just announced new API Endpoints for getting, ending and creating polls.

https://discuss.dev.twitch.tv/t/announcing-apis-and-eventsub-for-polls-and-predictions/31539

It would be really helpful to atleast get, end and create polls directly from Chatterino. I doubt Twitch will offer an API for voting in polls, but this issue isn't about voting anyway, the benefit is big enough IMO.

Mm2PL:
event sub over websocket when ResidentSleeper

zneix:
Please see https://github.com/Chatterino/Chatterino2/issues/2740#issuecomment-832218687

Eric1212:
I would say that polls missing from chatterino is sad to me. Hoping that some day we will find a direct access or a work around.

Jaid:
I would like to be able to recommend Chatterino to more streamer friends, but missing polls is what prevents them from replacing Twitch chat with Chatterino.

Felanbird:
> I would like to be able to recommend Chatterino to more streamer friends, but missing polls is what prevents them from replacing Twitch chat with Chatterino.

Take it up with Twitch, the API mentioned in this issue is completely useless for Chatterino, the reason we don't have Polls in Chatterino is purely on Twitch, hence the `blocked` label.

pajlada:
Scopes have will now be requested in https://github.com/Chatterino/website/pull/48 and https://github.com/Chatterino/website/pull/49

While the APIs are not super relevant for the average user, having the ability to view poll or prediction information as a streamer may come in handy and it's not a risky scope to add even if we're not using it right now.

BishopT:
same comment as in #2740 but for poll:

https://dev.twitch.tv/docs/api/polls/#:~:text=the%20endpoint%20requires%20a%20user%20access%20token%20with%20scope%20channel%3Amanage%3Apolls.%20this%20scope%20works%20for%20creating%2C%20ending%2C%20and%20getting%20polls

Felanbird:
> same comment as in #2740 but for poll:

same reply

Mm2PL:
23 lip 2023 10:01:53 tikodeur ***@***.***>:


>Is it the same issue for simpler accesses, like Get Polls or Get Predictions that only require channel:read tokens, or is it an overall user token issue? Asking from a viewer's perspective, not a broadcaster one.


[...]

PR #6605, opened 2025-11-27
feat: add /endpoll and /cancelpoll commands for broadcasters

iProdigy:
Follow-up to #6583; works towards #2739

* `/cancelpoll`: immediately hides a currently active poll
* `/endpoll`: prevents further voting on an active poll and leaves it visible for users to see the result

Separate PR will tackle the analogues for predictions


src/controllers/commands/builtin/twitch/Poll.cpp
```diff
@@ -67,4 +67,140 @@ QString createPoll(const CommandContext &ctx)
...
+    {
+        const QString err =
+            "The /endpoll command only works in Twitch channels";
+        if (ctx.channel != nullptr)
+        {
+            ctx.channel->addSystemMessage(err);
+        }
+        else
+        {
+            qCWarning(chatterinoCommands) << "Error parsing command:" << err;
```
pajlada: No parsing has been done, so this precheck can be changed a bit

src/controllers/commands/builtin/twitch/Poll.cpp
```diff
@@ -67,4 +67,140 @@ QString createPoll(const CommandContext &ctx)
...
+    {
+        const QString err =
+            "The /cancelpoll command only works in Twitch channels";
+        if (ctx.channel != nullptr)
+        {
+            ctx.channel->addSystemMessage(err);
+        }
+        else
+        {
+            qCWarning(chatterinoCommands) << "Error parsing command:" << err;
```
pajlada: Same here - we don't really do parsing, so this can be made simpler

src/controllers/commands/builtin/twitch/Poll.cpp
```diff
@@ -67,4 +67,140 @@ QString createPoll(const CommandContext &ctx)
...
+                        if (choice.votes > winner.votes)
+                        {
+                            winner = choice;
+                        }
+                    }
+                    const double percent =
+                        100.0 * winner.votes / std::max(totalVotes, 1);
+
+                    channel->addSystemMessage(
+                        QString("Ended poll: '%1' - '%2' had %3 votes (%4%)")
```
pajlada: Specify that %2 won with %3 votes, to clarify that it won.

Also handle the case where multiple options had the same number of votes

pajlada (changes requested):
Overall works well, cool!

pajlada (approved):

PR #6611, opened 2025-11-29
fix: wayland tooltip vertical positioning

vxtk:
Fixes an issue where emote tooltips would close instantly when spawning under the mouse cursor. Pretty sure the code was already supposed to move the tooltip above the cursor if it couldn't fit, but this commit makes it actually work on Wayland. Pretty sure this is a wayland only issue since it worked in xwayland before. Haven't tested on Windows, but I don't see a reason why it wouldn't work there.

**Difference:**

https://github.com/user-attachments/assets/c051bfb5-3ff6-4f46-a704-f2f3274a5a3d

### DIFF
--- a/src/util/WidgetHelpers.cpp
+++ b/src/util/WidgetHelpers.cpp
@@ -17,8 +17,6 @@ QPoint applyBounds(QScreen *screen, QPoint point, QSize frameSize)
 
     const QRect bounds = screen->availableGeometry();
 
-    // If the hover frame would go off the bottom of the screen it should be moved above the cursor
-
     if (point.x() < bounds.left())
     {
         point.setX(bounds.left());


src/util/WidgetHelpers.cpp
```diff
@@ -17,8 +17,12 @@ QPoint applyBounds(QScreen *screen, QPoint point, QSize frameSize, int height)
 
     const QRect bounds = screen->availableGeometry();
 
-    bool stickRight = false;
-    bool stickBottom = false;
+    // If the hover frame would go off the bottom of the screen it should be moved above the cursor
+    if (point.y() + frameSize.height() > bounds.bottom())
```
Nerixyz: Isn't this the same check as in line 39? If so, line 41 (`point.setY(bounds.bottom() - frameSize.height())`) would win.

Nerixyz:
There was also #6451 which fixed some stuff for Wayland.

src/util/WidgetHelpers.cpp
vxtk: yeah you are right idk why it worked anyway

vxtk:
i run the latest git version and that fix didnt work for me on hyprland

pajlada:
This PR in its current state doesn't change anything for me running Sway. This might be a hyprland-specific issue?

vxtk:
It seems like on Hyprland the tooltip never goes above the cursor and the flag `Qt::WA_TransparentForMouseEvents` that was added in #6451 is not respected.

Checked on Sway myself and it worked fine there.

After checking this again I realized that this is basically a refactor and I don't know why this fixes the issue on Hyprland. Closing this for now to investigate Hyprland directly.

PR #6613, opened 2025-11-30
refactor: split some highlight structs away from HighlightController

pajlada:
- move HighlightCheck to its own file
- move HighlightResult to its own file

and a little unrelated docker-compose file


src/controllers/highlights/HighlightCheck.hpp
```diff
@@ -0,0 +1,29 @@
...
+class Badge;
+
+enum class MessageFlag : std::int64_t;
+using MessageFlags = FlagsEnum<MessageFlag>;
+
+struct HighlightCheck {
+    using Checker = std::function<std::optional<HighlightResult>(
+        const MessageParseArgs &args, const std::vector<Badge> &badges,
+        const QString &senderName, const QString &originalMessage,
+        const MessageFlags &messageFlags, bool self)>;
```
Nerixyz, line 25: Not introduced here, but `messageFlags` can be passed by value.

PR #6619, opened 2025-12-04
fix: wording of "close visible tabs to X" for vertical tabs

### LINKED ISSUE #6618: Close multiple tabs submenu doesn't consider vertical tab layouts
### Checklist

### Describe your issue

The close multiple tabs submenu says "Close Visible Tabs to Left/Right" even when the tab layout is set to right/left which makes the tabs vertical.
It should change to "Close Visible Tabs to Top/Bottom" when a vertical layout is selected

see #6515 

### Screenshots

Image

### OS and Chatterino Version

Chatterino Nightly 2.5.4 (commit 2d446db2) built on 2025-12-01 with Qt 6.7.1, MSVC 194435219, Crashpad Running on Windows 10 Version 22H2, kernel: 10.0.19045

jupjohn:
This PR aims to "_fix_" the wording of "Close Visible Tabs to {Left/Right}" menu items when using a vertical tab direction. It also updates the menu when the setting is changed.

Demo on 2c93376a777b66ca98b03461a5b2fe3a6ce2edb4:

https://github.com/user-attachments/assets/153818f7-a7bd-40b4-adea-ca76de4ffcb5

Closes #6618.

### DIFF
--- a/src/widgets/helper/NotebookTab.cpp
+++ b/src/widgets/helper/NotebookTab.cpp
@@ -127,7 +127,6 @@ NotebookTab::NotebookTab(Notebook *notebook)
     this->closeMultipleTabsMenu_ = new QMenu("Close Multiple Tabs", this);
 
     const auto tabDirection = getSettings()->tabDirection.getEnum();
-    this->recreateCloseMultipleTabsMenu(tabDirection);
     this->menu_.addMenu(closeMultipleTabsMenu_);
     getSettings()->tabDirection.connect(
         [this](int val) {


src/widgets/helper/NotebookTab.cpp
```diff
@@ -125,9 +125,53 @@ NotebookTab::NotebookTab(Notebook *notebook)
                           });
 
     this->closeMultipleTabsMenu_ = new QMenu("Close Multiple Tabs", this);
+
+    const auto tabDirection = getSettings()->tabDirection.getEnum();
+    recreateCloseMultipleTabsMenu(tabDirection);
```
Nerixyz:
```suggestion
    this->recreateCloseMultipleTabsMenu(tabDirection);
```

src/widgets/helper/NotebookTab.cpp
```diff
@@ -125,9 +125,53 @@ NotebookTab::NotebookTab(Notebook *notebook)
...
+    this->menu_.addAction(highlightNewMessagesAction_);
+
+    this->menu_.addSeparator();
+
+    this->notebook_->addNotebookActionsToMenu(&this->menu_);
+}
 
-    closeMultipleTabsMenu_->addAction("Close All Visible Tabs", [this]() {
+void NotebookTab::recreateCloseMultipleTabsMenu(
+    const NotebookTabLocation tabDirection)
```
Nerixyz: The name `tabDirection` is confusing, because this is the tab location. For example, when the setting is `NotebookTabLocation::Top`, the direction is left-to-right.

It could also be nice to remove this parameter and read the setting here.

jupjohn: I was in the mindset of direction instead of location when writing this, will fix!

As for the parameter name: the setting's update signal passes through the new enum's value, so might as well pass that through since it's already there. If you still think it's a nicer unit having a setting lookup in this method, I'm happy to change that :+1:

jupjohn: Renamed parameter in 551b921d

src/widgets/helper/NotebookTab.cpp
jupjohn: Added in f9c4c372

Issue #6358, opened 2025-07-25
Application resets all data every few days


cqmpact:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Every few days the Chatterino app will:
- Log me out of all my accounts
- Delete ALL of my custom settings
- Only the channels in tabs stay

### OS and Chatterino Version

Chatterino 2.5.3 (commit 3f3a31db) built with Qt 6.7.1, MSVC 194334808, Crashpad Running on Windows 11 Version 24H2, kernel: 10.0.26100

pajlada:
Hi, sorry for the inconvenience.
Could you try downloading our latest Nightly build from https://github.com/Chatterino/chatterino2/releases/tag/nightly-build and seeing if that improves things for you?

For more information on what our Nightly build is, see https://wiki.chatterino.com/Help/#what-is-nightly-and-how-to-use-install-it

cqmpact:
Thanks for the recommendation! I'll give it a shot and report back if the issue persists.

8thony:
@cqmpact we haven't heard from you in a while.

The new version 2.5.4, has been released, which also has all the improvements. 
Does your issue still occur?

cqmpact:
> [@cqmpact](https://github.com/cqmpact) we haven't heard from you in a while.
> 
> The new version 2.5.4, has been released, which also has all the improvements. Does your issue still occur?

Hi there!
Thanks for checking in - I'm still on Nightly 2.5.3 and I haven't experienced any issues whatsoever.

I'll download stable and see if it happens again.

cqmpact:
It's been almost a month now without any issues on the new build.
Will leave the issue open until EOY in case it happens again.

jupjohn:
I've had a report of a reset from a user running `2.5.4`. Chatterino was running when she shut down, and she noted that her computer wasn't starting for a bit this morning so perhaps this was Windows' problem. Settings were recovered from backups.

```
Chatterino 2.5.4 (commit 117d06b9) built with Qt 6.7.1, MSVC 194435215, Crashpad
Running on Windows 10 Version 22H2, kernel: 10.0.19045
```

[...]

PR #6620, opened 2025-12-04
fix: save settings less often

pajlada:
Currently, when a user closes Chatterino or closes the settings page, we will save their `settings.json` file regardless of any setting changes that have been made.

For `commands.json`, we ensure a change has actually happened before we save it using the settings library's `CompareBeforeSet` (set on a per-setting setting, which will compare the old & new value and only ever called the Setting Manager's set function if there was a difference) and `OnlySaveIfChanged` (set on the settings manager, which actually decides whether to sparingly save or not).

This PR applies `OnlySaveIfChanged` (in an ugly way, I could technically _just _ set it but I might want the manual check to be explicit later) to the `settings.json` manager, and notably, enables `CompareBeforeSet` on the Hotkey defaults setting. Previously we've applied `CompareBeforeSet` on all of our `ChatterinoSetting`s, but since the Hotkey defaults setting is a little bit special, we never applied this there.

To test this, you can checkout this PR and revert 923a977 and 6b823d8 then call `/debug-test save-settings` on a freshly started Chatterino and see that no matter how many times you call it, we will overwrite `settings.json` even if nothing has changed.
If you only revert 6b823d8, you'll see that repeatedly calling `/debug-test save-settings` will only actually save to the file once. The reason for this initial save is the aforementioned Hotkey default setting which didn't have `CompareBeforeSet` applied.

Related to the recent conversation in #6358 

Discussion points:
This reduces the frequency of backups. In the latest release, starting and stopping Chatterino will always create a new backup, even if nothing changed. Is this change in behaviour a problem? I think it is not a problem.


jupjohn:
> This reduces the frequency of backups. In the latest release, starting and stopping Chatterino will always create a new backup, even if nothing changed. Is this change in behaviour a problem? I think it is not a problem.

I think not saving on start/stop because nothing's changed is a sane behaviour. It would also prevent an issue which I saw with one user who had a reset, then restarted Chatterino 9+ times due to ISP issues - preventing them from getting backups pre-reset. I like this.

jupjohn (approved):
Tested, LGTM. Happy for the discussion point to be addressed in another PR, if at all :+1:

PR #6622, opened 2025-12-04
feat: reverse search previously sent messages

Nerixyz:
I like how I can search the history of my commands in the terminal. Sometimes, when chatting, I wanted to search for a message I previously sent and didn't want to hit up-arrow 10x to find it. Being able to search through the history, solves this. By default, it's bound to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>R</kbd>:

https://github.com/user-attachments/assets/8dc44993-ff27-4bac-8570-020b726ed51d


jupjohn:
Would resolve #6513

pajlada (commented):
This functionality will pair up well with forward incremental searching as well.

Bash calls it "reverse incremental search" and "forward incremental search"
zsh & powershell calls it "backwards incremental search" and forward" incremental search"

The functionality should probably be called "Incremental search history" with actions for going backwards/reverse or forward.

The bindings I and others I've consulted would want to use are CTRL+R (which is already taken) for backwards/reverse and CTRL+S for forward. In my opinion, with this being an expert feature, I'd prefer this feature not have a default key binding.
I know this comes with the downside of worse discoverability, but we can add a key binding later, but we can't easily remove a key binding after we've added it, so I'd rather be cautious.

I'm just missing forward search and the ability to disable the looping, I guess as a parameter to the action

so something like
```c++
{"incremental-search-history", ActionDefinition{
  .displayName = "Incrementally search history",
  .argumentDescription = "[forward/backward] [loop/noloop]"
  .minCountArguments = 2,
  .maxCountArguments = 2,
}
```

I'd then bind CTRL+R to `incremental-search-history backward noloop` and CTRL+S to `incremental-search-history forward noloop`

It doesn't work super well with hidden input atm - I guess invoking the action should "force show" the input

Overall, cool feature I've wanted to have for quite a while

Nerixyz:
Kinda forgot this was still a draft.

src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -268,6 +268,21 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
               .argumentsPrompt = "Behavior:",
           }},
          {"undo", ActionDefinition{"Undo"}},
+         {"incremental-search-history",
+          ActionDefinition{
+              .displayName = "Incrementally search through the input history",
+              .argumentDescription = "[forward/backward] [loop/noloop]",
+              .minCountArguments = 2,
+              .maxCountArguments = 2,
```
Mm2PL: Neither of these argumnets are optional.

```suggestion
              .argumentDescription = "<forward/backward> <loop/noloop>",
              .minCountArguments = 2,
              .maxCountArguments = 2,
```

Coldblackice:
Did this get reverted? This was an awesome feature.

Nerixyz:
> Did this get reverted? This was an awesome feature.

No, it wasn't merged yet.

src/widgets/splits/SplitInput.cpp
```diff
@@ -1119,13 +1232,42 @@ void SplitInput::editTextChanged()
...
+            QTextCursor cursor = this->ui_.textEdit->textCursor();
+            QTextCharFormat format;
+            if (getTheme()->isLightTheme())
+            {
+                format.setBackground(QColor(20, 20, 255, 80));
+            }
+            else
+            {
+                format.setBackground(QColor(110, 110, 255, 80));
+            }
```
Mm2PL: maybe this underline's color could be themed?

src/widgets/splits/SplitInput.cpp
```diff
@@ -1466,6 +1609,185 @@ void SplitInput::updateFonts()
...
+    this->editTextChanged();
+}
+
+void SplitInput::updateHistorySearchStatus(bool failed, const QString &message)
+{
+    if (failed && !this->historySearchFailed)
+    {
+        QPalette palette = this->ui_.historySearchWrap->palette();
+        palette.setColor(QPalette::Text, Qt::red);
+        palette.setColor(QPalette::WindowText, Qt::red);
```
Mm2PL: This color could also be customized I think

Mm2PL (changes requested):
apparently i didn't click submit on this oopsie

I'm not fully sure if unfocusing the window should cancel history search

src/widgets/splits/SplitInput.cpp
Nerixyz: Done - `splits.input.searchHighlightBackground`.

src/widgets/splits/SplitInput.cpp
Nerixyz: Done - `splits.input.searchFailText`.

pajlada (commented):
If "Show input when it's empty" is disabled, pressing one of the incremental search hotkeys does not bring up the input. I think it would be good if it did because you've essentially pressed a button that says: "I am going to type something!!!"

-----------------

Behavioural-wise, this behaves a bit differently to my terminal. I don't know necessarily if this is good or not, but I want to bring it up. In short, the Chatterino feature only cares about single matches in a message, not multiple matches in a message.

If I type the following commands in my terminal:

 - echo "hi hi"
 - echo "lol hi"
 - 
and I backwards-incremental-search looping for "hi", it will hop to & highlight the lol hi message:

 - echo "lol <ins>hi</ins>"

If I backwards-incremental-search again, it will hop to & highlight the _last_ hi in the first message

 - echo "hi <ins>hi</ins>

Again, and it stays on the same message but moves to the first hi

 - echo "<ins>hi</ins> hi"

Again, and it jumps back to the "lol hi" message again

 - echo "lol <ins>hi</ins>"

If I now forwards-incremental-search, it keeps the same word highlighted but moves the cursor from the start of the matched word to the end of the matched word.


With this Chatterino feature, it always highlights the first match regardless of forward or backward-search, and 


---------------------

Another behavioural thing I've noticed that I don't know if it's correct or not without testing it more.

In my terminal, the cursor never actually leaves the message. If I backwards-incremental-search to a message and press the arrow keys to move my cursor, it moves the cursor in the message itself rather then search input.

Nerixyz:
> If "Show input when it's empty" is disabled, pressing one of the incremental search hotkeys does not bring up the input. I think it would be good if it did because you've essentially pressed a button that says: "I am going to type something!!!"

Should be fixed now.

> In short, the Chatterino feature only cares about single matches in a message, not multiple matches in a message.

I based my implementation on the behavior of PowerShell/PSReadLine. There, each entry only shows up once in the search. I can reproduce your behavior in zsh, though. Not sure what we should do. I think the current implementation is simpler, because it only needs to change the index in a list of filtered entries whereas with the zsh behavior, you'd also need to keep track of the position in the entry.

> In my terminal, the cursor never actually leaves the message. If I backwards-incremental-search to a message and press the arrow keys to move my cursor, it moves the cursor in the message itself rather then search input.

Yea and more importantly, it stops the search. Again, not sure if that should be done here. This could also be done in a followup PR.

pajlada (approved):

pajlada:
@Mm2PL Do you want another peek before I merge this in? There are a few things I might want to change as I noted in my comment above but I think it would be nice to have this merged in so I can actually test it easily together with other changes

Mm2PL (approved):

PR #6623, opened 2025-12-04
ci: add clazy

Nerixyz:
[Clazy](https://invent.kde.org/sdk/clazy) is a linter for Qt code from KDE. Since the newest release, 1.16, it can be loaded as a `clang-tidy` plugin. Conveniently, or because of that, the clang-tidy review action supports additional arguments to pass to `clang-tidy`.

Since 1.16 of Clazy is so new, it's not included in Ubuntu 25.10 (the image of the action). I created https://github.com/Nerixyz/clazy-ubuntu-builds to prebuild Clazy for this image.

All checks are currently enabled. We will likely need to turn off some verbose checks in the future.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,5 +1,5 @@
 freebsd_instance:
-  image_family: freebsd-14-2
+  image_family: freebsd-14-3
 
 task:
   install_script:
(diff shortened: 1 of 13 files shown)


Mm2PL:
Does this work in clangd finally?

Nerixyz:
> Does this work in clangd finally?

No, I don't think you can load clang or clang-tidy plugins yet.

pajlada (approved):

PR #6625, opened 2025-12-05
feat: show BTTV Pro badges

Nerixyz:
BTTV badges are broadcasted via their live updates, specifically in [User Update](https://betterttv.com/developers/websocket#user-update). We previously ignored that event. It's also used to broadcast personal emotes, but that depends on the emote refactor.

The badge only includes one size at 72px. Since we expect badges to be 18px, it uses the scale factor `18 / 72`.

Resolves #4323.

### DIFF
--- a/.CI/CreateUbuntuDeb.sh
+++ b/.CI/CreateUbuntuDeb.sh
@@ -20,10 +20,6 @@ deb_path="Chatterino-ubuntu-${ubuntu_release}-x86_64.deb"
 
 # Refactor opportunity:
 case "$ubuntu_release" in
-    20.04)
-        # Qt6 static-linked deb, see https://github.com/Chatterino/docker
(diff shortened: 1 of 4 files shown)


src/messages/MessageElement.hpp
```diff
@@ -105,6 +104,8 @@ enum class MessageElementFlag : int64_t {
     // - 7TV Contributor
     BadgeSevenTV = (1LL << 36),
 
+    BadgeBttv = (1LL << 6),
```
pajlada, line 109: Add comment like on the 7TV one. No need to fill out what types of badges, but that it's its own badge slot

src/providers/bttv/BttvBadges.cpp
```diff
@@ -0,0 +1,27 @@
...
+    return badgeJson["url"].toString();
+}
+
+EmotePtr BttvBadges::createBadge(const QString &id,
+                                 const QJsonObject & /* badgeJson */) const
+{
+    auto emote = Emote{
+        .name = EmoteName{},
+        .images = ImageSet(Image::fromUrl(Url{id}, 18.0 / 72.0)),
+        .tooltip = Tooltip{"BTTV Pro"},
```
pajlada, line 19:
```suggestion
        .tooltip = Tooltip{"BetterTTV Pro"},
```

pajlada (approved):

Issue #2740, opened 2021-05-04
Twitch Predictions


gempir:
**Describe your issue**
Twitch just announced new API Endpoints for getting, ending and creating Predictions!

https://discuss.dev.twitch.tv/t/announcing-apis-and-eventsub-for-polls-and-predictions/31539

It would be really helpful to at least get, end and create predictions from Chatterino, voting in them won't be possible similar to polls.

Mm2PL:
eventsub over websocket when ResidentSleeper

zneix:
All the new endpoints that were added today require Broadcaster's OAuth token, which makes our implementation not so useful. Perhaps we'd have to wait for Twitch's implementation to allow mods to modify predictions & polls

![](https://cdn.zneix.eu/iddq2Wj.png)

For more information, this was also discussed under the article gempir linked.

Mm2PL:
classic twitch fucking over the 3rd party devs ResidentSleeper

killertofus:
https://blog.twitch.tv/en/2021/05/24/polls-and-channel-points-predictions-have-leveled-up-with-twitch-api-and-eventsub-support/

BishopT:
Reading the doc as of today, it seems to me this has changed (or wrongly documented in past) : 
https://dev.twitch.tv/docs/api/predictions/#:~:text=the%20endpoint%20requires%20a%20user%20access%20token%20with%20scope%20channel%3Amanage%3Apredictions.%20this%20scope%20works%20for%20creating%2C%20ending%2C%20and%20getting%20predictions

Felanbird:
> it seems to me this has changed (or wrongly documented in past)

The documentation on the page you have linked, while not explicitly stating the limitations, does link to a page that does.

![image](https://user-images.githubusercontent.com/41973452/215907966-806bd5ae-5666-4298-bc44-99b3da0a2ad3.png)
![image](https://user-images.githubusercontent.com/41973452/215908004-7bdc7ef6-b66c-48f7-a90a-78c0010048fd.png)

Unfortunately but as expected, nothing has changed

Mm2PL:
![image](https://user-images.githubusercontent.com/25011746/215907988-c85df0ed-a5eb-46ac-b570-596386e69d11.png)
You seem to be wrong

pajlada:
As of #6583 #6612 and #6632 (and Chatterino v2.5.5), streamers can now perform these actions.

[...]

PR #6632, opened 2025-12-07
feat: add broadcaster-only /completeprediction command

iProdigy:
Follow-up to #6583; works towards #2740


src/controllers/commands/builtin/twitch/Prediction.cpp
```diff
@@ -218,4 +221,179 @@ QString cancelPrediction(const CommandContext &ctx)
     return "";
 }
 
+QString completePrediction(const CommandContext &ctx)
+{
+    const auto usage = QStringLiteral(
+        R"(Usage: /completeprediction --choice "<choice>" --index <index> - Selects a winner for an outstanding prediction. The choice title must exactly match the wording in the prediction. Alternatively, you may specify the one-based index of the winning outcome, but you may not specify both the choice name and index.)");
```
pajlada:
```suggestion
        R"(Usage: /completeprediction --choice "<choice>" or /completeprediction --index <index> - Selects a winner for an outstanding prediction. The choice title must exactly match the wording in the prediction. Alternatively, you may specify the one-based index of the winning outcome.)");
```

Issue #6637, opened 2025-12-09
Intermittent failure to capture Redeem messages


Ne0C0rt0sis:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Streamer has limited number of redeems available at the start of every stream. 
Stream properly starts once they are all redeemed.

Chatterino intermittently does not capture the redeem messages, has only become a problem since attempting to use the logging function then disabling it. Of note, enabling the logging function prevented any redeem messages from appearing in the chat window.

Have caught it in the past during the redeem period, had to restart Chatterino for it to start capturing again.

Was able to test with multiple instances at the same time and using the regular twitch chat, only the one that had logging enabled is having issues.

I am aware of Issue https://github.com/Chatterino/chatterino2/issues/1022 ,  unsure if applicable.

No screenshots available at the moment, will capture the next time it happens.

### OS and Chatterino Version

Chatterino 2.5.4 (commit 117d06b9) built with Qt 6.7.1, MSVC 194435215, Crashpad Running on Windows 10 Version 22H2, kernel: 10.0.19045

Nerixyz:
> Have caught it in the past during the redeem period, had to restart Chatterino for it to start capturing again.

I had the same problem before when resuming from a sleep.

I think we can migrate PubSub to the new WebSocket API... Would also allow us to remove websocketpp as a dependency.

Issue #6576, opened 2025-11-13
failes to build on debian


solomoncyj:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

```
cmake -B build -S . -G Ninja -DUSE_SYSTEM_QTKEYCHAIN=ON -DUSE_SYSTEM_QTKEYCHAIN=ON -DUSE_SYSTEM_QTKEYCHAIN=ON   
-- The C compiler identification is GNU 15.2.0
-- The CXX compiler identification is GNU 15.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- LTO: Disabled
-- Found Git: /usr/bin/git (found version "2.51.0")
-- Injected git values: 4a7572466ab765b231ca3cb5b98805ef2e4ab0bf (v2.5.4-61-g4a757246) modified: 1
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Performing Test HAVE_STDATOMIC
-- Performing Test HAVE_STDATOMIC - Success
-- Found WrapAtomic: TRUE
-- Found OpenGL: /usr/lib/x86_64-linux-gnu/libOpenGL.so
-- Found WrapOpenGL: TRUE
-- Found WrapVulkanHeaders: /usr/include
-- Qt version: 6.9.2
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libcrypto.so (found version "3.5.4")
-- Found Websocketpp: /usr/include
-- RapidJSON found. Headers: /usr/include
-- The following OPTIONAL packages have been found:

 * Git
 * Qt6WidgetsTools (required version >= 6.9.2)
 * OpenGL
 * WrapVulkanHeaders
[...]

PR #6638, opened 2025-12-09
refactor: move Twitch PubSub to use liveupdates

Nerixyz:
Twitch PubSub was the last user of websocketpp. It manually implemented most of the connection tracking and reconnecting. With the liveupdates, we have one system that can be used for all WebSocket services (BTTV, 7TV, Twitch PubSub).

With this PR, PubSub is migrated to it. Most patterns used in the old manager are representable. One thing that's missing is the ability to issue bulk subscriptions. I don't think this is necessary, and it would overcomplicate the design (we do the same in EventSub). 

Other than that, there's one modification needed to get nonces to work with the system (those are only needed for the debug counters).

Hopefully improves #6637 and #6576.


pajlada:
I'm OK with this approach, and OK that we don't do bulk subscribes given the amount of PubSub topics we subscribe to now.

src/providers/twitch/PubSubClient.cpp
```diff
@@ -1,224 +1,192 @@
...
-
-    return {message.topics, message.nonce};
+    this->sendText(R"({"type":"PING"})"_ba);
 }
 
-void PubSubClient::handleListenResponse(const PubSubMessage &message)
+QByteArray PubSubClient::encodeSubscription(const Subscription &subscription)
 {
+    PubSubListenMessage listen({subscription.topic});
+    this->nonces_[listen.nonce] = NonceInfo{.isListen = true};
```
pajlada: just formatting nit (will have to be autoformatted after)
```suggestion
    this->nonces_[listen.nonce] = NonceInfo{.isListen = true,};
```

src/providers/twitch/PubSubClient.cpp
```diff
@@ -1,224 +1,192 @@
...
+    return listen.toJson();
 }
 
-void PubSubClient::handleUnlistenResponse(const PubSubMessage &message)
+QByteArray PubSubClient::encodeUnsubscription(const Subscription &subscription)
 {
+    PubSubUnlistenMessage unlisten({subscription.topic});
+    this->nonces_[unlisten.nonce] = NonceInfo{.isListen = true};
```
pajlada: just formatting nit (will have to be autoformatted after)
```suggestion
    this->nonces_[unlisten.nonce] = NonceInfo{.isListen = true,};
```

tests/src/TwitchPubSubClient.cpp
```diff
@@ -107,38 +107,40 @@ TEST(TwitchPubSubClient, ServerRespondsToPings)
...
+    QTest::qWait(150);
 
-    ASSERT_EQ(pubSub.diag.connectionsOpened, 1);
-    ASSERT_EQ(pubSub.diag.connectionsClosed, 0);
-    ASSERT_EQ(pubSub.diag.connectionsFailed, 0);
-    ASSERT_EQ(pubSub.diag.messagesReceived, 2);
+    ASSERT_EQ(pubSub.wsDiag().connectionsOpened, 1);
+    ASSERT_EQ(pubSub.wsDiag().connectionsClosed, 0);
+    ASSERT_EQ(pubSub.wsDiag().connectionsFailed, 0);
+    ASSERT_EQ(pubSub.diag.messagesReceived, 1);  // LISTEN
```
pajlada, line 124: what changed here to make this 1 instead of 2?

tests/src/TwitchPubSubClient.cpp
```diff
@@ -177,33 +182,31 @@ TEST(TwitchPubSubClient, DisconnectedAfter1s)
...
+    QTest::qWait(500);
 
-    ASSERT_EQ(pubSub.diag.connectionsOpened, 1);
-    ASSERT_EQ(pubSub.diag.connectionsClosed, 0);
-    ASSERT_EQ(pubSub.diag.connectionsFailed, 0);
-    ASSERT_EQ(pubSub.diag.messagesReceived, 2);  // Listen RESPONSE & Pong
+    ASSERT_EQ(pubSub.wsDiag().connectionsOpened, 1);
+    ASSERT_EQ(pubSub.wsDiag().connectionsClosed, 0);
+    ASSERT_EQ(pubSub.wsDiag().connectionsFailed, 0);
+    ASSERT_EQ(pubSub.diag.messagesReceived, 1);  // Listen RESPONSE
```
pajlada, line 200: Is this because old pubsub code would send a ping asap, and new code waits to send a ping?

pajlada (changes requested):
You can remove the `ChatterinoWebSocketppLogger.hpp` file, the `websocketpp` submodule, and the `FindWebsocketpp.cmake` file

Overall looks good, will take a little bit of a closer look next review

tests/src/TwitchPubSubClient.cpp
Nerixyz: Previously, pings were sent out immediately. Now, we wait for the heartbeat/ping timer to expire once.

tests/src/TwitchPubSubClient.cpp
Nerixyz: Yes.

PR #6641, opened 2025-12-11
feat: include gifter in user info popup

### LINKED ISSUE #6640: Provide more detailed subscription info for subscribers in your own channel
We can use https://dev.twitch.tv/docs/api/reference#get-broadcaster-subscriptions when possible

From Discord where the request was specifically to see if the users's current subscription is a gift https://discord.com/channels/253897423439659008/459650160658481152/1448324427279499405

iProdigy:
Closes #6640

image

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -402,6 +402,7 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, Split *split)
                 .assign(&this->ui_.createdDateLabel);
             vbox.emplace<Label>("").assign(&this->ui_.followageLabel);
             vbox.emplace<Label>("").assign(&this->ui_.subageLabel);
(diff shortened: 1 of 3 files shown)


jupjohn:
One thing I noticed while testing this is that the usercard size may change once the gifter name is appended. If the user is about to click on a UI action, they may end up unintentionally clicking on something else - i.e. `unban` instead `1s timeout`, or `block` instead of `ignore highlights`. Even with the small name in my test, UI elements moved enough for me to potentially click the wrong thing. I imagine this would be worse with longer names.

Would putting the gifter info on a new line below prevent this, assuming that space is "reserved" no matter if the user was gifted or not?

src/widgets/dialogs/UserInfoPopup.cpp
Nerixyz: Could these be combined? I suppose in the "previously subscribed for X" we can't get the gifter?

PR #6649, opened 2025-12-13
chore: include new staff tenure badges

jupjohn:
This PR aims to update the known global badges file with the new 5, 10, and 15-year staff badges.


jupjohn:
I'll take the opportunity to move this to a workflow, given there's some test data available.

PR #6650, opened 2025-12-13
feat(plugins): read/update methods for messages in channels

Nerixyz:
This PR adds methods to read and update messages in a channel. Specifically, it adds

- `message_snapshot`
- `last_message`
- `replace_message`/`replace_message_at`
- `clear_messages`
- `find_message_by_id`
- `has_messages`
- `count_messages`

Since messages in Lua use `Message` (instead of `const Message`), we need to convert `std::shared_ptr<const Message>` to `std::shared_ptr<Message>`. Because interaction with `Messages` from Lua checks whether they are frozen, this is fine here (every message you get out of a channel is frozen ~> immutable).

### DIFF
--- a/.github/workflows/changelog-category-check.yml
+++ b/.github/workflows/changelog-category-check.yml
@@ -29,5 +29,5 @@ jobs:
             }
             return "";
 
-      - uses: pajlads/changelog-checker@v1.0.2
+      - uses: pajlads/changelog-checker@v1.0.4
         if: steps.label-checker.outputs.result != 'skip'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@
 - Dev: Cleaned up `AccountController` connections. (#6735)
 - Dev: Send multipart body for HTTP `PUT` requests. (#6746)
 - Dev: Removed unused variables. (#6748)
+- Dev: Updated `miniaudio` to 0.11.24. (#6754)
 
 ## 2.5.4
 


docs/lua-meta/globals.lua
```diff
@@ -151,6 +151,75 @@ function c2.Channel:add_system_message(message) end
...
+
+--- Check if the channel has any messages.
+---
+---@return boolean
+function c2.Channel:has_messages() end
+
+--- Count the number of messages in this channel.
+---
+---@return number
+function c2.Channel:count_messages() end
```
pajlada, line 203: What use case are you envisioning for `count_messages`?
Unless this is essential for this PR, I think it'd be best to discuss/implement separately

docs/lua-meta/globals.lua
```diff
@@ -151,6 +151,75 @@ function c2.Channel:add_system_message(message) end
...
+---@param n_items number Number of messages to retrieve. This is an upper bound, the actual number of messages returned might be lower.
+---@return c2.Message[]
+function c2.Channel:message_snapshot(n_items) end
+
+--- Get a list of messages in this channel (starting from the most recent messages).
+--- The snapshot is returned as a Lua table.
+---
+---@param n_items number Number of messages to retrieve. This is an upper bound, the actual number of messages returned might be lower.
+---@return c2.Message[]
+function c2.Channel:message_snapshot_as_table(n_items) end
```
pajlada: What benefit do we get from differentiating these?

docs/lua-meta/globals.lua
```diff
@@ -151,6 +151,75 @@ function c2.Channel:add_system_message(message) end
...
+---
+---@param index number A zero-based index (from the start) of the message to replace.
+---@param replacement c2.Message The replacement.
+function c2.Channel:replace_message_at(index, replacement) end
+
+--- Disable a message with an ID (set the `Disabled` flag).
+--- Note that this requires a re-layout to be visible. This functionality is not yet exposed to plugins.
+---
+---@param id string The ID of the message to disable.
+function c2.Channel:disable_message_by_id(id) end
```
pajlada: I'd prefer if we don't provide otherwise-achievable message-editing functions from Channel.
The plugin author should feasibly be able to call `find_message_by_id` and update the message's flag.

Same goes for `disable_all_messages`

pajlada (changes requested):

docs/lua-meta/globals.lua
Nerixyz: Some of the methods for `Channel` require an index. Something like `count_messages` is needed to determine how many messages are in the channel and to subsequently calculate the index. The indices are all from the top. So if you wanted to replace the second-to-last message, you'd need to pass `count_messages() - 2` as the index.

docs/lua-meta/globals.lua
Nerixyz: Nothing really (removed it). Could also be implemented in Lua as a helper.

docs/lua-meta/globals.lua
pajlada: Makes sense, thanks for the clarification

tests/lua/channel/messages.lua
```diff
@@ -0,0 +1,164 @@
+local chan = c2.Channel.by_name("mm2pl")
```
pajlada, line 5: Hijacking this PR to discuss licensing of things like tests.

My feeling is that they should be CC0-1.0, but I'm also open to them just following our code MIT license. Thoughts?

For the FileCopyrightText, it should be the creator of the file, but you're free to use whatever alias or real name or email (or no email) that you prefer

```suggestion
-- SPDX-FileCopyrightText: 2025 Nerixyz <nerixdev@outlook.de>
--
-- SPDX-License-Identifier: CC0-1.0

local chan = c2.Channel.by_name("mm2pl")
```

Nerixyz:
Ping - could this get in 2.6? Having access to the last messages in the channel might be nice, even if it's just for commands.

src/controllers/plugins/api/ChannelRef.hpp
```diff
@@ -85,6 +85,86 @@ struct ChannelRef {
...
+     * @exposed c2.Channel:replace_message
+     */
+    void replace_message(const MessagePtrMut &message,
+                         const MessagePtrMut &replacement);
+    /**
+     * Replace a specific message with a different one.
+     * 
+     * @lua@param message c2.Message The message to replace.
+     * @lua@param replacement c2.Message The replacement.
+     * @lua@param hint number A zero-based index (from the start) where the message is probably located. This is checked first. Otherwise the behavior is identical to the overload without this parameter.
```
Mm2PL: We should be consistent and use 1-based indexes.

src/controllers/plugins/api/ChannelRef.cpp
```diff
@@ -119,6 +120,70 @@ void ChannelRef::add_message(std::shared_ptr<Message> &message,
...
+
+void ChannelRef::clear_messages()
+{
+    this->strong()->clearMessages();
+}
+
+MessagePtr ChannelRef::find_message_by_id(const QString &id)
+{
+    return this->strong()->findMessageByID(id);
+}
```
Mm2PL:
```suggestion
MessagePtrMut ChannelRef::find_message_by_id(const QString &id)
{
    return std::const_pointer_cast<Message>(
        this->strong()->findMessageByID(id));
}
```


Needs to me changed to a MessagePtrMut to fix the below code sample
```lua
c = c2.Channel.by_name('pajlada')
lm = c:last_message()
borked = c:find_message_by_id(lm.id) -- can be any message as long as it comes from this function apparently
foo = c2.Message.new({
    flags = c2.MessageFlag.System|c2.MessageFlag.DoNotTriggerNotification,
    elements = {
        { type = "timestamp", color = "system" },
        { type = "text", text = "ayylmao test", color = "system" }
    },
})
c:replace_message(borked, foo) -- sol: no matching function call takes this number of arguments and the specified types

-- this should work
c:replace_message(lm, foo)
```

Mm2PL (approved):

PR #6652, opened 2025-12-13
feat(plugins): add `debug.traceback`

Nerixyz:
I love myself a nice stack trace, so let's add Lua's [`debug.traceback`](https://www.lua.org/manual/5.4/manual.html#pdf-debug.traceback). This can be used in `xpcall` as the message handler. It appends a stack trace to the message. This is really useful when testing (answering "which assert actually failed?").

The other functions of the `debug` library are all unsafe for us. Instead of reimplementing the function, we could also go with the approach we took for the `io` library. Since the `traceback` function is mostly a wrapper around `luaL_traceback`, which does the heavy lifting, I think it's fine here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Minor: Added broadcaster-only `/prediction`, `/cancelprediction`, `/lockprediction`, and `/completeprediction` commands. (#6583, #6612, #6632)
 - Minor: Added support for BetterTTV Pro subscriber badges. (#6625)
 - Minor: Added `debug.traceback` for plugins. (#6652)
+- Minor: Added title and duration options for `/clip` command. (#6669)
 - Minor: Added Markdown support to user notes. (#6490)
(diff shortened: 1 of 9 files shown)


src/controllers/plugins/PluginController.cpp
```diff
@@ -207,6 +208,13 @@ void PluginController::openLibrariesFor(Plugin *plugin)
         r["_IO_input"] = sol::nil;
         r["_IO_output"] = sol::nil;
     }
+    // set up debug lib
+    {
+        auto c2debug = lua.create_table();
+        g["debug"] = c2debug;
+
+        c2debug.set_function("traceback", lua::api::debugTraceback);
```
Mm2PL:
```suggestion
        auto debuglib = lua.create_table();
        g["debug"] = debuglib;

        debuglib.set_function("traceback", lua::api::debugTraceback);
```

PR #6653, opened 2025-12-14
feat: show mod/VIP badges from shared chats

jupjohn:
This PR aims to show mod & VIP badges from users in shared chats. It also adds the channel name to the mod/VIP badge's tooltip to further distinguish the source channel.

This can be tested by reverting `afbbadb8` and using the `/debug-test shared-chat-badge-messages`. You may have to run this command twice; the first time seems to not pick up the shared channel's profile picture or name.

TODO: update snapshot tests.

## Demo

https://github.com/user-attachments/assets/75c1e346-a307-4c09-aa64-45f09c64d3af

Closes #6651.

### DIFF
--- a/mocks/include/mocks/TwitchUsers.hpp
+++ b/mocks/include/mocks/TwitchUsers.hpp
@@ -18,7 +18,8 @@ public:
             .displayName = "forsen",
             .profilePictureUrl =
                 "https://static-cdn.jtvnw.net/jtv_user_pictures/"
-                "forsen-profile_image-48b43e1e4f54b5c8-300x300.png"};
+                "forsen-profile_image-48b43e1e4f54b5c8-300x300.png",
+        };
         return std::make_shared<TwitchUser>(u);
     }
 };


Nerixyz:
> `/debug-test shared-chat-badge-messages`

We should be using https://github.com/Chatterino/chatterino2/blob/658cbbdbdac7c4b15e6cf2e48fe0c5cca48f6c88/src/util/SampleData.cpp for this.

pajlada:
> > `/debug-test shared-chat-badge-messages`
> 
> We should be using [`658cbbd`/src/util/SampleData.cpp](https://github.com/Chatterino/chatterino2/blob/658cbbdbdac7c4b15e6cf2e48fe0c5cca48f6c88/src/util/SampleData.cpp) for this.

yeah, and reroute them to `#pajlada` too please

jupjohn:
> We should be using https://github.com/Chatterino/chatterino2/blob/658cbbdbdac7c4b15e6cf2e48fe0c5cca48f6c88/src/util/SampleData.cpp for this.

I didn't think it was worth adding these to sample data since it's just a test for this PR. If you guys want it added I'm happy to do that, but after I add snapshot tests I don't see a reason to have the messages in the program itself.

jupjohn:
> yeah, and reroute them to `#pajlada` too please

I've amended my command commit to target that channel :+1:

Issue #6627, opened 2025-12-06
settings.json - Settings are keeo reseting all the time!


w3bprinz:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Nearly every day my Chatterino Settings are gone, the whole settings.json is reverded back to default and I have to setup again or to use my settings.json.bkp Files to restore the settings.json...

Please fix this, this bug is ultra annoying...

I dont even understand why my settings.json is keeping reseting even when I normaly close Chatterino before restarting oder shut down my computer...

Chatterino should only use the settings.json on startup to read the settings and after that only when i change some settings to save the file, but i don't know why chatterino is doing that my settings.json is keeping reseted every f**** time.

### OS and Chatterino Version

Chatterino 7.5.4 (commit ddc4d335) built with Qt 6.9.1, MSVC 194435215, Crashpad Running on Windows 11 Version 24H2, kernel: 10.0.26100

Nerixyz:
Duplicate of #6358, but can you check if Chatterino crashes when closing the app. The crashdumps should be in `%APPDATA%/Chatterino2/Crashes/reports`. It would also be helpful to know how the `Settings.json` looks like before you start the app, and it resets (is there even a file?). **Don't copy-paste the exact contents here, because they contain your auth token**.

w3bprinz:
No reports under` %APPDATA%/Chatterino2/Crashes/reports`. the Settings.json looks good with all the Stuff in it, Token, Mod Comands, etc. and all settings I did, but when this happens that stuff is missing no account, no mod commands, no settings like highlights etc. just a "default" settings.json

w3bprinz:
Today again the settings.json got resetted.

And everytime I see "17:27 Failed to load channel badges - OAuth token is missing" this happens and this is nearly every day....

[settings.json](https://github.com/user-attachments/files/24037305/settings.json)

pajlada:
Could you try starting Chatterino, configuring something, closing Chatterino, then starting Chatterino again and seeing if the setting you configured stayed?

pajlada:
This should be much improved with the release of 2.5.5. Please open a new issue if you're noticing this issue again.

PR #6662, opened 2025-12-16
feat: Show restore dialog when settings fail to load

Nerixyz:
Since we get issues like https://github.com/Chatterino/chatterino2/issues/6627 or https://github.com/Chatterino/chatterino2/issues/6358, and I saw people complain about settings resetting on Twitch as well, I added a check for backups when we load settings.

This similar to https://github.com/Chatterino/chatterino2/pull/6363. However, it only checks when loading the main settings. The functions should be generic enough to adapt it to the window layout and other things.

This is how the dialog looks:
Screenshot

### DIFF
--- a/src/widgets/dialogs/RestoreBackupsDialog.cpp
+++ b/src/widgets/dialogs/RestoreBackupsDialog.cpp
@@ -13,6 +13,7 @@
 #include <QComboBox>
 #include <QDesktopServices>
 #include <QDialogButtonBox>
+#include <QKeyEvent>
 #include <QLabel>
 #include <QLocale>
 #include <QMessageBox>


src/util/FormatTime.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "common/FlagsEnum.hpp"
+
```
pajlada: unused include

src/util/Backup.cpp
```diff
@@ -0,0 +1,124 @@
...
+            .lastModified = entry.lastModified(),
+            .fileSize = entry.size(),
+            .state = state,
+        });
+    }
+
+    return backups;
+}
+
+void loadSettingFileWithBackups(const FileData &fileData,
```
pajlada: nit naming: `loadSettingsFileWithBackups`, or if it's meant to be generalized, `loadWithBackups` or `loadFileWithBackups`

src/util/Backup.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+class QJsonValue;
+
+namespace chatterino {
+class Paths;
+}  // namespace chatterino
+
+namespace chatterino::backup {
+
+bool anyBackupsOf(const QString &directory, const QString &filename);
```
pajlada: This can be moved to anon namespace in source file

src/util/Backup.hpp
```diff
@@ -0,0 +1,56 @@
...
+    std::filesystem::path dstPath;
+    QDateTime lastModified;
+    qint64 fileSize;
+    BackupState state;
+};
+
+std::vector<BackupFile> findBackupsFor(const QString &directory,
+                                       const QString &filename);
+
+struct FileData {
```
pajlada, line 42: Move struct to top of file with other structs

src/util/Backup.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <QStringBuilder>
+
+#include <algorithm>
+
+namespace {
+
+QRegularExpression regexForFile(const QString &file)
+{
+    return QRegularExpression(QStringView(u"^%1\\.(?:restore-)?bkp-\\d+$")
+                                  .arg(QRegularExpression::escape(file)));
```
pajlada: Where does the `restore-` part come from?

src/util/Backup.cpp
```diff
@@ -0,0 +1,124 @@
...
+            QJsonParseError err;
+            QJsonDocument::fromJson(ba, &err);
+            if (err.error == QJsonParseError::NoError)
+            {
+                state = BackupState::Ok;
+            }
+            else
+            {
+                state = BackupState::BadContents;
+            }
```
pajlada: If the intent is to load the contents as JSON to ensure it can be read as JSON we should use rapidjson instead of qt

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        bool retry = true;
+        while (retry)
+        {
+            retry = false;
+            std::error_code ec;
+            qCDebug(chatterinoSettings)
+                << "Copying" << stdPathToQString(data->path) << "to"
+                << stdPathToQString(data->dstPath);
```
pajlada, line 114: Rather than going through the copying step, could we allow the `load` function passed to `loadSettingFileWithBackups` to accept the path it should load from instead?
So loading would happen straight from e.g. `settings-bkp-01.json`

The Settings library supports this via `setPath` and `loadFrom` instead of just calling `load`

src/util/Backup.cpp
Nerixyz: Restoring backups used to back up the overwritten file with `...restore-bkp-{n}`.

src/widgets/dialogs/RestoreBackupsDialog.cpp
Nerixyz: Technically yes. However, we need to make sure that we immediately save the settings again (essentially creating a copy), because we currently only save if settings are changed. Otherwise, the user will see the same dialog the next time again.

pajlada: Good thinking, that approach sounds good to me.

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,159 @@
...
+        auto *data = get_if<backup::BackupFile>(&selected);
+        if (!data)
+        {
+            return;
+        }
+        auto url = QUrl::fromLocalFile(stdPathToQString(data->path));
+        QDesktopServices::openUrl(url);
+    });
+
+    this->refreshBackups();
```
pajlada, line 157: It would be good to set a fixed size for this dialog - that way it will properly show up as a popup on sway

this worked for me and replicates what QMessageBox uses:
```
    this->layout()->activate();
    this->setFixedSize(this->layout()->totalMinimumSize());
```

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,159 @@
...
+    hbox->addWidget(&this->showButton);
+    layout->addLayout(hbox);
+
+    this->corruptedBackupsWarning.hide();
+    layout->addWidget(&this->corruptedBackupsWarning);
+
+    layout->addSpacing(10);
+
+    auto *buttons =
+        new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel);
```
pajlada: I think we should use custom buttons, at a quick thought I'm thinking something like:
 - **GREEN** "Restore backup" or "Load backup"
 - **RED** "Discard backup" with an extra QMessageBox "This will not be shown again, your settings are being deleted, is this ok?" with the default button being set to cancel

but happy to iterate on that

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,159 @@
...
+    : QDialog(parent, Qt::Dialog)
+    , fileData(std::move(fileData))
+    , backupCombo(new QComboBox)
+    , showButton(u"Show"_s)
+    , corruptedBackupsWarning(
+          u"Some backups are damaged or otherwise unreadable."_s)
+{
+    this->setAttribute(Qt::WA_DeleteOnClose);
+    this->setWindowTitle(u"Restore backup of " %
+                         this->fileData.fileKind.toLower() % '?');
```
pajlada:
```suggestion
    this->setWindowTitle(u"Chatterino - Restore backup of " %
                         this->fileData.fileKind.toLower() % '?');
```

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    auto *layout = new QVBoxLayout(this);
+
+    auto *description =
+        new QLabel(this->fileData.fileKind % u" failed to load: " % prevError %
+                   u"<p>" % this->fileData.fileDescription %
+                   u"<p>There are backups of this file.<br>Do you want to "
+                   u"restore the selected backup?");
```
pajlada, line 77:
```suggestion
    auto *description =
        new QLabel("Chatterino " % this->fileData.fileKind % u" failed to load: " % prevError %
                   u"<p>" % this->fileData.fileDescription %
                   u"<p>There are backups of this file.<br>Do you want to "
                   u"restore the selected backup?");
```

src/widgets/dialogs/RestoreBackupsDialog.cpp
pajlada: Fixed in [`6018adf` (this PR)](https://github.com/Chatterino/chatterino2/pull/6662/commits/6018adfa4210957bbefecb34b62b44a5396859d2)

pajlada:
@Mm2PL Do you want a chance to test/review this before this gets merged in?

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,182 @@
...
+    layout->addWidget(&this->corruptedBackupsWarning);
+
+    layout->addSpacing(10);
+
+    auto *buttons = new QDialogButtonBox;
+    layout->addWidget(buttons);
+    buttons->addButton(QDialogButtonBox::Ok)->setText(u"Restore Backup"_s);
+    buttons->addButton(QDialogButtonBox::Cancel)->setText(u"Discard"_s);
+    // TODO: Add option to exit Chatterino without doing anything.
+    // This should be the default option to be used if the user exits the dialog through other means (e.g. ALT+F4)
```
Mm2PL: Is there reason not to do this in this PR?

Nerixyz:
Looks good to me now.

src/widgets/dialogs/RestoreBackupsDialog.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <QVBoxLayout>
+
+#include <chrono>
+#include <filesystem>
+
+using namespace Qt::Literals;
+
+namespace chatterino {
+
+RestoreBackupsDialog::RestoreBackupsDialog(backup::FileData fileData,
```
Mm2PL, line 45: I very much dislike the behavior of this dialog that when it's closed, it forgets the opportunity to restore settings.

src/widgets/dialogs/RestoreBackupsDialog.cpp
Nerixyz: I don't think so, added it.

src/widgets/dialogs/RestoreBackupsDialog.cpp
Nerixyz: No longer the case.

pajlada (approved):
CC @Mm2PL

src/widgets/dialogs/RestoreBackupsDialog.cpp
Mm2PL: Escape still causes this, we need something to eat the event.
```diff
diff --git a/src/widgets/dialogs/RestoreBackupsDialog.cpp b/src/widgets/dialogs/RestoreBackupsDialog.cpp
index 9c2b20073..bf8ea4c55 100644
--- a/src/widgets/dialogs/RestoreBackupsDialog.cpp
+++ b/src/widgets/dialogs/RestoreBackupsDialog.cpp
@@ -13,12 +13,14 @@
 #include <QComboBox>
 #include <QDesktopServices>
 #include <QDialogButtonBox>
+#include <QEvent>
 #include <QLabel>
 #include <QLocale>
 #include <QMessageBox>
 #include <QPushButton>
 #include <QString>
 #include <QStringBuilder>
+#include <Qt>
 #include <QVBoxLayout>
 
 #include <chrono>
@@ -171,6 +173,11 @@ void RestoreBackupsDialog::closeEvent(QCloseEvent * /*event*/)
     }
 }
 
+void RestoreBackupsDialog::keyPressEvent(QKeyEvent *event)
+{
+    event->setAccepted(event->key() == Qt::Key_Escape);
+}
+
 void RestoreBackupsDialog::refreshBackups()
 {
     this->backupCombo.clear();
diff --git a/src/widgets/dialogs/RestoreBackupsDialog.hpp b/src/widgets/dialogs/RestoreBackupsDialog.hpp
index 9f145c9cb..946c0467f 100644
--- a/src/widgets/dialogs/RestoreBackupsDialog.hpp
+++ b/src/widgets/dialogs/RestoreBackupsDialog.hpp
@@ -10,6 +10,7 @@
 #include <QComboBox>
 #include <QDialog>
 #include <QDialogButtonBox>
+#include <QEvent>
 #include <QLabel>
 #include <QPushButton>
 
@@ -25,6 +26,7 @@ public:
 
 protected:
     void closeEvent(QCloseEvent * /*event*/) override;
+    void keyPressEvent(QKeyEvent *event) override;
 
 private:
     void refreshBackups();
```

Nerixyz: Done.

PR #6665, opened 2025-12-19
feat: class lead mod badge as channel authority

jupjohn:
This PR aims to correctly flag the lead moderator badge as a channel authority badge instead of vanity (default.) This fixes the `Channel` badge filtering setting not filtering this badge.


jupjohn:
Worth squashing the two lead mod changelog entries into a single "Lead moderator support" line?

PR #6667, opened 2025-12-20
Add Lead Moderator support for managing moderators and VIPs

Leafyzito:
This PR adds support for Lead Moderators to manage moderators and VIPs in the same way as broadcasters.

Note: The client-side implementation is ready, but Twitch's API does not yet allow (as far as I know) Lead Moderators to add/remove moderators and VIPs. The UI will show the appropriate buttons and functionality, but API calls may fail until Twitch enables this on their backend. This prepares the client-side code for when Twitch enables the feature.

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -396,6 +396,7 @@ void IrcMessageHandler::parsePrivMessageInto(
             auto parsedBadges = parseBadges(badgesTag.toString());
             channel->setMod(parsedBadges.contains("moderator") ||
                             parsedBadges.contains("lead_moderator"));
+            channel->setLeadModerator(parsedBadges.contains("lead_moderator"));
             channel->setVIP(parsedBadges.contains("vip"));
             channel->setStaff(parsedBadges.contains("staff"));
         }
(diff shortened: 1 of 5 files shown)


iProdigy:
I expect the Helix API to retain the same broadcaster-only restriction, requiring lead moderators to use the first-party site (or undocumented API) to manage other mods

Leafyzito:
Does chatterino make use of undocumented API? It's something I never explored much, but if so, I could look into it.

pajlada:
> Does chatterino make use of undocumented API? It's something I never explored much, but if so, I could look into it.

We try to avoid this where possible

Leafyzito:
> We try to avoid this where possible

Should we wait a bit more and see if it becomes doable with Helix, or do we go the undocumented API way?

pajlada:
> > We try to avoid this where possible
> 
> Should we wait a bit more and see if it becomes doable with Helix, or do we go the undocumented API way?

I'd prefer to wait

Leafyzito:
I created a uservoice asking for this to be implemented on Helix.
Would appreciate if you guys could up vote it so maybe it could get some attention :)
https://twitch.uservoice.com/forums/310213-developers/suggestions/50848118-allow-lead-moderators-to-manage-moderator-and-vip

PR #6669, opened 2025-12-21
feat: add duration and title options to clip command

iProdigy:
Enabled by 2025-12-19 change: https://dev.twitch.tv/docs/change-log/

Sample usage:
* `/clip hello world`
* `/clip --duration 10 hello world`
* `/clip --duration 60`


pajlada (changes requested):
~~`/clip --title "foo bar"` creates a clip with the title `"foo bar"`, i think you need to parse this the same way as you did for polls/predictions~~
I didn't read the PR description, 1 moment :-)

pajlada (approved):

PR #6676, opened 2025-12-24
fix(macOS): command + delete

jacob-thompson:
#6111 fixed cmd+delete behavior for ResizingTextEdits, but the EditUserNotesDialog still has no cmd+delete behavior as it just uses QTextEdit

this reimplements the cmd+delete behavior as an event filter that can be installed onto any TextEdit that needs to handle it

### DIFF
--- a/src/widgets/dialogs/EditUserNotesDialog.cpp
+++ b/src/widgets/dialogs/EditUserNotesDialog.cpp
@@ -2,8 +2,8 @@
 
 #include "singletons/Theme.hpp"
 #include "util/LayoutCreator.hpp"
-#include "widgets/helper/CmdDeleteKeyFilter.hpp"
 #include "widgets/buttons/SvgButton.hpp"
+#include "widgets/helper/CmdDeleteKeyFilter.hpp"
 #include "widgets/MarkdownLabel.hpp"
 
 #include <QCheckBox>


src/widgets/helper/CmdDeleteKeyFilter.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+bool CmdDeleteKeyFilter::eventFilter(QObject *obj, QEvent *event)
+{
+    if (event->type() == QEvent::KeyPress)
+    {
+        auto *textEdit = qobject_cast<QTextEdit *>(obj);
+        const auto *keyEvent = dynamic_cast<QKeyEvent *>(event);
+        if (textEdit && keyEvent)  // types are verified
+        {
+#ifdef Q_OS_MACOS
```
Nerixyz: Extend this to the whole `if (event->type() == QEvent::KeyPress)` - we don't need to handle this on other platforms.

src/widgets/helper/CmdDeleteKeyFilter.cpp
```diff
@@ -0,0 +1,38 @@
...
+                                        QTextCursor::KeepAnchor);
+                }
+                cursor.removeSelectedText();
+                textEdit->setTextCursor(cursor);
+                return true;
+            }
+#endif
+        }
+    }
+    return QObject::eventFilter(obj, event);
```
Nerixyz:
```suggestion
    return false;
```

That's what the `QObject` implementation will do anyway.

src/widgets/helper/CmdDeleteKeyFilter.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <QObject>
+
+class CmdDeleteKeyFilter : public QObject
```
Nerixyz, line 7: Explain in a comment what this does, as well as why and where it's needed.

src/widgets/helper/CmdDeleteKeyFilter.cpp
```diff
@@ -0,0 +1,38 @@
+#include "CmdDeleteKeyFilter.hpp"
```
Nerixyz:
```suggestion
#include "widgets/helper/CmdDeleteKeyFilter.hpp"
```

pajlada (approved):

PR #6681, opened 2025-12-27
Enable highlighting all Twitch global badges

achow101:
Use the Twitch global badges that the application already knows to build the list of badges that can be highlighted. This allows new badges to be added to the highlights whenever Twitch adds a new badge without needing a software update.

This is somewhat similar to what FFZ allows for badge highlighting.

For simplicity, the badges displayed are only the first version of each badge, with the exception of 2 badges which don't use number versions. All versions of those badges will be listed.

### DIFF
--- a/src/providers/twitch/TwitchBadges.cpp
+++ b/src/providers/twitch/TwitchBadges.cpp
@@ -19,6 +19,8 @@
 #include <QThread>
 #include <QUrlQuery>
 
+#include <algorithm>
+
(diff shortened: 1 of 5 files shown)


pajlada:
Thanks for the PR - there's some highlight reworks going on right now that would collide with this so I'll hold off on this for now. I'll keep your PR open and see if I can incorporate (with attribution) the ongoing changes.

PR #6685, opened 2025-12-28
feat: Automatically refresh chatter list

drogeek:
Hi there!
I like to see who is connecting to my stream, but I noticed the chatter list was never refreshed. This MR refresh the list every 2 seconds.

### DIFF
--- a/src/widgets/ChatterListWidget.cpp
+++ b/src/widgets/ChatterListWidget.cpp
@@ -143,209 +143,21 @@ QString formatChattersError(HelixGetChattersError error, const QString &message)
 
 }  // namespace
 
-ChatterListWidget::ChatterListWidget(const TwitchChannel *twitchChannel,
-                                     QWidget *parent)
-    : BaseWindow({}, parent)
+void ChatterListWidget::setupUi()
 {
-    this->setWindowTitle("Chatter List - " + twitchChannel->getName());
-    assert(twitchChannel != nullptr);
+    this->setWindowTitle("Chatter List - " + this->twitchChannel_->getName());
+    assert(this->twitchChannel_ != nullptr);
 
     this->setAttribute(Qt::WA_DeleteOnClose);
 
-    auto *dockVbox = new QVBoxLayout();
-    auto *searchBar = new QLineEdit(this);
+    this->dockVbox_ = new QVBoxLayout(this);
+    this->searchBar_ = new QLineEdit(this);
 
-    auto *chattersList = new QListWidget();
-    auto *resultList = new QListWidget();
+    this->chattersList_ = new QListWidget(this);
+    this->resultList_ = new QListWidget(this);
 
-    auto *loadingLabel = new QLabel("Loading...");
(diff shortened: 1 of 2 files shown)


src/widgets/ChatterListWidget.cpp
```diff
@@ -143,41 +143,95 @@ QString formatChattersError(HelixGetChattersError error, const QString &message)
...
+    this->setLayout(this->dockVbox_);
 
-    auto *loadingLabel = new QLabel("Loading...");
-    searchBar->setPlaceholderText("Search User...");
+    this->setMinimumWidth(300);
+}
+
+void ChatterListWidget::clearUi()
+{
+    std::cout << "clearing the connected users UI" << std::endl;
```
Nerixyz, line 212: Don't use iostreams, use Qt's [`qCDebug`](https://doc.qt.io/qt-6/qloggingcategory.html#qCDebug) and friends.

src/widgets/ChatterListWidget.cpp
```diff
@@ -316,81 +369,45 @@ ChatterListWidget::ChatterListWidget(const TwitchChannel *twitchChannel,
...
-
-    dockVbox->addWidget(searchBar);
-    dockVbox->addWidget(loadingLabel);
-    dockVbox->addWidget(chattersList);
-    dockVbox->addWidget(resultList);
-    resultList->hide();
-
-    this->setStyleSheet(this->theme->splits.input.styleSheet);
-    this->setLayout(dockVbox);
+ChatterListWidget::ChatterListWidget(const TwitchChannel *twitchChannel,
```
Nerixyz, line 401: Move this constructor back to the top.

src/widgets/ChatterListWidget.hpp
```diff
@@ -17,6 +17,22 @@ class ChatterListWidget : public BaseWindow
     ChatterListWidget(const TwitchChannel *twitchChannel, QWidget *parent);
 
     Q_SIGNAL void userClicked(QString userLogin);
+
+private:
+    const TwitchChannel *twitchChannel_ = nullptr;
```
Nerixyz, line 22: It's probably better to store a `std::shared_ptr<TwitchChannel>` to avoid dangling references.

Nerixyz (commented):
Depending on the size of the chat, 2s might be too low as the interval. Not sure what the best interval is. Maybe it should be dynamic based on how long we take to load the list?

src/widgets/ChatterListWidget.cpp
drogeek: I simply removed it, it was there for debugging purposes

src/widgets/ChatterListWidget.hpp
drogeek: well, the memory isn't handled by this Widget, it's just passed by another component, I do not need to free it, and actually doing so makes the whole program crash

drogeek: nope, that's not it, but the program do crash when I close this widget for some reason

Issue #5978, opened 2025-02-23
Lua API for manipulating the window layout


Mm2PL:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

This API should be able to give plugins the ability to:
 - open new tabs
 - rename tabs
 - move tabs around
 - close tabs
 - add/remove splits
 - ideally also move splits between containers
 - ideally also change the width and height constraints on splits

The easy option seems to be presenting a tree of objects like:
```
Window
+-- Tab1
|   +-- SplitContainer1 (horizontal)
|       +-- SplitContainer1 (vertical)
|       |   +-- Split1
|       |   +-- SplitContainer2 (horizontal)
|       |       +-- Split2
|       |       +-- Split3
|       +-- SplitContainer3
|           ...
+-- Tab2
    ...
```
Which would correspond to something like:

![Image](https://github.com/user-attachments/assets/70b1cd71-db3c-4562-b815-555186bc8e99)

I am still not convinced that this is the best way to achieve our goals because this will leak the internal structure of a window to plugins. Should we want to change that or clean it up somehow this API will be in the way.

### OS and Chatterino Version

Chatterino 2.5.2 DEBUG (commit 9afd60534 modified) built with Qt 6.8.2 Running on Arch Linux, kernel: 6.13.2-arch1-1

Nerixyz:
Basic implementation of the read only API can be found in https://github.com/Nerixyz/chatterino2/tree/feat/window-api.

PR #6687, opened 2025-12-29
feat(plugins): add readonly window layout access

Nerixyz:
This implements the second part for https://github.com/Chatterino/chatterino2/issues/5978 - read-only access to the window layout.

This allows plugins to query the open channels and how they're presented. The following is an example of how to use the API to dump the window layout to JSON:

<details><summary>init.lua</summary>

```lua
local JSON = require("chatterino.json")

local function enum_to_string(enum)
	local rev = {}
	for k, v in pairs(enum) do
		rev[v] = k
	end
	return function(v)
		local r = rev[v]
		if r ~= nil then
			return r
		end
		return "unknown-value(" .. v .. ")"
	end
end
local SPCNT_to_string = enum_to_string(c2.SplitContainerNodeType)
local WindowType_to_string = enum_to_string(c2.WindowType)

---@generic T
---@generic U
---@param it T[]
---@param fn fun(v: T): U
---@return U[]
local function map(it, fn)
	local res = {}
	for i, v in ipairs(it) do
		res[i] = fn(v)
	end
	return res
end

---@param s c2.Split|nil
local function do_split(s)
	if not s then
		return JSON.null
	end
	return { channel = s.channel:get_name() }
end

---@param p c2.SplitContainerNode|nil
local function do_node(p)
	if not p then
		return JSON.null
	end
	return {
		type = SPCNT_to_string(p.type),
		horizontal_flex = p.horizontal_flex,
		vertical_flex = p.vertical_flex,
		split = do_split(p.split),
		children = map(p:children(), do_node),
	}
end

---@param p c2.SplitContainer|nil
local function do_container(p)
	if not p then
		return JSON.null
	end
	return {
		base_node = do_node(p.base_node),
		selected_split = do_split(p.selected_split),
		splits = map(p:splits(), do_split),
	}
end

---@param win c2.Window
local function do_window(win)
	local splits = {}
	for i = 1, win.notebook.page_count do
		table.insert(splits, do_container(win.notebook:page_at(i - 1)))
	end
	return {
		type = WindowType_to_string(win.type),
		notebook = {
			page_count = win.notebook.page_count,
			selected_page = do_container(win.notebook.selected_page),
			splits = splits,
		},
	}
end

c2.register_command("/windows", function(ctx)
	ctx.channel:add_system_message(JSON.stringify(map(c2.windows:all(), do_window)))
end)
```
</details> 

There aren't good tests for this, because we don't emulate the window layout in tests.


pajlada:
I've taken a quick glance this looks fine.

Open question: Should reading your open channels be locked behind a permission?

pajlada:
> Open question: Should reading your open channels be locked behind a permission?

I don't think so - not for reading.

For writing (i.e. opening a new channel / window, or modifying your split layout) we might want a permission, but that's not for this issue.

src/controllers/plugins/api/ChannelRef.hpp
```diff
@@ -288,12 +288,12 @@ struct ChannelRef {
 
     static void createUserType(sol::table &c2);
 
-private:
-    std::weak_ptr<Channel> weak;
-
     /// Locks the weak pointer and throws if the pointer expired
     std::shared_ptr<Channel> strong();
 
+private:
```
pajlada, line 330: can undo this change

src/controllers/plugins/api/WindowManager.cpp
```diff
@@ -0,0 +1,175 @@
...
+        sol::state_view(state).create_table(static_cast<int>(items.size()), 0);
+
+    for (size_t idx = 0; idx < items.size(); idx++)
+    {
+        tbl[static_cast<int>(idx + 1)] = QPointer(items[idx]);
+    }
+    return tbl;
+}
+
+struct SplitContainerNodeWrap {
```
pajlada, line 36: Can you add a comment here or explain why we need this wrapper struct?

pajlada (changes requested):

pajlada (approved):

docs/wip-plugins.md
```diff
@@ -894,6 +894,69 @@ All arguments accept an [`Image`](#image) or a `string` (URL).
 
 Requires the [network permission](#permissions).
 
+#### `Split`
+
+A split. See [Anatomy of a Chatterino window](https://wiki.chatterino.com/Glossary/#anatomy-of-a-chatterino-window).
+This holds a `channel` ([Channel]()) that's open in this split.
```
Mm2PL: bad link

docs/wip-plugins.md
```diff
@@ -894,6 +894,69 @@ All arguments accept an [`Image`](#image) or a `string` (URL).
...
+
+#### `SplitContainerNode`
```
Mm2PL, line 902: This should have its own Anatomy of a Chatterino Window that explains what a SplitContainer is. This could be a later PR.

docs/wip-plugins.md
```diff
@@ -894,6 +894,69 @@ All arguments accept an [`Image`](#image) or a `string` (URL).
...
+
+Get the notebook page at a specific index.
+`i` is the zero based index of the page.
+Returns the page ([`SplitContainer`](#splitcontainer)) contained at the specified index.
+
+#### `Window`
+
+It has the following fields:
+
+- `notebook` ([`SplitNotebook`](#splitnotebook)) The notebook of this window.
```
Mm2PL, line 942: Again, there is no good explanation of what a SplitNotebook is.

src/controllers/plugins/api/WindowManager.cpp
```diff
@@ -0,0 +1,180 @@
...
+    // Wrapped in a QPointer
+    c2.new_usertype<SplitNotebook>(
+        "SplitNotebook", sol::no_constructor,  //
+        "selected_page", sol::readonly_property([](SplitNotebook &self) {
+            return QPointer(self.getSelectedPage());
+        }),
+        "page_count", sol::readonly_property(&SplitNotebook::getPageCount),
+        "page_at", [](const SplitNotebook &self, int index) {
+            return QPointer(
+                dynamic_cast<SplitContainer *>(self.getPageAt(index)));
```
Mm2PL, line 155: Need to check against getPageCount for out of bounds indices.

Mm2PL (changes requested):
smol changes

docs/wip-plugins.md
Nerixyz: I'd rather do this in a later PR.

Mm2PL (approved):

PR #6690, opened 2025-12-30
refactor: remove use of raw nonstd::

pajlada:

### DIFF
--- a/src/controllers/plugins/PluginController.cpp
+++ b/src/controllers/plugins/PluginController.cpp
@@ -450,7 +450,7 @@ std::pair<bool, QStringList> PluginController::updateCustomCompletions(
                 qCDebug(chatterinoLua)
                     << "Got error from plugin " << pl->meta.name
                     << " while refreshing tab completion: "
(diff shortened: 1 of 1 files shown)


src/controllers/commands/common/ChannelAction.cpp
```diff
@@ -58,14 +58,14 @@ void PrintTo(const PerformChannelAction &a, std::ostream *os)
         << ", duration:" << std::to_string(a.duration) << '}';
 }
 
-nonstd::expected<std::vector<PerformChannelAction>, QString> parseChannelAction(
+Expected<std::vector<PerformChannelAction>, QString> parseChannelAction(
```
Nerixyz, line 61:
```suggestion
ExpectedStr<std::vector<PerformChannelAction>> parseChannelAction(
```

I don't know if you want to do this in this PR.

pajlada: I'm fine to leave this - I personally like `Expected<T, E>` but I'm fine if other contributors make their own code use the common aliases we provide.

PR #6691, opened 2025-12-30
Include third party badges in Message.badges

achow101:
Including third party badges in Message.badges allows filters to match on those badges in addition to the Twitch global badges. Since the third party badges don't have ids like the Twitch badges, the badge key uses the tooltip name.


pajlada:
Closing in favour of #6709

PR #6692, opened 2025-12-30
chore: simplify `USE_ALTERNATE_LINKER` code

pajlada:
We don't support this old version of clang, we don't need to use this
fancy macro. Just standard provide `-fuse-ld`, and error if the
alternate linker can't be found


Nerixyz:
Since 3.29, you can set [`CMAKE_LINKER_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html) which works cross platform.

Issue #2157, opened 2020-11-04
Solution like "Play highlight sound even when Chatterino is focused" for whispers


zneix:
**What should be added?**
Not really sure whether this is considered as a bug or a feature suggestion, but I would either like to have the option from the title apply for whisper messages as well or introduce a new thing that will allow us to get pings from whispers while having chatterino focused.

**Why should it be added?**
I often leave chatterino in focus while doing other things away from keyboard and have the option from issue title turned on, so I can get back to the pc and see what's going on. I would also love to be notified when someone whispers me, so I don't miss their messages.

PR #6695, opened 2025-12-30
feat: rework highlights & their UI

### LINKED ISSUE #2124: The columns in the highlights  ettings window don't enlarge when I enlarge the window
**Describe the bug**

**Screenshots**
 

 Can't adjust the column size so the text isn't all there

**Chatterino version**

Chatterino 2.2.3-beta (commit 3c6e6446c)

**Operating system**

Windows 10

**Additional information**

### LINKED ISSUE #2365: Highlights for specific channels.
What should be added?
Highlights for specific channels.

Why should it be added?
I wan't a specific emote to be highlighted, but it get's spammed a lot in some other channels.

Example:
billede
billede

### LINKED ISSUE #2401: Muting specific highlight sounds for specific channels
So, right now you can either mute all the notifications for a channel or none.
The problem for me is: My name is "Max", and in basically every channel people use "max" to shorten "maximum", so I would get pinged every 10 seconds or so.

If you could open up a menu i.e. through the three dots on a channel, where you choose which highlights you want to hear from only that channel and which not, would be very nice.

### LINKED ISSUE #6784: "Move Up" on topmost highlight entry causes it to shift to bottom
### Checklist

### Describe your issue

When pressing `Move Up` on a highlight entry that's at the top of the highlight patterns list, the highlight is moved to the bottom of the list.

I would expect it to not move at all, given it's at the top.

### Screenshots

https://github.com/user-attachments/assets/2d7732c6-e010-4632-b98c-70dcfab2d0d7

### OS and Chatterino Version

Chatterino 2.5.4 (commit [b69ab9315](https://github.com/Chatterino/chatterino2/commit/b69ab9315)) built with Qt 6.10.1
Running on Arch Linux, kernel: 6.18.7-2-cachyos

pajlada:
This rebuilds highlights to a fully user-customizable solution.
Users can now:
 - freely disable and enable any highlight.
 - reprioritize any highlight.
 - create filter highlights which allows for giga-advanced highlights.

Things that would be nice for future expansions:
 - Exporting/Importing individual highlights
 - Exporting/Importing list of highlights

Fixes https://github.com/Chatterino/chatterino2/discussions/3472
Fixes https://github.com/Chatterino/chatterino2/discussions/3865
Fixes https://github.com/Chatterino/chatterino2/discussions/3872
Fixes https://github.com/Chatterino/chatterino2/discussions/4191
Fixes https://github.com/Chatterino/chatterino2/discussions/5285
Partially fixes https://github.com/Chatterino/chatterino2/discussions/3616 (highlights only, not ignores)
Fixes https://github.com/Chatterino/chatterino2/issues/2365 through the use of filter highlights. Not super UX-friendly, but it gets the job done.
Fixes https://github.com/Chatterino/chatterino2/issues/2401 through the use of filter highlights.
Fixes https://github.com/Chatterino/chatterino2/issues/2124 - the checkboxes are not there and configuration is done in a dialog instead.
Fixes https://github.com/Chatterino/chatterino2/discussions/3710 by allowing a filter highlight.

Things I think should be fixed in a follow-up PR:
https://github.com/Chatterino/chatterino2/discussions/5360
https://github.com/Chatterino/chatterino2/discussions/6524
https://github.com/Chatterino/chatterino2/issues/2157
https://github.com/Chatterino/chatterino2/discussions/3584 - some groundwork has been laid for it in this PR, but leaving it for a follow-up.
https://github.com/Chatterino/chatterino2/issues/6794

When reviewing or testing this PR, please take a look at the included `highlights-rework.md` document for any known shortcomings or breaking changes.

<details>
<summary>Images</summary>

image

image

image

image

</details>


resources/sounds/REUSE.toml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2026 Rasmus Karlsson <rasmus.karlsson@pajlada.com>
+# SPDX-License-Identifier: CC0-1.0
+
+version = 1
+
+[[annotations]]
+path = "ping2.wav"
+SPDX-FileCopyrightText = "Copyright (c) 2016 fourtf"
+SPDX-License-Identifier = "MIT"
```
pajlada, line 9: @fourtf Is this an appropriate license for your `ping2.wav` sound?

pajlada: This was resolved in https://github.com/Chatterino/chatterino2/pull/7206

highlights-rework.md
Wissididom, line 1: For later: I think this should be either moved to the docs folder or removed before merging, because I feel like the root directory might not be the best place for it.

pajlada: Yep, the intent is to remove it. It's just where I keep random notes about the rework since I don't want to keep modifying the PR description over and over.

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+                getApp()->getTwitchBadges()->getBadgeIcons(
+                    highlights::twitchBadges(),
+                    [w](const QString &identifier,
+                        const std::shared_ptr<QIcon> &icon) {
+                        if (!w)
+                        {
+                            return;
+                        }
+
+                        int index = w->findData(identifier);
```
3r01: The pointer to the combo box will not be null if it is destroyed so this can cause a use-after-free if the form is closed while the request for badge icons hasn't returned yet. You can solve by capturing a `QPointer` in the lambda instead

src/controllers/highlights/Model.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    row[Column::Name]->setData(getIcon(highlight), Qt::DecorationRole);
+
+    if (const auto *h = std::get_if<BadgeHighlight>(&highlight))
+    {
+        getApp()->getTwitchBadges()->getBadgeIcon(
+            h->getBadgeName(),
+            [row](const QString &name, const std::shared_ptr<QIcon> &icon) {
+                (void)name;  // unused
+                row[Column::Name]->setData(*icon, Qt::DecorationRole);
```
3r01: If the row is deleted before the request for badge icons returns then this will cause use-after-free; `row` will still exist but the objects its pointers refer to will have already been destroyed.

src/providers/twitch/eventsub/MessageBuilder.cpp
```diff
@@ -668,6 +668,10 @@ MessagePtr makeAutomodHoldMessageHeader(
         u"AutoMod: Held a message for reason: " % reason %
         u". Allow will post it in chat. Allow Deny");
 
+    QVariantMap tagsMap;
+    Communi::TagsRef tags(tagsMap);
+    builder.parseHighlights(tags, {}, {}, channel);
```
3r01: Is it intentional to not pass the message text here?

src/singletons/Settings.cpp
```diff
@@ -53,8 +57,231 @@ namespace {
...
+        {
+            outcome.setSound("");
+        }
+        else
+        {
+            const auto &customSoundURL = customSoundSetting.getValue();
+            if (customSoundURL.isEmpty())
+            {
+                // Default ping sound
+                outcome.setSound("001-ping2");
```
3r01, line 76: What happens if I had set a custom global default before, but not a custom sound per highlight?

src/controllers/highlights/types/Outcome.hpp
```diff
@@ -0,0 +1,112 @@
...
+    void serialize(rapidjson::Value &ret,
+                   rapidjson::Document::AllocatorType &a) const;
+
+    bool deserialize(const rapidjson::Value &value);
+
+    friend QDebug operator<<(QDebug dbg, const Outcome &v);
+
+private:
+    std::shared_ptr<QColor> resolvedBackgroundColor =
+        std::make_shared<QColor>();
```
3r01, line 117: Not copy safe, copies of `Outcome` are going to point at the same `QColor`

src/controllers/highlights/Model.cpp
3r01, line 98: If the row is deleted before the request for badge icons returns then this will cause use-after-free; `row` will still exist but the objects its pointers refer to will have already been destroyed.

src/controllers/highlights/ConfigureDialog.cpp
3r01, line 228: The pointer to the combo box will not be null if it is destroyed so this can cause a use-after-free if the form is closed while the request for badge icons hasn't returned yet. Can solve by capturing a `QPointer` in the lambda instead

src/controllers/highlights/types/AllForward.hpp
```diff
@@ -0,0 +1,54 @@
...
+struct UserHighlight;
+struct BadgeHighlight;
+struct UncategorizedNotificationHighlight;
+
+// clang-format off
+/// Variant of all highlights.
+///
+/// When you add a new built-in highlight, it must be added to HighlightController::billTinHighlights and HighlightController::recreateMissingBillTinHighlights. Ensure the order of the latter makes sense. The BillTinHighlights test in HighlightController will also need to be updated.
+/// When you add a new user-defined highlight, it must be added to the HighlightControllerTest.BillTinHighlightsHighlightController test, and to the highlights/types/Common.cpp isUserDefined function.
+/// This variant must be kept in-sync with the one in All.hpp
```
Nerixyz: This has the same comment as in `All.hpp`. I think you can remove the declaration in `All.hpp` and include `AllForward.hpp` there.

src/controllers/highlights/types/Outcome.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    this->sound = newSound;
+
+    this->updateSoundURL();
+}
+
+std::shared_ptr<QColor> Outcome::getBackgroundColorWithDefault(
+    const QColor &defaultColor) const
+{
+    if (!this->resolvedBackgroundColor)
```
Nerixyz: Can the resolved background color ever be empty?

src/controllers/highlights/types/Outcome.hpp
```diff
@@ -0,0 +1,112 @@
...
+    /// An empty string means no sound is played.
+    /// A string matching one of our built-in sounds (e.g. "001-ping2") means it will play the given resource.
+    /// An absolute file URL, probably prefixed with file:/// means it will play the custom sound at that path.
+    QString sound;
+
+    void setSound(const QString &newSound);
+
+    /// Contains the URL that should actually be played when a highlight is triggered, or invalid/empty if no sound should be played.
+    /// Transient. Not stored as-is in the JSON.
+    QUrl soundURL;
```
Nerixyz: Is this ever set outside this class? If not, it could be private and the class could expose only a getter.

src/controllers/highlights/types/AnnouncementsHighlight.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+}  // namespace chatterino::highlights
+
+namespace pajlada {
+
+template <>
+struct Serialize<chatterino::highlights::AnnouncementsHighlight> {
+    using H = chatterino::highlights::AnnouncementsHighlight;
+
+    static rapidjson::Value get(const H &h,
```
Nerixyz, line 66: We usually didn't do this, but these methods can be declared in the source file. Not much to gain because we still need the `Serialize` template, so we need to include `pajlada/serialize/serialize.hpp`.

src/controllers/highlights/types/AnnouncementsHighlight.hpp
```diff
@@ -0,0 +1,116 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "controllers/highlights/types/Common.hpp"
+#include "controllers/highlights/types/Outcome.hpp"
+#include "pajlada/serialize/deserialize.hpp"
+#include "pajlada/serialize/serialize.hpp"
```
Nerixyz: We usually include these in angle brackets.

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+            l->addRow(lbl, w);
+
+            addSettingMenu(lbl, w, [this, w] {
+                std::visit(
+                    [](auto &&h) {
+                        h.outcome.showInMentions = std::nullopt;
+                    },
+                    this->data);
+                w->setChecked(shouldShowInMentions(this->data));
+            });
```
3r01, line 363: Missing `QSignalBlocker`

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+                             });
+            l->addRow(lbl, w);
+
+            addSettingMenu(lbl, w, [this, w] {
+                std::visit(
+                    [](auto &&h) {
+                        h.outcome.alert = std::nullopt;
+                    },
+                    this->data);
+                w->setChecked(shouldAlert(this->data));
```
3r01, line 389: Missing `QSignalBlocker`

src/controllers/highlights/types/Common.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+/// Get the configured sound string - same as stored in the settings.
+/// If a default sound is enabled, and no string is configured, it returns the default sound string.
+QString getSound(const AllHighlights &h);
+
+/// Gets the raw string the user has configured as the sound for the highlight.
+QString getSoundWithoutDefault(const AllHighlights &h);
+
+/// Returns the default sound of the highlight, as defined by the `SOUND_DEFAULT` static string.
+QStringView getDefaultSound(const AllHighlights &h);
```
Nerixyz, line 49: All the functions taking `const AllHighlights&` feel like they should be methods on `AllHighlights`. We can't do that, but could wrap the variant:
```cpp
struct AllHighlights {
  QStringView getID() const;
  // ...
  std::variant<...> v; // Or use a typedef here.
};
```

src/controllers/highlights/types/FilterHighlight.hpp
```diff
@@ -0,0 +1,173 @@
...
+#include "controllers/highlights/types/Outcome.hpp"
+#include "util/RapidjsonHelpers.hpp"
+
+#include <pajlada/serialize/common.hpp>
+#include <qcolor.h>
+#include <qdebug.h>
+#include <qpixmap.h>
+#include <qregularexpression.h>
+#include <QStringView>
+#include <qurl.h>
```
Nerixyz:
```suggestion
#include <QColor>
#include <QDebug>
#include <QStringView>
```

src/controllers/highlights/types/FirstMessageHighlight.cpp
```diff
@@ -0,0 +1,55 @@
...
+                return std::nullopt;
+            }
+
+            return HighlightResult{
+                .alert = highlight.outcome.alert.value_or(H::ALERT_DEFAULT),
+                .sound = highlight.outcome.soundURL,
+                .color = highlight.outcome.getBackgroundColor(),
+                .showInMentions = highlight.outcome.showInMentions.value_or(
+                    H::SHOW_IN_MENTIONS_DEFAULT),
+            };
```
Nerixyz: A lot of these look similar across highlight types. What if there was something like a `Outcome::makeSimpleResult<H>()`?

src/controllers/highlights/types/SubscribedThreadHighlight.hpp
```diff
@@ -0,0 +1,112 @@
...
+    static constexpr bool SHOW_IN_MENTIONS_DEFAULT = true;
+    static constexpr bool ALERT_DEFAULT = true;
+    static constexpr QStringView SOUND_DEFAULT = u"001-ping2";
+    static constexpr QColor BACKGROUND_COLOR_DEFAULT = QColor(143, 48, 24, 60);
+
+    // Default state:
+    // Enabled = true
+    // Show in mentions = true
+    // Flash taskbar = true
+    // Play sound = true
```
Nerixyz: Comment still necessary? The above statics basically say the same.

src/controllers/highlights/types/SubscriptionsHighlight.hpp
```diff
@@ -0,0 +1,114 @@
...
+    static constexpr bool SHOW_IN_MENTIONS_DEFAULT = false;
+    static constexpr bool ALERT_DEFAULT = false;
+    static constexpr QColor BACKGROUND_COLOR_DEFAULT =
+        QColor(196, 102, 255, 100);
+
+    SubscriptionsHighlight() = default;
+
+    // Default state:
+    // Enabled = true
+    // Show in mentions = unavailable (always false)
```
Nerixyz: Same as above, and this looks inaccurate, as `highlight.outcome.showInMentions.value_or(H::SHOW_IN_MENTIONS_DEFAULT)` can return `true`.

src/controllers/highlights/types/UncategorizedNotificationHighlight.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace chatterino::highlights {
+
+/// Matches any special "notifications" that aren't already categorized
+struct UncategorizedNotificationHighlight {
+    static constexpr QStringView ID = u"uncategorizednotification";
+    static constexpr QStringView ICON_RESOURCE =
+        u":/buttons/settings-darkMode.svg";
+
+    static constexpr QStringView DEFAULT_NAME = u"Uncategorized Notification";
```
Nerixyz, line 41: I think this could have a `DESCRIPTION` similar to the filter highlight that explains what this might include.

src/controllers/highlights/types/UserHighlight.hpp
```diff
@@ -0,0 +1,172 @@
...
+        this->username = newValue;
+    }
+
+    /// The display name/pretty name of this highlight.
+    /// If empty, we will try to auto-generate something that makes sense (e.g. "Text contains 'foo'")
+    QString name;
+
+    std::optional<bool> enabled;
+
+    Outcome outcome{BACKGROUND_COLOR_DEFAULT};
```
Nerixyz: Not specific to this highlight: This is the same in every[?] highlight - what if this was in a base class?

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+                       //
+                   },
+                   [this, formLayout](auto &&h) {
+                       auto *lbl = new QLabel("Enabled");
+                       auto *w = new QCheckBox;
+                       w->setChecked(highlights::isEnabled(this->data));
+
+                       QObject::connect(w, &QCheckBox::checkStateChanged,
+                                        [&](auto checkstate) {
+                                            h.enabled = checkstate;
```
Nerixyz: scary

src/messages/layouts/MessageLayout.cpp
```diff
@@ -385,84 +371,24 @@ void MessageLayout::updateBuffer(QPixmap *buffer,
...
-                  MessageFlag::RedeemedChannelPointReward)) &&
-             ctx.preferences.enableRedeemedHighlight)
-    {
-        // Blend highlight color with usual background color
-        backgroundColor =
-            blendColors(backgroundColor,
-                        *ctx.colorProvider.color(ColorType::RedeemedHighlight));
-    }
-    else if (this->message_->flags.has(MessageFlag::AutoMod) ||
-             this->message_->flags.has(MessageFlag::LowTrustUsers))
```
3r01, line 440: `MessageFlag::LowTrustUsers` and `MessageFlag::RedeemedChannelPointReward` aren't handled in the new system but were handled here before

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+                    }
+                    else if (auto sound = data.toString(); sound.isNull())
+                    {
+                        auto fileUrl = QFileDialog::getOpenFileUrl(
+                            this, tr("Open Sound"), QUrl(),
+                            tr("Audio Files (*.mp3 *.wav)"));
+                        if (fileUrl.isValid())
+                        {
+                            this->previousSoundIndex = index;
+                            // soundURLLabel->setText(fileUrl.toLocalFile());
```
Nerixyz: Commented code.

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,661 @@
...
+                        w->setCurrentText(defaultSoundXD->displayName);
+                        soundURLLabel->setText("Playing bill-tin " %
+                                               defaultSoundXD->displayName);
+                    }
+                }
+            });
+
+            /*
+            // TODO: Implement per-highlight sound volume support!
+            auto *slider = new QSlider(Qt::Orientation::Horizontal);
```
Nerixyz, line 582: Commented code.

src/controllers/highlights/HighlightController.cpp
```diff
@@ -456,41 +100,21 @@ void HighlightController::rebuildChecks(Settings &settings)
...
 
-    auto currentUser = getApp()->getAccounts()->twitch.getCurrent();
-    auto self = (senderName == currentUser->getUserName());
-
     for (const auto &check : *checks)
     {
-        if (auto checkResult = check.cb(args, twitchBadges, senderName,
-                                        originalMessage, messageFlags, self);
-            checkResult)
+        if (auto checkResult = check.cb(params); checkResult)
```
Nerixyz:
```suggestion
        if (auto checkResult = check.cb(params))
```

src/widgets/dialogs/ReplyThreadPopup.cpp
```diff
@@ -144,7 +144,9 @@ ReplyThreadPopup::ReplyThreadPopup(bool closeAutomatically, Split *split)
     layout->setContentsMargins(marginPx, marginPx, marginPx, marginPx);
 
     // Top Row
-    bool addCheckbox = getSettings()->enableThreadHighlight;
+    // TODO: how do I solve this?
+    // bool addCheckbox = getSettings()->enableThreadHighlight;
+    bool addCheckbox = true;
```
Nerixyz: Maybe `bool addCheckbox = getHighlights()->hasAnyReplyThreadHighlight()`? That would check if there's any enabled reply thread highlight.

src/widgets/settingspages/HighlightingWidget.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+namespace {
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+const auto &LOG = chatterinoHighlights;
+
+/// Tips that will be randomized once on launch and shown at the top of the settings highlights page
+const constexpr std::array<QStringView, 3> TIPS{
+    uR"(Filter highlights are handy for advanced highlights.)",
+    uR"(User highlights highlights any message from the given user.)",
```
Nerixyz:
```suggestion
    uR"(User highlights highlight any message from the given user.)",
```

src/widgets/Window.cpp
```diff
@@ -374,7 +374,8 @@ void Window::addShortcuts()
     HotkeyController::HotkeyMap actions{
         {"openSettings",  // Open settings
          [this](std::vector<QString>) -> QString {
-             SettingsDialog::showDialog(this);
+             SettingsDialog::showDialog(this,
+                                        SettingsDialogPreference::Highlights);
```
Nerixyz, line 378: Intentional?

tests/snapshots/Settings/cleanup/user-defined-highlights.json
```diff
@@ -0,0 +1,104 @@
...
+                    "id": "8dae323f-fc79-4a18-94e7-01423ed0d612",
+                    "name": "broadcaster",
+                    "sound": "003-cat-fox_alex-fx-jump-3",
+                    "type": "badge"
+                },
+                {
+                    "id": "8dae323f-fc79-4a18-94e7-01423ed0d612",
+                    "name": "moderator",
+                    "sound": "003-cat-fox_alex-fx-jump-3",
+                    "type": "badge"
```
Nerixyz, line 51: What if some property is different between the two here?

tests/src/HighlightController.cpp
```diff
@@ -228,232 +253,281 @@ class HighlightControllerTest : public ::testing::Test
...
+                    .testName = "[User Highlight] Match 001",
+                    .args = MessageParseArgs{},   // no special args
+                    .badges = {},                 // no badges
+                    .senderName = "pajlada",      // sender name
+                    .originalMessage = "hello!",  // original message
+                    .runContext = ctx,
+                },
+            .expected =
+                {
+                    // expected
```
Nerixyz: These comments aren't necessary now with the designated initializers.

Nerixyz (commented):
Some of these comments are just ideas.

src/widgets/Window.cpp
pajlada: For now, yes. Will revert before merge. It's part of the list of TODO's in `highlights-rework.md`:
> - Return the default settings "tab"

src/widgets/settingspages/HighlightingWidget.cpp
pajlada: Thanks, fixed in 6772d1c8086710e8bac59b130a124231bd5c2b54

src/controllers/highlights/types/AnnouncementsHighlight.hpp
pajlada: Fixed in 536f1793646c7ee74e6673b1402fe2e634197115

There are a few remaining quoted includes I found with `ast-grep scan` and this rule:
```
id: pajlada-quoted-include
language: Cpp
rule:
  pattern: '"$HEADER"'
constraints:
  HEADER:
    regex: "^pajlada/"
fix: "<$HEADER>"
```
but didn't end up changing them since I didn't introduce them (in this PR at least B))

tests/src/HighlightController.cpp
pajlada: Fixed in b7b681d14b18071ae59c2e8c52c643a906c3982d

tests/snapshots/Settings/cleanup/user-defined-highlights.json
pajlada: It will always retain the properties of the top-most (most prioritized) highlight. I have added more properties that differ to exacerbate the effect here 0f19796d7b45e35d23e5f8818964300403dcd299

src/singletons/Settings.cpp
pajlada: There's no perfect way to handle this, so since there will be a breaking change I'd rather just have that breaking change apply to every scenario.

If you've previously used a custom default sound, you'll need to change individual sounds going forward. Clarified this in 18ba276972f0c05a91eb1fdcf7aee1c196cf8666

src/providers/twitch/eventsub/MessageBuilder.cpp
pajlada: This is the header, it doesn't have any message content. So here, yes.
Just below this I do the same thing, and there we should include the message body. 19ac1a8179376969d6e2032246bdf441a3c375a1 fixes it

src/controllers/highlights/types/AllForward.hpp
pajlada: Good idea, resolved in 4ab09bc672a012d6df3035638d36892c39420f56

src/controllers/highlights/types/FilterHighlight.hpp
pajlada: Fixed in 735459b7d422e0e23fed95d59494da118baaf39e

src/controllers/highlights/types/SubscribedThreadHighlight.hpp
pajlada: Fixed in ccf5544245b5eb44d909f90b3a476784bc819c7a

src/controllers/highlights/types/SubscriptionsHighlight.hpp
pajlada: Fixed in c18db693849e0a0df037f358c8aa81cd57e8d20d

src/controllers/highlights/types/UncategorizedNotificationHighlight.hpp
pajlada: "Fixed" in b1c859728605579fb2587c1fc9fa0d07d0349517. I'm sure the description can be better.

src/controllers/highlights/ConfigureDialog.cpp
pajlada: Will keep this as it's a fast-followup for per-highlight volume sliders. Lazy of me.

src/controllers/highlights/ConfigureDialog.cpp
pajlada: Removed in 36837e3e5072892dcc374bdd702728d9354d297d

src/controllers/highlights/HighlightController.cpp
pajlada: Resolved in 3c13100d5f617524ab01f9e3bd6ce21786aca422

src/widgets/dialogs/ReplyThreadPopup.cpp
pajlada: Resolved in 15a54295a3986b1d9b0869ee3d863c295bcdb890 - a bit ugly but it's OK for now.

src/controllers/highlights/types/Outcome.hpp
pajlada: Fixed in b49d0a3edb971cac0d6b546aa682060d95552ca6

src/controllers/highlights/types/FirstMessageHighlight.cpp
pajlada: Yep, that ended up working well. Fixed in f35b01aced5851a92f4e26a1897e1e4d21b50c73

There was only one outlier that couldn't make use of this function (AnnouncementHighlights because it's responsible for reading the incoming IRC tags to decide the background color)

src/controllers/highlights/types/Outcome.cpp
pajlada: This was unused, so removed in 6a40cf07ba6c1d1e5b785e0fe1dfad347dd11fc2.
The function that actually reads this has a similar check but with asserts `getBackgroundColor`.

src/controllers/highlights/ConfigureDialog.cpp
pajlada: Fixed in 4b513513ed6a094982714d1d2eddc810b4561e00

src/controllers/highlights/ConfigureDialog.cpp
pajlada: Fixed in 4b513513ed6a094982714d1d2eddc810b4561e00

src/controllers/highlights/Model.cpp
pajlada: Fixed in 62b0135b101663f0ea1d27ce6b2ab1a9d8cb4822

src/controllers/highlights/ConfigureDialog.cpp
pajlada: Fixed in 62b0135b101663f0ea1d27ce6b2ab1a9d8cb4822

src/controllers/highlights/types/Outcome.hpp
pajlada: Noted in c87f100caf56e7779caefd982887d21f7ab94b17, not actually fixed

src/controllers/highlights/types/Common.hpp
pajlada: I see the gains from this but the refactor would be ass so I'm leaving it for now.

src/controllers/highlights/types/UserHighlight.hpp
pajlada: Without inheritance, which I'd prefer to avoid while using variants (not really a preference I've actually reasoned about), this change is also ass to make so I'll leave this for now.

src/controllers/highlights/types/AnnouncementsHighlight.hpp
pajlada: Not going to do this now, but if deemed reasonable this could be done in a follow-up PR.

highlights-rework.md
pajlada: I added a note about this in the same document so I don't forget it :) 9637e4915186b52254f4a76250597b432e6be5f7

src/messages/layouts/MessageLayout.cpp
pajlada: `MessageFlag::RedeemedChannelPointReward` is now being handled just like `RedeemedHighlight` as of 8ae3771a99008b730227f227677888ba2896b082

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,636 @@
...
+                                [fileUrl](auto &&h) {
+                                    h.outcome.setSound(fileUrl.toString());
+                                },
+                                this->data);
+                            soundURLLabel->setText(u"Playing custom " %
+                                                   fileUrl.toString());
+                            QSignalBlocker block(w);
+                            w->removeItem(numRows);
+                            w->addItem(fileUrl.fileName());
+                            w->setCurrentIndex(numRows);
```
3r01, line 526: Needs `previousSoundIndex = numRows;` here?

Or could we move custom sound out to a button maybe? It's not a real option and having it in the combo makes this selection/restore logic difficult to follow

src/controllers/highlights/Model.cpp
3r01: https://github.com/Chatterino/chatterino2/blob/6a8ae9dbc2e185984556cfdd17934f64ccaec299/src/common/SignalVectorModel.hpp#L76-L98

Keeping ownership of the model is not enough because if the remove event is called on it then it will delete its items but not be destroyed itself, line 96 in above quote 

Could solve by capturing the highlight id and looking the row up again when the callback runs

src/messages/MessageBuilder.cpp
3r01, line 1329: Needs `parseHighlights()`

8thony (commented):
On Windows the Highlight editor looks like a blank menu.
Image
Tested on Windows 10 & 11 with commit https://github.com/Chatterino/chatterino2/commit/10a144c4 built with Qt 6.8.3, from [this Actions run](https://github.com/Chatterino/chatterino2/actions/runs/35535802924), default settings

src/singletons/Settings.cpp
```diff
@@ -63,23 +290,546 @@ void _actuallyRegisterSetting(
...
+                break;
+            case pajlada::Settings::SettingManager::SaveResult::Skipped:
+                qCInfo(LOG) << "Skipped saving settings after migration";
+                break;
         }
     }
+}
 
-    return false;
+void Settings::migrateHighlights(bool isTest)
```
4rneee, line 331: How did you decide on the order of the highlights?

I'm not sure what a good order would be but I find the current order a bit confusing because it does neither represent how they are currently ordered in the settings (all system highlights in the beginning, then user highlights), nor does it reflect how highlights are [currently evaluated](https://github.com/Chatterino/chatterino2/blob/9c4ba35a974eb27db007955f82029ed7761b8535/src/messages/layouts/MessageLayout.cpp#L388-L465).

I think at least all system highlights should be in one group so that they are easier to find because they are a bit all over the place right now.

src/controllers/highlights/ConfigureDialog.cpp
```diff
@@ -0,0 +1,636 @@
...
+
+    QObject::connect(buttonBox, &QDialogButtonBox::accepted, this, [this] {
+        Q_EMIT this->confirmed(this->data);
+        this->close();
+    });
+    QObject::connect(buttonBox, &QDialogButtonBox::rejected, this,
+                     &BasePopup::close);
+    dialogLayout->addWidget(buttonBox, 0, Qt::AlignRight);
+
+    this->setLayout(dialogLayout);
```
3r01, line 633: Needs to be `this->getLayoutContainer()->setLayout(dialogLayout);` otherwise the highlight editor window will be blank on Windows as described by @8thony

src/singletons/Settings.cpp
pajlada: Order isn't perfect, I'd like to have the order match the old order as-best-as-possible.
I don't want to group highlights just because it looks nicer in the UI, it should roughly match user's priority expectations.

If I read this correctly, the order that's closest to how it worked before would be:
1. First message
1. Watch Streak
1. Subscriptions
1. Whispers
1. User-defined Message highlights
1. User-defined User highlights
1. Subscribed-to threads
1. User-defined Badge highlights
1. Announcements
1. Channel point redemptions
1. Automod
1. Low Trust users (interchangeable with automod)
1. Uncategorized notification

Does that seem reasonable to you?

4rneee: Yes that sound reasonable.

I'm somehow unable to find in the code why subscriptions are higher than announcements and user defined highlights but my testing confirmed your proposed order.

PR #6702, opened 2026-01-03
feat(xdg): return only base or user directories

pajlada:
Our current `getXDGDirectories` implementation returns both the user
_and_ system directory for the given user type.
For some use cases (e.g. dictionaries), we want to have more granural
control.

I have added two new functions:
 - `getXDGBaseDirectories` which returns the base directories from `XDG_CONFIG_DIRS`
 - `getXDGUserDirectories` which returns the user directory from `XDG_CONFIG_HOME`

The original `getXDGDirectories` function is still intact and uses the two new functions.

### DIFF
--- a/src/util/XDGDirectory.cpp
+++ b/src/util/XDGDirectory.cpp
@@ -8,29 +8,8 @@ namespace chatterino {
 
 #if defined(Q_OS_UNIX) and !defined(Q_OS_DARWIN)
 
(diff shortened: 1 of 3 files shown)


ghost:
does this fix #5555 ?

pajlada:
> does this fix #5555 ?

It does not, all this does was improve our internal functions for figuring out the user/base XDG directories

PR #6703, opened 2026-01-03
feat: linux-only system dictionary spell checking

pajlada:
This PR moves the spellchecking settings to the bottom of External Tools.

Additionally, it adds a dropdown there where a linux user gets a list of system-installed dictionaries that can be falled back to if no user-specified dictionary is installed.

the `loadDictionariesFromDirectory` function can be used to load dictionaries from the dictionaries dir later instead of relying only on `index.aff`

### DIFF
--- a/src/widgets/settingspages/ExternalToolsPage.cpp
+++ b/src/widgets/settingspages/ExternalToolsPage.cpp
@@ -18,6 +18,8 @@
 #include <QMessageBox>
 #include <QPushButton>
 
+#include <algorithm>
+
 namespace chatterino {
 
 namespace {
@@ -260,10 +262,11 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
             };
         };
         std::vector<std::pair<QString, QVariant>> dictList{{"None", ""}};
-        std::ranges::copy(
-            std::views::transform(
-                getApp()->getSpellChecker()->getSystemDictionaries(), toItem),
-            std::back_inserter(dictList));
+
+        std::ranges::transform(
+            getApp()->getSpellChecker()->getSystemDictionaries(),
+            std::back_inserter(dictList), toItem);
+
         if (dictList.size() > 1)
         {
             SettingWidget::dropdown(


src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -12,22 +15,77 @@
...
+}  // namespace
+
 class SpellCheckerPrivate
 {
 public:
-    static std::unique_ptr<SpellCheckerPrivate> tryLoad();
+    static std::unique_ptr<SpellCheckerPrivate> tryLoad(
+        const QString &path = {});
 
+    /// NOTE: To support multiple dictionaries at the same time, it seems like we need to store a list of Hunspell instances, each supporting a single dictionary, and then during the spell checking process check each hunspell instance.
```
Nerixyz, line 64:
```suggestion
    /// FIXME: To support multiple dictionaries at the same time, it seems like we need to store a list of Hunspell instances, each supporting a single dictionary, and then during the spell checking process check each hunspell instance.
```

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -117,4 +187,55 @@ std::vector<std::string> SpellChecker::suggestions(const QString &word)
...
+    {
+        qCDebug(chatterinoSpellcheck)
+            << "Looking for system dictionaries in" << searchDirectory;
+        for (const auto &dict : loadDictionariesFromDirectory(searchDirectory))
+        {
+            if (dict.isSymbolicLink)
+            {
+                // NOTE: We currently filter out symbolic links from system-loaded dictionaries.
+                // Without this, the list of dictionaries we "support" would be too high on Linux distros.
+                // Installing en_US dictionary symlinks en_GH/en_HK/en_IE/en_IN/en_JM/en_NA/en-NG/en_NZ etc to that installed dictionary.
```
Nerixyz:
```suggestion
                // NOTE: We currently filter out symbolic links from system-loaded dictionaries.
                // Without this, the list of dictionaries we "support" would be too high on Linux distros.
                // Installing en_US dictionary symlinks en_GH/en_HK/en_IE/en_IN/en_JM/en_NA/en_NG/en_NZ etc to that installed dictionary.
```
---

So we don't show `en_NZ` even though it exists, for example?

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 - Dev: Simplified uses of `getMessageSnapshot`. (#6607)
 - Dev: Disabled `llvm-prefer-static-over-anonymous-namespace` in clang-tidy. (#6610)
 - Dev: Started work on spell checking. (#6446)
+- Dev: Enabled linux-only system dictionary spell checking fallback. (#6703)
```
Nerixyz: Add to the entry above (since it's not a released feature yet).

src/controllers/spellcheck/SpellChecker.cpp
```diff
@@ -117,4 +187,55 @@ std::vector<std::string> SpellChecker::suggestions(const QString &word)
 #endif
 }
 
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
```
Nerixyz, line 190: Why shouldn't this be static?

I think we'd actually want this method (and maybe others) in a `SpellCheckController`. If we allow users to select a dictionary per split, we'll have multiple checkers per app instance.

Maybe that's for the future.

src/controllers/spellcheck/SpellChecker.cpp
pajlada: Correct, because the user doesn't _actually_ have en_NZ installed, they have en_US and en_NZ just points to en_US.

There may be scenarios where a distro installs real dictionaries outside of our search paths, and only includes symlinks in the `/usr/share/hunspell` (or other) search directories, but I haven't run into that in my limited testing.

src/controllers/spellcheck/SpellChecker.cpp
pajlada: I see the `SpellChecker` class as the `SpellCheckController`.

My thinking for multiple dictionaries was to store a vector of Hunspell instances inside `SpellCheckerPrivate`.
This might be the wrong way to think if we support per-split dictionaries, but even then I can see that working.
so yeah, this seems like a future thought

CHANGELOG.md
pajlada: Resolved in [`481d9e4` (#6703)](https://github.com/Chatterino/chatterino2/pull/6703/commits/481d9e4f276832fb5ff8df4efa5b382c9205f1fc) - I've left it as dev since I'm not sure it will be included in the next stable release

src/controllers/spellcheck/SpellChecker.cpp
pajlada: Comment was improved in [`d2cf8ae` (#6703)](https://github.com/Chatterino/chatterino2/pull/6703/commits/d2cf8ae0ae817ce41965b1ac10ca8c6e8d43d1af)

src/widgets/settingspages/ExternalToolsPage.cpp
```diff
@@ -227,6 +231,52 @@ void ExternalToolsPage::initLayout(GeneralPageView &layout)
...
+        std::vector<std::pair<QString, QVariant>> dictList{{"None", ""}};
+
+        std::ranges::transform(
+            getApp()->getSpellChecker()->getSystemDictionaries(),
+            std::back_inserter(dictList), toItem);
+
+        if (dictList.size() > 1)
+        {
+            SettingWidget::dropdown(
+                "Fallback spellchecking dictionary (requires restart)",
```
Nerixyz, line 273: More of a discussion:

Should we call this "Fallback spellchecking dictionary"? We will want to load multiple dictionaries from `Dictionaries/` too (presumably with `loadDictionariesFromDirectory`). These dictionaries should show up in a list like this as well. Or the other way around: we wouldn't want to have two settings (one for system dictionaries and one for local ones). They should all be in the same setting.

Given that we have plans to overwrite a dictionary per split, this would be something like a "Default dictionary".

pajlada: I don't intend this setting to stay as-is, so I won't touch it. The biggest impact this PR has is simplicity of testing dictionaries/experimenting with the feature, and adding the function to look for dictionaries in a given directory

PR #6705, opened 2026-01-03
added hunspell for building deps

killertofus:


pajlada (changes requested):
Since the spell checker isn't enabled by default, `hunspell` isn't actually a build dependency.
I'm OK with approving this if you also modify the cmake call in the manual instructions and add `-DCHATTERINO_SPELLCHECK=On`

pajlada (approved):

PR #6709, opened 2026-01-04
feat: add filter for external badges

pajlada:


pajlada:
@achow101 I reimplemented your changes but in a different way, storing external badges in a separate vector and making that available to filters. Please let me know if this work as you would expect.

The intent is for this to be available for highlights after the rework too.

src/messages/MessageBuilder.cpp
```diff
@@ -2472,6 +2472,10 @@ void MessageBuilder::appendChatterinoBadges(const QString &userID)
     {
         this->emplace<BadgeElement>(*badge,
                                     MessageElementFlag::BadgeChatterino);
+
+        /// e.g. "chatterino:Chatterino Top donator"
+        this->message().externalBadges.emplace_back(u"chatterino:" %
+                                                    (*badge)->tooltip.string);
```
Nerixyz: Could we set the ID of the badge to this string? Same for the other badges.

pajlada: I'd prefer to use something that is "sticky" too.

For the Chatterino badge, it returns the badges in an array with the only data being the image links, a list of Twitch User IDs with the badge, and the tooltip/text of the badge (e.g. "Chatterino Top Donator"). I don't think there's anything better to key off of in this scenario.

For the FrankerFaceZ, I already use the badge name (e.g. "subwoofer" or "bot"). Do you prefer something other than this?

For 7TV we could use the badge ID (e.g. `01JEY956QEPVSFKH7KWS9V6HMQ`) or the badge name (e.g. `NNYS 2024`). Do you know if the badge ID stays if the asset is updated? e.g. if the NNYS 2024 badge got a new look, would the ID change? If the ID changes, I'd prefer to use the badge name. If the ID stays the same, I'm fine to use the ID.

I haven't checked BetterTTV yet.

src/messages/Message.hpp
```diff
@@ -61,6 +61,13 @@ struct Message {
     /// Map of extra data associated with each Twitch badge
     std::unordered_map<QString, QString> twitchBadgeInfos;
 
+    /// List of external badges associated with this message
+    /// The badge should follow the following format: "provider:badgename". e.g.:
+    ///  - betterttv:pro
+    ///  - frankerfacez:mod
+    ///  - 7tv:sub
+    QStringList externalBadges;
```
achow101, line 69: Should this have a field in `toJson()`?

achow101:
Other than not necessarily knowing what to put as the badge, seems to work as expected.

src/messages/Message.hpp
pajlada: Good catch, fixed in [`92234ac` (#6709)](https://github.com/Chatterino/chatterino2/pull/6709/commits/92234ac1b581366cddc71bb3e1b67a51f9cf7efb)

src/messages/MessageBuilder.cpp
Nerixyz: The reason I suggested this is to avoid creating new strings for every badge in every message. It looks like we could use the emote name to store `provider:name`, because the emote name is never used. 7TV already uses the ID (although that might not be used either).

> Do you know if the badge ID stays if the asset is updated? e.g. if the NNYS 2024 badge got a new look, would the ID change?

No, but a reasonable assumption would be that the name stays the same, but the ID changes (much like for emotes).

pajlada: This PR now stores the "external badge names" in the emote name

test snapshot from `tests/snapshots/IrcMessageHandler/badges.json` shows the difference

let me know if I misunderstood your feedback

Nerixyz: Yes, that's great!

Issue #6704, opened 2026-01-03
`Sort Tabs Alphabetically` still works when `Lock Tab Layout` is on


brian6932:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

1. Open some tabs with different names.
1. Right click the tab bar, and press `Lock Tab Layout`.
1. `Sort Tabs Alphabetically` still works, and isn't grayed out.

### OS and Chatterino Version

eb373d04557e2034b9ef3d78cfbf07ad8ca7129f

PR #6710, opened 2026-01-04
fix: disable sortTabsAlphabetically when notebook layout locked (#6704)

ruinivist:
Greys out the "Sort Tabs Alphabetically" option if the tabs are locked.
- the assert should help ensure not being locked as a precondition
- this handles disabling for the click event for locking layout as well as the first initialization of menu if it's locked already from settings

image

first pr FeelsStrongMan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unversioned
 
-- Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6704)
 - Minor: Add a separate highlight option for watchstreak notifications. (#6571, #6581)
 - Minor: Badges now link to their home page like emotes in the context menu. (#6437)
 - Minor: Fixed usercard resizing improperly without recent messages. (#6496)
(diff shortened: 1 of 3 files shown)


src/widgets/Notebook.cpp
```diff
@@ -1632,4 +1638,10 @@ void SplitNotebook::forEachSplit(const std::function<void(Split *)> &cb)
     }
 }
 
+void SplitNotebook::setLockNotebookLayout(bool value)
+{
+    this->Notebook::setLockNotebookLayout(value);
+    this->sortTabsAlphabeticallyAction_->setEnabled(!value);
```
pajlada: No need for `this->` here

```suggestion
    Notebook::setLockNotebookLayout(value);

    this->sortTabsAlphabeticallyAction_->setEnabled(!value);
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6704)
```
pajlada: Move this changelog entry to among the other `Bugfix` entries, and change the number from `6704` to `6710` - we point to the PR that made the change rather than the issue it fixes

pajlada (changes requested):
Just these nits, and as extra credit you can move the definition of `sortTabsAlphabetically()` into the `protected:` scope in the header file

pajlada:

PR #6717, opened 2026-01-06
ci: abort clang-tidy-review on error

pajlada:
CC @Nerixyz


Nerixyz:
Yes, that was my intention. Currently with this PR, we don't abort, though.

PR #6718, opened 2026-01-06
added gentoo/suse build deps

killertofus:


BUILDING_ON_LINUX.md
```diff
@@ -20,8 +20,17 @@ The built binary should be exportable from the final image & able to run on your
 sudo apt install qt6-base-dev qt6-svg-dev qt6-image-formats-plugins libboost-dev libnotify-dev libssl-dev libsecret-1-dev pkg-config cmake g++ git hunspell
 ```
 
-### Arch Linux
+### OpenSuse
+```sh
+sudo zypper install cmake qt6-tools-devel qt6-multimedia-devel libboost_filesystem-devel boost-devel libopenssl-devel libboost_headers-devel qt6-svg-devel git libnotify-devel qt6-concurrent-devel qt6-qt5compat-devel libsecret-devel qt6-dbus-devel hunspell
```
Nerixyz: I don't know anything about OpenSuse.

- Why is `libboost_filesystem-devel` needed? Do we link to boost filesystem? If so, we should probably stop that.
- Why is `libboost_headers-devel` needed? Shouldn't the headers be included in `boost-devel`?

ghost: the opensuse build spec requires it https://build.opensuse.org/projects/home:MKamelll/packages/chatterino2/files/chatterino2.spec?expand=1

pajlada: take a peek at this newly cooked 2.5.4-compatible fedora spec - it only requires `boost-devel`

https://src.fedoraproject.org/rpms/chatterino2/blob/rawhide/f/chatterino2.spec

PR #6722, opened 2026-01-07
feat(spellcheck): ignore links for spellchecking

2547techno:
This PR ignores links for spellchecking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - Minor: Add categories to the emoji viewer. (#6598)
 - Minor: Added broadcaster-only `/poll`, `/cancelpoll`, and `/endpoll` commands. (#6583, #6605)
 - Minor: Added broadcaster-only `/prediction`, `/cancelprediction`, `/lockprediction`, and `/completeprediction` commands. (#6583, #6612, #6632)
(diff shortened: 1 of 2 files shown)


Mm2PL:
How is the performance of this with large amounts of links?

2547techno:
good question, i dont have metrics but with ~2k links (a.ca) i don't feel any sort of hiccup


https://github.com/user-attachments/assets/b873ce92-2afd-4370-91dc-3fe01094eeb1

src/widgets/splits/InputHighlighter.cpp
```diff
@@ -49,11 +50,17 @@ bool isIgnoredWord(TwitchChannel *twitch, const QString &word)
...
+    if (getApp()
+            ->getAccounts()
+            ->twitch.getCurrent()
+            ->twitchEmote(name)
+            .has_value())
+    {
+        return true;
+    }
+
+    auto link = linkparser::parse(word);
```
pajlada, line 63:
```suggestion
    // TODO: Replace this with a link parser variant that doesn't return the parsed data
    auto link = linkparser::parse(word);
```

pajlada (changes requested):

Issue #6578, opened 2025-11-14
Automatic username highlight does not update on login name change


jupjohn:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

When a user renames themselves on Twitch, the automatic username highlight will continue to ping on the original login instead of the new login - even after a restart.

This is a wider issue than just highlights as the account entry in the settings file still contains the old name, but this is the most common *symptom* that users notice.

### Screenshots

Most recent occurrence of this:

<img width="555" height="454" alt="Image" src="https://github.com/user-attachments/assets/1fea5faa-ec79-4fe3-a8ae-51e80d1db751" />

### OS and Chatterino Version

Chatterino 2.5.4 (commit f35592807) built with Qt 6.10.0 Running on Arch Linux, kernel: 6.17.7-zen1-1-zen

jupjohn:
Given that Chatterino will *eventually* move away from IRC, I feel like the best place to check this is on startup/connect against Helix.

Nerixyz:
We already check for missing scopes:
https://github.com/Chatterino/chatterino2/blob/c72d6707e36dbbea32ac1c7877a758155f3fda59/src/providers/twitch/TwitchAccountManager.cpp#L208
It shouldn't be too hard to extract the login name from the response. The "challenge" is to properly notify other components of the name change. Maybe this should invoke the same signal again?

I'd consider this a good first issue.

SonicSegFault:
Hi,

I didn’t find code where the username is refreshed, it seems static after boot so the "challenge" is just to implement the signal system? 

I don't think username is refreshed from Twitch's API, can someone correct me if I am wrong.

Mm2PL:
> *SonicSegFault* left a comment
> 
> I didn’t find code where the username is refreshed, it seems static after boot
I don't think we ever refresh the username. This is the core of the issue.

> so the "challenge" is just to implement the signal system?
> 
> I don't think username is refreshed from Twitch's API, can someone 
> correct me if I am wrong.

["https://id.twitch.tv/oauth2/validate"][1] gives us a reply that does 
have the login name, we already use that endpoint as nerixyz pointed 
[...]

PR #6723, opened 2026-01-07
Fix highlight mentions not updating when username changes

SonicSegFault:
This PR ensures mention highlights update correctly when the account username changes. 
Related to Issue:#6578

- It adds a notification on `TwitchAccount` when the username changes.
- Rebuilds highlight checks using the existing rebuild path

A `setUserName()` API on `TwitchAccount` that notifies dependents when the username changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@
 - Dev: Unwrapped `LimitedQueueSnapshot` to `std::vector`. (#6606)
 - Dev: Simplified uses of `getMessageSnapshot`. (#6607)
 - Dev: Disabled `llvm-prefer-static-over-anonymous-namespace` in clang-tidy. (#6610)
-- Dev: Added experimental spell checker support. (#6446, #6703, #6722, #6730)
+- Dev: Added experimental spell checker support. (#6446, #6703, #6722, #6730, #6731)
 - Dev: Added Clazy linting in CI. (#6623)
 - Dev: Added custom clang-tidy module linting in CI. (#6626)
 - Dev: CMake option `USE_ALTERNATE_LINKER` now errors if the given linker can't be found. (#6692)
--- a/src/controllers/commands/CommandController.cpp
+++ b/src/controllers/commands/CommandController.cpp
@@ -733,4 +733,36 @@ QStringList CommandController::getDefaultChatterinoCommandList()
     return this->defaultChatterinoCommandAutoCompletions_;
 }
 
+qsizetype CommandController::commandTriggerLen(QStringView text)
+{
+    auto words = text.split(' ');
+
+    qsizetype triggerLen = 0;
+    qsizetype spaces = 0;
+    QString commandName{};
+
+    for (qsizetype i = 0; i < words.length() && spaces <= this->maxSpaces_; ++i)
+    {
+        commandName += words[i];
+        triggerLen += words[i].length();
+
+        if (this->commands_.contains(commandName) ||
+            this->userCommands_.contains(commandName))
+        {
+            return triggerLen;
+        }
+
+        // ignore consecutive spaces
+        if (!words[i].isEmpty())
+        {
+            commandName += ' ';
+            ++spaces;
+        }
+        // account for the space between the current and next word
+        ++triggerLen;
+    }
+
+    return 0;
+}
+
 }  // namespace chatterino
--- a/src/controllers/commands/CommandController.hpp
+++ b/src/controllers/commands/CommandController.hpp
@@ -55,6 +55,13 @@ public:
     }
 #endif
 
+    /// Returns the length of the command trigger in `text`, including leading
+    /// spaces. If `text` does not start with a command trigger, 0 is returned.
+    /// Examples:
+    ///  - " /ban forsen" returns 5
+    ///  - "/non-existing-command" returns 0
+    qsizetype commandTriggerLen(QStringView text);
+
 private:
     void load(Paths &paths);
 
--- a/src/widgets/splits/InputHighlighter.cpp
+++ b/src/widgets/splits/InputHighlighter.cpp
@@ -8,6 +8,7 @@
 #include "common/Aliases.hpp"
 #include "common/LinkParser.hpp"
 #include "controllers/accounts/AccountController.hpp"
+#include "controllers/commands/CommandController.hpp"
 #include "controllers/spellcheck/SpellChecker.hpp"
 #include "messages/Emote.hpp"
 #include "providers/bttv/BttvEmotes.hpp"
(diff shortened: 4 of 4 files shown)


src/providers/twitch/TwitchAccount.cpp
```diff
@@ -194,6 +190,17 @@ void TwitchAccount::blockUserLocally(const QString &userID,
     this->ignoresUserLogins_.insert(blockedUser.name);
 }
 
+bool TwitchAccount::setUserName(const QString &newUserName)
```
Nerixyz, line 197: This function is never called.

src/controllers/highlights/HighlightController.cpp
```diff
@@ -442,6 +438,16 @@ HighlightController::HighlightController(Settings &settings,
...
 
     this->rebuildChecks(settings);
+
+    auto currentUser = accounts->twitch.getCurrent();
+    this->bConnections.emplace_back(
+        currentUser->userNameChanged.connect([this, &settings] {
+            qCDebug(chatterinoHighlights)
+                << "Rebuild checks because username changed";
+            this->rebuildChecks(settings);
+        }));
```
Nerixyz: This doesn't take changes to the current Twitch account into account (i.e. changing the current account).

SonicSegFault: > This doesn't take changes to the current Twitch account into account (i.e. changing the current account).

I think adding a disconnect to the prev `currentUserChanged` should do it. So we no longer hold a stale listener for the old account. and attach a new listener for the changed account

pajlada (changes requested):
Make the `userNameChanged` signal a `pajlada::Signals::NoArgSignal` and move it into `TwitchAccountManager`. `TwitchAccount::setUserName` should stay and return a bool, and `TwitchAccountManager` should call the signal based on whether that bool changed.

By making these changes, the logic in `HighlightController` stays cleaner since you don't have to mess with reconnecting signals, and `TwitchAccount` stays a little bit more data-oriented.

Thanks for taking a look at this! Feel free to ask if you have any questions.

oh also, no need to force push. we squash PRs at the end anyway :)

Nerixyz:
I just realized that we don't need any new signal (for highlights). They already use the current account when checking:
https://github.com/Chatterino/chatterino2/blob/9c4ba73af98e87a99569c674073091d70f276946/src/controllers/highlights/HighlightController.cpp#L480-L481

As for where we should update the name:

`addUser` is called when adding an account through the UI and when loading the user from the settings. We're mainly interested in the second case - from the issue:

> When a user renames themselves on Twitch, the automatic username highlight will continue to ping on the original login instead of the new login - even after a restart.

When restarting, we also update the current user (`currentUserChanged`) [here](https://github.com/Chatterino/chatterino2/blob/9c4ba73af98e87a99569c674073091d70f276946/src/providers/twitch/TwitchAccountManager.cpp#L365). Now, when the current user changes, we also check their token [here](https://github.com/Chatterino/chatterino2/blob/9c4ba73af98e87a99569c674073091d70f276946/src/providers/twitch/TwitchAccountManager.cpp#L212). That endpoint also returns the user's current name (see [docs](https://dev.twitch.tv/docs/authentication/validate-tokens#how-to-validate-a-token)). So `checkMissingScopes` should also take the account and update the username from the response. It needs to save the user data as well. See [LoginDialog.cpp](https://github.com/Chatterino/chatterino2/blob/9c4ba73af98e87a99569c674073091d70f276946/src/widgets/dialogs/LoginDialog.cpp#L65-L66)[^1].

You can test this locally by closing your Chatterino test instance and editing the `settings.json`: Change `accounts.uid{your ID}.username` to something else. When restarting Chatterino, you should see that highlights for your user work again.
Tip: try it once without your changes - you should see that highlights for your user don't work.

[^1]: Maybe we should move the saving somewhere else - in #5680 I added [`TwitchAccountData`](https://github.com/Nerixyz/chatterino2/blob/43b59a5a16dadc1d1b0eba473a7b1f5ba5a3334b/src/providers/twitch/TwitchAccount.hpp#L165-L176).

Nerixyz:
Okay, we probably _do_ need some signal for the UI (account switcher and split input). Maybe that's what I meant by 
> The "challenge" is to properly notify other components of the name change.

No one knows. But I think for now, it's fine if you just update the name in the account, as it doesn't impact functionality.

SonicSegFault:
Updated the account’s username based on the /oauth2/validate response.

I also looked into persisting the updated username and emitting a UI signal (for the account switcher / split input), but doing so here would require pulling in settings or UI-related headers into `TwitchAccountManager`, and it’s not entirely clear which signal is appropriate.

Since updating the account alone fixes the functional issue and matches the suggested approach, I kept the change minimal for now. 

Happy to follow up if there’s a preferred place to handle persistence or UI notification.

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Changelog
+#Changelog
```
pajlada: Seems like the changelog entries have been mangled somehow, please undo the unrelated changes

pajlada (approved):
This fixes the linked issue and looks good as-is.
Please fix the changelog issue and then I'm happy to merge this in.

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -50,20 +50,28 @@ QString missingScopes(const QJsonArray &scopesArray)
...
                 return;
             }
 
             const auto json = res.parseJson();
+
+            const auto login = json["login"_L1].toString();
+            if (!login.isEmpty() &&
+                login.compare(account->getUserName(), Qt::CaseInsensitive) != 0)
+            {
+                account->setUserName(login);
```
Nerixyz, line 73: We need to save the username afterward.

CHANGELOG.md
```diff
@@ -78,7 +78,7 @@
 - Dev: Unwrapped `LimitedQueueSnapshot` to `std::vector`. (#6606)
 - Dev: Simplified uses of `getMessageSnapshot`. (#6607)
 - Dev: Disabled `llvm-prefer-static-over-anonymous-namespace` in clang-tidy. (#6610)
-- Dev: Added experimental spell checker support. (#6446, #6703, #6722, #6730)
+- Dev: Added experimental spell checker support. (#6446, #6703, #6722, #6730, #6731)
```
Nerixyz: Add this as a separate entry.

src/providers/twitch/TwitchAccountManager.cpp
pajlada: No need to do that in this PR

pajlada (changes requested):
Sorry for the confusion / mixed signals (hehe). The signal to notify `HighlightController` of an update _is_ needed.

The linked part of `HighlightController` from https://github.com/Chatterino/chatterino2/pull/6723#issuecomment-3725678411 is responsible for checking if the message is your own, not whether someone typed your name or not.

So in the PRs current state, the issue only gets fixed on startup if the user performs some other change that would rebuild highlights (e.g. toggle one of the built-in highlights).

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -69,7 +70,8 @@ void checkMissingScopes(const std::shared_ptr<TwitchAccount> &account)
             if (!login.isEmpty() &&
                 login.compare(account->getUserName(), Qt::CaseInsensitive) != 0)
             {
-                account->setUserName(login);
+                account->setUserName(login);                
+                getApp()->getAccounts()->twitch.currentUserChanged();
```
pajlada: `currentUserChanged` is used for quite a lot of signals, so preferably this uses a separate signal, similar to how you implemented it before.

e.g.
```c++
// in TwitchAccountManager.hpp
pajlada::Signals::NoArgSignal currentUserNameChanged;
```

```c++
// near the bottom of HighlightController constructor
this->signalHolder_.managedConnect(
    accounts->twitch.currentUserNameChanged,
    [this, &settings] {
        qCDebug(chatterinoHighlights)
            << "Rebuild checks because user name changed";
        this->rebuildChecks(settings);
    });
```

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -50,20 +51,29 @@ QString missingScopes(const QJsonArray &scopesArray)
...
             }
 
             const auto json = res.parseJson();
+
+            const auto login = json["login"_L1].toString();
+            if (!login.isEmpty() &&
+                login.compare(account->getUserName(), Qt::CaseInsensitive) != 0)
+            {
+                account->setUserName(login);
+                getApp()->getAccounts()->twitch.currentUserNameChanged.invoke();
```
pajlada: we've already gotten the app here
```suggestion
                app->getAccounts()->twitch.currentUserNameChanged.invoke();
```

Issue #6635, opened 2025-12-08
Open in custom player doesn't work


kasper93:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

The stream URL is missing `:` when opening custom URI scheme.

1. Set URI scheme to `mpv://`
2. Try to open `<your_favorite_streamer>` stream in custom player
3. mpv opens, but errors on not knowing how to handle the stream URL with missing `:`
4. If we look at the log, we can see the issue
```
[   0.000][v][cplayer] Command line options: '--player-operation-mode=pseudo-gui' '--' 'mpv://https//www.twitch.tv/<your_favorite_streamer>'
```

To compare with working example:
1. Run `start mpv://https://www.twitch.tv/<your_favorite_streamer>`
```
[   0.000][v][cplayer] Command line options: '--player-operation-mode=pseudo-gui' '--' 'mpv://https://www.twitch.tv/<your_favorite_streamer>'
```

Note, if you don't have mpv registered to handle `mpv://` do:
1. Download mpv from https://nightly.link/mpv-player/mpv/workflows/build/master
2. Double click `mpv-register.bat`

### OS and Chatterino Version

Chatterino Nightly 2.5.4 (commit bd21db12) built on 2025-12-07 with Qt 6.7.1, MSVC 194435221, Crashpad Running on Windows 11 Version 24H2, kernel: 10.0.26100

teknsl:
It is written to open as `mpv://https://` but this is not fully legal in `QUrl`, it will disable the second scheme.

<details open>
<summary><code>src/util/CustomPlayer.cpp</code></summary>

```cpp
    QDesktopServices::openUrl(
        QUrl{scheme % u"https://www.twitch.tv/" % channelName});
```

</details>

kasper93:
You can urlescape the payload url, to make sure it is valid. `QUrl::toPercentEncoding` will help you with that.

pajlada:
Fixed in #6725

Thanks for the report!

PR #6725, opened 2026-01-08
Added support for percent-encoded url for custom player

SonicSegFault:
This PR updates `openInCustomPlayer` to percent‑encode Twitch URLs before embedding them in a custom URI scheme.

Possible solution to Issue:#6635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 - Dev: Unwrapped `LimitedQueueSnapshot` to `std::vector`. (#6606)
 - Dev: Simplified uses of `getMessageSnapshot`. (#6607)
 - Dev: Disabled `llvm-prefer-static-over-anonymous-namespace` in clang-tidy. (#6610)
-- Dev: Added experimental spell checker support. (#6446, #6703)
+- Dev: Added experimental spell checker support. (#6446, #6703, #6722, #6730)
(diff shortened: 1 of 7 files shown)


src/util/CustomPlayer.cpp
```diff
@@ -25,8 +25,10 @@ void openInCustomPlayer(QStringView channelName)
         return;
     }
 
-    QDesktopServices::openUrl(
-        QUrl{scheme % u"https://www.twitch.tv/" % channelName});
+    auto twitchUrl = "https://www.twitch.tv/" % channelName;
+    auto encodedTwitchUrl = QUrl::toPercentEncoding(twitchUrl);
+    QDesktopServices::openUrl(QUrl(QStringLiteral("%1://%2").arg(
+        scheme, QString::fromUtf8(encodedTwitchUrl))));
```
pajlada: You should be able to retain most of the syntax from before


```suggestion
    auto encodedTwitchUrl =
        QUrl::toPercentEncoding(u"https://www.twitch.tv/" % channelName);
    QDesktopServices::openUrl(QUrl{scheme % encodedTwitchUrl});

```

SonicSegFault: I think explicit conversion of `QString::fromUtf8` is needed for the `encodedTwitchUrl`

pajlada:

PR #6730, opened 2026-01-09
feat: load local directories for spellchecking

Nerixyz:
Previously, users could only have one local dictionary (`index.{aff,dic}`). With this PR, local dictionaries are loaded the same way as system dictionaries. In fact, all are loaded together into a single list. 

We save a default dictionary. It's saved as a path. For local dictionaries, we save a relative path (to ensure portability), and for system dictionaries, we save the absolute path.

### DIFF
--- a/src/controllers/spellcheck/SpellChecker.hpp
+++ b/src/controllers/spellcheck/SpellChecker.hpp
@@ -21,7 +21,9 @@ struct DictionaryInfo {
 
     /// The path to the dictionary without the .aff or .dic suffix (e.g. "/foo/bar/en_GB" or "en_GB")
     ///
-    /// Paths are either absolute or relative to the `Dictionaries/` directory.
+    /// Paths must be absolute if they're marked as a system dictionary,
+    /// otherwise they must be relative to the Chatterino Dictionaries
+    /// directory.
     QString path;
 
     bool isSymbolicLink;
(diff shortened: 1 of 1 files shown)


src/controllers/spellcheck/SpellChecker.hpp
```diff
@@ -19,10 +19,13 @@ struct DictionaryInfo {
     /// The name of the dictionary to be shown to users (e.g. "en_GB (System)")
     QString name;
 
-    /// The absolute path to the dictionary without the .aff or .dic suffix (e.g. "/foo/bar/en_GB")
+    /// The path to the dictionary without the .aff or .dic suffix (e.g. "/foo/bar/en_GB" or "en_GB")
+    ///
+    /// Paths are either absolute or relative to the `Dictionaries/` directory.
```
pajlada:
```suggestion
    /// Paths must be absolute if they're marked as a system dictionary, otherwise they must be relative to the Chatterino Dictionaries directory.
```

src/controllers/spellcheck/SpellChecker.hpp
```diff
@@ -37,10 +40,7 @@ class SpellChecker
     bool check(const QString &word);
     std::vector<std::string> suggestions(const QString &word);
 
-    /// Return a list of system-installed dictionaries.
-    ///
-    /// Currently only implemented on Linux.
-    std::vector<DictionaryInfo> getSystemDictionaries() const;
+    std::vector<DictionaryInfo> getAvailableDictionaries() const;
```
pajlada, line 49: Keep the essence of the comment
```suggestion
    /// Return a list of dictionaries from the system and from the Chatterino Dictionaries directory.
    ///
    /// System-dictionary loading is currently only implemented on Linux.
    std::vector<DictionaryInfo> getAvailableDictionaries() const;
```

PR #6731, opened 2026-01-09
feat(spellchecking): ignore commands for spellchecking

4rneee:
Now spellchecking ignores command triggers at the start of the message.

I saw in the [wiki](https://wiki.chatterino.com/Commands/#advanced) that you can also match command triggers at the end but didn't find that in the settings or the code. Is this outdated?

### DIFF
--- a/src/controllers/commands/CommandController.hpp
+++ b/src/controllers/commands/CommandController.hpp
@@ -57,7 +57,9 @@ public:
 
     /// Returns the length of the command trigger in `text`, including leading
     /// spaces. If `text` does not start with a command trigger, 0 is returned.
-    /// Example: " /ban forsen" returns 5
+    /// Examples:
+    ///  - " /ban forsen" returns 5
+    ///  - "/non-existing-command" returns 0
     qsizetype commandTriggerLen(QStringView text);
 
 private:


pajlada:
> I saw in the [wiki](https://wiki.chatterino.com/Commands/#advanced) that you can also match command triggers at the end but didn't find that in the settings or the code. Is this outdated?

Yeah, no need to worry about that

src/widgets/splits/InputHighlighter.cpp
```diff
@@ -101,15 +102,21 @@ void InputHighlighter::highlightBlock(const QString &text)
 #else
     auto it = this->wordRegex.globalMatch(textView);
 #endif
+    bool isFirst = true;
+    bool ignoreFirst = getApp()->getCommands()->isCommand(
+        text.section(' ', 0, 0, QString::SectionSkipEmpty));
```
Nerixyz: You could advance the iterator immediately here:

```cpp
if (isCommand() && it.hasNext())
{
    it.next();
}
```

src/controllers/commands/CommandController.cpp
```diff
@@ -733,4 +733,19 @@ QStringList CommandController::getDefaultChatterinoCommandList()
...
+    if (this->commands_.contains(word) || this->userCommands_.contains(word))
+    {
+        return true;
+    }
+#ifdef CHATTERINO_HAVE_PLUGINS
+    if (this->pluginCommands_.contains(word))
+    {
+        return true;
+    }
+#endif
```
pajlada: The command controller handles this already, so you don't have to check this.

src/widgets/splits/InputHighlighter.cpp
4rneee: Good idea, that's much cleaner.

4rneee:
I just applied Nerix' suggestion and fixed an issue where I skipped the first element of the iterator despite the command not being in it because it didn't match the `wordRegex`. An example where this happened was the command `//`.

src/widgets/splits/InputHighlighter.cpp
```diff
@@ -101,6 +102,13 @@ void InputHighlighter::highlightBlock(const QString &text)
 #else
     auto it = this->wordRegex.globalMatch(textView);
 #endif
+    auto firstWord = text.section(' ', 0, 0, QString::SectionSkipEmpty);
+    if (it.hasNext() && getApp()->getCommands()->isCommand(firstWord) &&
+        this->wordRegex.match(firstWord).hasMatch())
+    {
+        it.next();
+    }
```
pajlada: Use two if-statements instead, first one checking if `(it.hasNext())` is true.
Inside there, get firstWord & use your second if-statement to perform the command checks.

I understand why you need to check `isCommand` on the word, what purpose does the word regex match serve?

src/controllers/commands/CommandController.hpp
```diff
@@ -55,6 +55,11 @@ class CommandController final
     }
 #endif
 
+    /// Returns the length of the command trigger in `text`, including leading
+    /// spaces. If `text` does not start with a command trigger, 0 is returned.
+    /// Example: " /ban forsen" returns 5
```
pajlada: Would be nice if this includes an example that clarifies that non-existing commands will also return 0

src/controllers/commands/CommandController.cpp
```diff
@@ -733,4 +733,36 @@ QStringList CommandController::getDefaultChatterinoCommandList()
...
+        triggerLen += words[i].length();
+
+        if (this->commands_.contains(commandName) ||
+            this->userCommands_.contains(commandName))
+        {
+            return triggerLen;
+        }
+
+        // ignore consecutive spaces
+        if (!words[i].isEmpty())
```
Nerixyz, line 756: You could use `Qt::SkipEmptyParts` when splitting, then you won't have to account for empty strings.

src/controllers/commands/CommandController.cpp
```diff
@@ -733,4 +733,36 @@ QStringList CommandController::getDefaultChatterinoCommandList()
     return this->defaultChatterinoCommandAutoCompletions_;
 }
 
+qsizetype CommandController::commandTriggerLen(QStringView text)
+{
+    auto words = text.split(' ');
+
+    qsizetype triggerLen = 0;
+    qsizetype spaces = 0;
+    QString commandName{};
```
Nerixyz, line 742:
```suggestion
    QString commandName;
```

src/controllers/commands/CommandController.cpp
4rneee: I don't think I can, because I need to know the amount of empty parts that are being skipped to handle for example leading spaces. Otherwise "`/ban`" and "` /ban`" would both return 4

Issue #6734, opened 2026-01-11
Update account entry when username changes


pajlada:
Follow-up to https://github.com/Chatterino/chatterino2/pull/6723

As of https://github.com/Chatterino/chatterino2/pull/6723 - your username in the client is now updated on startup together with the scope check.
However, it does not update your account entry in `settings.json`.

Updating this might be a bit wacky since the way we do settings with regards to accounts is weird.
If you decide to take this on and run into any issues, feel free to ask here.

SonicSegFault:
Just to confirm my understanding:

we want to trigger `delayedItemsChanged` and update the corresponding `ChatterinoSetting`, we change `accounts_` signal vector for that. (not directly)

We don’t want to call `requestSave()` manually here, and instead rely on the existing save flow to persist the change to`settings.json`.

pajlada:
Give me a moment to ponder this and I'll get back to you

Nerixyz:
> we want to trigger `delayedItemsChanged` and update the corresponding `ChatterinoSetting`, we change `accounts_` signal vector for that. (not directly)

I'm not sure what you mean by this. We should avoid recreating the account, though, because it stores state and it would mean that the current account might need to reconnect.

> We don’t want to call `requestSave()` manually here

In my opinion, calling `requestSave()` would be fine. Moreover, something like the following from the login dialog would be fine for now:
https://github.com/Chatterino/chatterino2/blob/31ef6eeddf5db154882543a4a8a1533e620951f2/src/widgets/dialogs/LoginDialog.cpp#L65-L66

In the future, we should use a struct for the data we save (as I explained in https://github.com/Chatterino/chatterino2/pull/6723#issuecomment-3725678411).

> and instead rely on the existing save flow to persist the change to`settings.json`.

For accounts, we don't have a save flow other than the one in `LoginDialog`.

SonicSegFault:
Guess I might have been overcomplicating it, 

> I'm not sure what you mean by this. We should avoid recreating the account, though, because it stores state and it would mean that the current account might need to reconnect.

I wasn’t suggesting recreating the account object, my assumption was that mutating the `SignalVector` might trigger save via `delayedItemsChanged`

It seems that was not the case. As 
> For accounts, we don't have a save flow other than the one in `LoginDialog`.

I mean yeah, we could create a flow struct, but that might be overkill for this.

pajlada:
[...]

PR #6739, opened 2026-01-12
fix: update currently-selected username on username update

SonicSegFault:
This PR follows up on #6734.

While the username is now updated at runtime after OAuth validation, the updated login was not being persisted to settings.json. 

This change updates the setting's Twitch account list when a username change is detected, ensuring the updated login is persisted and remains consistent across restarts.


SonicSegFault:
Right now I’ve put the /accounts/current sync inline in `checkMissingScopes.`

Would it make sense to move this into `TwitchAccountManager` (e.g. `syncCurrentAccountPointer()`), just a feeling we might need it again. 
I'd want to avoid direct settings calls here and keep the save logic centralized. 
Happy to keep it inline if you prefer to keep the scope tight.

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -75,6 +76,17 @@ void checkMissingScopes(const std::shared_ptr<TwitchAccount> &account)
                     "/accounts/uid" + account->getUserId().toStdString();
                 pajlada::Settings::Setting<QString>::set(basePath + "/username",
                                                          login);
+                auto currentUsername =
+                    app->getAccounts()->twitch.getCurrent()->getUserName();
+                auto storedCurrent = pajlada::Settings::Setting<QString>::get(
+                    "/accounts/current");
```
pajlada: Instead of directly calling get and set on `/accounts/current`, utilize the `currentUsername` member in the `TwitchAccountManager` class

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -75,6 +76,17 @@ void checkMissingScopes(const std::shared_ptr<TwitchAccount> &account)
...
+                if (currentUsername.compare(storedCurrent,
+                                            Qt::CaseInsensitive) != 0)
+                {
+                    pajlada::Settings::Setting<QString>::set(
+                        "/accounts/current", currentUsername);
+                    getSettings()->requestSave();
```
pajlada: You will want to call requestSave either way since you've at least updated the account's username

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -71,6 +72,19 @@ void checkMissingScopes(const std::shared_ptr<TwitchAccount> &account)
...
+                    "/accounts/uid" + account->getUserId().toStdString();
+                pajlada::Settings::Setting<QString>::set(basePath + "/username",
+                                                         login);
+                auto &manager = app->getAccounts()->twitch;
+                auto currentUsername = manager.getCurrent()->getUserName();
+                if (currentUsername.compare(manager.currentUsername.getValue(),
+                                            Qt::CaseInsensitive) != 0)
+                {
+                    pajlada::Settings::Setting<QString>::set(
+                        "/accounts/current", currentUsername);
```
pajlada: You can use `manager.currentUsername` here too

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 - Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6710)
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
+- Bugfix: Defensively update current Twitch account username in checkMissingScopes. (#6734)
```
pajlada: You can add this PR number (so 6739, not 6734) to the `Bugfix: Fix highlight mentions not updating when username changes.` entry

pajlada (changes requested):

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 - Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6710)
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
+- Bugfix: Fix highlight mentions not updating when username changes. (#6739)
```
pajlada:
```suggestion
- Bugfix: Fix highlight mentions not updating when username changes. (#6723, #6739)
- Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
```

pajlada (approved):

Wissididom (approved):

PR #6743, opened 2026-01-15
feat: add search engine selection and context menu search functionality

Leafyzito:
Adds a "Search in {Engine}" option to the message context menu when text is selected, allowing users to search selected text in their preferred search engine.


pajlada:
I'm not sure this is functionality I want to add, at least not in this way.

For searching a full message's context, you can create a command with "Show in message menu" enabled like this: `/openurl https://duckduckgo.com/?q={1+}`

I am interested in making more things available with the command syntax, e.g. only the selected text, or the emote name/emote url you right-clicked.

Nerixyz:
> For searching a full message's context, you can create a command with "Show in message menu" enabled like this: `/openurl https://duckduckgo.com/?q={1+}`

You can also do more complex processing with plugins (e.g. URL encoding the query).

Leafyzito:
I just thought that this is such a nice, useful and simple feature that most people would like having by default.

8thony:
- There was a discussion about this feature in #2058 a while back.

You find some other perspectives and maybe some ideas to improve the functionality.

pajlada:
I'm not entirely opposed to this feature, just the current implementation. I would be more OK with this being a feature if:
1) it's not enabled by default
2) it requires users to provide their own search engine parameters (e.g. `https://duckduckgo.com/?q=` maybe with an optional name?), with us providing very few presets in the form of buttons or something in the setting page, and more examples on our wiki page.

not 100% sure, can keep discussing here if you have any thoughts

Leafyzito:
Right now, the dropdown sits under Miscellaneous with presets (Google, Bing, DuckDuckGo, Qwant, Startpage, Yahoo, Yandex, Ecosia, Baidu, Ask, AOL), and with Google selected by default.

If we disable the feature by default and keep those presets available in settings (not auto-selected), would that match what you had in mind?

pajlada:
I don't like us providing such a big list, even at this size it can't really be complete.
My personal preference is to not provide _any_ presets, but I understand that's not very user-friendly so my compromise is to only provide very few presets (DuckDuckGo, Bing, Google in this order) + a way to use a custom search engine instead.

Leafyzito:
This is my first contribution to a larger C++ project, so the implementation may not be the most optimized. It achieves the intended goal and follows the project’s patterns. Feedback is welcome.

pajlada:
No need to immediately act on the github bot comments you're receiving, they can be handled easily later

Leafyzito:
That should fix the problems with the checks. About the bot's comments, I'll handle them later tonight

Leafyzito:
I’ve updated the implementation to align better how you envisioned it. Please let me know what you think of it.

Nerixyz:
> I’ve updated the implementation to align better how you envisioned it. Please let me know what you think of it.

You don't need the model and the dialog anymore.

src/singletons/Settings.hpp
```diff
@@ -309,6 +309,10 @@ class Settings
     BoolSetting autoCloseUserPopup = {"/behaviour/autoCloseUserPopup", true};
     BoolSetting autoCloseThreadPopup = {"/behaviour/autoCloseThreadPopup",
                                         false};
+    BoolSetting searchEngineEnabled = {"/behaviour/searchEngineEnabled", false};
```
pajlada: Should be pathed at `/behaviour/searchEngine/enabled`

src/singletons/Settings.hpp
```diff
@@ -309,6 +309,10 @@ class Settings
...
+    QStringSetting searchEngine = {"/behaviour/searchEngine", ""};
```
pajlada: Remove this - it's unused

src/widgets/helper/ChannelView.cpp
```diff
@@ -2619,6 +2619,25 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
...
+        // Add search action when text is selected and search feature is enabled
+        if (getSettings()->searchEngineEnabled.getValue())
+        {
+            QString searchURL = getSettings()->searchEngineUrl.getValue();
+            QString searchName = getSettings()->searchEngineName.getValue();
+
+            if (!searchURL.isEmpty())
+            {
+                QString displayName =
+                    searchName.isEmpty() ? "Search" : searchName;
```
pajlada: If no search name is set, I think the action should just be "Search" instead of "Search with Search"

src/widgets/helper/ChannelView.cpp
```diff
@@ -2619,6 +2619,25 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
...
+
+                menu->addAction(
+                    "&Search with " + displayName, [this, searchURL] {
```
pajlada: Not a request for change: this adds "s" as a hotkey. This is consistent with Firefox. Is this OK?

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1214,6 +1214,98 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
         ->addTo(layout);
 
+    {
+        layout.addSubtitle("Search Engine");
+        SettingWidget::checkbox("Enable search in right-click context menu",
+                                s.searchEngineEnabled)
+            ->setTooltip(
+                "Allow searching selected text using a search engine from "
+                "the right-click context menu.")
+            ->addTo(layout);
```
pajlada, line 1230: Move the checkbox below the description

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1214,6 +1214,98 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+
+        layout.addDescription(
+            "Search engine which appears when you select text and "
+            "right-click a message. Select a search engine preset from the "
+            "dropdown below, or fill in your custom search engine URL and "
+            "name.");
+
+        // Preset dropdown
+        QStringList presetList = {"DuckDuckGo", "Bing", "Google", "Custom"};
```
pajlada: Remove "Custom" from the preset list and instead set a placeholder text for the combo box (e.g. "Select..." or something, not sure)

This placeholder text is shown whenever the current index is invalid (e.g. -1), so also set the current index to -1, and reset the current index to -1 an ytime a user has selected one of the valid presets. This allows the dropdown to actually act as an "action" dropdown rather than something the user selects.

With this, you can remove all the "dropdown based on current settings" initialization code you have

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1214,6 +1214,81 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                    s.searchEngineUrl = "https://www.bing.com/search?q=";
+                    s.searchEngineName = "Bing";
+                }
+                else if (preset == "Google")
+                {
+                    s.searchEngineUrl = "https://www.google.com/search?q=";
+                    s.searchEngineName = "Google";
+                }
+                // Reset to -1 after selection
+                presetCombo->blockSignals(true);
```
Nerixyz: Use [`QSignalBlocker`](https://doc.qt.io/qt-6/qsignalblocker.html) here.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2619,6 +2619,25 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
...
+
+            if (!searchURL.isEmpty())
+            {
+                QString actionText = searchName.isEmpty()
+                                         ? "&Search"
+                                         : "&Search with " + searchName;
+
+                menu->addAction(actionText, [this, searchURL] {
+                    QDesktopServices::openUrl(QUrl(
+                        searchURL + this->getSelectedText().trimmed()));
```
Nerixyz: Should the query get URL encoded?

Leafyzito: Is it needed? for what i've tested, it works fine as it is

Nerixyz: I think we need to - if you select text that contains ampersands, then stuff after them won't be added to the query - something like `foo & bar` will only search for `foo`.

Leafyzito: Yeah you're right, I'll implement it

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1214,6 +1215,82 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        // Preset dropdown
+        QStringList presetList = {"DuckDuckGo", "Bing", "Google"};
+        auto *presetCombo =
+            layout.addDropdown("Search Engine Preset", presetList,
+                               "Select a search engine preset");
+        presetCombo->setEditable(true);
+        presetCombo->lineEdit()->setPlaceholderText("Select...");
+        presetCombo->lineEdit()->setReadOnly(true);
+        presetCombo->setCurrentIndex(-1);
+        presetCombo->setEnabled(s.searchEngineEnabled.getValue());
```
pajlada: mini nit: this initial `setEnabled` call is not needed, as the `connect` call to a setting will be run once upon connection

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1214,6 +1215,82 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                "Allow searching selected text using a search engine from "
+                "the right-click context menu.")
+            ->addTo(layout);
+
+        // Preset dropdown
+        QStringList presetList = {"DuckDuckGo", "Bing", "Google"};
+        auto *presetCombo =
+            layout.addDropdown("Search Engine Preset", presetList,
+                               "Select a search engine preset");
+        presetCombo->setEditable(true);
```
pajlada: iirc you shouldn't need to make this QComboBox editable. Instead use `presetCombo->setPlaceHolderText(...)`, and remove all other lineEdit-related calls and things should still work as expected

let me know if this doesn't work for some reason and i can give it another look

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Add search engine selection for context menu search action.
```
pajlada: Add your PR number to the end of this changelog entry, similar to what you see on the changelog entries below yours.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2619,6 +2620,26 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
         menu->addAction("&Copy selection", [this] {
             crossPlatformCopy(this->getSelectedText());
         });
+
+        // Add search action when text is selected and search feature is enabled
+        if (getSettings()->searchEngineEnabled.getValue())
```
pajlada: Positioning in the menu: Right now this is under Copy selection and above Copy message - this feels wrong to me.
I think this should be below the reply/thread-related options, but I'm happy to discuss this.

Leafyzito:
Thanks for the feedback and guidance. I believe the latest changes align with your suggestions. Happy to adjust further if needed.

pajlada:

PR #6746, opened 2026-01-17
fix(http): Send multipart body for `PUT`

Nerixyz:
Not all request types had the same body handling - `PUT` requests couldn't send a multipart body. With this PR, they have the same handling.

I also added tests for the body in the request types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Minor: Added a setting to show the stream title in live messages. (#6572)
 - Minor: Add categories to the emoji viewer. (#6598)
 - Minor: Added broadcaster-only `/poll`, `/cancelpoll`, and `/endpoll` commands. (#6583, #6605)
-- Minor: Added broadcaster-only `/prediction`, `/cancelprediction`, `/lockprediction`, and `/completeprediction` commands. (#6583, #6612, #6632)
+- Minor: Added broadcaster-only `/prediction`, `/cancelprediction`, `/lockprediction`, and `/completeprediction` commands. (#6583, #6612, #6632, #6749)
 - Minor: Added support for BetterTTV Pro subscriber badges. (#6625, #6724)
 - Minor: Added `debug.traceback` for plugins. (#6652)
 - Minor: Added title and duration options for `/clip` command. (#6669)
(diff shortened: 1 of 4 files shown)


Nerixyz:
Sure we should still send the correct body in all requests, though.

pajlada:
> Sure we should still send the correct body in all requests, though.

Agreed, am I right in understanding that it's specifically PUT multipart requests this now aims to fix?

Nerixyz:
> Agreed, am I right in understanding that it's specifically PUT multipart requests this now aims to fix?

Yes.

Nerixyz:
Not sure, but the httpbox instance on https://braize.pajlada.com/httpbox might be using a different version than the one we use in tests. The tests passed for that public URL but fail in CI (fix was to set the content type).

pajlada:
I don't remember ever updating the public version, i'll take a peek at that tomorrow

Nerixyz:
> I don't remember ever updating the public version, i'll take a peek at that tomorrow

Or maybe the requests go through some proxy that sets the `content-type` header.

pajlada:
public httpbox is running through cloudflare and nginx, so some headers are being added (`cf-connecting-ip`, `x-real-ip`, `x-forwarded-proto` etc), but I don't see `content-type` when I GET the `httpbox/headers` param

remotely:
```
$ curl --silent https://braize.pajlada.com/httpbox/headers | cut -d':' -f1 
host
x-real-ip
x-forwarded-for
x-forwarded-proto
user-agent
cf-ray
accept
cdn-loop
cf-connecting-ip
cf-ipcountry
cf-visitor
accept-encoding
```

locally directly connecting to httpbox:
```
$ curl --silent localhost:9051/headers | cut -d':' -f1 
host
user-agent
accept
```

haven't tried sending the request from qt yet

pajlada:
I have updated braize to match local-running now, headers look the same to me from curl

CHANGELOG.md
```diff
@@ -89,6 +89,7 @@
 - Dev: Refactored types used for emoji parsing. (#6714)
 - Dev: `LinkParser` now parses `QStringView`s. (#6715)
 - Dev: Cleaned up `AccountController` connections. (#6735)
+- Dev: Unified body handling for HTTP requests. (#6746)
```
pajlada: Make the changelog entry match current pr title, i.e. something like Fix: properly send multipart body for PUT

pajlada (changes requested):

PR #6750, opened 2026-01-18
feat: add /monitor and /restrict commands for moderators

### LINKED ISSUE #6747: Add support for modification of low trust users
### Checklist

### Describe your issue

Obviously requires moderator status, needs new scope: chatterino/website#913

https://dev.twitch.tv/docs/api/reference#add-suspicious-status-to-chat-user
https://dev.twitch.tv/docs/api/reference#remove-suspicious-status-from-chat-user

@iProdigy, I think you wanted to do this?

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.4 DEBUG (commit a8ac28425 modified) built with Qt 6.10.1 Running on Arch Linux, kernel: 6.17.9-arch1-1

iProdigy:
Closes #6747

Add `/(un)monitor`, `/(un)restrict` commands

image

https://dev.twitch.tv/docs/api/reference/#add-suspicious-status-to-chat-user
https://dev.twitch.tv/docs/api/reference/#remove-suspicious-status-from-chat-user

No longer broken: https://github.com/twitchdev/issues/issues/1141

note: the api doesn't make a distinction between `/unmonitor` and `/unrestrict` so doing `/unmonitor` on a restricted user still "works" (makes the user no longer restricted)


src/controllers/commands/builtin/twitch/LowTrust.cpp
```diff
@@ -0,0 +1,171 @@
...
+    {
+        ctx.channel->addSystemMessage("You must be logged in to " % command %
+                                      " someone!");
+        return;
+    }
+
+    auto modId = currentUser->getUserId();
+    getHelix()->getUserByName(
+        ctx.words.at(1),
+        [chan = ctx.twitchChannel, modId, command, restrict](const auto &user) {
```
Nerixyz: The channel could get destroyed before the response arrives (same for the other callbacks).

iProdigy:
Endpoint was fixed by Twitch and the commands appear to work as expected

src/controllers/commands/builtin/twitch/LowTrust.cpp
pajlada: Capture `ctx.channel` instead of `ctx.twitchChannel` in `addSuspiciousTreatment` too, and call any functions like roomId before both helix calls

pajlada:
Just https://github.com/Chatterino/chatterino2/pull/6750#discussion_r2725341751 - otherwise looks good. Haven't tested yet

src/controllers/commands/builtin/twitch/LowTrust.cpp
iProdigy: thanks, I forgot I had two helper funcs to fix

PR #6752, opened 2026-01-19
feat: allow updating from nightly to stable

### LINKED ISSUE #6608: Allow updating from nightly to stable
If a user is on a nightly version, they will never be notified of newer version - nightly or stable. Nightly versions *must* be updates manually, which isn't something most users want to do.

To prevent users from being "stuck" on old nightly versions, Chatterino should allow updating when a newer stable version is available.

Spawned from #6599.

jupjohn:
This PR aims to remove the nightly guard from the update checker. This should allow nightly versions that are on a lower semantic version than stable/beta to notify of updates in their respective channels.

I've tested this by lowering the `CHATTERINO_VERSION` to `2.5.3` on this branch, and the update button shows as expected. When the value is set to `2.5.4`, no update button is shown.

Fixes #6608.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,10 @@
 - Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6710)
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723, #6739)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
+- Bugfix: Fixed incorrect deletion length in tab autocomplete when completion included emojis. (#6800)
+- Bugfix: Fixed WebSockets for 7TV or BTTV not connecting on Windows if IPv6 attempts fail. (#6805)
 - Bugfix: Fixed portable updates not showing an error if the updater is not present. (#6801)
+- Bugfix: Fixed liveupdates connection debug count becoming negative. (#6806)
 - Dev: Nightly builds are now defined through a build flag rather than the Modes file. (#6798)
 - Dev: Update release documentation. (#6498)
 - Dev: Make code sanitizers opt in with the `CHATTERINO_SANITIZER_SUPPORT` CMake option. After that's enabled, use the `SANITIZE_*` flag to enable individual sanitizers. (#6493)
@@ -112,6 +115,9 @@
 - Dev: Updated `pajlada-settings` to v0.5.0. (#6797)
 - Dev: Updated `pajlada-serialize` to v0.2.1. (#6797)
 - Dev: Updated `pajlada-signals` to v0.1.1. (#6797)
+- Dev: Balance IPv4 and IPv6 connection attempts. (#6804)
+- Dev: Factored out recent messages benchmark helper. (#6815)
+- Dev: Added `CHATTERINO_FORCE_LTO` CMake option to skip LTO check. (#6816)
 
 ## 2.5.4
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ cmake_policy(SET CMP0087 NEW) # evaluates generator expressions in `install(CODE
 cmake_policy(SET CMP0091 NEW) # select MSVC runtime library through `CMAKE_MSVC_RUNTIME_LIBRARY`
 include(FeatureSummary)
 include(FetchContent)
+include(CMakeDependentOption)
 
 list(APPEND CMAKE_MODULE_PATH
     "${CMAKE_SOURCE_DIR}/cmake"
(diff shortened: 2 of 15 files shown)


Mm2PL:
This used to suggest downgrades from `2.x.y+some-nightly-commits` to `2.x.y`. Is that still the case?

jupjohn:
> This used to suggest downgrades from `2.x.y+some-nightly-commits` to `2.x.y`. Is that still the case?

Are there builds using that naming scheme? AFAIK nightly builds are purely `2.*.*` with no suffix, and nightly builds use the current version instead of the next releasable version.

pajlada:
There are five self-installed (i.e. not managed by a package manager) types:
1.  Stable installer, installed from https://chatterino.com/#downloads
2.  Stable Portable, installed from https://chatterino.com/#downloads
3.  Nightly, installed from https://github.com/Chatterino/chatterino2/releases/tag/nightly-build 
chatterino-windows-x86-64-Qt-6.7.1.zip 
4.  Nightly Portable, installed like above but with custom modification to the modes file
5.  And, not one we really support or advertise, Nightly installer

1 & 2 are solved, and not something this PR aims to fix.

For 3, this would provide a confusing experience. The User has downloaded a zip file and extracted it, presumably somewhere like their Documents folder, Downloads folder, or their desktop. When they get prompted that there is an update available, it will download the updated installer which will install/update Chatterino that is installed under `C:\Program Files\` rather than wherever you're running from, so the next time the user goes to run Chatterino from however they ran it, there's a big chance it will run the old version and prompt for the same update yet again.

For 4, this can work if we provide a proper portable zip in the nightly release with the modes file already set, and the updater helper.

For 5, this PR would work correctly, since the Nightly installer installs in the same location as the Stable installer.

**The options I see are:**
Since this change only applies going forward, we could remove 3 as a supported method, and instead provide new support for 5 and better support for 4.

or

Provide different text-only instructions for Nightly builds, instructing users where to go to download the build themselves.

Thoughts?

sidenotes:
I noticed portable installations that lack `updater.1/ChatterinoUpdater.exe` just silently fail.

If a user has installed Chatterino at a non-standard path with the installer (e.g. `D:\MyApps`), do they have to update the path in the updater whenever they run the update? or do we infer that somehow

8thony:
> Thoughts?

All my thoughts below:

> so the next time the user goes to run Chatterino from however they ran it, there's a big chance it will run the old version

My idea is to make the icon of Chatterino Nightly different.
The best example would be Chrome or Firefox nightly or beta versions. They have a different color or text with "Dev" to make clear what version it is.

> If a user has installed Chatterino at a non-standard path with the installer (e.g. `D:\MyApps`), do they have to update the path in the updater whenever they run the update? or do we infer that somehow

No the installer checks the Programm Uninstall Key for Chatterino and installs the update in the same folder automatically. 
<details>
<summary>Click to Expand Regedit Uninstall Keys</summary>

For systemwide installs:
```
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{F5FE6614-04D4-4D32-8600-0ABA0AC113A4}_is1
```
For current User installs:
```
Computer\HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{F5FE6614-04D4-4D32-8600-0ABA0AC113A4}_is1
```
</details>


> For 5, this PR would work correctly, since the Nightly installer installs in the same location as the Stable installer.

I don't like this approach because then you can't test/install both versions side by side. 
For the nightly installer in general, I suggest changing the default install path to 
"C:\Program Files\Chatterino Nightly". 
This also needs a different AppId [here](https://github.com/Chatterino/chatterino2/blob/43ac9c38db3a1a8e1a6c882a7fa7967f25e0ec60/.CI/chatterino-installer.iss#L28C1-L28C6), for the Programm Uninstall registry entry so it does not overrite the stable entry and treats the nightly installation like a new program. Something like "AppId=chatterino_nightly" would be enough.

> I noticed portable installations that lack `updater.1/ChatterinoUpdater.exe` just silently fail.

We should build and bundle the Updater with CI as @Nerixyz explained and done [here](https://redirect.github.com/SevenTV/chatterino7/pull/228)
This makes sure we can't ship without an updater and makes it easier to get the release artifacts. 

For me personally, I always stay on nightly, and it would be good to make an option to disable update checks and make nightly updates possible.

Nerixyz:
> Since this change only applies going forward, we could remove 3 as a supported method, and instead provide new support for 5 and better support for 4.

-1, because users have learned to extract the nightly files themselves. **Forcing** them to adjust their workflow is not worth it. If we wanted to move towards only supporting 4 and 5 (using the `Chatterino.Nightly` files), then we should do that in steps without removing support for 3 right away.

Though I'm also not happy with easing installation of nightly versions. Not because "users should do it the hard way", but because I feel like this is a band-aid solution for Chatterino not having a frequent update channel (https://github.com/Chatterino/chatterino2/discussions/5773). As hinted at in the discussion, nightly is too fast and requires manual updating and stable/beta are too slow.

> Provide different text-only instructions for Nightly builds, instructing users where to go to download the build themselves.

This feels like the easiest solution. I'm not sure where we should link to:

1. Link to the installer (https://chatterino.com/download/Chatterino%20Installer.exe). This would be bad for portable installs or other manually extracted installations.
2. Link to the exact GitHub release (e.g. https://github.com/Chatterino/chatterino2/releases/tag/v2.5.4). Would be guesswork based on the `version` from the update response. The response could also include a link to the GitHub release.
3. Link to the GitHub releases page (https://github.com/Chatterino/chatterino2/releases). This will show nightly at the top.
4. Link to the Chatterino website (https://chatterino.com/#downloads).

> For me personally, I always stay on nightly, and it would be good to make an option to disable update checks and make nightly updates possible.

If you build it yourself, there's the `CHATTERINO_UPDATER` CMake option. But having a setting for this seems reasonable. 

> I noticed portable installations that lack `updater.1/ChatterinoUpdater.exe` just silently fail.

I'll add a check.

pajlada:
The flow from Nightly to Stable is not super streamlined now because it would be difficult to take into consideration everyone's installations, but with this change now we still run the updater to check for new versions, and if a new version is detected, we let the user know & point them to the download page

image

Do these changes make sense to you @jupjohn ?

pajlada (approved):

PR #6753, opened 2026-01-19
Refactor DebugCount

Mm2PL:
Switching to enums gives us a nice way to implement a base template that increments and decrements the DebugCount on (de)construction.

~~This fixes issues with Lua HTTP objects showing negative amounts, I don't believe we had an issue for that.~~ I undid the AutoDebugCount changes.

### DIFF
--- a/src/providers/liveupdates/BasicPubSubManager.hpp
+++ b/src/providers/liveupdates/BasicPubSubManager.hpp
@@ -228,7 +228,7 @@ private:
             qCWarning(chatterinoLiveupdates)
                 << "Retrying after" << id << "failed";
             auto nSubs = subs.size();
-            DebugCount::increase("LiveUpdates subscription backlog",
+            DebugCount::increase(DebugObject::LiveUpdatesSubscriptionBacklog,
                                  static_cast<int64_t>(nSubs));
             this->pendingSubscriptions_.insert(
                 this->pendingSubscriptions_.end(),


src/util/DebugCount.cpp
```diff
@@ -21,70 +23,63 @@ struct Count {
 };
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-UniqueAccess<std::map<QString, Count>> COUNTS;
+UniqueAccess<std::array<Count, static_cast<size_t>(DebugObject::Count)>> COUNTS;
+
+constexpr bool isBytes(DebugObject target)
```
Nerixyz, line 27: This doesn't handle all cases, which is why the tests are currently failing.

Mm2PL: It now is used and uses default case

src/util/DebugCount.hpp
```diff
@@ -6,37 +6,157 @@
...
+        DebugCount::decrease(target, 1);
+    }
 
-    static void configure(const QString &name, Flags flags);
+    static QString getDebugText();
+};
 
-    static void set(const QString &name, const int64_t &amount);
+template <DebugObject target>
+class AutoDebugCount
```
pajlada: Could we hold off on the `AutoDebugCount` implementation for a follow-up PR?

src/util/DebugCount.cpp
```diff
@@ -5,86 +5,65 @@
...
 using namespace chatterino;
 
 struct Count {
     int64_t value = 0;
-    DebugCount::Flags flags = DebugCount::Flag::None;
 };
 
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-UniqueAccess<std::map<QString, Count>> COUNTS;
+UniqueAccess<std::array<Count, static_cast<size_t>(DebugObject::Count)>> COUNTS;
```
pajlada, line 25: Can make use of magic_enum's enum_size function here, which means we don't have to keep `Count`

src/util/DebugCount.hpp
```diff
@@ -6,37 +6,157 @@
...
+    {
+        DebugCount::increase(target);
+    }
+
+    AutoDebugCount &operator=(const AutoDebugCount & /*unused*/) = default;
+    AutoDebugCount &operator=(AutoDebugCount &&) = default;
 };
 
 }  // namespace chatterino
+template <>
```
pajlada, line 80: new line
```suggestion

template <>
```

Mm2PL:
magic_enum didn't really like me when I tried that and produced 
megabytes of error messages. I CBA to debug why that pile of templates 
dislikes me today.

W dniu 24.01.2026 o 13:00, pajlada pisze:
> Can make use of magic_enum's enum_size function here, which means we don't have to keep `Count`

src/util/DebugCount.hpp
```diff
@@ -6,37 +6,130 @@
...
-
-    static void configure(const QString &name, Flags flags);
+    static void set(DebugObject target, int64_t amount);
 
-    static void set(const QString &name, const int64_t &amount);
-
-    static void increase(const QString &name, const int64_t &amount);
-    static void increase(const QString &name)
+    static void increase(DebugObject target, int64_t amount);
+    static void increase(DebugObject target)
```
Nerixyz, line 64: The `amount` could default to 1, then you won't need an additional overload. Same for the `decrease` method.

PR #6770, opened 2026-01-25
feat: add setting to search in incognito mode

4rneee:
Alternatively we could use the existing `Open links in incognito/private mode` setting.


src/widgets/helper/ChannelView.cpp
pajlada, line 2723: Could you update the action text too? "...in private mode" or something

4rneee: Sure, though this makes the menu quite large.

PR #6779, opened 2026-01-27
chore: add tests for input highlighting

Nerixyz:
Given that we had some issues with words not being separated correctly or parts not getting ignored when spell checking, I thought it would be good if we had some tests. With them, we can avoid regressions.


tests/src/InputHighlighter.cpp
```diff
@@ -0,0 +1,271 @@
...
+            .input = "/ban a user",
+            .words = {"a", "user"},
+        },
+        {
+            .input = "/my-command some text",
+            .words = {"some", "text"},
+        },
+        {
+            .input = "/my-command with spaces and some text",
+            // FIXME: 'with spaces' shouldn't show up
```
4rneee: I think this is expected because `/my-command` will be triggered instead of `/my-command with spaces`. If you change the command trigger to `/my command with spaces` 'with spaces' won't show up anymore.

Nerixyz: > `/my-command` will be triggered instead of `/my-command with spaces`

That's some weird behavior from the command controller. Commands with spaces are just weird, though.

PR #6781, opened 2026-01-29
feat(plugins): Add message signals

Nerixyz:
This PR adds `Channel:on_message{_appended, _replaced, s_cleared}`. They use the same functionality as the display name changed signal and forward the respective signals in `Channel`.

**They're all synchronous.** This makes the implementation fairly straight forward, but also a trivial stack overflow. Other than Chatterino crashing, this shouldn't have any consequences, because the OS will kill the process. Note that a plugin can already halt Chatterino by executing an infinite loop, because plugins run in the GUI thread.

We still have to be careful that we don't destroy the channel as part of a callback (or at least delay that). For now that's not an issue.


Mm2PL:
> **They're all synchronous.** This makes the implementation fairly straight forward, but also a trivial stack overflow. 

IMO the only viable solution (which keeps this sync) would be to limit 
callback depth. We have a ton of things that add messages (the plugin 
adding the message, another plugin adding another message, commands to 
name a few). We could stop calling plugins once they have been called 
into more than X times, where X is a sufficiently big number to not be 
hit in normal usage.

> Other than Chatterino crashing, this shouldn't have any consequences, because the OS will kill the process.

This is mostly a nice feature *I want*

docs/wip-plugins.md
```diff
@@ -422,6 +422,31 @@ channel:add_message(c2.Message.new({
...
+##### `Channel:on_message_appended(cb)`
+
+Callback when a message is added.
+This is called synchronously.
+
+`cb` receives:
+
+- `msg` ([`c2.Message`](#message))
+- `override_flags` ([`c2.MessageFlag`](#message) or `nil`)
+
```
Mm2PL, line 459: This should probably mention that other even own plugin actions also invoke callbacks.

CHANGELOG.md
```diff
@@ -27,7 +27,7 @@
 - Minor: Added Markdown support to user notes. (#6490)
 - Minor: Moderators and VIPs in shared chats now show their channel badges. (#6653)
 - Minor: Added message read/update methods to the `Channel` plugin API. (#6650)
-- Minor: Added a display name change event to `Channel` for plugins. (#6594)
+- Minor: Added a display name change and message events to `Channel` for plugins. (#6594, #6781)
```
pajlada: Undo changelog entry

pajlada (changes requested):
Not tested, but the implementation seems fine to me

pajlada (approved):

Mm2PL (commented):
pending comment

Mm2PL (approved):

PR #6782, opened 2026-01-30
Add a countdown for slow mode and timeouts 

achow101:
When a channel is in slow mode or a user is timed out, it is useful to know how long the user needs to wait before they are able to chat again. This does not prevent sending messages, it only counts down in a UI element to inform the user.

Showing the timer is controlled with a new setting that defaults to off.

There are a few situations I found where the timer is not accurate, and probably others that I didn't check for:
- If the user is untimed out, there does not appear to be a notification for that. The timer will reset once a message from the user is received back from the channel.
- If the channel is in slow mode and the message is held by automod, it's unclear when the timer begins. This PR starts the timer the next time a message from the user is received back from the channel.
- If the user becomes a mod or vip, there does not appear to be a notification for these events so the timer keeps going. Once a message is sent and the USERSTATE update is received, the timer will be updated.

Loosely based on/inspired by #3329

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Minor: Added a countdown timer to the input box to show when you can chat again during slow mode or after a timeout. This feature is disabled by default, and can be enabled with the "Show countdown on slow mode or when timed out" setting. (#6782)
 - Minor: Add search engine selection for context menu search action. (#6743, #6770)
 - Minor: Add a separate highlight option for watchstreak notifications. (#6571, #6581)
 - Minor: Badges now link to their home page like emotes in the context menu. (#6437)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added a countdown timer to the input box to show when you can chat again during slow mode or after a timeout. (#11804)
```
pajlada: number should be your PR #

src/widgets/splits/SplitInput.hpp
```diff
@@ -82,6 +82,8 @@ class SplitInput : public BaseWidget
      */
     void setInputText(const QString &newInputText);
 
+    void setSendWaitStatus(const QString &text) const;
```
pajlada, line 90: Could you add a comment here what the intended use for this is (i.e. for timeouts & slow mode)

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -979,6 +997,26 @@ void IrcMessageHandler::handleNoticeMessage(Communi::IrcNoticeMessage *message)
...
+        int seconds = remaining.toInt(&ok);
+        if (ok)
+        {
+            auto *tc = dynamic_cast<TwitchChannel *>(channel.get());
+            assert(tc != nullptr);
+            tc->setSendWait(seconds);
+        }
+    };
+
+    if (tags == "msg_slowmode")
```
pajlada, line 1041: This feels like a strange way to parse the value. Could you provide examples of these messages as comments and explain what you're doing?

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -979,6 +997,26 @@ void IrcMessageHandler::handleNoticeMessage(Communi::IrcNoticeMessage *message)
...
+
+    auto handleSendWait = [&channel](const QString &remaining) {
+        bool ok = false;
+        int seconds = remaining.toInt(&ok);
+        if (ok)
+        {
+            auto *tc = dynamic_cast<TwitchChannel *>(channel.get());
+            assert(tc != nullptr);
+            tc->setSendWait(seconds);
+        }
```
pajlada, line 1038: You're fine to assert here, but you should also actually check that tc isn't a nullptr before using it

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -345,6 +347,9 @@ class TwitchChannel final : public Channel, public ChannelChatters
     const QString &getDisplayName() const override;
     void updateDisplayName(const QString &displayName);
 
+    void syncSendWaitTimer();
```
pajlada, line 376: Add a comment to this function

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -617,6 +630,11 @@ void IrcMessageHandler::handleUserStateMessage(Communi::IrcMessage *message)
                 tc->setMod(modTag == "1");
             }
         }
+
+        if (tc->hasHighRateLimit())
```
pajlada, line 660: Comment here

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2341,4 +2357,41 @@ void TwitchChannel::listenSevenTVCosmetics() const
...
+    }
+    else
+    {
+        this->sendWaitUpdate.invoke(formatTime(remaining));
+    }
+}
+
+void TwitchChannel::setSendWait(int seconds)
+{
+    if (seconds <= 0)
```
pajlada, line 2379: Are we able to be a bit more defensive here, meaning we don't have to call syncSendWaitTimer if `setSendWait(0)` is called twice in a row?

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -511,6 +515,15 @@ void IrcMessageHandler::handleClearChatMessage(Communi::IrcMessage *message)
     else
     {
         chan->addOrReplaceTimeout(std::move(clearChat.message), time);
+
+        bool ok = false;
+        int remainingTime = message->tags().value("ban-duration").toInt(&ok);
+        if (ok)
+        {
+            auto *tc = dynamic_cast<TwitchChannel *>(chan.get());
```
pajlada: Same here, check that tc is not nullptr before using it

CHANGELOG.md
achow101: Whoops, wrong thing on the clipboard, fixed.

src/widgets/splits/SplitInput.hpp
achow101: Added a comment

src/providers/twitch/IrcMessageHandler.cpp
achow101: Added a comment

src/providers/twitch/IrcMessageHandler.cpp
achow101: Done

src/providers/twitch/TwitchChannel.hpp
achow101: Added a comment

src/providers/twitch/IrcMessageHandler.cpp
achow101: Added a comment

src/providers/twitch/TwitchChannel.cpp
achow101: I added an additional check for `sendWaitEnd_.has_value()` so `syncSendWSaitTimer()` is skipped if the timer has already been unset.

src/providers/twitch/IrcMessageHandler.cpp
achow101: Done

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -460,6 +460,10 @@ void IrcMessageHandler::handleRoomStateMessage(Communi::IrcMessage *message)
         if (auto it = tags.find("slow"); it != tags.end())
         {
             roomModes.slowMode = it.value().toInt();
+            if (roomModes.slowMode == 0)
+            {
+                twitchChannel->setSendWait(roomModes.slowMode);
+            }
```
Nerixyz: Should be done in `setRoomModes`.

src/widgets/helper/ChannelView.cpp
```diff
@@ -1103,6 +1103,14 @@ void ChannelView::setChannel(const ChannelPtr &underlyingChannel)
...
                 this->liveStatusChanged.invoke();
             });
+
+        this->channelConnections_.managedConnect(
+            twitchChannel->sendWaitUpdate, [this](const QString &text) {
+                if (auto *split = dynamic_cast<Split *>(this->parentWidget()))
+                {
+                    split->getInput().setSendWaitStatus(text);
+                }
+            });
```
Nerixyz: Why is the connection done here instead of in the `Split`/`SplitInput`?

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -511,6 +515,18 @@ void IrcMessageHandler::handleClearChatMessage(Communi::IrcMessage *message)
...
+        int remainingTime = message->tags().value("ban-duration").toInt(&ok);
+        if (ok)
+        {
+            auto *tc = dynamic_cast<TwitchChannel *>(chan.get());
+            assert(tc != nullptr);
+            if (tc != nullptr)
+            {
+                tc->setSendWait(remainingTime);
+            }
+        }
```
Nerixyz: This will match any timeout for any user - not just the current one.

Nerixyz (commented):
The message length label now has more space to the right:

Image

src/providers/twitch/IrcMessageHandler.cpp
achow101: Done

src/widgets/helper/ChannelView.cpp
achow101: Wasn't sure where to put it, so was following what #3329 did

Moved to `Split`

src/providers/twitch/IrcMessageHandler.cpp
achow101: Added a check for the current username.

achow101:
> The message length label now has more space to the right:

I'm not entirely sure how to resolve that. I tried using a vbox  but that ends up increasing the height of the input box a lot when neither label is showing anything, and that looked odd to me. The message length label could be swapped to the other side of the timer label, but then the timer label would have the same issue.

Nerixyz:
> I'm not entirely sure how to resolve that. I tried using a vbox but that ends up increasing the height of the input box a lot when neither label is showing anything, and that looked odd to me. The message length label could be swapped to the other side of the timer label, but then the timer label would have the same issue.

When no timeout is active, you can hide the label, then the space will be gone. The timeout label and message length label also use a different font.

achow101:
> When no timeout is active, you can hide the label, then the space will be gone.

I see. I've implemented that now.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unversioned
 
+- Major: Added a countdown timer to the input box to show when you can chat again during slow mode or after a timeout. (#6782)
```
pajlada: Added instructions for how to enable the feature, and changed the type from Major to Minor
```suggestion
- Minor: Added a countdown timer to the input box to show when you can chat again during slow mode or after a timeout. This feature is disabled by default, and can be enabled with the "Show countdown on slow mode or when timed out" setting. (#6782)
```

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -979,6 +1007,33 @@ void IrcMessageHandler::handleNoticeMessage(Communi::IrcNoticeMessage *message)
...
+                tc->setSendWait(seconds);
+            }
+        }
+    };
+
+    if (tags == "msg_slowmode")
+    {
+        // Notice received when the user sends a message too quickly during slow mode.
+        // @msg-id=msg_slowmode :tmi.twitch.tv NOTICE #channel :This room is in slow mode and you are sending messages too quickly. You will be able to talk again in 10 seconds.
+        handleSendWait(message->content().split(" ").value(21));
```
pajlada: Apply the same split behaviour in the `msg_timedout` case too

```suggestion
        handleSendWait(message->content().split(u' ').value(21));
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2341,4 +2364,43 @@ void TwitchChannel::listenSevenTVCosmetics() const
...
+                  this->sendWaitEnd_.value() - now)
+            : 0s;
+    if (remaining <= 0s)
+    {
+        this->sendWaitTimer_.stop();
+        this->sendWaitUpdate.invoke("");
+    }
+    else
+    {
+        this->sendWaitUpdate.invoke(formatTime(remaining));
```
pajlada: For a 14 day timeout, this will currently show up like this:

Image

That feels a little verbose to me. I think it would be enough if the user received the two most significant time intervals (i.e. `13d 23h` in this case)

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -381,6 +381,13 @@ void GeneralPage::initLayout(GeneralPageView &layout)
             "limit, or a lower limit enforced by a moderation bot")
         ->addTo(layout);
 
+    SettingWidget::checkbox("Show countdown on slow mode or when timed out",
```
pajlada, line 384: Utilize the `addKeywords` function to add some keywords the user might search for (e.g. `timeout` or `slowmode`)

CHANGELOG.md
achow101: Done as suggested

src/providers/twitch/IrcMessageHandler.cpp
achow101: Done

src/providers/twitch/TwitchChannel.cpp
achow101: Done

src/widgets/settingspages/GeneralPage.cpp
achow101: Done

achow101:
> but it would be good to clear the timer after we've detected a successful message from ourselves that's sent, or a user could be stuck with a 14d timer.

I changed this so that the timer starts/resets whenever we see a message from the user in the channel. This should resolve the untimeout issue. I think this also resolves the automod issue as well.

src/util/FormatTime.hpp
```diff
@@ -12,9 +12,13 @@
 namespace chatterino {
 
 // format: 1h 23m 42s
-QString formatTime(int totalSeconds);
-QString formatTime(const QString &totalSecondsString);
-QString formatTime(std::chrono::seconds totalSeconds);
+// 'components' controls how many most significant components the formatted time will have
+QString formatTime(int totalSeconds,
```
pajlada: No need to use an `optional` here, just use an int with the default value clearly defined in the default param (i.e. `QString formatTime(..., int components = 4);`

achow101: Done

PR #6792, opened 2026-02-02
feat(plugins): Add images and image sets

Nerixyz:
When adding message construction and introspection, I left out images. This PR intends to add them.

Images are special, because plugin authors can specify a URL that Chatterino will load if the image is shown. While the plugin can't see the response, it still causes Chatterino to issue a request, so _creating_ images will require the network permission.

Other than that, images and image sets are mostly straight forward. You can query them and on sets, you can change the images. Note that sets are value types, so getting an image set in Lua will always refer to a copy, not a reference.

When querying the empty image, Chatterino would try to access `frames_` which is not set there. I wonder why this never caused any problems before (guess we filter empty images before layout?). But I added checks in the accessors.


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 - Minor: Added action to reset `/watching`. (#6759)
 - Minor: Removed messaging about running flatpak. This is already apparent in the newer Flatpak runtimes. (#6768)
 - Minor: Add `/(un)monitor` and `/(un)restrict` commands for moderators. (#6750, #6785)
+- Minor: Added support for images and image sets in plugins. (#6792)
```
pajlada: Remove changelog entry

pajlada (changes requested):
Not tested, API looks good

docs/wip-plugins.md
```diff
@@ -812,6 +812,53 @@ Is this connection currently blocked?
...
+
+#### `ImageSet`
+
+A set of images. Each image should depict the same content at different sizes.
+
+Image sets have the following writable properties:
+
+- `image1` ([`Image`](#image)): The base image (1x).
+- `image2` ([`Image`](#image)): The first scaled image (often 2x, `scale=0.5`)
+- `image3` ([`Image`](#image)): The second scaled image (often 3/4x, `scale=0.25`)
```
Mm2PL: 3/4x reads like $\frac34 x$

```suggestion
- `image3` ([`Image`](#image)): The second scaled image (often 3x or 4x, `scale=0.25`)
```

src/controllers/plugins/api/Images.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+template <typename T, typename... Args>
+T noPermissionFn(Args... /* unused */)
+{
+    throw std::runtime_error("Missing network permission to create images");
+}
+
+void validateUrl(const QString &urlString)
+{
+    QUrl url(urlString);
```
Mm2PL, line 19: Should we allow plugins to use resources baked into our program?

src/controllers/plugins/api/Images.cpp
```diff
@@ -0,0 +1,130 @@
...
+void createUserTypes(sol::table &c2, const Plugin &plugin)
+{
+    auto *imageCtorPtr =
+        &noPermissionFn<ImagePtr, const QString &, sol::variadic_args>;
+    auto *imageSetCtorPtr = &noPermissionFn<ImageSet, sol::variadic_args>;
+    if (plugin.hasNetworkPermission())
+    {
+        imageCtorPtr = &imageCtor;
+        imageSetCtorPtr = &imageSetCtor;
+    }
```
Mm2PL: I would prefer this check to happen inside of the constructor function just like every other API.

src/controllers/plugins/api/Images.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+---Get the empty image
+---@return c2.Image
+function c2.Image.empty() end
+
+---A set of images. Each image should depict the same content at different sizes.
+---@class c2.ImageSet
+---@field image1 c2.Image The base image (1x).
+---@field image2 c2.Image The first scaled image (often 2x, `scale=0.5`)
+---@field image3 c2.Image The second scaled image (often 3/4x, `scale=0.25`)
```
Mm2PL: second `3/4x` mention here

Mm2PL (changes requested):
~~I cannot get this to work.~~ This requires element flags or the image won't be visible. Maybe add a warning that an image is being constructed without flags.

```lua
is = c2.ImageSet.new(
    "https://fourtf.com/chatterino/badges/supporter.png",
    "https://fourtf.com/chatterino/badges/supporter2x.png",
    "https://fourtf.com/chatterino/badges/supporter3x.png"
)
m = c2.Message.new({
    elements={
        {type='timestamp'},
        {type='scaling-image', images=is},
        {type='text', text='asdasd'}
    }
})
c2.Channel.by_name('pajlada'):add_message(m)
```

src/controllers/plugins/api/Images.cpp
pajlada: I think this should come later with another type of constructor and maybe permission

Nerixyz:
> This requires element flags or the image won't be visible. Maybe add a warning that an image is being constructed without flags.

Yes, this is a small pitfall. I made the flags required now, so you have to specify some (`::None` is technically still an option). Not sure about a warning. I think we should explain the flags more, though one part of it is ChannelView, which plugins can't use.

pajlada (approved):

src/controllers/plugins/LuaAPI.hpp
```diff
@@ -84,13 +84,20 @@ struct CompletionEvent {
 
 sol::table toTable(lua_State *L, const CompletionEvent &ev);
 
+/* @lua-fragment
+
+---@alias QSize {width: integer, height: integer}
+---@alias QSizeF {width: number, height: number}
```
Mm2PL: These also could be an array-table

Mm2PL (approved):
smol thing

Mm2PL (approved):

PR #6796, opened 2026-02-05
refactor: move message rate-limit to `TwitchChannel`

Nerixyz:
When writing #6795, I noticed that we enforce the rate-limit globally for all channels. So if you're in this rate-limit state in one channel, you won't be able to send messages in other channels (where the user has the same rate-limit). I tested with IRC and Helix that they don't enforce this globally (hitting the rate-limit in one channel and typing in another), so we should track it per channel.

I moved the code from `TwitchIrcServer`  to `TwitchChannel`. It's mostly 1:1 identical, except that `channel` is `this`.

For end-users this is more of a bugfix, since previously, they weren't able to send messages in other channels during that state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
 - Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6710)
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723, #6739)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
+- Bugfix: Fixed messages not sending in other channels when hitting the local rate-limit. (#6796)
 - Dev: Update release documentation. (#6498)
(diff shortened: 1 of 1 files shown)


Mm2PL:
> When writing #6795, I noticed that we enforce the rate-limit globally for 
> all channels. So if you're in this rate-limit state in one channel, you 
> won't be able to send messages in other channels (where the user has the 
> same rate-limit).
So if I were to spam #pajlada at 100/30s, then I could spam another 100 
messages in #mm2pl in that same 30 second window and I could send yet 
another 20 messages in #forsen (assuming I don't hit other global rate 
limits). This feels very suspect to me knowing this did not use to be 
the case for a very long time. I don't see many issues with the code 
itself but I really want to test TMI for this behavior.

Nerixyz:
Yeah, I didn't test this fully, looks like you can only send one message in another channel if you're rate-limited.

PR #6799, opened 2026-02-07
chore(windows): add command to relaunch with a new console

Nerixyz:
Recently, when debugging an issue with someone else, they were not comfortable launching Chatterino from the command line or using command line tools like `c2-logs` to get the logs. We shouldn't require users to learn how to use the command line to fix their issues.

With this PR, I'm adding `/debug-relaunch-with-console [logging-rules...]` that relaunches Chatterino with a new console and some logging rules, because in almost all cases, you'd want to inspect the logs. In the console, they can see all the logs and copy them.


src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -246,4 +248,47 @@ QString debugTest(const CommandContext &ctx)
...
+    {
+        env.insert(loggingRulesEnv, ctx.words.mid(1).join(';'));
+    }
+    else if (!env.contains(loggingRulesEnv))
+    {
+        // by default, enable all debug logging
+        env.insert(loggingRulesEnv, "chatterino.*.debug=true");
+    }
+
+    QProcess proc;
```
pajlada, line 275: My only concern with this is racy settings/window layout saves/loads.
Not sure if super feasible, but maybe we can pre-emptively save & then disable all saves before launching the new Chatterino instance?

PR #6800, opened 2026-02-08
fix: incorrect deletion length in tab autocomplete when completion included emojis

### LINKED ISSUE #6771: Tab autocomplete on emotes with emoji on name eating text.
### Checklist

### Describe your issue

If a emote (bttv/ffz/7tv) has a emoji in the name after using tab to autocomplete he will delete the words before the one how was completed.

https://github.com/user-attachments/assets/cfbc27ae-80de-4fe1-b31b-8aba3069da0c

### Screenshots

Emote example:

Image

### OS and Chatterino Version

Chatterino 2.5.4 (commit a2ba163a4) built with Qt 6.10.1
Running on EndeavourOS, kernel: 6.18.6-arch1-1

ruinivist:
Closes #6771 

`prefixSize` accounted for number of `QChar` objects in the string which could be different from the "visual length" used in `tc.movePosition(QTextCursor::PreviousCharacter, QTextCursor::KeepAnchor, prefixSize);`, due to emojis using multiple characters for representation.

Instead of relative offsets, this sets the index based position instead where both `tc.position()` and `prefixSize`  work with `QChar` units. 
https://doc.qt.io/qt-6/qtextcursor.html#position

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723, #6739)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
 - Bugfix: Fixed incorrect deletion length in tab autocomplete when completion included emojis. (#6800)
+- Bugfix: Fixed WebSockets for 7TV or BTTV not connecting on Windows if IPv6 attempts fail. (#6805)
 - Bugfix: Fixed portable updates not showing an error if the updater is not present. (#6801)
 - Dev: Nightly builds are now defined through a build flag rather than the Modes file. (#6798)
 - Dev: Update release documentation. (#6498)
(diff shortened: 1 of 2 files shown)


ruinivist:
question: Noticed this while testing, this `Cluele|ss ` and tab converts to -> `Clueless |ss `. I assume just moving the caret to the end like `Clueless |` would be better no? I can add that

Nerixyz:
> Noticed this while testing, this `Cluele|ss ` and tab converts to -> `Clueless |ss `. I assume just moving the caret to the end like `Clueless |` would be better no?

My code editor behaves like Chatterino here except that it doesn't insert a space, so for me, the current behavior is expected.

ruinivist:
> My code editor behaves like Chatterino

~Fair, but then I would expect it to break the emote like `Cluele |ss` instead of completing and breaking.~

Edit: nvm, I see what you mean

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
 - Bugfix: Disable "Sort Tabs Alphabetically" action when notebook layout is locked. (#6710)
 - Bugfix: Fix highlight mentions not updating when username changes. (#6723, #6739)
 - Bugfix: Fixed Return and Enter being treated as different keys on Mac OS. (#6726)
+- Bugfix: Fixed incorrect deletion length in tab autocomplete when using emojis. (#6800)
```
pajlada:
```suggestion
- Bugfix: Fixed incorrect deletion length in tab autocomplete when completion included emojis. (#6800)
```

pajlada (approved):

PR #6804, opened 2026-02-13
refactor(websockets): Balance IPv4 and IPv6 attempts

Nerixyz:
Since #6631, we try to connect to all resolved hosts.
However, we don't control the order. At least on Windows, the results are usually grouped into IPv4 and IPv6 (i.e. either all v4 records followed by v6 or vice versa). When v6 records come first and the machine fails to connect over v6 because the connection times out, the user needs to wait `30s * #v6-records` for a connection to happen. This can take some time. For example, 7TV's event API resolves seven records for me, so they'd wait 3 min 30 s.

With this PR, we balance the entries we try. If possible, an attempt to a v4 IP is followed by a v6 one and vice versa. I chose to start with v4 first, because that's more common, but we could also make that decision random.

### DIFF
--- a/src/common/websockets/detail/BalancedResolverResults.hpp
+++ b/src/common/websockets/detail/BalancedResolverResults.hpp
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
 #pragma once
 
 #include <boost/asio/ip/basic_resolver_results.hpp>
--- a/src/common/websockets/detail/BalancedResolverResults.cpp
+++ b/src/common/websockets/detail/BalancedResolverResults.cpp
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
 #include "common/websockets/detail/BalancedResolverResults.hpp"
 
 namespace chatterino::ws::detail {


src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -99,51 +99,47 @@ void WebSocketConnectionHelper<Derived, Inner>::onResolve(
...
 {
-    if (endpointIterator == this->resolvedEndpoints.end())
+    if (!entry)
     {
         this->fail("Ran out of resolved endpoints"sv, u"connect");
         return;
     }
 
-    const auto &endpoint = endpointIterator->endpoint();
+    auto ep = entry->endpoint();
```
pajlada: nit: keep old variable name (i.e. `endpoint` instead of `ep`)

src/common/websockets/detail/BalancedResolverResults.hpp
```diff
@@ -0,0 +1,58 @@
...
+    /// Returns true if `currentIdx` now points to an IPv4 entry.
+    bool advanceIPv4();
+
+    /// Advance the IPv6 iterator.
+    /// Returns true if `currentIdx` now points to an IPv6 entry.
+    bool advanceIPv6();
+
+    size_t nextIPv4Idx = 0;
+    size_t nextIPv6Idx = 0;
+    bool currentIsIPv6 = true;
```
pajlada: I think it would be easier to understand this variable name if it instead pointed to what will come next (i.e. `nextIsIPv6`)

pajlada (approved):

PR #6806, opened 2026-02-17
fix(liveupdates): bump connections when adding a client

Nerixyz:
Fixes a bug where the debug count of liveupdates connections could get negative when connection attempts fail. `onConnectionOpen` is only called once the WS handshake is done, so if the connection fails, we'd subtract too much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,8 @@
 - Dev: Updated `pajlada-serialize` to v0.2.1. (#6797)
 - Dev: Updated `pajlada-signals` to v0.1.1. (#6797)
 - Dev: Balance IPv4 and IPv6 connection attempts. (#6804)
+- Dev: Factored out recent messages benchmark helper. (#6815)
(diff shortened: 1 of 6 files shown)


src/providers/liveupdates/BasicPubSubManager.hpp
```diff
@@ -147,7 +147,6 @@ class BasicPubSubManager : public QObject
     {
         assertInGuiThread();
 
-        DebugCount::increase(DebugObject::LiveUpdatesConnection);
         this->addingClient_ = false;
         this->diag.connectionsOpened.fetch_add(1, std::memory_order_acq_rel);
```
pajlada, line 151: Is this also incorrect then? Or is it tracking a different thing

Nerixyz: No it's tracking the actually opened connections. For the tests, we could use either place.

pajlada (approved):

PR #6814, opened 2026-02-20
chore(filters): improve test coverage and add benchmarks

Nerixyz:
Filters had moderate test coverage. This improves the coverage to include more lines and adds a few benchmarks.


src/controllers/filters/FilterSet.hpp
```diff
@@ -25,6 +25,7 @@ class FilterSet
 public:
     FilterSet();
     FilterSet(const QList<QUuid> &filterIds);
+    FilterSet(QMap<QUuid, FilterRecordPtr> filters);
```
pajlada: I'm not fond of this constructor since it doesn't work like a normal filter set. I'd prefer if the tests were written in a way where they could use one of the existing constructors

pajlada (changes requested):

pajlada (approved):

PR #6815, opened 2026-02-20
refactor(benchmark): move recent messages app/bench to helper

Nerixyz:
Split from #6814.

This moves the `MockApplication` and benchmark runner out to a helper in order to be able to reuse both in other benchmarks.


benchmarks/src/MessageBuilding.hpp
```diff
@@ -0,0 +1,147 @@
+#pragma once
```
pajlada, line 5: license

benchmarks/src/MessageBuilding.cpp
```diff
@@ -0,0 +1,86 @@
+#include "MessageBuilding.hpp"
```
pajlada, line 5: license

pajlada (approved):

PR #6821, opened 2026-02-22
fix(spellcheck): treat words separated by `'` and `-` as one

### LINKED ISSUE #6789: Spellchecking: Hyphenated compound words not recognized as one word
### Checklist

### Describe your issue

When making a typo in a hyphenated compound word only the misspelled part is underlined and the spelling suggestion only suggests fixes for this part instead of the entire word.

There already was a short discussion of this in https://github.com/Chatterino/chatterino2/pull/6780#discussion_r2749405687.

### Screenshots

Image

### OS and Chatterino Version

Chatterino 2.5.4 (commit [30141748](https://github.com/Chatterino/chatterino2/commit/30141748)) built with Qt 6.10.1
Running on Arch Linux, kernel: 6.18.7-arch1-1

Nerixyz:
The following words should be treated as one:

- `don't`
- `doesn't`
- `wouldn't've` (My Hunspell dictionary doesn't like this, but it's a word)
- `follow-up` (#6789)

This is done by matching a word as `\p{L}+(?:['-]\p{L}+)*`. This ensures that `'` or `-` isn't matched at the start/end.

Fixes #6789.

### DIFF
--- a/src/common/Credentials.cpp
+++ b/src/common/Credentials.cpp
@@ -22,11 +22,7 @@
 
 #ifndef NO_QTKEYCHAIN
 #    ifdef CMAKE_BUILD
-#        if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-#            include "qt6keychain/keychain.h"
-#        else
-#            include "qt5keychain/keychain.h"
-#        endif
+#        include "qt6keychain/keychain.h"
 #    else
 #        include <qtkeychain/keychain.h>
 #    endif
--- a/src/common/LinkParser.cpp
+++ b/src/common/LinkParser.cpp
@@ -33,12 +33,6 @@ TldSet &tlds()
         }
         QTextStream stream(&file);
 
-#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-        // Default encoding of QTextStream is already UTF-8, at least in Qt6
-#else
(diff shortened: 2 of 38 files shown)


Nerixyz:
Hmm, thinking about this, I don't know if we should treat words with hyphens in them as one. Think about `my-super-duper-word`. That should be valid, no?

4rneee:
> Hmm, thinking about this, I don't know if we should treat words with hyphens in them as one. Think about `my-super-duper-word`. That should be valid, no?

Fair point. Firefox seems to allow such words but if one contains a typo the entire word is underlined and suggestions are for the entire word, although only the misspelled part changes in the suggestions.


Edit:
I just tested this PR and noticed that `my-super-duper-word` does not get underlined. If I make a typo in one of the words the entire word gets underlined which I think is good.
The only issue I noticed is that there is no suggestion for the entire word and you have to right-click the misspelled word in order to get a suggestion.

spellcheck

Nerixyz:
> I just tested this PR and noticed that `my-super-duper-word` does not get underlined.

Ah, seems like Hunspell does this for us (?).

> The only issue I noticed is that there is no suggestion for the entire word and you have to right-click the misspelled word in order to get a suggestion.

This should be fixed with #6822 (so we use the same "definition" of what a word is).

PR #6822, opened 2026-02-22
fix(spellcheck): use input highlighter's word segmentation to find word

### LINKED ISSUE #6819: Completions not correctly bound to word
### Checklist

### Describe your issue

Type "I love memees!" into the chat input box.
Right-click the the `memee` part and see it properly try to suggest words.
Right-click the `s` part and see it try to incorrectly suggest single letter fixes.

Does not happen with "I love memees.", but right-clicking `s` there provides no suggestions instead.

Some off by one error?

Using my system `en_GB` dictionary if that matters.

### Screenshots

Image

### OS and Chatterino Version

Chatterino Nightly 2.5.4 DEBUG (commit [dbfbcb8d7](https://github.com/Chatterino/chatterino2/commit/dbfbcb8d7) modified) built on 2026-02-22 with Qt 6.10.2
Running on Arch Linux, kernel: 6.18.9-arch1-2

Nerixyz:
We used to use `QTextCursor`'s definition of a word when right-clicking. This doesn't necessarily align with ours (e.g. https://github.com/Chatterino/chatterino2/pull/6821#issuecomment-3941244373). Furthermore, we didn't consider the character after a word as part of it. Since Qt does some rounding, it might determine that you clicked on the character after a word if you clicked on the right side of it. With this PR, that's fixed.

Here's a visualization of which characters we consider as part of a word (separated by even/odd):

```
     ███     █████  ██████   ████
this is some text#I#write in here
█████   █████     ██      ███
```

Fixes #6819.


pajlada:
Seems like it highlights words correctly

image

But when attempting to apply the suggestion, it doesn't do that correctly.

image

PR #6841, opened 2026-03-10
fix: watch streak filter

pajlada:
In v2.5.4, Watch Streak messages were incorrectly labeled as Subscriber Messages. This was fixed in #6571, but we didn't add a new filter to compensate for this change.
This means in v2.5.5-beta.1, people now have a hard time filtering out Watch Streak messages.
This PR fixes this by adding the `flags.watch_streak` filter.

Thanks @Gucci-z for reporting this issue.

When this change is merged in, remember to update the 2.5.5 website PR to include this in the changelog.


pajlada:
image

PR #6842, opened 2026-03-12
fix(updates): handle nightly downgrade message

pajlada:
added a bonus unit test for fun

This PR ensures a different message is displayed for nightly users who are running a "too-new-version".

### DIFF
--- a/src/singletons/Updates.cpp
+++ b/src/singletons/Updates.cpp
@@ -462,7 +462,7 @@ QString Updates::buildUpdateAvailableText() const
         if (this->isDowngrade())
         {
             return QString("The version online (%1) seems to be lower than the "
-                           "current (%2).\nEither a version war reverted or "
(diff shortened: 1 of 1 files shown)


src/singletons/Updates.cpp
```diff
@@ -459,6 +459,15 @@ QString Updates::buildUpdateAvailableText() const
     if (version.isNightly())
     {
         // Since Nightly builds can be installed in many different ways, we ask the user to download the update manually.
+        if (this->isDowngrade())
+        {
+            return QString("The version online (%1) seems to be lower than the "
+                           "current (%2).\nEither a version war reverted or "
```
Wissididom:
```suggestion
                           "current (%2).\nEither a version was reverted or "
```

pajlada: WAR

pajlada:
image

PR #6856, opened 2026-03-19
fix: update message length when completing

### LINKED ISSUE #6855: Message length label not updating when tabcompleting
### Checklist

### Describe your issue

When tabcompleting (emotes, commands or usernames) the message length is not updated until typing again.

I assume the issue originates from the fact that `SplitInput::editTextChanged` is not called after inserting the completion using `ResizingTextEdit::insertCompletion`. `ResizingTextEdit::textChanged` is already connected to `SplitInput::editTextChanged` but this seems not to be enough despite `ResizingTextEdit::insertCompletion` inserting text (maybe even a bug in Qt?).

I already tried adding an explicit `this->textChanged()` in `ResizingTextEdit::insertCompletion` and also tried directly connecting `ResizingTextEdit::insertCompletion` with `SplitInput::editTextChanged` but both approaches didn't work and I'm not sure why.

### Screenshots

https://github.com/user-attachments/assets/e0c504ee-89db-4f0e-880c-6110c2154224

### OS and Chatterino Version

Chatterino 2.5.5-beta.1 (commit [de9496a8](https://github.com/Chatterino/chatterino2/commit/de9496a8)) built with Qt 6.10.2
Running on Arch Linux, kernel: 6.19.8-arch1-1

Nerixyz:
When the completer completes, we block all signals. However, we do want to broadcast a text-changed signal. Since we can't unblock individual signals, we unconditionally emit a signal after we unblock all signals. If the text didn't change, this is technically wrong, but doesn't hurt.

Fixes #6855.


Mm2PL:
This introduces a weird behavior with Message Overflow = Prevent where tab completion is stopped because of the clipping of the message. The previous behavior of allowing that word to exist until otherwise modified was also weird.

2026-03-19_18-53-43 mkv

Nerixyz:
> This introduces a weird behavior with Message Overflow = Prevent where tab completion is stopped because of the clipping of the message.

I'm not sure what the expected behavior here is. Should the emotes complete but cut off? I think the current behavior is somewhat reasonable. You can't complete emotes if they'd overflow. Assuming the limit is correct, you wouldn't be allowed to send the message after completing anyway.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Bugfix: Make it possible to filter Watch Streak messages again. Previously, these messages were labeled `sub_messages`. This was fixed in #6571, but we didn't add a new filter to compensate. (#6741)
 - Bugfix: Improve nightly downgrade message. (#6842)
 - Bugfix: Improved blocked user loading on poor internet connections. (#6846)
+- Bugfix: Fixed message length label not updating when tab-completing. (#6856)
```
pajlada: Remove this changelog entry

PR #6857, opened 2026-03-19
feat: hide user notes in streamer mode by default

pyrrhoscliff:
Adds an option to hide user notes in usercards when Streamer Mode is enabled. 
The setting is on by default to protect streamer privacy, but can be disabled 
in Settings > Streamer Mode > "Hide user notes".

Related to #6510.

**Testing:**
1. Add a note to any user via their usercard
2. Enable Streamer Mode in Settings
3. Open that user's usercard and notes should be hidden
4. Uncheck "Hide user notes" in Streamer Mode settings
5. Open the usercard again and notes should now be visible


Mm2PL:
Feel free to ignore the 'changelog check' workflow. We're moving to a workflow where we make the changelog based on commit history rather than a CHANGELOG.md file.

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -1193,7 +1193,12 @@ void UserInfoPopup::updateNotes()
         this->ui_.notesPreview->setVisible(false);
         return;
     }
-
+    if (getApp()->getStreamerMode()->isEnabled() &&
```
Mm2PL, line 1201: This could display that notes are hidden in streamer mode.

Also it does not update when streamer mode status is changed while the usercard is already open. I don't think it would be an issue if this would display that notes were hidden.

pyrrhoscliff: Updated to show "Notes hidden in streamer mode." when notes are hidden, and added a connection to the streamer mode `changed` signal so the notes update live when streamer mode is toggled.

Mm2PL (approved):

pajlada:
With the current implementation, if a streamer shows off a user card on stream, the text "Notes hidden in streamer mode." will give away that there are user notes for that user.

Is that a concern? I think it's fine, but would love your input.

Mm2PL:
> With the current implementation, if a streamer shows off a user card on stream, the text "Notes hidden in streamer mode." will give away that there are user notes for that user.

I think it's fine too.

Mm2PL:

Issue #4613, opened 2023-05-09
Chatterino preventing computer from going to sleep, all sound options are disabled


Shadowkillerdragon:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Chatterino prevents my computer from going to sleep mode due to it detecting a audio stream from the application for the sound pings. I believe this started occuring in 2.4.2 as I was having it during that build as well. Prior versions to this I was able to have the application open and with the sound effects disabled it didnt take control like that.  First command request is with chatterino open and the second is with it closed.

### Screenshots

![chatterino](https://user-images.githubusercontent.com/11730650/236998404-7967eb89-d2fe-4bc1-9f42-0a1c74ede80d.png)

### OS and Chatterino Version

Windows 11 22H2 2.4.2/2,4,3

ShowMeBillyJo:
I have the same issue

pajlada:
This has been mostly fixed in https://github.com/Chatterino/chatterino2/pull/4978 - not available in a stable release yet though.

It's fixed in two ways:
1) The sound device will only be activated when we try to play a sound, meaning if you have no highlights that play sounds then it shouldn't prevent your computer from going to sleep. If a sound does play, the sound device **should** be released after 30 seconds of it being idle, allowing your computer to go to sleep.
2) We now support multiple sound backends, so instead of using the default miniaudio backend, you can use the "Null" backend which disables sound playback entirely, making sure Chatterino would never stop your computer from going to sleep for any sound-related reason.

PR #6859, opened 2026-03-21
feat: add setting to always keep sound engine alive

pajlada:
maybe useful for wireless headphone users who miss first ping?

maybe there's a better way to solve this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unversioned
 
-- Minor: Added a setting to disable sound device sleeping. This can help ensure highlight sounds play more reliably when using a wireless sound device. (#6861)
+- Minor: Added a setting to disable sound device sleeping. This can help ensure highlight sounds play more reliably when using a wireless sound device. (#6859)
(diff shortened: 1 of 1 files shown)


Bakint0:
Yes, it is useful for wireless headphone users! It fixed my problem for not getting pinged the first time but after that. So now I get pinged every time!

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1586,6 +1586,13 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         ->setTooltip("Change this only if you're noticing issues with sound "
                      "playback on your system")
         ->addTo(layout);
+    SettingWidget::checkbox("Keep sound backend alive (requires restart)",
+                            s.soundMiniaudioKeepEngineAlive)
+        ->setTooltip(
+            "This setting makes Chatterino output silence to your sound "
+            "device, even if no ping is being played. Try this setting if you "
+            "have issues with wireless devices not playing the first sound.")
```
Nerixyz: Mention that this might prevent the computer from going to sleep (https://github.com/Chatterino/chatterino2/issues/4613).

pajlada: Fixed in [`494866d` (this PR)](https://github.com/Chatterino/chatterino2/pull/6859/commits/494866d8e45378bfce9b4bfcec0ca3ae343b1bc4)

PR #6861, opened 2026-03-22
Optionally allow a message when any filter matches

achow101:
It can be useful to allow a message if any of the filters match it, not just when all of them match. This PR adds a checkbox to the filter selection dialog which enables this behavior if the user wants to use it for that split.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Minor: Update emoji data to Unicode 17.0. (#6471)
 - Minor: Added a setting to disable sound device sleeping. This can help ensure highlight sounds play more reliably when using a wireless sound device. (#6859)
+- Minor: Added an option to match on any of a filter in a split rather than all selected filters. (#6861)
 - Bugfix: Make it possible to filter Watch Streak messages again. Previously, these messages were labeled `sub_messages`. This was fixed in #6571, but we didn't add a new filter to compensate. (#6741)
 - Bugfix: Improve nightly downgrade message. (#6842)
 - Bugfix: Improved blocked user loading on poor internet connections. (#6846)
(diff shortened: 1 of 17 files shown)


pajlada:
I'm not 100% aligned on this, I'd like to talk more about how filters are applied post 2.5.5 but that's probably best done in a discussion rather than in a PR.

Next changes coming to Filters will be how they are matched (context map)

achow101:
I see. Is there an existing discussion for this or should I open a new one? And is there any existing work/ideas yet (I saw #6695 but I'm not sure if that's entirely related)? This is something that I would be interested in working on.

pajlada:
Feel free to create one, we've only had very casual conversations about this.

Nerixyz:
One thing I'd like to note is that we're essentially re-implementing a (chained) binary operation for the aggregation in `FilterSet::filter`. With the current design, that's the only way to do it, because we can't reuse an `Expression` in more than one other place. Though, I'm not sure if we'd have anything other than "any-of" and "all-of".

pajlada:
achow intends to make a github discussion where we can discuss it a bit more, but we spoke about this a bit:

Nerixyz has some pending changes that removes/reworks the `ContextMap` (where we map things like `author.name`. All functionality will still be the same, but instead of us duplicating data there, we provide functions that can be use on-demand by the filters to request the data. This will reduce the memory requirements for each message by a good bit.

Frontend-wise we have not written any code, so if this is something you'd like to look at that would be great. Let me explain the issues I have with the filter UI a bit:

1. The dialog starts off way too small/isn't scrollable. It should have a better minimum size, and the list of filters should be scrollable
2. The default behaviour of a filter is backwards in my head. I would love for the ability to apply a filter or the inverse of a filter inside the filter dialog itself. If I want to hide all channel point redemptions in one split and only show channel point redemptions in another split, I have to create two filters that are inverse of each other right now. Preferably, this is not the case.
3. Related to your PR: How filters behave when multiple are applied can also feel wrong. It would be great to be able to apply some logic operator to it (we'd probably use plain english rather than AND/OR). I'd like this to have this more as a dropdown rather than a checkbox so we can add new things later down the line (COUNT maybe?)

PR #6870, opened 2026-03-24
fix: use user ID in ban/timeout commands from user card when available

### LINKED ISSUE #6825: User card ban button uses username
### Checklist

### Describe your issue

If target is platform banned, they cannot be banned or timed out to delete their messages. Chatterino should ban/timeout by ID if it's available to prevent this situation from happening.

### Screenshots

`[16:44:56.568] Failed to ban, bad target name: misvrine`

### OS and Chatterino Version

Chatterino 2.5.4 DEBUG (commit 0f5c0bc3d modified) built with Qt 6.10.2
Running on Arch Linux, kernel: 6.18.8-arch2-1

voiceofgrog:
If a user is platform-banned, Twitch's API rejects ban/timeout commands
that use a username, returning "bad target name". This fix uses the
`id:` prefix syntax in the user card's ban/timeout buttons and the
`execModeratorAction` plugin handler when the user ID is available,
bypassing the username lookup entirely. Falls back to username if the
ID is not yet populated.

Fixes #6825

**Testing:**
- Verified normal timeout behavior is intact by testing on a live channel
- The platform-banned case is difficult to reproduce locally as it
  requires access to a suspended account

### DIFF
--- a/src/widgets/dialogs/UserInfoPopup.cpp
+++ b/src/widgets/dialogs/UserInfoPopup.cpp
@@ -201,17 +201,20 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, Split *split)
                         "moderation action to execute, see description in the "
                         "editor";
              }
+             const QString userTarget = this->userId_.isEmpty()
+                                            ? this->userName_
+                                            : "id:" + this->userId_;
              auto target = arguments.at(0);
              QString msg;
 
              // these can't have /timeout/ buttons because they are not timeouts
              if (target == "ban")
              {
-                 msg = QString("/ban %1").arg(this->userName_);
+                 msg = QString("/ban %1").arg(userTarget);
              }
              else if (target == "unban")
              {
-                 msg = QString("/unban %1").arg(this->userName_);
+                 msg = QString("/unban %1").arg(userTarget);
              }
              else
              {
@@ -241,7 +244,7 @@ UserInfoPopup::UserInfoPopup(bool closeAutomatically, Split *split)
                  }
                  const auto &button = timeoutButtons.at(buttonNum - 1);
                  msg = QString("/timeout %1 %2")
-                           .arg(this->userName_)
+                           .arg(userTarget)
                            .arg(calculateTimeoutDuration(button));
              }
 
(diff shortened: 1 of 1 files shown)


Mm2PL:
> The platform-banned case is difficult to reproduce locally as it requires access to a suspended account

I don't think that easily works because there will be no ID data in that usercard.
image

We might need to switch opening usercards to use ids from messages when building messages and we have the IDs. We already have support for this if a usercard is opened with a username like `id:117691339`. If you're willing to do more, you could add a code path that has both ID and username although that might be a little bit hacky.

voiceofgrog:
Thanks for the clarification and the screenshot. I'm planning to fix this in `appendUsername` in `MessageBuilder.cpp`, where the username link is set for normal chat messages. Since `userID` is already available there from the IRC tags, I can pass `id:<userID>` as the link target instead of the display name, so the user card opens with the ID already set.

There are several other `Link::UserInfo` usages in `MessageBuilder.cpp` where the ID also appears to be available. Should I address those in this PR as well, or keep the scope to the normal chat message case for now?

Mm2PL:
> There are several other `Link::UserInfo` usages in `MessageBuilder.cpp` where the ID also appears to be available. Should I address those in this PR as well, or keep the scope to the normal chat message case for now?

It'd be nice to have it for timeout/ban messages too. Both eventsub and 
irc deliver the target's user id but I'm not sure if it wouldn't require 
additional refactoring to move the data to where we build message elements.

voiceofgrog:
I'm planning on tackling both paths. For the normal chat message case, I'll update `appendUsername` in `MessageBuilder.cpp` to use `id:<userID>` as the link target. For the IRC timeout/ban case, `parseClearChatMessage` in `IrcMessageHandler.cpp` already reads `ban-duration` from the `CLEARCHAT` tags so I'll extract `target-user-id` there too and thread it through the `TimeoutMessageTag` constructors.

For the EventSub path, `appendUser` in the EventSub message builder uses `MentionElement`, which hardcodes `Link::UserInfo` with the login name and forbids `setLink` via `assert(false)`. Would you prefer I modify `MentionElement` to optionally accept an ID, or is there another approach you'd recommend?

I'll convert this to a draft while I work on it.

PR #6877, opened 2026-03-28
Use environment variables for test API addresses

achow101:
Instead of hardcoding the addresses for the test API servers, make it possible to set them via environment variables. PUBSUB_WSS_ADDR replaces 127.0.0.1:9050, HTTPBOX_ADDR replaces 127.0.0.1:9051, and PUBSUB_WS_ADDR replaces 127.0.0.1:9052.

I think this change also obsoletes `CHATTERINO_TEST_USE_PUBLIC_HTTPBIN` but I left that in to avoid breaking anyone that's using that.

### DIFF
--- a/tests/src/Test.cpp
+++ b/tests/src/Test.cpp
@@ -20,7 +20,7 @@ const QString HTTPBIN_BASE_URL =
 #else
 const QString HTTPBIN_BASE_URL =
     "http://" +
-    qEnvironmentVariable("CHATTERINO_TEST_HTTPBOX_ADDR", "127.0.0.1:9050");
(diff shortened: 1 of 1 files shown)


Nerixyz:
I'd prefer if the pubsub variables included `TEST` to indicate that they're not used in the main app.

Other than that, these are not environment variables, they're global variables right now.

achow101:
It was using `std::getenv` in `main()` to update the globals from environment variables.

I've updated the names to have `CHATTERINO_TEST_` prefixes and use `qEnvironmentVariable`.

PR #6882, opened 2026-03-29
fix: Print fully qualified name in EventSub generator

Nerixyz:
The JSON generation was broken with Clang 22, because the type names didn't include the namespace (e.g. `std::string` → `string`). We need to tell Clang to use the fully qualified name. This is possible with `Type.get_fully_qualified_name` which was added in Clang 21. 

This does have the consequence that more types are now fully qualified.


.github/workflows/eventsub-test.yml
```diff
@@ -40,3 +46,13 @@ jobs:
...
+
+      - name: Build
+        run: |
+          cmake -B build -S lib/twitch-eventsub-ws -DFORCE_JSON_GENERATION=On
+          make -C build -j"$(nproc)" twitch-eventsub-ws
+
+      - name: Check generated sources
+        run: |
+          git add -N lib/twitch-eventsub-ws/include lib/twitch-eventsub-ws/src
+          git --no-pager diff --exit-code lib/twitch-eventsub-ws/include lib/twitch-eventsub-ws/src
```
pajlada, line 58: nice nice

PR #6883, opened 2026-03-29
Use case() function in GitHub Actions

### LINKED ISSUE #6783: CI: Replace ` P && A || B`  with `case()`
GitHub Actions now has a [`case()`](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case) function. This can be used to replace our ternaries in the form `pred && item || default` with `case(pred, item, default)`. `case` can also support multiple predicates (like a switch statement). We use that a bit in our workflows.

Tim-KH:
Replaced the ternaries in the GitHub Actions with the new case() function as described in #6783


pajlada (approved):

pajlada:

Issue #5703, opened 2024-11-08
Deal with Channel::isTwitchChannel


Mm2PL:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Currently client-side channels are "Twitch" channels for `isTwitchChannel` and have `Twitch` in the channel type enum. However they cannot be cast to `TwitchChannel`.

We should rework channel types, to resolve the `type >= Twitch && type < TwitchEnd` but can't be cast to `TwitchChannel` problem.
```cpp
    enum class Type {
        None, // Unused? empty?
        Direct, // Actually indirect channel, not sure where it's used though
        Twitch, // Normal Twitch channel like #pajlada
	    // can execute commands, but not send messages, there are no moderator privileges, no vips, no broadcasters,
        // no stream titles, ...
        TwitchWhispers, 
        TwitchWatching,
        TwitchMentions, // this could be made into a generic channel, there is no point of it being Twitch specifically, it's not a TwitchChannel anyway.
        TwitchLive,
        TwitchAutomod,
        TwitchEnd,
        Misc, // this is for local testing channels, can be left as is
    };
```

The minimal modifications here would be:
```cpp
    enum class Type {
        None,
        Direct,
        Twitch, // Normal Twitch channel like #pajlada
        Whispers, 
        Watching,
        Mentions,
        Live,
        Automod,
        Misc, // this is for local testing channels, can be left as is
    };
// make Channel::isTwitchChannel into:
bool Channel::isTwitchChannel() const {
	return this->type_ == Channel::Type::Twitch;
}
```

There was an idea where instead of having an enum for a channel we instead would have some kind of flags for what traits supports, ex:
```
enum ChannelFlags {
    PERMISSIONS = (1 << 0), // mods, vips, ...
    TWITCH_COMMANDS = (1<<1), // allow for /w and friends to run
    LOCAL = (1<<2),
    STREAM_TITLE = (1<<3),
    INDIRECT = (1<<4), // collects messages from more than one other channel, this could be used to add channel links at the start of messages (like /mentions)
};
```

Channel types would then be:
| name             | flags                                      | comment                                        |
| ----             | -----                                      | ---                                            |
[...]

PR #6889, opened 2026-03-31
feat: Add chatterino-embed as a reusable shared library

Nerixyz:
Adds `chatterino-embed` that exposes a minimal interface to embed Chatterino into another Qt app. Its interface is opaque - it doesn't expose internal types. Currently, you can only add a split.

It can be built and installed with:
```
cmake -B build -G Ninja -DBUILD_EMBED=ON -DBUILD_APP=OFF -DCHATTERINO_UPDATER=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja -C build chatterino-embed
cmake --install build --component chatterino-embed --prefix build/install
```

Draft because it still contains other changes and no CI.


src/singletons/WindowManager.cpp
```diff
@@ -777,34 +746,35 @@ IndirectChannel WindowManager::decodeChannel(const SplitDescriptor &descriptor)
...
+        case Channel::Type::TwitchAutomod:
+            return getApp()->getTwitch()->getAutomodChannel();
+        case Channel::Type::Misc:
+            return getApp()->getTwitch()->getChannelOrEmpty(
+                descriptor.channelName_);
+
+        case Channel::Type::None:
+        case Channel::Type::Direct:
+        case Channel::Type::TwitchEnd:
+            break;  // FIXME: remove these
```
Mm2PL: you probably want to link #5703 here

PR #6891, opened 2026-03-31
refactor: Remove operator!=

### LINKED ISSUE #4895: C++20 refactor: Investigate removing operator!=

4rneee:
Supersedes #5079 

Resolves #4895

The only place (except git submodules) with `operator!=` is now `lib/semver/include/semver/semver.hpp` but I didn't want to touch that.

### DIFF
--- a/src/controllers/highlights/HighlightResult.cpp
+++ b/src/controllers/highlights/HighlightResult.cpp
@@ -56,6 +56,11 @@ bool HighlightResult::operator==(const HighlightResult &other) const
     return true;
 }
 
+bool HighlightResult::operator!=(const HighlightResult &other) const
+{
+    return !(*this == other);
+}
+
 bool HighlightResult::empty() const
 {
     return !this->alert && !this->playSound &&
--- a/src/controllers/highlights/HighlightResult.cpp
+++ b/src/controllers/highlights/HighlightResult.cpp
@@ -56,11 +56,6 @@ bool HighlightResult::operator==(const HighlightResult &other) const
     return true;
 }
 
-bool HighlightResult::operator!=(const HighlightResult &other) const
-{
-    return !(*this == other);
-}
-
 bool HighlightResult::empty() const
 {
     return !this->alert && !this->playSound &&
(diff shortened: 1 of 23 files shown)


4rneee:
Looking over this again I think I probably shouldn't have removed the operator in `src/common/FlagsEnum.hpp` since it does not necessarily do the same as !(lhs == rhs), right?

Nerixyz:
> Looking over this again I think I probably shouldn't have removed the operator in `src/common/FlagsEnum.hpp` since it does not necessarily do the same as !(lhs == rhs), right?

No, that's fine. We have tests for this: https://github.com/Chatterino/chatterino2/blob/eec6ea9401d1cc8205926c86dd67d57a38b4c691/tests/src/FlagsEnum.cpp#L112-L127

...and they pass.

The reason for this is that given `t != u`, the compiler will search for `!=(T, U)`, `!=(U, T)`, `==(T, U)`, and `==(U, T)`. See [Call to an overloaded operator (4, rewritten candidates)](https://en.cppreference.com/w/cpp/language/overload_resolution.html#Call_to_an_overloaded_operator). I think this landed with `operator <=>` in C++ 20.

So you could also remove the overload `operator==(T lhs, FlagsEnum rhs)`, because we already have `operator==(FlagsEnum lhs, T rhs)`.

src/controllers/highlights/HighlightResult.cpp
```diff
@@ -56,11 +56,6 @@ bool HighlightResult::operator==(const HighlightResult &other) const
     return true;
 }
 
-bool HighlightResult::operator!=(const HighlightResult &other) const
```
pajlada, line 59: Could you leave this one?

pajlada (changes requested):

src/controllers/highlights/HighlightResult.cpp
4rneee: Sure, may I ask why?

pajlada: I have another PR in the works where leaving this in will make my merge conflicts easier

pajlada (approved):

PR #6892, opened 2026-04-01
Link dynamically against Boost.JSON when building with vcpkg on Windows

### LINKED ISSUE #6597: Compilation fails on Windows with vcpkg
### Checklist

### Describe your issue

Compiling Chatterino with vcpkg on Windows fails for about 10 months now (specifically since https://github.com/Chatterino/chatterino2/pull/5837). It fails, because vcpkg builds Boost.JSON as a shared library by default (sets `BOOST_ALL_DYN_LINK`), but we expect to compile the source ourselves. Linking then fails to find dllimports.

### Screenshots

_No response_

### OS and Chatterino Version

\-

invisibl3cat:
When building with vcpkg, we get the Boost.JSON library as a precompiled binary. Instead of trying to vendor Boost.JSON by building it ourselves, we can just link against the binary shipped by vcpkg. This should fix #6597

The CMake logic to decide whether to build or link against Boost.JSON is perhaps a bit awkward but I could not think of anything better.

I have checked that the the code builds and runs on Windows-vcpkg and Linux with otherwise standard build flags.

(This is my first time touching Chatterino, sorry if I'm making a mess...)


pajlada:
The logic looks sane, but I believe it needs to be applied to the usage of boost-json inside the `twitch-eventsub-ws` library in our repo too.
Once you've done that I'll give this a proper try on Windows.
Thanks for looking into this!

invisibl3cat:
Sure, I'll take a look at `twitch-eventsub-ws` as well! But is there a reason why not having this fix in `twitch-eventsub-ws` would not lead to a build failure?

src/controllers/plugins/api/JSONParse.cpp
```diff
@@ -415,6 +415,10 @@ int jsonParse(lua_State *L)
 
 }  // namespace chatterino::lua::api
 
-#    include <boost/json/src.hpp>
+#    ifndef BOOST_ALL_DYN_LINK
+#        include <boost/json/src.hpp>
+#    else
+#        include <boost/json.hpp>
```
Nerixyz: You don't need to include anything if Boost.JSON is already built.

Nerixyz (commented):
Because we use Boost.JSON for both the eventsub library and the main one, it might be easier to make a separate CMake target that links to Boost.JSON, so we'd have something like

```mermaid
flowchart TD
    bjp(BoostJsonProvider) -->|links to| bj(Boost.JSON)
    tw(twitch-eventsub-ws) -->|links to| bjp
    c2(chatterino-lib) -->|links to| tw
    c2(chatterino-lib) -->|links to| bjp
```

And in BoostJsonProvider, we decide if we need to build it (because it was header-only) or not.

src/controllers/plugins/api/JSONParse.cpp
invisibl3cat: I thought so too but not including `<boost/json.hpp>` left me with 2 unresolved externals. At least that is the case on my setup with MSVS 2026.

invisibl3cat:
> Because we use Boost.JSON for both the eventsub library and the main one, it might be easier to make a separate CMake target that links to Boost.JSON, so we'd have something like
>
> And in BoostJsonProvider, we decide if we need to build it (because it was header-only) or not.

Okay. I will try to address this in `twitch-eventsub-ws` first and look into providing a common target then.

pajlada:
You can skip the common target part, if we're only using this in two places this will add unnecessary complexity.

When I intend to test this, I will test this with `-DCHATTERINO_PLUGINS=On` and `-DCHATTERINO_PLUGINS=Off`

src/CMakeLists.txt
```diff
@@ -951,6 +951,9 @@ target_link_libraries(${LIBRARY_PROJECT}
         )
 if (CHATTERINO_PLUGINS)
     target_link_libraries(${LIBRARY_PROJECT} PUBLIC lua sol2::sol2)
+    if (VCPKG_TARGET_TRIPLET)
+        target_link_libraries(${LIBRARY_PROJECT} PUBLIC Boost::json)
+    endif()
```
Nerixyz: Not necessary. We build Boost.JSON already.

CMakeLists.txt
```diff
@@ -162,7 +162,11 @@ if (WIN32)
 endif ()
 
 # Find boost on the system
-find_package(Boost REQUIRED CONFIG OPTIONAL_COMPONENTS headers)
+set(BOOST_REQUIRED_COMPONENTS "")
+if (VCPKG_TARGET_TRIPLET AND CHATTERINO_PLUGINS)
+    list(APPEND BOOST_REQUIRED_COMPONENTS json)
+endif ()
+find_package(Boost REQUIRED CONFIG COMPONENTS ${BOOST_REQUIRED_COMPONENTS} OPTIONAL_COMPONENTS headers)
```
Nerixyz, line 169: Same here.

src/controllers/plugins/api/JSONParse.cpp
Nerixyz: I see, including `boost/json/basic_parser_impl.hpp` at the top fixes this. Then we don't need this bottom block at all (not for conan or vcpkg).

invisibl3cat:
I do not think that there is a real issue with `twitch-eventsub-ws` with regard to Boost.JSON. Since `twitch-eventsub-ws` is only built as a static library, it does not run into the `dllexport` vs. `dllimport` mismatch that breaks linking of `chatterino-lib`. In case of a static library these decorators should be ignored by the compiler. However, when we know that we are linking against Boost.JSON DLL, we do not have to duplicate the Boost.JSON implementation in `twitch-eventsub-ws`. I have updated the PR to do this small optimization.

CMakeLists.txt
invisibl3cat: If I remove this directive I get link errors.

lib/twitch-eventsub-ws/src/json.cpp
```diff
@@ -1 +1,5 @@
-#include <boost/json/src.hpp>
+#ifdef BOOST_JSON_FROM_DLL
+#    include <boost/json.hpp>
```
pajlada: In the case of the `twitch-eventsub-ws` library, we don't need an extra include since this is the only contents of the source file

pajlada (changes requested):

lib/twitch-eventsub-ws/src/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ target_compile_definitions(${PROJECT_NAME} PUBLIC
     $<$<BOOL:${MSVC}>:BOOST_JSON_NO_LIB>
     $<$<BOOL:${MSVC}>:BOOST_CONTAINER_NO_LIB>
     QT_NO_KEYWORDS
+    $<$<BOOL:${VCPKG_TARGET_TRIPLET}>:BOOST_JSON_FROM_DLL>
```
Nerixyz: `BOOST_JSON_FROM_DLL` should have a different name. If it starts with `BOOST_JSON` it sounds like it's something that affects the library itself, not how we use it.

You also shouldn't define `BOOST_JSON_NO_LIB` in that case.

lib/twitch-eventsub-ws/src/json.cpp
invisibl3cat: You mean doing just this?

```
#ifndef BOOST_JSON_FROM_DLL
#    include <boost/json/src.hpp>
#endif // BOOST_JSON_FROM_DLL
```
I'll check that that it doesn't cause any weird linking issues... though it should work.

lib/twitch-eventsub-ws/src/CMakeLists.txt
invisibl3cat: Would `TWEVENTSUB_EXPECT_BOOST_JSON_FROM_DLL` be a more acceptable name for the defn?

Nerixyz: Yes - alternatively prefixed with `CHATTERINO_`

lib/twitch-eventsub-ws/src/json.cpp
pajlada: Yep

pajlada:
Overall, this change looks good to me. Will spend some time testing it out and report back.
Thanks for your quick turnaround to review comments!

pajlada (approved):
Tested a vcpkg build with and without plugins and it works fine with your change. Thank you!

pajlada:

PR #6893, opened 2026-04-01
Add CI job that builds with latest Qt versions

### LINKED ISSUE #6879: Add CI job that builds with the latest RC/beta Qt version
It would be nice to ensure changes we make continue to build on the latest release candidate or beta versions of Qt.

There can always be behavioural changes that we wouldn't catch here of course, but knowing it builds is still helpful imo.

vGregorGit:
## Summary
Added a new CI workflow that automatically detects and builds with the latest Qt beta/RC versions to ensure Chatterino remains compatible with newer Qt releases as requested in issue #6879.
 
## Changes
- Added [.github/workflows/build-qt-beta.yml](cci:7://file:///c:/Users/vgregor/Desktop/Projekte%202024/chatterino2/.github/workflows/build-qt-beta.yml:0:0-0:0) workflow
- Automatically detects latest Qt versions from Qt download servers
- Tests with detected versions on Ubuntu, Windows, and macOS

Fixes #6879

### DIFF
--- a/.github/workflows/build-qt-beta.yml
+++ b/.github/workflows/build-qt-beta.yml
@@ -22,25 +22,62 @@ env:
   CONAN_VERSION: 2.11.0
 
 jobs:
+  detect-latest-qt:
+    name: "Detect Latest Qt Versions"
+    runs-on: ubuntu-latest
+    outputs:
+      qt-versions: ${{ steps.detect.outputs.qt-versions }}
+      latest-beta: ${{ steps.detect.outputs.latest-beta }}
+      latest-stable: ${{ steps.detect.outputs.latest-stable }}
+    steps:
+      - name: Detect available Qt versions
+        id: detect
+        run: |
+          # Try to detect latest Qt versions from Qt's download servers
+          QT_VERSIONS="[]"
+          LATEST_BETA=""
+          LATEST_STABLE="6.11.0"  # fallback
+          
+          # Check for Qt 6.12 beta/RC versions first
+          for version in "6.12.0-rc1" "6.12.0-beta1" "6.12.0-beta2"; do
+            echo "Checking Qt $version..."
+            if curl -s -f -L "https://download.qt.io/development_releases/qt/${version%.*}/$version/qt-opensource-linux-x64-$version.run" > /dev/null 2>&1; then
+              echo "✓ Found Qt $version (beta/RC)"
+              QT_VERSIONS=$(echo "$QT_VERSIONS" | jq ". + [\"$version\"]")
+              if [ -z "$LATEST_BETA" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/build-qt-beta.yml
```diff
@@ -0,0 +1,255 @@
+---
+name: Build with Qt Beta/RC
+
+on:
+  push:
+    branches:
+      - master
+      - "bugfix-release/*"
+      - "release/*"
```
pajlada: No need to build in these two branches

test-workflows.ps1
```diff
@@ -0,0 +1,27 @@
+# Test script to validate CI workflow syntax
```
pajlada, line 1: No need to commit this test file

.github/workflows/build-qt-beta.yml
```diff
@@ -0,0 +1,255 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      qt-versions: ${{ steps.detect.outputs.qt-versions }}
+      latest-beta: ${{ steps.detect.outputs.latest-beta }}
+      latest-stable: ${{ steps.detect.outputs.latest-stable }}
+    steps:
+      - name: Detect available Qt versions
+        id: detect
+        run: |
+          # Try to detect latest Qt versions from Qt's download servers
```
pajlada: Are we able to utilize the `aqt` tool to see if there's a beta or RC version available instead? Having to manually update the beta versions we want to build with seems tedious

.github/workflows/build-qt-beta.yml
```diff
@@ -0,0 +1,255 @@
...
+
+      - name: Determine build type
+        id: build-type
+        shell: bash
+        run: |
+          if git describe --exact-match --match 'v*'; then
+            echo "NIGHTLY_BUILD=Off" >> "$GITHUB_OUTPUT"
+          else
+            echo "NIGHTLY_BUILD=On" >> "$GITHUB_OUTPUT"
+          fi
```
pajlada: You can remove this part of the build process, leave `CHATTERINO_NIGHTLY_BUILD` undefined

pajlada (changes requested):
Preferably the build process / workflow is smaller than this - some of what we include in our regular build process is only really relevant for builds that will be published, not if we're only checking if things build

pajlada (commented):
I'd like to wait with merging this in / testing a bit more until there's an actual beta / rc version available to confirm that it can be found with this action.

In the meantime, there's some build complications for the Windows & Ubuntu build you can look at. If building on multiple platforms is too complicated, I would be fine if this action _only_ builds on something like mac or windows or ubuntu to start.

Issue #5219, opened 2024-03-01
Muting specific highlight sounds for specific channels


Mm2PL:
### Discussed in https://github.com/Chatterino/chatterino2/discussions/2354

<div type='discussions-op-text'>

<sup>Originally posted by **MPH2210** January 10, 2021</sup>
So, right now you can either mute all the notifications for a channel or none.
The problem for me is: My name is max, and in legit every channel people use "max" to shorten "maximum", so I would get pinged every 10 seconds or so.

If you could open up a menu through the three dots on a channel, where you choose which highlights you want to hear from only that channel and which not, would be super cool.

Hope it was understandable what I meant.</div>

Possibly this could be made as a plugin (with impl of #4619). Not sure if that's a good idea though.

x3rt:
Would love to see this feature!
I often get referred to as 'x3' in channels I'm active in, but as you can probably guess that causes issues in other channels

ceriath:
this would also be very helpful for moderation, since you could have a highlight like `mods?` only in channels you are a mod in.

Skrufy:
dear god please add this shit my username is "shitty" and channels like Kai cenat people love to spam "hahahahahah thats so SHITTY!!!!!!!!!!!!!!!!!!!!!!" this shit is so annoying this should already be a feature

Nerixyz:
This is a duplicate of #2401.

PR #6896, opened 2026-04-02
Channel-specific highlights

Kehnanmind:
Addresses #2401, #5219 

This PR introduces the ability to filter highlight phrases by channel, allowing users to define where specific highlights should apply.
There are many different ways this feature could be implemented. The way I did is by adding a `Filter Channels` column in the already existing highlights table:

image

**Motivation**
Currently, highlight phrases are global - they trigger in every channel. This can be problematic for users who follow many channels and want the highlights to behave differently by channel.

This feature allows:
- Highlight/ping custom phrases only in specific channels;
- Reduce the amount of pings you get in unrelated channels;
- Create more flexible setups (e.g. ping in one channel, highlight-only in another);
- This can be especially useful for users that have common words in their names (e.g. username is fish123, but _only want to get pinged_ when "fish" is typed in some small channel);
- Leaving the field empty makes it global (same functionality as it is now);

**Implementation**
I tried to build this on the existing highlight system rather than introducing a new one:

- Added channel filtering to `HighlightPhrase` by introducing a new `QString channelNames_` field;
- Added `appliesToChannel(...) `helper that checks whether the current message’s channel matches (or allows everything if empty);
- Updated `highlightPhraseCheck(...) `to call `appliesToChannel(args.channelName)` before doing the actual match, so highlights are ignored early if they’re not meant for that channel;
- Extended `MessageParseArgs` with `QString channelName`;
- Passed the extra channel context to existing structures;
- Added a new “Filter channels” column in the UI;

Currently, this works for:
- Custom user-defined highlight phrases
- "**Your Username (automatic)"** built-in highlight

The "Filter Channels" column is  disabled for other built-in highlights (whispers, subscriptions, first messages, etc.)

If this approach doesn’t quite match the direction you’re aiming for, that’s totally fine. I still think channel-specific highlights/pings would be a really useful feature to have, so I wanted to share my take on it.

### DIFF
--- a/src/controllers/highlights/HighlightCheck.hpp
+++ b/src/controllers/highlights/HighlightCheck.hpp
@@ -25,9 +25,9 @@ using MessageFlags = FlagsEnum<MessageFlag>;
 struct HighlightCheck {
     using Checker = std::function<std::optional<HighlightResult>(
         const MessageParseArgs &args,
-        const std::vector<TwitchBadge> &twitchBadges, const QString &senderName,
-        const QString &originalMessage, const MessageFlags &messageFlags,
-        bool self)>;
+        const std::vector<TwitchBadge> &twitchBadges,
+        const QString &senderName, const QString &originalMessage,
+        const MessageFlags &messageFlags, bool self)>;
     Checker cb;
 };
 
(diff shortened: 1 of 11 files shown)


Nerixyz:
#6695 already intends to refactor highlights to support more complex conditions, channels included.

PR #6905, opened 2026-04-04
automod: dedupe hit terms

### LINKED ISSUE #6006: De-duping automod.message.hold matched terms

pajlada:
Fixes #6006


tests/snapshots/EventSub/automod-message-hold/blocked-terms-4.json
```diff
@@ -146,8 +146,8 @@
                         "3",
                         "blocked",
                         "terms",
-                        "\"🍋‍🟩\",",
                         "\"blockedterm\",",
+                        "\"🍋‍🟩\",",
```
pajlada, line 150: This order is not the same now since we use a set. could use boost multiindex or a vector that we dedupe "manually" but I don't think this matters

PR #6906, opened 2026-04-04
refactor: Move split (de)serialization to WindowDescriptors

Nerixyz:
This moves the split and container (de)serialization code to `WindowDescriptors.(c/h)pp`. The main intention is to have both parts side-by-side. Before, we had the serialization in `WindowManager` and the deserialization in `WindowDescriptors`. When changing fields, you had to update both.


src/common/WindowDescriptors.cpp
```diff
@@ -28,30 +35,190 @@ QJsonArray loadWindowArray(const QString &settingsPath)
...
+    }
+    return descriptor;
+}
+
+SplitDescriptor SplitDescriptor::fromSplit(const Split &split)
+{
+    SplitDescriptor descriptor;
+    descriptor.type_ =
+        qmagicenum::enumNameString(split.getIndirectChannel().getType());
+    descriptor.channelName_ = split.getChannel()->getName();
```
Mm2PL: NOTE: This is always encoded now, compared to being conditionally encoded.

Mm2PL (commented):
I don't see anything obviously wrong at first glance. The diff is kinda ass to read having mixed move and changes. I've not tested this xd

Mm2PL:
I'd love if you rebased this on top of 7255c65 or otherwise removed those changes from the total diff of this PR.

src/widgets/splits/SplitContainer.cpp
```diff
@@ -841,45 +844,16 @@ void SplitContainer::popup()
...
-            return "live";
-        case Type::TwitchAutomod:
-            return "automod";
-        case Type::Misc:
-            return "misc";
-    }
-}
-
 NodeDescriptor SplitContainer::buildDescriptorRecursively(
     const Node *currentNode) const
```
Mm2PL: Interesting that we have WindowDescriptors' buildDescriptorRecursively and this one. Do we need to have both?

Nerixyz: We don't. I removed the one in `WindowDescriptors`.

Nerixyz:
I'll do more small changes, but using this as a reference for what's left.

Nerixyz:
Closing in favor of the individual PRs.

PR #6908, opened 2026-04-05
fix: removed EventSub opt in settings

### LINKED ISSUE #6899: Remove EventSub opt in setting
the "Enable experimental Twitch EventSub support" setting should be removed and users should always connect to Twitch EventSub.

nemida:
Removed UI Checkbox
Removed the definition in Settings.hpp
Made it so that it connects to Twitch EventSub unconditionally.
Fixes #6899

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -130,14 +130,7 @@ SeventvEventAPI *makeSeventvEventAPI(Settings &settings)
 
 eventsub::IController *makeEventSubController(Settings &settings)
 {
-    bool enabled = settings.enableExperimentalEventSub;
-
-    if (enabled)
-    {
-        return new eventsub::Controller();
-    }
-
-    return new eventsub::DummyController();
+    return new eventsub::Controller();
 }
 
 const QString TWITCH_PUBSUB_URL = "wss://pubsub-edge.twitch.tv";
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -128,11 +128,6 @@ SeventvEventAPI *makeSeventvEventAPI(Settings &settings)
     return nullptr;
 }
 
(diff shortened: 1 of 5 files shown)


src/singletons/Settings.hpp
```diff
@@ -803,12 +803,6 @@ class Settings
         false,
     };
 
-    // Advanced
```
Wissididom, line 806: Shouldn't this comment stay?

src/Application.cpp
```diff
@@ -130,14 +130,7 @@ SeventvEventAPI *makeSeventvEventAPI(Settings &settings)
...
 eventsub::IController *makeEventSubController(Settings &settings)
 {
-    bool enabled = settings.enableExperimentalEventSub;
-
-    if (enabled)
-    {
-        return new eventsub::Controller();
-    }
-
-    return new eventsub::DummyController();
```
Wissididom, line 140: Would it cause any problem removing the DummyController alltogether if it isn't being used somewhere else?

src/singletons/Settings.hpp
nemida: Yes, my bad.

src/Application.cpp
nemida: I don't think so. I'll remove it in the next commit.

src/Application.cpp
```diff
@@ -130,14 +130,7 @@ SeventvEventAPI *makeSeventvEventAPI(Settings &settings)
...
+    return new eventsub::Controller();
```
pajlada: You can remove the `makeEventSubController` too and create it inside the Application constructor instead, similar to `Logging` and `Fonts` above it

pajlada (approved):

pajlada:

PR #6909, opened 2026-04-05
 Allow plugins to clone existing elements 

Mm2PL:
You can do stuff like:

```lua
c = c2.Channel.by_name('pajlada')
randomel = c:message_snapshot(1)[1]:elements()[3]
c:add_message(c2.Message.new {elements={randomel}})
```
This will copy the first badge of the user from the last message (if its a normal privmsg ofc).
A slightly more complicated plugin that needs functionality like this: https://github.com/Mm2PL/c2-formatting


src/messages/MessageElement.hpp
```diff
@@ -196,6 +196,9 @@ class MessageElement
     /// member.
     virtual std::string_view type() const = 0;
 
+    /// Creates a new identical message element. Used for Lua plugins.
```
Nerixyz:
```suggestion
    /// Creates a new identical message element.
```

Quite useful for other things as well.

src/messages/MessageElement.cpp
```diff
@@ -158,6 +155,15 @@ std::string_view ImageElement::type() const
     return std::remove_pointer_t<decltype(this)>::TYPE;
 }
 
+std::unique_ptr<MessageElement> ImageElement::clone() const
+{
+    auto im = std::make_unique<ImageElement>(this->image_, this->getFlags());
+    im->setLink(this->getLink());
+    im->setTooltip(this->getTooltip());
+    im->setTrailingSpace(this->hasTrailingSpace());
```
Nerixyz: This has to be done for all elements, and it's easy to miss one field (e.g. when updating some functionality). In Chatterino7, I added a `cloneFrom` to `MessageElement` that does this: https://github.com/SevenTV/chatterino7/blob/c37a982fc79742003e02e6674b1a2bfa89502aea/src/messages/MessageElement.cpp#L111-L117. It's called in every `::clone()`.

src/messages/MessageElement.cpp
```diff
@@ -872,6 +946,17 @@ std::string_view TextElement::type() const
 {
     return std::remove_pointer_t<decltype(this)>::TYPE;
 }
+std::unique_ptr<MessageElement> TextElement::clone() const
+{
+    auto text = this->words_.join(' ');
```
Nerixyz: We can reuse the words (utilize the COW of QStringList), because they're already split. You can add a private/protected constructor that takes a list of words and pass `this->words_` along.

Same for `SingleLineTextElement`, `LinkElement`, and `MentionElement`.

src/controllers/plugins/api/Message.cpp
```diff
@@ -852,16 +746,101 @@ void createUserType(sol::table &c2)
...
+                auto el = elementFromTable(tbl);
+                if (el)
+                {
+                    msg->elements.emplace_back(std::move(el));
+                }
+            },
+            [](Message *msg, ElementRef &element) {
+                checkWritable(msg);
+                msg->elements.emplace_back(
+                    std::move(element.constElement()->clone()));
```
Nerixyz: Shouldn't need a `std::move` here

Nerixyz (commented):
In the future, we should also allow plugins to clone whole messages that are then writable again. That should be fairly simple (e.g. https://github.com/SevenTV/chatterino7/blob/c37a982fc79742003e02e6674b1a2bfa89502aea/src/messages/Message.cpp#L107-L135).

As for testing, I'm not sure where the best place is. We have a lot of `Message`s in the eventsub and IRC tests.

src/messages/MessageElement.cpp
```diff
@@ -1141,6 +1276,17 @@ std::string_view MentionElement::type() const
     return std::remove_pointer_t<decltype(this)>::TYPE;
 }
 
+std::unique_ptr<MessageElement> MentionElement::clone() const
+{
+    auto elem = std::make_unique<MentionElement>(
+        this->words_.join(' '), this->userLoginName_, this->fallbackColor_,
```
Nerixyz: Could also use a `QStringList` constructor

src/messages/MessageElement.cpp
```diff
@@ -1141,6 +1276,17 @@ std::string_view MentionElement::type() const
...
+        this->userColor_);
+
+    elem->setTooltip(this->getTooltip());
+    elem->setTrailingSpace(this->hasTrailingSpace());
```
Nerixyz, line 1299:
```suggestion
    elem->cloneFrom(*this);
```

Though we know that the link and flags will be the same. Similar for `LinebreakElement` and others.

Nerixyz (commented):
The TypeScript definitions should be updated and possibly the Markdown docs.

Some tests would be great. Maybe by re-using the inputs we have for the IRC messages ([here](https://github.com/Chatterino/chatterino2/blob/0549d16896e5a0727d53efbda73f7fa55cb6e883/tests/src/IrcMessageHandler.cpp#L596))?

It doesn't have to check against the ground truth even. It's enough if we get a bunch of elements and check that they're still equal after cloning:

```cpp
// not tested
TEST_P(TestIrcMessageHandlerP, CloneElements)
{
    auto channel = makeMockTwitchChannel(u"pajlada"_s, *snapshot);

    VectorMessageSink sink;

    for (auto prevInput : snapshot->param("prevMessages").toArray())
    {
        auto *ircMessage = Communi::IrcMessage::fromData(
            prevInput.toString().toUtf8(), nullptr);
        ASSERT_NE(ircMessage, nullptr);
        IrcMessageHandler::parseMessageInto(ircMessage, sink, channel.get());
        delete ircMessage;
    }

    auto *ircMessage =
        Communi::IrcMessage::fromData(snapshot->inputUtf8(), nullptr);
    ASSERT_NE(ircMessage, nullptr);
    IrcMessageHandler::parseMessageInto(ircMessage, sink, channel.get());
    delete ircMessage;

    for (const auto &message : sink.messages())
    {
        for (const auto&original : message->elements)
        {
            auto originalObj = original->toJson();
            auto clonedObj = original->clone()->toJson();
            ASSERT_EQ(originalObj, clonedObj) 
                << "\noriginal:\n" << QJsonDocument(originalObj).toJson() 
                << "\ncloned:\n" << QJsonDocument(clonedObj).toJson();
        }
    }
}
```

src/messages/MessageElement.cpp
Mm2PL: This would call `MessageElement::setLink` which is an `assert(false)`.

tests/src/IrcMessageHandler.cpp
```diff
@@ -651,3 +651,38 @@ TEST(TestIrcMessageHandlerP, Integrity)
 {
     ASSERT_FALSE(UPDATE_SNAPSHOTS);  // make sure fixtures are actually tested
 }
+
+TEST_P(TestIrcMessageHandlerP, CloneElements)
+{
+    auto channel = makeMockTwitchChannel(u"pajlada"_s, *snapshot);
```
Nerixyz: I think we should use explicit `this->` here, because any change to this will raise a warning again.

Nerixyz (approved):

src/messages/MessageElement.hpp
```diff
@@ -196,10 +196,15 @@ class MessageElement
...
     virtual std::string_view type() const = 0;
 
+    /// Creates a new identical message element.
+    virtual std::unique_ptr<MessageElement> clone() const = 0;
+
 protected:
     MessageElement(MessageElementFlags flags);
     bool trailingSpace = true;
 
+    void cloneFrom(const MessageElement &source);
```
pajlada, line 207: comment

src/messages/MessageElement.hpp
```diff
@@ -264,19 +272,30 @@ class CircularImageElement : public MessageElement
...
+    };
+
 public:
     static constexpr std::string_view TYPE = "text";
 
     TextElement(const QString &text, MessageElementFlags flags,
                 const MessageColor &color = MessageColor::Text,
                 FontStyle style = FontStyle::ChatMedium);
+
+    TextElement(TextElement::CloneConstructorTag, QStringList words,
```
pajlada, line 288: add comment to all of these "public but actually private" constructors, just a one-liner similar to "for use in the internal clone method only"

src/controllers/plugins/api/Message.cpp
```diff
@@ -852,16 +790,100 @@ void createUserType(sol::table &c2)
...
+                auto el = elementFromTable(tbl);
+                if (el)
+                {
+                    msg->elements.emplace_back(std::move(el));
+                }
+            },
+            [](Message *msg, ElementRef &element) {
+                checkWritable(msg);
+                msg->elements.emplace_back(element.constElement()->clone());
+            }));
```
pajlada, line 809: can you add comments to these two overloads

pajlada (changes requested):

pajlada (approved):

pajlada (approved):

PR #6910, opened 2026-04-06
fix(macos): adjust default font size

Leafyzito:
~~I'm so used to the old scaling that I immediately thought of adding this, and probably other people will like having the option to go back too.~~

Changing the default font size to 16 was enough to get things back to normal.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,7 +191,7 @@ jobs:
 
       - name: Install Qt6
         if: startsWith(matrix.qt-version, '6.')
-        uses: jurplel/install-qt-action@v4.3.0
+        uses: jurplel/install-qt-action@v4.3.1
         with:
           cache: true
           cache-key-prefix: ${{ runner.os }}-QtCache-${{ matrix.qt-version }}-v2
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Install Qt6
         if: startsWith(matrix.qt-version, '6.')
-        uses: jurplel/install-qt-action@v4.3.0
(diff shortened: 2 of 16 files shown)


pajlada:
What is your monitor setup and what's the difference between legacy & non-legacy scaling for you?

Leafyzito:
Lately I've been using a 13.3-inch display Macbook Air M1 (macOS 26.4, build 25E246)

| Old scaling | New scaling |
| --- | --- |
| Screenshot 2026-04-06 at 5 20 29 PM | Screenshot 2026-04-06 at 5 21 19 PM |

The text in general got smaller and makes badges and emotes look too big and just weird

pajlada:
Have you tried the build from your PR? There were other mac-specific changes I made that are not reverted with legacy scaling

Nerixyz:
> The text in general got smaller and makes badges and emotes look too big and just weird

I think it would be better if we could adjust the font size correctly instead of relying on that flag.

Leafyzito:
> Have you tried the build from your PR? There were other mac-specific changes I made that are not reverted with legacy scaling

Yes, the screenshots were taken from my fork version, but the build from this PR behaves the same way. And everything looks fine to me

Leafyzito:
> I think it would be better if we could adjust the font size correctly instead of relying on that flag.

Setting the font size to 16 on the new scaling seems to put things back to normal. I guess this would fix the issue

Nerixyz:
> Setting the font size to 16 on the new scaling seems to put things back to normal. I guess this would fix the issue

This does sound large. I wonder if that also looks good on pajlada's mac.

If that's too large there, then we might need the flag on macOS. Though I'd still wonder why _we_ need it and other Qt apps don't.

pajlada:
fresh config on my mac mini, no scaling

image

default chat text size looks good compared to right-click menu

pajlada (approved):

PR #6913, opened 2026-04-07
Fix vertical text misalignment

### LINKED ISSUE #6616: Text can be vertically misaligned in an element
### Checklist

### Describe your issue

Sometimes depending on the characters used text can be rendered lower than is visible. This happens in most fonts but some don't have it, Noto Sans Arabic for example does not have this behavior.

### Screenshots

Image
font used: FiraCode Nerd Font, 11pt, zoom 0.9x

### OS and Chatterino Version

Chatterino 2.5.4 DEBUG (commit 041674b7f modified) built with Qt 6.10.1 Running on Arch Linux, kernel: 6.17.9-arch1-1

invisibl3cat:
This is an attempt to fix vertical text misalignment of chat messaged caused by some glyphs that cause the bounding box of the text they are a part of to have a different height. This might fix #6616


src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -451,12 +451,30 @@ void TextLayoutElement::paint(QPainter &painter,
...
+    // on its line.
+    // A simple way to compensate for this is to shift the top
+    // of the bounding rectangle we draw the text into back to
+    // the position where it would have been for a text composed only
+    // of the "usual" glyphs.
+    auto metrics = app->getFonts()->getFontMetrics(this->style_, this->scale_);
+
+    auto expectedBrectHeight =
+        app->getFonts()->getExpectedBrectHeight(this->style_, this->scale_);
+    auto actualBrectHeight = metrics.boundingRect(text).y();
```
Nerixyz: This should be done when laying out the text, not when paining.

Nerixyz:
Have you tried using `Qt::AlignVCenter` or `Qt::AlignBaseline`?

src/messages/layouts/MessageLayoutElement.cpp
invisibl3cat: I can update the logic but doing this in `getTextLayoutElement()` - which is where I assume you'd like this logic to be - is rather awkward. For example, you do not get to know if there is an RTL_EMBED character prepended to the string until `paint()` gets called. We need to know about this modification before we apply the text position correction, otherwise it would not fix up some words correctly.

Or did you mean to put this logic in a different place entirely?

invisibl3cat:
> Have you tried using `Qt::AlignVCenter` or `Qt::AlignBaseline`?

I have and neither of those work correctly. When Qt::TopAlign does not work, I don't think that any other vertical alignment modes can work. As far as I gather, the issue here is that the height of the bounding rectangle is inconsistent between individual `TextLayoutElement`s in a line and in that case the standard vertical alignment flags won't work.

lib/twitch-eventsub-ws/lib/date
pajlada, line 1: undo

tools/crash-handler
pajlada, line 1: undo

src/messages/layouts/MessageLayoutElement.cpp
```diff
@@ -504,6 +527,19 @@ int TextLayoutElement::getMouseOverIndex(QPointF abs) const
     return this->getSelectionIndexCount() - (this->hasTrailingSpace() ? 1 : 0);
 }
 
+void TextLayoutElement::setClipRegion()
```
pajlada: rename `updateClipRegion` instead

pajlada:
The clipping feels off (see the supibot message in the right-most chat). Default zoom, Cantarell 11pt on Arch Linux.

The second part of the links in the left-most chat also disappear

unsorted_20260412_12h08m35s
unsorted_20260412_12h08m36s

pajlada (changes requested):

invisibl3cat:
It turns out that my initial approach to this problem was rather wrong. I've made two major changes that feel a lot better:

- Instead of rendering the `TextLayoutElement` into a rectangle, I'm giving the `QPainter` just the initial *point*. Painting text into a rectangle would be useful if wanted the `QPainter` to take care of multiple lines and alignment for us. We probably don't want this because we're already laying out the individual blocks of the message ourselves. For example, I think that the clipped URLs were caused by QPainter rendering the part after the "/" on a new line which then got painted over.

- `QFontMetrics` has this funny behavior where it happily computes metrics for a font that is *not available* on the system. We then end up in a situation where we compute the layout believing that it will be rendered with the font for which we computed the metrics. In reality, `QPainter` will replace the font with the closest match and mess everything up. We need to fix up the fonts early to avoid this.

This should probably use some more testing on wider variety of systems to make sure that there are no lurking edge cases.

src/singletons/Fonts.cpp
```diff
@@ -256,7 +256,21 @@ Fonts::FontData Fonts::createFontData(FontStyle type, float scale)
...
+    // QFontMetrics does not care that it computed metrics
+    // for a font that will not be used for the actual rendering.
+    // By fixing up the font early we can prevent some rendering
+    // issues caused by the mismatch between what QFontMetrics
+    // computed and what got painted on the screen.
+
+    QWidget w;
+    w.setFont(font);
+
+    return w.font();
```
Nerixyz, line 273: If I read it correctly, this is equivalent to `font.resolve(QApplication::font())`. Haven't tested it though.

invisibl3cat: I'm not so sure. The docs say that

---
[QFont](https://doc.qt.io/qt-6/qfont.html) QFont::resolve(const [QFont](https://doc.qt.io/qt-6/qfont.html) &other) const

Returns a new [QFont](https://doc.qt.io/qt-6/qfont.html) that has attributes copied from other that have not been previously set on this font.

---

I haven't gone down the font handling rabbit hole deep enough but I think that a font can be setting all of the attributes regardless of whether it is available or not. Getting Qt to pick the closest match via a paint device seems to be the recommended way how to do this. It feels quite clumsy but we should have to do it only once for each font.

src/messages/layouts/MessageLayoutElement.cpp
Nerixyz: No longer relevant.

src/singletons/Fonts.cpp
Nerixyz: I was only going through the Qt code:

**`QWidget::setFont`**: https://github.com/qt/qtbase/blob/cd8a057fcf050f1fe7df1b59b9facb135d1518ff/src/widgets/kernel/qwidget.cpp#L4771-L4779
1. I don't think the `setAttribute` is relevant for us.
2. `d->naturalWidgetFont(...)`: https://github.com/qt/qtbase/blob/cd8a057fcf050f1fe7df1b59b9facb135d1518ff/src/widgets/kernel/qwidget.cpp#L4794
	1. Grabs [`QApplication::font(widget)`](https://github.com/qt/qtbase/blob/cd8a057fcf050f1fe7df1b59b9facb135d1518ff/src/widgets/kernel/qapplication.cpp#L1288) which should return `QApplication::font()`, because the widget doesn't have a parent.
	2. The other parts aren't relevant, because we don't have a parent or proxy widget.
	3. Calls `naturalFont.setResolveMask(0)` and returns that
3. Calls `QFont::resolve(naturalFont)`
4. Sets that font - though this is a bit more involved: https://github.com/qt/qtbase/blob/cd8a057fcf050f1fe7df1b59b9facb135d1518ff/src/widgets/kernel/qwidget.cpp#L4871

Though creating a temporary widget there should be fine - we do this once.

pajlada (commented):
Neat, this works well for my default-initialized font now.

I am questioning if we still need the clip region addition you've made - for what scenario is this necessary?

pajlada:
for reference, 3 screenshots of 2.5.5, your pr (labeled new), and your pr without the clipping (labeled new without clip)

unsorted_20260418_12h04m01s




unsorted_20260418_12h04m02s





unsorted_20260418_12h04m16s

invisibl3cat:
The clip is useful to prevent "bleedover" of some unusually tall or oddly aligned glyphs to the adjacent lines. See the attached screenshots:

With clip:
with_clip


Without clip:
without_clip

Without clipping, the Arabic characters at the end of the test message protrude to the line above, making the text a bit more difficult to read. FWIW I could only replicate this behavior with monospace fonts.

pajlada:
I think I'd rather leave the clipping out, and potentially discuss it further as a separate change. I don't think it's as clear of a win as the vertical text misalignment you've fixed with the other changes

invisibl3cat:
> I think I'd rather leave the clipping out, and potentially discuss it further as a separate change. I don't think it's as clear of a win as the vertical text misalignment you've fixed with the other changes

Fair enough, the clipping is now gone.

pajlada (approved):

PR #6916, opened 2026-04-10
Add setting to specify custom path for the native messaging manifest

### LINKED ISSUE #6758: Waterfox on Linux "watching" page not working.
### Checklist

### Describe your issue

Whenever I attempt to use the "watching" tab while on Waterfox (installed using [this method](https://www.waterfox.com/support/install-waterfox-linux/#install-waterfox-from-waterfox-builds-for-advanced-users), not through flatpak), it does not update the page successfully.
Have confirmed that it works normally on base Firefox. Chatterino extension is installed and is definitely working (if I select "replace Twitch page" it shows the relevant message about it only being on Windows).

Waterfox 6.6.8
Chatterino Native Host extension on Waterfox v1.8

Something else I noticed:
Sometimes when "Replace Twitch chat" is enabled, Twitch gives a "failed to load module" over chat and shows some errors in the console that don't appear otherwise. Unsure if related to the issue or not.

Image

Image

### Screenshots

Image

### OS and Chatterino Version

Chatterino 2.5.4 (commit 117d06b9) built with Qt 6.7.2 Running on Linux Mint 22.3, kernel: 6.8.0-90-generic

### LINKED ISSUE #6903: Allow custom browser integration manifest path
As noted in https://github.com/Chatterino/chatterino2/issues/6758 - not all browsers or manifest paths are supported.
This is difficult for us to actively maintain, so having an option to manually specify a custom path to write a manifest to would be useful.
We can solve this by adding two more settings under the Browser Integration section:
 - Custom path to write a manifest to. This should be the full path (e.g. `/foo/bar/baz.json`). Stretch goal would be to support parsing a limited set of XDG dirs (e.g. `$XDG_DATA_DIR`). Which format we use is not super important (maybe `%FOO%` is easier since we have clear boundaries).
 - Manifest format as a dropdown (e.g. "Chrome" or "Firefox").

4rneee:
Closes #6903 and #6758(?)

This setting is not added on Windows.

The current implementation only creates the file if the path already exists. No directories are created. This could be changed but would add complexity.

### DIFF
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -147,8 +147,8 @@ QJsonDocument getFirefoxManifest(const QStringList &extensionIDs)
 #ifndef Q_OS_WIN
 void writeManifestToCustomPath(const QJsonDocument &manifest)
 {
-    auto customPath =
-        parseCustomPath(getSettings()->customNativeMessagingManifestPath.getValue());
+    auto customPath = parseCustomPath(
+        getSettings()->customNativeMessagingManifestPath.getValue());
     if (!customPath.has_value())
     {
         return;
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -94,7 +94,7 @@ void registerNmManifest([[maybe_unused]] const Paths &paths,
 #endif
 }
 
-QJsonObject getBaseDocument()
+QJsonObject buildBaseDocument()
 {
     return QJsonObject{
         {u"name"_s, "com.chatterino.chatterino"_L1},
@@ -104,9 +104,9 @@ QJsonObject getBaseDocument()
     };
 }
 
-QJsonDocument getChromeManifest(const QStringList &extensionIDs)
+QJsonDocument buildChromeManifest(const QStringList &extensionIDs)
 {
-    auto obj = getBaseDocument();
+    auto obj = buildBaseDocument();
     QJsonArray allowedOriginsArr = {
         u"chrome-extension://%1/"_s.arg(EXTENSION_ID)};
 
@@ -125,9 +125,9 @@ QJsonDocument getChromeManifest(const QStringList &extensionIDs)
     return QJsonDocument{obj};
 }
 
-QJsonDocument getFirefoxManifest(const QStringList &extensionIDs)
+QJsonDocument buildFirefoxManifest(const QStringList &extensionIDs)
 {
-    auto obj = getBaseDocument();
+    auto obj = buildBaseDocument();
     QJsonArray allowedExtensions = {"chatterino_native@chatterino.com"};
 
     for (const auto &id : extensionIDs)
@@ -249,8 +249,8 @@ void registerNmHost(const Paths &paths)
         getSettings()->additionalExtensionIDs.getValue().split(
             ';', Qt::SkipEmptyParts);
 
-    QJsonDocument chromeManifest = getChromeManifest(extensionIDs);
-    QJsonDocument firefoxManifest = getFirefoxManifest(extensionIDs);
+    QJsonDocument chromeManifest = buildChromeManifest(extensionIDs);
+    QJsonDocument firefoxManifest = buildFirefoxManifest(extensionIDs);
 
     registerNmManifest(paths, CHROME, chromeManifest);
     registerNmManifest(paths, FIREFOX, firefoxManifest);
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -14,6 +14,7 @@
 #include "singletons/Settings.hpp"
 #include "util/IpcQueue.hpp"
 #include "util/PostToThread.hpp"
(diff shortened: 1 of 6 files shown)


tests/src/NativeMessaging.cpp
```diff
@@ -64,3 +64,29 @@ TEST_F(NativeMessagingFixture, writeManifestToWindowsCurrentDir)
...
     ASSERT_TRUE(QFile(combinePath(this->dir.path(), "test.json")).exists());
 }
+
+#ifndef Q_OS_WIN
+TEST(NativeMessaging, parseCustomPath)
+{
+    ASSERT_EQ(parseCustomPath("/my/custom/path/to/manifest.json"),
+              std::optional{"/my/custom/path/to/manifest.json"});
+
+    ASSERT_EQ(parseCustomPath(""), std::optional<QString>{});
```
pajlada: use `std::nullopt` instead for the second param

tests/src/NativeMessaging.cpp
```diff
@@ -64,3 +64,29 @@ TEST_F(NativeMessagingFixture, writeManifestToWindowsCurrentDir)
...
+
+    ASSERT_EQ(parseCustomPath("~/path/to/manifest.json"),
+              std::optional{QDir::homePath() % "/path/to/manifest.json"});
+
+    ASSERT_EQ(parseCustomPath("relative/path/to/manifest.json"),
```
pajlada: use `std::nullopt` instead for the second param

src/singletons/Settings.hpp
```diff
@@ -811,6 +812,16 @@ class Settings
...
     QStringSetting additionalExtensionIDs{"/misc/additionalExtensionIDs", ""};
 
+#ifndef Q_OS_WIN
+    QStringSetting customNativeMessagingManifestPath{
+        "/misc/customNativeMessagingManifestPath", ""};
+    EnumStringSetting<BrowserManifestFormat>
+        customNativeMessagingManifestFormat = {
+            "/misc/customNativeMessagingManifestFormat",
+            BrowserManifestFormat::Chrome,
+    };
```
pajlada, line 824: move these settings to `/misc/extension/...`

src/singletons/NativeMessaging.hpp
```diff
@@ -25,6 +25,10 @@ Expected<void, WriteManifestError> writeManifestTo(QString directory,
                                                    const QString &filename,
                                                    const QJsonDocument &json);
 
+#ifndef Q_OS_WIN
+std::optional<QString> parseCustomPath(QString path);
```
pajlada, line 32: add a comment for this

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -942,6 +942,33 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        layout.addLayout(form);
+        SettingWidget::lineEdit("Custom Native Messaging Manifest Path",
+                                s.customNativeMessagingManifestPath,
+                                "/full/path/to/native/messaging/manifest.json")
+            ->addTo(layout, form);
+
+        SettingWidget::dropdown("Custom Native Messaging Manifest Format",
+                                s.customNativeMessagingManifestFormat)
+            ->addTo(layout);
+    }
```
pajlada: Can remove the "Native Messaging" phrasing in these two settings. e.g. "Custom manifest path" (note the lack of titlecase) is enough

src/singletons/NativeMessaging.cpp
```diff
@@ -142,58 +245,27 @@ void registerNmHost(const Paths &paths)
...
         getSettings()->additionalExtensionIDs.getValue().split(
             ';', Qt::SkipEmptyParts);
 
-    // chrome
-    {
-        auto obj = getBaseDocument();
-        QJsonArray allowedOriginsArr = {
-            u"chrome-extension://%1/"_s.arg(EXTENSION_ID)};
+    QJsonDocument chromeManifest = getChromeManifest(extensionIDs);
+    QJsonDocument firefoxManifest = getFirefoxManifest(extensionIDs);
```
pajlada: rename the two functions called to `buildChromeManifest` and `buildFirefoxManifest` instead

src/singletons/NativeMessaging.cpp
```diff
@@ -93,6 +94,79 @@ void registerNmManifest([[maybe_unused]] const Paths &paths,
 #endif
 }
 
+QJsonObject getBaseDocument()
```
pajlada: rename to `buildBaseManifest` while you're already making changes to it

pajlada (changes requested):

src/singletons/Settings.hpp
4rneee: `/misc/additionalExtensionIDs` too? Then everything related to the extension would be under `/misc/extension/...`.
Or does this cause issues when upgrading?

pajlada: That'll cause issues for users who have previously configured it, so only move the new ones

pajlada (approved):

PR #6919, opened 2026-04-12
Add visual indicator for when your login expires by using the preexisting user button in the gui.

ExeArco:
This pull addresses my concerns brought up in discussion  #3213.
I would like to be completely transparent that I have only rudimentary knowledge of C++ and as such have no idea how the actual code base works. 
This code was 100% AI Generated by Sonnet 4.6 through Claude Code. 

This pull request adds a small visual indicator, based upon the existing design language using a modified version of the  icon ([account-darkMode.svg](https://github.com/Chatterino/chatterino2/blob/master/resources/buttons/account-darkMode.svg))  in resources/buttons. 

I have been running into a consistent issue with Chatterino where, after leaving it running over night, my token would expire naturally, and leave Chatterino in a strange zombie state where chats would still run, however the status of Channels would not update. This would cause me to be unaware of the state of my login token, and I would miss streams until I realized that it was broken. The only way I had to realize that it was broken was by noticing that certain streamers were either marked as streaming in incredibly odd times, or where the usual stream time for a certain streamer does not  go live at their usual time. 

I would like this error to be more easily noticeable at a glance, and therefore have borrowed preexisting UI elements to notify the user.

The following is an except from Claude explaining this:

> Background: Token Revocation vs. Natural Expiry (Zombie State)
> 
>   Chatterino maintains two independent connections to Twitch that authenticate separately:
> 
>   - IRC — a persistent WebSocket connection used for chat messages
>   - Helix API — periodic HTTP requests used for stream status, channel info, etc.
> 
>   Token revocation (e.g. disconnecting Chatterino from Twitch's Connections settings, or re-logging in) causes Twitch to immediately terminate the IRC connection. Chatterino then tries to reconnect, Twitch rejects the dead token with a Login authentication failed NOTICE,
>    and it's obvious something is wrong.
> 
>   Natural token expiry behaves differently. When a token silently expires via TTL, Twitch does not terminate existing connections. The IRC WebSocket stays alive — chat continues to work normally, messages arrive, you can see people talking. However, the Helix API begins
>   returning 401 Unauthorized on every request. Since the stream status poller (LiveController) was silently discarding these errors, stream statuses would freeze at whatever they were at expiry time. Streamers would appear offline when live, or live when offline, with no
>    indication anything was wrong.
> 
>   This creates a "zombie state": Chatterino looks and feels functional, but a silent, invisible failure is causing it to show stale data. The only way to notice was to reopen Chatterino (which would trigger a fresh IRC connection attempt, fail auth, and show the expired
>   indicator) or to notice that European streamers were somehow still live at 5am.
> 

I have ran chatterino-tests before and after adding my changes and both times 479 out of 490 tests pass. 
Additionally, in order to test that this successfully catches the "zombie" state that I was having, and that waiting for a token to expire manually can take weeks, I the zombie state was simulated to the best of my abilities using mitmproxy. 
I ran mitmproxy, accepted the CA system-wide, and rewrote any response from api.twitch.tv to return error 401, and ran Chatterino through that proxy. 
This allowed me to confirm that in that state, chat was still available, stream statuses weren't being updated properly, and it displayed the visual indicator notifying me that I needed to renew my login.


resources/buttons/account-expired.svg
pajlada, line 1: This needs to be remade by a human

src/widgets/Notebook.cpp
```diff
@@ -1497,16 +1499,29 @@ void SplitNotebook::addCustomButtons()
...
+    auto loginExpiredState = std::make_shared<bool>(false);
+
     userBtn->setVisible(!getSettings()->hideUserButton.getValue());
     getSettings()->hideUserButton.connect(
-        [this, userBtn](bool hide) {
+        [this, userBtn, loginExpiredState](bool hide) {
+            if (*loginExpiredState)
+            {
+                return;
+            }
```
pajlada: I don't think you need this `loginExpiredState` logic. 
Since you already force visibility to true in the new signals you added below, this would only come into effect if a user explicitly opens the settings dialog & toggles the "Show user button" setting, which I don't think we should interfere with.

pajlada (changes requested):
This currently fails if you're in anonymous mode - this might be most easily fixed by ensuring the helix requests don't actually get called if you're not signed in

ExeArco:
I have remade the icon using Inkscape.

Then, in the functionality, I fixed the anonymous-mode false positive, which oddly was actually only triggering for you because you had an other account that had an invalid token, and wouldn't have actually triggered when your actual accounts still have valid tokens. This confused me greatly at first. 

Claude flagged a potential leak with my code in relation to the addCustomButtons() section. 
Here is an excerpt from Claude explaining it in more detail:

> fix: tie account-button currentUserChanged connection to signalHolder_
>   
>   The currentUserChanged handler in addCustomButtons() used a raw
>   boost::signals2 connect that was never disconnected. The signal lives
>   on AccountController (app lifetime), but the lambda captures the
>   SplitNotebook (this) and userBtn, which don't. After a notebook is
>   destroyed (e.g. a closed popped-out window), the next account switch
>   fires the stale lambda against freed memory — a use-after-free, plus a
>   slow leak of connections that accumulate per notebook.
> 
>   Route it through this->signalHolder_.managedConnect so it's torn down
>   with the widget, matching loginExpired above it and the existing
>   currentUserChanged call sites in Window/Split/SplitHeader.

I have re ran chatterino-tests and 620/643 pass, and once again simulated the zombie state with mitmproxy.

src/widgets/Notebook.cpp
```diff
@@ -1497,13 +1499,22 @@ void SplitNotebook::addCustomButtons()
...
+        .dark = ":/buttons/account-expired.svg",
+        .light = ":/buttons/account-expired.svg",
+    };
 
+    auto *userBtn = this->addCustomButton<SvgButton>(normalAccountSrc);
     userBtn->setPadding({0, 0});
 
+    // Shared state: whether login is currently expired. Used to restore the
+    // normal icon (and the user's visibility setting) once they re-authenticate.
+    auto loginExpiredState = std::make_shared<bool>(false);
```
pajlada: I don't think this shared pointer should be necessary at all anymore

pajlada (changes requested):
For the boost::signals2 comment - we don't use that anymore so I think your robot is reading old code

pajlada (approved):
This seems reasonable, I will merge this in & hope to get some testing from others in the nightly version.
Thanks!

jupjohn:
Just saw this feature in action today - it works! I'll shove a few possible improvements into an issue.

Issue #6901, opened 2026-04-04
Make it easier to save logs to a file on Windows


pajlada:
Currently, if we require a user to send us logs to debug an issue they're having, it's very complicated.

A user can use [c2-logs](https://github.com/nerixyz/c2-logs), but it requires the user to download another project.
A user can use the `/debug-relaunch-with-console` command from #6799, but it spawns the old conhost so the user doesn't have an easy time to actually export the logs to a file.

I propose we add a command similar to `/debug-relaunch-with-console` that relaunches Chatterino with logs saving to a file enabled. When this flag is enabled, we'd overwrite the qt logging handler with one that saves logs to a file.

See https://doc.qt.io/qt-6/qtlogging.html#qInstallMessageHandler

8thony:
Thanks @invisibl3cat for adding this useful feature.

Technically you already satisfied this issue, but I found 2 little annoyances in your implementation.
- If I use `/debug-enable-logfile` or `/debug-disable-logfile` i dont get feedback that logging started or stopped.
- If I use `/debug-enable-logfile` with a path surrounded by quotation marks, I get an error `Unable to open log file`.

pajlada:
Closing as fixed - made #6974 to track your feedback @8thony

PR #6920, opened 2026-04-12
chore: Redirect log output to a file

invisibl3cat:
This PR adds two commands that redirect log output to a file instead of wherever they would be written otherwise. This tries to address #6901

The commands are:
`/debug-enable-logfile <path>` to enable immediate redirection of logs to a file.
`/debug-relaunch-with-logfile <path>` to restart Chatterino with logs being written to a file right after launch.

EDIT:
I also added `/debug-disable-logfile` that switches off logging to a file.


src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -34,6 +35,46 @@ using namespace Qt::StringLiterals;
 
 namespace chatterino::commands {
 
+static bool restartChatterino(const QProcessEnvironment &env)
+{
+    QProcess proc;
+    proc.setProgram(qApp->applicationFilePath());
+    // https://doc.qt.io/qt-6/debug.html#environment-variables-recognized-by-qt
```
Nerixyz, line 44: Move this to where we name the environment variables.

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -34,6 +35,46 @@ using namespace Qt::StringLiterals;
 
 namespace chatterino::commands {
 
+static bool restartChatterino(const QProcessEnvironment &env)
```
pajlada: move to anon namespace instead of making the functions static

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -34,6 +35,46 @@ using namespace Qt::StringLiterals;
...
+                QApplication::exit();
+            },
+            Qt::QueuedConnection);
+        return true;
+    }
+
+    return false;
+}
+
+static QProcessEnvironment setUpEnvironmentForLogging(
```
pajlada: move to anon namespace instead of making the functions static

src/util/LoggerToFile.hpp
```diff
@@ -0,0 +1,33 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#define CHATTERINO_REDIRECT_LOG_TO_FILE_ENVVAR "CHATTERINO_REDIRECT_LOG_TO_FILE"
```
pajlada: use a qstringview or something instead of a define

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -256,42 +297,77 @@ QString relaunchWithConsole(const CommandContext &ctx)
...
-            [] {
-                QApplication::exit();
-            },
-            Qt::QueuedConnection);
+        ctx.channel->addSystemMessage(
+            "Unable to open log file. Check the file path and that you have "
+            "write permissions to it");
+    }
+
+    return {};
```
Nerixyz, line 354: Print the absolute path of the logfile here. Maybe even with a link.

src/util/LoggerToFile.cpp
```diff
@@ -0,0 +1,87 @@
+#include "util/LoggerToFile.hpp"
+
+#include <QFileInfo>
+
+namespace {
+
+std::unique_ptr<chatterino::LoggerToFile> LOGGER = nullptr;
```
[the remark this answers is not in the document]
Nerixyz: You should be able to declare this `constinit` - that won't silence the warning but should be enough for us. This is mainly about having to run static constructors when starting.

src/util/LoggerToFile.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <QFile>
+#include <QString>
+
+#include <memory>
+#include <mutex>
+
+namespace chatterino {
+
+class LoggerToFile
```
pajlada: Make this a singleton instead (lives under `src/singletons/`) similar to `Settings`. The instance itself can then have an enable(path) & disable() (no need to implement that) and handle the previous logger with function calls instead of RAII.
Let it live as a raw pointer & leak on exit, no need to do anything on destruction.

pajlada (changes requested):

src/util/LoggerToFile.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        qInstallMessageHandler(this->originalHandler_);
+    }
+}
+
+void LoggerToFile::log(QtMsgType /*type*/,
+                       const QMessageLogContext & /*context*/,
+                       const QString &msg)
+{
+    this->logFile_.write((msg + "\n").toUtf8());
```
Nerixyz: - You need to flush the file here. Otherwise, the writes are buffered and if Chatterino crashes, we won't see any output.
- Use `qFormatLogMessage` to respect `QT_MESSAGE_PATTERN`
- Call the original handler if it exists. This way you can still see the messages in the stdout/stderr/debug output.

src/util/LoggerToFile.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    // Check that we can actually create the log file
+    {
+        QFile f(absFilePath);
+        bool success = f.open(QIODevice::WriteOnly);
+        if (!success)
+        {
+            return false;
+        }
+    }
```
Nerixyz: You can save the success as a member variable in the class and check that after you create the logger. The file should also have an error string ([`QIODevice::errorString`](https://doc.qt.io/qt-6/qiodevice.html#errorString)). You can return `ExpectedStr<void>`.

src/util/LoggerToFile.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    // Make sure that we destroy any previous logger before creating a new one
+    LOGGER = nullptr;
+
+    // make_unique requires the c-tor to be public, at least with VC++ 19.50.(lotta numbers)
+    auto *ptr = new LoggerToFile(filePath);
```
Nerixyz: Use something like `std::unique_ptr<LoggerToFile> logger(new LoggerToFile(filePath))`. This way you can't leak the pointer. After you check the error state (see above), you can move it to `LOGGER`.

src/controllers/commands/builtin/chatterino/Debugging.cpp
invisibl3cat: Done

src/controllers/commands/builtin/chatterino/Debugging.cpp
invisibl3cat: Done

src/util/LoggerToFile.hpp
invisibl3cat: `qEnvironmentVariable()` takes a raw const char * so I've replaced the `#define` with that.

src/util/LoggerToFile.hpp
invisibl3cat: Done

src/controllers/commands/builtin/chatterino/Debugging.cpp
invisibl3cat: I've expanded the error message to show the absolute path to the log file and the reason why we could not open it.

src/util/LoggerToFile.cpp
invisibl3cat: Done

invisibl3cat (commented):
Updated as requested

pajlada (approved):
I made some minor changes, including adding an SPDX/Copyright header to one of the files. Please let me know if all of this looks fine to you @invisibl3cat and I can merge this in.

Cool that you can still log to the normal logger

src/singletons/FileLogger.cpp
```diff
@@ -0,0 +1,130 @@
...
+    {
+        auto result = this->enable(env.logToFile);
+        if (!result.has_value())
+        {
+            auto error = result.error();
+            QString errorMessage = QString("Unable to open log file %1. Error "
+                                           "reported by the system was: %2")
+                                       .arg(error.absFilePath, error.errorDesc);
+
+            std::cerr << errorMessage.toLocal8Bit().constData() << '\n';
```
Nerixyz:
```suggestion
            std::cerr << errorMessage.toStdString() << '\n';
```

Not sure if we can assume the output of `toLocal8Bit` is null terminated (though in practice we probably can). Since the IO-streams support `std::string`, use that.

src/singletons/FileLogger.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+}  // namespace
+
+namespace chatterino {
+
+FileLogger *FileLogger::INSTANCE = nullptr;
+
+FileLogger::FileLogger()
+{
+    FileLogger::INSTANCE = this;
```
Nerixyz, line 32:
```suggestion
    assert(FileLogger::INSTANCE == nullptr);
    FileLogger::INSTANCE = this;
```

pajlada: resolved in [`39cb9e1` (this PR)](https://github.com/Chatterino/chatterino2/pull/6920/commits/39cb9e1fd7a52c787c8961640e0d17d0312e36ab)

src/singletons/FileLogger.cpp
pajlada: resolved in [`39cb9e1` (this PR)](https://github.com/Chatterino/chatterino2/pull/6920/commits/39cb9e1fd7a52c787c8961640e0d17d0312e36ab)

invisibl3cat:
Oops, didn't notice the Env file 😳.  Thanks for tidying up the patch!

One thing I remember about "\n" vs. std::endl on Windows: When I used "\n" as I usually do, Windows Console sometimes wouldn't flush the output right away which is awkward when looking for a debugging message. Using std::endl would, apparently, force the flush.

Anyway, this is all fine by me.

Issue #3906, opened 2022-08-06
ci: move freebsd build from CirrusCI to Actions


jammehcow:
### Checklist

- [X] I'm reporting a problem with Chatterino
- [X] I've verified that I'm running **the most** recent nightly build or stable release
- [X] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [X] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Having our FreeBSD pipeline running in the same environment/workflow spec as the rest of the builds would improve maintainability of the pipelines and allow users familiar with GitHub Actions to diagnose issues with the FreeBSD CI.

### OS and Chatterino Version

N/A

jammehcow:
Inspired by the [freebsd-vm project](https://github.com/vmactions/freebsd-vm) which allows you to run a FreeBSD VM on a macOS host.

jammehcow:
Going to look at redoing #3509 if @Mm2PL is fine with it. Would make the definition of FreeBSD runners less hacky.

Mm2PL:
> if @Mm2PL is fine with it
@jammehcow, go for it.

pajlada:
I've marked this as a low priority since we're not using Cirrus as a blocking CI. If it fails, or if it's slow, it doesn't slow us down.

pajlada:
Fixed in #6922

PR #6922, opened 2026-04-13
Add GitHub CI workflow for building on FreeBSD

Wissididom:
I saw somewhere that we wanted to have them in CI and also I saw today that [CirrusLabs joins OpenAI and shuts down Cirrus CI](https://cirruslabs.org/) - I just found https://github.com/Chatterino/chatterino2/issues/3906

Screenshot From 2026-04-13 20-03-50

This seems to work, at least in CI. I'm not sure if I should remove the `.cirrus.yml`, tho.


.github/workflows/build-freebsd.yml
```diff
@@ -0,0 +1,66 @@
...
+          fetch-depth: 0 # allows for tags access
+
+      - name: Build using FreeBSD
+        uses: vmactions/freebsd-vm@v1.4.4
+        with:
+          envs: "C2_ENABLE_LTO"
+          release: "14.4"
+          arch: x86_64
+          mem: 4096
+          cpu: 2
```
Nerixyz: Can these be inferred from the runner? For example from `nproc` and `/proc/meminfo`.

Wissididom: Probably, I might also be able to just remove both of them, but I thought I'd add them with the values I saw at Cirrus. Removed in 8818d85b
mem and cpu are command line switches that are added to the anyvm command.
Edit: Ok, apparently the default mem is 2048 and the default for cpu is "all host cores" (Source: https://github.com/anyvm-org/anyvm#release--arch--resources)

pajlada (changes requested):
You can remove the cirrus-specific config file alongside this change.

Thanks for noticing this!

pajlada (approved):

PR #6923, opened 2026-04-13
chore: remove references to temporaries in for loops

Nerixyz:
Clang's [`-Wrange-loop-analysis`](https://clang.llvm.org/docs/DiagnosticsReference.html#id890) is useful to detect references in range for-loops. We have this quite a lot when iterating over a `QJsonObject` or `QJsonArray`. They yield `QJsonValue(Const)Ref`s. These are already a reference.


pajlada (approved):

Nerixyz:
Should also fix some of the above, but in a separate change.

PR #6925, opened 2026-04-16
fix: don't mark announcements as subscriptions

### LINKED ISSUE #4163: Announcements listed when searching for subscriptions
### Checklist

### Describe your issue

When I search for subscriptions (with either `is:sub` or `is:subscription`) I noticed an announcement appeared there. I uploaded a screenshot in the editor below.

### Screenshots

image

### OS and Chatterino Version

Chatterino 2.3.5 DEBUG (commit 32d077c4) built with Qt 5.15.6 Running on Manjaro Linux, kernel: 5.18.19-3-MANJARO

TheMizeGuy:
## Summary

Announcements sent via Twitch's `/announce` command arrive over IRC as `USERNOTICE` messages, which share the same code path as subscription events (`sub`, `resub`, `subgift`, etc.). As a result, the code was setting `MessageFlag::Subscription` on announcement messages, causing them to appear in search results for `is:sub` / `is:subscription`.

This PR makes two small changes in `IrcMessageHandler`:

1. **Skip setting `MessageFlag::Subscription` when the message type is `announcement`** in both the system-message and user-content paths of `parseUserNoticeMessageInto` / `addMessage`.

2. **Switch the existing announcement check to use the resolved `msgType` parameter** instead of the raw `tags.value("msg-id")` tag. For shared-chat announcements, `msg-id` is `"sharedchatnotice"` and the real type lives in `source-msg-id`, which is already resolved into `msgType` earlier in the pipeline. This fixes a secondary issue where shared-chat announcements were losing their `Highlighted` flag.

Fixes #4163.

## Steps to reproduce (before this PR)

1. In a channel, have a mod/broadcaster run `/announce <text>`
2. Open search (Ctrl+F) and search `is:sub`
3. The announcement incorrectly appears in the results

After this PR, announcements no longer match `is:sub` / `is:subscription`, but still appear under `is:highlighted` as expected.

## Verification

- `announcement.json` and `shared-chat-announcement.json` snapshots were regenerated via `UPDATE_SNAPSHOTS = true` and verified; the `Subscription` flag is removed, and the shared-chat snapshot's `Highlighted` flag is now correctly preserved.
- All 101 `IrcMessageHandler` snapshot tests pass locally.
- Verified all 6 consumers of `MessageFlag::Subscription` (highlight color, reply-ability, layout, search predicate, tab mentions, user info popup) — none require the flag on announcements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unversioned
 
+- Bugfix: Fixed Twitch announcements appearing in search results when searching for subscriptions (`is:sub`/`is:subscription`). (#6925)
+
 ## 2.5.5
 
 - Minor: Update emoji data to Unicode 17.0. (#6471)
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -914,7 +914,7 @@ void IrcMessageHandler::parseUserNoticeMessageInto(Communi::IrcMessage *message,
         {
             msg->flags.set(MessageFlag::WatchStreak);
         }
-        else
+        else if (msgType != "announcement")
         {
             msg->flags.set(MessageFlag::Subscription);
         }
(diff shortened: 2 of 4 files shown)


pajlada:
Thanks for your contribution - this should be handled by https://github.com/Chatterino/chatterino2/pull/6927

PR #6927, opened 2026-04-18
feat(highlights): Support displaying colored announcements

### LINKED ISSUE #4163: Announcements listed when searching for subscriptions

### LINKED ISSUE #5218: Add support for /announce[color] command variations

4rneee:
Closes #5218

Since now announcements no longer have the `Subscription` flag this also closes #4163.

I am **very** unsure about the colors I chose and more than open for suggestions. I chose the default primary announcement color to be the same as the purple one since thats the behavior I noticed in the webchat.

announce


src/messages/Message.hpp
```diff
@@ -121,6 +122,8 @@ struct Message {
 
     uint32_t bits{0};
 
+    HelixAnnouncementColor announcementColor{HelixAnnouncementColor::Primary};
```
pajlada, line 125: add to `toJson`

src/messages/layouts/MessageLayout.cpp
```diff
@@ -418,6 +419,47 @@ void MessageLayout::updateBuffer(QPixmap *buffer,
                 blendColors(backgroundColor, *this->message_->highlightColor);
         }
     }
+    else if (this->message_->flags.has(MessageFlag::Announcement) &&
+             ctx.preferences.enableAnnouncementHighlight)
+    {
+        if (ctx.preferences.enableColoredAnnouncementHighlight)
+        {
+            switch (this->message_->announcementColor)
+            {
```
Nerixyz: What about turning this into something like
```cpp
ColorType colorType = [&]{
  if (!enableHighlight)
    return ColorType::AnnouncementHighlight;
  switch(announcementColor) {
    case HelixAnnouncementColor::Primary:
      return ColorType::AnnouncementHighlight;
    ...
  }
  return ColorType::AnnouncementHighlight;
}();

backgroundColor = blendColors(backgroundColor, *ctx.colorProvider.color(colorType));
```

src/messages/Message.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include "messages/MessageFlag.hpp"
 #include "providers/twitch/ChannelPointReward.hpp"
+#include "providers/twitch/api/Helix.hpp"
```
Nerixyz: Avoid including Helix here if possible, because it's a large header.

src/messages/MessageFlag.hpp
```diff
@@ -77,6 +77,7 @@ enum class MessageFlag : std::int64_t {
     /// Note: If this message is inside a reply thread, the root must not have the flag either.
     InvalidReplyTarget = (1LL << 42),
     WatchStreak = (1LL << 43),
+    Announcement = (1LL << 44),
```
Nerixyz, line 80: Rerun [scripts/make_luals_meta.py](https://github.com/Chatterino/chatterino2/blob/master/scripts/make_luals_meta.py) and update https://github.com/Chatterino/chatterino2/blob/b36ccdf776c6c1f57ab2a8ac7099ac829a34a9f7/docs/chatterino.d.ts#L444

src/singletons/Settings.hpp
```diff
@@ -577,6 +577,13 @@ class Settings
     QStringSetting watchStreakHighlightColor = {
         "/highlighting/watchStreak/color", ""};
 
+    BoolSetting enableAnnouncementHighlight = {
+        "/highlighting/announcement/enabled", true};
+    QStringSetting announcementHighlightColor = {
+        "/highlighting/announcement/color", ""};
+    BoolSetting enableColoredAnnouncementHighlight = {
+        "/highlighting/announcement/coloredAnnouncement/enabled", true};
+
```
Nerixyz, line 596: These are not visible in the settings.

Nerixyz:
> I am **very** unsure about the colors I chose and more than open for suggestions. I chose the default primary announcement color to be the same as the purple one since thats the behavior I noticed in the webchat.

I don't think we can provide a great color for _all_ users. As long as the colors roughly match the ones that we already have, it should be fine. We have users using the light theme. They will probably need to make adjustments anyway. As the dark theme is the default, it's more important that it looks good there.

src/messages/Message.hpp
4rneee: Right, I noticed that too. What's the best approach here? Should I create a new `providers/twitch/api/helix/HelixAnnouncement.hpp` which I then include in `Helix.hpp`?

src/singletons/Settings.hpp
4rneee: What do you mean by that? They are in the Highlights tab.

src/messages/Message.hpp
Nerixyz: Yea that or a `HelixEnums.hpp` that contains all enums.

src/singletons/Settings.hpp
Nerixyz: Oh right, nvm

src/messages/layouts/MessageLayout.cpp
4rneee: Sure, tho at this point I could also create a helper function since I also need to map `HelixAnnouncementColor` to `ColorType` in `Message::getScrollBarHighlight`.

Where would be a good place for this?

src/messages/MessageFlag.hpp
4rneee: Done. I also added  `WatchStreak`  in `chatterino.d.ts` because it was missing.

src/messages/layouts/MessageLayout.cpp
Nerixyz: Maybe in `ColorProvider.c/hpp` as `colorTypeFromHelixAnnouncementColor`? That's a bit verbose, but sadly we can't have member functions for enums.

4rneee:
> We have users using the light theme. They will probably need to make adjustments anyway. 

Right now these colors can't be adjusted.

>As the dark theme is the default, it's more important that it looks good there.

Agreed. Could we also provide colors depending on the theme? Custom themes probably should also be able to set them, right?

Nerixyz:
> Right now these colors can't be adjusted.

Ah, I missed that. I think it would be better to allow users to configure the colors given that they can modify the other colors. But not sure how much value that provides.

> Could we also provide colors depending on the theme? Custom themes probably should also be able to set them, right?

It's a bit tricky if you allow users to configure the colors as well. If you don't, it should be easier I think.

4rneee:
Should I address all those `performance-enum-size` warnings? I feel like we always ignore those. If we don't care we could remove the check since it only clutters the conversation.

4rneee:
> > Right now these colors can't be adjusted.
> 
> Ah, I missed that. I think it would be better to allow users to configure the colors given that they can modify the other colors. But not sure how much value that provides.
> 
> > Could we also provide colors depending on the theme? Custom themes probably should also be able to set them, right?
> 
> It's a bit tricky if you allow users to configure the colors as well. If you don't, it should be easier I think.

If I understand this correctly, we should either allow changing the colors through the UI or through the theme.

I'm not sure what a good UI would be to do that since I feel like adding a row for all four announcement colors fills up the settings too much for the value this provides. Maybe some kind of popup? 

I therefore tend to have them fixed by the theme. Users who dislike the colors can always disable it and still have announcements highlighted in the same color or create a custom theme.

Let me know what you think.

Nerixyz:
> Should I address all those `performance-enum-size` warnings? I feel like we always ignore those. If we don't care we could remove the check since it only clutters the conversation.

I don't see a reason why we wouldn't want to specify a smaller type. We don't need forwards compatibility here nor do we serialize this to a binary stream. Many of these enums are older than the clang-tidy CI, so we never got that warning before. But if we touch them anyway, we might as well fix that.

> I'm not sure what a good UI would be to do that since I feel like adding a row for all four announcement colors fills up the settings too much for the value this provides. Maybe some kind of popup?
> 
> I therefore tend to have them fixed by the theme. Users who dislike the colors can always disable it and still have announcements highlighted in the same color or create a custom theme.

Sure, that sounds like a good plan.

pajlada (changes requested):
> Could we also provide colors depending on the theme?

That seems reasonable if the colors you've provided don't look good in light mode

> Should I address all those performance-enum-size warnings?

I'd prefer you don't move all the other helix enums unless necessary for this change. Feel free to still call the file `HelixEnums.hpp` and follow up with a PR afterwards. You can then proceed to address the `performance-enum-size` warning specifically for the enums you've touched/moved.

4rneee:
> > Could we also provide colors depending on the theme?
> 
> That seems reasonable if the colors you've provided don't look good in light mode

Okay I will look into that.

> > Should I address all those performance-enum-size warnings?
> 
> I'd prefer you don't move all the other helix enums unless necessary for this change. Feel free to still call the file `HelixEnums.hpp` and follow up with a PR afterwards. You can then proceed to address the `performance-enum-size` warning specifically for the enums you've touched/moved.

So only move `HelixAnnouncementColor` now and then make another PR that moves the rest, right?

pajlada:
> So only move HelixAnnouncementColor now and then make another PR that moves the rest, right?

Correct. That way this PR doesn't have to be noisy with feedback from clang-tidy and the changes are also less noisy.

pajlada (approved):

4rneee:
What about specifying the colors in a custom theme? I was looking into that a bit before this got merged. Is this still something we would want?

pajlada:
> What about specifying the colors in a custom theme? ... Is this still something we would want?

I don't personally think that's a super-important feature, and I don't have an issue with waiting until someone formally requests it (this could be you!)

> I was looking into that a bit before this got merged. 

Feel free to make your PRs a draft if you don't want me to sneakily merge the PR in the future image

PR #6930, opened 2026-04-19
feat(plugins): Add bindings for emotes

Nerixyz:
This adds bindings to create emotes and to read emotes from messages. Notably, it doesn't add an `EmoteMap`. With this PR, plugins can build all elements except `LinkElement`.

Emotes can be created from a table with `c2.Emote.new_uncached`. When plugins can create emote providers, then we'd add a function to create cached emotes, as the cache lives in the provider and there's no global cache.

### DIFF
--- a/src/controllers/plugins/api/Emotes.hpp
+++ b/src/controllers/plugins/api/Emotes.hpp
@@ -1,0 +1,35 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#ifdef CHATTERINO_HAVE_PLUGINS
+
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api::emotes {
+
+/* @lua-fragment
+
+---@class c2.Emote
+---@field name string
+---@field images c2.ImageSet
+---@field tooltip string
+---@field home_page string URL to the platform specific emote page.
+---@field zero_width boolean
+---@field id string Platform specific ID.
+---@field author string Username of the emote creator.
+---@field base_name string|nil If this emote is aliased, this contains the original (base) name of the emote.
+
+---Create a new emote. This emote is not cached anywhere.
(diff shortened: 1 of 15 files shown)


src/messages/MessageElement.hpp
```diff
@@ -593,10 +600,12 @@ class FfzBadgeElement : public BadgeElement
     QJsonObject toJson() const override;
     std::string_view type() const override;
 
+    QColor color() const;
+
 protected:
     MessageLayoutElement *makeImageLayoutElement(const ImagePtr &image,
                                                  QSizeF size) override;
-    const QColor color;
+    const QColor color_;
```
pajlada: undo

src/messages/MessageElement.hpp
```diff
@@ -593,10 +600,12 @@ class FfzBadgeElement : public BadgeElement
     QJsonObject toJson() const override;
     std::string_view type() const override;
 
+    QColor color() const;
```
pajlada: rename `getColor`

src/controllers/plugins/api/Emotes.hpp
```diff
@@ -0,0 +1,35 @@
...
+---@field home_page string URL to the platform specific emote page.
+---@field zero_width boolean
+---@field id string Platform specific ID.
+---@field author string Username of the emote creator.
+---@field base_name string|nil If this emote is aliased, this contains the original (base) name of the emote.
+
+---Create a new emote. This emote is not cached anywhere.
+---@param tbl {name: string, images: c2.ImageSet, tooltip: string, home_page?: string, zero_width?: boolean, id?: string, author?: string, base_name?: string}
+---@return c2.Emote
+function c2.Emote.new_uncached(tbl) end
```
pajlada: Since this can be a bit of a problematic API if used incorrectly, I think we should be as clear as possible in how this is meant to be used.

pajlada (changes requested):
not a complete review, just a quick glance.

functionality as a whole seems reasonable

PR #6939, opened 2026-04-22
feat: Hide chat message timestamps when the Twitch channel is live

### LINKED ISSUE #6933: Option to "hide timestamps when streamer is live"
### Discussed in https://github.com/Chatterino/chatterino2/discussions/6742

<sup>Originally posted by **Samueltelegram** January 14, 2026</sup>
Right now you can disable timestamps or choose a message timestamp format. 
I'd like an added option which disabled timestamps when the streamer is live.
This is especially useful for if you only watch big streamers. 
Like in an offline chat you'd like to see how recent messages were, but at the same time for live watching with fast chats its redundant.
This removes chat bloat.</div>

This feature should add one more setting which allows the user to conditionally hide the timestamp in front of each message.
By default, the current behaviour of always showing the timestamp should stay.

If the user has already disabled the timestamp, this functionality will not have any effect.

Setting could be named something like: "Conditionally hide timestamp when:" with the options "Never" (default), and "When streamer is live"

invisibl3cat:
An idea how to address https://github.com/Chatterino/chatterino2/issues/6933

The original suggestion seems to consider a situation where a very fast going chat makes timestamps a useless noise. It is unclear what to do with messages that are not a "generic" chat but system messages, whispers etc. In many cases, the code that creates `TimestampElement` does not have access to the Twitch channel to check if the stream is live - at least as far as I can tell.

This PR hides timestamps only for "generic" chat and gifted subs messages. Do we want to hide timestamps for all other  messages too? If so, this would probably require a more invasive change.


Mm2PL:
Have you considered using a MessageElementFlag and hiding all timestamps using it? We have `WindowManager::updateWordTypeMask()` which already hides timestamps

Nerixyz:
> Have you considered using a MessageElementFlag and hiding all timestamps using it? We have `WindowManager::updateWordTypeMask()` which already hides timestamps

Wanted to comment this as well once this is undrafted. I think it should be enough to check `Channel::isLive` in [`ChannelView::getFlags`](https://github.com/Chatterino/chatterino2/blob/7bae9185ccdb408740cde2e6b1bff044e830eeb3/src/widgets/helper/ChannelView.cpp#L1391) for `underlyingChannel_`.

invisibl3cat:
> Have you considered using a MessageElementFlag and hiding all timestamps using it? We have `WindowManager::updateWordTypeMask()` which already hides timestamps

Ha... didn't notice that this was possible. Thanks for pointing it out! It feels like a less messy solution. This will, however, hide timestamps for all messages in the channel. Are we sure we want this? It might make sense to show timestamps for system messages, announcements, etc.

pajlada:
I think this should hide all timestamps in the channel yeah, regardless of message type

invisibl3cat:
There is one difference between my original solution and the simpler one suggested by @Mm2PL and @Nerixyz. Fixing up MessageElementFlags in ChannelView removes timestamps from all messages when the channel goes live, even from those that were sent before the channel went online. My initial attempt would always keep the "offline" messages timestamped... do we care about this?

Nerixyz:
> do we care about this?

No, given that the current implementation is much simpler.

src/singletons/Settings.hpp
```diff
@@ -159,6 +176,9 @@ class Settings
     BoolSetting showTimestamps = {"/appearance/messages/showTimestamps", true};
     BoolSetting animationsWhenFocused = {
         "/appearance/enableAnimationsWhenFocused", false};
+    EnumSetting<HideMessageTimestamp> hideMessageTimestamp = {
+        "/appearance/messages/hideMessageTimestamp",
+        HideMessageTimestamp::Never};
```
Nerixyz: If you really want this to be an enum, use `EnumStringSetting` - that's better for compatibility.

However, I don't see why this would need to be an enum. Specifically, when you add more situations where timestamps are hidden, the available options will explode. Say you'd add "when looking away" as another situation. Now you'd also need to add "when looking away or channel is live" as an option. You'll have $2^n$ entries in there with $n$ being the number of situations.

I'd suggest making this a boolean setting.

pajlada: This is not how the issue explained it, but a boolean setting is fine.

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -515,6 +515,12 @@ void GeneralPage::initLayout(GeneralPageView &layout)
             "When enabled, messages deleted by moderators will be hidden.")
         ->addTo(layout);
 
+    SettingWidget::checkbox("Hide message timestamps when channel is live",
+                            s.hideMessageTimestampsWhenLive)
+        ->setTooltip("When enabled, timestamps of all messages will be hidden "
+                     "when the channel is live")
```
pajlada: can remove the tooltip, the setting name is self-explanatory

pajlada (changes requested):
just this nit, otherwise looks good

pajlada (approved):

PR #6943, opened 2026-04-25
fix(highlights): prioritize custom highlitghts as message background

4rneee:
Previously when a message matched a custom highlight the background color was not always set because other message flags had higher priority. This behavior also differed from the scrollbar highlights and is now fixed.


4rneee:
Kind of related: Is there a reason why we don't allow sub (or rather any special) messages to get highlighted?

https://github.com/Chatterino/chatterino2/blob/1c56fb147e488d821806f77e86c279464392f579/src/providers/twitch/IrcMessageHandler.cpp#L1291

4rneee:
Since #7038, watchstreak (and other special messages) now show up in mentions which fixed the original issue this PR tried to address. Now I only changed the order in which the message background color is evaluated such that custom highlights have the highest priority and changed title and description accordingly.

I am not sure if the highlight rework fixes this anyways by allowing users to manually configure this priority. In that case I could just close this PR.

pajlada (commented):
This code is changing a lot with the highlights PR (see https://github.com/Chatterino/chatterino2/pull/6695/changes#diff-8fd31b561faf33cecc8e8db779ef914a4401e5fbaf4ac637b0077d60cd80ae18R389 for progress) - I think that should cover this issue

PR #6947, opened 2026-04-25
feat(plugins): Add custom channel providers 

Nerixyz:
This adds the ability for plugins to manage their own channels. Channels are still created by users in the select channel popup. Plugins provide a description for a channel along with fields they'd like the user to provide. An example channel is in the prose documentation (`wip-plugins.md`).

Since plugins can come and go, these channels can either be adopted or orphaned, and they can switch between these states.
When we start, all plugin channels are orphaned. When a plugin registers a provider, all channels belonging to this provider are adopted (calling `create` in the provider).  This essentially provides the plugin with a vtable for all (two) virtual functions.

Draft because it depends on #6906 and has no tests.


docs/lua-meta/globals.lua
```diff
@@ -7,17 +7,17 @@
 c2 = {}
 ---@enum c2.LogLevel
 c2.LogLevel = {
```
Mm2PL, line 9: rerun the script to fix indents

src/controllers/plugins/Plugin.cpp
```diff
@@ -2,12 +2,13 @@
 //
 // SPDX-License-Identifier: MIT
 
+#include "api/ChannelProviders.hpp"
```
Mm2PL: Move this include below the ifdef to get nicer sorting

src/controllers/plugins/PluginController.cpp
```diff
@@ -2,9 +2,9 @@
 //
 // SPDX-License-Identifier: MIT
 
+#include "api/ChannelProviders.hpp"
+#include "controllers/plugins/PluginChannel.hpp"
```
Mm2PL: move for nicer sorting

src/widgets/dialogs/SelectChannelDialog.hpp
```diff
@@ -86,17 +92,37 @@ class SelectChannelDialog final : public BaseWindow
 
         detail::AutoCheckedRadioButton *automod;
         QLabel *automodLabel;
+
+        MicroNotebook *notebook;
```
Mm2PL: Why does this use a new MicroNotebook instead of reusing the existing Notebook class? Can that one not do something we need here?

src/widgets/dialogs/SelectChannelDialog.cpp
Mm2PL, line 589: This window should implement the Action `openTab` now that it has tabs again.

src/widgets/dialogs/SelectChannelDialog.cpp
```diff
@@ -421,4 +545,108 @@ void SelectChannelDialog::addShortcuts()
...
+        delete item->widget();
+        delete item;
+    }
+
+    if (!provider)
+    {
+        return;
+    }
+
+    // Add the new widgets.
```
Mm2PL, line 636: I want this to select the first widget if the window already has data in it.

Mm2PL (changes requested):
This is pretty damn cool.

I'm not so sure about the current behavior of selecting linking to channels. Plugins have to give each channel a unique name and hope it's not annoying to the user. The annoyance maybe could be resolved by allowing to also control the display name.
Consider a channel arguments=`{"name": "#test"}` from provider `localhost:6667 (IRC)`. Logically that channel would be named `#test` but there could be another `#test` from a different provider or from Twitch and all links (from /mentions or plugins) will be ambiguous.

Nerixyz:
> Plugins have to give each channel a unique name and hope it's not annoying to the user. The annoyance maybe could be resolved by allowing to also control the display name.
> Consider a channel arguments=`{"name": "#test"}` from provider `localhost:6667 (IRC)`. Logically that channel would be named `#test` but there could be another `#test` from a different provider or from Twitch and all links (from /mentions or plugins) will be ambiguous.

This is a problem I'm aware of. There are some ideas on how to solve this. All these should be able to be added in a followup.

1. When `ChannelView` resolves links in the context of the platform of the channel its viewing. This is the easiest to implement, but it has downsides. The main issue is in channels that aggregate messages from other channels. 
1. Attach a platform to every message. Either as a numeric (non-persistent) ID or as a string. `ChannelView` would first try to resolve the name based on that platform.
1. Attach a (weak) source channel to every message. Similar to the idea above except that we already have the channel.

PR #6948, opened 2026-04-26
Show participants in a shared chat session in the split header

### LINKED ISSUE #6934: Show shared chat participants in Split Header
### Show Participant Names in Split Header

Currently, you have to hover the badges in the chat to see which streamers are part of the shared chat session. We could try to show the participants of a shared chat session in the split header. For example: `streamer (participant1, participant2)`.

We will probably need to poll [Get Shared Chat Session](https://dev.twitch.tv/docs/api/reference/#get-shared-chat-session) for this instead of using EventSub, because of the current subscription limits.

I'm not sure if we can do this in all Stream Together sessions (I don't think we can get the info from Twitch).

_Originally posted by @Nerixyz in https://github.com/Chatterino/chatterino2/discussions/5756#discussioncomment-15530430_

invisibl3cat:
This is an attempt to address https://github.com/Chatterino/chatterino2/issues/6934

The current logic avoids having to subscribe to EventSub by doing this:
When a shared chat message is detected, a [Get Shared Chat Session](https://dev.twitch.tv/docs/api/reference/#get-shared-chat-session) request is made. If the response returns an active shared chat session, a 60 second timer is started that periodically refreshes the state of the session. When no shared chat session is returned, the timer is stopped.

Besides this needing a bit more testing, there are a few optimizations that should be done before merging this. When a shared chat session is detected, this does another Twitch API call to translate broadcaster IDs to their display names. It is unnecessary to do this on every refresh because the participants in the shared chat are not likely to change very often.
Another thing to add could be some rate-limiting when the first API call to check for a shared chat session fails. Current code would just blast the API with every new shared chat message...

ATM I'd like to get some feedback on whether it is a workable solution to detect and display shared chat participants. Thanks! 👍


src/providers/twitch/api/HelixEnums.hpp
```diff
@@ -303,4 +303,14 @@ enum class HelixCreateEventSubSubscriptionError : std::uint8_t {
     Forwarded,
 };
 
+enum class HelixGetSharedChatSessionError {
```
pajlada:
```suggestion
enum class HelixGetSharedChatSessionError : std::uint8_t {
```

src/widgets/splits/SplitHeader.cpp
```diff
@@ -199,7 +200,14 @@ auto formatTitle(const TwitchChannel::StreamStatus &s, Settings &settings)
...
     {
-        title += " (live)";
+        if (sharedChatParticipants.isEmpty())
+        {
+            title += " (live)";
+        }
+        else
+        {
+            title += " (live with " + sharedChatParticipants.join(", ") + ")";
+        }
```
pajlada, line 211: lgtm

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2399,4 +2404,95 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+    getHelix()->getSharedChatSession(
+        this->roomId(),
+        [this](const HelixSharedChatSession &session) {
+            if (session.participantIds.empty())
+            {
+                this->nextSharedChatSessionUpdate_.stop();
+
+                this->sharedChatSessionParticipants_.clear();
+                this->sharedChatStatusChanged.invoke(
+                    this->sharedChatSessionParticipants_);
```
pajlada: return & remove the else below to make the scope a bit more bearable

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2399,4 +2404,95 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+{
+    if (!this->nextSharedChatSessionUpdate_.isActive())
+    {
+        this->refreshSharedChatSessionState();
+    };
+}
+
+void TwitchChannel::refreshSharedChatSessionState()
+{
+    getHelix()->getSharedChatSession(
```
pajlada, line 2556: i think you need to pass along `this` as a weak pointer - see how the other helix requests in here do it in `refreshBadges`

pajlada (commented):
Helix implementation is fine, the way you're making the request from TwitchChannel is mostly fine so this is an OK approach 👍

invisibl3cat:
Thanks for taking a look at this!

Besides addressing your comments, I've:
- Added a check if the list of participants changed and only call `fetchUsers()` when it did.
- Added rate-limiting of the initial probe so it doesn't get triggered with every shared chat message when we don't have any information about if there is a shared chat session or not.

@Nerixyz had some concerns about the use of QTimer for checking on the shared chat session state. Is there anything else that I could use?

There is also a kinda awkward issue that causes `IrcMessageHandler` tests to fail because `probeSharedChatSession()` attempts to call uninitialized Helix object. Using `isTest()` fixes this but it feels more like a workaround. Is there a better place where to call `probeSharedChatSession()` where we wouldn't run into this issue?

Nerixyz:
> @Nerixyz had some concerns about the use of QTimer for checking on the shared chat session state. Is there anything else that I could use?

I think this is fine now.

pajlada:
> There is also a kinda awkward issue that causes `IrcMessageHandler` tests to fail because `probeSharedChatSession()` attempts to call uninitialized Helix object. Using `isTest()` fixes this but it feels more like a workaround. Is there a better place where to call `probeSharedChatSession()` where we wouldn't run into this issue?

You could initialize a mock helix object, but there's no way in snapshots to interface with the mocks so you'd need to specify "it's ok for this function to be called 0 or more times" which is also a bit ugly, but at least it moves the ugliness to the test case. If you choose to do this, add some comment for why that's there / or just say that it's ugly

invisibl3cat:
> You could initialize a mock helix object, but there's no way in snapshots to interface with the mocks so you'd need to specify "it's ok for this function to be called 0 or more times" which is also a bit ugly, but at least it moves the ugliness to the test case. If you choose to do this, add some comment for why that's there / or just say that it's ugly

Initing a mock helix object for this should be sufficient I think. The tests now pass without any weird hacks. Would be nice is somebody else other than me tested this but otherwise I think this should be okay.

tests/src/IrcMessageHandler.cpp
```diff
@@ -561,6 +563,8 @@ class TestIrcMessageHandlerP : public ::testing::TestWithParam<QString>
 
         this->mockApplication->twitch.mockChannels.emplace(
             "twitchdev", this->twitchdevChannel);
+
+        Helix::initialize();
```
pajlada: This initializes real helix, so running tests will now potentially make network requests to Twitch's API in this test. See `tests/src/Commands.cpp` for an example of a test that mocks the Helix API

pajlada (changes requested):

src/providers/twitch/api/Helix.hpp
```diff
@@ -575,6 +575,21 @@ struct HelixPredictions {
     }
 };
 
+struct HelixSharedChatSession {
+    QStringList participantIds;
+
+    explicit HelixSharedChatSession(const QJsonObject &jsonObject)
+    {
+        const auto &participants = jsonObject.value("participants").toArray();
```
Nerixyz, line 616:
```suggestion
        const auto participants = jsonObject.value("participants").toArray();
```

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2399,4 +2405,137 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+    }
+}
+
+void TwitchChannel::refreshSharedChatSessionState()
+{
+    getHelix()->getSharedChatSession(
+        this->roomId(),
+        [this,
+         weak = weakOf<Channel>(this)](const HelixSharedChatSession &session) {
+            if (session.participantIds.empty())
```
Nerixyz: Use `this->weak_from_this()` instead of `weakOf` and check if the weak pointer is still alive in the lambda (`weak.lock()`).

tests/src/IrcMessageHandler.cpp
invisibl3cat: Ouch... I didn't realize this. I think I've figured out a way how to make the tests pass. It really isn't particularly neat but it's all isolated to the IrcMessageHandler test cases.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2399,4 +2405,137 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+
+            if (!participantsDiffer)
+            {
+                return;
+            }
+
+            getHelix()->fetchUsers(
+                session.participantIds, {},
+                [this, weak = weakOf<Channel>(this)](const auto &users) {
+                    this->sharedChatSessionParticipants_.clear();
```
Nerixyz: See above

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1635,6 +1635,14 @@ void GeneralPage::initLayout(GeneralPageView &layout)
                      "double-clicked")
         ->addTo(layout);
 
+    SettingWidget::intInput(
+        "Shared chat session status refresh interval (requires restart)",
```
Nerixyz: You can listen to the setting in the channel and update the timer interval. Then you don't need to restart the client.

tests/src/IrcMessageHandler.cpp
```diff
@@ -599,6 +612,18 @@ TEST_P(TestIrcMessageHandlerP, Run)
...
+    // and that we expect the method call to be made. Luckily we can identify
+    // the tests that need this special handling by the name of the snapshot.
+    // Not great, not terrible...
+    testing::StrictMock<mock::Helix> mockHelix;
+    if (makesGetSharedChatSessionCall(*snapshot))
+    {
+        initializeHelix(&mockHelix);
+        EXPECT_CALL(mockHelix, getSharedChatSession).Times(1);
+    }
+
```
Nerixyz: It would be great if we could avoid this. I'd be fine with checking `getApp()->isTest()` in `refreshSharedChatSessionState`. Or, always initializing helix and accepting any amount of calls (see `tests/src/Plugins.cpp` for an example). You have the downside that in case of test failures, you might get spammed with unrelated warnings from gmock.

src/providers/twitch/TwitchChannel.cpp
Nerixyz: I think we should add a separate `weakFromThis` for `TwitchChannel` so that we can directly get a `std::shared_ptr<TwitchChannel>` rather than `std::shared_ptr<Channel>`.

src/widgets/settingspages/GeneralPage.cpp
invisibl3cat: That would not work here. The code path that sets the timer interval is entered only when the shared chat session state changes because there is no reason to do that otherwise. I don't think that adding additional logic that could fiddle with timeout of a potentially running timer is worth the convenience. The intent behind this toggle was to make it easier to find a reasonable default value for the refresh interval.

Nerixyz: You can set the `interval` on a `QTimer` while it's running ([docs](https://doc.qt.io/qt-6/qtimer.html#interval-prop)). And if it's not running, it doesn't matter because we always overwrite it.

src/providers/twitch/api/Helix.hpp
invisibl3cat: Why? `toArray()` returns QJsonArray, not a Ref object. All other c-tors in Helix.hpp are explicitly taking a ref here. Or am I looking at the wrong thing here?

src/providers/twitch/TwitchChannel.cpp
invisibl3cat: > Use `this->weak_from_this()` instead of `weakOf` and check if the weak pointer is still alive in the lambda (`weak.lock()`).

Ok... thanks.

 Would it make sense to unify how the individual methods in `TwitchChannel.cpp`  get the weak pointer? ATM there are 4(?) slightly different approaches to the intent behind them gets a bit unclear.

pajlada: It's a good idea to unify them so they all use `this->weak_from_this` as Nerixyz points out - but not in this PR please.

tests/src/IrcMessageHandler.cpp
pajlada: I'm fine with this conditional approach, but you should be able to initialize mock helix & the mock inside the test fixture `SetUp` function. You have access to the snapshot there and can avoid duplicating the functionality in this & the `CloneElements` test.

If you can be bothered, you can also instead read a json parameter from the snapshot from something like a "helixExpectations" key & make expectations based on that. Not necessary for this PR though.

src/providers/twitch/TwitchChannel.cpp
Nerixyz: > Would it make sense to unify how the individual methods in `TwitchChannel.cpp` get the weak pointer? ATM there are 4(?) slightly different approaches to the intent behind them gets a bit unclear.

Yes, in a separate PR. But you shouldn't use `weak_from_this` directly. That gives you a `weak_ptr<Channel>`. You can add a `weakFromThis` that returns a `weak_ptr<TwitchChannel>`. It might be easier if you pair it with a `sharedFromThis`:

```cpp
std::shared_ptr<TwitchChannel> TwitchChannel::sharedFromThis() {
  return std::static_pointer_cast<TwitchChannel>(this->shared_from_this());
}
std::weak_ptr<TwitchChannel> TwitchChannel::weakFromThis() {
  return this->sharedFromThis();
}
```

invisibl3cat: > It's a good idea to unify them so they all use `this->weak_from_this` as Nerixyz points out - but not in this PR please.

Sure. I can submit a janitorial PR for this once this is done. Are you okay with leaving things as they are now and cleaning it up later?

pajlada: This is fine as-is for this PR yes

pajlada: Nerixyz's latest comment is what makes most sense for the follow-up janitorial PR

src/widgets/settingspages/GeneralPage.cpp
invisibl3cat: I'm slightly worried that changing the interval of an active timer stops and starts it again because the code relies on `isActive()` to check if there is a shared chat session going on. It would probably be fine and I can add code for this... but is it worth the additional wiring?

tests/src/IrcMessageHandler.cpp
invisibl3cat: I've adjusted the snapshots to include information about whether Helix is needed. It's probably the least awkward solution. Thanks for suggesting this!

invisibl3cat:
Needed a rebase against master

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1635,6 +1635,14 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        s.sharedChatSessionRefreshInterval,
+        {.min = 1, .max = 999, .singleStep = 1, .suffix = "s"})
```
pajlada: Clamp to something like 5 seconds minimum instead of 1

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2545,4 +2551,151 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+        this->refreshSharedChatSessionState();
+    }
+}
+
+void TwitchChannel::refreshSharedChatSessionState()
+{
+    getHelix()->getSharedChatSession(
+        this->roomId(),
+        [this,
+         weak = this->weak_from_this()](const HelixSharedChatSession &session) {
```
pajlada: If you merge from master, you can use `this->weakFromThis` instead to get a nicer pointer you don't have to dynamic-pointer-cast, here and in the getUsers call

src/providers/twitch/api/Helix.hpp
pajlada: Technically the other places are wrong, although this is not the end of the world. Feel free to apply Nerixyz's suggestion.

lib/magic_enum
pajlada, line 1: Undo the change to the magic_enum library

invisibl3cat: Oops... it crept in somehow, sorry about that. Should be fine now.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -184,6 +185,11 @@ TwitchChannel::TwitchChannel(const QString &name)
     });
     this->threadClearTimer_.start(5 * 60 * 1000);
 
+    QObject::connect(&this->nextSharedChatSessionUpdateTimer_, &QTimer::timeout,
```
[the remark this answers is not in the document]
pajlada: @invisibl3cat

src/widgets/settingspages/GeneralPage.cpp
pajlada: It looks like this works smoothly, so you can implement this @invisibl3cat

pajlada (approved):
The "Live with ..." text doesn't support your preferred display name setting, however that's best left for a follow-up PR after the "pinned message" PR has been merged in, which solved this with a nice helper function.

PR #6950, opened 2026-04-26
fix: check for `isAppAboutToQuit` in more places

Nerixyz:
When closing Chatterino with a popup window open, we go through the procedure handing `QApplication::aboutToQuit` [here](https://github.com/Chatterino/chatterino2/blob/8573688f3c3b5198f86d16f154789506b767023e/src/RunGui.cpp#L281-L290). There, we call `app->stop()`, which clears all components. However, the app instance from `getApp` or `tryGetApp` is still valid.

This feels more like a band-aid fix - we might want to clear the `INSTANCE` variable in `Application.cpp` when calling `stop()` immediately.

### DIFF
--- a/src/common/Channel.cpp
+++ b/src/common/Channel.cpp
@@ -58,7 +58,7 @@ Channel::Channel(const QString &name, Type type)
 Channel::~Channel()
 {
     auto *app = tryGetApp();
-    if (app && this->anythingLogged_)
+    if (app && !isAppAboutToQuit() && this->anythingLogged_)
     {
         app->getChatLogger()->closeChannel(this->name_, this->platform_);
     }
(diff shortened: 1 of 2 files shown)


pajlada (approved):
Yeah it's definitely a bit of a bandaid fix - the current logic is somewhat like:
1. User initiates a close
2. We mark the application as "we are about to quit" as a state of "don't start making new requests, maybe don't save certain things"
3. Application actually starts tearing itself down.

Some of the problems come from certain things living longer than the application (namely images & channels last I checked).

I think some of this can be improved by reducing our use of `getApp`, especially in destructors, that way we'd get a clearer view of what depends on what.

PR #6953, opened 2026-04-27
Add native messaging manifest for Microsoft Edge

### LINKED ISSUE #6836: Create Native Messaging Manifest for MS Edge
MS Edge isn't supported right now, because we don't create a manifest for it:

> 🤔Wasn't working for me in windows with edge, ffz & the chatterino native host plugin. I haven't used the feature before though. In the end only a nice to have I suppose. 

 _Originally posted by @Wynadorn in [#6489](https://github.com/Chatterino/chatterino2/issues/6489#issuecomment-4007443375)_

We create the manifests here: https://github.com/Chatterino/chatterino2/blob/de9496a8d7a0aa1ef3611633136db67a9d7f87a5/src/singletons/NativeMessaging.cpp#L51-L77

[Microsoft Docs](https://learn.microsoft.com/en-us/microsoft-edge/extensions/developer-guide/native-messaging?tabs=v3%2Cwindows) show the locations that Edge expects. Since it's the default browser on Windows, I think it'd be good to handle it. Should be enough to only handle it on Windows.

kill74:
Fixes #6836

## Summary

Adds Microsoft Edge native messaging host registration on Windows.

Edge is Chromium-based and uses the same native messaging manifest format as Chrome, so this reuses the existing Chrome manifest generation and registers it under Edge's native messaging registry path:

`HKCU\Software\Microsoft\Edge\NativeMessagingHosts\com.chatterino.chatterino`

## Testing

- Not built locally
- Verified the Edge registration is guarded behind `Q_OS_WIN`
- Verified Edge reuses the existing Chromium native messaging manifest generation path

### DIFF
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -77,6 +77,16 @@ const Config CHROME{
 #endif
 };
 
+// Edge is Chromium-based and uses the same native messaging manifest format
+// as Chrome. The registration path below is Windows-specific.
+#ifdef Q_OS_WIN
+const Config EDGE{
+    .fileName = u"native-messaging-manifest-edge.json"_s,
+    .registryKey =
+        u"HKCU\\Software\\Microsoft\\Edge\\NativeMessagingHosts\\com.chatterino.chatterino"_s,
+};
+#endif
+
 void registerNmManifest([[maybe_unused]] const Paths &paths,
                         const Config &config, const QJsonDocument &document)
 {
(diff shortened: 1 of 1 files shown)


pajlada:
Could you test this locally to confirm it works as expected? It looks fine

kill74:
Tested this locally on Windows and it works as expected.

What I tested:
- Built Chatterino from this branch
- Removed the existing Edge native messaging registry key before running it
- Started the built `chatterino.exe`
- Confirmed that the Edge registry key was created again at:
  `HKCU\Software\Microsoft\Edge\NativeMessagingHosts\com.chatterino.chatterino`
- Confirmed that the registry value points to:
  `C:/Users/Guilherme Sales/AppData/Roaming/Chatterino2/Misc/native-messaging-manifest-edge.json`
- Confirmed that the manifest file exists
- Confirmed that the manifest contains the expected extension origin:
  `chrome-extension://glknmaideaikkmemifbfkhnomoknepka/`
- Confirmed that the manifest `path` points to the built executable:
  `C:/Users/Guilherme Sales/Documents/Projects/chatterino2/build/bin/chatterino.exe`

So from my local test, the Edge native messaging registration is working.

PR #6954, opened 2026-04-27
ci: use official download-artifact action

### LINKED ISSUE #6682: ci: investigate github script + actions/download-artifact usability
Currently, we make use of the `dawidd6/action-download-artifact` action. We should investigate the https://github.com/actions/download-artifact `run-id` parameter to see if it, combined with https://github.com/actions/github-script could replace it.

kill74:
Fixes #6682.

## Summary

This PR replaces `dawidd6/action-download-artifact` in the installer workflow with the official `actions/download-artifact` action.

The workflow now handles both ways the installer workflow can be started:

- For `workflow_run`, it downloads the artifact from the triggering Build workflow run.
- For `workflow_dispatch`, it uses `actions/github-script` to find a successful Build workflow run for the current commit, then downloads the artifact from that run.

The artifact name and download path are kept unchanged.

## Testing

I verified the workflow changes manually by checking that:

- The installer workflow still downloads `chatterino-windows-x86-64-Qt-${{ matrix.qt-version }}.zip`.
- The artifact is still downloaded into `build/`.
- The manual `workflow_dispatch` path keeps the previous commit-based lookup behavior.

The full workflow behavior should be validated by GitHub Actions once this PR runs.


pajlada (changes requested):
You've included changes from your other PR

kill74:
Thanks, that was my mistake. I cleaned up the branch so this PR now only contains the installer workflow change.

.github/workflows/create-installer.yml
```diff
@@ -24,12 +24,51 @@ jobs:
...
           name: chatterino-windows-x86-64-Qt-${{ matrix.qt-version }}.zip
-          commit: ${{ github.sha }}
+          github-token: ${{ github.token }}
+          run-id: ${{ github.event.workflow_run.id }}
+          path: build/
+
+      - name: Find Build workflow run
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        id: find-build-run
+        uses: actions/github-script@v9
```
Nerixyz, line 34: This should work in the `workflow_run` case too, no? Then you could remove the case for `workflow_run` above. This also makes it easier to change the downloaded artifact as you don't have to update two places (e.g. if we do WOA builds).

kill74: Good point, thanks. I updated this so the workflow resolves the Build workflow run ID first for both trigger types, then uses a single `actions/download-artifact` step.

pajlada (approved):

PR #6958, opened 2026-04-28
feat: store user aliases by Twitch user ID

### LINKED ISSUE #6345: User-based nicknames
### Discussed in https://github.com/Chatterino/chatterino2/discussions/6195

<sup>Originally posted by **MrChuw** May  7, 2025</sup>
As the title suggests, use UserID instead of User Names to give nicknames, as there are users who are always changing their name.
</div>

Currently, you can create nicknames for user based on the usernames. It should be possible to create nicknames based on the user's account (i.e. their Twitch User ID).
We support attaching data to users based on their IDs using the UserDataController, this can be used to store user-based-aliases.

There are multiple possible ways a user could interact with the feature:
1) Add a chat command similar to `/unstable-set-user-color `
2) Modify user alias from user info popup
3) See & modify user aliases in settings dialog

For an initial implementation, I only need to see 2 (and optionally 1) implemented. 3) brings with it a lot of complexity, and requires more thought before implementation.

kill74:
Fixes #6345

## Summary

Adds user aliases backed by `UserData`, keyed by Twitch user ID instead of username.

Changes:
- Stores an optional `alias` field in `user-data.json`.
- Adds `UserDataController::setUserAlias`.
- Applies user ID aliases before username-based nicknames when rendering usernames.
- Adds a **Set alias** action to the user info popup.
- Updates the user data mock for tests.

## Testing

Tested locally on Linux.

1. Built Chatterino successfully.
2. Opened a Twitch channel.
3. Opened the user info popup for a chat user.
4. Clicked **Set alias**.
5. Entered a custom alias.
6. Confirmed new messages from that user render with the alias.
7. Restarted Chatterino.
8. Confirmed the alias persisted.
9. Cleared the alias.
10. Confirmed new messages fall back to the normal username behavior.

The settings-dialog management UI is intentionally not included because the initial implementation only requires editing aliases from the user info popup.

Co-authored-by: -
Tested-by: -
Reported-by: -
Reviewed-by: -
Parent-pr: -

### DIFF
--- a/mocks/include/mocks/UserData.hpp
+++ b/mocks/include/mocks/UserData.hpp
@@ -11,29 +11,62 @@ class UserDataController : public IUserDataController
 public:
     UserDataController() = default;
 
-    // Get extra data about a user
-    // If the user does not have any extra data, return none
+    // Get extra data about a user.
+    // If the user does not have any extra data, return none.
     std::optional<UserData> getUser(const QString &userID) const override
     {
-        auto it = this->userMap.find(userID);
+        const auto it = this->userMap.find(userID);
         if (it != this->userMap.end())
         {
             return it->second;
         }
+
         return std::nullopt;
     }
 
-    // Update or insert extra data for the user's color override
+    // Update or insert extra data for the user's color override.
     void setUserColor(const QString &userID,
                       const QString &colorString) override
     {
+        if (userID.isEmpty())
+        {
+            return;
+        }
+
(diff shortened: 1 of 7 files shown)


Mm2PL (changes requested):
This duplicates the nicknames feature while also breaking regex functionality. I don't see any good reason to outright replace that. Please stop with the AI-gen descriptions.

PR #6961, opened 2026-04-29
feat(plugins): Add signal for context menus

Nerixyz:
This PR adds the ability for plugins to add actions to context menus of ChannelViews. They can add actions, submenus and separators.

I added the signal to the `WindowManager` to allow plugins to listen to these events in any channel.

Here's an example plugin:

```lua
c2.windows:on_channelview_context_menu_requested(function(args)
	args.menu:add_action("Testing", function()
		args.channel:add_system_message("Clicked!")
	end)
	args.menu:insert_action(1, "First!", function()
		local msg = "text='" .. args.message.message_text .. "'"
		if args.message_element then
			msg = msg .. " element=" .. args.message_element.type
		end
		if args.split then
			msg = msg .. " split={channel=" .. args.split.channel:get_name() .. "}"
		end
		args.channel:add_system_message(msg)
	end)
	local sub = args.menu:add_menu("Submenu")
	sub:add_action("An action", function() end)
	sub:add_separator()
	sub:add_action("Second one", function() end)
end)
```


src/controllers/plugins/api/QMenuWrap.hpp
```diff
@@ -0,0 +1,56 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+#ifdef CHATTERINO_HAVE_PLUGINS
+#    include <sol/forward.hpp>
+
+namespace chatterino::lua::api::qmenu {
```
pajlada: somewhat similar to my feedback on the wip datetime stuff, I would prefer we don't include our implementation details in the plugin API.
Practically, this would mean calling this `menu` and `c2.Menu` instead of `qmenu` / `c2.QMenu`.

The functions on `c2.Menu` can mirror `QMenu` where it makes sense of course

src/singletons/WindowManager.hpp
```diff
@@ -162,6 +164,10 @@ class WindowManager final : public QObject
     pajlada::Signals::Signal<SplitContainer *> selectSplitContainer;
     pajlada::Signals::Signal<const MessagePtr &> scrollToMessageSignal;
 
+    pajlada::Signals::Signal<const ChannelView &, const MessageLayout &,
```
pajlada, line 177: A little comment for this signal would be nice

src/widgets/helper/ChannelView.cpp
```diff
@@ -2605,6 +2613,11 @@ void ChannelView::addContextMenuItems(
     // Add executable command options
     this->addCommandExecutionContextMenuItems(menu, layout);
 
+    menu->addSeparator();
```
pajlada, line 2642: I haven't tested this - is a separator at the end of a menu with no further menu items hidden? Thinking for the common use-case of no plugins adding any right-click entries

src/controllers/plugins/api/Menu.cpp
```diff
@@ -0,0 +1,101 @@
...
+}  // namespace
+
+namespace chatterino::lua::api::qmenu {
+
+void createUserType(sol::table &c2)
+{
+    c2.new_usertype<QMenu>(
+        "QMenu", sol::no_constructor,  //
+
+        "add_action",
```
Mm2PL, line 60: imo this should either be our own api or identically mirror the names of Qt functions its binding to

src/controllers/plugins/api/QMenuWrap.hpp
Nerixyz: Done in [`04f4367` (this PR)](https://github.com/Chatterino/chatterino2/pull/6961/commits/04f436774b426062ff2d75973210da6eb1ca4623).

src/singletons/WindowManager.hpp
Nerixyz: Done in [`318bf64` (this PR)](https://github.com/Chatterino/chatterino2/pull/6961/commits/318bf647c6d4dc0f3d79bfc1ccc40888542cc337).

src/widgets/helper/ChannelView.cpp
Nerixyz: From my testing, this doesn't cause any differences. Separators are already deduplicated in menus, so I'd expect this to also handle separators at the end.

src/controllers/plugins/api/Menu.cpp
Nerixyz: Could you elaborate on this? As per https://github.com/Chatterino/chatterino2/pull/6961#discussion_r3167225305, I changed the name to `Menu`.

src/controllers/plugins/api/Menu.hpp
```diff
@@ -0,0 +1,56 @@
...
+---A generic menu used for context menus.
+---@class c2.Menu
+c2.Menu = {}
+
+---Appends a new action to the menu.
+---@param text string
+---@param cb fun()
+function c2.Menu:add_action(text, cb) end
+
+---Inserts an action named `text` after `before`. If `before` is not found,
```
pajlada: It actually inserts before `before`, which matches Qts implementation.

src/controllers/plugins/api/Menu.hpp
```diff
@@ -0,0 +1,56 @@
...
+---@param text string
+---@param cb fun()
+function c2.Menu:insert_action(before, text, cb) end
+
+---Appends a new Menu with `title` to the menu.
+---@param title string
+---@return c2.Menu
+function c2.Menu:add_menu(title) end
+
+---Inserts a new Menu named `title` after `before`. If `before` is not found,
```
pajlada: It actually inserts before `before`, which matches Qts implementation.

src/controllers/plugins/api/Menu.hpp
```diff
@@ -0,0 +1,56 @@
...
+---the menu is inserted at the end. `before` can either be a name or a one-based
+---index.
+---@param before string|integer A name or index of an action.
+---@param title string
+function c2.Menu:insert_menu(before, title) end
+
+---Appends a new separator.
+function c2.Menu:add_separator() end
+
+---Inserts a new separator after `before`. If `before` is not found,
```
pajlada: It actually inserts before `before`, which matches Qts implementation.

pajlada (changes requested):

src/controllers/plugins/api/WindowManager.cpp
```diff
@@ -178,8 +184,54 @@ void createUserTypes(sol::table &c2)
...
+                                ref.emplace(msg, i);
+                                break;
+                            }
+                        }
+                    }
+                    std::optional<ChannelRef> maybeChan;
+                    if (auto underlyingChan = view.underlyingChannel())
+                    {
+                        maybeChan.emplace(std::move(underlyingChan));
+                    }
```
pajlada, line 228: Can you group up the various parameters with a newline inbetween?

i.e.

```
auto msg = ...;

auto ref = ...;

auto maybeChan = ...;
```

pajlada (approved):
This works as expected for me, with just a styling nit left that I don't really care if it's left unfixed.

src/controllers/plugins/api/Menu.cpp
pajlada: I believe this is resolved

Mm2PL (approved):

PR #6967, opened 2026-05-01
Add undo/redo hotkeys

### LINKED ISSUE #6817: Add Undo and Redo to Hotkeys
### Checklist

### Describe your issue

In Chatterino, the keyboard shortcuts for actions Undo and Redo are bound to CTRL+Z and CTRL+SHIFT+Z respectively, without the ability to change this in the Hotkeys tab. They're possibly taken from the Qt widget.

I would like if these were added to the Hotkeys tab, so that they can be rebound per user preference.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.4 (commit [117d06b92](https://github.com/Chatterino/chatterino2/commit/117d06b92)) built with Qt 6.10.1 (running on 6.10.2)
Running on KDE Flatpak runtime, kernel: 6.18.10-200.fc43.x86_64, running from Flatpak

Kehnanmind:
Fixes #6817

Adds Undo and Redo to the Hotkeys menu, allowing users to rebind them according to their preferences.
image

By default, `Ctrl+Z`and `Ctrl+Shift+Z` are handled by Qt. This means that even if these actions are rebound, the default shortcuts will still trigger undo/redo.

To fix this, the default Qt undo/redo shortcuts are disabled, ensuring that these actions are handled only through the configured hotkeys (A similar thing is already done for `Ctrl+C`)

### DIFF
--- a/src/controllers/hotkeys/HotkeyController.cpp
+++ b/src/controllers/hotkeys/HotkeyController.cpp
@@ -481,6 +481,13 @@ void HotkeyController::addDefaults(std::set<QString> &addedHotkeys)
         this->tryAddDefault(addedHotkeys, HotkeyCategory::SplitInput,
                             QKeySequence("Down"), "nextMessage",
                             std::vector<QString>(), "next message");
+
+        this->tryAddDefault(addedHotkeys, HotkeyCategory::SplitInput,
+                            QKeySequence("Ctrl+Z"), "undo",
+                            std::vector<QString>(), "undo");
+        this->tryAddDefault(addedHotkeys, HotkeyCategory::SplitInput,
+                            QKeySequence("Ctrl+Shift+Z"), "redo",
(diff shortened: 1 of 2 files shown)


pajlada (commented):
Hi, the hotkey you added targets the action `undo` and `redo`, but those actions are not defined. See https://github.com/Chatterino/chatterino2/pull/5856 as an example for what further actions you need to implement


The bindings you've blocked (CTRL+Z and CTRL+SHIFT+Z) are only some of the undo/redo bindings. See https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts

I'm not sure if there's a reasonable way to handle this for all platforms - open to discussions. What is your intent / goal with this change? Are you looking to bind another action to CTRL+Z/CTRL+SHIFT+Z?

Issue #6931, opened 2026-04-19
Favourite emotes


pajlada:
### Discussed in https://github.com/Chatterino/chatterino2/discussions/4317

<div type='discussions-op-text'>

<sup>Originally posted by **buzz** January 20, 2023</sup>
Make it possible to fav emotes with Ctrl+mouse click or similar. They should appear in a separate emote tab for quick access.

Only mention of this feature I found: #33</div>

It should be possible to favourite an emote in the emote popup window.

Once you have favourited one or more emotes, those emotes should show up in a new tab and/or at the top of the emote's own tab. Happy to explore either or both options.

Favourited emotes should be marked as such in their original section with an outline or star to indicate that they are favourited.

PR #6968, opened 2026-05-01
feat: Favourite emotes

invisibl3cat:
A proposal how to implement favourite emotes propsed in #6931 .

This adds a new "Favourite" tab that can be filled with emotes by Ctrl+clicking them in other tabs.

- Favourite emotes can be unfavourited by Ctrl+clicking them in the "Favourite" tab.
- Ctrl+Shift+clicking makes an emote favourite and also inserts it in the chat.
- <s>Favourite emotes are persisted as a semicolon-separated string in settings. Is there a better way we could do this?</s> String was replaced by a vector
- Currently only works for emotes and not emojis. Do we want to have favourite Emojis too?
- Should the "Favourite" tab be selected by default? Currently it is not.

image

While working on this I came across two things that I found curious:
- Why does https://github.com/Chatterino/chatterino2/blob/7f00e801f48212702907830fcccf9b2c2fcc430d/src/widgets/dialogs/EmotePopup.cpp#L53 take the vector as a value? This forces a copy unless the caller can move the vector to the function.

- Is there a reason to store Twitch emotes in a vector and not an EmoteMap? The lookup code that translates the emote name to favourite emote needs to iterate over all Twitch emotes which could be a bit slow.


Nerixyz:
> * take the vector as a value? This forces a copy unless the caller can move the vector to the function.

It's a bit convoluted, but the function needs to take _some_ sortable range. Since most (if not all?) emotes are stored in maps, we use a vector. This could also be a `std::vector&`.

> * Is there a reason to store Twitch emotes in a vector and not an EmoteMap?

Not really. Where do we do this?

invisibl3cat:
> > * take the vector as a value? This forces a copy unless the caller can move the vector to the function.
> 
> It's a bit convoluted, but the function needs to take _some_ sortable range. Since most (if not all?) emotes are stored in maps, we use a vector. This could also be a `std::vector&`.

Never mind, I'm dumb. The copy is obviously needed there.

> 
> > * Is there a reason to store Twitch emotes in a vector and not an EmoteMap?
> 
> Not really. Where do we do this?

Here

https://github.com/Chatterino/chatterino2/blob/7f00e801f48212702907830fcccf9b2c2fcc430d/src/providers/twitch/TwitchEmotes.hpp#L45

This struct stores EmotePtrs as a vector whereas other emote sources use EmoteMap.

Nerixyz:
> This struct stores EmotePtrs as a vector whereas other emote sources use EmoteMap.

It stores the emotes as a list, because we only ever use it in the emote popup. If it's easier for this to be an `EmoteMap`, you can change this.

invisibl3cat:
> > This struct stores EmotePtrs as a vector whereas other emote sources use EmoteMap.
> 
> It stores the emotes as a list, because we only ever use it in the emote popup. If it's easier for this to be an `EmoteMap`, you can change this.

It turns out that this vector is sorted by emote names so just replacing it with EmoteMap would likely change the display order of the Twitch emotes. I suppose I can leave it as it is for now and address it in another PR if it turns out that it makes sense to do so?

Nerixyz:
> It turns out that this vector is sorted by emote names so just replacing it with EmoteMap would likely change the display order of the Twitch emotes.

Every `EmoteMap` that the emote popup displays will be "sorted" (by going through that intermediate vector). So this shouldn't cause any issues wrt sorting.

invisibl3cat:
> > It turns out that this vector is sorted by emote names so just replacing it with EmoteMap would likely change the display order of the Twitch emotes.
> 
> Every `EmoteMap` that the emote popup displays will be "sorted" (by going through that intermediate vector). So this shouldn't cause any issues wrt sorting.

I didn't go particularly deep on this but is this sorting redundant then?
https://github.com/Chatterino/chatterino2/blob/7f00e801f48212702907830fcccf9b2c2fcc430d/src/providers/twitch/TwitchAccount.cpp#L501

Nerixyz:
> I didn't go particularly deep on this but is this sorting redundant then?

Yes, I think so. Since we already sort in the popup.

4rneee:
> This adds a new "Favourite" tab that can be filled with emotes by Ctrl+clicking them in other tabs.

I feel like Ctrl+clicking is not very intuitive. What about a checkbox/button in the right-click menu?

Two other thoughts:
- right now only the name is stored which leads to different emotes appearing in the favorite tab if two channels have different emotes with the same name. I'm not sure if this is intended or what the expected behavior would be.
- There is no way of removing favorite emotes that are no longer available in any channel. Is this a problem?

invisibl3cat:
> > This adds a new "Favourite" tab that can be filled with emotes by Ctrl+clicking them in other tabs.
> 
> I feel like Ctrl+clicking is not very intuitive. What about a checkbox/button in the right-click menu?

Ctrl+click was suggested as one of the triggers in the respective issue. Adding something more discoverable is definitely worth exploring.

> 
> Two other thoughts:
> 
>     * right now only the name is stored which leads to different emotes appearing in the favorite tab if two channels have different emotes with the same name. I'm not sure if this is intended or what the expected behavior would be.
> 

Can that actually happen? AFAIK emotes are not "namespaced" so if two channels tried to use the same name for an emote, it'd collide even on Twitch. I could be wrong, though...

>     * There is no way of removing favorite emotes that are no longer available in any channel. Is this a problem?
Unavailable emotes should be removed automatically. When we're loading emotes, any emote name that cannot be resolved to an emote gets dropped. See the relevant bits in `reloadEmotes()`.

pajlada (commented):
When no favourite emote is added the text under the "Favourites" tab should instruct the user as to how they can favourite an emote

As @4rneee mentioned - a right-click entry to add a favourite emote would be great. Both methods can be mentioned.

Some indicator that an emote has been favourited would be cool but if it's complicated I'm OK to leave that as a future feature. Right now, adding a favourite emote feels "empty". I don't get any feedback that anything happened.

4rneee:
> > ```
> > * right now only the name is stored which leads to different emotes appearing in the favorite tab if two channels have different emotes with the same name. I'm not sure if this is intended or what the expected behavior would be.
> > ```
> 
> Can that actually happen? AFAIK emotes are not "namespaced" so if two channels tried to use the same name for an emote, it'd collide even on Twitch. I could be wrong, though...

Not sure what you mean. I meant third party emotes often having variants with the same name. Example:

Here I added `TRICKLEIN` and `PagMan` in one channel as favorite. In another channel they also appear but are a "different" emote.
Not sure if this is expected or if we would want to differentiate between such variants by for example storing some kind of ID instead.

> > ```
> > * There is no way of removing favorite emotes that are no longer available in any channel. Is this a problem?
> > ```
> 
> Unavailable emotes should be removed automatically. When we're loading emotes, any emote name that cannot be resolved to an emote gets dropped. See the relevant bits in `reloadEmotes()`.

In the popup yes but not from the settings, or am I missing something?
My thought was that if you add emotes and then you close the channel or the emote gets removed these emotes would be "orphaned" and there is no way of removing them from the favorite list through the UI.
This is probably not a big issue but I still wanted to mention it.

invisibl3cat:
> > > ```
> > > * right now only the name is stored which leads to different emotes appearing in the favorite tab if two channels have different emotes with the same name. I'm not sure if this is intended or what the expected behavior would be.
> > > ```
> > 
> > 
> > Can that actually happen? AFAIK emotes are not "namespaced" so if two channels tried to use the same name for an emote, it'd collide even on Twitch. I could be wrong, though...
> 
> Not sure what you mean. I meant third party emotes often having variants with the same name. Example:  Here I added `TRICKLEIN` and `PagMan` in one channel as favorite. In another channel they also appear but are a "different" emote. Not sure if this is expected or if we would want to differentiate between such variants by for example storing some kind of ID instead.
> 

OK... I see what you're getting at. I'm not immediately sure how to resolve this when the appearance of the same emote is channel-dependent. We could store the EmoteSet ID too

> > > ```
> > > * There is no way of removing favorite emotes that are no longer available in any channel. Is this a problem?
> > > ```
> > 
> > 
> > Unavailable emotes should be removed automatically. When we're loading emotes, any emote name that cannot be resolved to an emote gets dropped. See the relevant bits in `reloadEmotes()`.
> 
> In the popup yes but not from the settings, or am I missing something? My thought was that if you add emotes and then you close the channel or the emote gets removed these emotes would be "orphaned" and there is no way of removing them from the favorite list through the UI. This is probably not a big issue but I still wanted to mention it.



> > > ```
> > > * right now only the name is stored which leads to different emotes appearing in the favorite tab if two channels have different emotes with the same name. I'm not sure if this is intended or what the expected behavior would be.
> > > ```
> > 
> > 
> > Can that actually happen? AFAIK emotes are not "namespaced" so if two channels tried to use the same name for an emote, it'd collide even on Twitch. I could be wrong, though...
> 
> Not sure what you mean. I meant third party emotes often having variants with the same name. Example:  Here I added `TRICKLEIN` and `PagMan` in one channel as favorite. In another channel they also appear but are a "different" emote. Not sure if this is expected or if we would want to differentiate between such variants by for example storing some kind of ID instead.
> 

Oh, I see what you're getting at. How about this? We look at which Twitch channel would the Emote be pasted into. If the channel has its own variant for the Emote, we display the image for that variant in the Favourites tab. If the channel does not have such emote, we just display the first candidate we find. AFAICT Twitch resolves these collisions on the backend so we cannot really know for sure which Emote will be displayed in the chat.

I don't think that storing the emote set ID would help because we cannot tell Twitch which variant of the Emote we want to display, can we?

If this makes sense to you, could you take another look when I push this change?

> > > ```
> > > * There is no way of removing favorite emotes that are no longer available in any channel. Is this a problem?
> > > ```
> > 
> > 
> > Unavailable emotes should be removed automatically. When we're loading emotes, any emote name that cannot be resolved to an emote gets dropped. See the relevant bits in `reloadEmotes()`.
> 
> In the popup yes but not from the settings, or am I missing something? My thought was that if you add emotes and then you close the channel or the emote gets removed these emotes would be "orphaned" and there is no way of removing them from the favorite list through the UI. This is probably not a big issue but I still wanted to mention it.

Sorry, the code in `reloadEmotes()` was not saving the newly constructed list of favourite Emotes right away. I've fixed that on my end. That should ensure that there are no orphaned entries.

src/widgets/helper/EmoteChannelView.cpp
Nerixyz, line 1: I think I'd prefer if the context menu customization was done with a signal from a `ChannelView`.
This feels more extensible. Say, if another feature needs to add context menu items to a `ChannelView`, it doesn't need to subclass it but can listen to the signal.

invisibl3cat:
I made a few changes to address the issue of orphaned emotes. I now use the list of emotes from `Settings` as the source of truth which allows us to check any Emotes that are currently unavailable for whatever reason. Those are displayed in a separate list on the `Favourite` tab. The list is channel-aware so it will display some Emotes as unavailable unless the user would paste them into the correct channel. Does that sound reasonable?

"Wrong" channel active
unavailable

"Correct" channel active
unavailable2

invisibl3cat:
@4rneee I've added a bit of logic to check the currently active channel's emote set first. If there are multiple channels sharing the same Emote name, the `Favourite` tab should now display the Emote that is appropriate for the active channel. If this makes sense to you, could you check that it works?

src/widgets/helper/EmoteChannelView.cpp
invisibl3cat: I'm not the biggest fan of the subclassing myself. If I understand your idea, you suggest  that `ChannelView`  emit a signal when it is about to display the context menu so that other parties can change its content?

Nerixyz: Yes, similar to how it was done here: https://github.com/Chatterino/chatterino2/pull/6961

pajlada: A signal `pajlada::Signals::Signal<QMenu*, element and layout if necessary> messageMenuCreated` is fine to add to `ChannelView` and listen to from EmotePopup, called after right before `menu->popup(QCursor::pos());`. No further context should be necessary for consumers of the signal. You do not need to add a separator before emitting the signal

src/singletons/Settings.hpp
```diff
@@ -441,6 +441,11 @@ class Settings
     BoolSetting enableSevenTVEventAPI = {"/emotes/seventv/eventapi", true};
     BoolSetting sendSevenTVActivity = {"/emotes/seventv/sendActivity", true};
 
+    ChatterinoSetting<std::vector<QString>> favouriteEmotes = {
+        "/emotes/favouriteEmotes", {}};
+    ChatterinoSetting<std::vector<QString>> favouriteEmojis = {
+        "/emotes/favouriteEmojis", {}};
```
pajlada: formatting - this will spread the setting out on a line a little better. clang-format will complain/fix this
```suggestion
    ChatterinoSetting<std::vector<QString>> favouriteEmotes = {
        "/emotes/favouriteEmotes", {},};
    ChatterinoSetting<std::vector<QString>> favouriteEmojis = {
        "/emotes/favouriteEmojis", {},};
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -42,6 +42,47 @@ namespace {
...
 
+bool emojiHasShortCode(const EmojiPtr &emoji, const QString &shortCode)
+{
+    auto it = std::ranges::find(emoji->shortCodes, shortCode);
+    return it != emoji->shortCodes.end();
+}
+
+bool isEmojiIdentifier(const QString &identifier)
+{
+    return identifier.startsWith(':') && identifier.endsWith(':') &&
```
pajlada: Unfortunately I don't think this is reliable - some emote providers allow you to use colons throughout the emote, including at the start and end. see `:bingbongbang:` in my channel

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -468,10 +557,152 @@ void EmotePopup::loadChannel(ChannelPtr channel)
...
+    /*
+     * Note that we are checking the persistent list of favourite emote names
+     * rather than the internal vector of favEmotes_. We do this because
+     * in order to populate the favEmotes_ list, we first need to download
+     * all Emotes we have access to from Twitch. This can take considerable
+     * time during which the persistent list and the internal list are
+     * effectively out of sync. If there is a connection issue, these two lists
+     * may not sync up at all. Using the persistent list is the safe choice.
+     */
+    auto emoteNames = Settings::instance().favouriteEmotes.getValue();
```
pajlada: Use `getSettings` instead of `Settings::instance()` - here and in other places

4rneee:
> @4rneee I've added a bit of logic to check the currently active channel's emote set first. If there are multiple channels sharing the same Emote name, the `Favourite` tab should now display the Emote that is appropriate for the active channel. If this makes sense to you, could you check that it works?

As discussed on pajlada's stream: the approach of only storing the emote name is fine. My original comment is therefore no longer a concern.

8thony:
thanks @invisibl3cat for your work,

I have some things I want to put up for consideration/discussion.

> the approach of only storing the emote name is fine

I have some concerns about this:

- What if an emote gets renamed?
- Some channels dislike specific emotes and add an emote with the same name but with a completely different meaning or text, like "I am dumb" instead. So I don't want to have these emotes show up as favorites. 
- Handling emotes that have the same name
	<details>
	<summary>Image</summary>

	

	</details>

But I also see the advantages:
- With different emote sets, like for seasons or events where emotes get a slightly different theme, this could be simpler from the user perspective.
- It is not relevant where the emote was uploaded. If the same emote is uploaded on different providers and different channels use different providers, the user has it in both scenarios in the favorites list.

The FFZ Browser Extension stores the favorite emotes with the emote ID, and if another channel doesn't have the exact same emote, it is not showing up. I like this more because I associate emote names with what I want to express, and if that doesn't match, I'm confused.

> Some indicator that an emote has been favourited would be cool but if it's complicated I'm OK to leave that as a future feature. Right now, adding a favourite emote feels "empty". I don't get any feedback that anything happened.

Something like the FFZ Browser Extension does (little yellow star in the bottom right corner).
This should show up in the emote popup but also in the emote completion popup.

> under the "Favourites" tab should instruct the user as to how they can favourite an emote

👍 I agree this is needed

Additional:
You must consider the next Feature people want is higher tab priorities for favorite emotes. 

If no emote is added yet, two rows of "No favourite emojis, add some" show up as of c86bd0d2

invisibl3cat:
> thanks @invisibl3cat for your work,
> 
> I have some things I want to put up for consideration/discussion.
> 
> > the approach of only storing the emote name is fine
> 
> I have some concerns about this:

We discussed this a bit on the stream and I've made some adjustments to the logic that assigns Emote to its name.

> 
>     * What if an emote gets renamed?

If that happens, it will show up as "Unavailable" and the user will need to fix it up manually. I'm not sure if we can do much better than that.

> 
>     * Some channels dislike specific emotes and add an emote with the same name but with a completely different meaning or text, like "I am dumb" instead. So I don't want to have these emotes show up as favorites.
> 
>     * Handling emotes that have the same name
>         
>         Image
>         
> 
> 
> But I also see the advantages:
> 
>     * With different emote sets, like for seasons or events where emotes get a slightly different theme, this could be simpler from the user perspective.
> 
>     * It is not relevant where the emote was uploaded. If the same emote is uploaded on different providers and different channels use different providers, the user has it in both scenarios in the favorites list.
> 
> 
> The FFZ Browser Extension stores the favorite emotes with the emote ID, and if another channel doesn't have the exact same emote, it is not showing up. I like this more because I associate emote names with what I want to express, and if that doesn't match, I'm confused.
> 

The Name -> Emote assignment logic is channel-aware. It tries to get an Emote for the name in the following order:

1) Local Twitch channel Emotes
2) Emotes for the currently active Twitch channel - the one we'd paste the name into
3) All other Twitch channels available to the user
4) Twitch channel BTTV, FFZ, 7TV Emotes, in this order
5) Global BTTV, FFZ, 7TV Emotes, in this order

It returns the first match it finds. This seems like a reasonable "best effort" approach to try to make sure that the Emote the user sends will be the same Emote the other chatters in the channel would see.

If you could suggest some channels that would allow me to check this behaviour or test it yourself that'd be great!

> > Some indicator that an emote has been favourited would be cool but if it's complicated I'm OK to leave that as a future feature. Right now, adding a favourite emote feels "empty". I don't get any feedback that anything happened.
> 
> Something like the FFZ Browser Extension does (little yellow star in the bottom right corner). This should show up in the emote popup but also in the emote completion popup.
> 

This is planned but we'd like to leave it for a follow-up improvement. The changes we've made so far are already a bit bigger than I expected.

> > under the "Favourites" tab should instruct the user as to how they can favourite an emote
> 
> 👍 I agree this is needed
> 

I've added some more descriptive text to the Favourite tab, suggestions welcome :)

> Additional: You must consider the next Feature people want is higher tab priorities for favorite emotes.
> 

That should not be difficult to add. But we should probably store the favourites in a data structure that's faster for lookups that a vector.

> If no emote is added yet, two rows of "No favourite emojis, add some" show up as of [c86bd0d](https://github.com/Chatterino/chatterino2/commit/c86bd0d2baad020c9ecf9b452d5267a9ebc88c19)

That should be gone now.

8thony:
> The Name -> Emote assignment logic is channel-aware. It tries to get an Emote for the name in the following order:
> 
> 1. Local Twitch channel Emotes
> 2. Emotes for the currently active Twitch channel - the one we'd paste the name into
> 3. All other Twitch channels available to the user
> 4. Twitch channel BTTV, FFZ, 7TV Emotes, in this order
> 5. Global BTTV, FFZ, 7TV Emotes, in this order
> 
> It returns the first match it finds. This seems like a reasonable "best effort" approach to try to make sure that the Emote the user sends will be the same Emote the other chatters in the channel would see.

Yes, I understand that. But I mean, for example, one channel heavily using this emote:
https://7tv.app/emotes/01F61B1440000991F7SWQNMVX7
But on a completely different channel, it's not appreciated, and they added an emote with the same name:
https://7tv.app/emotes/01FF687BFR000D76QJKE5PX0TX
Both individually would be shown as favorites if the emote popup opened from the chat, which added it. (No Shared Chat Session)



> If you could suggest some channels that would allow me to check this behaviour or test it yourself that'd be great!

Open both emotes I linked above and open from both sites a channel from the list there in Chatterino. 
Mark this shared emote in the first channel as a favorite and open the emote popup on the second channel, which has the different emote. This emote now also shows up as a favorite.

For the example I shared with the same name, go to https://www.frankerfacez.com/emoticon/131597-FeelsWeirdMan open a channel from the list there, and search for the emote. 
You find both the FFZ (Channel) and the 7TV (Global) one.

invisibl3cat:
I checked the behavior of the current code (using the most sophisticated example I could think of). These two channels both have a "67" Emote from 7TV but associate it with a different image. The `Favourite` tab looks like this, depending on which channel is currently active:

67_v1

67_v2

Is this not a sufficient enough indication for the user what they are about to paste? Or am I just not getting the problem you're trying to explain?

We could just switch to EmoteIds instead but as was already pointed out this would force the user to potentially favourite multiple Emotes that look pretty much the same...

8thony:
> We could just switch to EmoteIds instead but as was already pointed out this would force the user to potentially favourite multiple Emotes that look pretty much the same...

Yes, that's the question here: 
Does the user complain about an emote he has not added or that he has to favorite more emotes?

invisibl3cat:
> > We could just switch to EmoteIds instead but as was already pointed out this would force the user to potentially favourite multiple Emotes that look pretty much the same...
> 
> Yes, that's the question here: Does the user complain about an emote he has not added or that he has to favorite more emotes?



> > We could just switch to EmoteIds instead but as was already pointed out this would force the user to potentially favourite multiple Emotes that look pretty much the same...
> 
> Yes, that's the question here: Does the user complain about an emote he has not added or that he has to favorite more emotes?

It would not be that difficult to make this configurable and give the user a choice to store emotes either by names or IDs. @pajlada Would some extra complexity be an okay tradeoff here to resolve this usability issue "diplomatically"?

invisibl3cat:
I thought about this a little more and came up with a bit different approach. It needs some code adjustments in a few places other than just the `EmotePopup` but I don't think that it's that bad.

The current solution is fairly simple but has two awkward usability problems:
- If an Emote is unavailable, we cannot show the image for it, just some text.
- No seamless switching between "match by Emote name" and "match by Emote Id" modes.

To fix this, I tried to store not just the Emote names or Ids but the whole Emote. That's fairly easy because Emotes can be dumped to JSON. This allows me to always reconstruct the favourite Emote regardless of whether it is currently available or not.

The good:
- We can always show the Emote image
- Easy switching between "names" vs. "Id" matching because we know both.

The bad:
- Since we cannot attach arbitrary data to Links, the Emotes need to be passed around in their serialized form as JSONs.
- We also need to store these JSONs in the settings file.

The I'm not sure:
- We need to be able to create Emotes from all providers directly from JSONs. I'm not sure how much of an abuse of the emote handling code that is...

"Strict" matching by ID
avail_unavail
avail_unavail_strict

"Non-strict" mode matching by Names
avail_unavail_non_strict



The code depends on this https://github.com/Chatterino/chatterino2/pull/6985 so it's currently living in my fork here: https://github.com/invisibl3cat/chatterino2/tree/favourite-emotes-with-id-option

There is quite some room for optimization and tweaks but that can be done later...

Nerixyz:
> To fix this, I tried to store not just the Emote names or Ids but the whole Emote. That's fairly easy because Emotes can be dumped to JSON. This allows me to always reconstruct the favourite Emote regardless of whether it is currently available or not.

I don't think the absence of an image is a problem. I'm against saving the raw JSON of an emote - this ties us to the current structure of emotes. We should only include what we need. The name, ID, and possibly the home page should suffice. If an emote is unavailable, the user can click on it to open it in the browser. They can't insert the emote anyway.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -36,12 +37,102 @@
...
+#include <set>
 #include <utility>
 
 namespace {
 
 using namespace chatterino;
 
+constexpr auto makeShortCodesSet(const std::vector<QString> &shortCodes)
+{
+    std::set<QString> shortCodesSet;
```
pajlada: think you need to import util/QStringHash.hpp for things to compile here

src/widgets/dialogs/EmotePopup.hpp
```diff
@@ -39,6 +41,7 @@ class EmotePopup : public BasePopup
     ChannelView *channelEmotesView_{};
     ChannelView *subEmotesView_{};
     ChannelView *viewEmojis_{};
+    ChannelView *favEmotesAndEmojisView_{};
```
pajlada: Rename to `favouritesView_`

src/widgets/dialogs/EmotePopup.hpp
```diff
@@ -51,14 +54,27 @@ class EmotePopup : public BasePopup
     QLineEdit *search_;
     Notebook *notebook_;
 
+    std::vector<EmotePtr> favEmotes_;
+    std::unordered_map<QString, EmojiPtr> favEmojis_;
```
pajlada: use full names: `favouriteEmotes_` and `favouriteEmojis_`

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -664,6 +1030,115 @@ void EmotePopup::filterEmotes(const QString &searchText)
...
+        for (const auto &[setId, emoteSet] : *twitchEmotes)
+        {
+            auto emote = findEmoteByName(name, emoteSet.emotes);
+            if (emote)
+            {
+                return emote;
+            }
+        }
+
+        if (getSettings()->enableBTTVChannelEmotes)
```
pajlada: Twitch channel already has a `bttvEmote`, `ffzEmote`, `seventvEmote` function that handles this setting & checking emotes, we shouldn't need to re-implement it here unless I'm missing something.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -664,6 +1030,115 @@ void EmotePopup::filterEmotes(const QString &searchText)
...
+                findEmoteByName(name, *this->twitchChannel_->seventvEmotes());
+            if (emote)
+            {
+                return emote;
+            }
+        }
+    }
+
+    // Global
+    if (getSettings()->enableBTTVGlobalEmotes)
```
pajlada: Same for global emote fetching - see how MessageBuilder handles it in `parseEmote`

pajlada (changes requested):
Remove the `favouriteStateChanged` function and move the logic for whether it's an "add to favourite" or "remove from favourites" into the `QAction::triggered` lambda directly

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -36,12 +38,104 @@
...
+
+bool emojiHasShortCode(const EmojiPtr &emoji, const QString &shortCode)
+{
+    auto it = std::ranges::find(emoji->shortCodes, shortCode);
+    return it != emoji->shortCodes.end();
+}
+
+bool isEmojiIdentifier(const QString &emoteName)
+{
+    /*
```
Nerixyz: Use `//` (BCPL) comments

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -36,12 +38,104 @@
...
+     * Some Emotes can have a name that looks like this:
+     *
+     *  :<name>:
+     *
+     * This makes them look like Emoji short codes. To better tell Emotes and
+     * Emojis apart, we check the emoteName against a list of known Emoji
+     * short codes and we have a match, we trust that the idenfifier is
+     * an Emoji.
```
Nerixyz: What if an emote is named after an emoji? Chatterino currently displays the emoji there.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -36,12 +38,104 @@
...
+    return emoteName.mid(1, emoteName.length() - 2);
+}
+
+bool isFavouriteEmoteOrEmoji(const MessageLayoutElement *element)
+{
+    const auto &emoteName = element->getCreator().getLink().value;
+
+    if (isEmojiIdentifier(emoteName))
+    {
+        const auto &emojiNames = getSettings()->favouriteEmojis.getValue();
```
Nerixyz: Pretty sure this binds to a temporary, because `getValue` returns a `T`, not a `const T&`.

src/singletons/Settings.hpp
```diff
@@ -441,6 +441,9 @@ class Settings
     BoolSetting enableSevenTVEventAPI = {"/emotes/seventv/eventapi", true};
     BoolSetting sendSevenTVActivity = {"/emotes/seventv/sendActivity", true};
 
+    ChatterinoSetting<std::vector<QString>> favouriteEmotes = {"/emotes/favouriteEmotes", {}};
+    ChatterinoSetting<std::vector<QString>> favouriteEmojis = {"/emotes/favouriteEmojis", {}};
```
Nerixyz: You might want to make these QStringLists instead. A std::vector<QString> needs to copy all its data in the copy constructor - a QList is COW. It also has convenience methods like `contains` or `removeAll`.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -298,7 +395,30 @@ EmotePopup::EmotePopup(QWidget *parent)
     QObject::connect(this->search_, &QLineEdit::textChanged, this,
                      &EmotePopup::filterEmotes);
 
-    auto clicked = [this](const Link &link) {
+    auto clicked = [this](const Link &link, Qt::KeyboardModifiers modifiers) {
+        if (modifiers & Qt::KeyboardModifier::ControlModifier)
```
```suggestion
        if (modifiers.testFlag(Qt::KeyboardModifier::ControlModifier))
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -468,10 +631,203 @@ void EmotePopup::loadChannel(ChannelPtr channel)
...
+        }
+    }
+
+    this->updateFavouriteEmotesAndEmojis();
+    saveFavouriteEmojis(this->favouriteEmojis_);
+}
+
+void EmotePopup::addFavouriteEmote(const EmoteName &name)
+{
+    /*
```
Nerixyz: Use `//`

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -468,10 +631,203 @@ void EmotePopup::loadChannel(ChannelPtr channel)
...
+                return emoteName == emote->name.string;
+            });
+        if (it == this->favouriteEmotes_.end())
+        {
+            unavailableEmotes.push_back(emoteName);
+        }
+    }
+    if (!unavailableEmotes.empty())
+    {
+        static const QString explainUnavailability =
```
Nerixyz: You can use string literals here (`using namespace Qt::Literals; u"str"_s`;

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -664,6 +1046,91 @@ void EmotePopup::filterEmotes(const QString &searchText)
...
+
+        auto emote = findEmoteByName(name, *emotesToTry);
+        if (emote)
+        {
+            return emote;
+        }
+
+        auto twitchEmotes =
+            *getApp()->getAccounts()->twitch.getCurrent()->accessEmoteSets();
+        /*
```
Nerixyz: Use `//`.

invisibl3cat:
I've addressed the suggestions made on the stream and in the PR so that we can try to get the initial idea merged and get some user feedback.

One silly issue I know of is that the Favourite checkbox appears everywhere. I tried to fix this but there is no good way how to tell which elements represent something that can be (un)favourited and which do not.

I think that the explanatory tooltip for unavailable emotes is a good thing to have. It could, perhaps, look a bit more streamlined.

Thanks for all the time you spend reviewing my evening musings over Chatterino code 👍

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -468,10 +632,203 @@ void EmotePopup::loadChannel(ChannelPtr channel)
...
+
+        MessageBuilder builder;
+        builder->flags.set(MessageFlag::Centered);
+        builder
+            .emplace<TextElement>(
+                "Currently unavailable favourite emotes",
+                MessageElementFlags{MessageElementFlag::Text,
+                                    MessageElementFlag::AlwaysShow},
+                MessageColor::System)
+            ->setTooltip(explainUnavailability);
```
pajlada: I'd prefer not using a tooltip - it looks kinda goofy for me. I think adding extra text is OK

invisibl3cat:
Updated as requested. Note that I had to add a new `elementClicked` signal to `ChannelView` to get what element was clicked on a left-click event too. What was discussed on stream worked only on a right-click because that was the handler for the context menu.

Checking the flags should also fix the issue of "Favourite" checkbox showing up for nonsensical items.

invisibl3cat:
Rebased against master

Nerixyz:
Merging with master/main is usually preferred here, because the commits are squashed at the end and rebasing usually discards the "mark file as viewed" state for reviewers.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -664,6 +1030,91 @@ void EmotePopup::filterEmotes(const QString &searchText)
...
+        if (emote)
+        {
+            return emote;
+        }
+
+        emote = this->twitchChannel_->seventvEmote(name);
+        if (emote)
+        {
+            return emote;
+        }
```
pajlada, line 1094: Order does not match our standard emote order, which is:
1. FFZ Channel
2. BetterTTV Channel
3. 7TV Channel
4. FFZ Global
5. BetterTTV global
6. 7TV Global

src/widgets/dialogs/EmotePopup.cpp
pajlada: @invisibl3cat you can do this too, sorry i marked as resolved too early

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -298,8 +376,53 @@ EmotePopup::EmotePopup(QWidget *parent)
...
+            else if (this->viewEmojis_ == page)
+            {
+                this->addFavouriteEmoji(toEmojiShortCode(identifier));
+            }
+            else
+            {
+                this->addFavouriteEmote(EmoteName{identifier});
+            }
+
+            if (!(modifiers & Qt::KeyboardModifier::ShiftModifier))
```
pajlada: testFlag here too

pajlada (commented):
Just a few unresolved comments, otherwise this looks good to go.

pajlada (approved):

pajlada:
Going to merge this in as-is to get a bit of a user-base to test it out. I'm open to changing functionality before we make the next release

PR #6976, opened 2026-05-03
Provide better feedback to the user when they toggle logging to file

### LINKED ISSUE #6974: Logging command enhancements
> Technically you already satisfied this issue, but I found 2 little annoyances in your implementation.
> - If I use `/debug-enable-logfile` or `/debug-disable-logfile` i dont get feedback that logging started or stopped.
> - If I use `/debug-enable-logfile` with a path surrounded by quotation marks, I get an error `Unable to open log file`. 

 _Originally posted by @8thony in [#6901](https://github.com/Chatterino/chatterino2/issues/6901#issuecomment-4366352500)_

invisibl3cat:
This addresses some user convenience issues brought up by @8thony wrt  #6901

The error message displayed when logging to file fails to enable now specifically mentions that the path to the file should not be quoted. The error message is also displayed in red.

Fixes #6974

### DIFF
--- a/src/controllers/commands/builtin/chatterino/Debugging.cpp
+++ b/src/controllers/commands/builtin/chatterino/Debugging.cpp
@@ -321,6 +321,8 @@ QString disableLogfile(const CommandContext &ctx)
 {
     FileLogger::instance().disable();
 
+    ctx.channel->addSystemMessage("Logging to file disabled");
+
     return {};
 }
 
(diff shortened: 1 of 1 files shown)


pajlada (approved):

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -341,14 +343,24 @@ QString enableLogfile(const CommandContext &ctx)
...
         auto error = result.error();
 
-        ctx.channel->addSystemMessage(
+        MessageBuilder builder;
+        builder.emplace<TextElement>(
             QString("Unable to open log file '%1'. Error reported by "
-                    "the system was: %2")
-                .arg(error.absFilePath, error.errorDesc));
+                    "the system was: %2 (Hint: Do not quote the file path "
+                    "even if it contains spaces)")
```
Nerixyz, line 358: We can detect and remove the quotes. If we fail to open the file, we can append some diagnostic about the quotes.

PR #6982, opened 2026-05-04
If path to logfile is quoted, try to intelligently dequote it

invisibl3cat:
Feel free to just drop  this if you think the added complexity is not worth the benefit.

@Nerixyz suggested in https://github.com/Chatterino/chatterino2/issues/6974 that instead of telling users to not quote path to log files we could just handle the quotes. Doing this opens a bit of a can of worms which is why I initially didn't do it.

- `/debug-enable-logfile "something" something` is now an invalid input because of the trailing characters
- Some common filesystems allow to have the double quote as part of path so we need to have escaping to allow double quotes inside a quoted string
- The usual escape character -  `\` - is also a path separator in Windows

The current code should handle most inputs a user may consider to be the "correct" way  to enter a path.

If you decide this is worth merging I'd like to have a unit test for this. Any suggestions how to add it without having to expose the `dequoteAndUnescapeFilePath` function?


src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -59,6 +59,70 @@ bool restartChatterino(const QProcessEnvironment &env)
     return false;
 }
 
+QString dequoteAndUnescapeFilePath(QString filePath, bool &ok)
```
Nerixyz: Can return a `std::optional` or `ExpectedStr` here.

src/controllers/commands/builtin/chatterino/Debugging.cpp
```diff
@@ -59,6 +59,70 @@ bool restartChatterino(const QProcessEnvironment &env)
...
+{
+    ok = false;
+
+    filePath = filePath.trimmed();
+    if (filePath.startsWith("\""))
+    {
```
Nerixyz, line 66: This feels too complex to me. I'd be fine if we just strip leading and trailing quotes. The chances of someone invoking this as `/command "my path.txt" epic argument` are near zero. After all, it's a debug command. And if someone wants quotes in their filename, they can double quote the path (`""myfile""` → `"myfile"`).

invisibl3cat: If we go ahead with this, shouldn't we try to make the quoted strings behave close to like quoted strings in common shells do? I imagine this is what most users would expect. Using multiple quotes is not something I'd think of...

Nerixyz: Hmm yea, it's not that bad. I was only concerned that the function is relatively large.

pajlada:
Remove first & last quotes if first & last quotes exist - nothing more. I don't think it's worth the complexity to handle escaping/unescaping things

invisibl3cat:
When @8thony brought this up he didn't mention that the quotes probably came from Windows automatically enquoting a copy-pasted path. I then assumed a user that'd expect to work with strings like they would in a terminal shell. If Windows puts quotes around a copy-pasted path, the easiest thing is to just remove them.

pajlada (approved):

PR #6985, opened 2026-05-06
Store EmoteId for all Emotes

invisibl3cat:
Unless I'm missing something, this is probably just an oversight?

I noticed this while playing with favourite Emotes stored by Emote IDs.

### DIFF
--- a/src/providers/bttv/BttvEmotes.cpp
+++ b/src/providers/bttv/BttvEmotes.cpp
@@ -93,6 +93,7 @@ std::pair<Outcome, EmoteMap> parseGlobalEmotes(const QJsonArray &jsonEmotes,
             .tooltip = Tooltip{name.string + "<br>Global BetterTTV Emote"},
             .homePage = Url{EMOTE_LINK_FORMAT.arg(id.string)},
             .zeroWidth = ZERO_WIDTH_EMOTES.contains(name.string),
+            .id = id,
         });
 
         emotes[name] = cachedOrMakeEmotePtr(std::move(emote), currentEmotes);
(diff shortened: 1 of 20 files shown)


4rneee:
I think the same goes for global bttv emotes:
https://github.com/Chatterino/chatterino2/blob/6ddd9af2a313e07848280bb0ed7909e43db52a69/src/providers/bttv/BttvEmotes.cpp#L82-L96

and all ffz emotes:
https://github.com/Chatterino/chatterino2/blob/6ddd9af2a313e07848280bb0ed7909e43db52a69/src/providers/ffz/FfzEmotes.cpp#L92-L100

invisibl3cat:
> I think the same goes for global bttv emotes:
> 
> https://github.com/Chatterino/chatterino2/blob/6ddd9af2a313e07848280bb0ed7909e43db52a69/src/providers/bttv/BttvEmotes.cpp#L82-L96
> 
> and all ffz emotes:
> 
> https://github.com/Chatterino/chatterino2/blob/6ddd9af2a313e07848280bb0ed7909e43db52a69/src/providers/ffz/FfzEmotes.cpp#L92-L100

Is this intentional or is it just an oversight due to the fact that we're not really accessing the `.id` field in the Emote anywhere?

Nerixyz:
> Is this intentional or is it just an oversight due to the fact that we're not really accessing the `.id` field in the Emote anywhere?

The `id` field was added for 7TV emotes and the event API. We didn't need it elsewhere.

invisibl3cat:
> > Is this intentional or is it just an oversight due to the fact that we're not really accessing the `.id` field in the Emote anywhere?
> 
> The `id` field was added for 7TV emotes and the event API. We didn't need it elsewhere.

I see. Is it okay to set it for all emotes so that we can identify emotes by their Id and not just a name? At the very least it'd help with the Favourite emotes implementation.

Nerixyz:
> I see. Is it okay to set it for all emotes so that we can identify emotes by their Id and not just a name? At the very least it'd help with the Favourite emotes implementation.

Oh, yea that's fine.

PR #6990, opened 2026-05-10
feat(plugins): Allow messages to be cloned

Nerixyz:
After #6909, we can clone message elements. With this PR, we can clone whole messages. The messages are non-frozen afterward, so plugins can edit their properties and add/remove elements.


pajlada (approved):
This works as expected for me 👍

PR #6997, opened 2026-05-13
deps: remove direct dependency on Boost.Variant

Nerixyz:
We compile with C++ 23 now and have compiled with 20 and 17 for a while. These standards support `std::variant` which we use in several places. A few older places still used `boost::variant`. `std::variant` supersedes `boost::variant`. It doesn't require an extra dependency and (usually) has built in support in debuggers.

I also removed includes for `boost/current_function` and `boost/foreach`. We didn't use them. We still depend on Boost.Variant through Boost.Signals2, though.

### DIFF
--- a/src/PrecompiledHeader.hpp
+++ b/src/PrecompiledHeader.hpp
@@ -4,8 +4,6 @@
 
 #ifdef __cplusplus
 #    include <boost/circular_buffer.hpp>
-#    include <boost/current_function.hpp>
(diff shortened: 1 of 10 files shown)


pajlada:
`C:\dev\chatterino2\src\widgets\Notebook.cpp(28): fatal error C1083: Cannot open include file: 'boost/foreach.hpp': No such file or directory`

pajlada (approved):

PR #6999, opened 2026-05-15
overlay: fix scroll wheel & reset hotkey zoom

4rneee:
While playing around with the overlay I noticed that zooming in/out with keybindings only changes the overlay zoom while zooming in/out with the mouse changed all windows.  
This PR fixes this so that zooming with the mouse only changes the overlay scale.

I also noticed that the zoom reset hotkey resets the overall UI scale and not the overlay zoom when the overlay is focused. ~Is this intended? If not, I could fix this here too.~ This is now also fixed.


Nerixyz:
Thank you for noticing and fixing this!

> I also noticed that the zoom reset hotkey resets the overall UI scale and not the overlay zoom when the overlay is focused. Is this intended? If not, I could fix this here too.

I think it should reset just the overlay zoom, since the other hotkeys also affect _just_ the overlay. It might be a bit weird, because the overlay scale is actually `globalScale * overlayScaleFactor`. But having only the scale factor reset seems better, because you can't reset it otherwise.

src/widgets/OverlayWindow.cpp
```diff
@@ -225,6 +225,29 @@ void OverlayWindow::applyTheme()
...
 
+void OverlayWindow::wheelEvent(QWheelEvent *event)
+{
+    // ignore horizontal mouse wheels
+    if (event->angleDelta().x() != 0)
+    {
+        return;
+    }
+
+    if ((event->modifiers() & Qt::ControlModifier) != 0U)
```
Nerixyz:
```suggestion
    if (event->modifiers().testFlag(Qt::ControlModifier))
```

Nerixyz (approved):

pajlada (approved):

PR #7002, opened 2026-05-16
fix: load newly added filters in FilterSet

BRADYguang:
This PR fixes a functional bug in Chatterino2's filter system.

Problem:
When users add new filters in the settings, the new filter rules do not take effect at runtime. Users have to restart the application to make new filters work.

Root Cause:
The original reloadFilters() function only updated or removed existing filters, but did not load newly created filters.

Fix:
Added logic to automatically load and apply newly added filters during the reload process.

This is a non-trivial functional change that directly modifies the runtime behavior of the application.


pajlada:
Hi, could you please provide clear steps for reproducing the bug? A video would work too.

PR #7004, opened 2026-05-17
feat: Add `/(un)pin`

Nerixyz:
This adds the `/pin` and `/unpin` commands as well as action in the context menu of messages.

The `/pin` command can do the following:
- `/pin`: Show the currently pinned message. Intended to be temporary until we have the PubSub/EventSub topic for this.
- `/pin some message here`: Send a message and pin it for 20 minutes.
- `/pin -d none some message here` (or `/pin -d until-end msg`): Send a message and pin it until the stream ends.
- `/pin --id abcd123 -d 15m`: Pin a message with ID `abcd123` for 15 min.
- `/pin -h` or `/pin --help`: Show usage.

`/unpin` either takes no arguments or the ID to unpin. When no arguments are specified, it unpins the currently pinned message.

Having three layers of submenus in the channel view seems a bit weird. But on the other hand, having multiple `Pin for x minutes` is also weird.

### DIFF
--- a/src/controllers/commands/builtin/twitch/Pin.cpp
+++ b/src/controllers/commands/builtin/twitch/Pin.cpp
@@ -1,0 +1,394 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#include "Application.hpp"
+#include "common/Channel.hpp"
+#include "controllers/accounts/AccountController.hpp"
+#include "controllers/commands/CommandContext.hpp"
+#include "messages/Message.hpp"
+#include "messages/MessageBuilder.hpp"
+#include "messages/MessageElement.hpp"
+#include "providers/twitch/api/Helix.hpp"
+#include "providers/twitch/TwitchAccount.hpp"
+#include "providers/twitch/TwitchChannel.hpp"
+#include "util/Expected.hpp"
+#include "util/FormatTime.hpp"
+#include "util/Helpers.hpp"
+
+#include <controllers/commands/builtin/twitch/Pin.hpp>
+#include <QCommandLineParser>
+#include <QProcess>
+
+namespace {
+
+using namespace Qt::Literals;
+using namespace chatterino;
+
+struct SendPinnedAction {
+    QString text;
+
+    void run(const std::shared_ptr<TwitchChannel> &chan,
+             const TwitchAccount &moderator) const;
+};
+
+struct PinAction {
+    QString messageID;
+    std::optional<std::chrono::seconds> duration;
+
+    void run(const std::shared_ptr<TwitchChannel> &chan,
+             const TwitchAccount &moderator) const;
+};
+
+using Action = std::variant<SendPinnedAction, PinAction>;
+
+void SendPinnedAction::run(const std::shared_ptr<TwitchChannel> &chan,
+                           const TwitchAccount &moderator) const
+{
+    getHelix()->sendChatMessage(
+        {
+            .broadcasterID = chan->roomId(),
+            .senderID = moderator.getUserId(),
+            .message = this->text,
+            .pin = true,
+        },
+        [weak = std::weak_ptr(chan),
+         origText = this->text](const HelixSentMessage &res) {
+            auto chan = weak.lock();
+            if (!chan)
+            {
+                return;
+            }
+
+            if (res.isSent)
+            {
+                chan->addMessage(
+                    MessageBuilder::makePinSuccessMessage(origText, res.id),
+                    MessageContext::Original);
+                return;
+            }
+
+            if (res.dropReason)
+            {
+                chan->addSystemMessage(res.dropReason->message);
+            }
+            else
+            {
+                chan->addSystemMessage("Your message was not pinned.");
+            }
+        },
+        [weak = std::weak_ptr(chan)](auto error, auto message) {
+            auto chan = weak.lock();
+            if (!chan)
+            {
+                return;
+            }
+
+            if (message.isEmpty())
+            {
+                message = "(empty message)";
+            }
+
+            using Error = HelixSendMessageError;
+
+            auto errorMessage = [&]() -> QString {
+                switch (error)
+                {
+                    case Error::MissingText:
+                        return "You can't pin an empty message.";
+                    case Error::BadRequest:
+                        return "Failed to pin message: " + message;
+                    case Error::Forbidden:
+                        return "You are not allowed to pin messages in this "
+                               "channel.";
+                    case Error::MessageTooLarge:
+                        return "Your message was too long.";
+                    case Error::UserMissingScope:
+                        return "Missing required scope. Re-login with your "
+                               "account and try again.";
+                    case Error::Forwarded:
+                        return message;
+                    case Error::Unknown:
+                    default:
+                        return "Unknown error: " + message;
+                }
+            }();
+            chan->addSystemMessage(errorMessage);
+        });
+}
+
+void PinAction::run(const std::shared_ptr<TwitchChannel> &chan,
+                    const TwitchAccount &moderator) const
+{
+    chan->pinMessageAs(this->messageID, this->duration, moderator);
+}
+
+ExpectedStr<Action> parseAction(const CommandContext &ctx)
+{
+    QCommandLineParser parser;
+    parser.setSingleDashWordOptionMode(QCommandLineParser::ParseAsLongOptions);
+    parser.setOptionsAfterPositionalArgumentsMode(
+        QCommandLineParser::ParseAsOptions);
+    QCommandLineOption idOption(u"id"_s, {}, u"ID"_s);
+    QCommandLineOption durationOption(QStringList{u"d"_s, u"duration"_s}, {},
+                                      u"duration"_s);
+    parser.addOptions({idOption, durationOption});
+    parser.parse(QProcess::splitCommand(ctx.words.join(" ")));
+
+    std::optional<std::chrono::seconds> duration;
+    if (parser.isSet(durationOption))
+    {
+        const auto dur = parseDurationToSeconds(parser.value(durationOption));
+        if (dur <= 0)
+        {
+            return makeUnexpected(u"Duration must be positive."_s);
+        }
+        duration.emplace(dur);
+    }
+
+    std::optional<QString> id;
+    if (parser.isSet(idOption))
+    {
+        id.emplace(parser.value(idOption));
+    }
+
+    auto positionals = parser.positionalArguments();
+
+    if (id)
+    {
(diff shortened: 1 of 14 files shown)


src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    auto action = parseAction(ctx);
+    if (!action)
+    {
+        ctx.channel->addSystemMessage(action.error());
+        return {};
+    }
+
+    std::visit(
```
Mm2PL: This feels slightly overcomplicated. Unless there's more complex parsing logic needed I'd prefer a simpler approach without the structs and variant, like:
```cpp
void runCommand(QString args)
{
    /* parse the text... */
    if (id)
    {
        chan->pinMessageAs(*id, duraction, current_account);
    }
    else if (text)
    {
        getHelix()->sendChatMessage(/* ... */);
    }
}
```

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,394 @@
...
+            .messageID = *std::move(id),
+            .duration = duration,
+        };
+    }
+
+    if (duration)
+    {
+        return makeUnexpected(
+            u"No duration can be specified when sending and pinning a message. "
+            "Send the message first, and pin it afterward."_s);
```
Mm2PL: We could work around this limitation by sending a message and attempting to pin it instantly (helix gives the id to us). What do you think?

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,394 @@
...
+    {
+        assert(false && "should have /pin");
+        return {};
+    }
+    args = args.subspan(1);
+
+    if (args.size() == 1 && (args[0] == u"-h" || args[0] == u"--help"))
+    {
+        ctx.channel->addSystemMessage(
+            u"Pin a chat message or show the currently pinned one. Usage: /pin --duration <seconds> --id <id> [message]..."_s);
```
Mm2PL: Nit: I think this usage should be split in two:
`Usage: /pin [--duration <duration>] --id <messageID> or /pin [message]...`. Because `/pin --id a9c01967-014c-41f7-be7c-cc4d0489314f My new pinned message` does not logically make sense.

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,394 @@
...
+            {
+                pinMessageText = pinMessageText.left(50) + "…";
+            }
+
+            builder
+                .emplace<TextElement>(pinMessageText, MessageElementFlag::Text,
+                                      MessageColor::Text)
+                ->setLink({Link::JumpToMessage, result->messageID});
+            builder->messageText = pinMessageText;
+            builder->searchText = pinMessageText;
```
Mm2PL: Could this live in MessageBuilder?

Mm2PL (changes requested):

src/controllers/commands/builtin/twitch/Pin.cpp
Nerixyz: If the user has permissions, that will work just fine. However, if the user isn't a mod and uses the command, it will send a message and then fail to pin that message. We could check if the user has mod status and only send the message in that case, though.

src/providers/twitch/api/Helix.cpp
```diff
@@ -3806,6 +3810,164 @@ void Helix::deleteEventSubSubscription(const QString &subscriptionID,
         .execute();
 }
 
+void Helix::pinChatMessage(
+    const QString &broadcasterID, const QString &moderatorID,
+    const QString &messageID, std::optional<std::chrono::seconds> duration,
+    ResultCallback<> successCallback,
```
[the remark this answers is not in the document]
Nerixyz: We always pass this by value in the other overloads. If we want to, we can change that in bulk.

src/controllers/commands/builtin/twitch/Pin.cpp
Mm2PL: We can use an api call checking if there is a pinned message. That will also give us a way to not send the message if there is something pinned already.

src/controllers/commands/builtin/twitch/Pin.cpp
Nerixyz: Yes, moved it there.

src/controllers/commands/builtin/twitch/Pin.cpp
Nerixyz: Done.

src/controllers/commands/builtin/twitch/Pin.cpp
Nerixyz: I'm now checking `hasModRights` here, because getting the pinned message seems to be available to everyone (actually a good thing).

iProdigy: The Send Chat Message API endpoint has been updated with a new optional pin parameter to support sending and pinning a message in a single request.

Nerixyz: We can't control the duration of the pinned message with the "Send Chat Message" endpoint.

Now that I think about it, we could also use the update endpoint to change the duration.

Nerixyz: Did that in [`d47f38b` (this PR)](https://github.com/Chatterino/chatterino2/pull/7004/commits/d47f38beec7a24b3e1c39551df0e87d6809af06d).

src/providers/twitch/api/Helix.cpp
```diff
@@ -3806,6 +3810,226 @@ void Helix::deleteEventSubSubscription(const QString &subscriptionID,
...
```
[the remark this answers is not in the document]
Nerixyz: See above suggestion.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2178,6 +2188,143 @@ void TwitchChannel::deleteMessagesAs(const QString &messageID,
...
+void TwitchChannel::pinOrUpdateMessage(
+    bool update, const QString &messageID,
+    std::optional<std::chrono::seconds> duration,
+    const TwitchAccount &moderator, QString textHint)
+{
+    auto fn = &IHelix::pinChatMessage;
+    if (update)
+    {
+        fn = &IHelix::updatePinnedChatMessage;
+    }
```
pajlada: I would prefer to see the lambda saved & we have two paths for update/no update
e.g.
```c++
auto onSuccess = ...;
auto onFailure = ...;

if (update) {
  getHelix()->updatePinnedChatMessage(a,b,c,d,onSuccess,onFailure);
} else {
  getHelix()->pinChatMessage(a,b,c,d,onSuccess,onFailure);
}
```

src/messages/MessageBuilder.hpp
```diff
@@ -257,6 +258,12 @@ class MessageBuilder
                                               const QString &actor,
                                               uint32_t count = 1);
 
+    static MessagePtrMut makePinSuccessMessage(const QString &textOrID,
+                                               const QString &id);
```
pajlada: Why do we accept the ID potentially twice?

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        ctx.channel->addSystemMessage("You must be logged in to pin messages!");
+        return "";
+    }
+
+    if (ctx.words.empty())
+    {
+        assert(false && "should have /pin");
+        return {};
+    }
```
Mm2PL, line 243: Does msvc hate the simple assert?

```suggestion
    assert(!ctx.words.empty() && "should have /pin")
```

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        ctx.channel->addSystemMessage("You must be logged in to pin messages!");
+        return "";
+    }
+
+    if (ctx.words.empty())
+    {
+        assert(false);
+        return {};
+    }
```
Mm2PL, line 306: Does msvc hate the simple assert?

```suggestion
    assert(!ctx.words.empty() && "should have /unpin")
```

src/controllers/commands/builtin/twitch/Pin.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        ctx.channel->addSystemMessage(
+            u"The /unpin command only works in Twitch channels"_s);
+        return {};
+    }
+
+    auto currentUser = getApp()->getAccounts()->twitch.getCurrent();
+    if (currentUser->isAnon())
+    {
+        ctx.channel->addSystemMessage("You must be logged in to pin messages!");
```
Mm2PL: I think this error message should differentiate /pin and /unpin
```suggestion
        ctx.channel->addSystemMessage(
            u"You must be logged in to unpin messages!"_s);
```

src/controllers/commands/builtin/twitch/Pin.cpp
Nerixyz: No, it should be fine. I was mainly worried about UB of indexing into the list in release mode, hence the return.

src/messages/MessageBuilder.hpp
Nerixyz: Artifact from before, should be fixed now.

pajlada (approved):

src/providers/twitch/TwitchAccountManager.cpp
```diff
@@ -165,7 +165,8 @@ const std::vector<QStringView> AUTH_SCOPES{
     u"moderator:manage:banned_users",  // for ban/unban/timeout/untimeout api & channel.moderate eventsub topic
 
     // https://dev.twitch.tv/docs/api/reference#delete-chat-messages
-    u"moderator:manage:chat_messages",  // for delete message api (/delete, /clear) & channel.moderate eventsub topic
+    // https://dev.twitch.tv/docs/api/reference#pin-chat-message
+    u"moderator:manage:chat_messages",  // for delete message api (/delete, /clear, /pin) & channel.moderate eventsub topic
```
Wissididom: Should it stay "for delete message api"? I mean `/pin` doesn't really have to do with deletion I think.

Mm2PL:
```suggestion
    u"moderator:manage:chat_messages",  // for message moderation api (/delete, /clear, /pin) & channel.moderate eventsub topic
```

Mm2PL (approved):
LGTM

PR #7012, opened 2026-05-24
fix: `reply-parent-display-name` escapes in `stripLeadingReplyMention`

### LINKED ISSUE #7011: Replies to users with space-trailing display names don't strip out username
### Checklist

### Describe your issue

When replying to a user whose display name ends in a space (which is a Twitch data issue,) the username isn't stripped from the start of the reply.

## Source message

~~https://paste.ivr.fi/mukilijupa.sql~~ (had escaped backslashes from recent-messages)
https://paste.ivr.fi/emolosacym.sql

## Rendered

Image

### OS and Chatterino Version

Chatterino 2.5.5 (commit [ed7fe100c](https://github.com/Chatterino/chatterino2/commit/ed7fe100c)) built with Qt 6.11.0 (running on 6.11.1)
Running on Arch Linux, kernel: 7.0.9-1-cachyos

brian6932:
fixes #7011

_Sorry for the spam, I was curious if I could remove the branch, but it seems like it's required for `[blocked user]` tests._

This is a monkey patch fix for now, I didn't want to touch the parser in such a PR in case of breakage, but ideally MessageBuilder.cpp should do unescaping of all tags first.

The following tags should be unescaped:
- [x] `badge-info`
- [x] `source-badge-info`
- [ ] `client-nonce`
- [ ] `msg-param-sub-plan-name`
- [x] `system-msg`
- [x] `reply-parent-msg-body`
- [ ] `reply-parent-display-name`
- [ ] `msg-param-origin-id`
- [ ] `msg-param-displayName`
- [x] `display-name`

### DIFF
--- a/src/providers/twitch/IrcMessageHandler.cpp
+++ b/src/providers/twitch/IrcMessageHandler.cpp
@@ -106,7 +106,7 @@ int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
     if (const auto it = tags.find("reply-parent-display-name");
         it != tags.end())
     {
-        auto displayName = it.value().toString();
+        auto displayName = parseTagString(it.value().toString());
 
         if (content.length() <= 1 + displayName.length())
         {
@@ -118,7 +118,8 @@ int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
             content.at(1 + displayName.length()) == ' ' &&
             content.indexOf(displayName, 1) == 1)
         {
-            int messageOffset = 1 + displayName.length() + 1;
+            // Reply prefix's "@" + displayName + " "
+            qsizetype messageOffset = 1 + displayName.length() + 1;
             content.remove(0, messageOffset);
             return messageOffset;
         }
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -333,12 +333,8 @@ void appendBadges(MessageBuilder *builder,
             {
                 auto infoValue = badgeInfoIt->second;
                 auto predictionText =
-                    infoValue
-                        .replace(R"(\s)", " ")  // standard IRC escapes
-                        .replace(R"(\:)", ";")
-                        .replace(R"(\\)", R"(\)")
-                        .replace("⸝", ",");  // twitch's comma escape
-                // Careful, the first character is RIGHT LOW PARAPHRASE BRACKET or U+2E1D, which just looks like a comma
+                    parseTagString(infoValue).replace("⸝", ",");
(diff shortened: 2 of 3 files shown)


jupjohn:
LGTM 

image

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -118,9 +123,10 @@ int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
             content.at(1 + displayName.length()) == ' ' &&
             content.indexOf(displayName, 1) == 1)
         {
-            int messageOffset = 1 + displayName.length() + 1;
+            // Reply prefix's "@" + displayName + " "
+            qsizetype messageOffset = 1 + displayName.length() + 1;
             content.remove(0, messageOffset);
-            return messageOffset;
+            return (int)messageOffset;
```
Nerixyz: Return `qsizetype` instead.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -106,7 +106,12 @@ int stripLeadingReplyMention(const QVariantMap &tags, QString &content)
     if (const auto it = tags.find("reply-parent-display-name");
         it != tags.end())
     {
-        auto displayName = it.value().toString();
+        auto displayName = it.value()
+                               .toString()
+                               .replace(R"(\s)", " ")  // standard IRC escapes
+                               .replace(R"(\:)", ";")
+                               .replace(R"(\\)", R"(\)")
+                               .replace("⸝", ",");
```
Nerixyz: Use QString literals here (`u""_s`) to avoid creating intermediate strings.

Also, it looks like `\\s` will be incorrectly unescaped to `\ ` instead of `\s`. This could be a standalone function. That way, we could test it better and/or write the replacements manually.

Nerixyz (commented):
A test for this in https://github.com/Chatterino/chatterino2/tree/11aca323a1ce89d887ce8bcef8bf53617429bdc3/tests/snapshots/IrcMessageHandler would be great.

src/providers/twitch/IrcMessageHandler.cpp
pajlada: In IrcHelpers.hpp we have `parseTagString` that should be usable instead. Once you've added a unit test, you can swap to that function and ensure it works as expected.

brian6932: fwiw I took that from: https://github.com/Chatterino/chatterino2/blob/47e4abf5470c1b08b3d3ff1a2289b4e865990b31/src/messages/MessageBuilder.cpp#L337-L340
So I guess I'll replace it there with `parseTagString` as well.

Nerixyz (approved):

pajlada (approved):

PR #7014, opened 2026-05-24
feat: add pinned message banner

realtlos:
Adds a pinned message banner to each split. When a moderator pins a message, a small panel appears between the channel header and the chat view showing who pinned it, the message text, the original sender, and a countdown if the pin has a duration.

The banner auto-hides after a few seconds. It can be kept open by toggling the pin button that appears in the split header whenever a message is pinned, or by enabling the new "Always show pinned channel message" setting.

Mods get a ⋮ menu on the banner to:
- Unpin the message
- Re-pin it with a specific duration (1/5/10/20/30 minutes or until stream ends)
- Hide the banner locally without unpinning

The current pin is fetched via `GET /helix/chat/pins` on channel join and when the user gains mod rights. `pinned-chat-updates-v1` PubSub events keep it in sync: pin/update events re-fetch from Helix, unpin clears it locally. The context menu pin actions from #7004 also trigger a refresh.

(screenshots below) [updated]

image
image


mocks/include/mocks/Helix.hpp
Nerixyz, line 1: These changes are not necessary.

src/providers/twitch/pubsubmessages/PinnedChatUpdates.cpp
```diff
@@ -0,0 +1,23 @@
+// SPDX-FileCopyrightText: 2025 Contributors to Chatterino <https://chatterino.com>
```
Nerixyz:
```suggestion
// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
```

Same in all the other new files.

src/providers/twitch/PubSubClient.cpp
```diff
@@ -162,6 +162,46 @@ void PubSubClient::handleResponse(const PubSubMessage &message)
...
+        {
+            qCDebug(chatterinoPubSub)
+                << "Malformed pinned-chat-updates-v1 message";
+            return;
+        }
+
+        const auto &innerMessage = *oInnerMessage;
+        // strip the "pinned-chat-updates-v1." prefix
+        const auto channelId =
+            message.topic.mid(static_cast<int>(sizeof("pinned-chat-updates-v1.") - 1));
```
Nerixyz:
```suggestion
            message.topic.sliced(static_cast<qsizetype>(sizeof("pinned-chat-updates-v1.") - 1));
```

> If you know that _position_ and _n_ cannot be out of bounds, use [sliced](https://doc.qt.io/qt-6/qstring.html#sliced)() instead in new code, because it is faster.

src/providers/twitch/PubSubManager.hpp
```diff
@@ -49,6 +49,13 @@ class PubSub
         Signal<const QJsonObject &> redeemed;
     } pointReward;
 
+    struct {
+        /// Emitted when a message is pinned or its pin is updated.
+        Signal<const QString &> pinned;
```
Nerixyz, line 55: Mention that the signal argument here is the channel name.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -926,6 +927,12 @@ void TwitchChannel::setMod(bool value)
         this->mod_ = value;
 
         this->userStateChanged.invoke();
+
+        if (value)
+        {
+            // Gained mod privileges — fetch the current pin
```
Nerixyz:
```suggestion
            // Gained mod privileges - fetch the current pin
```

Same in some other comments.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,129 @@ bool TwitchChannel::isLoadingRecentMessages() const
     return this->loadingRecentMessages_.test();
 }
 
+void TwitchChannel::refreshPinnedMessage()
+{
+    if (!this->hasModRights())
+    {
+        return;
+    }
```
Nerixyz: At least when I tested it, you could get the currently pinned message without mod rights (which is a sane thing for apps to do). So I think you can remove this check.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,129 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+    auto currentAccount = getApp()->getAccounts()->twitch.getCurrent();
+    if (!currentAccount || currentAccount->isAnon())
+    {
+        return;
+    }
+
+    getHelix()->getPinnedChatMessage(
+        this->roomId(), currentAccount->getUserId(),
+        [this](std::optional<HelixPinnedChatMessage> msg) {
+            // Preserve a client-side endsAt if the API doesn't return one
```
Nerixyz: Why is this needed? If the API returns that the pinned message doesn't have an end time, then it should get unpinned when the stream ends.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,129 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+    {
+        return;
+    }
+    this->pinnedMessage_ = std::nullopt;
+    this->pinnedMessageChanged.invoke();
+}
+
+void TwitchChannel::pinMessage(const QString &messageId,
+                               std::optional<int> durationSeconds)
+{
```
Nerixyz: This is already in `pinMessageAs`/`updatePinnedMessageAs`. If you need more functionality, you should update these or add a wrapper.

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -12,6 +12,7 @@
 #include "common/UniqueAccess.hpp"
 #include "providers/ffz/FfzBadges.hpp"
 #include "providers/ffz/FfzEmotes.hpp"
+#include "providers/twitch/api/Helix.hpp"
```
Nerixyz: Instead of including helix, forward declare `HelixPinnedChatMessage`.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2756,7 +2756,6 @@ void ChannelView::addMessageContextMenuItems(QMenu *menu,
                 twitchChannel->deleteMessagesAs(
                     id, getApp()->getAccounts()->twitch.getCurrent().get());
             });
-
```
Nerixyz, line 2785: Undo

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+namespace chatterino {
+
+namespace {
+
+constexpr auto MUTED_STYLE = "color: #adadb8;";
+
+// Small flat icon button helper
+QPushButton *makeIconButton(const QString &text, const QString &tooltip,
+                            QWidget *parent, double fontSize = 12)
```
Nerixyz: Don't use a `QPushButton`, use a `LabelButton` or `SvgButton` (`widgets/buttons`). They will respect the current theme.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,381 @@
...
+    outerBox->setContentsMargins(0, 0, 0, 0);
+    outerBox->setSpacing(0);
+
+    // 1px top border — separates from the split header above
+    auto *topBorder = new QWidget(this);
+    topBorder->setFixedHeight(1);
+    topBorder->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
+    topBorder->setStyleSheet(
+        QStringLiteral("background-color: palette(shadow);"));
+    outerBox->addWidget(topBorder);
```
Nerixyz: I think it's easier to overwrite `paintEvent` and paint a background + border. Then you can also use the current theme colors (`getTheme()`).

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,381 @@
...
+    headerRow->addStretch(1);
+    this->menuButton_ =
+        makeIconButton(QStringLiteral("\u205D"),  // ⁝ tricolon
+                       QStringLiteral("Mod options"), this, 10);
+    this->menuButton_->hide();
+    headerRow->addWidget(this->menuButton_);
+
+    contentBox->addLayout(headerRow);
+
+    // Message body
```
Nerixyz, line 76: General question: Do we want to show emotes in the message? We could also do this in a followup PR.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    this->pinnedByLabel_->setText(
+        QStringLiteral("Pinned by <b>%1</b>")
+            .arg(pin->pinnedBy.displayName.toHtmlEscaped()));
+
+    // deduplicate double @mention that appears when pinning a reply
+    QString displayText = pin->messageText;
+    {
+        static const QRegularExpression doubleAt(
+            QStringLiteral("^(@[^\\s]+) \\1(?=\\s|$)"),
```
Nerixyz: This will also strip leading mentions if the message is not a reply. For example `@user is the winner` → `is the winner`.

Nerixyz (commented):
Thank you for working on this!
Most of these comments are more nitpicks.

One important aspect is that the popup should follow the selected theme.

src/widgets/splits/PinnedMessageWidget.cpp
realtlos: > This will also strip leading mentions if the message is not a reply. For example `@user is the winner` → `is the winner`.

I tired a few different approaches here before settling with this one but now see the issues this can cause. Any recommendations or ideas on how we should handle replies?

Nerixyz: I don't know how Twitch handles these in their UI, but I think it would be fine if we don't handle that for now.

src/providers/twitch/TwitchChannel.cpp
realtlos: You're right this is not needed, I was overthinking it. The Helix API only omits `ends_at` for indefinite pins, and PubSub events don't carry any pin data anyway, they just kick off a re-fetch. Will remove this.

src/providers/twitch/TwitchChannel.hpp
realtlos: > Instead of including helix, forward declare `HelixPinnedChatMessage`.

`HelixPinnedChatMessage` is used as `std::optional<T>` both as a return type and as a member variable, which requires a complete type so a forward declaration alone won't work here i don't think. We'd need to switch the member to `std::unique_ptr` to make it work. Is that worth doing?

src/widgets/splits/PinnedMessageWidget.cpp
realtlos: > General question: Do we want to show emotes in the message? We could also do this in a followup PR.

Yeah, I'd be happy to add emote rendering, though given the complexity maybe a follow up PR makes more sense? What do you think? Either or works for me.

src/providers/twitch/TwitchChannel.hpp
Nerixyz: Ah, I missed that we store it. My main concern is the size of `Helix.hpp`. I think a `unique_ptr` would make sense. It would also avoid copying the struct every time it's returned. You should add a comment that it may be null, because we don't have a good distinction between nullable and non-nullable `unique_ptr`s.

Leafyzito:
Why is "Manually unpinned or end of stream" only 1 option? Can't we have "Indefinite" or "Manually unpin" and "End of stream" separately? Personally I think that makes more sense.

iProdigy:
> Why is "Manually unpinned or end of stream" only 1 option? Can't we have "Indefinite" or "Manually unpin" and "End of stream" separately? Personally I think that makes more sense.

the api doesn't give us that level of granularity: https://dev.twitch.tv/docs/api/reference#pin-chat-message

`duration_seconds 	Integer 	No 	The number of seconds the message should be pinned for. Minimum: 30. Maximum: 1800. If not specified, the message will be pinned until the stream ends.`

2547techno:
IMO "Manually unpinned or" is redundant because its implied for every pin duration. Just "End of Stream" would make the menu cleaner and half the width. Thoughts?

src/widgets/splits/PinnedMessageWidget.cpp
pajlada: follow-up makes most sense if it's more than a few lines of code

src/widgets/DraggablePopup.hpp
```diff
@@ -71,8 +72,8 @@ class DraggablePopup : public BaseWindow
         .light = ":/buttons/pinDisabled-lightMode.svg",
     };
     SvgButton::Src pinEnabledSource_{
-        .dark = ":/buttons/pinEnabled.svg",
-        .light = ":/buttons/pinEnabled.svg",
+        .dark = ":/buttons/pinnedMessage-chat.svg",
+        .light = ":/buttons/pinnedMessage-chat.svg",
```
Mm2PL: Find and replace mistake. This is the icon for automatically closing Reply Threads and User Cards.

src/widgets/PluginRepl.hpp
```diff
@@ -74,8 +74,8 @@ class PluginRepl : public BaseWindow
             .light = ":/buttons/pinDisabled-lightMode.svg",
         };
         SvgButton::Src pinEnabledSource_{
-            .dark = ":/buttons/pinEnabled.svg",
-            .light = ":/buttons/pinEnabled.svg",
+            .dark = ":/buttons/pinnedMessage-chat.svg",
+            .light = ":/buttons/pinnedMessage-chat.svg",
```
Mm2PL: This change is not intended right?

src/widgets/PluginRepl.cpp
```diff
@@ -520,6 +520,8 @@ void PluginRepl::themeChangedEvent()
 
     pal.setColor(QPalette::Base, getTheme()->splits.input.background);
     this->ui.input->setPalette(pal);
+
+    this->ui.pin->setColor(getTheme()->isLightTheme() ? QColor(0x4d, 0x4d, 0x4d) : QColor(0xb7, 0xb7, 0xb7));
```
Mm2PL: This pin button is for putting the repl to always be the topmost window.

src/widgets/DraggablePopup.cpp
```diff
@@ -113,11 +114,22 @@ void DraggablePopup::togglePinned()
...
+void DraggablePopup::themeChangedEvent()
+{
+    BaseWindow::themeChangedEvent();
+
+    if (this->pinButton_)
+    {
+        this->pinButton_->setColor(this->theme->isLightTheme() ? QColor(0x4d, 0x4d, 0x4d) : QColor(0xb7, 0xb7, 0xb7));
+    }
+}
+
```
Mm2PL: This is the pin button for windows that automatically close. I'd prefer this to not be touched.

Image

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    this->pinnedByLabel_->setText(
+        QStringLiteral("Pinned by <b>%1</b>")
+            .arg(pin->pinnedBy.displayName.toHtmlEscaped()));
+
+    this->messageLabel_->setPlainText(pin->messageText);
+
+    {
+        const QString sentAt =
+            pin->startsAt.toLocalTime().toString(QStringLiteral("h:mm AP"));
```
Mm2PL: Not everyone likes `1:23 PM` as a time format. This should be configurable, guess based on presence of am/pm in the _Message timestamp format_ setting or reuse that timestamp format.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,379 @@
...
+        QFont f = this->pinnedByLabel_->font();
+        f.setPointSize(11);
+        this->pinnedByLabel_->setFont(f);
+    }
+    headerRow->addWidget(this->pinnedByLabel_);
+    headerRow->addStretch(1);
+    this->menuButton_ = new LabelButton(QStringLiteral("\u205D"));
+    this->menuButton_->setToolTip(QStringLiteral("Mod options"));
+    this->menuButton_->setFixedSize(30, 22);
+    this->menuButton_->hide();
```
Mm2PL: Could this be a DrawnButton(DrawnButton::Symbol::Kebab)? It then would look the same as the one in the SplitHeader.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,379 @@
...
+    headerRow->addWidget(this->menuButton_);
+
+    contentBox->addLayout(headerRow);
+
+    // Message body
+    this->messageLabel_ = new QTextEdit(this);
```
Mm2PL: The text in this is very hard to read in light themes. You're fine to just ignore this if you're planning on changing this to a MessageView or a clone of it (you'd need to do some changes there to actually display emotes not just a SingleLineTextElement).

Mm2PL (dismissed):
There's still some bugs to iron out

Right clicking on the overlay textedit opens two menus, one for the QTextEdit and one for the entire Notebook (the widget that holds tabs and all splits).
Image

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,379 @@
+﻿// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#include "widgets/splits/PinnedMessageWidget.hpp"
+
+#include "Application.hpp"
+#include "controllers/accounts/AccountController.hpp"
```
Mm2PL, line 8: missed an include
```suggestion
#include "controllers/accounts/AccountController.hpp"
#include "providers/twitch/api/Helix.hpp"
```

src/widgets/splits/PinnedMessageWidget.cpp
realtlos: > The text in this is very hard to read in light themes. You're fine to just ignore this if you're planning on changing this to a MessageView or a clone of it (you'd need to do some changes there to actually display emotes not just a SingleLineTextElement).

I do plan to replace it with a MessageView later, but I can fix the light theme readability in the meantime if you'd like.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,67 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+    auto currentAccount = getApp()->getAccounts()->twitch.getCurrent();
+    if (!currentAccount || currentAccount->isAnon())
+    {
+        return;
+    }
+
+    const auto msgId = this->pinnedMessage_->messageID;
+    getHelix()->unpinChatMessage(
+        this->roomId(), currentAccount->getUserId(), msgId,
+        [this] {
```
Nerixyz: See other comment.

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -602,6 +628,9 @@ class TwitchChannel final : public Channel, public ChannelChatters
     eventsub::SubscriptionHandle eventSubChannelChatUserMessageHoldHandle;
     eventsub::SubscriptionHandle eventSubChannelChatUserMessageUpdateHandle;
 
+    /// May be null if no message is currently pinned.
+    std::unique_ptr<HelixPinnedChatMessage> pinnedMessage_;
```
Nerixyz:
```suggestion
    std::unique_ptr<const HelixPinnedChatMessage> pinnedMessage_;
```

I assume this can be const since we don't modify it anywhere.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,67 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+{
+    auto currentAccount = getApp()->getAccounts()->twitch.getCurrent();
+    if (!currentAccount || currentAccount->isAnon())
+    {
+        return;
+    }
+
+    getHelix()->getPinnedChatMessage(
+        this->roomId(), currentAccount->getUserId(),
+        [this](std::optional<HelixPinnedChatMessage> msg) {
```
Nerixyz: You can't capture `this` safely. You can use `this->weakFromThis()` to get a weak reference that you can lock in the callback.

resources/buttons/pinnedMessage-chat.svg
Nerixyz, line 1: Is this stitched together from Fluent-UI icons? Just asking if we need to provide credit for this somewhere. We already list Fluent-UI.

realtlos: > Is this stitched together from Fluent-UI icons? Just asking if we need to provide credit for this somewhere. We already list Fluent-UI.

This design was actually created entirely by me, using only the message outline and a few reference images from Google for inspiration. The artwork itself was hand drawn in an online [SVG editor](https://yqnn.github.io/svg-path-editor/) without directly tracing or copying any references. I also worked with a friend during parts of the process, and I’m happy to provide proof of the design’s development or any other information if needed.

Nerixyz: > The artwork itself was hand drawn in an online [SVG editor](https://yqnn.github.io/svg-path-editor/)

That's great! I love that site.

src/providers/twitch/TwitchChannel.cpp
```diff
@@ -2550,4 +2558,79 @@ bool TwitchChannel::isLoadingRecentMessages() const
...
+            auto self = weak.lock();
+            if (!self)
+            {
+                return;
+            }
+            self->pinnedMessage_ =
+                msg ? std::make_unique<const HelixPinnedChatMessage>(
+                          std::move(*msg))
+                    : nullptr;
+            self->pinnedMessageChanged.invoke();
```
pajlada: I'd prefer if this didn't use a ternary operator - just have two separate paths for if msg is set or not set. It's fine to repeat yourself here

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -399,6 +402,30 @@ class TwitchChannel final : public Channel, public ChannelChatters
...
+    void refreshPinnedMessage();
+
+    /**
+     * Clears the pinned message for this channel immediately (e.g. on unpin
+     * PubSub event).
+     */
+    void clearPinnedMessage();
+
+    /// Returns the currently pinned message, or null if none is pinned.
+    const HelixPinnedChatMessage *getPinnedMessage() const;
```
pajlada, line 418: Instead of returning a raw pointer, return a const reference to the `std::unique_ptr<...>` instead.
Returning a raw pointer implies, to me, that the pointer may be stored for future use later, which is not how it's used.

src/widgets/splits/PinnedMessageWidget.cpp
pajlada: I'm fine with leaving this as-is for now, and it becoming a MessageView with emote support etc in a follow-up PR.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    const bool isMod = this->channel_->hasModRights();
+    this->menuButton_->setVisible(isMod);
+
+    this->show();
+
+    this->autoHideTimer_->stop();
+    if (!getSettings()->alwaysShowPinnedMessage && !this->userToggled_)
+    {
+        this->autoHideTimer_->start(5000);
```
pajlada: This should, imo, be visible for at least 30 seconds

src/widgets/splits/PinnedMessageWidget.hpp
```diff
@@ -0,0 +1,64 @@
...
+public:
+    explicit PinnedMessageWidget(QWidget *parent = nullptr);
+
+    // Pass nullptr to detach from any channel.
+    void setChannel(TwitchChannel *channel);
+
+    // Called by the header pin button to toggle manual visibility.
+    void toggleUserPinned();
+
+    [[nodiscard]] bool hasPinnedMessage() const;
```
pajlada: Looks like this function is unused.
If you have future plans for it, add a comment to the function & let me know in a response here.
If you do _not_ have future plans for it, remove the function.

src/widgets/splits/SplitHeader.cpp
```diff
@@ -394,6 +408,10 @@ void SplitHeader::initializeLayout()
                          this->split_->openChatterList();
                      });
 
+    QObject::connect(this->pinButton_, &Button::leftClicked, this, [this]() {
+        this->split_->togglePinnedBanner();
```
pajlada, line 413: If possible, I'd love if the color of the "pinned message" banner in the split header changes color if the pinned message is currently visible or not.

src/widgets/DraggablePopup.cpp
```diff
@@ -118,11 +114,15 @@ void DraggablePopup::togglePinned()
...
 }
+
 Button *DraggablePopup::createPinButton()
 {
     this->pinButton_ = new SvgButton(this->pinDisabledSource_, this, {3, 3});
     this->pinButton_->setScaleIndependentSize(18, 18);
     this->pinButton_->setToolTip("Pin Window");
+    this->pinButton_->setColor(this->theme->isLightTheme()
+                                   ? QColor(0x4d, 0x4d, 0x4d)
+                                   : QColor(0xb7, 0xb7, 0xb7));
```
pajlada: The changes to this file seem unrelated - please undo them all.
If I'm mistaken, please correct me :-)

mocks/include/mocks/Helix.hpp
pajlada: Need to resolve Nerixyz's comments

pajlada (changes requested):
the pr works, neat! a bunch of nits you can start resolving 🙏

mocks/include/mocks/Helix.hpp
```diff
@@ -505,33 +505,33 @@ class Helix : public IHelix
                 (override));
 
     MOCK_METHOD(
-        void, pinChatMessage,
```
pajlada, line 508: @realtlos You've reordered these functions - please undo that change. When you're done, the `mocks/include/mocks/Helix.hpp` file should show as having 0 changes

src/singletons/Settings.hpp
```diff
@@ -249,6 +249,8 @@ class Settings
                                      false};
     BoolSetting headerGame = {"/appearance/splitheader/showGame", false};
     BoolSetting headerUptime = {"/appearance/splitheader/showUptime", false};
+    BoolSetting alwaysShowPinnedMessage = {
+        "/appearance/alwaysShowPinnedMessage", false};
```
pajlada: This is behaviour of the pinned message hide-logic, not appearance. Move to the `/// Behaviour` section instead

pajlada (commented):
The menu button takes up a lot of space in the height, causing a big space between the "pinned by" message and the actual pinned message, would be nice if that didn't happen

Image

EDIT: This is exaggerated by zooming your Chatterino in - so the text currently doesn't scale.
If that's easy to resolve, feel free to resolve it, otherwise you can make a TODO comment for the followup PR that changes the message rendering to a MessageView

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+    this->messageLabel_->setPlainText(pin->messageText);
+
+    {
+        const QString sentAt = pin->startsAt.toLocalTime().toString(
+            getSettings()->timestampFormat);
+        this->footerLabel_->setText(
+            QStringLiteral("Sent by %1 \u00B7 %2")
+                .arg(pin->sender.displayName.toHtmlEscaped(), sentAt));
+    }
```
pajlada: It would be nice if this could take the `usernameDisplayMode` setting into consideration. Should maybe be a function inside the `HelixMinimalUser` struct that accepts the `UsernameDisplayMode` enum and formats things based on that

src/widgets/DraggablePopup.hpp
pajlada: @realtlos this was not actually resolved, same with the other changes to DraggablePopup.cpp

pajlada (changes requested):
Just the unrelated changes in DraggablePopup, then this looks mergable to me.

There's some polish that's missing but I think that's best done in a follow-up PR.

realtlos:
> Just the unrelated changes in DraggablePopup, then this looks mergable to me.
> 
> There's some polish that's missing but I think that's best done in a follow-up PR.

Just to confirm should I revert all the `DraggablePopup` changes (the `themeChangedEvent`, `setColor`, `#include "singletons/Theme.hpp"`, and the wrong icon in .hpp), or just the clang-format reformatting from the last commit?

pajlada:
> > Just the unrelated changes in DraggablePopup, then this looks mergable to me.
> > There's some polish that's missing but I think that's best done in a follow-up PR.
> 
> Just to confirm should I revert all the `DraggablePopup` changes (the `themeChangedEvent`, `setColor`, `#include "singletons/Theme.hpp"`, and the wrong icon in .hpp), or just the clang-format reformatting from the last commit?

All of them - none of the changes you've made to DraggablePopup are actually relevant to your pinned message widget right?

realtlos:
> > > Just the unrelated changes in DraggablePopup, then this looks mergable to me.
> > > There's some polish that's missing but I think that's best done in a follow-up PR.
> > 
> > 
> > Just to confirm should I revert all the `DraggablePopup` changes (the `themeChangedEvent`, `setColor`, `#include "singletons/Theme.hpp"`, and the wrong icon in .hpp), or just the clang-format reformatting from the last commit?
> 
> All of them - none of the changes you've made to DraggablePopup are actually relevant to your pinned message widget right?

Reverted all the DraggablePopup changes as requested.

I also included a small follow up fix: I noticed long pinned messages were getting cut off in narrow splits. I swapped the message body from a `QTextEdit` to a `QLabel` inside a `QScrollArea` so the text wraps to the banner width, the banner sizes to its content, and it scrolls once it exceeds the max height.

src/providers/twitch/TwitchChannel.hpp
```diff
@@ -399,6 +402,31 @@ class TwitchChannel final : public Channel, public ChannelChatters
...
+    void refreshPinnedMessage();
+
+    /**
+     * Clears the pinned message for this channel immediately (e.g. on unpin
+     * PubSub event).
+     */
+    void clearPinnedMessage();
+
+    /// Returns the currently pinned message, or null if none is pinned.
+    const std::unique_ptr<const HelixPinnedChatMessage> &getPinnedMessage()
```
Nerixyz:
```suggestion
    const HelixPinnedChatMessage *getPinnedMessage()
```

You can unwrap the unique pointer here.

src/common/enums/UsernameDisplayMode.hpp
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+namespace chatterino {
+
+enum UsernameDisplayMode : int {
```
Nerixyz, line 9: Can we turn this into a `enum class`? I don't remember if the settings allow this. Similarly for `int` -> `uint8_t`.

src/widgets/splits/PinnedMessageWidget.hpp
```diff
@@ -0,0 +1,76 @@
...
+    /// (scaled) maximum height. A vertical scrollbar appears past the cap.
+    void updateMessageHeight();
+
+    TwitchChannel *channel_ = nullptr;
+    pajlada::Signals::SignalHolder signalHolder_;
+
+    // Header row
+    QLabel *pinnedByLabel_ = nullptr;
+    QLabel *countdownLabel_ = nullptr;
+    DrawnButton *menuButton_{};  // mod menu
```
Nerixyz:
```suggestion
    /// Mod Menu.
    DrawnButton *menuButton_ = nullptr;
```

src/widgets/splits/PinnedMessageWidget.hpp
```diff
@@ -0,0 +1,76 @@
...
+    // Body
+    QScrollArea *messageScrollArea_ = nullptr;
+    QLabel *messageLabel_ = nullptr;
+    QLabel *footerLabel_ = nullptr;
+
+    QTimer *progressTimer_ = nullptr;
+    QTimer *autoHideTimer_ = nullptr;
+    int messageMaxHeight_ = 110;  // scaled cap for the message body
+    bool userToggled_ = false;    // true while user manually pinned the widget
+    QDateTime pinEndsAt_;         // invalid when no end time
```
Nerixyz:
```suggestion
    /// Scaled cap for the message body.
    int messageMaxHeight_ = 110;
    /// True while user manually pinned the widget.
    bool userToggled_ = false;
    /// Invalid when no end time.
    QDateTime pinEndsAt_;
```

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,416 @@
...
+    contentBox->setSpacing(3);
+
+    // Header row: "Pinned by <user>"  [⋮]
+    auto *headerRow = new QHBoxLayout();
+    headerRow->setSpacing(4);
+
+    headerRow->addWidget(this->pinnedByLabel_);
+    headerRow->addStretch(1);
+    this->menuButton_->setScaleIndependentSize(28, 28);
+    this->menuButton_->setToolTip(QStringLiteral("Mod options"));
```
Nerixyz:
```suggestion
    this->menuButton_->setToolTip(u"Mod options"_s);
```

And add `using namespace Qt::Literals` at the top of the file. These literals are more compact. Same for the other uses of `QStringLiteral` in this file.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,416 @@
...
+    , pinnedByLabel_(new QLabel(this))
+    , countdownLabel_(new QLabel(this))
+    , menuButton_(new DrawnButton(DrawnButton::Symbol::Kebab, {}, this))
+    , messageLabel_(new QLabel(this))
+    , footerLabel_(new QLabel(this))
+    , progressTimer_(new QTimer(this))
+    , autoHideTimer_(new QTimer(this))
+{
+    this->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Minimum);
+    this->setAttribute(Qt::WA_OpaquePaintEvent);
```
Nerixyz: Does this work if Chatterino is transparent? This is only implemented on Linux right now.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,416 @@
...
+                return;
+            }
+            auto currentAccount = getApp()->getAccounts()->twitch.getCurrent();
+            if (!currentAccount || currentAccount->isAnon())
+            {
+                return;
+            }
+            std::optional<std::chrono::seconds> duration =
+                seconds ? std::make_optional(std::chrono::seconds(*seconds))
+                        : std::nullopt;
```
Nerixyz: You already added `using namespace std::chrono_literals` at the top. So you can pass the arguments as `123s`.

src/widgets/splits/PinnedMessageWidget.cpp
```diff
@@ -0,0 +1,416 @@
...
+                                           });
+        this->signalHolder_.managedConnect(channel->userStateChanged, [this] {
+            this->refresh();
+        });
+    }
+
+    this->refresh();
+}
+
+void PinnedMessageWidget::showModMenu()
```
Nerixyz: You can add a `QMenu` to the `DrawnButton` with [`setMenu`](https://github.com/Chatterino/chatterino2/blob/305149a7957571e097af5e3c92f4ab1605ec050a/src/widgets/buttons/Button.hpp#L94-L105).

src/widgets/splits/SplitHeader.cpp
```diff
@@ -837,6 +856,28 @@ void SplitHeader::handleChannelChanged()
             twitchChannel->streamStatusChanged, [this]() {
                 this->updateChannelText();
             });
+
+        this->channelConnections_.managedConnect(
+            twitchChannel->pinnedMessageChanged, [this]() {
+                auto ch = this->split_->getChannel();
+                auto *tc = dynamic_cast<TwitchChannel *>(ch.get());
+                this->updatePinButton(tc != nullptr &&
+                                      tc->getPinnedMessage() != nullptr);
```
Nerixyz: The `tc != nullptr && tc->getPinnedMessage() != nullptr` check is duplicated below. You could move it inside the `updatePinButton` method.

src/common/enums/UsernameDisplayMode.hpp
pajlada: Let's leave that for a separate PR

src/widgets/splits/PinnedMessageWidget.cpp
realtlos: > Does this work if Chatterino is transparent? This is only implemented on Linux right now.

Good catch, it doesn't. I'll drop `WA_OpaquePaintEvent` to match `SplitHeader`, which doesn't set it and just fills the background directly. Thanks!

pajlada:
for me, the height does not get correctly resized. is this an issue for you on windows(if that's your dev os)?
https://github.com/user-attachments/assets/7e52e563-adc2-4626-bbd4-2ccdfad2873f

realtlos:
> for me, the height does not get correctly resized. is this an issue for you on windows(if that's your dev os)? https://github.com/user-attachments/assets/7e52e563-adc2-4626-bbd4-2ccdfad2873f

just checked and it seems to not be an issue on windows. I can take a look on linux and figure out the issue.

pajlada:
Thanks, appreciate it. If you take a look and can't figure it out, we can leave it a bit broken on Linux. It's an operating system for nerds anyway :nerd_face:

pajlada (approved):

pajlada:

PR #7035, opened 2026-06-07
refactor(plugins): Use `QStringList` for enabled plugins

Nerixyz:
`QStringList` has two advantages over `std::vector<QString>` for settings:
1. Cheap to copy (COW). Setting values need to be copied when they're accessed, because they're behind a mutex.
2. Better helper functions (see changes).


src/util/serialize/StringList.hpp
```diff
@@ -0,0 +1,55 @@
...
+#pragma once
+
+#include <pajlada/serialize.hpp>
+#include <QString>
+#include <QStringList>
+
+namespace pajlada {
+
+template <>
+struct Serialize<QStringList> {
```
pajlada: Could we make this a templated QList serialize/deserialize instead? similar to how the bill-tin std::vector<T> works

pajlada (commented):
I'm OK with the change - the standard container functions is definitely missing some nice things

src/util/serialize/StringList.hpp
Nerixyz: Sure.

pajlada (approved):

pajlada:
@Mm2PL did you want a chance to take a look at this one?

Mm2PL (approved):

PR #7038, opened 2026-06-07
refactor: Move uncategorized notifications to their own message flag

### LINKED ISSUE #7046: Messages attached to subscriptions that meet the requirements to be filtered don't show up in mentions tab
### Checklist

### Describe your issue

Messages that would trigger certain filters that I have in order to make them show up in my mentions tab don't show up if they're in sub messages. The sound still plays, and chatterino still flashes on the taskbar, but the message itself doesn't show up in mentions, and the tab that the message was sent in doesn't get highlighted. This happens with both author.name filters and "message" highlights. I don't have any other types so I don't know if it's the same for them. 

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.5 (commit [1bc6e5aa](https://github.com/Chatterino/chatterino2/commit/1bc6e5aa))

pajlada:
Currently, when a "special" message is sent from Twitch, and we don't have a good category for it, we categorize it as a Subscriptions.

This is confusing when working with message parsing, and difficult with the highlights rework I'm working on.

In this PR, I've added a new message flag `UncategorizedNotification` that gets used for that instead.

Some of the changes to snapshot tests look a bit off, but I have done visual inspections and everything still work as expected

image


tests/snapshots/IrcMessageHandler/mod-streak.json
```diff
@@ -156,7 +156,6 @@
             ],
             "flags": "Collapsed|WatchStreak",
             "frozen": false,
-            "highlightColor": "#64c466ff",
```
Nerixyz, line 159: This highlight color is removed.

tests/snapshots/IrcMessageHandler/watch-streak.json
```diff
@@ -147,7 +147,6 @@
             ],
             "flags": "Collapsed|WatchStreak",
             "frozen": false,
-            "highlightColor": "#64c466ff",
```
Nerixyz, line 150: Same here.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -1280,15 +1315,9 @@ void IrcMessageHandler::addMessage(Communi::IrcMessage *message,
...
-            else
-            {
-                msg->flags.set(MessageFlag::Subscription);
-            }
-
-            if (tags.value("msg-id") != "announcement")
+            else if (!KNOWN_MESSAGE_TYPES.contains(msgType))
             {
-                // We want announcements to be able to show up in mentions
-                msg->flags.unset(MessageFlag::Highlighted);
```
4rneee, line 1291: This also fixes sub/special messages not showing up in mentions 👍

src/messages/MessageBuilder.cpp
```diff
@@ -1707,10 +1707,22 @@ std::pair<MessagePtrMut, HighlightAlert> MessageBuilder::makeIrcMessage(
...
+        if (msgIdTypes.contains("highlighted-message"))
+        {
+            builder->flags.set(MessageFlag::RedeemedHighlight);
+        }
+
+        if (msgIdTypes.contains("viewermilestone") ||
+            msgIdTypes.contains("modiversary"))
+        {
+            builder->flags.set(MessageFlag::WatchStreak);
+        }
```
4rneee, line 1725: Is there a reason to move Watchstreak messages to `makeIrcMessage` but leave Announcement messages in `IrcMessageHandler::addMessage`? I think it would be good if all special message types are handled in one location.

tests/snapshots/IrcMessageHandler/mod-streak.json
pajlada: Because the `Highlighted` flag is not set, `highlightColor` is actually never use. So from the perspective of this PR & previous behaviour, thing still work as expected.

In the grand scope of the highlight rework, this will need to be fixed, but from my perspective I'm better of fixing that when I actually apply the `Highlighted` flag in that PR.

tests/snapshots/IrcMessageHandler/watch-streak.json
pajlada: https://github.com/Chatterino/chatterino2/pull/7038/changes#r3407785687

src/messages/MessageBuilder.cpp
pajlada: Mostly because I want to touch as few things as possible. This PR is spun off from the big highlights rework PR and I think there will be a few more similar PRs from there.

PR #7052, opened 2026-06-14
refactor: Move more message flag parsing into MessageBuilder

pajlada:

### DIFF
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -586,7 +586,7 @@ MessageBuilder::MessageBuilder(SystemMessageTag, const QString &text,
 MessagePtrMut MessageBuilder::makeSystemMessageWithUser(
     const QString &text, const QString &loginName, const QString &displayName,
     const MessageColor &userColor, const QTime &time,
-    const Communi::IrcMessage *ircMessage)
+    const Communi::IrcMessage &ircMessage)
 {
     MessageBuilder builder;
     builder.emplace<TimestampElement>(time);
@@ -609,7 +609,7 @@ MessagePtrMut MessageBuilder::makeSystemMessageWithUser(
     builder->messageText = text;
     builder->searchText = text;
 
-    auto tags = ircMessage->tags();
+    auto tags = ircMessage.tags();
 
     builder.parseMessageTags(tags);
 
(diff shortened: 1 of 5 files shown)


src/messages/MessageBuilder.cpp
```diff
@@ -562,7 +585,8 @@ MessageBuilder::MessageBuilder(SystemMessageTag, const QString &text,
 
 MessagePtrMut MessageBuilder::makeSystemMessageWithUser(
     const QString &text, const QString &loginName, const QString &displayName,
-    const MessageColor &userColor, const QTime &time)
+    const MessageColor &userColor, const QTime &time,
+    const Communi::IrcMessage *ircMessage)
```
Nerixyz:
```suggestion
    const Communi::IrcMessage &ircMessage)
```

This can't be null, so we should use a reference to signal this to callers.

src/providers/twitch/IrcMessageHandler.cpp
```diff
@@ -798,6 +785,7 @@ void IrcMessageHandler::parseUserNoticeMessageInto(Communi::IrcMessage *message,
         return;
     }
 
+    // TODO: Why are we ONLY allowing these message types to have an additional message with their content added?
```
Nerixyz, line 785: Yea, this feels a bit wrong. I think we should show the message by default and leave it out for types where it's not needed. This way we don't need to push an update every time Twitch adds a new message type with content for users to be able to see that.

src/messages/MessageBuilder.cpp
pajlada: Resolved in [`549d84a` (this PR)](https://github.com/Chatterino/chatterino2/pull/7052/commits/549d84a8995c16ecd3d8901712f18de2b5c58a25)

src/messages/MessageBuilder.cpp
```diff
@@ -84,6 +84,8 @@ const QRegularExpression allUsernamesMentionRegex("^" + regexHelpString);
 
 const QRegularExpression SPACE_REGEX("\\s");
 
+const QString ANONYMOUS_GIFTER_ID = "274598607";
```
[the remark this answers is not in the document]
pajlada: resolved in [`646870a` (this PR)](https://github.com/Chatterino/chatterino2/pull/7052/commits/646870a7a07a419bc0443aa67d627adf3f4f35e2)

PR #7058, opened 2026-06-18
feat: Set window roles for session restoration

### LINKED ISSUE #6655: Set window roles for session restoration on Wayland
To solve #6484 on supported compositors:

While we don't control the compositor, we should forward as much information to it as possible. For session restoration, the compositor needs to have some ID for the windows we're creating.

Qt 6.11 adds support for the experimental [`xx-session-management-v1`](https://wayland.app/protocols/xx-session-management-v1) (https://github.com/qt/qtbase/commit/0094c73c3ce255f9c0b51f491caee518aef078df - currently, only supported by mutter). To tap into this, we need to set the `windowRole` of the top level widget (through [`setWindowRole`](https://doc.qt.io/qt-6/qwidget.html#setWindowRole)). This has to be set when creating that top widget.
The identifier is up to us. We could store this in the [window descriptor](https://github.com/Chatterino/chatterino2/blob/9f46cae6c06e1045f58a3630c5b3d5956a65b177/src/common/WindowDescriptors.hpp#L82-L95).

For the main window, it's pretty simple (use `main`), but for popups it's a bit more complicated. Popups should remain at their positions when restarting Chatterino and new popups should open at the position where the last popup was closed. Maybe we could use something like `popup-{uuid}` and keep track of the last closed identifier? It's best to test this with a compositor that supports the protocol.

The following windows should get a role:

- Main windows
- Popups
- Overlays
- Emote popup

Nerixyz:
This sets the window roles for our windows similar to what I described in #6655. All roles are prefixed with `chatterino.` to avoid collision in the case where we're embedded.

- Main Window: `chatterino.main`
- Popups: `chatterino.popup.{id}`
- Emote popup: `chatterino.emote-popup`
- Settings: `chatterino.settings`
- Overlay: `chatterino.overlay`

Right now, you need to pass `-session {session-id}_0` when starting, but I hope future Qt versions add a way to set the session ID after the `QGuiApplication` is created. Note that because of this, restoration won't work out of the box. That's expected. You can see the session status in the logs of `qt.qpa.wayland`.

Tested with KDE 6.7 and Qt 6.12-beta1.

Closes #6655.


Mm2PL:
Can we make pid files in `/run/user/ID/` (or whatever `$XDG_blah` points to it) to give each started Chatterino instance a unique session number?

Nerixyz:
> Can we make pid files in `/run/user/ID/` (or whatever `$XDG_blah` points to it) to give each started Chatterino instance a unique session number?

I think we should save the session ID in the window layout file. This should match what happens on macOS, Windows, and X11 where we restore the previous window layout. If there isn't a session ID, we let the compositor give us one.

Nerixyz:
> I think we should save the session ID in the window layout file.

This is how I imagine it to work: https://github.com/Nerixyz/chatterino2/commit/70f2a841ab52e6dc536694b8f82259792f00b005. I'll open a change request in Qt for this.

src/widgets/Window.hpp
```diff
@@ -34,6 +34,9 @@ class Window : public BaseWindow
     WindowType getType();
     SplitNotebook &getNotebook();
 
+    void setPopupID(size_t ID);
```
pajlada: `id` instead of `ID`

pajlada (changes requested):

pajlada (approved):

Mm2PL (approved):
The code looks fine, but I don't have time to test everything

PR #7063, opened 2026-06-22
feat: add tab history navigation

### LINKED ISSUE #6063: Keep selected tab history & add hotkey to move through tab history
When changing your selected tab, the tab you changed from should be added to your tab history.

A user should be able to add a hotkey (e.g. `alt+left arrow`) to focus the previous tab.

### Implementation details to think about

Branching history: Focusing the previous tab then clicking another tab - now the "focus "next" tab" has no target right?

### Other implementations

Discord allows you to press a mouse side button to go to the previously selected channel
VS Code allows you to jump to the last cursor position (even across files)

Feature request from https://github.com/Chatterino/chatterino2/discussions/6060

Leafyzito:
Adds browser-style tab history navigation to main windows, popup windows and emote popup.

Default bindings: ``Ctrl+Alt+Left`` / ``Ctrl+Alt+Right`` to go back/forward in tab history; ~~mouse back/forward buttons work too~~ - to be added in a future PR

Closed tabs are skipped and pruned from the stack instead of being reopened
The quick switcher shows recently visited tabs when the search box is empty

Closes #6063


src/widgets/TabHistory.hpp
```diff
@@ -0,0 +1,36 @@
...
+    std::optional<QWidget *> peekForward() const;
+    std::vector<QWidget *> backStackMostRecentFirst() const;
+    void discardBackTop();
+    void discardForwardTop();
+
+private:
+    static void removeFromStack(std::vector<QWidget *> &stack, QWidget *page);
+
+    std::vector<QWidget *> back_;
+    std::vector<QWidget *> forward_;
```
Nerixyz: I think you can use a `boost::circular_buffer(_space_optimized)` or `std::deque` here - they're more efficient when removing items from the start. For the current position, you can store an index.

src/util/TabHistory.hpp
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <QWidget>
+
+#include <optional>
+#include <vector>
+
+namespace chatterino {
+
+class TabHistory
```
Nerixyz, line 21: This could use some documentation and tests.

src/widgets/Notebook.cpp
```diff
@@ -374,7 +387,92 @@ void Notebook::select(QWidget *page, bool focusPage)
...
+
+    return *previous;
+}
+
+std::vector<QWidget *> Notebook::getVisitHistoryPages() const
+{
+    std::vector<QWidget *> pages;
+    for (auto *page : this->tabHistory_.backStackMostRecentFirst())
+    {
+        if (this->containsPage(page))
```
Nerixyz: Isn't this always true?

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -404,6 +404,32 @@ void EmotePopup::addShortcuts()
              }
              return "";
          }},
+        {"selectTabHistory",
+         [this](std::vector<QString> arguments) -> QString {
```
```suggestion
         [this](const std::vector<QString> &arguments) -> QString {
```

src/widgets/TabHistory.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace {
+
+constexpr size_t MAX_TAB_HISTORY_SIZE = 50;
+
+}  // namespace
+
+namespace chatterino {
+
+void TabHistory::recordVisit(QWidget *from, QWidget *to)
```
Nerixyz: `to` isn't used other than a check if it's the same as `from`.

src/util/TabHistory.hpp
Nerixyz, line 1: This file doesn't contain a widget. It should go somewhere else. Maybe in `util/` or some new subdirectory there?

Leafyzito:
later today i'll add tests and address clang's comments too

src/util/TabHistory.hpp
```diff
@@ -0,0 +1,62 @@
...
+    std::optional<QWidget *> peekForward() const;
+
+    /// Returns back-stack entries from most recently visited to oldest.
+    std::vector<QWidget *> backStackMostRecentFirst() const;
+
+    /// Removes the top back entry without navigating.
+    void discardBackTop();
+
+    /// Removes the top forward entry without navigating.
+    void discardForwardTop();
```
Nerixyz, line 51: What's the "top" entry? At least for the backward case this kind-of makes sense, but for the forward case I could think of either.

tests/src/TabHistory.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+class TabHistoryTest : public ::testing::Test
+{
+protected:
+    QWidget parent;
+};
+
+}  // namespace
+
+TEST_F(TabHistoryTest, RecordVisitBuildsBackStack)
```
Nerixyz: Test that the forward history entries are not included.

tests/src/TabHistory.cpp
```diff
@@ -0,0 +1,156 @@
...
+    QWidget b(&this->parent);
+    QWidget c(&this->parent);
+    TabHistory history;
+
+    history.recordVisit(&a);
+    history.recordVisit(&b);
+
+    auto back = history.goBack(&c);
+    ASSERT_TRUE(back.has_value());
+    EXPECT_EQ(*back, &b);
```
Nerixyz: You should be able to use `EXPECT_EQ(back, &b)`. Same for the other comparisons with `std::optional`.

tests/src/TabHistory.cpp
```diff
@@ -0,0 +1,156 @@
...
+    history.recordVisit(&b);
+    history.goBack(&c);
+    ASSERT_TRUE(history.canGoForward());
+
+    history.recordVisit(&b);
+
+    EXPECT_FALSE(history.canGoForward());
+}
+
+TEST_F(TabHistoryTest, RemovePage)
```
Nerixyz: Test how this behaves with the `peekBack`/`peekForward`.

src/util/TabHistory.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    if (current != nullptr)
+    {
+        this->history_.push_back(current);
+    }
+
+    return target;
+}
+
+std::optional<QWidget *> TabHistory::goForward(QWidget *current)
```
Nerixyz: Why do we not track the current entry in this class? It feels weird that `goForward` needs to insert a page. Shouldn't it increment the current page index and return that page?

src/util/TabHistory.hpp
```diff
@@ -0,0 +1,59 @@
...
+    std::vector<QWidget *> backStackMostRecentFirst() const;
+
+    /// Removes the most recent back entry (the one @ref peekBack would return).
+    void discardBackTop();
+
+    /// Removes the next forward entry (the one @ref peekForward would return).
+    void discardForwardTop();
+
+private:
+    static constexpr size_t MAX_TAB_HISTORY_SIZE = 50;
```
pajlada: Is this the maximum number of tabs stored inside `history_`?

src/widgets/Notebook.cpp
```diff
@@ -38,6 +38,12 @@
 #include <ranges>
 #include <utility>
 
+namespace {
+
+constexpr size_t MAX_TAB_HISTORY_NAV_ATTEMPTS = 50;
```
pajlada: Is this different to `TabHistory`'s `MAX_TAB_HISTORY_SIZE`?
If not, re-use it.
If it is, document both entries.

src/widgets/Notebook.hpp
```diff
@@ -70,7 +72,13 @@ class Notebook : public BaseWidget
     /**
      * @brief Selects the Notebook tab containing the given page.
      **/
-    virtual void select(QWidget *page, bool focusPage = true);
+    virtual void select(QWidget *page, bool focusPage = true,
+                        bool recordInHistory = true);
+
+    void selectHistoryBack(bool focusPage = true);
+    void selectHistoryForward(bool focusPage = true);
```
pajlada: Remove the default value here please

src/widgets/Notebook.cpp
```diff
@@ -374,7 +387,83 @@ void Notebook::select(QWidget *page, bool focusPage)
...
+
+        this->tabHistory_.discardForwardTop();
+    }
+}
+
+bool Notebook::selectHistoryPage(bool forward, bool focusPage)
+{
+    for (size_t attempt = 0; attempt < MAX_TAB_HISTORY_NAV_ATTEMPTS; ++attempt)
+    {
+        this->pruneInvalidHistoryEntries();
```
pajlada: Do we need to prune invalid history entries every iteration?

src/widgets/Notebook.cpp
```diff
@@ -374,7 +387,83 @@ void Notebook::select(QWidget *page, bool focusPage)
...
+            break;
+        }
+
+        this->tabHistory_.discardForwardTop();
+    }
+}
+
+bool Notebook::selectHistoryPage(bool forward, bool focusPage)
+{
+    for (size_t attempt = 0; attempt < MAX_TAB_HISTORY_NAV_ATTEMPTS; ++attempt)
```
pajlada: nit: This function should maybe just be duplicated in goBackwards/goForwards functions since it's like 10 lines of code, that way you don't need the ternary forward param

src/widgets/Notebook.cpp
```diff
@@ -374,7 +387,83 @@ void Notebook::select(QWidget *page, bool focusPage)
...
+    {
+        this->pruneInvalidHistoryEntries();
+
+        const bool canNavigate = forward ? this->tabHistory_.canGoForward()
+                                         : this->tabHistory_.canGoBack();
+        if (!canNavigate)
+        {
+            return false;
+        }
+
```
pajlada: Is this necessary if goForward/goBack already returns a nullopt? you could return false if target is a nullopt

src/util/TabHistory.cpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+        return;
+    }
+
+    if (!this->history_.empty())
+    {
+        this->history_.erase(
+            this->history_.begin() +
+                static_cast<std::ptrdiff_t>(this->currentIndex_ + 1),
+            this->history_.end());
```
pajlada, line 23: Could you add a comment for what this does?
If I understand it correctly, it removes any entries beyond your current index, so if you've gone back in history and push a new history entry it will cut off any entries you just navigated through

```
a, b, c
         ^
```
user hasn't navigated, so nothing will be removed because it tries to remove
```
a, b, c, ???
            ^
```

but if user has navigated backwards
```
a, b, c
    ^
```

it will delete c

src/controllers/hotkeys/HotkeyController.cpp
```diff
@@ -552,6 +566,21 @@ void HotkeyController::addDefaults(std::set<QString> &addedHotkeys)
...
+                            QKeySequence("Ctrl+Alt+Right"), "selectTabHistory",
+                            {"next"}, "tab history forward");
+        this->tryAddDefault(addedHotkeys, HotkeyCategory::Window,
+                            QKeySequence(Qt::Key_Back), "selectTabHistory",
+                            {"previous"},
+                            "tab history back (mouse back button)");
+        this->tryAddDefault(addedHotkeys, HotkeyCategory::Window,
+                            QKeySequence(Qt::Key_Forward), "selectTabHistory",
+                            {"next"},
+                            "tab history forward (mouse forward button)");
```
pajlada: these don't work for me on linux, do they work for you on windows or mac?

pajlada (changes requested):
I'm not super certain on default bindings - whether they should be added or not.
Mostly nits, otherwise works well (minus the mouse bindings)

apa420:
I tested this pr and after flashing my QMK keypad with KC_WWW_BACK and KC_WWW_FORWARD I was able to get this working. However using the side mouse buttons mouse 4 and 5 did *not* work.

Chatterino Nightly 2.5.5 (commit [5389a4b9](https://github.com/Chatterino/chatterino2/commit/5389a4b9)) built on 2026-06-27 with Qt 6.8.3, MSVC 195136248, Crashpad
Running on Windows 10 Version 21H2, kernel: 10.0.19044

src/controllers/hotkeys/HotkeyController.cpp
Leafyzito: i didn't test mouse buttons, as im currently on a trackpad only setup. would appreciate some kind of input about this on maybe how to fix, if possible

pajlada: I don't think this is possible unless your mouse sends keyboard events, so I'd just remove those default bindings.
Our Hotkey Manager just doesn't support mouse events right now, so it would need to wait for that - or we hardcode the mouse button handling.

Leafyzito: yeah fair, i'll remove the mouse back/forward hotkey defaults for now. proper mouse button support outside the hotkey manager sounds like the right approach, to be done in a follow-up pr

pajlada (changes requested):
Please remove all default bindings, and if you want to, look into mouse hotkeys in a follow-up PR.
If mouse-hotkeys are implemented, one more follow-up PR can add default mouse buttons for the "popup tab history back" and "popup tab history forward" actions.

Leafyzito:
When I get back to my normal setup (with a mouse) I'll look into adding mouse support to hotkeys.

src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp
```diff
@@ -98,40 +100,52 @@ void QuickSwitcherPopup::updateSuggestions(const QString &text)
...
-    // Add items for navigating to different splits
-    for (auto *sc : openPages(this->window))
+    if (text.isEmpty())
     {
-        const QString &tabTitle = sc->getTab()->getTitle();
-        const auto splits = sc->getSplits();
-
-        // First, check for splits on this page
-        for (auto *split : splits)
+        for (auto *page : this->window->getNotebook().getVisitHistoryPages())
```
Nerixyz, line 105: If you didn't visit any page yet, none are shown here, right?

src/widgets/Notebook.cpp
```diff
@@ -178,26 +178,28 @@ void Notebook::removePage(QWidget *page)
         if (this->items_.count() == 1)
         {
             // Deleting only tab, select nothing
-            this->select(nullptr);
+            this->select(nullptr, true, false);
```
Nerixyz, line 181: It's hard to see from just this line what the booleans mean.

Nerixyz (commented):
There are a few uncovered cases in `TabHistory`: https://app.codecov.io/github/Chatterino/chatterino2/commit/ee5775c8e2050cf0febcb1d8d2e3993e78f0017b/blob/src/util/TabHistory.cpp?dropdown=coverage

src/widgets/dialogs/switcher/QuickSwitcherPopup.cpp
Leafyzito: Yeah, that's right. If you haven't switched tabs yet, nothing is shown.

src/widgets/Notebook.cpp
pajlada: This is fine in this PR but a follow-up PR could make it into a struct with multiple bools instead

pajlada:
> There are a few uncovered cases in `TabHistory`: [app.codecov.io/github/Chatterino/chatterino2/commit/ee5775c8e2050cf0febcb1d8d2e3993e78f0017b/blob/src/util/TabHistory.cpp?dropdown=coverage](https://app.codecov.io/github/Chatterino/chatterino2/commit/ee5775c8e2050cf0febcb1d8d2e3993e78f0017b/blob/src/util/TabHistory.cpp?dropdown=coverage)

If you'd like to try to improve this, feel free in this PR. If you'd rather have this done in a follow-up PR, I'm OK with that too. @Leafyzito

src/controllers/hotkeys/HotkeyController.cpp
```diff
@@ -581,6 +581,15 @@ void HotkeyController::clearRemovedDefaults()
...
                                       "setTabVisibility", {"toggleLiveOnly"});
+
+    this->warnForRemovedHotkeyActions(HotkeyCategory::Window,
+                                      "selectTabHistory", {"previous"});
+    this->warnForRemovedHotkeyActions(HotkeyCategory::Window,
+                                      "selectTabHistory", {"next"});
+    this->warnForRemovedHotkeyActions(HotkeyCategory::PopupWindow,
+                                      "selectTabHistory", {"previous"});
+    this->warnForRemovedHotkeyActions(HotkeyCategory::PopupWindow,
+                                      "selectTabHistory", {"next"});
```
pajlada: These never existed in the main branch, so these are not needed.

src/controllers/hotkeys/ActionNames.hpp
```diff
@@ -84,6 +84,10 @@ inline const std::map<HotkeyCategory, ActionDefinitionMap> actionNames{
               },
               .argumentsPrompt = "Direction:",
           }},
+         {"selectTabHistoryBack",
+          ActionDefinition{"Select previous tab from visit history"}},
+         {"selectTabHistoryForward",
+          ActionDefinition{"Select next tab from visit history"}},
```
pajlada: You've added these as Split hotkeys, but they're not defined under Split.cpp's `addShortcuts` function - are the hotkeys needed at all under the Split scope? I'd imagine no

pajlada (changes requested):

pajlada (approved):

PR #7077, opened 2026-06-29
Feat: search emotes by tags and original name

### LINKED ISSUE #7073: Add the ability to search for 7TV emotes in the Chatterino emote menu by their Tags on 7TV
### Discussed in https://github.com/Chatterino/chatterino2/discussions/7072

<sup>Originally posted by **Lilbint** June 28, 2026</sup>
Hi, just a suggestion I had. Sometimes I forget the name of an emote, but I know it's most likely tagged under a certain tag on the website, so it'd be nice to be able to just search that tag in the emote menu and have it pop up. Granted, this is only useful if the user who uploaded the emote tagged it properly, but just would be an extra handy tool to have when trying to search for a specific emote. </div>

Tags should be searchable with a prefix (probably `tag:yourtaghere`)

As a bonus, if you're modifying these files, you could make the original emote name also be searchable without any prefix (e.g. if I add the emote "forsenE" in my chat aliased as "forsenO", searching both "forsenE" and "forsenO" should show the emote)

4rneee:
Fixes #7073 

Currently emotes are only shown if the specified tag matches exactly (case-insensitive). If multiple tags are specified they are treated as OR.

The new validator for the `search_` widget has one quirk: If you input `tag:a xd` and have your cursor behind the a(`tag:a| xd`), you are not able to remove the a (or anything from `tag:a` unless you also remove the space).


src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -107,6 +107,14 @@ CreateEmoteResult createEmote(const QJsonObject &activeEmote,
             : createTooltip(emoteName.string, author.string, isGlobal);
     auto imageSet = SeventvEmotes::createImageSet(emoteData, false);
 
+    auto tagsArray = emoteData["tags"].toArray();
+    QStringList tags;
+    tags.reserve(tagsArray.size());
+    for (auto &tag : tagsArray.toVariantList())
```
Nerixyz:
```suggestion
    for (const auto tag : tagsArray)
```
`QJsonArray` should yield something like a `QJsonValueRef`.

4rneee:
Diff to test the "contains" instead of exact tag match:
```diff
diff --git a/src/widgets/dialogs/EmotePopup.cpp b/src/widgets/dialogs/EmotePopup.cpp
index d9b701df..c2b15d65 100644
--- a/src/widgets/dialogs/EmotePopup.cpp
+++ b/src/widgets/dialogs/EmotePopup.cpp
@@ -225,12 +225,16 @@ bool emoteMatchesSearchAndTags(const EmotePtr &emote, const QString &text,
                                const QStringList &tags)
 {
     bool tagsMatch = tags.empty();
-    for (const auto &tag : tags)
+    for (auto tag = tags.constBegin(); tag != tags.constEnd() && !tagsMatch;
+         ++tag)
     {
-        if (emote->tags.contains(tag, Qt::CaseInsensitive))
+        for (const auto &emoteTag : emote->tags)
         {
-            tagsMatch = true;
-            break;
+            if (emoteTag.contains(*tag, Qt::CaseInsensitive))
+            {
+                tagsMatch = true;
+                break;
+            }
         }
     }
 
 ```

File: src/messages/Emote.hpp [hunk 2]
```diff
@@ -7,6 +7,8 @@
 #include "common/Aliases.hpp"
 #include "messages/ImageSet.hpp"
 
+#include <qstringlist.h>
```
pajlada:
```suggestion
#include <QStringList>
```

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -26,12 +26,12 @@
 #include "singletons/WindowManager.hpp"
 #include "util/Helpers.hpp"
 #include "widgets/helper/ChannelView.hpp"
-#include "widgets/helper/TrimRegExpValidator.hpp"
 #include "widgets/Notebook.hpp"
 #include "widgets/Scrollbar.hpp"
 
 #include <QAbstractButton>
 #include <QHBoxLayout>
+#include <qregularexpression.h>
```
pajlada: dupe include
```suggestion
```

4rneee:
I decided to make the "contains match" the default for now because otherwise you dont see any emotes until you finished typing the text, which doesn't seem right. You can checkout the previous commit if you want to test the alternative behavior.

An open question is still if we somewhere want to show the tags to the user. The only place I can think of would be the tooltip but that might become very large if an emote has many tags.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -364,7 +419,7 @@ EmotePopup::EmotePopup(QWidget *parent)
     layout2->setSpacing(8);
 
     this->search_->setPlaceholderText("Search all emotes...");
-    this->search_->setValidator(new TrimRegExpValidator(searchRegex));
+    this->search_->setValidator(new QRegularExpressionValidator(searchRegex));
```
Nerixyz, line 419: You need to `#include <QRegularExpressionValidator>` to fix the builds without PCH.

src/widgets/dialogs/EmotePopup.cpp
```diff
@@ -316,14 +315,38 @@ void loadEmojis(Channel &channel, const std::vector<EmojiPtr> &emojiMap,
...
 }
 
-// Create an emote
-EmoteMap filterEmoteMap(const QString &text, const EmoteMap &emotes)
+bool emoteMatchesSearchAndTags(const EmotePtr &emote, const QString &text,
+                               const QStringList &tags)
+{
+    bool tagsMatch = tags.empty();
+    for (auto tag = tags.constBegin(); tag != tags.constEnd() && !tagsMatch;
+         ++tag)
```
```suggestion
    for (const auto &tag : tags)
```
You already break if `tagsMatch` is set. Could think about renaming `tags` to indicate that it's the user's query. That confused me at the start.

4rneee: > You already break if `tagsMatch` is set

Only the inner loop but I want to also stop the outer one. I could also use `goto` if thats preferred.

Nerixyz: Oh, right. You could also add a condition at the end that checks for `tagsMatch`. Another alternative is to use ranges. Like `tagsMatch = std::ranges::any_of(emote->tags, [&]...);`.

src/providers/seventv/SeventvEmotes.cpp
```diff
@@ -107,6 +107,14 @@ CreateEmoteResult createEmote(const QJsonObject &activeEmote,
             : createTooltip(emoteName.string, author.string, isGlobal);
     auto imageSet = SeventvEmotes::createImageSet(emoteData, false);
 
+    auto tagsArray = emoteData["tags"].toArray();
+    QStringList tags;
+    tags.reserve(tagsArray.size());
+    for (const auto tag : tagsArray)
```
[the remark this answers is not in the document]
Nerixyz: Or you could make `tagsArray` `const`.

pajlada (approved):

PR #7079, opened 2026-07-04
refactor: Use magic_enum for encoding channels

Nerixyz:
Split from #6906.

This changes the en-/decoding of channels in the window layout to use magic-enum for translating between enum and string. 
This way, the names are consistent when en-/deconding. And more importantly, when adding channel types, you should get compiler warnings in switch statements for uncovered cases.


pajlada (approved):

Mm2PL (approved):
Only change i noticed is having empty splits (type None) now have their type encoded. LGTM

PR #7080, opened 2026-07-04
refactor: unsingleton Modes

teknsl:

### DIFF
--- a/src/singletons/Settings.cpp
+++ b/src/singletons/Settings.cpp
@@ -6,6 +6,7 @@
 
 #include "Application.hpp"
 #include "common/Args.hpp"
+#include "common/Modes.hpp"
 #include "controllers/filters/FilterRecord.hpp"
 #include "controllers/highlights/HighlightBadge.hpp"
 #include "controllers/highlights/HighlightBlacklistUser.hpp"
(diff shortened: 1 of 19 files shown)


src/singletons/Updates.cpp
```diff
@@ -50,8 +50,9 @@ const QString CHATTERINO_OS = u"unknown"_s;
 
 namespace chatterino {
 
-Updates::Updates(const Paths &paths_, Settings &settings)
+Updates::Updates(const Modes &modes, const Paths &paths_, Settings &settings)
```
pajlada: Rename the param to `modes_` here and in the header

src/singletons/Settings.cpp
```diff
@@ -150,10 +151,14 @@ bool Settings::toggleMutedChannel(const QString &channelName)
...
 
-Settings::Settings(const Args &args, const QString &settingsDirectory,
+Settings::Settings(const Modes &modes, const Args &args,
+                   const QString &settingsDirectory,
                    const SettingsArgs &settingsArgs)
     : prevInstance_(Settings::instance_)
     , disableSaving(args.dontSaveSettings)
+    , createShortcutForToasts(
+          "/notifications/createShortcutForToasts",
+          not(modes.isPortable || modes.isExternallyPackaged))
```
pajlada: keep the old ternary logic instead of ! or not

pajlada (commented):
benchmarks/src/main.cpp needs modes too

pajlada (approved):

PR #7084, opened 2026-07-05
refactor: uncouple Args from Paths

teknsl:
`makeCustomChannelLayout` fits better in `WindowManager.cpp`
(like how `activateChannel` is used) or anywhere other than
`Args.cpp` but i think that can be done in a different PR,
this just makes the existing way work.

`WindowManager` already stores the file path that
`makeCustomChannelLayout` needs so i pass that instead of
`Paths` avoiding the path construction.

`parseCustomChannels` is suitable for testing but would like
some pointers for how to do that properly.

in a future PR if its something wanted, `--activate` could
reuse `parseCustomChannels` to be able to take in multiple
channels.


src/common/Args.cpp
```diff
@@ -195,7 +227,11 @@ Args::Args(const QApplication &app, const Paths &paths)
 
     if (parser.isSet(channelLayout))
     {
-        this->applyCustomChannelLayout(parser.value(channelLayout), paths);
+        this->customChannels = parseCustomChannels(parser.value(channelLayout));
+        if (not this->customChannels.empty())
```
Mm2PL:
```suggestion
        if (!this->customChannels.empty())
```
We don't use alternative operator representation.

src/common/Args.cpp
```diff
@@ -91,11 +91,43 @@ std::optional<Args::Channel> parseActivateOption(QString input)
     };
 }
 
+std::vector<Args::Channel> parseCustomChannels(const QString &string)
```
Nerixyz: Can you refactor this in a followup and keep the old version?

src/common/Args.cpp
```diff
@@ -297,49 +336,21 @@ void Args::applyCustomChannelLayout(const QString &argValue, const Paths &paths)
...
-
-            window.tabs_.emplace_back(std::move(tab));
-        }
+        assert(channel.provider == ProviderId::Twitch);
+
+        TabDescriptor tab = {.selected_ = window.tabs_.empty(),
+                             .rootNode_ = SplitNodeDescriptor{{
+                                 .type_ = "twitch",
+                                 .channelName_ = channel.name,
+                             }}};
```
Nerixyz:
```suggestion
        TabDescriptor tab{.selected_ = window.tabs_.empty(),
                             .rootNode_ = SplitNodeDescriptor{{
                                 .type_ = "twitch",
                                 .channelName_ = channel.name,
                             }},};
```
+reformat; we usually put a comma at the end in multiline braced initializers.

src/common/Args.cpp
teknsl: Sure, were you thinking extracting the parsing code from old `applyCustomChannelLayout` into `parseCustomChannels` and still output a vector or a different way

Nerixyz: Yea, you can't copy it 1:1, but as close as possible to the original code.

pajlada (approved):

PR #7086, opened 2026-07-05
reuse: annotate some more source files

pajlada:


src/controllers/plugins/PluginMeta.cpp
```diff
@@ -1,3 +1,4 @@
+// SPDX-FileCopyrightText: 2023 Contributors to Chatterino <https://chatterino.com>
 // SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
```
Nerixyz: This header is already here. It could be updated to say 2023.

PR #7092, opened 2026-07-09
refactor(window-layout): Build descriptor for serializing

### LINKED ISSUE #7132: Duplicating a tab does not include all characteristics from the splits inside the tab
It does not transfer over the following:

- split moderation mode
- split ratio/sizes
- split spell checking override

Nerixyz:
It's useful for other components to get a split descriptor from a split (e.g. chatterino-embed). To avoid differences between the descriptor and the JSON output (e.g. adding a field to the split data), this makes the descriptor mandatory. Instead of going `SplitContainer* -> JSON`, we do `SplitContainer* -> NodeDescriptor -> JSON`. In a followup, I'm going to add `*Descriptor::toJSON`.


src/widgets/splits/SplitContainer.cpp
```diff
@@ -870,15 +870,11 @@ QString channelTypeToString(Channel::Type value) noexcept
 NodeDescriptor SplitContainer::buildDescriptorRecursively(
     const Node *currentNode) const
 {
-    if (currentNode->children_.empty())
+    if (currentNode->children_.empty() && currentNode->split_)
```
pajlada, line 847: Why do we need to check that `currentNode->split_` is not a null pointer all of a sudden?

src/widgets/splits/SplitContainer.cpp
```diff
@@ -870,15 +870,11 @@ QString channelTypeToString(Channel::Type value) noexcept
...
     {
-        const auto channelType =
-            currentNode->split_->getIndirectChannel().getType();
-
-        SplitNodeDescriptor result;
-        result.type_ = channelTypeToString(channelType);
```
pajlada, line 879: `channelTypeToString` looks to be unused - since that's now being done inside Split's `buildDescriptor` with `magicenum::enumNameString`, it should be removed.

src/widgets/splits/SplitContainer.cpp
```diff
@@ -870,15 +870,11 @@ QString channelTypeToString(Channel::Type value) noexcept
...
-        result.channelName_ = currentNode->split_->getChannel()->getName();
```
pajlada, line 880: It looks like we used to _always_ set `channelName_`, regardless of the channel type.
With this refactor, we only set `channelName_` if it's a `Twitch` or `Misc` type. Is this intended? I imagine yes, and I imagine the other channel types don't have any relevant channel names at all, and I imagine for a third time that whatever channel name we would put there was never encoded into our `window-layout.json` file.

src/singletons/WindowManager.cpp
```diff
@@ -676,92 +679,62 @@ void WindowManager::encodeTab(SplitContainer *tab, bool isSelected,
...
+                       {
+                           data.insert("name"_L1, split.channelName_);
+                       }
+                       obj.insert("data", data);
+
+                       QJsonArray filters;
+                       WindowManager::encodeFilters(split.filters_, filters);
+                       obj.insert("filters", filters);
+
+                       if (split.spellCheckOverride)
```
pajlada: Use the explicit `has_value` function here

src/widgets/splits/SplitContainer.cpp
Nerixyz: We need this for empty tabs - they don't have children and no `split_`. This wasn't a problem for duplicating tabs, because `Notebook` checks for the splits:
https://github.com/Chatterino/chatterino2/blob/544c5903af862c986973623afbfeb86321f2b412/src/widgets/Notebook.cpp#L229-L231

src/widgets/splits/SplitContainer.cpp
Nerixyz: > and I imagine for a third time that whatever channel name we would put there was never encoded into our `window-layout.json` file.

Exactly, this is the important part. Before, the `buildDescriptor` was only used for duplicating tabs. The part in `Split::buildDescriptor` mainly intends to mirror the old `WindowManager::encodeChannel`. There, we only included the channel for specific types.

pajlada (approved):

PR #7093, opened 2026-07-10
feat: command-line arguments for setting portable directory

### LINKED ISSUE #7032: Add command-line arguments to overwrite portable mode and app directory
On Unix, you can use `XDG_DATA_HOME` to change the directory where Chatterino will place its settings. On Windows and macOS, you can't use this environment variable. 

We should still provide a way to overwrite the directory where the settings are loaded and saved from/to. Furthermore, there should be an easy option to start Chatterino in portable mode.

From pajlada in Twitch chat:
> --portable and --portable-dir
> for testing a clean build, just --portable straight inside the build dir would be nice

- Add `--portable` to launch in portable mode. This will take precedence over the `modes` file.
- Add `--portable-dir <dir>` to set the configuration directory. This takes precedence over the default (portable) directory.

**Tips**

- See [`Paths::initRootDirectory`](https://github.com/Chatterino/chatterino2/blob/e84e532fe1d4416479166b70140daee96087df03/src/singletons/Paths.cpp#L89), [`Args`](https://github.com/Chatterino/chatterino2/blob/e84e532fe1d4416479166b70140daee96087df03/src/common/Args.cpp), [`Modes`](https://github.com/Chatterino/chatterino2/blob/e84e532fe1d4416479166b70140daee96087df03/src/common/Modes.cpp), and [`main()`](https://github.com/Chatterino/chatterino2/blob/e84e532fe1d4416479166b70140daee96087df03/src/main.cpp#L53).
- `Args` currently depends on `Paths`. This should be flipped. It only depends on `Paths` for `Args::applyCustomChannelLayout`. The channel layout string can be saved and applied later (when `Paths` is constructed).

teknsl:
* Closes #7032
* Based on #7084 

Adds `--portable` and `--portable-dir` command line options for using a custom data directory


src/common/Modes.cpp
```diff
@@ -4,15 +4,29 @@
...
 
-Modes::Modes()
+Modes::Modes(const Args &args)
 {
-    QFile file(combinePath(QCoreApplication::applicationDirPath(), "modes"));
+    QString parentDirectory;
+
+    if (args.portableDirectory.has_value())
+    {
+        this->isPortable = true;
```
Nerixyz, line 18: Is it worth having a `-portable` line to disable portable mode?

src/singletons/Paths.hpp
```diff
@@ -64,12 +63,9 @@ class Paths
 
 private:
     void initAppFilePathHash();
-    void initCheckPortable();
-    void initRootDirectory(const Modes &modes);
+    void initRootDirectory(const Args &args, const Modes &modes);
     void initSubDirectories();
 
-    std::optional<bool> portable_;
```
Nerixyz, line 71: This wasn't used, right?

src/common/Args.hpp
```diff
@@ -44,7 +44,7 @@ class Args
     };
 
     Args() = default;
-    Args(const QApplication &app, const Paths &paths);
+    Args(const QApplication &app);
```
Nerixyz:
```suggestion
    explicit Args(const QApplication &app);
```

src/common/Modes.hpp
```diff
@@ -6,11 +6,14 @@
 
 namespace chatterino {
 
+class Args;
+
 class Modes
 {
 public:
-    Modes();
+    Modes(const Args &args);
```
Nerixyz:
```suggestion
    explicit Modes(const Args &args);
```

src/main.cpp
```diff
@@ -50,15 +50,16 @@ int main(int argc, char **argv)
         Version::instance().appUserModelID().c_str());
 #endif
 
-    const Modes modes;
+    const Args args(a);
+    const Modes modes(a);
```
Nerixyz:
```suggestion
    const Modes modes(args);
```

You should also see an error here if you make the `Args` constructor `explicit`.

src/common/Modes.cpp
pajlada: I don't think that's necessary - if we see a use for it we can refine this logic then.

src/common/Args.cpp
```diff
@@ -151,6 +183,15 @@ Args::Args(const QApplication &app, const Paths &paths)
         "like you have to use this, please reach out to our issue tracker at "
         "https://github.com/Chatterino/chatterino2/issues");
 
+    QCommandLineOption portable(
+        "portable",
+        "Starts Chatterino in portable mode in the current directory");
+
+    QCommandLineOption portableDir(
+        "portable-dir",
```
pajlada: I saw `--portable-dir` more as a "if portable mode is enabled, use the specified directory as the portable directory"

So if a user doesn't want to use the `modes` file but enable portable mode, they would run: `chatterino --portable`
If a user wants to respect whatever portable mode is in the `modes` file and change the directory _if_ portable mode is enabled, they would run: `chatterino --portable-dir /tmp/xd`
If a user wants to enable portable mode in a custom dir, they would run: `chatterino --portable --portable-dir /tmp/xd`

teknsl: ~do you mean "doesnt want to use `modes` file" as in doesnt _have_ to use the
modes file or dont read the modes file at all if `--portable` is passed? with
earlier form if the executable is for example located in /usr/local/bin it
would be a bit silly attempting reading a `modes` file there~

i missed the fact that the current behaviour is to actually read the modes file even if the executable is located under `/usr/local/bin`. i do not particularly like this behaviour but it is not relevant to this PR.

so to clarify the behaviour a little, is this what you want?
with `--portable` the application executes as if `$BINDIR/modes` had the portable line but `$BINDIR/modes` will still be attempted read (iirc you said you might want to do more with the `modes` file in the future). then `--portable-dir` only sets the portable directory if portable mode has been enabled through either of those.

src/singletons/Paths.hpp
teknsl: nope, dead code

src/common/Args.cpp
pajlada: Correct.

pajlada (approved):

Issue #7021, opened 2026-05-28
Date and time API for Plugins


Nerixyz:
There has never been a dedicated issue for this - only discusson outside of this repository.

Plugins need a date and time API. Lua technically has `os.clock`, `os.date`, `os.time` and `os.difftime`, but for one, we don't include the `os` library and more importantly, these use ISO C functions and don't support zones.

Qt has [`QDateTime`](https://doc.qt.io/qt-6/qdatetime.html) which is what we use in most places. It has the desired features.
There was an attempt to directly expose this to Lua (https://github.com/Chatterino/chatterino2/tree/feature/plugin-api-datetime / https://github.com/Chatterino/chatterino2/commit/a911163f525e4cf73a93ff6d9726c04282d74392).
The comment there was that it's too tightly coupled to Qt and that we should make our own API surface.
In the end, the usertype can (and probably will) still store a `QDateTime`(/`QDate`,`QTime`).

I think we can start with a bare-bones API that provides the basics.
This should address three main use-cases:
1. Parsing and stringifying timestamps.
   HTTP APIs often encode timestamps as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) strings.
2. Getting the current date and time.
3. Handling Unix timestamps (offsets from 1970-01-01 00:00 UTC).

Here's a proposed initial API surface:

```lua
---A zoned date and time.
---@class c2.DateTime
c2.DateTime = {}

---Parse a date from an ISO 8601 string with milliseconds (yyyy-MM-ddTHH:mm:ss.zzz)
---@param str string
---@return c2.DateTime
function c2.DateTime.from_iso_string(str) end

---Parse a date from an ISO 8601 string without milliseconds (yyyy-MM-ddTHH:mm:ss)
---@param str string
---@return c2.DateTime
function c2.DateTime.from_iso_string_without_ms(str) end

---Format the datetime as an ISO string with milliseconds (yyyy-MM-ddTHH:mm:ss.zzz)
---@return string
function c2.DateTime:to_iso_string() end

---Format the datetime as an ISO string without milliseconds (yyyy-MM-ddTHH:mm:ss)
---@return string
function c2.DateTime:to_iso_string_without_ms() end

---Get the current datetime in the system's local zone.
---@return c2.DateTime
function c2.DateTime.current_local() end

---Get the current datetime in the UTC zone (00:00).
---@return c2.DateTime
function c2.DateTime.current_utc() end

[...]

PR #7096, opened 2026-07-11
feat(plugins): Add `c2.DateTime`

Nerixyz:
This adds the `DateTime` API described in #7021.

Internally, `DateTime` is a `QDateTime`, but it doesn't expose the functions 1:1. I only added the functions described in the issue. While testing, I noticed that we're missing a way to convert between local and UTC time zones. I'll add this in a followup.


pajlada (approved):

docs/wip-plugins.md
```diff
@@ -981,6 +981,50 @@ Get all open windows.
...
+
+Parse a date from an ISO 8601 string with milliseconds (yyyy-MM-ddTHH:mm:ss.zzz)
+
+##### `DateTime:to_iso_string()`
+
+Format the datetime as an ISO string with milliseconds (yyyy-MM-ddTHH:mm:ss.zzz)
+
+##### `DateTime:to_iso_string_without_ms()`
+
+Format the datetime as an ISO string without milliseconds (yyyy-MM-ddTHH:mm:ss)
```
Mm2PL: I think it'd be nice to put these formats as code blocks

```suggestion
##### `DateTime.from_iso_string(str)`

Parse a date from an ISO 8601 string with milliseconds (`yyyy-MM-ddTHH:mm:ss.zzz`)

##### `DateTime:to_iso_string()`

Format the datetime as an ISO string with milliseconds (`yyyy-MM-ddTHH:mm:ss.zzz`)

##### `DateTime:to_iso_string_without_ms()`

Format the datetime as an ISO string without milliseconds (`yyyy-MM-ddTHH:mm:ss`)
```

Consider also putting the timezone identifier `±hh:mm` into these

Mm2PL (changes requested):

Mm2PL:
I may have failed at reading the patch description oops. Only thing left in the review is the documentation styling

Mm2PL (approved):

PR #7016, opened 2026-05-25
feat(plugins): Add remote repositories

Nerixyz:
Adds the ability to install plugins from within Chatterino from remote repositories. Users can add repositories on their own. Chatterino comes with one default repository.

This is intended to provide the bare minimum to install plugins and even for that it's relatively large. I'll try to extract some parts into separate PRs.

One thing I'm not too sure about is the terminology. While developing, I changed it at some point, so it might not be consistent right now. Here's the current state:

- (Plugin) **Repository**: A service/host for plugins. It is identified by a base URL and must provide a `meta.json` file that contains all available plugins. In most cases, this should correspond to a Git repo, but it might also be a standalone service (e.g. GH pages).
- **Remote Plugin**: Description of a plugin's metadata, ID, and download URL.

The metadata provided by the repository is intentionally kept simple to allow users to host it on services like GitHub or GH pages. A metadata file looks like this:

```json
{
  "metadata": {
    "name": "My Plugin Store"
  },
  "plugins": [
    {
      "id": "my-plugin",
      "download": "plugins/my-plugin",
      "meta": {
        "name": "My Plugin",
        "description": "This is a description of my plugin.",
        "authors": [],
        "version": "1.0.0",
        "license": "MIT",
        "files": ["init.lua"]
      }
    }
  ]
}
```

A schema file for this is also present in `docs/plugin-repo-meta.schema.json`. The repository structure might look like this:
```
.
│   meta.json
│   
└───plugins
    ├───my-plugin
    │       init.lua
    │       
    └───nerixyz.plugin
            info.json
            init.lua
```

Plugins are not single files (e.g. zips). They're a collection of files plus some metadata. The plugin metadata is contained in the repository metadata. It's not a separate file, because we need the metadata already when enumerating the plugins in the repository. If it was another file, we'd need to check that the metadata there is consistent with the one from the repository. When a plugin includes an "info.json" file in its `"files"`, we skip it. We always use the one from the repo.

The following is a small script to host a local plugin server. It expects the plugins in `src/`. Run it with `python bundle.py --serve`.

<details><summary>Script</summary>

```py
# Generate and run file server: python bundle.py --serve
import os
import shutil
import argparse
from dataclasses import dataclass
from pathlib import Path
import json
from typing import Any
import re


@dataclass
class Config:
    out_dir: Path
    src_dir: Path
    serve: bool
    port: int
    name: str


PATH_RE = re.compile(r"^[\w_\-+\./]+$")


def validate_path(path: str):
[...]

PR #7097, opened 2026-07-11
feat(plugins): Add remote plugins without UI

Nerixyz:
This is the "backend" part of #7016. It implements all the functionality for remote plugins without any UI. You interact with it through a temporary command.

```sh
# List all remote plugins
/debug-plugin list
# Install plugin by ID
/debug-plugin install ID
# Update plugin by ID
/debug-plugin update ID
# Remove plugin by ID
/debug-plugin remove ID [--erase-data]
```

I also added tests for "downloading" and removing.

The schema for the remote metadata is the same as in #7016. You can also use the same test server script as in that PR.
Quick summary:

- We have remote "repositories" which store the plugin files.
- Each repository holds a `meta.json` that lists all available plugins and their files.
- Local plugins store their remote URL. This way we can detect "related" plugins.

### DIFF
--- a/src/controllers/plugins/PluginController.cpp
+++ b/src/controllers/plugins/PluginController.cpp
@@ -32,6 +32,7 @@
 #    include "singletons/Settings.hpp"
 #    include "singletons/WindowManager.hpp"
 #    include "util/FilesystemHelpers.hpp"
+#    include "util/PostToThread.hpp"
 #    include "widgets/splits/SplitContainer.hpp"
 #    include "widgets/Window.hpp"
 
@@ -556,8 +557,9 @@ void PluginController::download(const DownloadArgs &args)
         .onDone = args.onDone,
     });
 
-    auto pushResult = [this, state](ExpectedStr<void> res) {
-        if (isAppAboutToQuit() || !tryGetApp())
+    auto pushResult = [this, lifetime = std::weak_ptr{this->lifetime},
+                       state](ExpectedStr<void> res) {
+        if (lifetime.expired())
         {
             return;
         }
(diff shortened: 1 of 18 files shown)


src/controllers/plugins/PluginController.cpp
```diff
@@ -367,6 +371,290 @@ bool PluginController::reload(const QString &id)
...
+
+    auto fetchFile = args.fetchFile;
+    if (!fetchFile)
+    {
+        fetchFile =
+            +[](const QUrl &url,
+                const std::function<void(ExpectedStr<QByteArray>)> &cb) {
+                NetworkRequest(url)
+                    .timeout(30'000)
+                    .followRedirects(true)
```
pajlada: Is this necessary?

src/controllers/plugins/PluginController.cpp
```diff
@@ -367,6 +371,290 @@ bool PluginController::reload(const QString &id)
...
+                f.write(*res);
+            }
+            pushResult({});
+        });
+    }
+}
+
+ExpectedStr<void> PluginController::finishDownload(
+    const RemotePlugin &remote, const std::filesystem::path &pluginDir)
+{
```
pajlada, line 646: Would love room for checksumming all the downloaded files here or somewhere nearby - it's fine if that's a stub for now

pajlada (commented):
Haven't looked at everything - overall I think this looks like a good approach. Thanks for working on this!

src/controllers/plugins/PluginController.cpp
Nerixyz: Not really, I removed it. If someone needs this in the future, we can add it back.

src/controllers/plugins/PluginController.cpp
Nerixyz: Adding checksums can be as simple as making `files` accept and array of either strings or objects (path+hash). Before we save the file, we check the hash. 

I don't think we need it in the initial version, though.

PR #7099, opened 2026-07-12
feat: add setting to show shared chat badge for all messages

### LINKED ISSUE #6244: Missing shared chat badge for host channel
### Checklist

### Describe your issue

Make it possible to align messages during shared chat sessions

### Screenshots

Image

### OS and Chatterino Version

Chatterino 2.5.3 (commit 31d1aaeb9) built with Qt 6.9.0 Running on Arch Linux, kernel: 6.14.6-arch1-1

4rneee:
The current default behavior is to only show the badge for messages from other participants. This PR introduces a setting to also show the badge for messages from the selected channel. 
In my opinion this helps to read shared chats because I sometimes mistake other badges as shared chat badges.


4rneee:
I just thought about this and think we could even make this the default (with or without the setting). I think the current behavior mirrored twitch's web chat but I noticed that now there are no badges in the web chat and you only see where the message was sent from when you hover over the username.
I'd argue that having a badge in front of every message improves the shared chat experience a lot.

src/singletons/Settings.hpp
```diff
@@ -414,6 +414,11 @@ class Settings
     IntSetting sharedChatSessionRefreshInterval = {
         "/behaviour/sharedChatSessionRefreshInterval", 60};
 
+    BoolSetting sharedChatAlwaysShowBadge = {
+        "/behaviour/sharedChatAlwaysShowBadge",
+        false,
```
pajlada: I think we should leave the setting but have the setting true by default, mirroring Twitch's implementation by default

pajlada (approved):

PR #7103, opened 2026-07-14
feat: add customizable recent-messages fallback list

3r01:
Recent message history is currently dependent upon a single instance of [robotty/recent-messages2](https://github.com/robotty/recent-messages2) hosted by @RAnders00 at `https://recent-messages.robotty.de/`. This instance does not have perfect uptime and users are not easily able to change to another instance if it fails (currently they can only switch using an env variable).

This PR addresses that problem by introducing several fallback instances in a configurable list, and a mechanism for querying them in order should any fail. When a failure happens, the fallback tries are shown in the chat, so that the user can see/audit their list if they wish.

image

---

The built-in provider list is as follows:

- https://recent-messages.robotty.de/api/v2/recent-messages/%1 - the existing instance by @RAnders00 
- https://recentmessages.ivr.fi/api/v2/recent-messages/%1 - by @Leppunen, for which I have obtained explicit permission.
- https://rm.iore.tv/api/%1 - by myself, given with my permission.
- https://recent-messages.zneix.eu/api/v2/recent-messages/%1 - by @zneix
- https://rm.lilb.dev/api/v2/recent-messages/%1 - by twitch user lilb_lxryer, for which I have obtained explicit permission.
- https://logs.zonian.dev/rm/%1 - by @ZonianMidian 

I have reached out to @zneix and @ZonianMidian to ask for their explicit permission to include their instances in the built-in list, but have not received a response at the time of writing (perhaps they could comment here). [The documentation](https://recent-messages.zneix.eu/api) states `You are free to use this API from client-side web apps and applications`, but I can remove those without explicit permission if that decision is reached in this PR.

---

Users can customize the list in the advanced settings UI. They are able to enable/disable individual providers, reorder them, and add/remove their own providers. Users can reorder the built-in providers but cannot edit or remove them.

image

In case we need to update the list of built-in providers in future, each built-in provider has a stable identity so that chatterino can reconcile updates with the user's configured list. It does this according to the following policy: if the built-ins remain in the default relative order within the user's settings, then any new entries are added in their default relative position. However, if the user has re-ordered the built-in providers, then any new entries are simply appended, to avoid unexpected changes to the user's custom order. Built-in providers that are removed from the list of defaults will also be removed from the user's settings upon load, using their stable id.

---

Backwards compatibility with `CHATTERINO2_RECENT_MESSAGES_URL` is supported - if the env variable is set, then it overrides the fallback list to act as sole provider, and displays a notice in the advanced settings UI.

image

---

I have added tests, with the exception that no network tests are included, so the fallback behaviour must be tested manually. To do that, you can insert an invalid provider at the top of the list, and a working one below, then connect to a channel and observe the fallback results. See the first screenshot in this PR for an example using `https://example.invalid.host/recent-messages/%1`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -430,6 +430,10 @@ set(SOURCE_FILES
         providers/recentmessages/Api.hpp
         providers/recentmessages/Impl.cpp
         providers/recentmessages/Impl.hpp
+        providers/recentmessages/Provider.cpp
+        providers/recentmessages/Provider.hpp
+        providers/recentmessages/ProviderModel.cpp
+        providers/recentmessages/ProviderModel.hpp
 
         providers/seventv/SeventvAPI.cpp
         providers/seventv/SeventvAPI.hpp
--- a/src/common/Env.cpp
+++ b/src/common/Env.cpp
@@ -74,9 +74,7 @@ bool readBoolEnv(const char *envName, bool defaultValue)
 
 Env::Env()
     : recentMessagesApiUrl(
-          qEnvironmentVariable("CHATTERINO2_RECENT_MESSAGES_URL",
-                               "https://recent-messages.robotty.de/api/v2/"
-                               "recent-messages/%1"))
+          qEnvironmentVariable("CHATTERINO2_RECENT_MESSAGES_URL"))
     , linkResolverUrl(qEnvironmentVariable(
           "CHATTERINO2_LINK_RESOLVER_URL",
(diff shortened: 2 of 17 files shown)


pajlada:
I'm not interested in reviewing / adding this change as is. If you're interested in making big changes to a system like this, please open a discussion first or talk to a contributor first.

Nerixyz:
I'm not opposed to this in general. However, I think the UI for this should be less "visible". Maybe in a dialog. Specifically because we shouldn't need the fallbacks in most cases.

PR #7104, opened 2026-07-14
feat: custom recent-messages URL in settings UI

3r01:
Adds a checkbox and text field for a custom message history URL (instance of [robotty/recent-messages2](https://github.com/robotty/recent-messages2))

Keeps backwards compatibility with the env variable `CHATTERINO2_RECENT_MESSAGES_URL`.  The priority order is:

1. `CHATTERINO2_RECENT_MESSAGES_URL` if populated.
2. The custom URL in settings, if populated and checkbox checked.
3. Chatterino default

<details>
<summary>screenshot</summary>
image</details>
<details>
<summary>screenshot with populated example</summary>
image</details>

I have manually tested this.

### DIFF
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -738,6 +738,9 @@ public:
 
     BoolSetting loadTwitchMessageHistoryOnConnect = {
         "/misc/twitch/loadMessageHistoryOnConnect", true};
+    BoolSetting useCustomMessageHistoryUrl = {
+        "/misc/twitch/useCustomMessageHistoryUrl", false};
+    QStringSetting messageHistoryUrl = {"/misc/twitch/messageHistoryUrl", ""};
     IntSetting twitchMessageHistoryLimit = {
         "/misc/twitch/messageHistoryLimit",
         800,
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -22,6 +22,7 @@
 #include "controllers/nicknames/Nickname.hpp"
 #include "controllers/sound/ISoundController.hpp"
 #include "providers/emoji/EmojiStyle.hpp"
+#include "providers/recentmessages/Api.hpp"
 #include "singletons/NativeMessaging.hpp"
 #include "singletons/Toasts.hpp"
 #include "util/RapidJsonSerializeQString.hpp"  // IWYU pragma: keep
@@ -738,7 +739,10 @@ public:
 
     BoolSetting loadTwitchMessageHistoryOnConnect = {
         "/misc/twitch/loadMessageHistoryOnConnect", true};
-    QStringSetting messageHistoryUrl = {"/misc/twitch/messageHistoryUrl", ""};
+    QStringSetting messageHistoryUrl = {
+        "/misc/twitch/messageHistoryUrl",
+        recentmessages::DEFAULT_API_URL.toString(),
+    };
     IntSetting twitchMessageHistoryLimit = {
         "/misc/twitch/messageHistoryLimit",
         800,
--- a/src/singletons/Settings.hpp
+++ b/src/singletons/Settings.hpp
@@ -22,6 +22,7 @@
 #include "controllers/nicknames/Nickname.hpp"
 #include "controllers/sound/ISoundController.hpp"
 #include "providers/emoji/EmojiStyle.hpp"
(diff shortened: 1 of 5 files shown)


3r01:
alternative to #7103 rejected as overly complex

Nerixyz:
Haven't tested this, but looks good to me overall. Now that I think about https://github.com/Chatterino/chatterino2/pull/7103: Maybe this text field would be enough. We could allow multiple entries there in a semicolon separated list.

3r01:
> Haven't tested this, but looks good to me overall. Now that I think about #7103: Maybe this text field would be enough. We could allow multiple entries there in a semicolon separated list.

I would be happy to do it that way if fallbacks are acceptable; pajlada said that he preferred not to have them. It would solve the single-point-of-failure problem, where this PR alone will not.

src/singletons/Settings.hpp
```diff
@@ -738,6 +738,9 @@ class Settings
 
     BoolSetting loadTwitchMessageHistoryOnConnect = {
         "/misc/twitch/loadMessageHistoryOnConnect", true};
+    BoolSetting useCustomRecentMessagesUrl = {
+        "/misc/twitch/useCustomRecentMessagesUrl", false};
+    QStringSetting recentMessagesUrl = {"/misc/twitch/recentMessagesUrl", ""};
```
pajlada: Is there some practical reason for why you'd want a checkbox + custom URL instead of a URL with a default value?

3r01: If, for example, either the default or your custom instance is down, you may want to quickly switch between your custom URL and the default, or vice-versa. A checkbox allows you to do so without having to re-enter a long string of text.

pajlada:
I don't think fallbacks for individual backends makes sense, nor is it worth the added complexity.

A single customizable URL is OK, but imo it devalues the effort put into maintaining the recent messages service and complicates support efforts from our end.
Rather than us resolving issues with one provider, we're now incentiviced to resolve issues with any provider a user puts in. When configurable with environment variables, there's a barrier to entry meaning it's mostly users who know what they're doing who would configure this.

If you want better uptime - support the provider of the service.
If you have concerns other than uptime, bring those up.

src/singletons/Settings.hpp
pajlada: Switch to using a single QStringSetting with a default value instead of a checkbox + QStringSetting.
The setting path should be `/misc/twitch/messageHistoryUrl`

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -1551,6 +1552,18 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+                            s.useCustomRecentMessagesUrl)
+        ->addTo(layout);
+
+    SettingWidget::lineEdit("Custom message history URL", s.recentMessagesUrl,
+                            recentmessages::DEFAULT_API_URL.toString())
+        ->setTooltip(
+            "Use %1 where the channel name should be inserted, for example: " +
+            recentmessages::DEFAULT_API_URL.toString())
+        ->conditionallyEnabledBy(s.useCustomRecentMessagesUrl)
+        ->addTo(layout);
```
pajlada: When the checkbox is removed, the text should either be "Message history URL", "Message history provider", or maybe just "URL" / "Provider" (not sure how that would look).

pajlada (changes requested):

src/widgets/settingspages/GeneralPage.cpp
3r01: How is this?

image

3r01:
I have made the requested changes. Apologies for the messy branch history.

image

src/singletons/Settings.hpp
```diff
@@ -764,6 +739,10 @@ class Settings
 
     BoolSetting loadTwitchMessageHistoryOnConnect = {
         "/misc/twitch/loadMessageHistoryOnConnect", true};
+    QStringSetting messageHistoryUrl = {
+        "/misc/twitch/messageHistoryUrl",
+        recentmessages::DEFAULT_API_URL.toString(),
```
pajlada: Leave this default empty, that way it will only show up as the placeholder in the LineEdit

pajlada (approved):

PR #7106, opened 2026-07-15
feat: add global filters that apply to all channels/splits

### LINKED ISSUE #7044: Allow turning on showing joins and parts granularly as well, instead of only globally
### Checklist

### Describe your issue

Right now Chatterino only has global settings checkboxes for "Show joined users" and "Show parted users". It is obvious why turning this on for all tabs/splits/channels is detrimental and very undesirable (unless you use a Chatterino instance with only a single channel...). In fact, it seems it would have made more sense than the present  situation if turning it globally was impossible and only granular control with no wildcard was offered. Of course, that isn't a reason to deprive of the user the control to choose either option (especially considering some existing users could be using the only presently available option).

The workaround that is currently possible of turning this on globally and then manually adding a filter that hides these messages in every single split ever created isn't tenable.
If a filter could do the reverse - selectively show the messages - or if (and incidentally, here are some good new ideas!) there were options to designate a filter as "globally active" or as "automatically added to all new splits" (might as well then also include a one-time button to add a filter to all existing splits, or a customizable mass-add option), then using this workaround would at least be sensible... as a workaround. It would still be backwards, though. Chatterino already, thankfully, includes a few per-channel and per-split preferences where it makes sense to have them, rather than only letting the user turn them on globally and then struggle to work around them in every case they're unwanted, so this wouldn't be the first one.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino Nightly 2.5.5 (commit [00ffa0e4](https://github.com/Chatterino/chatterino2/commit/00ffa0e4)) built on 2026-05-28 with Qt 6.7.1, MSVC 195136244, Crashpad
Running on Windows 10 Version 22H2, kernel: 10.0.19045

LeastSussyImposter:
These changes aim to bring [global filters](https://github.com/Chatterino/chatterino2/issues/7044#issuecomment-4881587337) that would apply to all channels and splits.
There are numerous use cases this could apply to for users. I'm sure people will get creative.

I added a `Global` property to filters, checking this property automatically applies the filter to all channels and splits.

<details>
<summary>Example: Only show users joined/parted messages on a specific channel.</summary>
Fixes #7044.</br></br>

Adding the following as a global filter: `!(flags.system_message && message.content startswith "Users " && channel.name != "MyChannel")`, user join/part messages will only show on `MyChannel`.

Here is what it looks like in the filter settings:
image
</details>

I have tested my changes on Arch Linux and Windows, I do not own a Mac to test my changes there.

Let me know if this format works for you.
I'm also willing to modify the filter documentation page should this PR be approved.

Thank you!

### DIFF
--- a/benchmarks/src/Filters.cpp
+++ b/benchmarks/src/Filters.cpp
@@ -40,7 +40,7 @@ public:
             // ensure deterministic order
             auto id = QUuid(static_cast<uint>(i), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
             auto filter = std::make_shared<FilterRecord>(
-                QString::number(i), this->filterTexts.at(i), id);
+                QString::number(i), this->filterTexts.at(i), false, id);
             if (!filter->valid())
             {
                 qCDebug(chatterinoApp) << i << this->filterTexts[i];
--- a/src/controllers/filters/FilterModel.cpp
+++ b/src/controllers/filters/FilterModel.cpp
@@ -13,7 +13,7 @@ namespace chatterino {
 
 // commandmodel
 FilterModel::FilterModel(QObject *parent)
-    : SignalVectorModel<FilterRecordPtr>(3, parent)
+    : SignalVectorModel<FilterRecordPtr>(4, parent)
 {
 }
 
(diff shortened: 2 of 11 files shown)


Nerixyz:
> There are numerous use cases this could apply to for users.

Could you list some besides hiding joined/parted users in specific channels?

LeastSussyImposter:
> > There are numerous use cases this could apply to for users.
> 
> Could you list some besides hiding joined/parted users in specific channels?

Blocking certain words/emotes people don't want to see, block all kind of Twitch messages like resubs, watch streaks, chat bot commands sent from other users like if the message start with '!' or other prefixes. Just from the top of my head 😄

Nerixyz:
> Blocking certain words/emotes people don't want to see,

This is something you can do with ignores today.

> block all kind of Twitch messages like resubs, watch streaks,

This is something you can't do with ignores.

---

Are we okay with the overlap? With this PR, you could define a filter for a message and an ignore phrase. One important difference is that filters are only visual. They don't actually "remove" the messages, they only hide it. This makes a difference with highlights. If you ignore+block a message, it won't cause highlights to trigger. If you filter out a message, highlights will still trigger.

pajlada:
Another option would be for ignores to have a "filter type", similar to how I'm adding a filter type to highlights. I'm OK with looking at this PR for now

PR #7109, opened 2026-07-16
fix: make pinned message text selectable

4rneee:
Previously, the text could not be interacted with.


Nerixyz:
This is something I didn't think about when recommending `MessageView` for pinned messages. Right now, `MessageView` doesn't support selection or click events in general. I'm not sure what the best solution would be. Of course, we could copy the selection code from `ChannelView`. But that has the downside of two implementations drifting apart.

pajlada:
What interactions are we looking to support from the pinned message?

I think we want links to be clickable.
I think we want emotes to show up.
I think we want to allow copying of the message. This would not need full text selection - this could be a right-click "copy message" option.
We could allow full text selection, which some users might prefer, but that feels complicated.

pajlada (approved):
This is fine for now

PR #7111, opened 2026-07-17
Respect preferred username display mode for participants in shared chat

invisibl3cat:
As a follow-up to the shared chat PR, this fixes up the issue where the names of the participants did not respect the preferred username display format.

I'm not sure how I feel about using the HelixMinimalUser directly like that. Is it worth factoring it out?

### DIFF
--- a/src/widgets/splits/SplitHeader.cpp
+++ b/src/widgets/splits/SplitHeader.cpp
@@ -208,8 +208,7 @@ auto formatTitle(const TwitchChannel::StreamStatus &s, Settings &settings,
         }
         else
         {
-            const auto mode = static_cast<UsernameDisplayMode>(
-                getSettings()->usernameDisplayMode.getValue());
+            const auto mode = getSettings()->usernameDisplayMode.getEnum();
             QStringList names;
             for (const auto &p : sharedChatParticipants)
             {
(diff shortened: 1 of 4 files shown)


src/widgets/splits/SplitHeader.cpp
```diff
@@ -201,13 +202,22 @@ auto formatTitle(const TwitchChannel::StreamStatus &s, Settings &settings,
...
-        if (sharedChatParticipants.isEmpty())
+        if (sharedChatParticipants.empty())
         {
             title += " (live)";
         }
         else
         {
-            title += " (live with " + sharedChatParticipants.join(", ") + ")";
+            const auto mode = static_cast<UsernameDisplayMode>(
+                getSettings()->usernameDisplayMode.getValue());
```
pajlada: can use `getEnum` directly
```suggestion
            const auto mode = getSettings()->usernameDisplayMode.getEnum();
```

pajlada (changes requested):
No need for a further refactor, this looks good at a glance to me other than the little `getEnum` usage (i know the pinned message didn't use it, missed it in that pr)

pajlada (approved):

PR #7113, opened 2026-07-18
chore: Scale down pinned message banner

Nerixyz:
I think the text in the pinned message banner is too large in relation to the default UI font width, so I scaled it down.

| Before| After |
|--------|--------|
|  |   |


pajlada (approved):
yep!

PR #7114, opened 2026-07-18
fix: Update pined message height if viewport resized

Nerixyz:
If the split is not 100dp wide, the calculated message height can be wrong. If the split is larger than 100dp and the message would wrap at that size, the widget would be too tall. And vice versa for splits smaller than 100dp.

We do calculate the height in our resize event, but we use the size of our children. Their resize events are delivered after we're resized. So we check if we need to resize after we show the widget. This does result in a small flash where the content is sized incorrectly, but better than it being incorrect the entire time.


pajlada (approved):
neat

PR #7122, opened 2026-07-20
feat(plugins/datetime): Add local/UTC conversion

Nerixyz:
I mentioned in #7096 that it's helpful to convert between UTC and local timezones. This PR adds methods for this.

### DIFF
--- a/docs/wip-plugins.md
+++ b/docs/wip-plugins.md
@@ -1025,6 +1025,22 @@ Convert a datetime to a Unix timestamp (offset from 1970-01-01 00:00 UTC) in mil
 
 Convert a datetime to a Unix timestamp (offset from 1970-01-01 00:00 UTC) in seconds.
 
+##### `DateTime:is_local()`
+
+Check if the datetime is in the user's local/system timezone.
+
+##### `DateTime:is_utc()`
+
+Check if the datetime is in UTC.
+
+##### `DateTime:to_local()`
+
+Returns a copy of this datetime converted to the user's local/system timezone.
+
+##### `DateTime:to_utc()`
+
+Returns a copy of this datetime converted to UTC.
+
 ### Input/Output API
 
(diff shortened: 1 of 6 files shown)


pajlada (approved):

docs/wip-plugins.md
```diff
@@ -1025,6 +1025,22 @@ Convert a datetime to a Unix timestamp (offset from 1970-01-01 00:00 UTC) in mil
 
 Convert a datetime to a Unix timestamp (offset from 1970-01-01 00:00 UTC) in seconds.
 
+##### `DateTime:is_local()`
+
+Check if the datetime is in the user's local/system timezone.
```
Mm2PL: A local date time is one *without* a timezone. So if I live in UTC:
`1970-01-01T00:00:00` will be a local datetime and `1970-01-01T00:00:00Z` will be the same point in time represented with a timezone. I want this to be more clear in the documentation.

Maybe something like:

```suggestion
Check if the datetime is a local time. Local times are represented without a timezone. This is distinct from a date time with your system timezone.
```

docs/wip-plugins.md
```diff
@@ -1025,6 +1025,22 @@ Convert a datetime to a Unix timestamp (offset from 1970-01-01 00:00 UTC) in mil
...
+
+##### `DateTime:is_utc()`
+
+Check if the datetime is in UTC.
+
+##### `DateTime:to_local()`
+
+Returns a copy of this datetime converted to the user's local/system timezone.
```
Mm2PL: This also needs to be changed.

```suggestion
Returns a copy of this datetime converted to the user's local time. Local time is represented without a timezone. That is `1970-01-01T00:00:00` is a local time but `1970-01-01T00:00:00Z` is not.
```

Mm2PL (changes requested):
Documentation for local times is slightly inaccurate.

Mm2PL (approved):

PR #7126, opened 2026-07-21
feat: improve memory usage of FFZ badges

Nerixyz:
FFZ badges consume a decent amount of memory in Chatterino. Before this PR, they consumed 18.4 MiB. 

We can do two cheap optimizations that cut the use in half:
- Use [`QVarLengthArray`](https://doc.qt.io/qt-6/qvarlengtharray.html) to store the badges for a user instead of `std::set`. `std::set` always allocates, even if there's only one badge. Since most users only have one, we create many small allocations. With `QVarLengthArray<T, N>`, `N` elements are stored inline. This reduced the memory usage to 11.9 MiB.
- Use `boost::unordered_flat_map` instead of `std::unordered_map`. Because of relaxed requirements for the Boost map, it doesn't need to store as much auxiliary data. This reduced the memory usage to 9.3 MiB.

A further optimization could be the use of a `boost::container::flat_multimap<std::string, int>`. This gets us to 4.6 MiB. I haven't measured the performance of this. We'd have to assume that FFZ returns us unique users per badge. The comparison with `QString` for the lookup can be done by comparing `QString < QLatin1StringView` (as we know badge IDs are latin1).

### DIFF
--- a/src/providers/ffz/FfzBadges.cpp
+++ b/src/providers/ffz/FfzBadges.cpp
@@ -107,7 +107,8 @@ void FfzBadges::load()
                     auto userIDString = QString::number(user.toInt());
 
                     auto [userBadges, created] = this->userBadges.emplace(
-                        std::move(userIDString), badgeID);
+                        std::move(userIDString),
+                        QVarLengthArray<int, 2>{badgeID});
                     if (!created)
                     {
                         // User already had a badge assigned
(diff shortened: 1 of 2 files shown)


pajlada (approved):
Nice, easy win. Feel free to look into further optimizations but this is already great.

PR #7134, opened 2026-07-25
deps: Update libcommuni and use `Communi::TagsRef`

Nerixyz:
This updates `libcommuni` to include https://github.com/Chatterino/libcommuni/pull/13 which refactors the tag handling (see PR for the reasons).

The only change here is the replacement of the map and the lookup functions.

### DIFF
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -523,15 +523,15 @@ EmotePtr parseEmote(TwitchChannel *twitchChannel, const EmoteName &name)
     return {};
 }
 
-std::pair<QString, bool> parseMessageType(const QVariantMap &tags)
+std::pair<QString, bool> parseMessageType(Communi::TagsRef tags)
 {
-    auto msgId = tags.value("msg-id").toString();
+    auto msgId = tags.getOrEmpty("msg-id");
 
     bool mirrored = msgId == "sharedchatnotice";
 
     if (mirrored)
     {
-        msgId = tags.value("source-msg-id").toString();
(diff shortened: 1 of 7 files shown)


pajlada:
I snuck in my changes from the highlights reword that disables the hype chat functionality of Chatterino, which have caused some merge conflicts for you

src/messages/MessageBuilder.cpp
```diff
@@ -2253,16 +2242,16 @@ void MessageBuilder::parseThread(const QString &messageContent,
...
                 channel);
             if (ignored)
             {
                 body = QString("[Blocked user]");
             }
             else
             {
-                auto name = replyDisplayName->toString();
-                body = parseTagString(replyBody->toString());
+                auto name = *std::move(replyDisplayName);
```
pajlada: Out of curiosity, what is the purpose of or why do we need the move here?

Nerixyz: Idea: Avoid copying (i.e. ref-counting). Not super necessary here, because we'll copy `name` later anyway. An alternative could be to declare `const auto &name = *replyDisplayName`.

pajlada: Would that optimization be relevant in the scenario where the underlying `QVariantMap` actually points to a `QStringView` instead?
If no, I'd rather not use `std::move` because I think it reduces the readability/understandability of the code here.

pajlada: same feedback goes for the other `std::move` uses in the same file, didn't wanna spam

pajlada (approved):

PR #7137, opened 2026-07-26
fix: make Notebook::findItem safe

### LINKED ISSUE #7131: Notebook::getItem is unsafe
It should return an `std::optional<Item>` instead of an item pointer. In some circumstances, this can cause a crash when the user attempts to duplicate a tab.

`Notebook::select` has a branch where it makes a modification to the item it finds - that place should iterate/find the item itself instead of relying on `findItem`.

4rneee:
Closes #7131

### DIFF
--- a/src/widgets/Notebook.cpp
+++ b/src/widgets/Notebook.cpp
@@ -368,8 +368,8 @@ void Notebook::select(QWidget *page, bool focusPage, bool recordInHistory)
         // Hide the previously selected page
         this->selectedPage_->hide();
 
-        auto item = std::ranges::find_if(
-            this->items_.begin(), this->items_.end(), [this](const auto &item) {
+        auto item =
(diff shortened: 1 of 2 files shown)


src/widgets/Notebook.cpp
```diff
@@ -368,8 +368,11 @@ void Notebook::select(QWidget *page, bool focusPage, bool recordInHistory)
         // Hide the previously selected page
         this->selectedPage_->hide();
 
-        auto *item = this->findItem(this->selectedPage_);
-        if (!item)
+        auto item = std::ranges::find_if(
+            this->items_.begin(), this->items_.end(), [this](const auto &item) {
```
Nerixyz: Can this take just `this->items_`?

pajlada (approved):
Thanks!

PR #7138, opened 2026-07-27
refactor: move JSON serialization to descriptors

Nerixyz:
The last main part from #6906: Produce the JSON in the descriptors instead of in the window layout. This mainly moves the content from `encodeNodeRecursively` into `toJson` functions on the descriptors.


pajlada (approved):
awesome

PR #7139, opened 2026-07-28
refactor: remove dependency on QtSvgWidgets

Nerixyz:
To ease embedding in OBS, this removes the dependency on QtSvgWidgets. OBS only ships QtSvg, not QtSvgWidgets. We only use `QSvgWidget`. This is a simple class to implement ([source](https://github.com/qt/qtsvg/blob/487dc571a1c02e9ecc6ccf56eb2eafc99fc59375/src/svgwidgets/qsvgwidget.cpp)), as it calls out to `QSvgRenderer` to do the heavy work. Hence, we now include a reimplementation.


src/widgets/helper/SvgWidget.cpp
```diff
@@ -0,0 +1,48 @@
+#include "widgets/helper/SvgWidget.hpp"
```
pajlada, line 5: Missing copyright header

src/widgets/helper/SvgWidget.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
pajlada, line 5: Missing copyright header

pajlada (changes requested):
Seems reasonable, just these two nits

pajlada (approved):

Mm2PL (approved):

PR #7142, opened 2026-07-29
build(windows): Enable segment heap if possible

Nerixyz:
With Visual Studio 2026 18.6, Microsoft included a script to easily enable the segment heap for apps ([blog](https://devblogs.microsoft.com/cppblog/segment-heap-support-for-c-projects-in-visual-studio/)). The segment heap was already available for some years, but it hasn't been enabled by default. With the provided script we can add a manifest to enable it. In my basic testing it reduced the memory usage from ~75MiB to ~70MiB with one tab and 4 messages. So it should be a free win for us.

### DIFF
--- a/cmake/EnableSegmentHeap.cmake
+++ b/cmake/EnableSegmentHeap.cmake
@@ -1,0 +1,17 @@
+if (NOT WIN32 OR MINGW)
+    return()
+endif()
+
+if(NOT DEFINED ENV{VSINSTALLDIR})
+    message(WARNING "Missing VSINSTALLDIR environment variable - not enabling segment heap.")
+    return()
+endif()
+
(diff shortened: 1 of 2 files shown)


cmake/EnableSegmentHeap.cmake
```diff
@@ -0,0 +1,17 @@
+if (NOT WIN32 OR MINGW)
+    return()
+endif()
```
pajlada: I would prefer if this is moved to the main `CMakeLists.txt` file.


Additionally, a copyright header would be great


```suggestion
# SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
#
# SPDX-License-Identifier: MIT

```

Feel free to use MIT or something more permissive like CC0-1.0

pajlada (changes requested):
Just a nit, otherwise great. This seems like a free win.

pajlada (approved):

Issue #6205, opened 2025-05-11
Chatterino doesn't refresh after sleep on Plasma 6


aknight2015:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

OS: Debian 12
Desktop: Plasma 6 with Wayland
Manager: GDM3
Whenever I put my laptop to sleep and resume Chatterino still displays the channels live, there were reported live before the sleep. After I wake it up I'll wait like 10 minutes and there's no change. I have to close it and reopen it in order for it to update.

### Screenshots

I wish I could, but I can't record my computer going to sleep.

### OS and Chatterino Version

Chatterino 2.5.3 (Commit 3f3a31db) build with Qt 6.7.2 on Debian 12 (Bookworm)

Nerixyz:
It would be great if you could capture the output of Chatterino. See [Viewing Logs](https://wiki.chatterino.com/Developer%20Logs) on how to get those. `chatterino.*.debug=true` should be enough I think.

aknight2015:
Sure. Do I use the CLI to launch the chatterino appimage using the above mentioned flags?

aknight2015:
I've run Chatterino with those flags. Where does it put the log file?
`#!/bin/bash
export QT_LOGGING_RULES="chatterino.*.debug=true"
./Chatterino-x86_64.AppImage`
if anyone else has trouble I made the above a script. That way in the future you can just share that with people.

pajlada:
It doesn't log to a file, it should log to your terminal.
You can log it to a file for later viewing like this:
`2>chatterino.log ./Chatterino-x86_64.AppImage`
or when you run your script
`2>chatterino.log ./your-script.sh`

aknight2015:
Crap. I thought it went to a file. I'll add that to the script and run it again. Good news is that the problem persisted, so it can be reproduced on my end.

aknight2015:
[chatterino.log](https://github.com/user-attachments/files/20236943/chatterino.log)

Here's the log file. Sorry it's so large.

pajlada:
[...]

PR #7150, opened 2026-08-02
fix: Clear connection cache when we hit a timeout in eventsub::SubscriptionRequest

invisibl3cat:
If the IP address of the machine changes while Chatterino is running, the connection that is used to issue subscription requests can get stuck in an unusable state from which it may not recover. When that happens, subscription attempts keep throwing errors like:

```
chatterino.twitch.eventsub: Unhandled error, retrying subscription "TimeoutError" eventsub::SubscriptionRequest[channel.chat.user_message_hold.1]{broadcaster_user_id=110240192, user_id=503262342}
```
The broken connection will still be using the old IP address and its outbound queue will keep growing.

```
      ESTAB 0      0      192.168.1.198:45778   34.212.92.60:443  users:(("chatterino",pid=7037,fd=61))
      ESTAB 0      594    192.168.1.196:53678    78.47.52.88:443  users:(("chatterino",pid=7037,fd=74))
      ESTAB 0      0      192.168.1.198:54988  172.67.68.177:443  users:(("chatterino",pid=7037,fd=44))
      ESTAB 0      0      192.168.1.198:36384  37.27.171.119:443  users:(("chatterino",pid=7037,fd=46))
-->   ESTAB 0      130572 192.168.1.196:34568  13.227.192.16:443  users:(("chatterino",pid=7037,fd=48))
      ESTAB 0      0      192.168.1.198:48480 199.232.18.214:443  users:(("chatterino",pid=7037,fd=49))
```

After sufficient time the outbound queue eventually fills up, next attempt to write to it will fail, allowing Chatterino to detect an error and create a new connection. That, however, is not very reliable and can take a long time.

Since the problematic connection is handled internally by QNetworkAccessManager, options how to deal with this are limited. What seems to work is calling `clearConnectionCache()` when a network request made through QNetworkAccessManager fails with a timeout.

I can reproduce the issue reliably by adjusting connection options in NetworkManager to make my computer appear like a different machine every time it connects to my WiFi router. DHCP then assigns it a different IP when I disconnect and connect back, triggering the issue.

This might solve https://github.com/Chatterino/chatterino2/issues/6205

### DIFF
--- a/src/common/network/NetworkTask.cpp
+++ b/src/common/network/NetworkTask.cpp
@@ -194,6 +194,8 @@ void NetworkTask::timeout()
         << this->data_->typeString() << "[timed out]"
         << this->data_->request.url().toString();
 
+    NetworkManager::accessManager->clearConnectionCache();
+
     this->data_->emitError({NetworkResult::NetworkError::TimeoutError, {}, {}});
     this->data_->emitFinally();
 }
--- a/src/common/network/NetworkTask.cpp
+++ b/src/common/network/NetworkTask.cpp
@@ -194,6 +194,19 @@ void NetworkTask::timeout()
         << this->data_->typeString() << "[timed out]"
         << this->data_->request.url().toString();
 
+    // If we have a stuck connection, QNetworkAccessManager does not have
+    // any means to detect it as stuck and may try to keep using it to issue
+    // HTTP requests. This will cause all issued requests to fail with
+    // a timeout and no indication that they were probably never delivered
+    // to the remote server.
+    //
+    // Calling clearConnectionCache() removes all connections from cache,
+    // forcing QNetworkAccessManager internals to open up a new healthy
+    // connection.
+    //
+    // Since we have no good way of detecting if the request timed out because
+    // of a stuck connection, clearing the connection cache on every timeout
+    // seems like the only way how to reliably recover from such a situation.
     NetworkManager::accessManager->clearConnectionCache();
 
     this->data_->emitError({NetworkResult::NetworkError::TimeoutError, {}, {}});


invisibl3cat:
One thing to note here is that this behavior is most likely specific to Linux and maybe some BSDs, at least as far as my quick research goes. Windows should set the socket to a failed state when a situation like this occurs. I have not written any test code to verify it though.

src/common/network/NetworkTask.cpp
```diff
@@ -194,6 +194,8 @@ void NetworkTask::timeout()
         << this->data_->typeString() << "[timed out]"
         << this->data_->request.url().toString();
 
+    NetworkManager::accessManager->clearConnectionCache();
```
pajlada, line 210: This needs a proper comment for why we're clearing the connection cache

pajlada (commented):
Haven't really looked into this / its validity, but if it's harmless and seems to fix the issue for you, I'm happy to merge this in after taking a little bit of a closer look

src/common/network/NetworkTask.cpp
```diff
@@ -194,6 +194,21 @@ void NetworkTask::timeout()
...
+    // HTTP requests. This will cause all issued requests to fail with
+    // a timeout and no indication that they were probably never delivered
+    // to the remote server.
+    //
+    // Calling clearConnectionCache() removes all connections from cache,
+    // forcing QNetworkAccessManager internals to open up a new healthy
+    // connection.
+    //
+    // Since we have no good way of detecting if the request timed out because
+    // of a stuck connection, clearing the connection cache on every timeout
```
Nerixyz, line 208: There's `QIODevice::readyRead` and `QNetworkReply::downloadProgress` that somewhat indicate that a connection was established.

invisibl3cat: I don't think any of these would help here. To connection object itself is buried deeply in the Qt network access infrastructure so we cannot probe it ourselves. The bigger issue is that the socket will look perfectly fine from the application perspective. The issue we're facing is a connection that appears to be established but cannot actually transmit anything.

Nerixyz: > but cannot actually transmit anything.

So if we find out that we transmitted something, we know that we're not in this situation.

invisibl3cat: True, but the only way for us to reliably detect that something was actually sent from the socket is by receiving a reply to it.

PR #7153, opened 2026-08-03
fix: use channel type from `IndirectChannel`

Nerixyz:
We need to use the type from the indirect channel, otherwise the `Watching` type won't be retained.


pajlada (approved):
👍

PR #7158, opened 2026-08-05
refactor: extract SplitBanner from PinnedMessageWidget

janzezno:
Moves the generic banner out of `PinnedMessageWidget`  into a new `SplitBanner` base class, so future banners shown between the split header and the chat don't have to reimplement it.

`SplitBanner` provides:
- the header label, header row and content box layout
- countdown label
- the shared scale and paint handling

`PinnedMessageWidget` now derives from it and keeps only its usage specific parts.

No behavior change - the pinned message banner looks and behaves as before.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -865,6 +865,8 @@ set(SOURCE_FILES
         widgets/splits/InputHighlighter.hpp
         widgets/splits/Split.cpp
         widgets/splits/Split.hpp
+        widgets/splits/SplitBanner.cpp
+        widgets/splits/SplitBanner.hpp
         widgets/splits/SplitCommon.hpp
         widgets/splits/SplitContainer.cpp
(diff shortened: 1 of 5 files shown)


pajlada:
Hey, thanks for your contribution.
Do you have a future use case in mind for this widget after you've refactored it out?

The reason I ask is: Making something generic/general is not always the right approach when it comes to code cleanliness, so if we don't have a specific use case for it, it might be best to leave specialized until we have a reason to make it generic.

janzezno:
Yes - a prediction banner for Kick in SevenTV/chatterino7#448. Same position as the pinned banner and the same chassis. Extracting the base was a suggestion by @Nerixyz on that PR. `PinnedMessageWidget` is the only in-tree user, the other is downstream in the fork.

Nerixyz:
You don't need to merge every time `master` updates unless there are conflicts.

Issue #146, opened 2017-12-17
Streamlink and settings for it probably don't work


fourtf:

pajlada:
Streamlink has basic functionality, the "choose" setting just doesn't work as expected imo but maybe I just don't know how it's supposed to work.
I restructured the code a bit in ef9aedb2cdc50889470f8938c0e7f15a91ff7571 to easier give some console error messages in case the path is wrong.

Confuseh:
So the issue with the "choose" not working might be due to `default-stream=<quality>` being used [in config], since it relies on the last line starting with "Available streams" that gets printed to stdout, when no quality and no `--default-stream` are specified.
There is no simple ways to prevent this, since streamlink does not include an option just to show quality options. It is possible to add `--default-stream someinvalidbullshithere` to force an error, but then "Available streams" part is indented [ I guess can be resolved with a trim ].

PR #7162, opened 2026-08-06
feat: allow opening arbitrary urls with streamlink

4rneee:
For #7156

If the argument is not a URL it is interpreted as twitch channel name so that the current behavior is unchanged.

I also renamed related functions and variable names to reflect this new behavior.

The `/streamlink` command pre-processes the argument by removing leading `@` or `#` and a trailing `,`, which shouldn't cause issues for URLs, right?

Unrelated: while reading through the code I noticed that we spawn the streamlink process with `--default-stream=KKona`. Is there any lore behind this? KKona

### DIFF
--- a/src/controllers/commands/builtin/Misc.cpp
+++ b/src/controllers/commands/builtin/Misc.cpp
@@ -340,15 +340,15 @@ QString streamlink(const CommandContext &ctx)
         else
         {
             ctx.channel->addSystemMessage(
-                "/streamlink [channel]. Open specified Twitch channel in "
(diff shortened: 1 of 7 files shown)


4rneee:
> Unrelated: while reading through the code I noticed that we spawn the streamlink process with --default-stream=KKona. Is there any lore behind this?

Found the answer myself: https://github.com/Chatterino/chatterino2/issues/146#issuecomment-352262951

pajlada (approved):

PR #7166, opened 2026-08-08
fix: don't crash on invalid plugins

pajlada:
This makes our plugin storage use a variant for the different plugin
states:

- UnloadedPlugin (used if the plugin's meta was invalid)
- PluginPtr (a normal valid plugin, which may still contain lua errors)

How to test the crash that doesn't crash with this PR:
Add an invalid permission in your `info.json` file

### DIFF
--- a/src/controllers/plugins/Plugin.hpp
+++ b/src/controllers/plugins/Plugin.hpp
@@ -42,7 +42,7 @@ namespace chatterino {
 
 /// A plugin that hasn't been loaded.
 ///
-/// Most likely, it's metadata is invalid.
+/// Most likely, its metadata is invalid.
 class UnloadedPlugin
 {
 public:
--- a/src/controllers/plugins/Plugin.hpp
+++ b/src/controllers/plugins/Plugin.hpp
@@ -75,8 +75,8 @@ public:
     {
         // The PluginMeta here must be valid, otherwise it should be initialized
         // as an UnloadedPlugin
-        assert(meta.isValid());
-        assert(state != nullptr);
+        assert(this->meta.isValid());
+        assert(this->state_ != nullptr);
     }
 
     ~Plugin();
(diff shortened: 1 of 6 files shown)


src/controllers/plugins/Plugin.hpp
```diff
@@ -37,6 +40,25 @@ struct SignalCallback;
 
 namespace chatterino {
 
+/// A plugin that hasn't been loaded.
+///
+/// Most likely, it's metadata is invalid.
```
Mm2PL: typo
```suggestion
/// Most likely, its metadata is invalid.
```

src/controllers/plugins/PluginController.hpp
```diff
@@ -45,8 +46,9 @@ class PluginController
     // This is required to be public because of c functions
     Plugin *getPluginByStatePtr(lua_State *L);
 
-    // TODO: make a function that iterates plugins that aren't errored/enabled
-    const std::map<QString, std::unique_ptr<Plugin>> &plugins() const;
+    void forEachPlugin(
+        FunctionRef<void(const std::unique_ptr<Plugin> &)>) const;
```
Mm2PL, line 51: Maybe a comment here would be nice
```suggestion
    // Run `cb` on every loaded plugin including those with load errors
    void forEachPlugin(
        FunctionRef<void(const std::unique_ptr<Plugin> &)> cb) const;
```

src/controllers/plugins/Plugin.hpp
pajlada: Applied in [`6f9ea68` (this PR)](https://github.com/Chatterino/chatterino2/pull/7166/commits/6f9ea683eeaf857e4c3826b792e3cc3fc96480a4)

src/controllers/plugins/PluginController.hpp
pajlada: Applied in [`6f9ea68` (this PR)](https://github.com/Chatterino/chatterino2/pull/7166/commits/6f9ea683eeaf857e4c3826b792e3cc3fc96480a4)

Mm2PL (approved):

PR #7168, opened 2026-08-09
fix: Use Websocket ping - pong to detect a stuck connection

invisibl3cat:
If a Websocket connection fails in a way that the connection appears serviceable from the application's perspective but cannot really transfer any data, there is currently no way for us to detect this. This may happen after some network configuration changes or if there is an issue somewhere along the way between us and the remote peer.
We can leverage Websocket ping and pong frames to periodically probe the remote peer if we do not see any other traffic going through the Websocket connection.

### DIFF
--- a/src/common/websockets/detail/WebSocketConnectionImpl.cpp
+++ b/src/common/websockets/detail/WebSocketConnectionImpl.cpp
@@ -10,6 +10,7 @@
 #include <boost/asio/strand.hpp>
 #include <boost/beast/core/bind_handler.hpp>
 #include <boost/beast/websocket/ssl.hpp>
+#include <boost/bind/bind.hpp>
 
 using namespace std::literals::string_view_literals;
 
@@ -17,6 +18,15 @@ namespace chatterino::ws::detail {
 
 namespace asio = boost::asio;
 namespace beast = boost::beast;
+
+// Use a long ping probe interval when there is regular traffic
+// going through the socket
+static constexpr std::chrono::seconds HEALTH_CHECK_INTERVAL_LONG =
+    std::chrono::seconds{60};
+// Use a short intervat when there is no traffic except for the
+// ping probe so that we can detect a failed socket quickly
+static constexpr std::chrono::seconds HEALTH_CHECK_INTERVAL_SHORT =
+    std::chrono::seconds{5};
 
 // MARK: WebSocketConnectionHelper
 
@@ -28,6 +38,8 @@ WebSocketConnectionHelper<Derived, Inner>::WebSocketConnectionHelper(
     : WebSocketConnection(std::move(options), id, std::move(listener), pool,
                           ioc)
     , stream(std::move(stream))
+    , healthCheckTimer(this->stream.get_executor())
+    , pingProbesTried(0)
 {
 }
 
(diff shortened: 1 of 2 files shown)


src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -247,6 +259,56 @@ void WebSocketConnectionHelper<Derived, Inner>::onWsHandshake(
...
+        this->fail(boost::asio::error::network_down, u"ping");
+        return;
+    }
+
+    this->stream.async_ping({}, [](const boost::system::error_code ec) {
+        (void)ec;
+    });
+    this->pingProbesTried++;
+
+    // Use the short intervat so that we can detect a failed connection quickly
```
Wissididom:
```suggestion
    // Use the short interval so that we can detect a failed connection quickly
```

src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -18,6 +19,15 @@ namespace chatterino::ws::detail {
 namespace asio = boost::asio;
 namespace beast = boost::beast;
 
+// Use a long ping probe interval when there is regular traffic
+// going through the socket
+static constexpr std::chrono::seconds HEALTH_CHECK_INTERVAL_LONG =
+    std::chrono::seconds{60};
+// Use a short intervat when there is no traffic except for the
```
Wissididom:
```suggestion
// Use a short interval when there is no traffic except for the
```

src/common/websockets/detail/WebSocketConnectionImpl.hpp
```diff
@@ -85,6 +89,9 @@ class WebSocketConnectionHelper : public WebSocketConnection,
     /// When we successfully resolve the host, we try to connect by
     /// iterating over these results.
     BalancedResolverResults resolvedEndpoints;
+
+    boost::asio::steady_timer healthCheckTimer;
+    int pingProbesTried;
```
Nerixyz: Use an unsigned type and initialize it here.

src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -18,6 +19,15 @@ namespace chatterino::ws::detail {
...
 namespace beast = boost::beast;
 
+// Use a long ping probe interval when there is regular traffic
+// going through the socket
+static constexpr std::chrono::seconds HEALTH_CHECK_INTERVAL_LONG =
+    std::chrono::seconds{60};
+// Use a short interval when there is no traffic except for the
+// ping probe so that we can detect a failed socket quickly
+static constexpr std::chrono::seconds HEALTH_CHECK_INTERVAL_SHORT =
+    std::chrono::seconds{5};
```
Nerixyz: The comment and the variable name are confusing. They make it sounds like the ping check has two different modes _that depend on the application messages_. So if the application doesn't send/receive anything, we'd send pings every 5s.

Looking at the code, that's not what we do.

src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -247,6 +259,54 @@ void WebSocketConnectionHelper<Derived, Inner>::onWsHandshake(
         this->readBuffer,
         beast::bind_front_handler(&WebSocketConnectionHelper::onReadDone,
                                   this->shared_from_this()));
+
+    this->stream.control_callback(
+        boost::bind(&WebSocketConnectionHelper::onControlFrame, this,
+                    boost::placeholders::_1, boost::placeholders::_2));
```
Nerixyz: Why can't this and the one below use `beast::bind_front_handler`?

src/common/websockets/detail/WebSocketConnectionImpl.cpp
invisibl3cat: True, the current text is a bit misleading. Updated phrasing should describe the actual logic correctly.

src/common/websockets/detail/WebSocketConnectionImpl.cpp
invisibl3cat: I just replaced the binds with lambdas.

src/common/websockets/detail/WebSocketConnectionImpl.cpp
```diff
@@ -247,6 +257,58 @@ void WebSocketConnectionHelper<Derived, Inner>::onWsHandshake(
...
+template <typename Derived, typename Inner>
+void WebSocketConnectionHelper<Derived, Inner>::onControlFrame(
+    boost::beast::websocket::frame_type frame_type, std::string_view payload)
+{
+    (void)payload;
+
+    if (frame_type == boost::beast::websocket::frame_type::pong)
+    {
+        this->pingProbesTried = 0;
+        this->healthCheckTimer.expires_after(HEALTH_CHECK_INTERVAL_LONG);
```
pajlada: Whenever we call `expires_after` on a steady timer, it will cancel any existing `async_wait` calls. This means `onHealthCheck` will never be called after this.
We will need some mechanism to handle this case - either don't reset the timer, or explicitly call `async_wait` again

invisibl3cat: Dang, such a lame mistake :) It should be okay now.

FTR it is possible to not see the periodic pinging if there is other data being regularly received over the socket. But that doesn't seem to happen in practice.

pajlada (approved):

PR #7174, opened 2026-08-12
feat(commands): add element.copytext placeholder

### LINKED ISSUE #7145: Allow commands to access the element that was right-clicked
Commands can be executed by typing them in chat: e.g. `/my-command arg1 arg2`, and commands can also be executed by right-clicking a message and selecting it. Commands are available in the right-click menu if "Show in message menu" is enabled.

It would be cool if commands could have access to certain things about the actual element that was right-clicked, e.g. if I right-click an emote, could the command access that specific emote's copy text? (i.e. the name?).
This would allow me to create a command that searches for an emote, something like `/openurl https://pajlada.se/my-emote-search.php?query={element.copytext}`

4rneee:
This adds a `element.copytext` placeholder for commands that are executed through the right click menu.

Resolves #7145

### DIFF
--- a/src/widgets/helper/ChannelView.cpp
+++ b/src/widgets/helper/ChannelView.cpp
@@ -2962,16 +2962,11 @@ void ChannelView::addCommandExecutionContextMenuItems(
     auto *cmdMenu = new QMenu(menu);
     executeAction->setMenu(cmdMenu);
 
-    QString elementCopyText = "";
+    QString elementCopyText;
     if (hoveredElement != nullptr)
     {
         hoveredElement->addCopyTextToString(elementCopyText);
-        // remove trailing space which addCopyTextToString adds if the hovered
-        // element is followed by a space
-        if (elementCopyText.endsWith(' '))
-        {
-            elementCopyText.truncate(elementCopyText.length() - 1);
-        }
+        elementCopyText = elementCopyText.trimmed();
     }
 
     for (auto &cmd : cmds)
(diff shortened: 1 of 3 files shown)


src/widgets/helper/ChannelView.cpp
```diff
@@ -2961,6 +2962,18 @@ void ChannelView::addCommandExecutionContextMenuItems(
     auto *cmdMenu = new QMenu(menu);
     executeAction->setMenu(cmdMenu);
 
+    QString elementCopyText = "";
```
Nerixyz:
```suggestion
    QString elementCopyText;
```

src/widgets/helper/ChannelView.cpp
```diff
@@ -2961,6 +2962,18 @@ void ChannelView::addCommandExecutionContextMenuItems(
...
+    if (hoveredElement != nullptr)
+    {
+        hoveredElement->addCopyTextToString(elementCopyText);
+        // remove trailing space which addCopyTextToString adds if the hovered
+        // element is followed by a space
+        if (elementCopyText.endsWith(' '))
```
Nerixyz: Why not use `QString::trimmed` here? Tbf, most elements don't have more than one space at the end.

src/widgets/helper/ChannelView.cpp
```diff
@@ -2983,9 +2997,8 @@ void ChannelView::addCommandExecutionContextMenuItems(
             // Execute command through right-clicking a message -> Execute command
             QString value = getApp()->getCommands()->execCustomCommand(
                 inputText.split(' '), cmd, true, channel, layout->getMessage(),
-                {
-                    {"input.text", userText},
-                });
+                {{"input.text", userText},
+                 {"element.copytext", elementCopyText}});
```
Nerixyz:
```suggestion
                 {"element.copytext", elementCopyText},});
```
and format

src/widgets/helper/ChannelView.cpp
4rneee: I only wanted to remove the explicitly added space but `trimmed` works too

Mm2PL:
> This adds a `element.copytext` placeholder for commands that are executed through the right click menu.
Please make a PR to the wiki ( https://github.com/chatterino/wiki ) that 
documents this new placeholder

pajlada (approved):

PR #7176, opened 2026-08-15
fix: Plugin API: passing a MessageElement to Message:append_element

brilliantdrink:
When passing a `MessageElement` instead of an init table to `Message:append_element` in a lua plugin, it would error with "Invalid message type" suggesting that it calls `elementFromTable`, i.e. acts as if a `MessageElementInit` table was passed.

Reordering the sol::overload so that the `Message:append_element(MessageElement)` one comes fixes this. I'm not 100% sure on this but it seems that the passed message element somehow matches as a `sol::table` under the hood so the MessageElementInit overload matches before MessageElement gets checked. With the order from this pr both passing a MessageElement and a MessageElementInit works as should.

Small example to test:

```lua
---@param msg c2.Message
local function change_message(msg)
    -- not :clone()ing here to get an empty elements array
    local newMessage = c2.Message.new(msg)
    newMessage:append_element({
        type = 'text'
        text = 'foo'
    })
    for index, element in ipairs(msg:elements()) do
        newMessage:append_element(element)
    end
end

c2.Channel.by_name('myChannel'):on_message_appended(change_message)
```

### DIFF
--- a/src/controllers/plugins/api/Message.cpp
+++ b/src/controllers/plugins/api/Message.cpp
@@ -767,6 +767,12 @@ void createUserType(sol::table &c2)
         },
         "append_element",
         sol::overload(
+            // Message:append_element(MessageElement)
+            // Create a clone of the given element and add it to the message
+            [](Message *msg, ElementRef &element) {
+                checkWritable(msg);
+                msg->elements.emplace_back(element.constElement()->clone());
+            },
             // Message:append_element(MessageElementInit)
             // Create a new element
             [](Message *msg, const sol::table &tbl) {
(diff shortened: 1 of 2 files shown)


Nerixyz:
Thank you, good catch! Can you add a test in https://github.com/Chatterino/chatterino2/blob/bf87302eb4837397a498d7a669860e859949b588/tests/lua/message/clone.lua (or some other file)? You can add a new function to the `tests` entry. Then we can make sure to not miss this again.

brilliantdrink:
@Nerixyz thanks for responding so quickly. I've added a test as a new `element.lua` file, its asserting `append_element` is executing without error for text inits and text element, and that the element was actually added.

pajlada (approved):

PR #7178, opened 2026-08-15
chore(twitch): Cleanup TwitchIrcServer

Nerixyz:
This cleans up some minor issues I found when adding different read connection strategies.

It consists of
- Passing methods directly to `QObject::connect`.
- Remove unused members.
- Remove unused arguments.
- Inline call to `open` when calling `initializeConnection` so it can be `static`.


src/providers/twitch/TwitchIrcServer.cpp
```diff
@@ -356,7 +344,7 @@ void TwitchIrcServer::initializeConnection(IrcConnection *connection,
     connection->setPort(Env::get().twitchServerPort);
     connection->setSecure(Env::get().twitchServerSecure);
 
-    this->open(type);
+    connection->open();
```
pajlada, line 347: This change no longer locks the connection mutex

pajlada (changes requested):

pajlada (approved):

Issue #7117, opened 2026-07-19
Add a light theme for the settings dialog


pajlada:
### Discussed in https://github.com/Chatterino/chatterino2/discussions/4823

We should add a light theme for the settings dialog (and other dark-mode-only dialogs like the color picker).
This should not respect your exact theme, but be a good light theme.
It should be based on the theme's `isLightTheme` function.
Theme has an `updated` signal that could be used.

PR #7182, opened 2026-08-17
feat: Add light theme variant for the settings UI

invisibl3cat:
This my first stab at implementing a light version of the settings UI. The code needs more than a few touch-ups but I'd like to get some feedback on the visuals. It is mostly based off the "Light" theme we already have. Currently needs a Chatterino restart for the theme to apply.

dark1
dark2
light1
light2

Addresses https://github.com/Chatterino/chatterino2/issues/7117


treejadey:
Hi, I think this PR is pretty good (from a visual standpoint, at least), but in my opinion  the colors could be a bit better. I did a really basic mockup (by eye) of what I think the colors should look like:

Frame 1 (1)

As you can see, it makes the colors a bit brighter, the inner part gets brighter instead of darker, and the elements are white to make them stand out more, as well as with the colors being darker and more readable.

---

The colors used in the image are:

- `#F2F2F2` for the **outer** background
- `#F9F9F9` for the **inner** background

- `#363636` for the darker text
- `#2D5992` for the dark blue color
- `#6F839D` for the washed out blue color on the right side
- `#888888` for the washed out color in the search bar

invisibl3cat:
The light theme is based off the light theme that is currently used for the main window. I suppose they should be kept in sync so changing the color palette itself is probably out of scope of this PR. I'd be open to do a follow-up with some tweaks to the palette if you have any suggestions.

PR #7183, opened 2026-08-17
Fix incognito links on MacOS and Opera

badoge:
IncognitoBrowser.cpp didn't handle the MacOS stuff so the links context menu in chat and the settings didn't have incognito options.

All browsers listed in getPrivateSwitch were tested on MacOS 26.5.2 except for the firefox forks because I didn't feel like it :) 

Safari doesn't support opening a private window using a flag, and the workaround that everyone uses is very scuffed so I didn't add it.

When testing I noticed that Opera never opened links in a private window even on other operating systems. And it turned out to be because we were using the wrong flag "-newprivatetab" when we should have been using "-incognito" because Opera is Chromium based.

Code was mostly clanker generated :)


src/util/IncognitoBrowser.cpp
```diff
@@ -68,44 +73,27 @@ namespace chatterino::incognitobrowser::detail {
...
-        lowercasedBrowserExecutable.chop(4);
-    }
-#endif
+    auto lowercasedBrowserExecutable =
+        QFileInfo(QDir::cleanPath(browserExecutable)).baseName().toLower();
 
     for (const auto &switch_ : switches)
     {
-        if (lowercasedBrowserExecutable == switch_.first)
+        if (lowercasedBrowserExecutable.contains(switch_.first))
```
pajlada: Undo change to firefox & contains usage here

src/util/MacOsHelpers.h
```diff
@@ -1,3 +1,13 @@
 #pragma once
 
+#include <qglobal.h>
```
pajlada: include `QString` instead

pajlada (changes requested):

pajlada (approved):

PR #7188, opened 2026-08-20
ci: Harden actions and dependabot (a bit)

Nerixyz:
To prevent issues from using GitHub actions and dependabot, I hardened the workflows using [zizmor](https://docs.zizmor.sh/). There are still 77 open findings. I only applied the ones that are easy wins.

- Pin actions to a commit hash instead of tag - tags are not immutable (see [`unpinned-uses`](https://docs.zizmor.sh/audits/#unpinned-uses))
- Disable `persist-credentials` on checkout - we don't need it (see [`artipacked`](https://docs.zizmor.sh/audits/#artipacked))
- Set cooldown on dependabot updates (see [`dependabot-cooldown`](https://docs.zizmor.sh/audits/#dependabot-cooldown)). For our own packages, we can manually create PRs.

I left the docker images unpinned, because they're frequently updated, but zizmor would like to see them pinned as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
       interval: "daily"
     labels:
       - "ci"
+    cooldown:
+      default-days: 7
   - package-ecosystem: "gitsubmodule"
     directory: "/"
     schedule:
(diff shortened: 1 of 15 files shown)


pajlada (approved):
Sweet! Thanks for the contribution, and for sharing the link to the project. Nice free wins

PR #7192, opened 2026-08-22
messageelement: allow for exhaustive flag test

pajlada:
this allows most elements to be configured with the "exhaustiveFlags"
flag which ensures that _all_ of the element's flags must be set in the
context for it to be laid out.

### DIFF
--- a/src/messages/MessageElement.cpp
+++ b/src/messages/MessageElement.cpp
@@ -117,6 +117,12 @@ void MessageElement::cloneFrom(const MessageElement &source)
     this->trailingSpace = source.trailingSpace;
 }
 
+bool MessageElement::matchesFlags(const MessageElementFlags &contextFlags) const
+{
+    return this->exhaustiveFlags ? contextFlags.hasAll(this->getFlags())
+                                 : contextFlags.hasAny(this->getFlags());
+}
+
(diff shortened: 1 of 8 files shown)


src/messages/MessageElement.cpp
```diff
@@ -117,6 +117,12 @@ void MessageElement::cloneFrom(const MessageElement &source)
     this->trailingSpace = source.trailingSpace;
 }
 
+bool MessageElement::matchesFlags(const MessageElementFlags &contextFlags) const
```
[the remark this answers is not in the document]
pajlada: resolved in https://github.com/Chatterino/chatterino2/pull/7192/commits/f8a415f2bf853fd856cc0a665b3f29fec5b4d5c0

src/messages/MessageElement.cpp
```diff
@@ -117,6 +117,12 @@ void MessageElement::cloneFrom(const MessageElement &source)
     this->trailingSpace = source.trailingSpace;
```
Nerixyz, line 117: `exhaustiveFlags` needs to be included here otherwise it's not preserved when an element is cloned.

pajlada: https://github.com/Chatterino/chatterino2/pull/7195

PR #7194, opened 2026-08-22
feat: Add strong view aliases

Nerixyz:
We have some strong aliases/typedefs like `UserId`, `UserName`, `EmoteName`, etc. Currently, they wrap a `QString`. In some cases, we don't want to use owned types. Notably, when looking up items in a map. There, it would be fine to look them up by `QStringView`. However, `QStringView` isn't "tagged" to contain a user ID, username, etc.

This PR adds `-View` types as "tagged" versions of `QStringView`. So now `EmoteName` has a corresponding `EmoteNameView`. The view can be created from the owned type, and they can be compared to each other.

The containers don't know about this relationship - for example, `std::unordered_map<EmoteName, V>` doesn't know that you can do lookup by `EmoteNameView`. To tell them, we need to provide both a transparent hasher (i.e. a hasher working for both types) and a transparent comparator. For the hasher, there's `EmoteNameHash` which forwards to the `QStringView` hash. For the comparison, [`std::equal_to<>`](https://en.cppreference.com/cpp/utility/functional/equal_to_void) can be used. See the tests for examples.

I didn't modify any uses of the types yet.


src/common/Aliases.hpp
```diff
@@ -11,34 +11,89 @@
...
+        }                                                          \
+        QStringAliasCompare(name##View);                           \
+    };                                                             \
+    struct name##Hash {                                            \
+        using is_transparent = void;                               \
+        /* Only allows NameView and Name to be hashed. */          \
+        size_t operator()(name##View n) const                      \
+        {                                                          \
+            return qHash(n.string);                                \
+        }                                                          \
```
pajlada, line 79: Is the View struct implicitly constructed here if we attempt to hash an `EmoteSetId`?

tests/src/Aliases.cpp
```diff
@@ -0,0 +1,209 @@
...
+    EXPECT_FALSE(map.contains(EmoteIdView{u"no in here"}));
+
+    // Can't check at compile time that passing a raw QString(View) here will
+    // fail, because `contains` isn't SFINAE friendly. The following two lines
+    // should fail to compile when uncommented.
+    // EXPECT_TRUE(map.contains("baz"));
+    // EXPECT_TRUE(map.contains(u"baz"));
+
+    // Check that the map still works even if `std::equal_to<>` isn't used.
+    // This should be avoided in practice.
```
pajlada, line 176: What's the difference between the default `KeyEqual` (`std::equal_to<Key>`) and `std::equal_to<>`?

src/common/Aliases.hpp
Nerixyz: Yes

tests/src/Aliases.cpp
Nerixyz: `std::equal_to<Key>` requires the arguments to be `Key`, whereas `std::equal_to<>` allows any argument type (i.e. it only requires `a == b` to compile even if they have different types).

pajlada (approved):

PR #7195, opened 2026-08-23
plugins: include exhaustiveFlags in clone

pajlada:
- **plugins: MessageElement:to_json() returns a JSON string of the element**
- **include exhaustiveFlags in MessageElement::cloneFrom**

### DIFF
--- a/docs/lua-meta/globals.lua
+++ b/docs/lua-meta/globals.lua
@@ -592,6 +592,9 @@ function c2.Menu:insert_separator(before) end
 c2.MessageElementBase = {}
 -- ^^^ this is kinda fake - this table doesn't exist in Lua, we only declare it to add methods
 
+--- Returns the JSON representation of the element
+function c2.MessageElementBase:to_json() end
+
 --- Add flags to this element
 ---
 ---@param flags c2.MessageElementFlag
--- a/docs/lua-meta/globals.lua
+++ b/docs/lua-meta/globals.lua
@@ -592,7 +592,8 @@ function c2.Menu:insert_separator(before) end
 c2.MessageElementBase = {}
 -- ^^^ this is kinda fake - this table doesn't exist in Lua, we only declare it to add methods
 
---- Returns the JSON representation of the element
+--- Returns the pretty-printed JSON representation of the element.
+--- This is meant for debugging and is subject to change.
 function c2.MessageElementBase:to_json() end
 
 --- Add flags to this element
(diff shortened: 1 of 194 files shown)


pajlada:
@Nerixyz  - I've included `to_json` in MessageElement here, which returns a JSON string representation of the message element (i.e. `toJson`). I'm using this to compare message elements inside unit tests.

docs/lua-meta/globals.lua
```diff
@@ -592,6 +592,9 @@ function c2.Menu:insert_separator(before) end
 c2.MessageElementBase = {}
 -- ^^^ this is kinda fake - this table doesn't exist in Lua, we only declare it to add methods
 
+--- Returns the JSON representation of the element
```
Nerixyz: I don't mind this since it's not too much effort for us. But the comment should indicate that it's for debugging only, pretty-printed, and subject to change.

src/messages/MessageElement.cpp
```diff
@@ -136,6 +137,7 @@ QJsonObject MessageElement::toJson() const
         },
         {"tooltip"_L1, this->tooltip_},
         {"flags"_L1, qmagicenum::enumFlagsName(this->flags_.value())},
+        {"exhaustiveFlags"_L1, this->exhaustiveFlags},
```
Nerixyz: This can be optional in the JSON output. So it's only included if it's `true`. That's the case for all our elements right now and it would reduce the diff a bit.

Mm2PL:
Can you please add exhaustive_flags on `c2.MessageElementInitBase` for lua meta? It was missed in #7192.

docs/lua-meta/globals.lua
pajlada: Fixed in https://github.com/Chatterino/chatterino2/pull/7195/commits/f2b81fe2af0d46f7a15e069edfa2af75ab2591ba

src/messages/MessageElement.cpp
pajlada: Fixed in https://github.com/Chatterino/chatterino2/pull/7195/commits/62d4ced3fda6b7cc7815a1e18bc30b3eed94cdbc

pajlada:
> Can you please add exhaustive_flags on `c2.MessageElementInitBase` for lua meta? It was missed in #7192.

Fixed in https://github.com/Chatterino/chatterino2/pull/7195/commits/ac2261b2225b9eaae1cbd0a1ce7c08d9f803f588

PR #7197, opened 2026-08-23
feat: add setting to individually always show a tab

### LINKED ISSUE #7193: Add option to individual tabs: "Always show tab"
### Checklist

# *Edited post to match response from pajlada

Feature wanted: Add a setting when right clicking a tab that says something along the lines of "Always show tab".

I've always wanted to have the feature to hide certain channels when they are offline individually, but let them show when they are live to help reduce clutter for unwanted chats. This feature would also help solve [issue #5027](https://github.com/Chatterino/chatterino2/issues/5027). 

Image

### OS and Chatterino Version

Chatterino 2.5.5 (commit [1bc6e5aa](https://github.com/Chatterino/chatterino2/commit/1bc6e5aa)) built with Qt 6.7.1, MSVC 194435223, Crashpad
Running on Windows 10 Version 22H2, kernel: 10.0.19045

4rneee:
Resolves #7193 

I was not sure how to handle the `Hide all tabs` visibility setting because "Hide all" and "Always show" contradict each other. For now tabs are hidden when `Hide all tabs` is selected even if they have `Always show Tab` checked.

### DIFF
--- a/src/widgets/helper/NotebookTab.cpp
+++ b/src/widgets/helper/NotebookTab.cpp
@@ -166,7 +166,8 @@ NotebookTab::NotebookTab(Notebook *notebook)
                      });
     this->menu_.addAction(this->highlightNewMessagesAction_);
 
-    this->alwaysShowAction_ = new QAction("Show tab while offline", &this->menu_);
+    this->alwaysShowAction_ =
+        new QAction("Show tab while offline", &this->menu_);
     this->alwaysShowAction_->setToolTip("Show the tab while offline even if "
                                         "\"Only show live tabs\" is selected");
     this->alwaysShowAction_->setCheckable(true);
--- a/src/widgets/helper/NotebookTab.cpp
+++ b/src/widgets/helper/NotebookTab.cpp
@@ -166,6 +166,17 @@ NotebookTab::NotebookTab(Notebook *notebook)
                      });
     this->menu_.addAction(this->highlightNewMessagesAction_);
 
+    this->alwaysShowAction_ =
+        new QAction("Always show Tab", &this->menu_);
+    this->alwaysShowAction_->setCheckable(true);
+    this->alwaysShowAction_->setChecked(this->alwaysShow_);
+    QObject::connect(this->alwaysShowAction_, &QAction::triggered,
+                     [this](bool checked) {
+                         this->alwaysShow_ = checked;
+                         this->notebook_->refresh();
+                     });
+    this->menu_.addAction(this->alwaysShowAction_);
+
     this->menu_.addSeparator();
 
     this->notebook_->addNotebookActionsToMenu(&this->menu_);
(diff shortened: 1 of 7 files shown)


src/widgets/helper/NotebookTab.cpp
```diff
@@ -166,6 +166,16 @@ NotebookTab::NotebookTab(Notebook *notebook)
...
 
+    this->alwaysShowAction_ = new QAction("Always show Tab", &this->menu_);
+    this->alwaysShowAction_->setCheckable(true);
+    this->alwaysShowAction_->setChecked(this->alwaysShow_);
+    QObject::connect(this->alwaysShowAction_, &QAction::triggered,
+                     [this](bool checked) {
+                         this->alwaysShow_ = checked;
+                         this->notebook_->refresh();
+                     });
+    this->menu_.addAction(this->alwaysShowAction_);
```
pajlada, line 180: I think I would like this checkbox to be inside the "Tab visibility" submenu.

Additionally, given the current behaviour (which I agree with) that it is only shown if "Only show live tabs", maybe the text could be updated to match that behaviour.
Something like "Always show even if offline"? idk if it's a bit wordy

pajlada (changes requested):

src/widgets/helper/NotebookTab.cpp
romeogiggletoes: What about "Show while offline"? It omits the word 'always' to avoid confusion with the hide all tabs feature.

I wonder if adding it to the ''Tab visibility'' section is a good idea since this submenu shows up when you just right click a empty space too. I have a feeling this menu is broken into individual tab settings and global tab settings as shown in the picture below. I suppose you could make it to where this option in ''Tab visibility'' only shows up when right clicking a tab? 
tabs

pajlada: > I wonder if adding it to the ''Tab visibility'' section is a good idea since this submenu shows up when you just right click a empty space too. I have a feeling this menu is broken into individual tab settings and global tab settings as shown in the picture below. I suppose you could make it to where this option in ''Tab visibility'' only shows up when right clicking a tab?

You're right, let's leave this where it is

romeogiggletoes: > You're right, let's leave this where it is

To me, it's actually quite a conundrum. I feel like it also should categorically be under Tab visibility though like you suggested for it to be neat, and since this option is only used when 'Only show lives tabs' is enabled, I intrinsically want this new option near it. However it would go against the Individual and Global sections and I'd imagine would be harder to implement. What do you think about having the "Show Tab While Offline" option/checkbox only appear under 'Tab visibility' when right clicking a tab?

tabb


Unrelated, my OCD is kicking in and noticed not all options have unified capitalization. 'Top most window' and 'Tab visibility' as well as the submenu options should also be capitalized since the 'Close Multiple Tabs' submenu is fully capitalized.

4rneee: I changed the name to `Show tab while offline` for now and added a tooltip to explain the setting in more detail.

Should I also rename the functions and variables to reflect this new name?

PR #7199, opened 2026-08-24
feat: wrap ascii art at default web chat width

3r01:
limit the width of messages containing ascii art to match twitch web chat (340px) so that they display as intended

a message is classified as ascii art if it contains 40 or more graphemes with code points in the So or Sk categories

screenshots:

<details> <summary>before</summary>

chatterino_LHj2Sdei4e

</details>

<details> <summary>after</summary>

chatterino_fT7JyOmkU9

</details>


Nerixyz (commented):
This feels like it should be a setting. Either for detection or layout. Also do we only care about braille art or everything else too (current state)?

3r01:
> This feels like it should be a setting. Either for detection or layout.

Alright. I will add a setting for it under the advanced section. Default on or off?

> Also do we only care about braille art or everything else too (current state)?

Currently it covers graphemes with code points in the [So](https://www.compart.com/en/unicode/category/So) or [Sk](https://www.compart.com/en/unicode/category/Sk) categories, which includes many more symbols than just braille; emojis, block/boxes, dingbats, and other miscellaneous symbols, e.g. `😀`, `♥`, `★`, `█`, `▓`, `●`, `⣿`, `➜`, `™`, `^`, and `ˇ`. Messages that do not contain 40 of these are left at the width decided by the window.

I think this covers most messages that rely on web chat's native width for formatting, and I have tested a bunch of them, but it is difficult to be exhaustive. 

I figure that requiring 40 of these graphemes is enough that you would very rarely, if ever, see any false positives. 

The `So` category appears to contain symbols from some Asian languages such Tamil and Bengali but, from what I can tell, they are symbols within those languages rather than characters used in ordinary writing and would not appear 40 times in a typical message of the respective language.

src/messages/AsciiArt.hpp
```diff
@@ -0,0 +1,13 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <QString>
+
+namespace chatterino::detail {
```
pajlada: No need to use `detail` here - it's used for APIs that we would technically want to keep source-only but want to add unit tests for

src/messages/MessageBuilder.cpp
```diff
@@ -1804,10 +1803,8 @@ std::pair<MessagePtrMut, HighlightAlert> MessageBuilder::makeIrcMessage(
         });
     twitchEmotes.erase(uniqueEmotes.begin(), uniqueEmotes.end());
 
-    // words
-    QStringList splits = content.split(' ');
-
-    builder.addWords(splits, twitchEmotes, textState);
+    builder->isAsciiArt = detail::isAsciiArt(content);
+    builder.addWords(content.split(u' '), twitchEmotes, textState);
```
pajlada: This detection should be done behind a setting as Nerixyz mentioned, it should be off-by-default.

tests/src/AsciiArt.cpp
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: 2026 Contributors to Chatterino <https://chatterino.com>
+//
+// SPDX-License-Identifier: MIT
+
+#include "messages/AsciiArt.hpp"
+
+#include "common/Literals.hpp"
+#include "messages/Message.hpp"
+
+#include <gtest/gtest.h>
```
pajlada: Don't include gtest directly, include our `Test.hpp` instead

pajlada (changes requested):

3r01:
I have made the requested changes.

In my previous comment I said that I would use the advanced section but it seemed to fit better under the Messages section. 

It is off by default.

image

pajlada:
Under Messages is good

3r01:
I forgot to include `QList` and didn’t notice because it was included by the windows precompiled header. This is now fixed.

src/messages/AsciiArt.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    const auto category = QChar::category(codePoint);
+    return category == QChar::Symbol_Other ||
+           category == QChar::Symbol_Modifier;
+}
+
+qsizetype countArtCells(QStringView text)
+{
+    const auto value = text.toString();
+    QTextBoundaryFinder boundaries(QTextBoundaryFinder::Grapheme, value);
```
Nerixyz: The `QTextBoundaryFinder` constructor can take a `QStringView` directly.

src/messages/AsciiArt.cpp
```diff
@@ -0,0 +1,66 @@
...
+    while (true)
+    {
+        const auto end = boundaries.toNextBoundary();
+        if (end == -1)
+        {
+            break;
+        }
+
+        const auto cluster = QStringView{value}.sliced(start, end - start);
+        if (std::ranges::any_of(cluster.toUcs4(), isArtCodePoint))
```
Nerixyz: We don't have a QStringIterator, but this can use the [same logic](https://github.com/qt/qtbase/blob/e67d1b219d2a40f62bee251c87f1dff0ce5b04c2/src/corelib/text/qstringiterator_p.h#L137-L149) to avoid the temporary list.

Something like
```cpp
const char16_t *pos = str.utf16();
const char16_t *end = str.utf16() + str.size();
for (;pos != end;++pos) {
    QChar uc = *pos;
    char32_t u = [&] {
        if (uc.isSurrogate()) {
            if (uc.isHighSurrogate() && pos + 1 != end && pos[1].isLowSurrogate()) {
                return QChar::surrogateToUcs4(uc, *(++pos));
            }
            return QChar::ReplacementCharacter;
        } 
        return uc.unicode();
    }();

    // ifArtCodePoint(u)
}
```

src/messages/AsciiArt.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+}  // namespace
+
+bool isAsciiArt(const QString &content)
+{
+    qsizetype artCells = 0;
+
+    for (const auto &segment :
+         QStringView{content}.split(u' ', Qt::SkipEmptyParts))
```
Nerixyz:
```suggestion
    for (const auto segment : content.tokenize(u' ', Qt::SkipEmptyParts))
```

src/messages/AsciiArt.hpp
```diff
@@ -0,0 +1,13 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <QString>
+
+namespace chatterino {
+
+bool isAsciiArt(const QString &content);
```
Nerixyz: This can take a `QStringView`.

src/messages/Message.hpp
```diff
@@ -50,6 +50,7 @@ struct Message {
     QString id;
     QString searchText;
     QString messageText;
+    bool isAsciiArt{};
```
Nerixyz: Could this be a message flag?

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -577,8 +585,13 @@ bool MessageLayoutContainer::fitsInLine(qreal width) const
...
 qreal MessageLayoutContainer::remainingWidth() const
 {
-    return (this->width_ - int(MARGIN.left() * this->scale_) -
-            int(MARGIN.right() * this->scale_) -
+    auto contentWidth = this->width_ - int(MARGIN.left() * this->scale_) -
+                        int(MARGIN.right() * this->scale_);
+    if (this->asciiArtWidth_)
+    {
+        contentWidth = std::min(contentWidth, *this->asciiArtWidth_);
+    }
```
Nerixyz: Couldn't we initialize the `width_` differently if the message is ascii art? Then we don't have this special case here and wouldn't need to store the ascii art width.

Nerixyz (commented):
A snapshot test for this would be great. Check https://github.com/Chatterino/chatterino2/blob/05d6b2afa2fddd79df5102467a88ebedc57dd63e/tests/snapshots/IrcMessageHandler/clearchat-stack-always.json for an example that customizes the settings and [Modifying message building](https://github.com/Chatterino/chatterino2/tree/05d6b2afa2fddd79df5102467a88ebedc57dd63e/tests#modifying-message-building) on how to update the tests.

3r01:
Thank you @Nerixyz, I have made those changes.

3r01:
Applied clang-tidy suggestions

src/messages/layouts/MessageLayoutContainer.cpp
```diff
@@ -28,6 +28,8 @@ using namespace chatterino;
 
 constexpr QMargins MARGIN{8, 4, 8, 4};
 constexpr qreal COMPACT_EMOTES_OFFSET = 4;
+// Target width used to match Twitch's desktop chat layout.
```
pajlada:
```suggestion
/// Target width used to match Twitch's desktop chat layout.
```

using `///` makes it a proper "document" comment that will show up in people's IDEs

src/messages/MessageFlag.hpp
```diff
@@ -80,6 +80,7 @@ enum class MessageFlag : std::int64_t {
     Announcement = (1LL << 44),
     /// Notifications that do not yet have special handling/categorization
     UncategorizedNotification = (1LL << 45),
+    AsciiArt = (1LL << 46),
```
pajlada, line 84: Add a comment here (`///`) describing what this flag means on a message

pajlada (changes requested):

3r01:
Thanks @pajlada, I have made those changes.

pajlada (approved):

pajlada:

PR #7203, opened 2026-08-30
fix: make subscriptions a single message

pajlada:


src/messages/MessageBuilder.hpp
```diff
@@ -284,7 +288,8 @@ class MessageBuilder
                        bool trimSubscriberUsername);
     void parseMessageID(Communi::TagsRef tags);
     /// Parses most of them message flags based on the given tags
```
Wissididom:
```suggestion
    /// Parses most of the message flags based on the given tags
```
I'm not sure if that should be changed since it seems somewhat unrelated, but it's something I noticed.

src/messages/MessageBuilder.cpp
```diff
@@ -1143,11 +1141,12 @@ void MessageBuilder::appendOrEmplaceText(const QString &text,
         back->getFlags() != MessageElementFlag::Text ||  //
```
Nerixyz:
```suggestion
        back->getFlags() != messageFlags ||  //
```

The elements don't get combined right now.

PR #7216, opened 2026-09-01
fix: prevent duplicate live notifications from /watching

### LINKED ISSUE #7164: Live notifications trigger when channel is opened in `/watching`
### Checklist

### Describe your issue

Steps to reproduce:

1. Add channel `c` to the live notifications.
2. (disable suppress live notifications on startup - makes it easier to debug when launching/closing)
3. Remove all tabs with `c`.
4. Add a `/watching` tab.
5. Restart Chatterino (if you previously had a tab with `c`).
6. Wait for `c` to go live.
7. Observe a live notification.
8. Open `c` in the browser.
9. ⚡ Observe another live notification.

We shouldn't've sent the second notification. The reason we did is that we create a new `TwitchChannel` when the `/watching` channel changes to `c`. There, we fetch the live status and issue a notification.

### Screenshots

_No response_

### OS and Chatterino Version

Chatterino 2.5.5 DEBUG (commit [491cda1a5](https://github.com/Chatterino/chatterino2/commit/491cda1a5) modified) built with Qt 6.12.0, MSVC 195136248

3r01:
Fixes #7164.

> Live notifications trigger when channel is opened in `/watching`
>
> Steps to reproduce:
> 
> 1. Add channel `c` to the live notifications.
> 2. (disable suppress live notifications on startup - makes it easier to debug when launching/closing)
> 3. Remove all tabs with `c`.
> 4. Add a `/watching` tab.
> 5. Restart Chatterino (if you previously had a tab with `c`).
> 6. Wait for `c` to go live.
> 7. Observe a live notification.
> 8. Open `c` in the browser.
> 9. ⚡ Observe another live notification.
> 
> We shouldn't've sent the second notification. The reason we did is that we create a new `TwitchChannel` when the `/watching` channel changes to `c`. There, we fetch the live status and issue a notification.

Fixed by tracking the last notified stream ID per channel and ignoring subsequent duplicate calls.

Added a test that covers various possible state changes, including the one described in the original issue.


src/controllers/notifications/NotificationController.hpp
```diff
@@ -74,6 +75,9 @@ class NotificationController final
     /// Channels are identified by their login name (case insensitive).
     std::map<QString, FakeChannel, QCompareCaseInsensitive> fakeChannels_;
 
+    /// Maps channel IDs to their most recently notified stream IDs
+    std::map<QString, QString> lastNotifiedStreamIds_;
```
pajlada, line 79: Stream IDs are unique, so this can be an unordered set instead of a map.

Additionally, should there be some mechanism that removes a stream from this list if we know the stream has gone offline? I'm thinking to prevent an ever-growing list of stream IDs.

pajlada (changes requested):

src/controllers/notifications/NotificationController.hpp
3r01: > Additionally, should there be some mechanism that removes a stream from this list if we know the stream has gone offline? I'm thinking to prevent an ever-growing list of stream IDs.

This was why I chose a map, this way the growth is capped at the number of monitored channels. I believe also that the offline event doesn't carry a stream ID, so you need to track stream IDs to channel IDs somewhere to make removal work.

pajlada (approved):

Issue #7186, opened 2026-08-19
Add view support for Twitch's gifs feature


Mm2PL:
### Checklist

- [x] I'm reporting a problem with Chatterino
- [x] I've verified that I'm running **the most** recent nightly build or stable release
- [x] I've looked for my problem on the [wiki](https://wiki.chatterino.com/Help/)
- [x] I've searched the [issues and pull requests](https://github.com/Chatterino/chatterino2/issues?q=) for similar looking reports

### Describe your issue

Twitch recently added an a/b tested feature where Tier 2 and Tier 3 subscribers can post GIFs. It seems to work essentially as emotes but there is no list of acceptable images and they need to be loaded after the message is received.

As of 2026-08-19, the message is as follows:
```
@badge-info=subscriber/30;badges=broadcaster/1,subscriber/0;color=#033700;display-name=TwitchDev;emotes=;first-msg=0;flags=;gifs=0-33|joSNxeswxuc74Juo8X|https://media4.giphy.com/media/joSNxeswxuc74Juo8X/giphy.gif?cid=095d7a5dzizsiwgabonagkmigggv8v1spfai91ac3x0dsiy0&ep=v1_gifs_trending&rid=giphy.gif&ct=g;id=401abf17-7e99-45d6-9bdf-43934e839327;mod=0;returning-chatter=0;room-id=12826;subscriber=1;tmi-sent-ts=1783632907018;turbo=0;user-id=141981764;user-type= :twitchdev!twitchdev@twitchdev.tmi.twitch.tv PRIVMSG #twitch :[Y A Y Yes GIF by Djemilah Birnie]
```
Formatted:
```
@badge-info=subscriber/30;
badges=broadcaster/1,subscriber/0;
color=#033700;
display-name=TwitchDev;
emotes=;
first-msg=0;
flags=;
gifs=0-33|joSNxeswxuc74Juo8X|https://media4.giphy.com/media/joSNxeswxuc74Juo8X/giphy.gif?cid=095d7a5dzizsiwgabonagkmigggv8v1spfai91ac3x0dsiy0&ep=v1_gifs_trending&rid=giphy.gif&ct=g;
id=401abf17-7e99-45d6-9bdf-43934e839327;
mod=0;
returning-chatter=0;
room-id=12826;
subscriber=1;
tmi-sent-ts=1783632907018;
turbo=0;
user-id=141981764;
user-type=
 :twitchdev!twitchdev@twitchdev.tmi.twitch.tv 
[...]

PR #7218, opened 2026-09-01
refactor: Message tokenization

Nerixyz:
I'm not sure about this yet. It's not really required for #7186, but I think it could help to separate the concern of parsing vs adding the actual elements.

The idea is to have a kind-of generator/iterator that generates `text | emote | emoji` events that the message builder then consumes. This is a push-based model where the tokenizer pushes events into the visitor.

There's only one snapshot test that changed - and I think for the better. The `&` in `&Kappa` had a trailing space set, but it shouldn't've, because it's next to the emote.


Nerixyz:
I'll wait with this, but I still think it's worth doing.

PR #7220, opened 2026-09-02
fix: store `length` for Twitch specials

Nerixyz:
Storing a length instead of an `end` makes more sense. We sometimes move the specials (e.g. in ignores). Both can be derived from each other.

*This is part of a stack*
1. #7219
2. #7220 ← this one
3. #7221
4. #7222
5. #7223

### DIFF
--- a/src/common/websockets/WebSocketPool.hpp
+++ b/src/common/websockets/WebSocketPool.hpp
@@ -9,6 +9,7 @@
 #include <QUrl>
 
 #include <memory>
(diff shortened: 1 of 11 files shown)


pajlada:
I think it would be clearer if we went for `start` and `length` instead of `start` and `end`, if we're going to be modifying this code.

pajlada (approved):

Mm2PL (approved):

PR #7222, opened 2026-09-02
feat: add Twitch gifs as links

Nerixyz:
Adds Twitch gifs as another special item besides emotes. Right now, they're added as links. I'll add an image version in a followup.

Closes #7186.

*This is part of a stack*
1. #7219
2. #7220
3. #7221
4. #7222 ← this one
5. #7223

PR #7223, opened 2026-09-02
feat: add Twitch gifs as images

Nerixyz:
This adds a setting to show Twitch GIFs as an inline image (default: on). It's similar to how it's shown in web chat: 

Code_2026-09-02_15-23-53

*This is part of a stack*
1. #7219
2. #7220
3. #7221
4. #7222
5. #7223 ← this one

### DIFF
--- a/docs/chatterino.d.ts
+++ b/docs/chatterino.d.ts
@@ -536,6 +536,7 @@ declare namespace c2 {
         LowercaseLinks = 0,
         RepliedMessage = 0,
         ReplyButton = 0,
+        TwitchGif = 0,
         Default = 0,
     }
 
--- a/docs/lua-meta/globals.lua
+++ b/docs/lua-meta/globals.lua
@@ -846,6 +846,7 @@ c2.MessageElementFlag = {
     EmoteImage = 0,
     EmoteText = 0,
     Emote = 0,
+    TwitchGif = 0,
     ChannelPointReward = 0,
     ChannelPointRewardImage = 0,
     BitsStatic = 0,
--- a/src/messages/MessageBuilder.cpp
+++ b/src/messages/MessageBuilder.cpp
@@ -2083,14 +2083,32 @@ void MessageBuilder::addTwitchGif(const QString &id, QStringView originalText)
 {
     auto original = originalText.toString();
     QString link = u"https://i.giphy.com/" % id % u".webp";
-    auto *el = this->emplace<LinkElement>(
-        LinkElement::Parsed{
-            .lowercase = original,
-            .original = original,
-        },
-        link, MessageElementFlag::Text, MessageColor::Link);
+    if (getSettings()->showTwitchGifs)
+    {
+        ImageSet set{
+            Image::fromUrl(
+                Url{u"https://media4.giphy.com/media/" % id % u"/100.webp"},
+                1.0, {100, 100}),
+            Image::fromUrl(
+                Url{u"https://media4.giphy.com/media/" % id % u"/200.webp"},
+                0.5, {200, 200}),
+        };
+        this->emplace<LinebreakElement>(MessageElementFlag::TwitchGif);
+        this->emplace<ScalingImageElement>(set, MessageElementFlag::TwitchGif)
+            ->setLink(Link{Link::Url, link})
+            ->setTooltip(original.toHtmlEscaped());
+    }
+    else
+    {
+        auto *el = this->emplace<LinkElement>(
+            LinkElement::Parsed{
+                .lowercase = original,
+                .original = original,
+            },
+            link, MessageElementFlag::Text, MessageColor::Link);
 
-    getApp()->getLinkResolver()->resolve(el->linkInfo());
+        getApp()->getLinkResolver()->resolve(el->linkInfo());
+    }
 }
 
 bool MessageBuilder::isEmpty() const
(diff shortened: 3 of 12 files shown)


src/controllers/ignores/IgnoreController.cpp
```diff
@@ -204,40 +204,39 @@ bool isIgnoredMessage(IgnoredMessageParameters &&params)
...
+            twitchSpecials.begin(), twitchSpecials.end(),
             [pos, len](const auto &item) {
                 // returns true for emotes outside the range
                 return !((item.start >= pos) && item.start < (pos + len));
             });
-        std::vector<TwitchEmoteOccurrence> emotesInRange(it,
-                                                         twitchEmotes.end());
-        twitchEmotes.erase(it, twitchEmotes.end());
-        return emotesInRange;
+        std::vector<TwitchSpecialOccurrence> specialsInRange(
```
3r01: If an ignore replacement starts inside a gif placeholder then the gif isn't removed and keeps its old length, so the tokenizer can consume the wrong range

```suggestion
    auto removeSpecialsInRange = [&twitchSpecials](SizeType pos, SizeType len) {
        const auto rangeEnd = pos + len;

        auto it = std::partition(
            twitchSpecials.begin(), twitchSpecials.end(),
            [pos, len, rangeEnd](const auto &item) {
                const auto specialEnd = item.start + item.length;
                const bool intersects =
                    len == 0 ? item.start < pos && pos < specialEnd
                             : item.start < rangeEnd && pos < specialEnd;
                return !intersects;
            });

        std::vector<TwitchSpecialOccurrence> specialsInRange(
```

src/providers/twitch/TwitchIrc.cpp
```diff
@@ -17,9 +17,60 @@
...
+    }
+    if (from > to || std::cmp_less(from, messageOffset))
+    {
+        qCDebug(chatterinoTwitch) << "Out of bounds range:" << occurrence
+                                  << "offset:" << messageOffset;
+        return;
+    }
+    to -= messageOffset;
+    from -= messageOffset;
+    if (to >= codepointToUtf16Idx.size())
```
3r01: One off? `codepointToUtf16Idx` includes the message end as its last value, but `to + 1` is accessed below, so `to == size() - 1` can still pass here and then read past the end

```suggestion
    if (to >= codepointToUtf16Idx.size() - 1)
```

Nerixyz: Thanks, that's actually an issue in #7220 (applied with https://github.com/Chatterino/chatterino2/compare/37c3bbf4d72d55a737aeba01363fa5240a91e996..0060c2870eaf56c9bbd3a4e9e9ca424a16fff569).

src/controllers/ignores/IgnoreController.cpp
Nerixyz: I think this is an issue with the existing implementation (for emotes) as well, but hasn't surfaced, because we haven't used the `end`/`length` before. I opened #7224 to fix this.

3r01:
I think `QMovie::CacheNone` might be a better fit, `Image` decodes and stores every frame and rejects the image when the decoded size exceeds 20 MiB. I have been hitting that quite frequently with `200.webp` images. `100.webp` are about a quarter of the size but that doesn't mean that you won't hit the limit with a long enough gif. Also with `Image` every loaded gif keeps all of its frames in RAM, where `QMovie::CacheNone` doesn't.

There is a CPU tradeoff because `QMovie` decodes the frames during playback, but in my testing it's not costly at all, just you have to make sure that gifs that aren't on screen aren't being rendered.

I tried something like this

```cpp
struct AnimatedImagePlayback {
    QByteArray data;
    QBuffer buffer{&data};
    QMovie movie{&buffer};

    AnimatedImagePlayback(QByteArray data)
        : data(std::move(data))
    {
        buffer.open(QIODevice::ReadOnly);
        movie.setCacheMode(QMovie::CacheNone);
    }
};
```

then on load:

```cpp
auto playback = std::make_unique<AnimatedImagePlayback>(result.getData());

if (!playback->movie.isValid() || !playback->movie.jumpToFrame(0))
{
    return;
}

currentPixmap = playback->movie.currentPixmap();
frameDelay =
    std::max(GIF_FRAME_LENGTH, playback->movie.nextFrameDelay());
```

and advance it from the existing GIF timer:

```cpp
elapsed += GIF_FRAME_LENGTH;

while (elapsed >= frameDelay)
{
    elapsed -= frameDelay;

    if (!playback->movie.jumpToNextFrame())
    {
        break;
    }

    currentPixmap = playback->movie.currentPixmap();
    frameDelay = std::max(GIF_FRAME_LENGTH,
                          playback->movie.nextFrameDelay());
}
```

You can also track when it was last painted and stop advancing it once it has been off screen for a short while, just need to recreate the `QMovie` from the cached data when it comes back on screen

Nerixyz:
> `Image` decodes and stores every frame and rejects the image when the decoded size exceeds 20 MiB. I have been hitting that quite frequently with `200.webp` images.

That's what I anticipated in https://github.com/Chatterino/chatterino2/issues/7186#issuecomment-5349447324 too (except I had the wrong limit in mind).

I think your idea is better for a followup - I thought the stack should be relatively easy to review to get basic support for GIFs out fast. But using QMovie sounds like a decent way of doing this in the long run. At the end we may want a system similar to the image GC but for movies too. Maybe some parts can be shared there?

3r01:
GIFs are rendered smaller here than in web chat. Is that intentional? Twitch uses max-height 140px and max-width 100%

Nerixyz:
> GIFs are rendered smaller here than in web chat. Is that intentional? Twitch uses max-height 140px and max-width 100%

Yes

pajlada (approved):

PR #7230, opened 2026-09-04
fix: prevent input focus from remaining on another tab

3r01:
Fixes an issue where input focus can remain on a different tab after switching. Once in this state, switching tabs further does not recover the focus; you have to click the affected input.

This was difficult to pin down and reproduce. Here are some logs by @zneix describing the issue last year:

>#pajlada [18/10/2025] zneix: one thing I never understood about chatterino - what is the cause of a bug where if you switch tabs with keyboard or mouse, the focused split sometimes will remain on the "old" tab
>#pajlada [18/10/2025] zneix: I swear this existed even before I made my first commit
>#pajlada [18/10/2025] zneix: and I can clearly see that by the split header not having streamer name in blue, but its in white
>#pajlada [18/10/2025] zneix: its so hard to capture this. idk how to repro
>#pajlada [18/10/2025] zneix: pajlada https://kappa.lol/lcsT4M
>#pajlada [18/10/2025] zneix: maybe I've described this wrong, but essentially - sometimes I would switch tab and focus will never follow, but only to 1 certain split/tab; I then would have to click with my mouse on that split and issue goes away
>#pajlada [18/10/2025] zneix: I encounter this maybe ~once a week
>#pajlada [18/10/2025] zneix: zero idea what's a cause mm2pl, sorry
>#pajlada [18/10/2025] zneix: I've pressed chat with mouse to remedy the issue, now the problem is gone and idk how to bring it back
>#pajlada [18/10/2025] zneix: I would have to check but I believe it fixes itself if I click the input field (where my typed text appears before I send it)

The video zneix linked demonstrates the bug:

https://github.com/user-attachments/assets/de7cf1b3-95ca-4610-8153-d69688d43353

There are several paths that reproduce this issue. The simplest that I found is as follows:

1. Start chatterino with an existing config that has at least two tabs (called A, B here)
2. Do not visit tab B after startup
3. With A selected, deactivate chatterino by clicking another application
4. While the chatterino window is inactive, hover your mouse over the tab strip and scroll to tab B, then back to tab A
5. Alt+tab back into chatterino; the input on tab A should now be focused
6. Ctrl+tab to tab B
7. Observe that tab B's input box is not focused, and if you type something it will appear in tab A's input box instead
8. Observe also that no amount of ctrl+tab or alt+tab recovers focus onto tab B; as far as I can tell, you must click into it with your mouse

This happens because when restoring the saved layout, chatterino creates every tab with `addPage(false)`, and only calls `select(page)` for whichever tab was selected when the layout was saved. `Notebook::Item::selectedWidget` starts as `nullptr` for any restored tab you haven't visited yet. In the reproduction above, switching to and away from tab B while the window is inactive results in the page itself being remembered as its `selectedWidget`. When the tab is later selected, this widget is detected as not being a child of the page, but no fallback focus is applied and focus remains on tab A.

Fixed by falling back to focusing the page when its `selectedWidget` is not a child of that page.


zneix:
I wanna take a look at this later, thank you for mentioning me and also finding a way to reproduce the bugged state; I wasn't quite able to get that down myself. This is something that haunted me ever since I started using chatterino in early 2019

pajlada (approved):
Thanks, happy to merge once zneix has had a look / test

zneix (approved):
tested out the change. tysm for this fix, highly appreciated

PR #7231, opened 2026-09-05
fix: show streamlink message in selected window

4rneee:
Previously the message was always shown in the main window, even when opening something in streamlink while focusing a popup.


Nerixyz:
I think this PR itself is fine, but I do wonder what we should do if Chatterino is embedded. There, we still create a main window, but that's not shown. Maybe there should be a last selected split?

pajlada (approved):

PR #7237, opened 2026-09-06
fix: Show account and follow age in local time

### LINKED ISSUE #7236: "Following since" field in user popout does not follow the system timezone
### Checklist

### Describe your issue

The Following since XXXX-XX-XX field in the user popout does not correctly follow the system timezone. In my case, my system timezone is Brasília Time (UTC-3), but the date appears to follow UTC instead.

### Screenshots

Image

In the user popout, the date shows 2026-09-06, while my system date is 2026-09-05.

### OS and Chatterino Version

Chatterino 2.5.5 (commit [1bc6e5aa](https://github.com/Chatterino/chatterino2/commit/1bc6e5aa)) built with Qt 6.7.1, MSVC 194435223, Crashpad

Nerixyz:
We show all user-facing date/times in local time (e.g. the timestamp in chat). As pointed out in #7236, we didn't do this in usercards. This PR fixes that for the creation date and the followage.

Fixes #7236.


src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -966,12 +966,13 @@ void UserInfoPopup::updateUserData()
 
         this->setWindowTitle(TEXT_TITLE.arg(
             user.displayName, this->underlyingChannel_->getName()));
-        this->ui_.createdDateLabel->setText(
-            TEXT_CREATED.arg(user.createdAt.section("T", 0, 0)));
+        auto createdAt =
+            QDateTime::fromString(user.createdAt, Qt::ISODateWithMs);
+        auto createdStr = createdAt.toLocalTime().toString(Qt::ISODate);
```
pajlada: Left is old, right is your PR
image

The format shouldn't change


```suggestion
        auto createdStr = createdAt.toLocalTime().date().toString(Qt::ISODate);
```

src/widgets/dialogs/UserInfoPopup.cpp
```diff
@@ -966,12 +966,13 @@ void UserInfoPopup::updateUserData()
...
+        this->ui_.createdDateLabel->setText(TEXT_CREATED.arg(createdStr));
         this->ui_.createdDateLabel->setToolTip(
-            formatLongFriendlyDuration(
-                QDateTime::fromString(user.createdAt, Qt::ISODateWithMs),
-                QDateTime::currentDateTimeUtc()) +
+            formatLongFriendlyDuration(createdAt,
+                                       QDateTime::currentDateTimeUtc()) +
```
pajlada, line 975: This now compares a local QDateTime with UTC, does `formatLongFriendlyDuration` handle this?

pajlada (changes requested):

src/widgets/dialogs/UserInfoPopup.cpp
Nerixyz: Oh right I did this while testing

src/widgets/dialogs/UserInfoPopup.cpp
Nerixyz: > This now compares a local QDateTime with UTC

The parsed datetime is in UTC already, as it ends with `Z`.

pajlada (approved):

PR #7248, opened 2026-09-06
fix: use message timestamps for R9K similarity check

3r01:
The R9K filter currently checks the delay between messages using the time they were parsed rather than their timestamps. This means that messages loaded from Recent Messages history can be erroneously treated as close because they're all parsed together. `parseTime` also only contains the time of day, so the check can give incorrect results across midnight.

Fixed by comparing `serverReceivedTime` from both messages instead.

Added a test covering this.

### DIFF
--- a/src/messages/MessageSimilarity.cpp
+++ b/src/messages/MessageSimilarity.cpp
@@ -68,7 +68,7 @@ float inMessages(const MessagePtr &msg, const T &messages)
          messages | std::views::reverse |
              std::views::take(getSettings()->hideSimilarMaxMessagesToCheck))
     {
-        if (prevMsg->parseTime.secsTo(QTime::currentTime()) >=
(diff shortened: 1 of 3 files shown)


3r01:
If this and #7244 are merged, then all consumers of `parseTime` will have been removed (aside from the plugin API). At that point I think it would make sense to consider removing `parseTime` entirely, or else replacing it with a full timestamp, to avoid similar bugs in future.

Nerixyz:
> If this and #7244 are merged, then all consumers of `parseTime` will have been removed (aside from the plugin API). At that point I think it would make sense to consider removing `parseTime` entirely, or else replacing it with a full timestamp, to avoid similar bugs in future.

I don't think there's any use for `parseTime` - even in plugins. Removing `parseTime` and renaming `serverReceivedTime` to `timestamp` is something we could do.

pajlada (approved):

PR #7253, opened 2026-09-08
fix: prevent use after free in settings dialog

3r01:
The settings dialog's parent is set the first time it is opened, so if it is first opened from a popup window, closing that popup also deletes the settings dialog but leaves the static pointer dangling.

Repro:

1. Restart chatterino
2. Open a popup window
3. While the popup is focused, open settings via hotkey (default `ctrl+P`)
4. Close settings and the popup
5. Open settings from the main window
6. Observe that chatterino crashes

Fixed by using a `QPointer` for the settings dialog, and preventing the timer used to open it from running if its parent has been deleted


Nerixyz (commented):
It feels weird to me that we parent the settings on a popup in the first place. There's always a main window. Why not use this as the parent?

One reason against it is that we may embed Chatterino in other apps, and having a main window there is weird (but that's how it is today).

The `QPointer` change is fine as is.

3r01:
> It feels weird to me that we parent the settings on a popup in the first place. There's always a main window. Why not use this as the parent?

You are right. I hadn't considered it, I was just following a train of thought about stale ownership after #7230 and #7252. I will change it to use the main window as parent.

pajlada (approved):

PR #7254, opened 2026-09-08
feat: Fill remaining space in horizontal tab layout

Nerixyz:
In #6153 I described this. There, I thought this would require more work.

| Off | On |
|---|---|
| <video src="https://github.com/user-attachments/assets/56b16cbd-162e-4ec9-a1a1-d70d9c87f212"> | <video src="https://github.com/user-attachments/assets/1694d784-7539-458a-b964-cabfcbd40e72"> |

The main change to get this to work was splitting the computation of the minimum width and the actual repositioning. Lines first lay out as before, but when we wrap, we extend each tab with its fraction of the remaining width.

Closes #6153.


pajlada:
https://github.com/user-attachments/assets/68705848-9d6d-41d1-8515-235e83a40bbc

Seems to not fully work as expected for me on Linux, might be related to the settings & user buttons?

pajlada (approved):
Nice, works as expected for me now. Sometimes a bit jumpy but no one resizes Chatterino like this

https://github.com/user-attachments/assets/744e5f27-0e27-4c87-ad7e-94cce6b3fa74

PR #7256, opened 2026-09-08
fix: count message length in unicode codepoints

3r01:
Some emojis count twice towards the input length because `QString::length()` is used, which counts UTF-16 code units rather than unicode codepoints.

image

The indicator turns red and reads 500 when the input contains 250 of these, but twitch allows you to send 500.

Fixed by counting unicode codepoints instead.


pajlada (approved):
Works well in my testing.
Funnily enough, a wall of lions count as ascii art

PR #7259, opened 2026-09-09
refactor: extract image frame storage

3r01:
Extract frame storage into `Frames::CachedFrames` behind a `Frames::Storage` interface in preparation for adding a dynamically decoded storage type

Added tests

### DIFF
--- a/src/messages/Image.cpp
+++ b/src/messages/Image.cpp
@@ -26,6 +26,8 @@
 #include <QTimer>
 
 #include <atomic>
+#include <numeric>
+#include <utility>
 
 // Duration between each check of every Image instance
 const auto IMAGE_POOL_CLEANUP_INTERVAL = std::chrono::minutes(1);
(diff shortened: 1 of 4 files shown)


src/messages/Image.cpp
```diff
@@ -119,15 +170,72 @@ int64_t Frames::memoryUsage() const
...
+        if (!playback.movie.jumpToNextFrame())
+        {
+            playback.finished = true;
+            this->gifTimerConnection_.disconnect();
+            return;
+        }
+        const auto decodeFinished = std::chrono::steady_clock::now();
+        const auto delay = playback.movie.nextFrameDelay();
+
+        // QMovie subtracts decoding time from the remaining delay.
```
Nerixyz: It may be easier to use [`QImageReader`](https://doc.qt.io/qt-6/qimagereader.html) directly - that seems to be what `QMovie` does under the hood regardless: https://github.com/qt/qtbase/blob/dd7b529dd3e7b855c966841de460a53cba4f8283/src/gui/image/qmovie.cpp#L337-L386

I think for our purpose, we can assume `jumpToImage(0)` always succeeds. Then, `nextImageDelay` is always static. Could also read all delays into a list at the start if that makes it simpler.

src/messages/Image.hpp
```diff
@@ -48,23 +50,29 @@ class Frames
...
+    std::optional<QSize> size() const;
 
 private:
+    struct Movie;
+
     int64_t memoryUsage() const;
+    void advanceMovie();
     void processOffset();
     QList<Frame> items_;
+    std::unique_ptr<Movie> movie_;
```
Nerixyz: I wonder if it's worth to let `Frames` or its data be polymorphic (not for performance, but code style). Because right now, most things branch on `if (movie_)`. Could also try `std::variant`. Then there would be a `CachedFrames(Data)` and `DynamicFrames(Data)`.

src/messages/Image.hpp
```diff
@@ -126,6 +137,7 @@ class Image : public std::enable_shared_from_this<Image>
     const QSize expectedSize_{16, 16};
     std::atomic_bool empty_{false};
 
+    const bool useMovie_{false};
```
Nerixyz: In the future, we could also "detect" this. There are some 3x and 4x 7TV emotes that exceed the memory limit, which results in tooltips being smaller. Dynamically decoding them could be a solution to this.

3r01: Yes, it is possible, but they would lose sync with emotes that render via cached frames. Keeping sync with cached frames is more headache than it is worth in my opinion. Qt's gif/webp readers don't support jumping to an arbitrary frame, and I didn't like the idea of having to play catch-up with decoding (or any of the other solutions I could think of to recover sync once it is lost).

I suppose sync isn't a huge problem for tooltips anyway?

src/messages/Image.hpp
3r01: I don't like branching on `if (movie_)` either. I had tried to abstract previously but ended up with a very bloated solution, so discarded it. I will give it another try with your suggestions in mind.

src/messages/Image.cpp
3r01: You are right, I will switch to `QImageReader`.

pajlada (changes requested):
Please make this a refactor PR instead where you don't implement the QMovie-style logic/dynamic frames stuff, but instead add tests & refactor the Frames storage into `Frames::Storage` and `Frames::CachedFrames` (so leave `Frames::DynamicFrames` out of this PR.

src/messages/Image.hpp
```diff
@@ -50,17 +50,15 @@ class Frames
     void clear();
     bool empty() const;
     bool animated() const;
-    void advance();
     std::optional<QPixmap> current() const;
-    std::optional<QPixmap> first() const;
+    std::optional<QSize> size() const;
```
pajlada: Rename to something like `imageSize` or `frameSize`, with a comment, something along the lines of "Returns the size of the first frame if it has been calculated/loaded"

src/messages/Image.cpp
```diff
@@ -34,13 +36,149 @@ const auto IMAGE_POOL_IMAGE_LIFETIME = std::chrono::minutes(10);
...
+    virtual bool empty() const = 0;
+    virtual bool animated() const = 0;
+    virtual void start(GIFTimer *timer) = 0;
+    virtual std::optional<QPixmap> current() const = 0;
+    virtual std::optional<QSize> size() const = 0;
+
+    pajlada::Signals::Connection gifTimerConnection;
+};
+
+struct Frames::CachedFrames : Storage {
```
pajlada, line 59: Add a comment for what this struct does, something along the lines of "Loads and stores all frames in memory"

pajlada (changes requested):

src/messages/Image.cpp
```diff
@@ -34,13 +36,149 @@ const auto IMAGE_POOL_IMAGE_LIFETIME = std::chrono::minutes(10);
...
+    Storage &operator=(Storage &&) = delete;
+
+    virtual int64_t memoryUsage() const = 0;
+    virtual bool empty() const = 0;
+    virtual bool animated() const = 0;
+    virtual void start(GIFTimer *timer) = 0;
+    virtual std::optional<QPixmap> current() const = 0;
+    virtual std::optional<QSize> size() const = 0;
+
+    pajlada::Signals::Connection gifTimerConnection;
```
Nerixyz: You can use a `ScopedConnection` here - this will automatically disconnect when destroyed.

src/messages/Image.hpp
Nerixyz: > I suppose sync isn't a huge problem for tooltips anyway?

Not a big problem in my opinion.

tests/src/Image.cpp
```diff
@@ -0,0 +1,99 @@
...
+    EXPECT_FALSE(frames.animated());
+    EXPECT_FALSE(frames.current());
+    EXPECT_FALSE(frames.size());
+}
+
+TEST(Image, StaticCachedFrame)
+{
+    ImageApplication app;
+    detail::Frames frames(QList<detail::Frame>{
+        {.image = makePixmap({1, 1}, Qt::red), .duration = 20}});
```
Nerixyz:
```suggestion
        {.image = makePixmap({1, 1}, Qt::red), .duration = 20},});
```
And reformat, same for the others.

pajlada (approved):

PR #7262, opened 2026-09-10
feat: add mouse button hotkey support

Leafyzito:
Adds mouse button support to the hotkey manager so users can bind Mouse Back/Forward (and other extra buttons) to any action.

Left, Right, and Middle clicks cannot be bound. Mouse bindings ignore keyboard modifiers.

AI generated/assisted


Nerixyz:
It will be easier to review if you split this into multiple PRs. For example adding stuff that doesn't touch mouse input directly separately.

Mm2PL:
hell nah i am not reviewing a slopped +1000-140 PR

PR #7266, opened 2026-09-11
fix: 1px border rendering around text input at fractional dpi scaling

### LINKED ISSUE #6091: SplitInput border partially disappearing when resizing
### Checklist

### Describe your issue

When resizing the window, the border of the SplitInput widget sometimes partially disappears

### Screenshots

https://github.com/user-attachments/assets/9446c01f-5f1d-4594-82e3-e6eac8762030

### OS and Chatterino Version

Chatterino 2.5.2 (commit 95eee280) built with Qt 6.8.2 Running on Arch Linux, kernel: 6.13.7-arch1-1

3r01:
Fixes #6091.

> When resizing the window, the border of the SplitInput widget sometimes partially disappears

This happens on monitors with fractional dpi scaling applied.

~~Fixed by moving the borders to the widgets' stylesheets instead.~~ Fixed by drawing the borders as `QRect` instead of using the pen.

Tested on Windows 11 at 125% and 150% scaling.

The same issue seems to affect the borders around the edges of the application's windows as well, but I think it is a Windows issue because it also happens in other applications. Either way, I spent some time trying but wasn't able to find a solution.


Nerixyz:
Please avoid using style sheets. See [Use style sheets for prototyping](https://doc-snapshots.qt.io/qt6-dev/qtwidgets-styling-approaches.html#use-style-sheets-for-prototyping) and [Say No to Qt Style Sheets](https://www.kdab.com/say-no-to-qt-style-sheets/). We don't need a full QStyle here either.

The difference that the style sheet makes is that instead of using the pen of the painter to draw the border, it draws each border as an individual `QRect` and enables antialiasing (not sure if that's needed here).

Nerixyz:
Thank you!

pajlada (approved):

PR #7272, opened 2026-09-12
chore: Add backup functionality to window layout

Nerixyz:
Someone on Discord mentioned that their window layout reset occasionally. We already have some backup restoration functionality for settings, so this PR uses that for the window layout. Most of this PR is switching to `ExpectedStr` and passing the errors on.

When parsing command line arguments, we also check the window layout for the initial bounds. But here I opted to not do restoration, as it's only a hint to where the window should be placed.

### DIFF
--- a/src/common/Args.cpp
+++ b/src/common/Args.cpp
@@ -342,17 +342,19 @@ std::optional<WindowLayout> Args::makeCustomChannelLayout(
 
     // Load main window layout from config file so we can use the same geometry
     const QRect configMainLayout = [windowLayoutFile] {
-        const WindowLayout configLayout =
-            WindowLayout::loadFromFile(windowLayoutFile);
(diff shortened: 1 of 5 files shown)


src/common/Args.cpp
```diff
@@ -342,17 +342,19 @@ std::optional<WindowLayout> Args::makeCustomChannelLayout(
...
-            }
+                if (window.type_ != WindowType::Main)
+                {
+                    continue;
+                }
 
-            return window.geometry_;
+                return window.geometry_;
+            }
         }
```
pajlada, line 358: If we have an error message here, maybe we can print that still

pajlada (changes requested):
Just this small nit, otherwise looks good.
Testing it in the meantime.

EDIT: Testing went well

pajlada (approved):

PR #7275, opened 2026-09-13
feat: manual registration for browser integration in portable chatterino

3r01:
Chatterino skips registration for browser integration in portable mode, presumably because the registration writes to Windows registry.

Added a button in settings to manually register for browser integration. It only appears on Windows in portable mode.

image

### DIFF
--- a/src/singletons/NativeMessaging.cpp
+++ b/src/singletons/NativeMessaging.cpp
@@ -79,14 +79,16 @@ const Config CHROME{
 #endif
 };
 
-bool registerNmManifest([[maybe_unused]] const Paths &paths,
-                        const Config &config, const QJsonDocument &document)
+ExpectedStr<void> registerNmManifest([[maybe_unused]] const Paths &paths,
+                                     const Config &config,
+                                     const QJsonDocument &document)
 {
 #ifdef Q_OS_WIN
-    if (!writeManifestTo(paths.miscDirectory, u"."_s, config.fileName,
-                         document))
+    auto result =
+        writeManifestTo(paths.miscDirectory, u"."_s, config.fileName, document);
+    if (!result)
     {
-        return false;
+        return makeUnexpected(result.error());
     }
 
     QSettings registry(config.registryKey, QSettings::NativeFormat);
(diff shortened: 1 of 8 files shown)


src/singletons/NativeMessaging.cpp
```diff
@@ -79,21 +79,32 @@ const Config CHROME{
 #endif
 };
 
-void registerNmManifest([[maybe_unused]] const Paths &paths,
+bool registerNmManifest([[maybe_unused]] const Paths &paths,
```
pajlada: Return an ExpectedStr<void> instead, that way you can forward the error from writeManifestTo too.

src/singletons/NativeMessaging.cpp
```diff
@@ -240,22 +257,19 @@ namespace chatterino {
...
 
     QJsonDocument chromeManifest = buildChromeManifest(extensionIDs);
     QJsonDocument firefoxManifest = buildFirefoxManifest(extensionIDs);
 
-    registerNmManifest(paths, CHROME, chromeManifest);
-    registerNmManifest(paths, FIREFOX, firefoxManifest);
+    const bool chromeRegistered =
+        registerNmManifest(paths, CHROME, chromeManifest);
+    const bool firefoxRegistered =
+        registerNmManifest(paths, FIREFOX, firefoxManifest);
```
pajlada: Once `registerNmManifest` returns an error string, print here (qdebug probably since for many this will happen every time if they don't have that browser installed?) if something went wrong. No need to early circuit since we'll want to try both

pajlada (changes requested):
Just nits, overall this is nice

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -949,6 +950,27 @@ void GeneralPage::initLayout(GeneralPageView &layout)
...
+        layout.addButton("Register browser integration", [this] {
+            if (registerNmHost(getApp()->getPaths()))
+            {
+                QMessageBox::information(this, "Registration Successful",
+                                         "Browser integration registered.");
+            }
+            else
+            {
+                QMessageBox::warning(this, "Registration Failed",
+                                     "Failed to register browser integration.");
```
pajlada: This should be able to include the error from `registerNmHost` to inform the user what went wrong

src/singletons/NativeMessaging.cpp
```diff
@@ -186,20 +197,23 @@ Expected<void, WriteManifestError> writeManifestTo(QString directory,
...
-            << "Failed to open" << filename << "in" << directory;
-        return makeUnexpected(WriteManifestError::FailedToCreateFile);
+        return makeUnexpected(
+            QString(u"Failed to open "_s % filename % u" in "_s % directory));
+    }
+    const auto data = json.toJson();
+    if (file.write(data) != data.size() || !file.flush())
+    {
+        return makeUnexpected(
+            QString(u"Failed to write "_s % filename % u" in "_s % directory));
```
pajlada: No need to `_s` the string literals used in makeUnexpected

src/widgets/settingspages/GeneralPage.cpp
```diff
@@ -949,6 +950,27 @@ void GeneralPage::initLayout(GeneralPageView &layout)
         formatRichNamedLink(FIREFOX_EXTENSION_LINK, "Download for Firefox"));
 
 #ifdef Q_OS_WIN
+    if (Modes{getApp()->getArgs()}.isPortable)
```
pajlada: Just store modes in Application instead of reinitializing it here

pajlada (changes requested):

pajlada (approved):

PR #7280, opened 2026-09-13
refactor: rename serverReceivedTime to timestamp

3r01:
`serverReceivedTime` is a misleading name because, depending on the context, it is used to hold timestamps provided by twitch or that we generate locally, e.g. emote add/remove notices have their timestamps generated locally.

This is a breaking change for the plugin api, `server_received_time` -> `timestamp`. I'm not sure how you want to handle that, so tagging @Mm2PL and @Nerixyz for review. A compatibility alias came to mind but I didn't add one.

Discussion from #7248:

>If this and #7244 are merged, then all consumers of `parseTime` will have been removed (aside from the plugin API). At that point I think it would make sense to consider removing `parseTime` entirely, or else replacing it with a full timestamp, to avoid similar bugs in future.
>
>_Originally posted by @3r01 in https://github.com/Chatterino/chatterino2/issues/7248#issuecomment-5566889394_
            

>I don't think there's any use for `parseTime` - even in plugins. Removing `parseTime` and renaming `serverReceivedTime` to `timestamp` is something we could do.
>
> _Originally posted by @Nerixyz in https://github.com/Chatterino/chatterino2/issues/7248#issuecomment-5567319892_


src/messages/Message.cpp
```diff
@@ -153,8 +156,7 @@ QJsonObject Message::toJson() const
         {"channelName"_L1, this->channelName},
         {"usernameColor"_L1, this->usernameColor.name(QColor::HexArgb)},
         {"count"_L1, static_cast<qint64>(this->count)},
-        {"serverReceivedTime"_L1,
-         this->serverReceivedTime.toString(Qt::ISODate)},
+        {"serverReceivedTime"_L1, this->timestamp.toString(Qt::ISODate)},
```
Nerixyz: The JSON field has to be renamed too. It would be fine if you did that in a followup I think.

Nerixyz (commented):
I can see two approaches for the plugin API:

1. Just rename it (current state). Plugins are still in alpha state, and we mentioned that there will be breaking changes.
2. Handle `server_received_time` as a fallback for some time. Maybe with a log to warn users?

Either way, we should have some place where we document this.

src/messages/Message.cpp
3r01: I will do it here, I think these changes are all-or-nothing.

pajlada:
Last change of removing parseTime was also technically breaking. As Nerixyz mentioned, since plugins are considered not-yet-stable, we are free to break things as needed.

PR #7283, opened 2026-09-13
fix: show inline twitch gif tooltips without link previews enabled

3r01:
Twitch gifs as images currently don't have a tooltip if link previews are disabled. This is contrary to how existing inline images (emotes) work in chat. You may want to see the name of the gif without also having link previews enabled.

Fixed by showing tooltips their regardless of the link previews setting, provided the gif is displayed as an inline image. 

No change is made to gifs as links.


Nerixyz:
An issue with this is that we're now using the link resolver even though the user didn't enable it.

3r01:
> An issue with this is that we're now using the link resolver even though the user didn't enable it.

Are we? As far as I can see the text comes from the twitch message's placeholder text, there's no link resolver request.

Nerixyz:
Oh, I thought this intends to show the tooltip when gifs aren't inline.

3r01:
No, apologies if my wording wasn't clear. Currently the tooltips only work if the link previews setting is enabled, even when the gif is displayed inline.

I think inline gifs as images should show their tooltip like emotes do, and we already have the data for that. No change is made to gifs displayed as links.

PR #7284, opened 2026-09-13
fix: display special characters in twitch gif tooltips

3r01:
The gif title is html escaped in the tooltip but Qt displays it as plain text because it contains no html tag

Fixed by wrapping the escaped title in `<span>` so it renders correctly

Before:
image

After:
image


3r01:
Feedback from stream:

>this feels hacky, can we not tell Qt that it's plaintext?

We could, but the tooltip widget has no way to pass the text format down at the moment, it just always uses Qt::AutoText. If it is preferable to change the tooltip widget to allow passing a text format, then let me know and I will do it that way instead.

PR #7288, opened 2026-09-14
feat: add emote ignores

### LINKED ISSUE #867: Add an option to ignore emotes the same way you could in c1
The current solution requires me to type the emote twice AND add some invisible character as a replacement.
I really don't think this is user friendly.
I would like to be able to use it the same way i was able to in c1.
Just a simple line separated list.
Some kind of import button that opens a window to do that would be perfectly fine aswell.

3r01:
Fixes #867. (I think? it is not the exact same solution discussed there)

Ignoring an emote through the existing phrase ignores requires setting up a pattern and replacement. I think it would be more user friendly to allow an emote to be ignored through the context menu, so this PR adds that and an accompanying `Emotes` tab under the `Ignores` settings menu.

Ignored emotes are shown as text in chat. You can optionally use regex.

Replaced emotes still hold their tooltips and context menus, so that you can right click again in chat to remove a non-regex ignore, or hover to view the emote that was ignored.

Added tests

https://github.com/user-attachments/assets/8c3041af-4d5b-43eb-a5c2-9827cdf8fd24

image


Nerixyz:
Would you mind elaborating on why you chose to check for ignored emotes when laying out messages? One benefit is that changes to the ignores "update" past messages.

3r01:
> Would you mind elaborating on why you chose to check for ignored emotes when laying out messages? One benefit is that changes to the ignores "update" past messages.

That was the only reason. Having said that, I couldn't figure out a way to have layered emotes update when ignores change without a more significant refactor because they are grouped when the message is built.

Nerixyz:
Messages built by someone else (plugins) are another aspect. By doing ignores during layout, they get emote ignores "for free". Now, the alternative would be to do this in the message builder ([what Chatterino 1 did](https://github.com/fourtf/chatterino/blob/d9ebd668e34415e590d082b96bf889c45f882243/Chatterino.Common/Message.cs#L527-L532)). Not sure which is better.

PR #7289, opened 2026-09-15
feat: add command completion popup

3r01:
Things that can complete in Chatterino can use a popup menu when typed with a symbol prefix; `@` for usernames and `:` for emotes.

Commands currently do not fit this pattern when they are prefixed with `/` or `.`, so this PR adds the option. It is on by default. The checkbox is in the Advanced section:

image

I assume the reason that it was not added previously is that there may be a conflict in behaviour with tab cycling. In this PR we handle that by handing off the popup's completion candidates to the tab completer, so that you can continue tab cycling through the list after the popup closes.

image


3r01:
Leaving this as draft pending the outcome of #7290 because that will change the implementation if merged.

PR #7294, opened 2026-09-16
feat: add hotkeys to select reply targets

3r01:
Adds hotkeys to select a reply target.

Defaults to enabled, `Ctrl+Up` to select older and `Ctrl+Down` to select newer. 

These keys are usually used for paragraph up/down in text editing, but twitch input doesn't accept paragraphs, and we already override up/down keys which would usually be for navigating lines.

The selected message is outlined, and this outline only applies to replies initiated by these hotkeys. The view scrolls to show a selected reply if it is off-screen. It skips messages you can't reply to (e.g. system messages). When you are already at the newest message, pressing the "newer" hotkey cancels the reply.

If you start navigating while at the bottom of the view, you will be returned there after the reply ends (unless you scroll manually during the process, then your manually scrolled location will be respected).

https://github.com/user-attachments/assets/c901c6c3-b7a0-4c69-8d75-70b59d7c6d5c


3r01:
I had thought about making this system into a more general keyboard navigation system, so that you could navigate/select messages and then use keyboard shortcuts to perform different actions on them. Initiating a reply would be one of those actions (i.e. `Ctrl+Up` to the message you want, then `R` to initiate a reply and re-focus the input). 

I am not sure if that is preferable. The list of actions that came to mind was quite short: copy message on `C`, reply on `R`.
Moderation actions would fit as well, but then we would have to add a hotkey field to the "moderation buttons" and "user timeout buttons" settings tables (or add a separate list under the hotkeys menu, but then either you connect the configurations - how? - or you don't, and the user has two separate lists to manage), and it's suddenly becoming way more complex than is warranted for a niche feature.

Mm2PL:
>I had thought about making this system into a more general keyboard navigation system, so that you could navigate/select messages and then use keyboard shortcuts to perform different actions on them. Initiating a reply would be one of those actions (i.e. `Ctrl+Up` to the message you want, then `R` to initiate a reply and re-focus the input). 

This as an idea has floated around for a while. But the idea if adding 
an entire mode system* or half-assing a mode system is weird.

*Mode as in modal editor, some state that changes the meaning of keys 
pressed. Like vim: in Normal mode, `w` moves the cursor one word to the 
right, in Insert mode `w` inserts the letter.

> I am not sure if that is preferable. The list of actions that came to mind as quite short: copy message on `C`, reply on `R`.
It'd be a shame if this couldn't be customized but maybe we could reuse 
the right-click menu.

3r01:
> This as an idea has floated around for a while. But the idea if adding an entire mode system* or half-assing a mode system is weird.

Yes, I could see that it was quickly becoming a large and complex feature if I went down that route, so I kept it to hotkeys for selecting a reply target.

>maybe we could reuse the right-click menu. 

If we did want a more general system, I think the context menu would be a good way to do it. You can already navigate most things by keyboard. For message actions, the only missing piece is a way to select a message and open its context menu without using the mouse.

Another option would be not to enter a mode at all and keep everything as configurable hotkeys. There could be hotkeys for selecting an older or newer message (as added here for replies), followed by hotkeys that act on the selected message. For example, you might assign `reply to selected message`, `copy selected message`, `open selected message context menu`. You could assign them to whatever keys you choose as long as they don't conflict with existing keys. That way ordinary keys would keep their behaviour and the input can remain focused, so you don't really enter a different "mode" at any point.

PR #7296, opened 2026-09-16
feat: cancel replies with escape

3r01:
Escape can be used to cancel or close in some contexts, like completion popups, usercards, emote picker, quick switcher, search, etc., but replies do not follow the same semantics.

When replying, you can now press `Escape` to cancel. 

If both a reply and completion popup are active, the completion popup consumes `Escape` first.


Mm2PL:
I don't think we want a new hardcoded shortcut.

3r01:
> I don't think we want a new hardcoded shortcut.

Well, I could make it a hotkey if that is preferable. Still, it seems odd to me that you can use Esc in most places that you would expect to cancel or close, but not replies. It is hardcoded for similar input semantics in `BasePopup`, `GenericListView`, `BackwardsSearchLineEdit`.

PR #7299, opened 2026-09-17
feat: nicknames by userid

### LINKED ISSUE #4398: Possible to have two nicknames for the same user.
### Checklist

### Describe your issue

Similar to https://github.com/Chatterino/chatterino2/issues/2497 you can have two nicknames for the same person.

I assume this same fix that is being work on for https://github.com/Chatterino/chatterino2/issues/2497 could be applied to nicknames as well. ( https://github.com/Chatterino/chatterino2/pull/4322 )

### Screenshots

image
image

### OS and Chatterino Version

Chatterino Nightly 7.4.0 (commit d7a04203 modified) built on 2023-02-18 with Qt 5.15.2, Windows SDK, MSVC 193431942, Crashpad Running on Windows 10 Version 2009, kernel: 10.0.22621

### LINKED ISSUE #6345: User-based nicknames

3r01:
Fixes #6345.
Fixes #4398.

Currently nicknames match on usernames rather than user IDs, so they can stop applying when a user changes their name. This PR adds nicknames based on twitch User IDs.

You can manage a user's nickname through a new `Nickname` button in the usercard, or through the existing `Nicknames` settings page.

When adding a nickname through settings, you can look up the target account by username or User ID. Either way, the saved entry uses the ID.

Existing username and regex entries keep their current behaviour. The settings page distinguishes between the three types as `Twitch account`, `Username (legacy)`, `Regex`.

I also added highlighting and a warning for duplicate or overlapping entries in the `Nicknames` settings page (#4398), though this does not cover regex entries because it is impractical to do that.

I have separated the changes into commits to make them easier to review, but I can split them into separate PRs if that is preferable.

https://github.com/user-attachments/assets/617e4b90-9427-4cfa-b082-62ff20b440a1


Mm2PL:
We have a system for attaching data to accounts, it's called UserData. I think nicknames should live there even if they are exposed in the current UI and presented similarly to existing ones.

3r01:
> We have a system for attaching data to accounts, it's called UserData. I think nicknames should live there even if they are exposed in the current UI and presented similarly to existing ones.

I can do that. I saw UserData but it seemed awkward to keep the list order synced across two places so I didn't use it, but I will take another look.

3r01:
It uses UserData now. I couldn't find a good way to keep the list order in the settings page when the data is stored in two places like this. It just becomes too brittle because they aren't serialised as part of the same transaction. So, instead, all legacy and regex entries can still be reordered among themselves but always come above the new account-based entries. That way you can override account-based entries with regex if you want. They're still presented all in the same list.

PR #7302, opened 2026-09-17
fix: correct arguments for multi-word custom commands

3r01:
In custom commands with multi-word triggers, numbered placeholders can expand to include parts of the trigger as arguments. For example, with the trigger `/spaced trigger`, entering `/spaced trigger one two` expands `{1}` to `trigger` instead of `one`.

The [chatterino wiki documents](https://wiki.chatterino.com/Commands/#using-placeholders) placeholders as capturing the Nth word after the trigger:

> `{1}`, `{2}`, `{3}` and so on can be used to insert the 1st, 2nd, 3rd, ... word after the trigger.

Fixed by excluding the whole trigger from the arguments when expanding placeholders.

Added tests.


pajlada:
I don't know if this is something I want to actually support. I would rather see multi-word custom commands removed instead. Would love to hear people's reasons for having multi-word custom commands.

3r01:
> I don't know if this is something I want to actually support. I would rather see multi-word custom commands removed instead. Would love to hear people's reasons for having multi-word custom commands.

I do not use them myself and I was surprised to see that they were supported in the first place, so I don't want to argue for keeping them if removing is an option.

They were added in b469c2415462c598233c846f56c5ce2349342d0e but I couldn't find any associated discussion or anyone talking about them since.

